PR #7, opened 2015-07-27
Feature/callbacks

alexeyklyukin:
Move the functionality that was provided by the AWS module (tagging the EC2 instances/EBS) into the callbacks, that can perform arbitrary actions on various events in Patroni, i.e. start, stop, restart, but also promote/demote.

Move the aws.py from helpers/ to scripts/, a new location for pluggable functionality.

Update unit-tests accordingly.


helpers/postgresql.py
```diff
@@ -240,6 +242,26 @@ def is_leader(self):
...
+        if not self.callback or cb_name not in self.callback:
+            return False
+        cmd = self.callback[cb_name]
+        if is_leader is None:
+            try:
+                is_leader = self.is_leader(check_only=True)
+            except psycopg2.OperationalError as e:
+                logger.warning("unable to perform {0} action, cannot obtain the cluster role: {1}".format(cb_name, e))
+                return False
+        name = self.scope
```
sli67: Wouldn't "scope" be a better variable name for self.scope?

helpers/postgresql.py
```diff
@@ -253,18 +275,37 @@ def start(self):
         ret = subprocess.call(self._pg_ctl + ['start', '-o', self.server_options()]) == 0
         ret and self.load_replication_slots()
         self.save_configuration_files()
-        if self.on_change_callback:
-            self.on_change_callback('replica' if os.path.exists(self.recovery_conf) else 'master')
+        if ret and 'on_start' in self.callback:
```
sli67: Shouldn't such magic strings better be declared as constants?

alexeyklyukin: We can, although it's the question of whether ON_START is less magical than "on_start" :-)

helpers/postgresql.py
alexeyklyukin: Good point.

PR #8, opened 2015-08-07
Move the functionality of restoring a replica from backup into an external script

alexeyklyukin:
Move the functionality of restoring a replica from backup into an external script. Now one can plugin-in new methods for bringing replicas to life (EBS snapshots, cloning from some external source, etc) without hacking into Patroni.

scripts/restore.py provides former functionality of choosing between WAL-E and pg_basebackup. In order for it to work, the spilo image has to install WAL-E and export environment variables.

### DIFF
--- a/postgres0.yml
+++ b/postgres0.yml
@@ -46,6 +46,7 @@ postgresql:
     env_dir: /home/postgres/etc/wal-e.d/env
     threshold_megabytes: 10240
     threshold_backup_size_percentage: 30
+  restore: /usr/bin/true
   #recovery_conf:
     #restore_command: cp ../wal_archive/%f %p
   parameters:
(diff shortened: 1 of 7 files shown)


postgres0.yml
```diff
@@ -46,6 +46,7 @@ postgresql:
     env_dir: /home/postgres/etc/wal-e.d/env
     threshold_megabytes: 10240
     threshold_backup_size_percentage: 30
+  restore: /usr/bin/true
```
CyberDem0n: We should not reference anything with absolute path...
Because for example in Ubuntu `true` is located just in `/bin`...

tests/test_postgresql.py
```diff
@@ -121,7 +121,8 @@ def set_up(self):
                              'callbacks': {'on_start': '/usr/bin/true', 'on_stop': '/usr/bin/true',
```
CyberDem0n: why not just `true`?

postgres0.yml
alexeyklyukin: Good point, actually, the reason I included the full path is due to usage of abspath before passing the command to subprocess.call. But that abspath is completely unnecessary: if the user wants to pass the command from PATH, we shouldn't block it.

PR #9, opened 2015-08-11
Write some documentation in a dockstring format for AbstractDCS class

CyberDem0n:


helpers/dcs.py
```diff
@@ -74,19 +120,33 @@ def current_leader(self):
...
+        :param connection_string: how this instance can be accessed by other instances
+        :param ttl: ttl for member key, optional parameter. If it is None `~self.member_ttl will be used`
+        :returns: `!True` on success otherwise `!False`
+        """
 
     @abc.abstractmethod
     def take_leader(self):
-        """take_leader"""
+        """This method should create leader key with value = `~self._name` and ttl=`~self.ttl`
+        Since it could be called only on initial cluster bootstrap it could create this key not atomically."""
```
feikesteenbergen: I don't fully understand this sentence, does it mean that this function is only used during bootstrapping? 

could --> should?

PR #10, opened 2015-08-26
python-etcd

CyberDem0n:
* Refactor etcd helper to work via python-etcd module with high-level interface.
* Set FD_CLOEXEC flag on rest api socket in order to prevent its inheritance on fork.


alexeyklyukin:
Looks good to me.

PR #23, opened 2015-09-04
93compat

jberkus:
Makes replication slots optional (controlled by the configuration option "use_slots").  This allows Patroni to be used with 9.3.

### DIFF
--- a/README.md
+++ b/README.md
@@ -67,6 +67,7 @@ For an example file, see `postgres0.yml`.  Below is an explanation of settings:
   * *connect_address*: ip address + port through which Postgres is accessible from other nodes and applications.
   * *data_dir*: file path to initialize and store Postgres data files
   * *maximum_lag_on_failover*: the maximum bytes a follower may lag before it is not eligible become leader
+  * *use_slots*: whether or not to use replication_slots.  Must be False for PostgreSQL 9.3.
   * *pg_hba*: list of lines which should be added to pg_hba.conf
     * *- host all all 0.0.0.0/0 md5*
   * *replication*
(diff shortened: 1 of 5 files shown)


jberkus:
I'm confused.  The check above says it failed, but when I look at the travis report, all 83 tests passed.  What am I missing?

CyberDem0n:
flake8 doesn't like some changes:

$ python setup.py flake8

running flake8
helpers/postgresql.py:46:1: W293 blank line contains whitespace
helpers/postgresql.py:261:121: E501 line too long (121 > 120 characters)
helpers/postgresql.py:367:121: E501 line too long (127 > 120 characters)

The command "python setup.py flake8" exited with 1.

jberkus:
Sigh, really?

Anyway, spacing changes submitted.

jberkus:
What's going on with the in_healthiest_node test?  I didn't touch any of that code.

jberkus:
OK, I fixed one bug: The original code for is_healthiest_node contains '0/0000000', which isn't a valid LSN, but '0/0000000'::pg_lsn seems to work anyway, whereas it doesn't work with pg_xlog_location_diff.  However, the test is still failing for reasons I can't see.  Suggestions?

CyberDem0n:
flake8 might be sounds silly but sometimes it really helps. For example it also reports unused variables.

Don't waste your time. I will fix failing test later. Purpose of unit tests in our case is to go through as much as possible code-paths. But we are not running queries against real databases, tests just mock some methods of psycopg2 and depending on query itself and other parameters return some expected values.

CyberDem0n:
There is one "feature" I'd like to have.
I think `use_slots` option in config file should not be mandatory and by default we must assume that it is turned 'on'. 

- self.use_slots = config['use_slots']
+ self.use_slots = config.get('use_slots', True)


Besides this everything else looks fine.

jberkus:
Ah, I wasn't sure that config_get() worked that way; I'm more used to ConfigObj.  Yeah, that makes sense.

jberkus:
Yah, FYI, all tests pass on a docker image with 9.3 installed ... and use_slots set to yes!

jberkus:
Mostly I get really aggrivated with pep8/flake8 making me reformat my SQL strings in really ridiculous ways.

PR #24, opened 2015-09-04
minor tweaks

radek-senfeld:
* trottelfest checking ttl > 2 * loop_wait
* user configurable .pgpass filepath

### DIFF
--- a/patroni.py
+++ b/patroni.py
@@ -18,6 +18,8 @@ logger = logging.getLogger(__name__)
 class Patroni:
 
     def __init__(self, config):
+        assert config["etcd"]["ttl"] > 2 * config["loop_wait"]
+
         self.nap_time = config['loop_wait']
         self.postgresql = Postgresql(config['postgresql'])
         self.ha = Ha(self.postgresql, self.get_dcs(self.postgresql.name, config))
--- a/patroni.py
+++ b/patroni.py
@@ -18,8 +18,6 @@ logger = logging.getLogger(__name__)
 class Patroni:
 
     def __init__(self, config):
-        assert config["etcd"]["ttl"] > 2 * config["loop_wait"]
-
         self.nap_time = config['loop_wait']
         self.postgresql = Postgresql(config['postgresql'])
         self.ha = Ha(self.postgresql, self.get_dcs(self.postgresql.name, config))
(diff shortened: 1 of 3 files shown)


patroni.py
```diff
@@ -18,6 +18,8 @@
 class Patroni:
 
     def __init__(self, config):
+        assert config["etcd"]["ttl"] > 2 * config["loop_wait"]
```
CyberDem0n: And what about other configuration stores?
Currently we already support ZooKeeper and working on Consul support.

patroni.py
```diff
@@ -88,6 +91,11 @@ def run(self):
             try:
                 if self.ha.state_handler.is_leader():
                     self.ha.cluster and self.ha.state_handler.create_replication_slots(self.ha.cluster)
+
+                    # SkyDNS2 support: publish leader
```
feikesteenbergen: We'd like to keep the code as generic as possible. This could be implemented by the ACTION_ON_ROLE_CHANGE callback (to call a script for example). This does however require some more documentation

scripts/restore.py
```diff
@@ -56,10 +56,16 @@ def replica_fallback_method(self):
 
     def run(self):
         """ creates a new replica using either pg_basebackup or WAL-E """
+
+        logger.warning("Starting database restore ...")
```
feikesteenbergen: logger.info I suppose? Everything is as excpected, therefore not a warning.

patroni.py
radek-senfeld: That was my thinking as well but for our use case callback cannot be used because it's being called just once - on a role change. We need periodic record refresh to keep record's TTL short = faster response when anything goes wrong.

Sorry for that, it wasn't meant to be part of my PR. Still being a git noob, I don't know how to divide future commits from PR I've already made.. :-(

scripts/restore.py
radek-senfeld: You're right.. Fixed.. :-)

patroni.py
radek-senfeld: Good point! Fixed for etcd, no exp with ZK nor Consul..

CyberDem0n:
Synchronous replication is now supported and it's also possible to configure custom location for pgpass.
For SkyDNS you should use callbacks.

Issue #18, opened 2015-09-02
support 9.3 by allowing disabling replication slots


jberkus:
... this should be possible.

jberkus:
Taken care of with PR #23

CyberDem0n:
Oh!
We can actually make this feature more smart and even make patroni compatible with postgresql  < 9.3

1) We should query major version in Postgresql.__init__
2) If version < 9.4 forcefully set `use_slots` to False
3) Get rid from `pg_xlog_location_diff` which exists only from 9.3 and use function `utils.lsn_to_bytes` (https://github.com/zalando/patroni/blob/master/helpers/utils.py#L45) instead.
4) Profit

CyberDem0n:
Hmm, for postgres version < 9.3 formula is a little bit different...
Ok, will move this method into Postgresql and make it version-specific.

CyberDem0n:
`2` wont work for slaves...
We have to know in advance that version < 9.4 in order to create proper recovery.conf but without starting cluster it's not really possible to determine version.

But at least everything else is doable.

https://github.com/zalando/patroni/pull/25

jberkus:
We don't care about supporting < 9.3  Here's why:

1. Before 9.3, you couldn't remaster over streaming replication, making failover/failback and master change *much* more complicated.

2. 9.3 was 2 years ago, and is the supported version in current stable Ubuntu, Debian, and Red Hat.

jberkus:
For that reason, and others: -1 from me.  

I feel that this complicates the code in order to support versions of PostgresQL which we have other problems supporting.

CyberDem0n:
Ok, dropped unneeded  code.

PR #25, opened 2015-09-04
Feature/lsn to bytes

CyberDem0n:
Use different algorithm when calculating amount of bytes written to xlog depending on server version

### DIFF
--- a/helpers/postgresql.py
+++ b/helpers/postgresql.py
@@ -35,7 +35,6 @@ def parseurl(url):
 
 
 class Postgresql:
-    _SERVER_VERSION = 90400
 
     def __init__(self, config):
         self.config = config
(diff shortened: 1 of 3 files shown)


helpers/postgresql.py
```diff
@@ -39,6 +35,7 @@ def parseurl(url):
 
 
 class Postgresql:
+    _SERVER_VERSION = 90400
```
feikesteenbergen: In issue #18 it is already mentioned that we might be able to support 9.3, can we make this a non-magic variable?

feikesteenbergen:
+1

helpers/postgresql.py
CyberDem0n: By default we assume that we are running on 9.4
Postgres version determine default behaviour of lsn_to_bytes and bytes_to_lsn methods.

When connection to instance is established this value would be overwritten with actual value (see line 85)

jberkus:
Why are we calculating LSN to Bytes in the python code at all?  Why not just pull it from the database when we pull the LSN?

jberkus:
If that wasn't clear: -1 from me.

Supporting versions earlier than 9.3 is going to be difficult for several reasons, not the least of which is the lack of remastering support (and in 9.0, the lack of any kind of timeline switching).  Given that adding the LSN function isn't *sufficient* to support 9.1 and 9.2, why clutter up the code?

PR #30, opened 2015-09-08
Make sure initialize flag is reset on failure.

alexeyklyukin:
Cleanup the initialize flag if the initializing node fails
to bootstrap its PostgreSQL database.

Rename dcs.race to initialize, since we only call it for the
initialize flag. Factored out PostgreSQL bootstrapping code
into a separate function.


patroni.py
```diff
@@ -52,18 +57,18 @@ def initialize(self):
...
+                except:
+                    # bail out and clean the initialize flag.
+                    self.cleanup_on_failed_initialization()
+                    raise
                 self.ha.dcs.take_leader()
-                self.postgresql.start()
-                self.postgresql.create_replication_user()
-                self.postgresql.create_connection_users()
             else:
                 while True:
```
feikesteenbergen: What if the initialisation of the master fails? It will loop forever

tests/test_zookeeper.py
```diff
@@ -141,7 +141,7 @@ def test_get_cluster(self):
         self.zk.delete_leader()
 
     def test_race(self):
```
feikesteenbergen: rename?

patroni/__init__.py
```diff
@@ -50,19 +55,20 @@ def initialize(self):
...
         if self.postgresql.data_directory_empty():
-            # racing to initialize
-            if self.ha.dcs.race('/initialize'):
-                self.postgresql.initialize()
-                self.ha.dcs.take_leader()
-                self.postgresql.start()
-                self.postgresql.create_replication_user()
-                self.postgresql.create_connection_users()
-            else:
-                while True:
```
CyberDem0n: Hmm,
why we are not waiting for leader anymore?

And what if initialization failed?

patroni/zookeeper.py
```diff
@@ -220,6 +220,11 @@ def delete_leader(self):
         if isinstance(self.leader, Leader) and self.leader.name == self._name:
             self.client.delete(self.client_path('/leader'))
 
+    def cancel_initialization(self):
+        node = self.get_node(self.initialize_key)
```
CyberDem0n: get_node returns either tuple or None if the node doesn't exists
First element of tuple contains value of the node
Second - znode object

patroni/zookeeper.py
```diff
@@ -220,6 +220,11 @@ def delete_leader(self):
...
+        if node and node == self._name:
+            self.client.delete(self.client_path(self.initialize_key))
```
CyberDem0n: If you want to really be sure that you are removing the right node, you should also pass version=znode.mzxid as an argument

patroni/__init__.py
alexeyklyukin: I don't see us waiting for the leader in the code fragment you've highlighted. What has changed is that now we try to bootstrap the PostgreSQL cluster before taking the leader: doing it the other way around provides no benefits, but bears an additional responsibility to remove the leader flag during cleanup_on_failed_initialization.

Actually, I see one problem here: if the initialiser node suddenly dies after initidb (say, it cannot not create the users), and then restarts - it will think it was fully initialised and will try to start as a normal member that was started after a stop, potentially acquiring the master role (which would screw up the whole cluster), or a replica, in which case it won't be able to catchup with the master, because it was not initialised from it.

alexeyklyukin: I think the solution is to remove the data directory during the cleanup_on_failed_initialization.

patroni/dcs.py
```diff
@@ -165,7 +165,7 @@ def take_leader(self):
         overwriting the key if necessary."""
 
     @abc.abstractmethod
-    def race(self, path):
+    def initialize(self):
         """Race for cluster initialization.
         :param path: usually this is just '/initialize'
```
CyberDem0n: initialize method does not have any arguments

alexeyklyukin: changed, thank you.

PR #34, opened 2015-09-09
Define initialize, leader, optime and members string constansts in Ab…

CyberDem0n:
…stractDCS

Also define following properties:
* initialize_path
* members_path
* member_path
* leader_path
* leader_optime_path

And replace all occurrences of these strings or client_path calls in a
etcd and zookeeper implementations with given constants and properties.


feikesteenbergen:
+1

PR #50, opened 2015-09-23
add support for pg_rewind.

alexeyklyukin:


patroni/postgresql.py
```diff
@@ -69,6 +70,23 @@ def __init__(self, config):
         self._cursor_holder = None
         self.members = []  # list of already existing replication slots
         self.retry = Retry(max_tries=-1, deadline=10, max_delay=1, retry_exceptions=PostgresConnectionException)
+        self.init_pg_rewind()
+
+    def init_pg_rewind(self):
+        try:
+            self._pg_rewind_present = ('username' in self._pg_rewind and
```
alexeyklyukin: We should explore the possibility of checking if pg_rewind failed because this options are not set and distinguish it from other failures. Rationale: those options might not be set in yam config, but in postgresql.conf, and we cannot check it at this stage.

patroni/postgresql.py
```diff
@@ -120,16 +138,19 @@ def initialize(self):
     def delete_trigger_file(self):
         os.path.exists(self.trigger_file) and os.unlink(self.trigger_file)
 
-    def sync_from_leader(self, leader):
-        r = parseurl(leader.conn_url)
-
+    def write_pgpass(self, record, append=False):
         pgpass = 'pgpass'
-        with open(pgpass, 'w') as f:
+        with open(pgpass, 'w' if not append else 'a') as f:
```
alexeyklyukin: Append is not necessary, as except for pg_rewind this file is only used by pg_basebackup, and not by streaming replication (which gets the pass from recovery.conf).

patroni/postgresql.py
```diff
@@ -296,12 +317,40 @@ def write_recovery_conf(self, leader):
...
-            run_callback and self.call_nowait(ACTION_ON_ROLE_CHANGE)
+            change_role = self.role == 'master'
+
+            if leader and change_role and self._pg_rewind_present:
+                self.stop()
+                if self.pg_rewind(leader):
+                    ret = self.start()
+                else:
+                    ret = False
+                    self.remove_data_directory()
```
alexeyklyukin: Removing the old data directory might be dangerous, because the former master might have some data not propagated to the new one (in the case of the split brain). Consider moving the data, or have a configuration option that defines what to do.

Issue #48, opened 2015-09-22
Add tags to replicas to support many replicas


jberkus:
Replicas should have the following attributes available.  This would be added as tags within the replica description, both in the API and in the SIS.  If the tags are absent, default behavior remains the same; in the config, the tags below are set to True or False:

**nofailover**: this replica should not be a candidate for automatic failover.  Used for ephemeral replicas, and multi-replica setups where there are a few designated failover targets.   If the tag is missing, the replica is assumed to be a failover target.

**noloadbalance**: this replica should not be part of any load-balancing, if HAProxy autoconfig is being used.  

**clonefrom**:  take basebackups from this replica *instead of* from the master.  If multiple CloneFrom replicals are available, one is randomly chosen for new basebackups.  If cloning using WALE or other archival backup, this tag is ignored.

**replicatefrom**: not true or false, but the IP address/hostname of another replica.  Used to support cascading replication.

kmoppel:
Also an "environment" tag should be there to distinguishing between release/live environments etc.
Even better would be the possibility to add "custom" user defined tags, with no meaning to patroni. Those tags should be persisted and easily inspectable via (future) patroni-cli for example.

valgog:
@kmoppel you are now introducing the idea of tags for the whole cluster. Until now the tags suggested by Josh are tagging instances in the cluster.

jberkus:
The only global tag I'm proposing to add will be:

**disabled**: if this tag is set for the cluster, no auto-failover will take place.  This tag is meant for disabling auto-failover during manual failover and similar expected downtime.  It should *always* be set with a TTL.

jberkus:
I don't think I agree with the idea of an "environment" tag.  In what way is that not simply part of the "scope"?

As for user-defined custom tags, nothing keeps patroni-cli from adding any tags it wants.  I haven't conceived of any I need yet though.

kmoppel:
@valgog yes, exactly for the whole cluster. @jberkus ok yes, the "environment" concept could be an user defined concept (tag) also - keeping mandatory parts from operational requirements only might be a good line to keep as patroni grows in features.
But custom tags in general would be very useful in my opinion - marking environments, applications using some cluster, criticality level, being part of a sharded cluster etc, will definitely help large setups.

jberkus:
[...]

PR #54, opened 2015-09-25
Patroni needs a plugin system

jinty:
Basically, there are a number of places where it would be useful to have a plugin system more powerful than "run external script". For example, the DCS class should be a plugin and preferably in a separate package to simplify dependencies. In my specific case I have 2 examples of things I would like to do:

 * Set the name of the node automatically (i.e. from AWS EC2 metadata) so I can have identical config files on each node
 * Add extra metadata to the data the node stores in the DCS. Specifically I want to add the EC2 availability zone so I can set HAProxy weights to have appservers pefer database servers in their availability zone.

I have made this pull request which contains the basic machinery I think is appropriate. Then I have a proof of concept branch (https://github.com/jinty/patroni/tree/plugins-example) which implements a default node name event and plugin for that event. See:
   https://github.com/jinty/patroni/commit/c35d89d35c5337067bdf915d355ac2d6da37d6e0
   https://github.com/jinty/patroni/commit/9ea29f1920032ce5592007fea93c0e257a309b69

### DIFF
--- a/patroni/__init__.py
+++ b/patroni/__init__.py
@@ -4,6 +4,7 @@ import sys
 import time
 import yaml
 
+from patroni.plugin import get_plugins
 from patroni.api import RestApiServer
 from patroni.etcd import Etcd
 from patroni.ha import Ha
@@ -17,6 +18,7 @@ logger = logging.getLogger(__name__)
 class Patroni:
 
     def __init__(self, config):
+        self.plugins = get_plugins(config, self)
         self.nap_time = config['loop_wait']
         self.postgresql = Postgresql(config['postgresql'])
         self.ha = Ha(self.postgresql, self.get_dcs(self.postgresql.name, config))
--- a/patroni/plugin.py
+++ b/patroni/plugin.py
@@ -1,0 +1,126 @@
+"""Plugin Machinery
+
+Plugins are objects which subscribe to events. An object subscribes to an event
+by implementing a method with the event name. There is a single namespace of event
+names, plugins can subscribe to as many events as they choose.
+
+Events come in 2 flavours:
+
+    * multiple: Multiple plugins can subscribe to a multiple event. They will all be run
+                in the order of their declaration in the config file. The return value
+                may or may not be used.
+    * single: Only one of these subscribers is allowed at a time.
+
+Events must document the arguments they will be called with.
+
+Registration and Usage
+----------------------
+
+Plugins are registered by adding them to the entry points in setup.py. e.g.
+
+    entry_points={
+        'patroni.plugin': [
+            'myplugin = mymodule:MyPlugin',
+        ],
+    },
+
+Plugins are only actually used if they appear in the configuration file in an
+ordered list of plugins to use:
+
+    plugins:
+        - myplugin1
+        - myplugin2
+
+Each plugin should provide it's own documentation about what configuration it
+needs.
+"""
+import logging
+from pkg_resources import iter_entry_points
+
+EVENTS = (
+        {'name': 'postgresql_name', 'type': 'single', 'required': False},
+        )
+
+def get_plugins(config, app):
+    plugins = load(config)
+    plugins = configure(plugins, config, app)
+    return get_event_handler(plugins, EVENTS)
+
+def load(config):
+    """Gets the plugin factories from the config file.
+
+    Plugins are sorted by the order they are specified in the config file.
+    """
+    plugin_names = [c.strip() for c in config.get('plugins', [])]
+    seen = set([])
+    for i in plugin_names:
+        if i in seen:
+            raise AssertionError('Duplicate plugin: {}'.format(i))
+        seen.add(i)
+    plugins = {}
+    available_plugins = []
+    for i in iter_entry_points('patroni.plugin'):
+        available_plugins.append(i.name)
+        if i.name in plugin_names:
+            if i.name in plugins:
+                raise Exception('Duplicate plugins for name {}, one was: {}'.format(i.name, plugin_factory))
+            plugin_factory = i.load(require=False) #EEK never auto install ANYTHING
+            plugins[i.name] = plugin_factory
+    not_seen = set(plugin_names) - set(plugins)
+    if not_seen:
+        raise Exception('Plugins were configured in the config file, but I could NOT find them: {}\n'
+                'Available plugins: {}'.format(not_seen, available_plugins))
+    return [(i, plugins[i]) for i in plugin_names] # set order to what is specified in config file
+
+def configure(plugins, *args, **kw):
+    """Setup all plugins by calling them with the passed arguments"""
+    c_plugins = []
+    for name, plugin_factory in plugins:
+        plugin = plugin_factory(name, *args, **kw)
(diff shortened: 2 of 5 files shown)


CyberDem0n:
You are absolutely right about implementing support of different  distributed configuration stores as a plugins, I've also thought about it.

Regarding your problems:
> Set the name of the node automatically (i.e. from AWS EC2 metadata)

Actually it's really easy to do even now. We are simply generating config.yml before starting up patroni:  https://github.com/zalando/spilo/blob/master/postgres-appliance/postgres_ha.sh
May be you a right about having some interface for setting up node name (for example default one can just use hostname). But from another hand it will make application harder to debug if you messed up with the configuration.... What about possibility to set (override) node name via ENV variable or command line parameter?

> Add extra metadata to the data the node stores in the DCS

Right now it's kind of tricky... In the beginning we decided to keep backward compatibility with the Governor (it is keeping only PostgreSQL connection url in etcd). So we extended this connection url with one parameter, application_name=http://rest.api.host:port/patroni. Psycopg2 (actually libpq) accepts such url without a problem, but currently the only possible way to add some extra stuff there is either extending application_name or for example by abusing "tty" parameter. We can also solve this problem radically: brake backward compatibility and start using json there...

> Specifically I want to add the EC2 availability zone so I can set HAProxy weights to have appservers pefer database servers in their availability zone.

Hmm, do you want to load-balance read-only queries as well? Updates can be executed only on master which would be running in one AZ.

jinty:
On Mon, Sep 28, 2015 at 01:04:42PM -0700, Alexander Kukushkin wrote:
> You are absolutely right about implementing support of different  distributed configuration stores as a plugins, I've also thought about it.

Have you thought about what kind of plugin system you want? I've found
entry points to be very flexible as a registration system.

> Regarding your problems:
> > Set the name of the node automatically (i.e. from AWS EC2 metadata)
> 
> Actually it's really easy to do even now. We are simply generating config.yml before starting up patroni:  https://github.com/zalando/spilo/blob/master/postgres-appliance/postgres_ha.sh
> May be you a right about having some interface for setting up node name (for example default one can just use hostname). But from another hand it will make application harder to debug if you messed up with the configuration.... What about possibility to set (override) node name via ENV variable or command line parameter?

I am aiming to use AWS auto scaling, so the only time I can write the
config.yaml or environment variable file is at boot. But on ec2 there's
a sane default that's almost always usable (i.e. the instance-id), I
thought some kind of ec2-specific plugin would be a good idea. There
doesn't seem to be any good way of detecting if code is running on ec2.

However, I'm also confused at the need for a name. Surely the
hostname:port combination is unique enough to use as an ID?

> > Add extra metadata to the data the node stores in the DCS
> 
> Right now it's kind of tricky... In the beginning we decided to keep backward compatibility with the Governor (it is keeping only PostgreSQL connection url in etcd). So we extended this connection url with one parameter, application_name=http://rest.api.host:port/patroni. Psycopg2 (actually libpq) accepts such url without a problem, but currently the only possible way to add some extra stuff there is either extending application_name or for example by abusing "tty" parameter. We can also solve this problem radically: brake backward compatibility and start using json there...
> 
> > Specifically I want to add the EC2 availability zone so I can set HAProxy weights to have appservers pefer database servers in their availability zone.
> 
> Hmm, do you want to load-balance read-only queries as well? Updates can be executed only on master which would be running in one AZ.

Yep, I want to load-balance read-only queries. I already do so in my
current setup, albeit with manual failover.

-- 
Brian Sutherland

CyberDem0n:
>Have you thought about what kind of plugin system you want? I've found entry points to be very flexible as a registration system.

Not yet. We don't have that much configuration stores. And frankly speaking API is not stable yet.

> There doesn't seem to be any good way of detecting if code is running on ec2.

We actually did it on purpose. I think Patroni itself should not care about the place where it is running.
Today we are running it on AWS but tomorrow it might be google compute engine for example. And actually everybody can easily adapt it. The only you have to do - create startup script which will generate proper configuration file. We are already doing it on AWS. We are also running Patroni + PostgreSQL in an auto scaling group. And even more, we are forced to run everything inside docker container. So we created simple bash script which generates config.yml from some environment variables plus http://169.254.169.254/latest/meta-data and only then starting up Patroni. Coding on bash is easier than on python for most people.

> However, I'm also confused at the need for a name. Surely the hostname:port combination is unique enough to use as an ID?

Well, it's unique and even more, within one cluster usually just hostname is unique enough (probably nobody wants to run HA cluster on the same host).


> Yep, I want to load-balance read-only queries. I already do so in my current setup, albeit with manual failover.

Aha, basically you want to be able to write arbitrary information into configuration store. But again in most cases this is just one time operation (on start).
For me it's really looks similar to this issue: https://github.com/zalando/patroni/issues/48
We can write whatever tags we want. And some of them could have special meaning (like in issues 48).

Thoughts?

jinty:
On Tue, Sep 29, 2015 at 11:29:51AM -0700, Alexander Kukushkin wrote:
> >Have you thought about what kind of plugin system you want? I've found entry points to be very flexible as a registration system.
> 
> Not yet. We don't have that much configuration stores. And frankly speaking API is not stable yet.

Ok

> > There doesn't seem to be any good way of detecting if code is running on ec2.

Hmm, I actually meant in general, not just within patroni.

> We actually did it on purpose. I think Patroni itself should not care about the place where it is running.
> Today we are running it on AWS but tomorrow it might be google compute engine for example. And actually everybody can easily adapt it. The only you have to do - create startup script which will generate proper configuration file. We are already doing it on AWS. We are also running Patroni + PostgreSQL in an auto scaling group. And even more, we are forced to run everything inside docker container. So we created simple bash script which generates config.yml from some environment variables plus http://169.254.169.254/latest/meta-data and only then starting up Patroni. Coding on bash is easier than on python for most people.

Ok

> > However, I'm also confused at the need for a name. Surely the hostname:port combination is unique enough to use as an ID?
> 
> Well, it's unique and even more, within one cluster usually just hostname is unique enough (probably nobody wants to run HA cluster on the same host).
> 
> 
> > Yep, I want to load-balance read-only queries. I already do so in my current setup, albeit with manual failover.
> 
> Aha, basically you want to be able to write arbitrary information into configuration store. But again in most cases this is just one time operation (on start).

For my case, it is a one-time operation. For other information, like
server load, it's probably better to communicate that to HAProxy via an
agent-check. 

But there is one usecase where you do want volatile information: If you
want to build some kind of admin UI which can display the current status
of all hosts just by looking at the DCS. Then it becomes interesting to
put things like replication lag in the information for each host.

I've also been developing something like patroni, you can see the layout
in the DCS here:
    https://github.com/jinty/zgres/blob/master/README.md

> For me it's really looks similar to this issue: https://github.com/zalando/patroni/issues/48
> We can write whatever tags we want. And some of them could have special meaning (like in issues 48).
> 
> Thoughts?

Yep, it is very similar.

Given that the way to use patroni is to generate the config file at boot
it's a good idea to put these tags in the config file.

Perhaps you want to force all tags patroni doesn't use to have a period
(.) in them. That way in future, when you introduce new special tags, you
can be certain of backwards compat.

-- 
Brian Sutherland

PR #59, opened 2015-10-12
Feature/pg rewind unclean shutdown

alexeyklyukin:
Add support for pg_rewind.

Make sure pg_rewind can be called in both the case when the former master lost the connect (and hence, was shut down cleanly), and in the case when the master crashed. In the latter case the single-user mode is employed to start the postgresql and shut it down cleanly in order to have a proper shut down status in pg_controldata.

Add new config parameter section, pg_rewind, which must contain the 'username' and 'password' parameters of a database superuser, in order to run pg_rewind. If this section is missing or not pg_rewind is present, the cluster will just try to restart itself.

Add some aux functions like controldata and read_postmaster_opts in postgresql.py, they might be useful in implementing other features (i.e. checking if postgresql is running or making sure the new node belongs to the same cluster.

### DIFF
--- a/patroni/ha.py
+++ b/patroni/ha.py
@@ -96,7 +96,8 @@ class Ha:
         # try to see if we are the former master that crashed. If so - we likely need to run pg_rewind
         # in order to join the former standby being promoted.
         pg_controldata = self.state_handler.controldata()
-        if not has_lock and pg_controldata.get('Database cluster state', '') == 'in production':  # crashed master
+        if not has_lock and pg_controldata and\
+                pg_controldata.get('Database cluster state', '') == 'in production':  # crashed master
             self.state_handler.require_rewind()
 
         # XXX: follow the leader calls stop, which might take quite some time.
--- a/patroni/ha.py
+++ b/patroni/ha.py
@@ -66,8 +66,15 @@ class Ha:
         if self.state_handler.is_healthy():
             return False
         has_lock = self.has_lock()
-        self.state_handler.write_recovery_conf(None if has_lock else self.cluster.leader)
-        self.state_handler.start()
+
+        # try to see if we are the former master that crashed. If so - we likely need to run pg_rewind
+        # in order to join the former standby being promoted.
+        pg_controldata = self.state_handler.controldata()
+        if not has_lock and pg_controldata.get('Database cluster state', '') == 'in production':  # crashed master
+            self.state_handler.require_rewind()
+
+        # XXX: should we call ha.follow_the_leader here instead?
+        ret = self.state_handler.follow_the_leader(None if has_lock else self.cluster.leader, recovery=True)
         if has_lock:
             logger.info('started as readonly because i had the session lock')
             self.load_cluster_from_dcs()
--- a/patroni/postgresql.py
+++ b/patroni/postgresql.py
@@ -343,7 +343,7 @@ recovery_target_timeline = 'latest'
                     ret = self.start()
                 else:
                     ret = False
-                    self.move_data_directory()
+                    self.remove_data_directory()
                     logger.error("unable to rewind the former leader")
             else:
                 ret = self.restart()
--- a/patroni/postgresql.py
+++ b/patroni/postgresql.py
@@ -135,14 +135,17 @@ class Postgresql:
     def delete_trigger_file(self):
         os.path.exists(self.trigger_file) and os.unlink(self.trigger_file)
 
+    def write_pgpass(self, record, append=False):
+        pgpass = 'pgpass'
+        with open(pgpass, 'w' if not append else 'a') as f:
+            os.fchmod(f.fileno(), 0o600)
+            f.write('{host}:{port}:*:{user}:{password}\n'.format(**record))
+        return pgpass
+
     def sync_from_leader(self, leader):
         r = parseurl(leader.conn_url)
 
-        pgpass = 'pgpass'
-        with open(pgpass, 'w') as f:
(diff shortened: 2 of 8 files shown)


patroni/postgresql.py
```diff
@@ -79,6 +81,33 @@ def __init__(self, config):
...
+            we have either wal_log_hints or checksums turned on
+        """
+        # low-hanging fruit: check if pg_rewind configuration is there
+        if not self.pg_rewind or\
+           not (self.pg_rewind.get('username', '') and self.pg_rewind.get('password', '')):
+            return False
+
+        cmd = ['pg_rewind', '--help']
+        try:
+            ret = subprocess.call(cmd, stdout=open(os.devnull, 'w'), stderr=subprocess.STDOUT)
```
CyberDem0n: stdout=open(os.devnull, 'w')
Hmm, probably this code will lead to FD leak.

Isn't it better to use `subprocess.check_output([], stderr=subprocess.STDOUT)` instead.
It will raise CalledProcessError if the exit code was non-zero.

patroni/postgresql.py
```diff
@@ -135,16 +164,19 @@ def initialize(self):
     def delete_trigger_file(self):
         os.path.exists(self.trigger_file) and os.unlink(self.trigger_file)
 
-    def sync_from_leader(self, leader):
-        r = parseurl(leader.conn_url)
-
+    def write_pgpass(self, record, append=False):
         pgpass = 'pgpass'
-        with open(pgpass, 'w') as f:
+        with open(pgpass, 'w' if not append else 'a') as f:
```
CyberDem0n: pgpass is only used either by pg_basebackup or by pg_rewind. There is no need to append we can simply rewrite it every time.

patroni/postgresql.py
alexeyklyukin: I think the GC should close the FD when the function variables goes out of scope, namely, when the function terminates...

patroni/postgresql.py
alexeyklyukin: Good point, will fix.

patroni/ha.py
```diff
@@ -92,8 +92,18 @@ def bootstrap(self):
 
     def recover(self):
         has_lock = self.has_lock()
-        self.state_handler.write_recovery_conf(None if has_lock else self.cluster.leader)
-        if not self.state_handler.start():
+
+        # try to see if we are the former master that crashed. If so - we likely need to run pg_rewind
+        # in order to join the former standby being promoted.
+        pg_controldata = self.state_handler.controldata()
```
CyberDem0n: `controldata` returns None in some cases which does not have get method.

patroni/postgresql.py
```diff
@@ -326,12 +358,120 @@ def write_recovery_conf(self, leader):
...
+            return result
+
+    def single_user_mode(self, command=None, options={}):
+        """ run a given command in a single-user mode. If the command is empty - then just start and stop """
+        cmd = ['postgres', '--single', '-D', self.data_dir]
+        for opt in sorted(options):
+            cmd.extend(['-c', '{0}={1}'.format(opt, options[opt])])
+        # need a database name to connect
+        cmd.append('postgres')
+        p = subprocess.Popen(cmd, stdin=subprocess.PIPE, stdout=open(os.devnull, 'w'), stderr=subprocess.STDOUT)
```
CyberDem0n: The same problem. `open(os.devnull, 'w')` will return fd which would never be closed...

patroni/postgresql.py
```diff
@@ -326,12 +358,120 @@ def write_recovery_conf(self, leader):
...
+            return p.wait()
+        return 1
+
+    def cleanup_archive_status(self):
+        status_dir = os.path.join(self.data_dir, 'pg_xlog', 'archive_status')
+        if os.path.isdir(status_dir):
+            for f in os.listdir(status_dir):
+                path = os.path.join(status_dir, f)
+                try:
+                    if os.path.isfile(path):
```
CyberDem0n: `isfile` method follows symlinks.
So it will return True if the symlink points to the real file.

Basically you need to change order of if and elif blocks

patroni/postgresql.py
```diff
@@ -354,7 +494,9 @@ def promote(self):
             return True
         ret = subprocess.call(self._pg_ctl + ['promote']) == 0
         if ret:
-            self.set_role('master')
+            self._role = 'master'
```
CyberDem0n: should call `self.set_role('master')`

patroni/postgresql.py
```diff
@@ -326,12 +358,120 @@ def write_recovery_conf(self, leader):
...
+            if not leader or not self._need_rewind:  # do not rewind until the leader becomes available
+                ret = self.restart()
+            else:  # we have a leader and need to rewind
+                if self.is_running():
+                    self.stop()
+                # at present, pg_rewind only runs when the cluster is shut down cleanly
+                # and not shutdown in recovery. We have to remove the recovery.conf if present
+                # and start/shutdown in a single user mode to emulate this.
+                # XXX: if recovery.conf is linked, it will be written anew as a normal file.
+                if os.path.isfile(self.recovery_conf):
```
CyberDem0n: `isfile` method follows symlinks and fill fail if recovery.conf is a symlink

patroni/ha.py
alexeyklyukin: Fixed, pg_controldata always returns a hash (might be empty), and the invokers check if the result is non-False before doing anything with it.

PR #68, opened 2015-10-16
Fixed format for option httpchk, fixed cursor object

hryamzik:


patroni/api.py
```diff
@@ -202,13 +202,15 @@ def __init__(self, patroni, config):
     def query(self, sql, *params):
         cursor = None
         try:
-            with self.patroni.postgresql.connection().cursor() as cursor:
-                cursor.execute(sql, params)
-                return [r for r in cursor]
+            cursor = self.patroni.postgresql.connection().cursor()
```
alexeyklyukin: The with support for the cursor was introduced in psycopg2 2.5, which was released 2.5 years ago, so it's not an unreasonable requirement (good point is that we should actually formalise it in requirements).

CyberDem0n: But actually your code has some bugs...
finally block would be executed in any case, for example when postgresql.connection() raised exception.
In this case cursor would be None and calling close method on None will raise AttributeError exception....

Current implementation doesn't work for you just because you are using outdated version of psycopg2.
Yeah, it's kind of our fault that we didn't specified minimal version in a requirements.txt and we will fix it soon.
And there are number of other reasons not to have outdated version of psycopg2 driver...

PR #69, opened 2015-10-16
Feature/clusterid

alexeyklyukin:
Add cluster id to the DCS cluster and make sure the new node may join only if its ID matches the one for the cluster.

### DIFF
--- a/patroni/dcs.py
+++ b/patroni/dcs.py
@@ -240,8 +240,11 @@ class AbstractDCS:
         overwriting the key if necessary."""
 
     @abc.abstractmethod
-    def initialize(self):
+    def initialize(self, create_new=True, sysid=None):
         """Race for cluster initialization.
+
+        :param create_new: False if the key should already exist (in the case we are setting the system_id)
+        :param sysid: PostgreSQL cluster system identifier, if specified, is written to the key
         :returns: `!True` if key has been created successfully.
 
         this method should create atomically initialize key and return `!True`
--- a/patroni/dcs.py
+++ b/patroni/dcs.py
@@ -240,7 +240,7 @@ class AbstractDCS:
         overwriting the key if necessary."""
 
     @abc.abstractmethod
-    def initialize(self, create_new=True, sysid=None):
+    def initialize(self, create_new=True, sysid=""):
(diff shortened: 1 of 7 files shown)


patroni/postgresql.py
```diff
@@ -105,6 +106,13 @@ def can_rewind(self):
                 data.get('Data page checksum version', '0') != '0'
         return False
 
+    @property
+    def sysid(self):
+        if not self._sysid:
+            data = self.controldata()
```
CyberDem0n: self.controldata() returns empty dictionary under python3
this happens because subprocess.check_output returns bytestring instead of normal string.
solution - replace line #403 with:
data = data.decode().splitlines()  # yeah, decode is very important. it will covert bytestring into string

patroni/postgresql.py
```diff
@@ -105,6 +106,13 @@ def can_rewind(self):
...
+            self._sysid = data and data.get('Database system identifier', None)
```
CyberDem0n: self._sysid would be assigned to {} (empty dict) if the data is an empty dict...

patroni/zookeeper.py
```diff
@@ -139,7 +139,7 @@ def _inner_load_cluster(self):
             self.fetch_cluster = True
 
         # get initialize flag
-        initialize = self._INITIALIZE in nodes
+        initialize = self.get_node(self._INITIALIZE)[0] if self._INITIALIZE in nodes else None
```
CyberDem0n: self.get_node(self.initialize_path)

patroni/dcs.py
```diff
@@ -240,8 +240,11 @@ def take_leader(self):
         overwriting the key if necessary."""
 
     @abc.abstractmethod
-    def initialize(self):
+    def initialize(self, create_new=True, sysid=None):
```
CyberDem0n: Does it make seance to set default value of sysid argument to an empty string?
This will help to get rid from following unnecessary checks:
1) sysid or "" in the etcd.py
2) sysid if sysid else "" in the zokeeper.py
3) sysid.encode("utf-8") if sysid else ""  in the zokeeper.py

PR #71, opened 2015-10-20
Feature/configurable pgpass

alexeyklyukin:
Make pgpass location configurable.

One can use pgpass configuration parameter in the postgres
subsection of Patroni. By default pgpass is written in ~/.
Mock actual writes to pgpass in the tests.


drnic:
Cool, thx

Issue #74, opened 2015-10-21
Patroni fails when Cluster name is only numeric


feikesteenbergen:
Scenario: Created Spilo with version 18

```plain
Oct 21 09:21:11 ip-172-31-173-176 docker/88c787eafc06[785]: TypeError: cannot concatenate 'str' and 'int' objects
Oct 21 09:21:36 ip-172-31-173-176 docker/88c787eafc06[785]: Traceback (most recent call last):
Oct 21 09:21:36 ip-172-31-173-176 docker/88c787eafc06[785]:   File "/home/postgres/patroni/patroni.py", line 6, in <module>
Oct 21 09:21:36 ip-172-31-173-176 docker/88c787eafc06[785]:     main()
Oct 21 09:21:36 ip-172-31-173-176 docker/88c787eafc06[785]:   File "/home/postgres/patroni/patroni/__init__.py", line 67, in main
Oct 21 09:21:36 ip-172-31-173-176 docker/88c787eafc06[785]:     patroni = Patroni(config)
Oct 21 09:21:36 ip-172-31-173-176 docker/88c787eafc06[785]:   File "/home/postgres/patroni/patroni/__init__.py", line 22, in __init__
Oct 21 09:21:36 ip-172-31-173-176 docker/88c787eafc06[785]:     self.dcs = self.get_dcs(self.postgresql.name, config)
Oct 21 09:21:36 ip-172-31-173-176 docker/88c787eafc06[785]:   File "/home/postgres/patroni/patroni/__init__.py", line 31, in get_dcs
Oct 21 09:21:36 ip-172-31-173-176 docker/88c787eafc06[785]:     return Etcd(name, config['etcd'])
Oct 21 09:21:36 ip-172-31-173-176 docker/88c787eafc06[785]:   File "/home/postgres/patroni/patroni/etcd.py", line 148, in __init__
Oct 21 09:21:36 ip-172-31-173-176 docker/88c787eafc06[785]:     super(Etcd, self).__init__(name, config)
Oct 21 09:21:36 ip-172-31-173-176 docker/88c787eafc06[785]:   File "/home/postgres/patroni/patroni/dcs.py", line 126, in __init__
Oct 21 09:21:36 ip-172-31-173-176 docker/88c787eafc06[785]:     self._base_path = '/service/' + self._scope
Oct 21 09:21:36 ip-172-31-173-176 docker/88c787eafc06[785]: TypeError: cannot concatenate 'str' and 'int' objects
```

[...]

PR #75, opened 2015-10-21
 Make it possible to override default namespace (/service/) from a config file

CyberDem0n:
If the namespace is not specified in a config file /service/ would be used.

Also it's possible to use just '/' as a namespace.
In this case we would have the following structure in a DCS:
  /scope1
  /scope2
  ...

### DIFF
--- a/patroni/dcs.py
+++ b/patroni/dcs.py
@@ -123,7 +123,7 @@ class AbstractDCS:
         """
         self._name = name
         self._namespace = '/{}'.format(config.get('namespace', '/service/').strip('/'))
-        self._base_path = '/'.join([self._namespace, config['scope']])
+        self._base_path = '/'.join([self._namespace, str(config['scope']]))
 
         self._cluster = None
         self._cluster_thread_lock = Lock()
--- a/patroni/dcs.py
+++ b/patroni/dcs.py
@@ -123,7 +123,7 @@ class AbstractDCS:
         """
         self._name = name
         self._namespace = '/{}'.format(config.get('namespace', '/service/').strip('/'))
-        self._base_path = '/'.join([self._namespace, str(config['scope'])])
+        self._base_path = '/'.join([self._namespace, config['scope']])
 
         self._cluster = None
         self._cluster_thread_lock = Lock()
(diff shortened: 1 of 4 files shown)


patroni/dcs.py
```diff
@@ -122,8 +122,8 @@ def __init__(self, name, config):
             i.e.: `zookeeper` for zookeeper, `etcd` for etcd, etc...
         """
         self._name = name
-        self._scope = config['scope']
-        self._base_path = '/service/' + self._scope
+        self._namespace = '/{}'.format(config.get('namespace', '/service/').strip('/'))
+        self._base_path = '/'.join([self._namespace, config['scope']])
```
feikesteenbergen: str(config['scope']) to address issue #74?

CyberDem0n: Well, the same way one could put list or dict objects there instead of str...
```yaml
scope:
  foo: bar
```

```yaml
scope: [foo, bar]
```

Such objects can be successfully converted to str, but... is this really expected behaviour?

alexeyklyukin: I'd propose to check the isinstance for the string type and exit right away if it is not True . Because on python2 a string can be either str or unicode, and there is no unicode type in python3, I'd suggest using six, since we already require it:

    import six
    ...
    if not isinstance(self._scope, six.string_types) and not instance(self._scope, six.text_type):
        logger.fatal('Scope should be an integer, supplied {}'.format(self._scope))
        sys.exit(1)

tests/test_postgresql.py
```diff
@@ -346,7 +346,7 @@ def test_remove_data_directory(self):
         self.p.remove_data_directory()
 
     @patch('subprocess.check_output', MagicMock(return_value=0, side_effect=pg_controldata_string))
-    @patch('subprocess.check_output', side_effect=subprocess.CalledProcessError)
+    @patch('subprocess.check_output', side_effect=subprocess.CalledProcessError(1, ''))
```
alexeyklyukin: This test is wrong on more than one layer (and your fix doesn't do much). Just leave it out the pull request, I will fix it with a separate one.

PR #79, opened 2015-10-23
Move basebackup method to postgresql.py, make other methods extensions

jberkus:
This is my PR to merge the default replica method, basebackup, into the mainstream postgresql.py.  Some notes:

1. The create_replica test is failing for reasons I can't figure out.  The body of the test consists of passing an invalid connection string and having create_replica fail.  When I walk through the code on my own, with the same parameters, it fails as it should, but it "succeeds" in the test for reasons I can't discern.  Probably I should just rewrite the test to actually create a replica, except is there a postgresql running at that point?

2. I had to make substantial changes to the connstring static method.  As far as I can tell, this is used only by create_replica at present, but please check to make sure I didn't break something else.

3. I split out restore-from-wal-e into its own library in scripts/.  However, this probably means some of the other files in scripts/ should be deleted.

Other comments?

### DIFF
--- a/patroni/postgresql.py
+++ b/patroni/postgresql.py
@@ -216,7 +216,7 @@ class Postgresql:
                 ret = self.basebackup(leader, env)
                 if ret == 0:
                     # if basebackup succeeds, exit with success
-                    return 0
+                    break
             else:
                 # user-defined method; check for configuration
                 # not required, actually
@@ -247,17 +247,18 @@ class Postgresql:
 
                 try:
                     # call script with the full set of parameters
-                    ret = subprocess.call(shlex.split(cmd) + shlex.split(method_config), env=env)
+                    ret = subprocess.call(shlex.split(cmd) + params, env=env)
                     # if we succeeded, stop
                     if ret == 0:
-                        return ret
+                        break
                 except Exception as e:
                     logger.exception('Error creating replica using method {0}: {1}'.format(replica_method, e.str))
                     ret = 1
 
         # write the recovery.conf
         if ret == 0:
-            self.write_recovery_conf(leader)
+            ret = self.write_recovery_conf(leader)
+            return 0
 
         # out of methods, return 1
         return 1
--- a/patroni/postgresql.py
+++ b/patroni/postgresql.py
@@ -241,9 +241,9 @@ class Postgresql:
                 else:
                     cmd = replica_method
                     method_config = {"scope": self.scope,
-                                    "role": "replica",
-                                    "datadir": self.data_dir,
-                                    "connstring": connstring}
+                                     "role": "replica",
+                                     "datadir": self.data_dir,
+                                     "connstring": connstring}
 
                 try:
                     # call script with the full set of parameters
--- a/patroni/postgresql.py
+++ b/patroni/postgresql.py
@@ -9,6 +9,7 @@ import time
 from patroni.exceptions import PostgresConnectionException, PostgresException
 from patroni.utils import Retry, RetryFailedError
 from six.moves.urllib_parse import urlparse
+from threading import Lock
 
 logger = logging.getLogger(__name__)
 
@@ -47,6 +48,8 @@ class Postgresql:
         self.replication = config['replication']
         self.superuser = config['superuser']
         self.admin = config['admin']
+        self.pgpass = config.get('pgpass', None) or os.path.join(os.path.expanduser('~'), 'pgpass')
+        self.pg_rewind = config.get('pg_rewind', {})
         self.callback = config.get('callbacks', {})
         self.use_slots = config.get('use_slots', True)
         self.schedule_load_slots = self.use_slots
@@ -56,7 +59,6 @@ class Postgresql:
         self.postmaster_pid = os.path.join(self.data_dir, 'postmaster.pid')
         self.trigger_file = config.get('recovery_conf', {}).get('trigger_file', None) or 'promote'
         self.trigger_file = os.path.abspath(os.path.join(self.data_dir, self.trigger_file))
-        self._role = 'replica'
 
         self._pg_ctl = ['pg_ctl', '-w', '-D', self.data_dir]
 
@@ -67,8 +69,50 @@ class Postgresql:
 
         self._connection = None
         self._cursor_holder = None
-        self.members = []  # list of already existing replication slots
-        self.retry = Retry(max_tries=-1, deadline=10, max_delay=1, retry_exceptions=PostgresConnectionException)
+        self._need_rewind = False
+        self._sysid = None
+        self.replication_slots = []  # list of already existing replication slots
+        self.retry = Retry(max_tries=-1, deadline=5, max_delay=1, retry_exceptions=PostgresConnectionException)
+
+        self._state = 'stopped'
+        self._state_lock = Lock()
+        self._role = 'replica'
+        self._role_lock = Lock()
+
+        if self.is_running():
+            self._state = 'running'
+            self._role = 'master' if self.is_leader() else 'replica'
+
+    @property
+    def can_rewind(self):
+        """ check if pg_rewind executable is there and that pg_controldata indicates
+            we have either wal_log_hints or checksums turned on
+        """
+        # low-hanging fruit: check if pg_rewind configuration is there
+        if not self.pg_rewind or\
+           not (self.pg_rewind.get('username', '') and self.pg_rewind.get('password', '')):
+            return False
+
+        cmd = ['pg_rewind', '--help']
+        try:
+            ret = subprocess.call(cmd, stdout=open(os.devnull, 'w'), stderr=subprocess.STDOUT)
+            if ret != 0:  # pg_rewind is not there, close up the shop and go home
+                return False
+        except OSError:
+            return False
+        # check if the cluster's configuration permits pg_rewind
+        data = self.controldata()
+        return data.get('wal_log_hints setting', 'off') == 'on' or data.get('Data page checksum version', '0') != '0'
+
+    @property
+    def sysid(self):
+        if not self._sysid:
+            data = self.controldata()
+            self._sysid = data.get('Database system identifier', "")
+        return self._sysid
+
+    def require_rewind(self):
+        self._need_rewind = True
 
     def get_local_address(self):
         listen_addresses = self.listen_addresses.split(',')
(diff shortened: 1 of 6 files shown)


CyberDem0n:
Hi Josh.

Currently I've merged your pull  request into the separate branch of out repository: https://github.com/zalando/patroni/compare/restore/movebasebackup

Besides that I've fixed some obvious errors, pep8 formatting and unit test for create_replica method. Please pull it into your repository.

patroni/postgresql.py
```diff
@@ -165,49 +113,85 @@ def data_directory_empty(self):
...
+                if ret == 0:
+                    # if basebackup succeeds, exit with success
+                    return 0
+            else:
+                # user-defined method; check for configuration
+                # not required, actually
+                if replica_method in self.config:
+                    # look to see if the user has supplied a full command path
+                    # if not, use the method name as the command
+                    if "command" in self.config[replica_method]:
```
CyberDem0n: It looks like you are trying to implement the case when replica_method is set to some 'program_name', but there is no special configuration section for it...
In this case you will get cmd undefined....

patroni/postgresql.py
```diff
@@ -165,49 +113,85 @@ def data_directory_empty(self):
...
+                    # add the default parameters
+                    method_config.update({"scope": self.scope, 
+                                          "role" : "replica",
+                                          "datadir" : self.data_dir,
+                                          "connstring" : self.connstring})
+                    params = ["--{0}={1}".format(arg, val) for arg, val in method_config.iteritems()]
+                    
+                try:
+                    # call script with the full set of parameters
+                    ret = subprocess.call(shlex.split(cmd) + shlex.split(method_config), env=env)
```
CyberDem0n: I guess you wanted to call: subprocess.call(shlex.split(cmd) + **params**, env=env)

But! On this line we would have cmd and params undefined if there is no special configuration section for a given replica_method in the configuration file.

alexeyklyukin:
>I split out restore-from-wal-e into its own library in scripts/. However, this probably means some of the other files in scripts/ should be deleted.

I don't see it among the files to be committed? Did you 'git add' it?

patroni/postgresql.py
jberkus: hmmm?  but then replica_method would not be in self.confg, and thus that whole code section would not be executed.

patroni/postgresql.py
jberkus: Again, if there is no section defined, you never get to that code.

jberkus:
Will check for wal-e-restore.sh.  Also I see that my changed to postgres0.yml didn't merge in; fixing.

patroni/postgresql.py
CyberDem0n: try block is not located inside "if replica_method in self.config:"...
It would be executed in any case...

jberkus: that's a bug then, probably an artifact of the problematic code merge.  New commits coming.

jberkus:
So, updated.

However, there's a major design issue.

Many restore scripts will need to write a restore_command to recovery.conf.  For example, wale_restore.py needs to write "restore_command = envdir /etc/wale.d wal_e wal_fetch %f %p" to the recovery.conf in order to work.

Since the restore script only returns 1 or 0, there's no clear way to communicate this back to postgresql.py.  Obviously this can be set up by the user by manually supplying a restore command, but isn't that asking a lot?

patroni/postgresql.py
```diff
@@ -189,25 +189,78 @@ def sync_from_leader(self, leader):
...
-            self.delete_trigger_file()
-        except:
-            logger.exception('Error when creating replica')
-            ret = 1
-        if ret != 0:
-            self.set_state('failed to build replica from {host}:{port}'.format(**master_connection))
-        return ret
+                try:
+                    # call script with the full set of parameters
+                    ret = subprocess.call(shlex.split(cmd) + shlex.split(method_config), env=env)
```
alexeyklyukin: The shlex.split(method_conifg) will raise an exception, since method_config is a dictionary, and shlex.split expects a string. Since you are declaring the 'params' variable, but not using them anywhere, my guess is that you originally prepared them to be passed into subprocess.call.

patroni/postgresql.py
```diff
@@ -189,25 +189,78 @@ def sync_from_leader(self, leader):
...
+                    # if we succeeded, stop
+                    if ret == 0:
+                        return ret
+                except Exception as e:
+                    logger.exception('Error creating replica using method {0}: {1}'.format(replica_method, e.str))
+                    ret = 1
+
+        # write the recovery.conf
+        if ret == 0:
```
alexeyklyukin: This code is not reachable, because of 
if ret == 0:
    return ret
I think it makes more sense to break from the loop over replica methods than to return immediately.

patroni/scripts/wale_restore.py
```diff
@@ -0,0 +1,227 @@
...
+        else:
+            self.wal_e.iam_string = ''
+
+    def setup(self):
+        pass
+
+    def replica_method(self):
+        return self.create_replica_with_pg_basebackup
+
+    def replica_fallback_method(self):
```
alexeyklyukin: Support for replica_fallback_method should be removed.

patroni/scripts/wale_restore.py
```diff
@@ -0,0 +1,227 @@
...
+
+    def run(self):
+        """ creates a new replica using either pg_basebackup or WAL-E """
+        method_fn = self.replica_method()
+        ret = method_fn() if method_fn else 1
+        if ret != 0 and self.replica_fallback_method() is not None:
+            ret = (self.replica_fallback_method())()
+        return ret
+
+    def create_replica_with_pg_basebackup(self):
```
alexeyklyukin: This we don't need anymore.

patroni/scripts/wale_restore.py
```diff
@@ -0,0 +1,227 @@
...
+        else:
+            self.wal_e.cmd = 'envdir {0} wal-e {1} '.\
+                format(self.wal_e.dir, self.wal_e.iam_string)
+
+    def replica_method(self):
+        if self.should_use_s3_to_create_replica():
+            return self.create_replica_with_s3
+        return None
+
+    def replica_fallback_method(self):
```
alexeyklyukin: We don't want to fallback to pg_basebackup when wal-e is chosen as the only replica_method.

postgres0.yml
```diff
@@ -34,10 +34,6 @@ postgresql:
   data_dir: data/postgresql0
   maximum_lag_on_failover: 1048576 # 1 megabyte in bytes
   use_slots: True
-  pgpass: /tmp/pgpass0
```
alexeyklyukin: Please, do not remove the settings for pgpass and pg_rewind

postgres0.yml
```diff
@@ -46,16 +42,20 @@ postgresql:
     password: rep-pass
     network:  127.0.0.1/32
   superuser:
-    username: postgres
```
alexeyklyukin: The superuser username should not be removed as well.

postgres0.yml
```diff
@@ -67,4 +67,3 @@ postgresql:
     archive_timeout: 1800s
     max_replication_slots: 5
     hot_standby: "on"
-    wal_log_hints: "on"
```
alexeyklyukin: Do not remove this

postgres1.yml
```diff
@@ -34,10 +34,6 @@ postgresql:
   data_dir: data/postgresql1
   maximum_lag_on_failover: 1048576 # 1 megabyte in bytes
   use_slots: True
-  pgpass: /tmp/pgpass1
```
alexeyklyukin: Same issues as with postgres0.yml

jberkus:
gah.  Ok, I need to set up a better way to edit on my test server; clearly I'm getting version conflicts.

jberkus:
Any ideas for how to resolve the issue with restore_command?

jberkus:
New commit.

First, I don't get this change to test_create_replica:

    @patch('time.sleep', Mock())
    def test_create_replica(self):
        self.p.delete_trigger_file = Mock(side_effect=OSError())
        with patch('subprocess.call', Mock(side_effect=[1, 0])):
            self.assertEquals(self.p.create_replica({'host': '', 'port': '', 'user': ''}, ''), 0)
        with patch('subprocess.call', Mock(side_effect=[Exception(), 0])):
            self.assertEquals(self.p.create_replica({'host': '', 'port': '', 'user': ''}, ''), 0)

Those connection strings are invalid, so replica creation should fail.  No?  Is there a way we can test actually creating a replica using basebackup?  Is there a running master at that stage of the test?

Second, I was not able to test wale_restore, because I'm currently fighting permissions issues on S3.  If someone else can test it, great; otherwise, I'll get to it tommorrow.

alexeyklyukin:
Hi Josh,

I think archive_command in recovery.conf should be handled by the user via the patroni yml configuration: I don't see a universal way to construct it for each possible replica method.

alexeyklyukin:
As for the tests, I think they actually uncover an issue with the code. When the replica_method is set to base backup (because there is no specific configuration in the tests), it tries to call the base backup in line 216 of postgresql.py, returning non-zero value (as subprocess.call returns 1 for the first call), but right after that it goes into the else and re-executes the same command in line 250, returning 0, so the whole test actually returns 0.

I think here both the test and the code it covers are wrong.

And as a note, generally, our tests are more coverage tests : we don't actually run PostgreSQL, or etcd, and do not create replicas. The goal of those tests is to make sure that the code does't have obvious syntax errors. What we need are regression tests that actually start Patroni, etcd (in the docker), do actions like failover and verify the result. Do you have any experience with writing such tests in Python?

jberkus:
Alexey: restore_command, not archive_command.  Not that we don't need to deal with archive_command too!  You're right, for now let's punt and leave it up to the user.

Will deal with the code issue for the create_replica test.

As for doing integration testing in python ... not yet, but I'm figuring it out!  I've got a docker-compose setup, and am working on doing the rest to automatically run patroni through its paces.

alexeyklyukin:
Doh, of course it's restore_command!

jberkus:
Alexey:

> As for the tests, I think they actually uncover an issue with the code. When the replica_method is set to base backup (because there is no specific configuration in the tests), it tries to call the base backup in line 216 of postgresql.py, returning non-zero value (as subprocess.call returns 1 for the first call), but right after that it goes into the else and re-executes the same command in line 250, returning 0, so the whole test actually returns 0.

I don't get this.  How could it ever reach line 250?  That requires a non-basebackup method to be defined.  

Also, why is it testing for ==0?  Shouldn't it be looking for != 0?

alexeyklyukin:
Yes, you are right, like 250 is unreachable if basebackup is selected, I was confused by the indentation on the github screen. I'm testing this branch on AWS now.

jberkus:
Apparently someone added a test this doesn't pass.  Please do test the AWS plugin.

alexeyklyukin:
I did some tests on AWS, checking the WAL-E replica method, as well as the ability to fall-back to the basebackup if it's configured. I've fixed a couple of mistakes (mostly in the WAL-E script, but also one in the Patroni itself), merged recent changes and produced a pull request out of it https://github.com/zalando/patroni/pull/103. Please, take a look. The whole thing is missing the tests, and the old WAL-E script is yet to be removed, but mostly it's finished.

PR #82, opened 2015-10-27
Feature/patroni cli or ctl tbd

feikesteenbergen:


tests/test_ha.py
```diff
@@ -27,10 +27,10 @@ def get_cluster_not_initialized_without_leader():
 
 def get_cluster_initialized_without_leader(leader=False, failover=None):
     m = Member(0, 'leader', 28, {'conn_url': 'postgres://replicator:rep-pass@127.0.0.1:5435/postgres',
-                                 'api_url': 'http://127.0.0.1:8008/patroni'})
+                                 'api_url': 'http://127.0.0.1:8008/patroni', 'role':'replica'})
```
CyberDem0n: Why do you need 'role' here?

patroni/cli.py
```diff
@@ -0,0 +1,323 @@
...
+    logging.basicConfig(format='%(asctime)s - %(levelname)s - %(message)s', level=LOGLEVEL)
+
+
+def get_dcs(config, scope):
+    scheme, hostname, port = map(config.get('dcs', {}).get, ('scheme', 'hostname', 'port'))
+
+    if scheme == 'etcd':
+        return Etcd(name=scope, config={'scope': scope, 'host': '{}:{}'.format(hostname, port)})
+
+    raise Exception('Can not find suitable configuration of distributed configuration store')
```
alexeyklyukin: No Zookeeper support yet?

patroni/cli.py
```diff
@@ -0,0 +1,323 @@
...
+        if force:
+            master = cluster.leader.member.name
+        else:
+            master = click.prompt('Master', type=str, default=cluster.leader.member.name)
+
+    if cluster.leader.member.name != master:
+        raise Exception('Member {} is not the leader of cluster {}'.format(master, cluster_name))
+
+    candidate_names = [str(m.name) for m in cluster.members if m.name != master]
+    candidate_names.sort()
```
alexeyklyukin: I think it would be good to stop here if len(candidate_names) == 0

tests/test_ha.py
feikesteenbergen: Because I return the role to the client.

patroni/cli.py
feikesteenbergen: Indeed, no support yet

patroni/ctl.py
```diff
@@ -0,0 +1,613 @@
...
+        time.sleep(watch)
+        counter += 1
+        if clear:
+            click.clear()
+        yield 0
+
+
+def build_connect_parameters(conn_url, connect_parameters={}):
+    params = connect_parameters.copy()
+    parsed = parseurl(conn_url)
```
alexeyklyukin: We should check that parsurl does not return an error.

patroni/ctl.py
```diff
@@ -0,0 +1,613 @@
...
+
+    if file is not None:
+        command = file.read()
+
+    config, dcs, cluster = ctl_load_config(cluster_name, config_file, dcs)
+
+    cursor = None
+    for _ in watching(w, watch, clear=False):
+
+        output, cursor = query_member(cluster=cluster, cursor=cursor, member=member, role=role, command=command)
```
alexeyklyukin: It does not make much sense to expose the cursor outside of query_member. One way to do it without using a global variable is to convert query_member into a generator, yielding the cursor on every call.

patroni/ctl.py
```diff
@@ -0,0 +1,613 @@
...
+                message = 'No connection to role={} is available'.format(role)
+            logging.debug(message)
+            return [[timestamp(0), message]], None
+
+        cursor.execute('SELECT pg_is_in_recovery()')
+        in_recovery = cursor.fetchone()[0]
+
+        if in_recovery and role == 'master' or not in_recovery and role == 'replica':
+            cursor.connection.close()
+            return None, None
```
alexeyklyukin: It makes sense to return the same values as in line 310 (i.e. return the same values on error).

patroni/ctl.py
```diff
@@ -0,0 +1,613 @@
...
+
+    if cluster.leader:
+        confirm = click.prompt('This cluster currently is healthy. Please specify the master name to continue')
+        if confirm != cluster.leader.name:
+            raise PatroniCtlException('You did not specify the current master of the cluster')
+
+    dcs.client.delete(dcs._base_path, recursive=True)
+
+
+def wait_for_leader(dcs, timeout=30):
```
alexeyklyukin: This assumes Patroni loop_wait is not more than 30 seconds, which is in most cases true, but can be configured differently in yaml configuration. Maybe it makes sense for Patroni to expose its configuration via dcs...

patroni/ctl.py
```diff
@@ -0,0 +1,613 @@
...
+    dcs = get_dcs(config, cluster_name)
+    cluster = dcs.get_cluster()
+
+    return config, dcs, cluster
+
+
+@ctl.command('restart', help='Restart cluster member')
+@click.argument('cluster_name')
+@click.argument('member_names', nargs=-1)
+@click.option('--role', '-r', help='Restart only members with this role', default='any',
```
alexeyklyukin: --role=any doesn't make much sense, since one can just omit --role altogether.

patroni/ctl.py
```diff
@@ -0,0 +1,613 @@
...
+        if r.status_code == 200:
+            logging.debug(r)
+            logging.debug(r.text)
+            cluster = dcs.get_cluster()
+            click.echo(timestamp() + ' Failing over to new leader: {}'.format(cluster.leader.member.name))
+        else:
+            click.echo('Failover failed, details: {}, {}'.format(r.status_code, r.text))
+            return
+    except:
+        logging.exception(r)
```
alexeyklyukin: I'd propose to make it a separate option.

PR #89, opened 2015-11-03
Remove the useless codeline.

alexeyklyukin:


patroni/__init__.py
```diff
@@ -20,7 +20,6 @@ def __init__(self, config):
         self.nap_time = config['loop_wait']
         self.postgresql = Postgresql(config['postgresql'])
         self.dcs = self.get_dcs(self.postgresql.name, config)
-        host, port = config['restapi']['listen'].split(':')
```
antban: probably sutable can be renamed to suitable too

alexeyklyukin: Fixed, thank you!

PR #91, opened 2015-11-08
Store host:port connect_address in KV to be used by routing systems

drnic:
In addition to `conn_url` this PR adds `conn_address` which contains just `host:port`. This is very simple for templating tools to directly use in haproxy configuration, for example.

An example confd that uses the additional `conn_address` in an haproxy.cfg:

```
{{range $index, $name := ls "/routing/allocation"}}
	{{range $member := getvs (printf "/service/%s/members/*" $name)}}
		{{$data := json $member}}
		{{if eq $data.role "master"}}
listen {{$name}}
  mode tcp
  bind *:{{(printf "/routing/allocation/%s" $name | getv)}}
  server leader {{$data.conn_address}}
		{{end}}
	{{end}}
{{end}}
```

### DIFF
--- a/patroni/ha.py
+++ b/patroni/ha.py
@@ -49,6 +49,7 @@ class Ha:
     def touch_member(self):
         data = {
             'conn_url': self.state_handler.connection_string,
+            'conn_address': self.state_handler.connect_address,
             'api_url': self.patroni.api.connection_string,
             'state': self.state_handler.state,
             'role': self.state_handler.role,
--- a/patroni/postgresql.py
+++ b/patroni/postgresql.py
@@ -67,9 +67,9 @@ class Postgresql:
         self._pg_ctl = ['pg_ctl', '-w', '-D', self.data_dir]
 
         self.local_address = self.get_local_address()
-        connect_address = config.get('connect_address', None) or self.local_address
+        self.connect_address = config.get('connect_address', None) or self.local_address
         self.connection_string = 'postgres://{username}:{password}@{connect_address}/postgres'.format(
-            connect_address=connect_address, **self.replication)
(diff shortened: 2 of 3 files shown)


CyberDem0n:
@drnic, @bodymindarts, I don't know all capabilities of ``confd `` in Nov 2015, but with the latest version it's definitely possible to extract ``host:port`` from *conn_url*:
```
        {{$data := json $member}}
        {{$conn_addr := base (replace (index (split $data.conn_url "/") 2) "@" "/" -1)}}
```

And one more thing: there is no need to search for member with ``$data.role == "master"`` and I would even say that it is totally wrong.
Instead of doing this, let the `haproxy`` to do it's job, just put all members into haproxy config and configure health-check there. You can take [haproxy.cfg](https://github.com/zalando/patroni/blob/master/haproxy.cfg) as an example. It (haproxy) determines where is the master by performing ``OPTIONS /`` request on Patroni REST API. Patroni responds with status code 200 if it is running as the master. Otherwise it will respond with 502.

If you want to do read load-balancing, it's possible to use another endpoint, ``GET /replica`` or ``OPTIONS /replica``. In this case you will get 200 only when the node running as replica.

bodymindarts:
@CyberDem0n thanks a lot for your investigation. Looks like the string extraction will work fine.
We are currently looking for role master because we are not exposing the patroni endpoint to the outside network but will think about if we want to.

bodymindarts:
@CyberDem0n thanks again for looking into that. The confd version we had running didn't support that kind of string manipulation but after updating confd we don't rely on this PR anymore and have been able to smoothly update to v0.90!
PR can be closed /cc @drnic

PR #92, opened 2015-11-09
Call touch_member right after demote and promote.

CyberDem0n:
This is necessary to propagate actual information about node into DCS.


alexeyklyukin:
Makes sense to me. +1

Issue #88, opened 2015-11-02
error if user comments out postgresql/parameters section


jberkus:
Two places in postgresql.py, it does this:

        for setting, value in self.config['parameters'].items():

There's two problems with this code:

1. if the user comments out the parameters section because there are no custom parameters, patroni errors

2. we shouldn't be reaching into config inside a function; this should be passed to self.server_parameters in the init function.

I'll fix this later if someone else doesn't do it first.

feikesteenbergen:
Fixed by PR #93

PR #93, opened 2015-11-11
Do not fail when there is no postgresql/parameters section in a config file

CyberDem0n:
https://github.com/zalando/patroni/issues/88


alexeyklyukin:
Looks good :+1:

Issue #87, opened 2015-10-30
we should enable data page checksums


jberkus:
If patroni is initdb'ing, we should enable data page checksums because:

* it's a good idea
* performance impact is not noticeable under most circumstances
* prior to 9.5, it's required for pg_rewind to be safe to use

CyberDem0n:
I guess we should solve this problem more general. We can create special section in the config.yaml file which will contain all needed parameters and options for initdb

jberkus:
Why?

If Patroni is doing initdb, I don't see why it should give the user an option to turn off checksums.

a1exsh:
Do we have any recent benchmark measuring performance penalty of enabling
data checksums? Last time I saw one, the effect was not nearly negligible,
I believe.

jberkus:
Hmmm, I need to publish.  On workloads I've tested overhead has been < 5%.  Compared to the safety of detecting corruption as soon as it happens, that's negligable.  Also, I'm pretty sure that pg_rewind without checksums is unsafe, and you're building pg_rewind into Patroni.

a1exsh:
Would be glad to see the benchmark. :-)  pg_rewind requires checksums OR
logging of hint bits iirc.

alexeyklyukin:
Patroni can be attached to an already running cluster, or maybe in the future used to initialise new replicas out of the master that is completely unaware of patroni. On both of those cases we cannot guarantee the checksums will be on. We might solve this by fetching the checksum value parameter out of an existing master, but I think it would be easier to just make initdb options configurable.

jberkus:
We don't initdb in either of those cases.

alexeyklyukin:
Oh, right. But we also won't be able to rely on the checksums being on for pg_rewind in those cases. I think we should make this option configurable and choose the default value depending on the results of the benchmarks.

feikesteenbergen:
Implemented with custom options.

Issue #86, opened 2015-10-29
initialization fails if superuser.name=postgres


jberkus:
With HEAD as of a couple days ago, if you set the following:

  superuser:
    username: postgres
    password: zalando

For a new database system, you will get this error on initialization:

```
2015-10-29 19:40:06,302 INFO: established a new patroni connection to the postgres cluster
ERROR:  role "postgres" already exists
STATEMENT:  CREATE ROLE "postgres" WITH LOGIN SUPERUSER PASSWORD 'super'
2015-10-29 19:40:06,316 INFO: removing initialize key after failed attempt to initialize the cluster
2015-10-29 19:40:06,317 INFO: closed patroni connection to the postgresql cluster
waiting for server to shut down...LOG:  received immediate shutdown request
.WARNING:  terminating connection because of crash of another server process
DETAIL:  The postmaster has commanded this server process to roll back the current transaction and exit, because another server process exited abnormally and possibly corrupted shared memory.
HINT:  In a moment you should be able to reconnect to the database and repeat your command.
 done
server stopped
2015-10-29 19:40:07,349 INFO: renaming data directory to /pgdata/data_2015-10-29-19-40-07
2015-10-29 19:40:07,350 ERROR: Error communicating with Postgresql. Will try again later
Traceback (most recent call last):
  File "build/bdist.linux-x86_64/egg/patroni/ha.py", line 380, in _run_cycle
    return self.bootstrap()  # new node
  File "build/bdist.linux-x86_64/egg/patroni/ha.py", line 79, in bootstrap
    self.state_handler.bootstrap()
  File "build/bdist.linux-x86_64/egg/patroni/postgresql.py", line 647, in bootstrap
    self.create_connection_users()
  File "build/bdist.linux-x86_64/egg/patroni/postgresql.py", line 588, in create_connection_users
    self.superuser['username']), self.superuser['password'])
  File "build/bdist.linux-x86_64/egg/patroni/postgresql.py", line 160, in query
    return self.retry(self._query, sql, *params)
  File "build/bdist.linux-x86_64/egg/patroni/utils.py", line 153, in __call__
    return func(*args, **kwargs)
  File "build/bdist.linux-x86_64/egg/patroni/postgresql.py", line 153, in _query
    raise e
ProgrammingError: role "postgres" already exists
```

Then initialization will fail ... and be reattempted, in a continuous loop.

For all of the roles, we should check if they already exist, and only create them if they don't (but set passwords if they do).

Will work on patch this week.

[...]

PR #94, opened 2015-11-11
Make it possible to specify custom options for initdb

CyberDem0n:
In the initial implementation we were using the only option
--encoding=UTF8. In order to have pg_rewind working with postgresql-9.3
we have to enable data-checksums. The naive approach was to enable it
globally but taking into account some performance degradation it's better
not to do it but make it possible to configure it.

In addition to that fix all problems with setting up password of default
postgres user: execute CREATE ROLE | ALTER ROLE depending on content of
pg_authid

This pull request addresses following issues: https://github.com/zalando/patroni/issues/86 and https://github.com/zalando/patroni/issues/87

### DIFF
--- a/patroni/postgresql.py
+++ b/patroni/postgresql.py
@@ -559,6 +559,7 @@ recovery_target_timeline = 'latest'
     def create_or_update_role(self, name, password, options):
         self.query("""DO $$
 BEGIN
+    SET local synchronous_commit = 'local';
     PERFORM * FROM pg_authid WHERE rolname = %s;
     IF FOUND THEN
         ALTER ROLE "{0}" WITH LOGIN {1} PASSWORD %s;
--- a/patroni/postgresql.py
+++ b/patroni/postgresql.py
@@ -169,11 +169,8 @@ class Postgresql:
 
     @staticmethod
     def initdb_allowed_option(name):
-        allowed_options = set(['auth', 'auth-host', 'auth-local', 'encoding', 'data-checksums',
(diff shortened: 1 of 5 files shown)


patroni/postgresql.py
```diff
@@ -164,9 +167,42 @@ def query(self, sql, *params):
     def data_directory_empty(self):
         return not os.path.exists(self.data_dir) or os.listdir(self.data_dir) == []
 
+    @staticmethod
+    def initdb_allowed_option(name):
+        allowed_options = set(['auth', 'auth-host', 'auth-local', 'encoding', 'data-checksums',
```
alexeyklyukin: Hm... why not just pass the user options directly to initdb and examine the return code in order to find out if it didn't like some of the options. You have to do it anyway (i.e. what if the user supplies a value for the option that does not need one, i.e. --data-checksums=true).

patroni/postgresql.py
```diff
@@ -164,9 +167,42 @@ def query(self, sql, *params):
...
+                               'locale', 'lc-collate', 'lc-ctype', 'lc-messages', 'lc-monetary',
+                               'lc-numeric', 'lc-time', 'text-search-config', 'xlogdir', 'debug', 'noclean'])
```
a1exsh: I know initdb options are changing infrequently, but does it really make sense to hard-code the known list here?  It will fail anyway if some option or passed value is invalid.

alexeyklyukin:
Looks good :+1:

PR #100, opened 2015-11-18
Handle unexpected exceptions in etcd.

alexeyklyukin:
Previously, patroni would die after receiving an exception
other than RetryFailedError, etcd.EtcdException from etcd.
We have observed an AttributeError raised by etcd on some
occasions. With this change, we demote ourselves, but not
terminate on such exceptions.


feikesteenbergen:
+1

PR #103, opened 2015-11-23
Pgexperts restore/movebasebackup

alexeyklyukin:
Configurable replica creation implemented by @jberkus + several fixes from @CyberDem0n and me. Tested locally, as well as on AWS with WAL-E.


patroni/postgresql.py
```diff
@@ -223,24 +223,72 @@ def sync_from_leader(self, leader):
...
     @staticmethod
     def build_connstring(conn):
-        return "host={host} port={port} user={user}".format(**conn)
+        """
+        >>> Postgresql.build_connstring({'host': '127.0.0.1', 'port': '5432'}) == 'host=127.0.0.1 port=5432 '
+        True
+        """
+        mconn = ""
+        for param, val in sorted(conn.items()):
+            mconn = mconn + "{0}={1} ".format(param, val)
```
CyberDem0n: Do we really need this awkward space at the end?
And actually we can have one-liner here:

```python
return ' '.join('{}={}'.format(param, val) for param, val in sorted(conn.items()))
```

patroni/postgresql.py
```diff
@@ -223,24 +223,72 @@ def sync_from_leader(self, leader):
...
+
+        return mconn
+
+    def create_replica(self, leader, env):
+        # create the replica according to the replica_method
+        # defined by the user.  this is a list, so we need to
+        # loop through all methods the user supplies
+        connstring = leader.conn_url
+        # get list of replica methods from config
+        replica_list = self.config.get('create_replica_method', 'basebackup')
```
feikesteenbergen: So the create_replica_method is not a list defined in yaml, but a comma seperated list?
What if we implement it in such a way that you can provide a list inside the yaml?

patroni/postgresql.py
```diff
@@ -223,24 +223,72 @@ def sync_from_leader(self, leader):
...
+                    break
+            else:
+                cmd = replica_method
+                # user-defined method; check for configuration
+                # not required, actually
+                if replica_method in self.config:
+                    method_config = self.config[replica_method].copy()
+                    # look to see if the user has supplied a full command path
+                    # if not, use the method name as the command
+                    if "command" in method_config:
```
CyberDem0n: Just a nitpicking:
lines 267-270 could be rewritten as:
```python
cmd = method_config.pop('command', cmd)
```

patroni/postgresql.py
```diff
@@ -223,24 +223,72 @@ def sync_from_leader(self, leader):
...
+                                          "connstring": connstring})
+                    params = ["--{0}={1}".format(arg, val) for arg, val in method_config.items()]
+                else:
+                    method_config = {"scope": self.scope,
+                                     "role": "replica",
+                                     "datadir": self.data_dir,
+                                     "connstring": connstring}
+                try:
+                    # call script with the full set of parameters
+                    ret = subprocess.call(shlex.split(cmd) + params, env=env)
```
CyberDem0n: `params` could be not defined here, because it is not declared in `else` block from line 277

patroni/postgresql.py
```diff
@@ -663,3 +711,29 @@ def remove_data_directory(self):
...
+    def basebackup(self, leader, env):
+        # creates a replica data dir using pg_basebackup.
+        # this is the default, built-in create_replica_method
+        # tries twice, then returns failure (as 1)
+        # uses "stream" as the xlog-method to avoid sync issues
+        master_connection = leader.conn_url
+        bbfailures = 0
+        maxfailures = 2
+        ret = 1
+        while bbfailures < maxfailures:
```
CyberDem0n:
```python
for bbfailures in range(0, maxfailures):
```
could save some lines of code and more predictable.

alexeyklyukin: Fixed, thank you!

patroni/postgresql.py
alexeyklyukin: you mean doing:

create_replica_method:
 - basebackup
 - wale
 - rsync

instead of
create_replica_method: basebackup, wale, rsync

why?

patroni/postgresql.py
```diff
@@ -663,3 +700,28 @@ def remove_data_directory(self):
...
             logger.exception('Could not remove data directory %s', self.data_dir)
             self.move_data_directory()
+
+    def basebackup(self, leader, env):
+        # creates a replica data dir using pg_basebackup.
+        # this is the default, built-in create_replica_method
+        # tries twice, then returns failure (as 1)
+        # uses "stream" as the xlog-method to avoid sync issues
+        master_connection = leader.conn_url
+        bbfailures = 0
```
CyberDem0n: This line is not needed any more.

patroni/postgresql.py
jberkus: @feikesteenbergen : yeah, that seems like a good idea.  Less vulnerable to punctuation wierdness.  Wanna add it?

feikesteenbergen:
+1

PR #104, opened 2015-11-25
Bugfixes for Patronictl and the Development Docker environment

feikesteenbergen:
For easier development using Docker the $HOSTNAME variable will be used to
name the running Patroni. Bumped some _segments postgresql settings to ensure
WAL files are not removed very quickly.

Increased the timeout for the post request for Patroni, as some operations
(failover) may take considerable time to complete.

The failover to a specific member was broken in patronictl as it used a wrong
key to specify the member to failover to.

Pretty printing fix for xlog lag, to prevent false negatives to show up and have
good alignment.

### DIFF
--- a/patroni/ctl.py
+++ b/patroni/ctl.py
@@ -111,7 +111,7 @@ def post_patroni(member, endpoint, content, headers={'Content-Type': 'applicatio
     url = urlparse(member.api_url)
     logging.debug(url)
     return requests.post('{}://{}/{}'.format(url.scheme, url.netloc, endpoint), headers=headers,
-                         data=json.dumps(content), timeout=30)
+                         data=json.dumps(content), timeout=60)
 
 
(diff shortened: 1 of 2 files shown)


patroni/ctl.py
```diff
@@ -111,7 +111,7 @@ def post_patroni(member, endpoint, content, headers={'Content-Type': 'applicatio
     url = urlparse(member.api_url)
     logging.debug(url)
     return requests.post('{}://{}/{}'.format(url.scheme, url.netloc, endpoint), headers=headers,
-                         data=json.dumps(content), timeout=5)
+                         data=json.dumps(content), timeout=30)
```
CyberDem0n: I would make timeout a little bit bigger then 30 seconds. Corresponding code in the `Api` which does manual failover will wait 30 seconds for a new leader in configuration store and then generate a response. This means that it could take more then 30 seconds to process request and your code might experience timeout...

PR #108, opened 2015-12-09
Call a checkpoint on master before pg_rewind.

alexeyklyukin:
PostgreSQL does not run a checkpoint during promition.
Since pg_rewind relies on the last checkpoint to get the timeline,
there is a short race condition right after the promotion, when
it can get the timeline wrong and fail. We work around this by
calling the checkpoint manually.

Make sure our test configuration does both archive and recovery.


slitsche:
+1

PR #112, opened 2015-12-22
Feature/patronictl password

feikesteenbergen:


Dockerfile
```diff
@@ -14,14 +14,19 @@ RUN apt-get upgrade -y
 
 ENV PGVERSION 9.4
 RUN apt-get install python python-yaml python-requests python-boto postgresql-${PGVERSION} python-dnspython python-kazoo python-pip -y
-RUN apt-get install python-dev postgresql-server-dev-${PGVERSION} -y
-RUN pip install python-etcd psycopg2
+RUN apt-get install python-dev postgresql-server-dev-${PGVERSION} python-prettytable -y
+ADD requirements-py2.txt /tmp/
+RUN pip install -r /tmp/requirements-py2.txt
```
a1exsh: Why mix `python-*` packages and pip?  Let's stick to one of these.

docker/entrypoint.sh
```diff
@@ -79,6 +79,11 @@ then
     ETCD_CLUSTER="127.0.0.1:4001"
 fi
 
+mkdir -p ~postgres/.config/patroni
+cat > ~postgres/.config/patroni/patronictl.yaml <<__EOF__
+{dcs_api: 'etcd://${ETCD_CLUSTER}', namespace: /service/}
+__EOF__
+
```
a1exsh: Did you mean `.json` instead of `.yaml` here?  Also, just below these lines we have `.yml` vs. `.yaml`.
And that doesn't look like correct JSON to me either: you need to double quote the keys and values, unless it is any of the `null`, `true`, `false` or numeric.

feikesteenbergen: It's valid yaml, it was generated using yaml.dump:

    >>> print( yaml.dump({'api_url':'etcd://localhost:4001', 'namespace': '/service/'}) )
    {api_url: 'etcd://localhost:4001', namespace: /service/}

  I chose yaml to have some similarity with the stups tooling, which all save their config in `~/.config/<tool>/tool.yaml` as well.

a1exsh: I must admit I no longer know what a YAML file looks like ;-)  I've opened the latest spec for v1.2 and I must say these guys are totally nuts.

alexeyklyukin:
+1

PR #118, opened 2016-01-18
Feature/scheduled failover

feikesteenbergen:
Allow failovers to be scheduled in the future. Requires changing of the failover key format and some sanity checks to when it will take place.


patroni/ha.py
```diff
@@ -270,6 +272,23 @@ def demote(self, delete_leader=True):
...
+            if delta > 10:
+                logging.debug('Awaiting failover, {0}, {1}, {2}'.format(now.isoformat(),
+                              failover.planned_at.isoformat(),
+                              delta))
+                return
+            elif delta < -15:
+                logger.info('Found a stale failover value, cleaning up: {}'.format(failover.planned_at))
+                self.dcs.manual_failover('', '', self.cluster.failover.index)
+                return
+            time.sleep(max(delta, 0))
```
a1exsh: What if a failover is scheduled 24h in the future and the process doing this sleep() dies for some reason: what will happen to it upon restart?  Will it notice a failover pending again and sleep for as long as it still should?

patroni/utils.py
```diff
@@ -12,39 +14,62 @@
...
+        return None
+
+    if timestamp.tzinfo is None:
+        timestamp = pytz.timezone(tzlocal.get_localzone().zone).localize(timestamp)
+
+    return timestamp
+
+
+def utcnow_timezone_aware():
+    return datetime.datetime.utcnow().replace(tzinfo=pytz.utc)
```
a1exsh: Doesn't this do the same thing: `datetime.now(pytz.utc)`?

patroni/ha.py
feikesteenbergen: If the failover is in the future, nothing is done, we log that we have read the value and return.

The time.sleep basically sleeps for a maximum of 10 and a minimum of 0 seconds

patroni/dcs.py
```diff
@@ -85,13 +86,39 @@ def conn_url(self):
...
+            t = [a.strip() for a in value.split(':')] + ['']
+            return Failover(index, t[0], t[1], None) if t[0] or t[1] else None
 
+        if data is None:
+            return None
+
+        if data.get('planned_at'):
+            data['planned_at'] = parse_datetime(data['planned_at'])
+
+        return Failover(index, data.get('leader'), data.get('member'), data.get('planned_at'))
```
alexeyklyukin: Extra line missing (flake8 is complaining)

patroni/dcs.py
```diff
@@ -85,13 +86,39 @@ def conn_url(self):
...
```
alexeyklyukin: New line is missing (flake8 is complaining)

patroni/api.py
```diff
@@ -175,13 +175,29 @@ def do_POST_failover(self):
...
+                planned_at = localize_datetime(planned_at)
+                if planned_at < utcnow_timezone_aware():
+                    data = b'Cannot schedule failover in the past'
+                    status_code = 422
+                elif self.server.patroni.dcs.manual_failover(leader, member, planned_at):
+                    data = b'Failover scheduled'
+                    status_code = 200
+            except (ValueError, TypeError):
+                logger.exception('Invalid planned failover time: {}'.format(request['planned_at']))
+                data = b'Unable to parse planned timestamp. Should be in ISO 8601 format'
```
alexeyklyukin: I think this error message is outdated, as parse_datetime will happily accept a string representing a date that is not in the ISO8601 format, i.e 'January 21st 2016'

patroni/api.py
```diff
@@ -175,13 +175,29 @@ def do_POST_failover(self):
...
-                self.server.patroni.dcs.event.set()
-                status_code, data = self.poll_failover_result(cluster.leader and cluster.leader.name, member)
+
+        data = b''
+        if request.get('planned_at'):
+            try:
+                planned_at = parse_datetime(request['planned_at'])
+                planned_at = localize_datetime(planned_at)
+                if planned_at < utcnow_timezone_aware():
+                    data = b'Cannot schedule failover in the past'
```
alexeyklyukin: Should we also reject the dates representing this very moment? If we accept them and move on, would it be possible for those dates to be already too far in the past once we revisit this part of the code in the next iteration of event loop?

alexeyklyukin: Another idea is to call the is_failover_possible unconditionally after recording the key from the plannned failover, so that the check if we need to failover will happen right away.

patroni/ctl.py
```diff
@@ -449,10 +450,11 @@ def reinit(cluster_name, member_names, config_file, dcs, force):
 @click.argument('cluster_name')
 @click.option('--master', help='The name of the current master', default=None)
 @click.option('--candidate', help='The name of the candidate', default=None)
+@click.option('--planned', help='Timestamp of a planned failover in ISO 8601 format', default=None)
```
alexeyklyukin: Also does not require ISO 8601 format.

patroni/ctl.py
```diff
@@ -490,6 +492,22 @@ def failover(config_file, cluster_name, master, candidate, force, dcs):
...
 
+    if planned is None and not force:
+        planned = click.prompt('When should the failover take place (e.g. 2015-10-01T14:30) ', type=str, default='now')
+
+    if (planned or 'now') == 'now':
+        planned_at = None
+    else:
+        planned_at = parse_datetime(planned)
+        if planned_at is None:
+            message = 'Unable to parse planned timestamp ({}). It should be in ISO 8601 format.'
```
alexeyklyukin: Also does not require ISO 8601 format.

patroni/api.py
alexeyklyukin: Not necessary ISO 8601, parse_datetime accepts other formats, i.e. "January 27, 2016"

patroni/dcs.py
```diff
@@ -85,12 +86,39 @@ def conn_url(self):
...
-        t = [a.strip() for a in value.split(':')] + ['']
-        return Failover(index, t[0], t[1]) if t[0] or t[1] else None
+        if value is None:
+            return None
+
+        try:
+            data = json.loads(value)
+        except ValueError:
+            t = [a.strip() for a in value.split(':')] + ['']
+            return Failover(index, t[0], t[1], None) if t[0] or t[1] else None
```
alexeyklyukin: You should check that the value.split list really contains 2 members

patroni/dcs.py
```diff
@@ -216,8 +244,29 @@ def attempt_to_acquire_leader(self):
     def set_failover_value(self, value, index=None):
         """Create or update `/failover` key"""
 
-    def manual_failover(self, leader, member, index=None):
-        return self.set_failover_value(leader + (':' + member if member else ''), index)
+    def manual_failover(self, leader, member, planned_at=None, index=None):
```
alexeyklyukin: Looks like all invocations of this method already do the parse/localize time conversion, so there is no need to call it agin inside.

patroni/utils.py
```diff
@@ -12,39 +14,62 @@
...
+        timestamp = pytz.timezone(tzlocal.get_localzone().zone).localize(timestamp)
+
+    return timestamp
+
+
+def utcnow_timezone_aware():
+    return datetime.datetime.now(pytz.utc)
 
 
 def calculate_ttl(expiration):
```
alexeyklyukin: Looks like this function is not used anywhere

patroni/dcs.py
feikesteenbergen: by adding `['']` at the end we ensure the list always contains at least 2 items, taken verbatim from the previous version.

patroni/api.py
feikesteenbergen: True, we could relax this requirement. Instead of telling people it should be in this format we could mention that that is an unambiguous format that will always work, regardless of locale.

patroni/dcs.py
feikesteenbergen: That may be true, however calling `localize_datetime` on an already timezone returns the same value.
It was added to ensure that whenever we write a value, we attach a timezone to it, even though currently we don't have anyone calling the function with unaware timezones.

patroni/api.py
alexeyklyukin: I think we should always use the ISO 8601 in the API, but provide a more user-friendly interface for patronictl. I don't think calling the parse_datetime and localize_datetime everywhere just in case in a sane way to go, it doesn't make the code more readable and it's easy to forget it somewhere. Instead, what about calling it only in patronictl, and just enforcing the ISO 8601 format with strptime in the API endpoint?

patroni/dcs.py
alexeyklyukin: gotcha, this would indeed work, but is somewhat harder to read than the code that just checks the length of the list...

patroni/api.py
feikesteenbergen: Sounds sane to me

feikesteenbergen: strptime is not thread safe, therefore the parsing is done using dateutil.

PR #123, opened 2016-01-29
Provide more context for healthchecks and monitoring.

feikesteenbergen:
Include version numbers of both PostgreSQL and patroni for the /patroni endpoint.
Scope is also returned.


alexeyklyukin:
Looks good to me

PR #124, opened 2016-02-01
Bugfix: Ensure to inject the superuser username when connecting.

feikesteenbergen:
When running patroni with a superuser different than the os-user, the connection
was being established using the os-username. This fixes this.


alexeyklyukin:
+1

PR #131, opened 2016-02-10
Bugfix/superuser username

alexeyklyukin:
Make sure username is translated into the user when calling
psycopg2.connect. Also, fix the sample configuration files to use
username everywhere.

### DIFF
--- a/patroni/postgresql.py
+++ b/patroni/postgresql.py
@@ -132,6 +132,8 @@ class Postgresql:
         if not self._connection or self._connection.closed != 0:
             r = parseurl('postgres://{}/postgres'.format(self.local_address))
             r.update(self.superuser)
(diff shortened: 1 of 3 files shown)


feikesteenbergen:
+1

patroni/postgresql.py
```diff
@@ -132,6 +132,8 @@ def connection(self):
         if not self._connection or self._connection.closed != 0:
             r = parseurl('postgres://{}/postgres'.format(self.local_address))
             r.update(self.superuser)
+            if r.get('username'):
+                r['user'] = r['username']
```
a1exsh: I think you also need to remove 'username' from the `r` dict?

avaczi:
+1

PR #132, opened 2016-02-10
Create superuser with the name specified in config file

CyberDem0n:
And later use this name to connect to the cluster


alexeyklyukin:
Looks good to me

CyberDem0n:
There are still some problems with the `checkpoint` method...

alexeyklyukin:
The new version fixed the issue with constant reinitialization attempt after a series of "establishing a new patroni connection to the postgres cluster" messages.

PR #134, opened 2016-02-10
Feature/scheduled failover squashed

feikesteenbergen:
Scheduled failover allows scheduling of a failover in the future.

### DIFF
--- a/patroni/api.py
+++ b/patroni/api.py
@@ -104,7 +104,7 @@ class RestApiHandler(BaseHTTPRequestHandler):
 
     @check_auth
     def do_POST_restart(self):
-        status_code = 503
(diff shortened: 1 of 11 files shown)


patroni/api.py
```diff
@@ -176,13 +179,31 @@ def do_POST_failover(self):
         member = request.get('member', None)
         cluster = self.server.patroni.ha.dcs.get_cluster()
         status_code = 503
```
alexeyklyukin: Shouldn't we return 500 internal error instead of 503 service unavailable for exceptions?

feikesteenbergen: Makes sense, I'll add it

PR #143, opened 2016-02-23
Add support for ZooKeeper/Exhibitor DCS URI in patronictl ... -d

a1exsh:


a1exsh:
Well, line 364 seems to be aware of the limitation of `ZooKeeper` class: no recursive delete.  So someone has to implement that first.

CyberDem0n:
Actually Feike is just using dcs.client object, which is not necessary available for different DCS implementations.
So far it's available for  Etcd and ZooKeeper and this client object even has similar interfaces of `delete` method:
```python
KazooClient.delete(self, path, version=-1, recursive=False)
etcd.Client.delete(self, key, recursive=None, dir=None, **kwdargs)
```

So recursive `delete` for ZooKeeper will work as well. But for example it will not work for Consul.
Instead of doing this ugly hack, one should add the new abstract method to the AbstractDCS class, something like `delete_cluster` and implement it in Etcd and ZooKeeper classes.

a1exsh:
@CyberDem0n oh, right: I didn't pay attention to the `.client` indirection.

a1exsh:
In the hindsight: should we have made client a non-public property, i.e. `_client`?

a1exsh:
Last Travis-CI build was cancelled (timeout?).  It is suggested that closing and re-opening the PR can help.

alexeyklyukin:
Yeah, cancelled it yesterday, it was stuck after 1h. Might be something wrong with the tests....

CyberDem0n:
Tests are wrong.
You should run them locally: python setup.py test

a1exsh:
Spotted it: the tests are hanging on python3.  They were green on every commit on python2, which is my default.  *scratches head*

PR #144, opened 2016-02-24
Feature/acceptance tests

alexeyklyukin:
Abuse of the lettuce acceptance tests in order to check the real-world scenarios when running Patroni. At the moment we just check the basic replication and failover, but adding new tests is quite easy. The drawbacks are that it takes some time to run the tests, and there is no support for parallel execution

PR #151, opened 2016-03-10
First implementation of cloning from the replica.

alexeyklyukin:
At the moment we just replace the master with the node at the 'clonefrom' tag if it's present. Master
should be available anyway, otherwise, it will not even try to do cloning.

Acceptance tests:
https://github.com/zalando/patroni/pull/144/commits


CyberDem0n:
+1

alexeyklyukin:
:+1:

feikesteenbergen:
:+1:

PR #152, opened 2016-03-10
Run acceptance tests on travis

CyberDem0n:
There are a few problems with travis and lettuce:
* lettuce is not compatible with python3
* postgresql-9.5 is not yet available on travis
* there is no pg_rewind available

Since travis is using virtualenv to enable different python version there is no obvious (not hacky) way to run different versions of python with all necessary modules installed at the same time. I.e. it's not possible to execute lettuce (python2) which will start up patroni with python3.
 
So far the only possible workaround for all problems - 'sudo: required' in .travis.yml which unfortunately makes initialisation of build much slower.
But at least sudo gives possibility to install postgresql-9.5 plus install lettuce not into virtual environment.

### DIFF
--- a/features/basic_replication.feature
+++ b/features/basic_replication.feature
@@ -11,6 +11,6 @@ Feature: basic replication
     When I kill postgres0
     Then postgres1 role is the primary after 30 seconds
     When I start postgres0
-    Then postgres0 role is the secondary after 10 seconds
+    Then postgres0 role is the secondary after 15 seconds
     When I add the table bar to postgres1
     Then table bar is present on postgres0 after 10 seconds
--- a/features/cascading_replication.feature
+++ b/features/cascading_replication.feature
@@ -8,5 +8,5 @@ Scenario: check a base backup from the replica
 	And I create label with "postgres0" in postgres0 data directory
 	And I create label with "postgres1" in postgres1 data directory
 	And I configure and start postgres2 with a tag clonefrom postgres1
-	Then replication works from postgres0 to postgres2 after 15 seconds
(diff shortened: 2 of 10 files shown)


features/basic_replication.feature
```diff
@@ -11,6 +11,6 @@ Feature: basic replication
     When I kill postgres0
     Then postgres1 role is the primary after 30 seconds
     When I start postgres0
-    Then postgres0 role is the secondary after 10 seconds
+    Then postgres0 role is the secondary after 15 seconds
```
CyberDem0n: Timeout was increased because travis is pretty slow...

features/cascading_replication.feature
```diff
@@ -8,5 +8,5 @@ Scenario: check a base backup from the replica
 	And I create label with "postgres0" in postgres0 data directory
 	And I create label with "postgres1" in postgres1 data directory
 	And I configure and start postgres2 with a tag clonefrom postgres1
-	Then replication works from postgres0 to postgres2 after 15 seconds
-	And there is a label with "postgres1" in postgres2 data directory
\ No newline at end of file
+	Then replication works from postgres0 to postgres2 after 30 seconds
```
CyberDem0n: Timeout was increased because travis is pretty slow

features/cascading_replication.feature
```diff
@@ -8,5 +8,5 @@ Scenario: check a base backup from the replica
...
+	And there is a label with "postgres0" in postgres2 data directory
```
CyberDem0n: clonefrom feature is not yet merged

alexeyklyukin:
+1

PR #156, opened 2016-03-15
Implement `delete_cluster` interface in for all available dcs

CyberDem0n:
In addition to that rename confusing `Etcd.client` and
`ZooKeeper.client` into `_client`. This attribute is available from
AbstractDCS and people had wrong impression that it provides the same
interface for different DCS implementations, which is obviously not the
case. For Etcd it has type etcd.Client and for ZooKeeper - KazooClient.


feikesteenbergen:
:+1:

alexeyklyukin:
:+1:

PR #160, opened 2016-03-18
Rename ambiguous `Failover.member` to candidate

CyberDem0n:
But! 'member' is still accepted by REST API and also name 'member' is
used to strore/read this value to/from DCS (for backward comatibility)


alexeyklyukin:
:+1:

alexeyklyukin:
Actually, I think it rightfully complains that do_failover_result has a parameter candidate that is unused.

CyberDem0n:
I know. Currently `poll_failover_result` method just checks that leader has changed.
But it also should check that leader is matching with the `candidate`.

alexeyklyukin:
:+1:

alexeyklyukin:
+1

PR #163, opened 2016-03-24
Bugfix: In some cases Ha.cluster can be None after calling `get_cluster`

CyberDem0n:
Such situation is causing patroni crash. Usually it was happening during
manual failover, after former master has demoted and `reset_cluster`
method has been called. In this case `fetch_cluster` was `False` and
`_load_cluster` method was returning value from `self._cluster`, which
was `None`.

### DIFF
--- a/patroni/zookeeper.py
+++ b/patroni/zookeeper.py
@@ -93,8 +93,8 @@ class ZooKeeper(AbstractDCS):
         self._client.add_listener(self.session_listener)
 
         self._my_member_data = None
-        self.fetch_cluster = True
-        self.last_leader_operation = 0
+        self._fetch_cluster = True
+        self._last_leader_operation = 0
 
         self._client.start(None)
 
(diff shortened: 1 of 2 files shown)


patroni/zookeeper.py
```diff
@@ -165,15 +165,15 @@ def _inner_load_cluster(self):
...
+        optime = self.get_node(self.leader_optime_path) if self._OPTIME in nodes and self._fetch_cluster else None
+        self._last_leader_operation = 0 if optime is None else int(optime[0])
+        self._cluster = Cluster(initialize, leader, self._last_leader_operation, members, failover)
 
     def _load_cluster(self):
         if self.exhibitor and self.exhibitor.poll():
             self._client.set_hosts(self.exhibitor.zookeeper_hosts)
 
-        if self.fetch_cluster:
+        if self._fetch_cluster or self._cluster is None:
```
CyberDem0n: This line is fixing a problem.

Everything else was changed just to give a hint that `_fetch_cluster` and `_last_leader_operation` are "private"

alexeyklyukin:
:+1:

PR #164, opened 2016-03-24
Speed up reattach of former leader to the cluster

CyberDem0n:
Instead of starting it up in "read-only", it will wait 2 seconds, to give
a time to somebody to promote and after it will execute normal `recover`
procedure.


feikesteenbergen:
2 seconds looks like a short period of time for someone else to promote, what would be the downside of making this take, say 10 seconds?

CyberDem0n:
Leader lock is removed immediately after shut down, so usually 2 seconds is enough to get notification about missing leader, run all health checks and take leader lock.
And actually acceptance tests are proving that.

alexeyklyukin:
:+1:

PR #166, opened 2016-03-30
Correct implementation of 'clonefrom' feature

CyberDem0n:
According to https://github.com/zalando/patroni/issues/48 'clonefrom'
tag should be boolean and it should be used to mark node as a suitable
for creation of a new replica from. If there are more then one such node
in the cluster (with tag clonefrom=true), one of them will be chosen
randomly.

### DIFF
--- a/features/cascading_replication.feature
+++ b/features/cascading_replication.feature
@@ -1,7 +1,7 @@
 Feature: cascading replication
 	We should check that patroni can do base backup and streaming from the replica
 
-Scenario: check a base backup from the replica
+Scenario: check a base backup and streaming replication from a replica
 	Given I start postgres0
 	And postgres0 is a leader after 10 seconds
 	And I configure and start postgres1 with a tag clonefrom true
--- a/features/cascading_replication.feature
+++ b/features/cascading_replication.feature
@@ -4,10 +4,10 @@ Feature: cascading replication
 Scenario: check a base backup from the replica
 	Given I start postgres0
 	And postgres0 is a leader after 10 seconds
-	And I start postgres1
+	And I configure and start postgres1 with a tag clonefrom true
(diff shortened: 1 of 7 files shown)


features/cascading_replication.feature
```diff
@@ -4,10 +4,10 @@ Feature: cascading replication
 Scenario: check a base backup from the replica
```
alexeyklyukin: Let's update the scenario to "check a base backup and streaming replication from a replica."

alexeyklyukin:
:+1:

alexeyklyukin:
We still need the functionality of cloning from a specific member or set of members, i.e. those that is located in the same region/availability zone/datacenter as the replica to create. The former clonefrom could implement it, because you could set the node to clone from explicitly.

What we can do here is to agree on a certain tag so that all nodes having this tag are considered close enough to be the clone destination for each other (iff they have also the clonefrom tag set), and provide a new callback that calls a script during startup that will gather and set arbitrary tags on the node. For instance, on AWS the callback can detect the current region and set it as a value of the tag above. This way, we can provide a vendor-independant mechanism to choose the preferred replicas to clone (and also to replicate) from.

CyberDem0n:
I like the idea of making decisions based on some specific "tag". Actually to implement this idea there is no need to have a new callback. Generation of tags could be done from start up script, which generates `config.yaml` (either from scratch or from template).

alexeyklyukin:
The idea behind the programmatically settable tags is that we might not know where the instance is running at the time we generate the patroni.yaml. Imagine the Spilo appliance running on top of Taupage, it will have the same configuration no matter which region we start the Spilo on, and then can detect the region via a callback script and sets its value for a "locality" tag.

PR #168, opened 2016-04-04
Read Patroni configuration from the environment.

alexeyklyukin:
The environment variable name is PATRONI_CONFIGURATION.
If both the filename and the variable supplied, the file takes priority.

This is useful to generate the Patroni configuration dynamically.


feikesteenbergen:
:+1:

avaczi:
+1

Issue #167, opened 2016-03-31
rejoin former master fails


jankeirse:
I'm not entirely sure if this is expected behaviour or not. If I stop a master (systemctl stop patroni) a slave takes over. This works fine. If I then want to rejoin the master as slave using just systemctl start patroni it fails: 

> < 2016-03-31 09:13:38.119 CEST >ERROR:  requested starting point 0/26000000 on timeline 5 is not in this server's history
> < 2016-03-31 09:13:38.119 CEST >DETAIL:  This server's history forked from timeline 5 at 0/250001A8.

I can fix it by manually running pg_rewind on the former master:
` sudo -u postgres pg_rewind -D /var/lib/pgsql/9.5/data --source-server='host=currentmaster password=pwforpostgres' 
`
After running that command the former master can join again as slave and everything appears to work. Looking at the code I see that there is quite a bit of code there for running pg_rewind already so I am wondering if I have hit a bug or if this is how it is expected to work and I should always manually intervene before rejoining a former master even if it was shut down manually (not crashed.)

alexeyklyukin:
You have to configure the username and the password for pg_rewind first (i.e. https://github.com/zalando/patroni/blob/master/postgres0.yml). Create a pg_rewind section in the patroni configuration file and specify there the superuser username and the password, i.e.
pg_rewind:
    username: postgres
    password: pwforpostgres

I assume you have set either the wal_log_hints or enabled the checksum, because this is a prerequisite to running pg_rewind.

CyberDem0n:
Hmm, usually pg_rewind is not necessary if patroni (postgres) was shut down cleanly.
How it works:

1. `systemctl stop patroni` sends (hopefully) `SIGTERM` to patroni
2. patroni handles this signal and shuts down postgres with: `pg_ctl stop -m fast`
3. after clean shutdown of postgres patroni removes `leader` lock from etcd (or zookeeper)
4. one of replicas acquires leader lock and promote.

What can go wrong?
It could happen that patroni was not able to shut down postgres during 10 seconds (https://github.com/zalando/patroni/blob/master/extras/startup-scripts/patroni.service#L21) and was killed with `SIGKILL`. In this case replica has a chance to promote earlier then former master was shut down.

In order to get full picture and answer all questions I need to get:

1. config yaml file from the former master
2. patroni logs from the former master (starting from the moment when you run  `systemctl stop patroni`)
3. postgres logs from the former master (starting from the moment when you run  `systemctl stop patroni`)

[...]

PR #169, opened 2016-04-05
Add KillMode to avoid issues when restarting

jankeirse:
This fixes https://github.com/zalando/patroni/issues/167


CyberDem0n:
:+1:

PR #171, opened 2016-04-12
Remove restrictions on running pg_rewind.

### LINKED ISSUE #167: rejoin former master fails

alexeyklyukin:
Previously, pg_rewind was called only if a crashed master tried
to rejoin the cluster. It didn't cover the important case of a
master shut down cleanly, but with a combination of a smart
shutdown and subsequently a fast shutdown. Since out pg_rewind
code does not depend on the "uncleanness" of the master's shutdown,
we can call it unconditionally in all cases where the former master
tries to rejoin as a replica.

This resolves  #167.

### DIFF
--- a/patroni/postgresql.py
+++ b/patroni/postgresql.py
@@ -79,7 +79,7 @@ class Postgresql(object):
 
         self._state = 'stopped'
         self._state_lock = Lock()
-        self._role = 'replica' if os.path.exists(self.recovery_conf) else 'master'
+        self._role = self.get_postgres_role_from_data_directory()
         self._role_lock = Lock()
 
         if self.is_running():
@@ -123,6 +123,9 @@ class Postgresql(object):
                 local_address = 'localhost'  # connection via localhost is preferred
                 break
         return local_address + ':' + self.port
+
+    def get_postgres_role_from_data_directory(self):
+        return 'replica' if os.path.exists(self.recovery_conf) else 'master'
 
     @property
     def _connect_kwargs(self):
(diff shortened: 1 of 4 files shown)


patroni/postgresql.py
```diff
@@ -564,13 +560,12 @@ def cleanup_archive_status(self):
     def follow(self, leader, recovery=False):
         if self.check_recovery_conf(leader) and not recovery:
             return True
-
         change_role = self.role == 'master'
```
CyberDem0n: Unfortunately it will not fix https://github.com/zalando/patroni/issues/167
Because role would be 'replica' there

patroni/postgresql.py
```diff
@@ -73,14 +73,13 @@ def __init__(self, config):
...
         self._cursor_holder = None
-        self._need_rewind = False
         self._sysid = None
         self.replication_slots = []  # list of already existing replication slots
         self.retry = Retry(max_tries=-1, deadline=5, max_delay=1, retry_exceptions=PostgresConnectionException)
 
         self._state = 'stopped'
         self._state_lock = Lock()
-        self._role = 'replica'
+        self._role = 'replica' if os.path.exists(self.recovery_conf) else 'master'
```
CyberDem0n: The same statement is used at  https://github.com/zalando/patroni/blob/feature/pg_rewind_unconditionally/patroni/postgresql.py#L350
Should we create method for that?

feikesteenbergen:
👍

CyberDem0n:
:+1:

PR #172, opened 2016-04-13
Use ssl as a default for replication if enabled

feikesteenbergen:
Some setups require all network connections to use SSL.
As the first pg_hba entry is allowing non-ssl replication connections,
we could not enforce this requirement.

This patch changes the behaviour of Patroni for clusters running SSL.
If replication connections without ssl enabled are still to be allowed
to the cluster, they should be explicitly specified in the Patroni configuration.

Some tests were added to ensure we don't write hostssl records if ssl is disabled.

### DIFF
--- a/patroni/postgresql.py
+++ b/patroni/postgresql.py
@@ -448,11 +448,14 @@ class Postgresql(object):
 
     def write_pg_hba(self):
         with open(os.path.join(self.data_dir, 'pg_hba.conf'), 'a') as f:
-            f.write('\nhost replication {username} {network} md5\n'.format(**self.replication))
-            for line in self.config.get('pg_hba', []):
-                if line.split()[0].strip() == 'hostssl' and self.server_parameters.get('ssl', 'off').lower() != 'on':
-                    continue
-                f.write(line + '\n')
+            ssl = self.server_parameters.get('ssl', 'off').lower() == 'on'
+
+            replication_record = '{method} replication {username} {network} md5'.format(
+                                 method='hostssl' if ssl else 'host', **self.replication)
+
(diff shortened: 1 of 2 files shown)


patroni/postgresql.py
```diff
@@ -448,11 +448,14 @@ def check_replication_lag(self, last_leader_operation):
...
-                if line.split()[0].strip() == 'hostssl' and self.server_parameters.get('ssl', 'off').lower() != 'on':
-                    continue
-                f.write(line + '\n')
+            ssl = self.server_parameters.get('ssl', 'off').lower() == 'on'
+
+            replication_record = '{method} replication {username} {network} md5'.format(
+                                 method='hostssl' if ssl else 'host', **self.replication)
+
+            for line in ['', replication_record] + self.config.get('pg_hba', []):
+                if ssl or not line.startswith('hostssl'):
```
a1exsh: So this will reject any `hostnossl` or `host` lines supplied by the user, right?

a1exsh: Oh, it's the other way round: it will reject user-supplied `hostssl` if we don't have SSL enabled.  Any reason to filter this explicitly?  I'd just let it go to the config, it won't work either way.

alexeyklyukin:
:+1:

patroni/postgresql.py
feikesteenbergen: Probably to quickly be able to disable ssl in your development environment. Removing it may be the best, it would require some test configurations to be rewritten.

feikesteenbergen:
@CyberDem0n had a good remark: Why do all this magic anyway?
Good point, will implement a removal of code. closing this PR.

PR #173, opened 2016-04-13
 Disable sudo, reshuffle travis tasks and introduce caching

CyberDem0n:
Without sudo travis is executing build tasks using docker and waiting
time to start build in this case is really small, usually not longer then 10 seconds.

postgresql-9.5 is installed via addons.apt.packages (without sudo)
But ports 5432 and 5433 are busy. So I had to adjust environment.py to
assign port from higher range.

And a few words about build tasks:
First task is used for executing unit tests for all different python versions
The second one is used for executing acceptance tests against etcd
The third one is used for executing acceptance tests against zookeeper

Ccceptance tests are executed with python2.7 and python3.5

In addition that I've introduced caching of python virtual environment. It really helps to reduce time needed to install python modules.

Also this pull request contains small changes in acceptance tests and scheduled failover. It was necessary to make acceptance tests more stable.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -51,15 +51,19 @@ script:
 
         if [[ $TEST_SUITE == "behave" ]]; then
             if [[ $pv != "3.4" ]]; then
+                echo Running acceptance tests using python${pv}
                 if ! PATH=.:$PATH $TEST_SUITE; then
                     grep . features/output/*/*postgres?.*
                 fi
             fi
         else
+            echo Running unit tests using python${pv}
             $TEST_SUITE test
             $TEST_SUITE flake8
         fi
     done
+
+    set +e
 after_success:
   - coveralls
   - if [[ $TEST_SUITE != "behave" ]]; then python-codacy-coverage -r coverage.xml; fi
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,43 +1,65 @@
-sudo: required
+sudo: false
 language: python
+python:
+- "3.5"
 addons:
+  apt:
+    packages:
+    - postgresql-contrib-9.5
   postgresql: "9.5"
 env:
   global:
-    - BOTO_CONFIG='' ETCDVERSION=2.2.5 ZKVERSION=3.4.6
+  - ETCDVERSION=2.3.1 ZKVERSION=3.4.6
   matrix:
-    - TEST_SUITE="python setup.py test"
-    - TEST_SUITE="behave"
-python:
-  - "2.7"
-  - "3.4"
-  - "3.5"
+  - TEST_SUITE="python setup.py"
+  - DCS="etcd" TEST_SUITE="behave"
+  - DCS="exhibitor" TEST_SUITE="behave"
+cache:
+  directories:
+  - $HOME/virtualenv/python2.7.9
+  - $HOME/virtualenv/python3.4.2
+  - $HOME/virtualenv/python3.5.0
 install:
   - |
+    set -e
+
     if [[ $TEST_SUITE == "behave" ]]; then
-        sudo bash -c '
-            /etc/init.d/postgresql stop
-            apt-get -y remove --purge postgresql-9.1 postgresql-9.2 postgresql-9.3 postgresql-9.4
-            apt-get -y autoremove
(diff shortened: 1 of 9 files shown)


.travis.yml
```diff
@@ -1,43 +1,70 @@
...
+        if [[ $DCS == "etcd" ]]; then
+            curl -L https://github.com/coreos/etcd/releases/download/v${ETCDVERSION}/etcd-v${ETCDVERSION}-linux-amd64.tar.gz | tar xz -C . --strip=1 --wildcards --no-anchored etcd
+        fi
+
+        if [[ $DCS == "exhibitor" ]]; then
+            curl -L http://www.apache.org/dist/zookeeper/zookeeper-${ZKVERSION}/zookeeper-${ZKVERSION}.tar.gz | tar xz
+            mv zookeeper-${ZKVERSION}/conf/zoo_sample.cfg zookeeper-${ZKVERSION}/conf/zoo.cfg
+            zookeeper-${ZKVERSION}/bin/zkServer.sh start
+            while true; do
+                echo -e 'HTTP/1.0 200 OK\nContent-Type: application/json\n\n{"servers":["127.0.0.1"],"port":2181}' | nc -l 8181 &> /dev/null
```
alexeyklyukin: What's the purpose of lines 35-37?

.travis.yml
```diff
@@ -1,43 +1,70 @@
...
-python:
-  - "2.7"
-  - "3.4"
-  - "3.5"
+  - TEST_SUITE="python setup.py"
+  - DCS="etcd" TEST_SUITE="behave"
+  - DCS="exhibitor" TEST_SUITE="behave"
+cache:
+  directories:
+  - $HOME/virtualenv/python2.7.9
```
alexeyklyukin: The virtualenv directories use python2.7, python3.4 and python3.5.

.travis.yml
```diff
@@ -1,43 +1,70 @@
...
+                echo -e 'HTTP/1.0 200 OK\nContent-Type: application/json\n\n{"servers":["127.0.0.1"],"port":2181}' | nc -l 8181 &> /dev/null
+            done&
+        fi
     fi
-  - pip install -r requirements.txt
-  - pip install behave codacy-coverage coverage coveralls
+
+    for pv in "2.7" "3.4" "3.5"; do
+        source ~/virtualenv/python${pv}/bin/activate
+        for p in '-r requirements.txt' 'behave codacy-coverage coverage coveralls flake8 mock pytest-cov pytest'; do
```
alexeyklyukin: Should we do 'test-requirements.txt' inside the patroni directory, listing all the test dependencies? Makes it easy to install all pre-requisites for running the test locally.

features/basic_replication.feature
```diff
@@ -6,12 +6,12 @@ Feature: basic replication
     And postgres0 is a leader after 10 seconds
     And I start postgres1
     When I add the table foo to postgres0
-    Then table foo is present on postgres1 after 15 seconds
+    Then table foo is present on postgres1 after 20 seconds
 
   Scenario: check the basic failover
     When I kill postgres0
-    Then postgres1 role is the primary after 30 seconds
+    Then postgres1 role is the primary after 32 seconds
```
alexeyklyukin: 32 seconds instead of 30? Why?

.travis.yml
CyberDem0n: It "emulates" exhibitor REST api

.travis.yml
CyberDem0n: setup.py installs missing dependencies automatically.
That means locally it's possible to run unit tests even without installing any requirements.
`behave` is not covered, but I don't think this is the issue.

features/basic_replication.feature
CyberDem0n: In worst case it takes 30 seconds to expire `leader` key. Plus ~2 seconds for running all health-checks: https://github.com/zalando/patroni/blob/master/patroni/ha.py#L159

.travis.yml
CyberDem0n: These directories are symlinks

alexeyklyukin:
:+1:

PR #174, opened 2016-04-15
Consul support for patroni

CyberDem0n:


patroni/consul.py
```diff
@@ -0,0 +1,231 @@
...
+            try:
+                self._client.kv.put(self.member_path, connection_string, acquire=self._session)
+            except:
+                logger.exception('touch_member')
+        return True
+
+    @catch_consul_errors
+    def attempt_to_acquire_leader(self):
+        ret = self._client.kv.put(self.leader_path, self._name, acquire=self._session)
+        ret or logger.info('Could not take out TTL lock')
```
CyberDem0n: Codacy Issue found: [Expression "(ret) or (logger.info('Could not take out TTL lock'))" is assigned to nothing](https://www.codacy.com/app/cyberdemn/patroni/file/2412073583/issues/source?bid=3127594&fileBranchId=3235575#l180)

patroni/consul.py
```diff
@@ -0,0 +1,231 @@
...
+    def get(self, callback, path, params=None, timeout=None):
+        uri = self.uri(path, params)
+        if timeout is None and isinstance(params, dict) and 'index' in params:
+            timeout = (float(params['wait'][:-1]) if 'wait' in params else 300) + 1
+        return callback(self.response(self.session.get(uri, verify=self.verify, timeout=timeout)))
+
+
+class ConsulClient(base.Consul):
+
+    def connect(self, host, port, scheme, verify=True):
```
CyberDem0n: Codacy Issue found: [Method could be a function](https://www.codacy.com/app/cyberdemn/patroni/file/2412073583/issues/source?bid=3127594&fileBranchId=3235575#l43)

patroni/consul.py
```diff
@@ -0,0 +1,231 @@
...
+        return True
+
+    def client_path(self, path):
+        return super(Consul, self).client_path(path)[1:]
+
+    @staticmethod
+    def member(node):
+        return Member.from_node(node['ModifyIndex'], os.path.basename(node['Key']), node['Session'], node['Value'])
+
+    def _load_cluster(self, timeout=None):
```
CyberDem0n: Codacy Issue found: [Arguments number differs from overridden '_load_cluster' method](https://www.codacy.com/app/cyberdemn/patroni/file/2412073583/issues/source?bid=3127594&fileBranchId=3235575#l106)

patroni/consul.py
```diff
@@ -0,0 +1,231 @@
...
+    def touch_member(self, connection_string, ttl=None):
+        create_member = self.referesh_session()
+        cluster = self.cluster
+        member_exists = cluster and any(m.name == self._name for m in cluster.members)
+        if create_member and member_exists:
+            self._client.kv.delete(self.member_path)
+        if create_member or not member_exists:
+            try:
+                self._client.kv.put(self.member_path, connection_string, acquire=self._session)
+            except:
```
CyberDem0n: Codacy Issue found: [No exception type(s) specified](https://www.codacy.com/app/cyberdemn/patroni/file/2412073583/issues/source?bid=3127594&fileBranchId=3235575#l173)

patroni/consul.py
```diff
@@ -0,0 +1,231 @@
...
+        host, port = config.get('host', '127.0.0.1:8500').split(':')
+        self._client = ConsulClient(host=host, port=port)
+        self._scope = config['scope']
+        self._session = None
+        self._new_cluster = False
+        self._cluster_index = None
+        self.create_or_restore_session()
+
+    def create_or_restore_session(self):
+        index, member = self._client.kv.get(self.member_path)
```
CyberDem0n: Codacy Issue found: [Unused variable 'index'](https://www.codacy.com/app/cyberdemn/patroni/file/2412073583/issues/source?bid=3127594&fileBranchId=3235575#l70)

patroni/consul.py
CyberDem0n: Codacy Issue found: [Expression "(ret) or (logger.info('Could not take out TTL lock'))" is assigned to nothing](https://www.codacy.com/app/cyberdemn/patroni/file/2412073583/issues/source?bid=3127594&fileBranchId=3235575#l180)

patroni/consul.py
CyberDem0n: Codacy Issue found: [Method could be a function](https://www.codacy.com/app/cyberdemn/patroni/file/2412073583/issues/source?bid=3127594&fileBranchId=3235575#l43)

patroni/consul.py
CyberDem0n: Codacy Issue found: [Arguments number differs from overridden '_load_cluster' method](https://www.codacy.com/app/cyberdemn/patroni/file/2412073583/issues/source?bid=3127594&fileBranchId=3235575#l106)

patroni/consul.py
CyberDem0n: Codacy Issue found: [No exception type(s) specified](https://www.codacy.com/app/cyberdemn/patroni/file/2412073583/issues/source?bid=3127594&fileBranchId=3235575#l173)

patroni/consul.py
CyberDem0n: Codacy Issue found: [Unused variable 'index'](https://www.codacy.com/app/cyberdemn/patroni/file/2412073583/issues/source?bid=3127594&fileBranchId=3235575#l70)

patroni/consul.py
```diff
@@ -0,0 +1,237 @@
...
+        self._scope = config['scope']
+        self._session = None
+        self._new_cluster = False
+        self._cluster_index = None
+        self.create_or_restore_session()
+
+    def create_or_restore_session(self):
+        _, member = self._client.kv.get(self.member_path)
+        self._session = (member or {}).get('Session')
+        if self.referesh_session(retry=True):
```
feikesteenbergen: referesh_session -> refresh_session

patroni/consul.py
```diff
@@ -0,0 +1,237 @@
...
+    def client_path(self, path):
+        return super(Consul, self).client_path(path)[1:]
+
+    @staticmethod
+    def member(node):
+        return Member.from_node(node['ModifyIndex'], os.path.basename(node['Key']), node['Session'], node['Value'])
+
+    def _do_load_cluster(self, timeout=None):
+        if not timeout and self._new_cluster:
+            self._new_cluster = False
```
feikesteenbergen: Perhaps rename this one? As it might be a bit confusing.

features/environment.py
```diff
@@ -128,13 +129,18 @@ def _make_patroni_test_config(self, name, dcs, tags):
             config['tags'] = tags
 
         if dcs != 'etcd':
-            etcd = config.pop('etcd')
-            config['zookeeper'] = {'scope': etcd['scope'], 'session_timeout': etcd['ttl'],
-                                   'reconnect_timeout': config['loop_wait']}
-            if dcs == 'exhibitor':
-                config['zookeeper']['exhibitor'] = {'hosts': ['127.0.0.1'], 'port': 8181}
+            dcs_config = config.pop('etcd')
```
alexeyklyukin: Actually, the code in __init__.py from Patroni will take the first one of etcd, zookeeper and consul available, so instead of taking the etcd configuration and changing it to fit consul or zookeeper we should change the sample configuration to have values for all three.

CyberDem0n: How it will help?
It simply will always use etcd...

alexeyklyukin: No, for the behave tests we are parsing the configuration file manually, so instead of:

    dcs_config = config.pop('etcd')
    # do whatever it takes to make a zookeeper configuration out of it
    config['zookeeper'] = dcs_config
 
we would do

    config.pop('etcd')
    config.pop('consul')

leaving only zookeeper config in place.

I think the way we are currently doing this is quite obscure, i.e. I spent some time trying to figure out why do we pop the etcd config for either consul or zookeeper.

alexeyklyukin: As a bonus, we provide all 3 configurations in a sample config file, and we can add a comment stating 'please, comment the configuration for the DCs that are not used'.

CyberDem0n: I don't agree, In sample config files should be only one DCS section available. Otherwise it's really confusing.

patroni/consul.py
```diff
@@ -0,0 +1,230 @@
...
+
+    def _patch_default_timeout(self):
+        request_func = getattr(self.session.request, '__func__' if six.PY3 else 'im_func')
+        defaults_attr_name = '__defaults__' if six.PY3 else 'func_defaults'
+        defaults = list(getattr(request_func, defaults_attr_name))
+        code = request_func.__code__ if six.PY3 else request_func.func_code
+        defaults[code.co_varnames[code.co_argcount - len(defaults):code.co_argcount].index('timeout')] = 5
+        setattr(request_func, defaults_attr_name, tuple(defaults))  # monkeypatching
+
+    def get(self, callback, path, params=None, timeout=None):
```
alexeyklyukin: The get function from std.HTTPClient does not have a timeout parameter (https://github.com/cablehead/python-consul/blob/master/consul/std.py), so it'll always be None and the `if timeout is None` check is redundant. 

I think you should add comments explaining what you are doing here, how about:

    # Set a default timeout for the request.session.request function, that is used
    # internally  by the methods request.session.get, request.session.post and
    # others. We monkey-patch here to avoid overriding each individual method from
    # request.sesson.*. By default, the timeout is not set. It means that a new
    # session may hang almost indefinitely waiting for the server to respond,
    # which is not what we want in Patroni.

And for the get function:

    # The get function is overridden to handle a special case of it being called
    # with an index and wait parameters. That form indicates that a user needs to
    # wait for the given key to change its value, with a wait timeout supplied. We
    # don't want our monkey-patched timeout to be less than the value of the wait
    # parameter, therefore, we set it to either the value of wait or a default of 5 minutes.

features/environment.py
alexeyklyukin: I don't think it would confuse anyone if we add a comment stating the order the DC sections are picked and instruct the user to comment out those sections they don't want. It won't change the default behavior of picking etcd as well. And it will make the code in behave tests much less confusing, since there will be no need to pick up etcd section and transform it somehow to consul or zookeeper. I won't object those changes to be outside of this pull-request though.

patroni/consul.py
```diff
@@ -0,0 +1,230 @@
...
+        except (ConsulException, RequestException):
+            return False
+    return wrapper
+
+
+class Consul(AbstractDCS):
+
+    def __init__(self, name, config):
+        super(Consul, self).__init__(name, config)
+        self.ttl = int((config.get('ttl') or 30)/2)
```
alexeyklyukin: Is there any reason why do you divide the value from the configuration by 2? Should it be documented?

CyberDem0n: https://www.consul.io/docs/agent/http/session.html
> The TTL field is a duration string, and like LockDelay it can use "s" as a suffix for seconds. If specified, it must be between 10s and 86400s currently. When provided, the session is invalidated if it is not renewed before the TTL expires. The lowest practical TTL should be used to keep the number of managed sessions low. When locks are forcibly expired, such as during a leader election, sessions may not be reaped for up to **double** this TTL, so long TTL values (>1 hour) should be avoided. See the session internals page for more documentation of this feature.

My experiments showed that session expires after 2*tll time, so I decided to divide it by two.

patroni/consul.py
```diff
@@ -0,0 +1,230 @@
...
+            path = self.client_path('/')
+            _, results = self._client.kv.get(path, recurse=True)
+
+            if results is None:
+                raise NotFound
+
+            nodes = {}
+            for node in results:
+                node['Value'] = (node['Value'] or b'').decode('utf-8')
+                nodes[os.path.relpath(node['Key'], path)] = node
```
alexeyklyukin: should we decode the value of the Key as well? What if someone specifies the non-ASCII node name in the configuration...

patroni/consul.py
```diff
@@ -0,0 +1,230 @@
...
+            try:
+                self._client.kv.delete(self.member_path)
+            except Exception:
+                return False
+
+        if not create_member and member and data == self._my_member_data:
+            return True
+
+        try:
+            self._client.kv.put(self.member_path, data, acquire=self._session)
```
alexeyklyukin: Given that your session locking behavior is set to "delete" you don't need to delete the member path manually after the new session is created, do you? The locked keys should be dropped automatically.

alexeyklyukin:
👍
I think a couple comments could be added (Basically, describing the monkey-patch, the reasoning behind TTL/2).

patroni/consul.py
CyberDem0n: Don't think that it is necessary (decoding of a key name). At least we are not doing it neither for etcd nor for zookeeper.
But decoding of value is a different story. Consul returns value encoded in base64. Decoding of base64 is done in the python-consul module, but, for python3 base64 decode function returns `bytestring`. Thats why it must be later decoded into normal string.

patroni/consul.py
CyberDem0n: You are right. If the session disappeared from consul it will remove member key.
But... deleting of non-existing key from consul doesn't hurt (consul returns status=success even if there is no such key)

PR #177, opened 2016-04-20
Remove pg_hba injection and filtering

feikesteenbergen:
Previously we explicitly injected a replication record into pg_hba.conf.
This doesn't allow users to explicitly write their configurations.

This change will just write the lines specified by the user.


CyberDem0n:
You forgot to update README.rst

feikesteenbergen:
+1

CyberDem0n:
:+1:

Issue #178, opened 2016-04-20
failover fails if master is down and slave is not 'healthiest'


jankeirse:
I've had a sittuation where I deliberately shut down the only slave and then the master. I then 'lost' the master and restarted the slave, but it refused to become master because it was:
'Following a different leader because i am not the healthiest node'. 
Except that there was no other node (the former master was lost, no other slave exists.) I am not entirely sure why it considered itself unhealthy, but manually promoting with pg_ctl promote worked and got the cluster back up and running. 
I'm afraid I don't have enough details to reproduce the scenario. If I try to replicate what I did again it just works fine.

CyberDem0n:
Hi!
It would be good to see your config files.
I suppose you have in there something like: `postgresql.maximum_lag_on_failover: 1048576 # 1 megabyte in bytes`

Every `ttl` seconds master write it's xlog position (in bytes) to etcd (/service/<scope>/optime/leader)

Replica could be considered healthy only if it is behind master not much then `postgresql.maximum_lag_on_failover` bytes, otherwise you have high chances to lose some data when promoting.

jankeirse:
The configfile used is here: https://github.com/zalando/patroni/files/204183/patroni.yaml.txt
  The slave may indeed have been behind the master, this is found in the logfile for the slave after startup:
```
recovery command file "recovery.conf" specified neither primary_conninfo nor restore_command
The database server will regularly poll the pg_xlog subdirectory to check for files placed there.
```
Strangely the recovery.conf file does contain the proper restore_command now.

CyberDem0n:
Do you run both patroni/postgres on the same host?

jankeirse:
There are 2 virtual machines (on different physical machines.) They both have 1 instance of patroni that starts 1 instance of postgresql. They share /pgarchive/archive/ (which is an NFS share.)

CyberDem0n:
Fixed in master

jankeirse:
Thanks!

PR #179, opened 2016-04-20
Append values from postgresql.recovery_conf to recovery.conf

CyberDem0n:
Previously these values were appended only when the leader was known and
it was preventing replica from restoring xlogs with `restore_command`.

This pull request is intended to solve https://github.com/zalando/patroni/issues/178


feikesteenbergen:
👍

feikesteenbergen:
+1

PR #183, opened 2016-04-21
Refactor api.py

CyberDem0n:
All responses to the client sent by single method `_write_response`
encode('utf-8') is done only inside this method. It make easier to
support existing code because eliminates need to put b'' everywhere


feikesteenbergen:
+1

PR #185, opened 2016-04-22
Implement noloadbalance support

CyberDem0n:
Mostly this tag is necessary to give a hint to load balancer
auto-configuration tool that node should not be included into
LB configuration.
In addition to that Patroni also should not return status_code=200
for a health check if the tag is present and value is not `False`.


feikesteenbergen:
+1

PR #188, opened 2016-04-23
urllib3.exceptions.HTTPError fixes for python 3.5.1

CyberDem0n:
Somehow when you import only urllib3 it's not possible work with
urllib3.exceptions.HTTPError exception (it looks like it is imported
from some other place. from urllib3.exceptions import HTTPError solves
the problem.

And refactoring of watch method, while loop is not necessary there.

### DIFF
--- a/patroni/etcd.py
+++ b/patroni/etcd.py
@@ -6,13 +6,13 @@ import random
 import requests
 import socket
 import time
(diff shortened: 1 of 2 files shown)


feikesteenbergen:
+1

PR #192, opened 2016-05-09
Reset _sysid and don't call pg_controldata when restore of backup in progress

CyberDem0n:
Otherwise there were some errors in a log from rest-api healthcheck endpoint

### DIFF
--- a/patroni/postgresql.py
+++ b/patroni/postgresql.py
@@ -261,6 +261,8 @@ class Postgresql(object):
                    for replica_method in replica_methods)
 
     def create_replica(self, clone_member, env):
+        self.set_state('creating replica')
+        self._sysid = None
         # create the replica according to the replica_method
         # defined by the user.  this is a list, so we need to
         # loop through all methods the user supplies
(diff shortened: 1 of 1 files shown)


patroni/postgresql.py
```diff
@@ -78,6 +78,7 @@ def __init__(self, config):
         self.replication_slots = []  # list of already existing replication slots
         self.retry = Retry(max_tries=-1, deadline=5, max_delay=1, retry_exceptions=PostgresConnectionException)
 
+        self._creating_replica = False
```
alexeyklyukin: I think `create_replica_in_progress` would be a more meaningful name.

alexeyklyukin: And it would be better to return this status via the API as well instead of sysid == 0 nonsense.

CyberDem0n: Ok, I've found a better solution.
We should use `Postgresql.state` property for that. It is already exposed via REST API.
And regarding `sysid == 0`: if sysid is not defined  it is not exposed.

alexeyklyukin: Good idea, so in addition to the running and stopped we should also have something like "initializing"

alexeyklyukin:
👍

PR #193, opened 2016-05-09
Set application_name to node name in primary_conninfo

CyberDem0n:
It will make it simplier to identify node the in pg_stat_replication view


alexeyklyukin:
👍

Issue #190, opened 2016-05-09
Do not wipe-out the data directory if the replica creation method returned 0, but the start failed


alexeyklyukin:
We had an issue in the field:
- the replica has been initialized successfully from a multi-GB backup
- the replica failed to start, because max_connections there were set to the lower master than on the master
- the replica is wiped out and the backup is started anew.

Should we change the logic to reinitialize the replica only if the replica method fails, and do nothing if the actual start was unsuccessful?

PR #195, opened 2016-05-09
Remove data directory only if replica creation failed

CyberDem0n:
And follow the right node after replica creation (it was following the same node from which it took the backup).

Fixes https://github.com/zalando/patroni/issues/190

### DIFF
--- a/patroni/ha.py
+++ b/patroni/ha.py
@@ -67,11 +67,13 @@ class Ha(object):
         self.dcs.touch_member(json.dumps(data, separators=(',', ':')))
 
     def clone(self, clone_member=None, msg='(without leader)'):
-        if self.state_handler.bootstrap(cluster_initialized=True, clone_member=clone_member):
+        if self.state_handler.clone(clone_member):
             logger.info('bootstrapped %s', msg)
+            cluster = self.dcs.get_cluster()
+            node_to_follow = self._get_node_to_follow(cluster)
+            self.state_handler.follow(node_to_follow, True)
         else:
             logger.error('failed to bootstrap %s', msg)
-            self.state_handler.stop('immediate')
             self.state_handler.remove_data_directory()
 
     def bootstrap(self):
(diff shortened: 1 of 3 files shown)


patroni/ha.py
```diff
@@ -67,11 +67,14 @@ def touch_member(self):
         self.dcs.touch_member(json.dumps(data, separators=(',', ':')))
 
     def clone(self, clone_member=None, msg='(without leader)'):
-        if self.state_handler.bootstrap(cluster_initialized=True, clone_member=clone_member):
+        if self.state_handler.clone(clone_member):
             logger.info('bootstrapped %s', msg)
+            node_to_follow = self._get_node_to_follow(self.dcs.get_cluster())
+            if node_to_follow and node_to_follow.name == self.state_handler.name:
```
alexeyklyukin: Should we do the check for node_to_follow.name == self.state_handler.name inside the _get_node_to_follow and just return None in this case?

patroni/postgresql.py
```diff
@@ -126,7 +126,12 @@ def get_local_address(self):
         return local_address + ':' + self.port
 
     def get_postgres_role_from_data_directory(self):
```
alexeyklyukin: This function is called from the set_role, so it should return only those roles accepted by set_role (which is not 'uninitialized').

CyberDem0n: `set_role` doesn't apply any restrictions on value

patroni/ha.py
CyberDem0n: fixed

patroni/postgresql.py
```diff
@@ -693,41 +700,32 @@ def sync_replication_slots(self, cluster):
...
+        ret = self.create_replica(clone_member) == 0
+        if ret:
+            self.delete_trigger_file()
+            self.restore_configuration_files()
+        return ret
 
+    def bootstrap(self):
+        """
             Failure during initdb always leads to an exception, since there is
             no point in continuing if initdb fails. For the rest of the cases,
```
alexeyklyukin: This comment is outdated, since this function is not returning False anymore

alexeyklyukin:
+1

PR #196, opened 2016-05-11
Write postgres options to postgresql.conf

CyberDem0n:
Originally we were passing postgresql options as an argument of `pg_ctl
start`. It was nice and convenient because doesn't require to touch
configuration files but this method has one significant drawback: it
wasn't possible to change values of options which were passed as an
arguments without restart (event for the case when option reqires only
reload). Instead of doing that (passing options as arguments) we will:
1) rename original postgresql.conf to postgresql-base.conf
2) write options into postgresql.conf which has `include
  'postgresql-base.conf'` on the first line

In addition to that this commit makes some attributes of `Postgresql`
class private (prefixes them with '_')

### DIFF
--- a/patroni/postgresql.py
+++ b/patroni/postgresql.py
@@ -44,10 +44,11 @@ class Postgresql(object):
     def __init__(self, config):
         self.config = config
         self.name = config['name']
-        self.server_parameters = config.get('parameters') or {}
+        self._server_parameters = self.get_server_parameters(config)
+        self._listen_addresses, self._port = (config['listen'] + ':5432').split(':')[:2]
+
         self.scope = config['scope']
-        self.listen_addresses, self.port = config['listen'].split(':')
-        self.data_dir = config['data_dir']
+        self._data_dir = config['data_dir']
         self.replication = config['replication']
         self.superuser = config.get('superuser') or {}
         self.admin = config.get('admin') or {}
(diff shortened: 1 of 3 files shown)


patroni/postgresql.py
```diff
@@ -57,15 +58,17 @@ def __init__(self, config):
...
-        self.trigger_file = config.get('recovery_conf', {}).get('trigger_file') or 'promote'
-        self.trigger_file = os.path.abspath(os.path.join(self.data_dir, self.trigger_file))
-
-        self._pg_ctl = ['pg_ctl', '-w', '-D', self.data_dir]
+        self._schedule_load_slots = self.use_slots
+        self._postgresql_conf = os.path.join(self._data_dir, 'postgresql.conf')
+        self._postgresql_base_conf_name = 'postgresql-base.conf'
+        self._postgresql_base_conf = os.path.join(self._data_dir, self._postgresql_base_conf_name)
+        self._recovery_conf = os.path.join(self._data_dir, 'recovery.conf')
+        self._configuration_to_save = (self._postgresql_conf, os.path.join(self._data_dir, 'pg_hba.conf'))
```
alexeyklyukin: If we don't include postgresql_base_conf here, than we'll break WAL-E restore from the backup (it will backup postgesql.conf, which would include postgresql-base.conf, but not postgresql-base.conf, since it excludes *.conf from the PGDATA, resulting in copying the wrong postgresql.conf to postgresql-base.conf).

alexeyklyukin:
+1

PR #202, opened 2016-05-25
Feature/dynamic configuration

CyberDem0n:


tests/test_config.py
```diff
@@ -0,0 +1,30 @@
...
+
+class TestConfig(unittest.TestCase):
+
+    @patch('os.path.isfile', Mock(return_value=True))
+    @patch('json.load', Mock(side_effect=Exception))
+    @patch.object(builtins, 'open', MagicMock())
+    def setUp(self):
+        self.config = Config(config_env='postgresql: {data_dir: foo}')
+
+    def test_reload_local_configuration(self):
```
CyberDem0n: Codacy Issue found: [Method could be a function](https://www.codacy.com/app/cyberdemn/patroni/file/2630856527/issues/source?bid=3127594&fileBranchId=3318079#l16)

patroni/postgresql.py
```diff
@@ -92,18 +107,70 @@ def __init__(self, config):
...
+    def get_major_version(self):
+        if not self.data_directory_empty():
+            try:
+                with open(os.path.join(self._data_dir, 'PG_VERSION')) as f:
+                    return float(f.read())
+            except Exception:
+                logger.exception('Failed to read PG_VERSION from %s', self._data_dir)
+        return 0.0
+
+    def get_server_parameters(self, config):
```
CyberDem0n: Codacy Issue found: [Method could be a function](https://www.codacy.com/app/cyberdemn/patroni/file/2630856516/issues/source?bid=3127594&fileBranchId=3318079#l123)

patroni/postgresql.py
```diff
@@ -90,39 +107,58 @@ def __init__(self, config):
...
+                for r in self.query("""SELECT name, setting, context
                                          FROM pg_settings
-                                        WHERE context in ('internal', 'postmaster')
-                                          AND name IN (""" + ', '.join('%s' for _ in changes.keys()) + ')',
+                                        WHERE name IN (""" + ', '.join('%s' for _ in changes.keys()) + ')',
                                     *(list(changes.keys()))):
                     if server_parameters[r[0]] is None or str(server_parameters[r[0]]) != str(r[1]):
-                        self._restart_pending = True
-                        break
+                        reload_pending = True
```
alexeyklyukin: reload_pending will always be set even for those parameters that are pending restart.

patroni/config.py
```diff
@@ -0,0 +1,168 @@
...
+                    fd = None
+                    json.dump(self.dynamic_configuration, f)
+                tmpfile = os.rename(tmpfile, self._cache_file)
+                self._cache_needs_saving = False
+            except Exception:
+                logger.exception('Exception when saving file: %s', self._cache_file)
+                if fd:
+                    try:
+                        os.close(fd)
+                    except Exception:
```
alexeyklyukin: I think OSError would be better here.

CyberDem0n: Not really better, this method is called from main HA loop and in this case any other exception will break the loop....

patroni/dcs/etcd.py
```diff
@@ -215,6 +215,13 @@ def get_etcd_client(config):
                 sleep(5)
         return client
 
+    def set_ttl(self, ttl):
+        ttl = int(ttl)
+        if self._ttl != ttl:
+            self.reset_cluster()
```
alexeyklyukin: I suppose setting the even should wake up the other members to re-read the new TTL from the DCS-based configuration, but why reset_cluster()?

CyberDem0n: This is a trick to update TTL of leader and member keys in etcd. Update is happening only when `update_leader` and `touch_member` are called. I don't want to call them explicitly, but want to run next iteration of ha loop immediately.

`set_ttl` is called from `Patroni.reload_config`, after ha_loop, but before `schedule_next_run`, which calls `dcs.watch`

`watch` method is implemented is such way that it will not "watch" for "leader" key if cluster object is None, but will wait for event `AbstractDCS.watch` method. Line 222 is taking care about generating this event.

alexeyklyukin: This is exactly the kind of explanation that should become the comment to this method.

patroni/api.py
```diff
@@ -65,15 +68,15 @@ def check_auth_header(self):
 
     def _write_status_response(self, status_code, response, options=False):
```
alexeyklyukin: 'options' parameter should have a less confusing name, how about 'method_OPTIONS'?

patroni/api.py
```diff
@@ -110,6 +113,45 @@ def do_GET_patroni(self):
...
+                RestApiHandler._patch_config(config[name], value)
+            elif value is None:
+                config.pop(name, None)
+            else:
+                config[name] = value
+
+    @check_auth
+    def do_PATCH_config(self):
+        content_length = int(self.headers.get('content-length', 0))
+        request = json.loads(self.rfile.read(content_length).decode('utf-8'))
```
alexeyklyukin: This should process the failure of json decoder correctly. We want to return 400 BAD REQUEST on this.

patroni/api.py
```diff
@@ -110,6 +113,45 @@ def do_GET_patroni(self):
...
+        cluster = self.server.patroni.ha.dcs.get_cluster()
+        data = cluster.config.data.copy()
+        RestApiHandler._patch_config(data, request)
+        if data != cluster.config.data:
```
alexeyklyukin: That would apply such "changes" as quoting the value, i.e.

http PATCH http://127.0.0.1:8008/config <<<'{"ttl": "20"}'
HTTP/1.0 200 OK

http PATCH http://127.0.0.1:8008/config <<<'{"ttl": "20"}'
HTTP/1.0 200 OK

patroni/api.py
```diff
@@ -110,6 +113,45 @@ def do_GET_patroni(self):
...
+            self.server.patroni.ha.dcs.set_config_value(json.dumps(data, separators=(',', ':')), cluster.config.index)
+            self._write_json_response(200, data)
```
alexeyklyukin: Would it make sense to always schedule the reload after successfully changing the configuration?

patroni/api.py
```diff
@@ -110,6 +113,45 @@ def do_GET_patroni(self):
...
+        else:
+            self._write_response(304, '', '')
+
+    @check_auth
+    def do_POST_reload(self):
+        try:
+            if not self.server.patroni.config.reload_local_configuration(True):
+                return self._write_response(304, '', '')
```
alexeyklyukin: Should we also have some endpoints that one can query to determine if the restart is necessary after the parameter change? Right now, AFAIK one can submit the changes to the configuration but there is no way to ask patroni whether those were applied after the reload....

patroni/api.py
CyberDem0n: What is your suggestion?

patroni/api.py
CyberDem0n: Any GET request which returns status of a given node (GET /, GET /replica, GET /master, GET /patroni) will have "restart_pending": true in this case.

patroni/api.py
CyberDem0n: Any GET request which returns status of a given node (GET /, GET /replica, GET /master, GET /patroni) will have "restart_pending": true in this case.

patroni/api.py
CyberDem0n: It will do it on the next iteration of HA loop.

POST /reload endpoint serves different purpose, it reloads configuration from the config.yml file.

patroni/api.py
alexeyklyukin: The easiest way would be to go through each key in a dict (including nested ones) and compare the values directly, stripping the quotes when necessary. This would also give you specific parameters that were changed. Then, given this information, you can ask PG whether those parameters can be applied after reload, and if not, indicate that the restart is required (it's better to do the latter as a response to reload).

alexeyklyukin: This only makes sense if it prevents unnecessary restarts. For reloads, we don't care much.

CyberDem0n: "stripping the quotes when necessary" -- and how are you going to compare `int` and `string`?
Unfortunately there is no simple solution for this task, because one need to know data type for each parameter or at least to know data types of some of the parameters, like "ttl": int, "loop_wait": int, "use_slots": bool and so on. Everything except could be compared after casting old and new value to string.

Reload of everything is done asynchronously, on the next iteration of HA loop.

`POST /reload` endpoint is an equivalent of SIGHUP, but again, it is not performing real `pg_ctl reload`, but re-reads config.yml. Actual reload will happen asynchronously, on the next iteration.

alexeyklyukin: >  and how are you going to compare int and string?
I assume that you won't get int and string together for a single parameter value, would you? Otherwise, I don't understand your question. You won't catch some more complicated cases, like changing '1MB' shared buffers to '1024KB', but this is not the goal: the goal is to avoid unnecessary restarts when the value being changed is accidentally quoted...

patroni/config.py
```diff
@@ -0,0 +1,185 @@
...
+                    return True
+            except Exception:
+                logger.exception('Exception when reloading local configuration from %s', self.config_file)
+                if dry_run:
+                    raise
+
+    def _process_postgresql_parameters(self, parameters, is_local=False):
+        ret = {}
+        for name, value in (parameters or {}).items():
+            if (is_local and name not in self.__DEFAULT_CONFIG['postgresql']['parameters']) \
```
alexeyklyukin: Why do we allow overriding DEFAULT_CONFIG options only from DCS, but not locally?
We shouldn't allow changing options not mentioned below, port and listen_address, altogether.

alexeyklyukin: Ok, the answer is that in postgresql.py both listen_address and port are completely ignored when they are applied. Still, I think it makes sense to mention them here as well.

patroni/postgresql.py
```diff
@@ -627,26 +724,19 @@ def promote(self):
...
 
-    def create_or_update_role(self, name, password, options):
+    def create_or_update_user(self, name, password, options):
         self.query("""DO $$
 BEGIN
     SET local synchronous_commit = 'local';
     PERFORM * FROM pg_authid WHERE rolname = %s;
     IF FOUND THEN
-        ALTER ROLE "{0}" WITH LOGIN {1} PASSWORD %s;
+        ALTER USER "{0}" WITH {1} PASSWORD %s;
```
alexeyklyukin: ALTER USER does not add LOGIN to the non-login role (CREATE USER does).

patroni/api.py
CyberDem0n: Here we are just making decision, write new data into DCS or not.
Decision about "restart" is made in absolutely different place: https://github.com/zalando/patroni/pull/202/files#diff-d8c1582337c8426ef750420dc6b76949R150

Also DCS contains not only postgres parameters, but some other, like ttl and loop_wait, which must be integers (or at least floats)

patroni/config.py
CyberDem0n: > Why do we allow overriding DEFAULT_CONFIG options only from DCS, but not locally?

Because it's better to have these options exactly the same everywhere.

jberkus:
So I'm a bit concerned about the possible impact of comparing every config value with the DCS tree every single HA loop.  Imagine that someone has their HA poll interval set to 3 seconds because of SLA requirements.  What about having a timestamp instead?  That is, under the config tree, we have a "config_last_updated" timestamp, and we only apply & reload if the local cached timestamp is lower?

CyberDem0n:
Comparison of config values is comparable in performance with json deserialization (it's fast). But if you like, I can optimize it. Instead of putting timestamp into config we can use internal mechanisms of configuration stores. Every configuration store already has information about "transactionId" which created/changed a given key. So first we could compare this transactionId and only after the real value.
Although for me it really looks like premature optimization...

jberkus:
Maybe it is. I just can't help but think of the 200+ configuration variables PostgreSQL has.

Hmmm, does this logic still work if a configuration var has been deleted?  That is, does it reset to default?

CyberDem0n:
> think of the 200+ configuration variables PostgreSQL has.
It is hard to imagine that somebody will change all of them.

> Hmmm, does this logic still work if a configuration var has been deleted? That is, does it reset to default?

It depends. Current implementation keeps the latest available copy of configuration in the data directory. If  config key was completely removed from the "store", current master will restore it from the local copy. Replicas will not apply empty configuration.

jberkus:
See, I knew this was going to end up being really complicated ...

patroni/postgresql.py
```diff
@@ -628,24 +725,21 @@ def promote(self):
...
     ELSE
-        CREATE ROLE "{0}" WITH LOGIN {1} PASSWORD %s;
+        CREATE USER "{0}" WITH {1} PASSWORD %s;
```
feikesteenbergen: ROLE

features/patroni_api.feature
```diff
@@ -13,13 +13,34 @@ Scenario: check API requests on a stand-alone server
...
+        Then I receive a response code 304
+        When I add tag new_tag new_value to postgres0 config
+        And I issue an empty POST request to http://127.0.0.1:8008/reload
+        Then I receive a response code 202
+
+Scenario: check dynamic configuration change via DCS
+        Given I issue a PATCH request to http://127.0.0.1:8008/config with {"ttl": 20, "loop_wait": 1, "postgresql": {"parameters": {"max_connections": 101}}}
+        Then I receive a response code 200
+        And I receive a response loop_wait 1
+        And Response on GET http://127.0.0.1:8008/patroni contains restart_pending after 11 seconds
```
feikesteenbergen: Shall we rename to pending_restart, we'll be compatible with 
9.5+ then:
https://www.postgresql.org/docs/9.5/static/view-pg-settings.html

patroni/postgresql.py
```diff
@@ -92,18 +108,102 @@ def __init__(self, config):
...
+                        unit = changes['wal_segment_size'] if r[0] in ('min_wal_size', 'max_wal_size') else r[2]
+                        new_value = changes.pop(r[0])
+                        if new_value is None or not compare_values(r[3], unit, r[1], new_value):
+                            if r[4] == 'postmaster':
+                                self._pending_restart = True
+                                if r[0] in ('listen_addresses', 'port'):
+                                    listen_address_changed = True
+                            else:
+                                pending_reload = True
+                for param, value in changes.items():
```
CyberDem0n: Codacy Issue found: [Unused variable 'value'](https://www.codacy.com/app/cyberdemn/patroni/file/2684766216/issues/source?bid=3127594&fileBranchId=3318079#l167)

patroni/postgresql.py
```diff
@@ -92,18 +108,102 @@ def __init__(self, config):
...
+            changes.update({p: None for p, v in self._server_parameters.items() if not ('.' in p or p in changes)})
+            if changes:
+                if 'wal_segment_size' not in changes:
+                    changes['wal_segment_size'] = '16384kB'
+                # XXX: query can raise an exception
+                for r in self.query("""SELECT name, setting, unit, vartype, context
+                                         FROM pg_settings
+                                        WHERE name IN (""" + ', '.join(['%s'] * len(changes)) + """)
+                                        ORDER BY 1 DESC""", *(list(changes.keys()))):
+                    if r[4] == 'internal':
```
feikesteenbergen: I vote for not checking internal parameters

patroni/utils.py
```diff
@@ -33,6 +34,121 @@ def calculate_ttl(expiration):
...
+    if value in ('on', 'true', 'yes', '1'):
+        return True
+    if value in ('off', 'false', 'no', '0'):
+        return False
+
+
+def split_int_unit(value, strict=True):
+    value = str(value)
+    l = len(value) - 1
+    while l >= 0 and not value[l].isdigit():
```
alexeyklyukin: isdigit doesn't support hex values.

patroni/config.py
```diff
@@ -0,0 +1,198 @@
...
+                        return not deep_compare(new_configuration, self.__effective_configuration)
+                    self._local_configuration = configuration
+                    self.__effective_configuration = new_configuration
+                    return True
+            except Exception:
+                logger.exception('Exception when reloading local configuration from %s', self.config_file)
+                if dry_run:
+                    raise
+
+    def _process_postgresql_parameters(self, parameters, is_local=False):
```
CyberDem0n: Codacy Issue found: [Method could be a function](https://www.codacy.com/app/cyberdemn/patroni/file/2739774072/issues/source?bid=3127594&fileBranchId=3318079#l133)

alexeyklyukin:
+1

PR #203, opened 2016-05-26
Bugfix: pg_rewind can work only with master

CyberDem0n:


alexeyklyukin:
I'm confused, which problem this "fix" is solving?

CyberDem0n:
pg_rewind creates temporary tables therefore it can work only with master.
Since we support cascading replication `follow` method could get not leader, but some another node to follow, which is obviously in recovery...

alexeyklyukin:
+1

Issue #199, opened 2016-05-18
Password visibility


feikesteenbergen:
Since #79, no filtering of the connection parameters is done, therefore the password can be visible, for example during pg_basebackup:

    /usr/lib/postgresql/9.5/bin/pg_basebackup --pgdata=/home/postgres/pgdata/data --xlog-method=stream \
    --dbname=postgres://standby:PASSWORDHERE@172.31.157.130:5432/postgres

PR #204, opened 2016-05-26
Remove password from backup connection string

CyberDem0n:
It's written into pgpass file anyway and ENV variable is set.

Fixes: https://github.com/zalando/patroni/issues/199


feikesteenbergen:
👍

PR #206, opened 2016-05-30
BUGFIX: demote is already running in a thread

CyberDem0n:
It should not call `recover` but `state_handler.follow` directly
(like it is already done in the `clone`).
Otherwise it tries to create a new thread and all async functionality
became borken...


feikesteenbergen:
👍

PR #207, opened 2016-05-30
Edited README + added new SETTINGS.rst file

LappleApple:
Check it out and let me know what you think. Here is the Settings file with the YAML Config Settings:
https://github.com/zalando/patroni/blob/master/SETTINGS.rst

We still need to do a bit more work on the intro paragraph to make Patroni's advantages/benefits a bit more clear. I also might do a bit more magic with the headers/subheads, to stack things a bit more clearly.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -1,34 +1,64 @@
-|Build Status| |Coverage Status|
-
 Patroni: A Template for PostgreSQL HA with ZooKeeper, etcd or Consul
 ------------------------------------------------------------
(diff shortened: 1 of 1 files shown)


alexeyklyukin:
+1

alexeyklyukin:
I think those are good changes and I've merged them. I think we can mention Spilo as an example of the Docker-based deployment with Patroni. The other thing we can do is to add the "Target audience" section, mentioning there that Patroni is a project for DB Engineers, DevOps, SREs and everyone looking to quickly deploy HA PostgreSQL.

PR #210, opened 2016-06-07
make dbname and config file configurable

misja:
I ran into the issue of the database name (`postgres`) and config file name (`postgresql.conf`) being hardcoded when trying to set up patroni for handling a [pipelinedb](https://www.pipelinedb.com/) cluster. When initialized, pipelinedb will create a default database `pipeline` and a config file named `pipelinedb.conf`, hence a need to have these values configuarable. @jberkus, any chance you could chime in here?

### DIFF
--- a/patroni/postgresql.py
+++ b/patroni/postgresql.py
@@ -60,7 +60,8 @@ class Postgresql(object):
         self.callback = config.get('callbacks') or {}
         self.use_slots = config.get('use_slots', True)
         self._schedule_load_slots = self.use_slots
-        self._postgresql_conf = os.path.join(self._data_dir, 'postgresql.conf')
+        self._postgresql_conf = os.path.join(self._data_dir,
+                                             config.get('config', 'postgresql.conf'))
         self._postgresql_base_conf_name = 'postgresql.base.conf'
         self._postgresql_base_conf = os.path.join(self._data_dir, self._postgresql_base_conf_name)
         self._recovery_conf = os.path.join(self._data_dir, 'recovery.conf')
--- a/patroni/postgresql.py
+++ b/patroni/postgresql.py
@@ -44,6 +44,7 @@ class Postgresql(object):
     def __init__(self, config):
         self.config = config
         self.name = config['name']
+        self.database = config.get('database', 'postgres')
         self._server_parameters = self.get_server_parameters(config)
         self._listen_addresses, self._port = (config['listen'] + ':5432').split(':')[:2]
 
@@ -73,8 +74,8 @@ class Postgresql(object):
 
         self.local_address = self.get_local_address()
         connect_address = config.get('connect_address') or self.local_address
-        self.connection_string = 'postgres://{username}:{password}@{connect_address}/postgres'.format(
-            connect_address=connect_address, **self.replication)
+        self.connection_string = 'postgres://{username}:{password}@{connect_address}/{database}'.format(
+            connect_address=connect_address, database=self.database, **self.replication)
 
         self._connection = None
         self._cursor_holder = None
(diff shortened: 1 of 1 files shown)


CyberDem0n:
I don't have experience with pipelinedb. Are there `pg_ctl` and `pg_controldata` binaries available?
Does it support pg_rewind?

Besides that your change looks fine to me.

jberkus:
@misja : I thougth Pipeline changed stuff to normal names because it was causing issues?  

In general, I'm not keen on making this kind of thing easily configurable; Patroni works partly because it's an opinionated system.  Especially since we're talking *only* pipelineDB with this issue.  So likely as not you'd need a fork to support PipelineDB; it would be a simple patch.

misja:
@CyberDem0n yes, they are available, it is more or less regular postgres but as @jberkus mentioned they did rename binaries so need to create symlinks. @jberkus, I fully agree with you, I'm just trying to get this to work with the current distribution - maybe you could mention this when you're in touch with the pipelinedb people? :wink:

jberkus:
Opened issue with PipelineDB.  Let's keep this PR open until we hear from them.

misja:
thanks @jberkus !

jberkus:
See response here: https://github.com/pipelinedb/pipelinedb/issues/1557

The upshot of this is that we just need to run a post-init script, something which I think that we already have a provision for?

patroni/postgresql.py
```diff
@@ -59,7 +60,8 @@ def __init__(self, config):
         self.callback = config.get('callbacks') or {}
         self.use_slots = config.get('use_slots', True)
         self._schedule_load_slots = self.use_slots
-        self._postgresql_conf = os.path.join(self._data_dir, 'postgresql.conf')
+        self._postgresql_conf = os.path.join(self._data_dir,
+                                             config.get('config', 'postgresql.conf'))
```
CyberDem0n: Oh. May be you didn't noticed, but recently we had changed how Patroni works with config files.
After initializing of new database it renames ``postgresql.conf`` into ``postgresql.base.conf``.
Later it writes into ``postgresql.conf`` something like:
```
include 'postgresql.base.conf'
param1 = 'value1'
param2 = 'value2'
etc...
```

In your case it will rename ``pipelinedb.conf`` into ``postgresql.base.conf``. It will not hurt, but doesn't look nice.

I would propose rename ``config`` into ``config_base_name`` and at the end to have something like:
```diff
@@ -59,8 +59,9 @@ class Postgresql(object):
         self.callback = config.get('callbacks') or {}
         self.use_slots = config.get('use_slots', True)
         self._schedule_load_slots = self.use_slots
-        self._postgresql_conf = os.path.join(self._data_dir, 'postgresql.conf')
-        self._postgresql_base_conf_name = 'postgresql.base.conf'
+        config_base_name = config.get('config_base_name', 'postgresql')
+        self._postgresql_conf = os.path.join(self._data_dir, config_base_name + '.conf')
+        self._postgresql_base_conf_name = config_base_name + '.base.conf'
         self._postgresql_base_conf = os.path.join(self._data_dir, self._postgresql_base_conf_name)
         self._recovery_conf = os.path.join(self._data_dir, 'recovery.conf')
         self._configuration_to_save = (self._postgresql_conf, self._postgresql_base_conf,
```

jberkus: hmmm, if we're renaming the files, then PipelineDB's suggested solution isn't going to work, because the new postgresql.conf won't get picked up.

CyberDem0n: It will work. If the ``config_base_name`` was redefined in the config file Patroni will write not into postgresql.conf, but into pipelinedb.conf

misja: Just pushed the suggested change. Note that I also understand the pipelinedb comments and could work with it, so this pull request really is about making life easier.

feikesteenbergen:
👍

PR #211, opened 2016-06-08
Implement possibility to configure Patroni via environment

CyberDem0n:


docs/ENVIRONMENT.rst
```diff
@@ -0,0 +1,54 @@
+==================================
+Environment Configuration Settings
+==================================
+
+Some of configuration parameters defined in the configuration file is possible to override via Environment variables
```
alexeyklyukin: I'd change the description to the following one:

It is possible to override some of the configuration parameters defined in the Patroni configuration file using the system environment variables. This document lists all environment variables handled by Patroni. The values set via those variables always take precedence over the ones set in the Patroni configuration file.

docs/ENVIRONMENT.rst
```diff
@@ -0,0 +1,54 @@
...
+This document list all possible environment variables handled by Patroni.
+Environment variable always takes precedence on configuration file.
+
+Global/Universal
+----------------
+-  **PATRONI\_NAME**: name of the node where the current instance of Patroni is running. Must be unique for the cluster.
+-  **PATRONI\_NAMESPACE**: path within configuration store where Patroni will keep information about cluster. Default value: "/service"
```
alexeyklyukin: within [the] configuration store ... about [the] cluster.

docs/ENVIRONMENT.rst
```diff
@@ -0,0 +1,54 @@
...
+-  **PATRONI\_SCOPE**: cluster name
+
+Bootstrap configuration
+-----------------------
+It is possible to define users which will be created right after initializing of a new cluster by defining following environment variables:
```
alexeyklyukin: I'd change the following to:

It is possible to create new database users right after the successful initialization of a new cluster. This process is defined by the following variables:

docs/ENVIRONMENT.rst
```diff
@@ -0,0 +1,54 @@
...
+
+-  **PATRONI\_<username>\_PASSWORD='<password>'**
+-  **PATRONI\_<username>\_OPTIONS='list,of,options'**
+
+Example: defining of ``PATRONI_admin_PASSWORD=strongpasswd`` and ``PATRONI_admin_OPTIONS='createrole,createdb'`` will cause creation of user **admin** with the password **strongpasswd**, which is allowed to create other users and databases.
```
alexeyklyukin: Example: defining ``PATRONI_admin_PASSWORD=strongpasswd`` and ``PATRONI_admin_OPTIONS='createrole,createdb'`` will cause creation of the user **admin** with the password **strongpasswd** that is allowed to create other users and databases.

docs/ENVIRONMENT.rst
```diff
@@ -0,0 +1,54 @@
...
+Etcd
+----
+-  **PATRONI\_ETCD\_HOST**: the host:port for the etcd endpoint.
+
+PostgreSQL
+----------
+-  **PATRONI\_POSTGRESQL\_LISTEN**: IP address + port that Postgres listens to. Multiple comma-separated addresses are permitted, as long as the port component is appended after to the last one with a colon, i.e. ``listen: 127.0.0.1,127.0.0.2:5432``. Patroni will use the first address from this list to establish local connections to the PostgreSQL node.
+-  **PATRONI\_POSTGRESQL\_CONNECT\_ADDRESS**: IP address + port through which Postgres is accessible from other nodes and applications.
+-  **PATRONI\_POSTGRESQL\_DATA\_DIR**: file path to initialize and store Postgres data files.
+-  **PATRONI\_POSTGRESQL\_PGPASS**: path to `pgpass` file which would be created by Patroni when it is necessary (for example, before executing pg\_basebackup). This locations must be accessible for writing by Patroni.
```
alexeyklyukin: -  **PATRONI\_POSTGRESQL\_PGPASS**: path to the [.pgpass password file](https://www.postgresql.org/docs/9.3/static/libpq-pgpass.html). Patroni creates this file before executing pg\_basebackup and under some other circumstances. The location must be writable by Patroni.

docs/ENVIRONMENT.rst
```diff
@@ -0,0 +1,54 @@
...
+-  **PATRONI\_REPLICATION\_USERNAME**: replication username; user will be created during initialization. Replicas will use this user to access master via streaming replication
```
alexeyklyukin: the user

docs/ENVIRONMENT.rst
```diff
@@ -0,0 +1,54 @@
...
+-  **PATRONI\_REPLICATION\_PASSWORD**: replication password; user will be created during initialization.
```
alexeyklyukin: the user

docs/ENVIRONMENT.rst
```diff
@@ -0,0 +1,54 @@
...
+-  **PATRONI\_SUPERUSER\_USERNAME**: name for the superuser, set during initialization (initdb) and later used by Patroni to connect to the postgres. Also this user is used by pg_rewind.
+-  **PATRONI\_SUPERUSER\_PASSWORD**: password for the superuser, set during initialization (initdb).
+
+REST API
+-------- 
+-  **PATRONI\_RESTAPI\_CONNECT\_ADDRESS**: IP address and port through which restapi is accessible.
```
alexeyklyukin: IP address and port to access the REST API.

docs/ENVIRONMENT.rst
```diff
@@ -0,0 +1,54 @@
...
+-  **PATRONI\_RESTAPI\_LISTEN**: IP address and port that Patroni will listen to, to provide health-check information for HAProxy.
+-  **PATRONI\_RESTAPI\_USERNAME**: Basic-auth username to protect dangerous REST API endpoints.
+-  **PATRONI\_RESTAPI\_PASSWORD**: Basic-auth password to protect dangerous REST API endpoints.
+-  **PATRONI\_RESTAPI\_CERTFILE**: Specifies a file with the certificate in the PEM format. If the certfile is not specified or is left empty, the API server will work without SSL.
```
feikesteenbergen: As certificates/private keys themselves may be put into the environment variables themselves, we could also allow the PEM formatted certificate/key to be in this variable? Or introduce some variables to accomplish this?
```patch
@@ -47,7 +47,9 @@ REST API
 -  **PATRONI\_RESTAPI\_USERNAME**: Basic-auth username to protect dangerous REST API endpoints.
 -  **PATRONI\_RESTAPI\_PASSWORD**: Basic-auth password to protect dangerous REST API endpoints.
 -  **PATRONI\_RESTAPI\_CERTFILE**: Specifies a file with the certificate in the PEM format. If the certfile is not specified or is left empty, the API server will work without SSL.
+-  **PATRONI\_RESTAPI\_CERTIFICATE**: The public certificate to use for the API server in PEM format.
 -  **PATRONI\_RESTAPI\_KEYFILE**: Specifies a file with the secret key in the PEM format.
+-  **PATRONI\_RESTAPI\_KEY**: The private key to use for the API server in PEM format.
```

docs/ENVIRONMENT.rst
```diff
@@ -0,0 +1,54 @@
...
+-  **PATRONI\_REPLICATION\_USERNAME**: replication username; user will be created during initialization. Replicas will use this user to access master via streaming replication
+-  **PATRONI\_REPLICATION\_PASSWORD**: replication password; user will be created during initialization.
+-  **PATRONI\_SUPERUSER\_USERNAME**: name for the superuser, set during initialization (initdb) and later used by Patroni to connect to the postgres. Also this user is used by pg_rewind.
+-  **PATRONI\_SUPERUSER\_PASSWORD**: password for the superuser, set during initialization (initdb).
+
+REST API
+-------- 
+-  **PATRONI\_RESTAPI\_CONNECT\_ADDRESS**: IP address and port through which restapi is accessible.
+-  **PATRONI\_RESTAPI\_LISTEN**: IP address and port that Patroni will listen to, to provide health-check information for HAProxy.
+-  **PATRONI\_RESTAPI\_USERNAME**: Basic-auth username to protect dangerous REST API endpoints.
```
alexeyklyukin: I'd replace dangerous with unsafe to stick to the REST terminology.

patroni/config.py
```diff
@@ -151,10 +159,87 @@ def _safe_copy_dynamic_configuration(self, dynamic_configuration):
...
+            elif name in config:  # only variables present in __DEFAULT_CONFIG allowed to be overriden from DCS
                 config[name] = int(value)
         return config
 
+    @staticmethod
+    def _build_environment_configuration():
+        ret = defaultdict(dict)
+
+        def _popenv(name):
+            return os.environ.pop('PATRONI_' + name.upper(), None)
```
feikesteenbergen: This will break with ``PATRONI_admin_PASSWORD``, as it will change this into ``PATRONI_ADMIN_PASSWORD`` right?

docs/ENVIRONMENT.rst
```diff
@@ -0,0 +1,54 @@
...
+
+Etcd
+----
+-  **PATRONI\_ETCD\_HOST**: the host:port for the etcd endpoint.
+
+PostgreSQL
+----------
+-  **PATRONI\_POSTGRESQL\_LISTEN**: IP address + port that Postgres listens to. Multiple comma-separated addresses are permitted, as long as the port component is appended after to the last one with a colon, i.e. ``listen: 127.0.0.1,127.0.0.2:5432``. Patroni will use the first address from this list to establish local connections to the PostgreSQL node.
+-  **PATRONI\_POSTGRESQL\_CONNECT\_ADDRESS**: IP address + port through which Postgres is accessible from other nodes and applications.
+-  **PATRONI\_POSTGRESQL\_DATA\_DIR**: file path to initialize and store Postgres data files.
```
alexeyklyukin: The location of the Postgres data directory, either existing or to be initialized by Patroni.

docs/ENVIRONMENT.rst
alexeyklyukin: the file

docs/ENVIRONMENT.rst
```diff
@@ -0,0 +1,54 @@
...
+-  **PATRONI\_SUPERUSER\_PASSWORD**: password for the superuser, set during initialization (initdb).
+
+REST API
+-------- 
+-  **PATRONI\_RESTAPI\_CONNECT\_ADDRESS**: IP address and port through which restapi is accessible.
+-  **PATRONI\_RESTAPI\_LISTEN**: IP address and port that Patroni will listen to, to provide health-check information for HAProxy.
+-  **PATRONI\_RESTAPI\_USERNAME**: Basic-auth username to protect dangerous REST API endpoints.
+-  **PATRONI\_RESTAPI\_PASSWORD**: Basic-auth password to protect dangerous REST API endpoints.
+-  **PATRONI\_RESTAPI\_CERTFILE**: Specifies a file with the certificate in the PEM format. If the certfile is not specified or is left empty, the API server will work without SSL.
+-  **PATRONI\_RESTAPI\_KEYFILE**: Specifies a file with the secret key in the PEM format.
```
alexeyklyukin: the file

patroni/config.py
```diff
@@ -45,7 +46,11 @@ def __init__(self, config_file=None, config_env=None):
         self._config_file = None if config_env else config_file
         self._modify_index = -1
         self._dynamic_configuration = {}
-        self._local_configuration = yaml.safe_load(config_env) if config_env else self._load_config_file()
+        if config_env:
+            self._local_configuration = yaml.safe_load(config_env)
+        else:
+            self.__environment_configuration = self._build_environment_configuration()
```
alexeyklyukin: You can call _build_environment_configuration directly inside the _load_config_file. Make it the staticmethod by passing the self.__config_file as a parameter.

patroni/config.py
```diff
@@ -151,10 +159,87 @@ def _safe_copy_dynamic_configuration(self, dynamic_configuration):
...
+                        value = os.environ.pop(param)
+                        value = value if suffix == 'HOST' else value and _parse_list(value)
+                        if value:
+                            ret[name.lower()][suffix.lower()] = value
+                    # PATRONI_<username>_PASSWORD=<password>, PATRONI_<username>_OPTIONS=<option1,option2,...>
+                    # CREATE USER "<username>" WITH <OPTIONS> PASSWORD '<password>'
+                    elif suffix == 'PASSWORD':
+                        password = os.environ.pop(param)
+                        if password:
+                            users[name] = {'password': password}
```
alexeyklyukin: We should lowercase the name here, so that 2 names that differ only by their case result in a single dictionary key.

patroni/config.py
CyberDem0n: No, it's not possible.
`_load_config_file` can be executed more than once, but `_build_environment_configuration` MUST be executed only once, because it removes all `PATRONI_` variables from `os.environ`. I did it intensionally, because in the future we may want to dump `self.__environment_configuration` into the file and **re-execute** Patroni (just to reset environment).

patroni/config.py
CyberDem0n: it is not used for PATRONI_admin_PASSWORD

patroni/config.py
CyberDem0n: Should we?
Role names in postgres are case sensitive...

alexeyklyukin: Only if you quote them, otherwise the case is ignored.

```
# CREATE ROLE admin;
CREATE ROLE
# CREATE ROLE ADMIN;
ERROR:  role "admin" already exists
```

CyberDem0n: They are quoted: https://github.com/zalando/patroni/blob/master/patroni/postgresql.py#L639

alexeyklyukin: That's wrong. We should not quote role names unconditionally, since it breaks the assumption that an unquoted role name is case-insensitive from the user perspective. We should pass the role as is. If there are no quotes in the name - we should also lowercase it when reading the configuration from the environment variable. If there are quotes - we should pass it unchanged. Actually, the environment variable name cannot contain quotes (and spaces, and perhaps many other characters permitted in the role name), so we should have a separate parameter for the name.

CyberDem0n: > That's wrong. We should not quote role names unconditionally, since it breaks the assumption that an unquoted role name is case-insensitive from the user perspective.

Well, nobody stops user to put into [config](https://github.com/zalando/patroni/blob/feature/dynamic_configuration/specs/patroni_bootstrap.yml#L43) something like:
```yaml
  users:
    admin:
      password: admin
      options:
        - createrole
        - createdatabase
    Admin:
      password: Admin
      options:
        - supeuser
```


> If there are no quotes in the name - we should also lowercase it when reading the configuration from the environment variable. 

Only from environment variables? Behavior is not consistent with the config file....

> If there are quotes - we should pass it unchanged

Usually nobody does it. It requires to write something like:
```bash
FOO='"bar"'
```
what it strange and ugly... The same applies to the yaml file. If the value is quoted, it would be treated as a string and you'll never see quotes there. Unless you will not do the same trick as with shell.

> Actually, the environment variable name cannot contain quotes (and spaces, and perhaps many other characters permitted in the role name)

And I believe this is great! It automatically eliminates a lot of problems :)

alexeyklyukin: >If there are no quotes in the name - we should also lowercase it when reading the configuration >from the environment variable.
>> Only from environment variables? Behavior is not consistent with the config file....

You are right, we should enforce the same rule uniformly for all sources of the Patroni configuration.

>>If there are quotes - we should pass it unchanged
>Usually nobody does it. It requires to write something like:
>
>FOO='"bar"'
>what it strange and ugly... The same applies to the yaml file. If the value is quoted, it would be >treated as a string and you'll never see quotes there. Unless you will not do the same trick as >with shell.

Quoted user names are rarely used as well, so clunkiness of expressing them in the YAML is not really an issue. The idea is to lowercase all user names by default, unless the name contains quotes, in which case it should be passed verbatim to the "CREATE USER". We should also eliminate mandatory quoting from the "CREATE USER/ALTER ROLE" statement arguments.

> And I believe this is great! It automatically eliminates a lot of problems :)

And adds an inconsistency behavior, since it is possible to define  certain user names only in the configuration file, but not when getting the names from the environment variables. Besides, if we introduce a Patroni-specific limit on user names - shouldn't we document it?

feikesteenbergen: I think that not supporting everything is fine, or else we will be restricted by the expressiveness of bash.

alexeyklyukin:
+1

feikesteenbergen:
👍

alexeyklyukin:
+1

PR #213, opened 2016-06-08
Reimplement pg_ctl status in python

CyberDem0n:
subprocess.call was causing problems when the server was running under high load.


alexeyklyukin:
+1

feikesteenbergen:
👍

PR #220, opened 2016-06-16
Feature/patronictl newconf

CyberDem0n:
Make patronictl use config.py for loading config_file. config.py is not only loading config_file but also can build configuration from environment variables.


alexeyklyukin:
+1

Issue #186, opened 2016-04-22
patronictl doesn't send authorization header


CyberDem0n:
Some of Patroni  REST API endpoints (restart, reinitialize, failover) require basic auth (if it is configured) but patronictl always send requests without `Authorization` header and there is no possibility to specify user and password

PR #221, opened 2016-06-16
patronictl will send authorization header if it is configured

CyberDem0n:
username:password can be configured in the 'restapi' section of config file or via environment.

this pull request should solve https://github.com/zalando/patroni/issues/186


alexeyklyukin:
+1

PR #223, opened 2016-06-16
Acceptance tests for patronictl

CyberDem0n:
Call patronictl.py when it's possible instead of doing REST API calls.


alexeyklyukin:
+1

PR #225, opened 2016-06-17
sample haproxy templates for confd

CyberDem0n:
and utilize them in a docker (start confd on the same node with etcd)


jberkus:
I'll poke at this as soon as I can load it up on the test cluster.

alexeyklyukin:
+1

PR #226, opened 2016-06-21
Setup signal handler before creating dcs

CyberDem0n:
Otherwise it was swallowing SysExit exception in an infinite loop.


alexeyklyukin:
+1

PR #227, opened 2016-06-27
Feature/scheduled restarts

alexeyklyukin:
Implement scheduled and conditional restarts. The following conditions are supported for both:

- with_pending_restart_flag: restart only if the pending restart flag is set (because of the configuration change)
- role: restart if the Postgres role is set to a certain value
- postgres_version (x.y.z) - restart if the current Postgres version is less than the one specified.


features/steps/patroni_api.py
```diff
@@ -103,18 +103,35 @@ def scheduled_failover(context, at_url, from_host, to_host, in_seconds):
...
 
 
+@step('I issue a scheduled restart at {url:url} in {in_seconds:d} seconds with {data}')
+def scheduled_restart(context, url, in_seconds, data):
+    data = data and json.loads(data) or {}
+    restart_options = ['"schedule": "{0}"'.format((datetime.now(pytz.utc) + timedelta(seconds=int(in_seconds))).isoformat())]
+    for key in data:
+        if key == 'schedule':
+            continue
+        restart_options.append('"{0}": "{1}"'.format(key, data[key]))
```
a1exsh: Could this for loop be replaced with something like `json.dumps(data.update(schedule=...))`?

patroni/api.py
```diff
@@ -162,18 +169,82 @@ def do_POST_reload(self):
...
+                data = "Unknown filter for the scheduled restart: {0}".format(k)
+                break
+        else:
+            if 'schedule' not in request:
+                try:
+                    status, data = self.server.patroni.ha.restart()
+                    status_code = 200 if status else 503
+                except Exception:
+                    logger.exception('Exception during restart')
+                    data = "Schedule required for the scheduled restart"
```
a1exsh: It sounds odd to mention scheduled restart at this point, since the request didn't mention schedule.

patroni/api.py
```diff
@@ -162,18 +169,82 @@ def do_POST_reload(self):
...
+                else:
+                    data = "Another restart is already scheduled"
+                    status_code = 409
         self._write_response(status_code, data)
 
     @check_auth
+    def do_DELETE_restart(self):
+        self.server.patroni.ha.delete_future_restart()
+        data = "scheduled restart deleted"
+        code = 200
```
a1exsh: Would it not make sense to check if a restart is actually scheduled?

patroni/ha.py
```diff
@@ -286,31 +291,43 @@ def demote(self, delete_leader=True):
...
             now = datetime.datetime.now(pytz.utc)
             try:
-                delta = (failover.scheduled_at - now).total_seconds()
+                delta = (scheduled_at - now).total_seconds()
 
                 if delta > self.patroni.nap_time:
-                    logging.info('Awaiting failover at %s (in %.0f seconds)', failover.scheduled_at.isoformat(), delta)
-                    return
+                    logging.info('Awaiting {0} at %s (in %.0f seconds)'.format(action_name),
+                                 scheduled_at.isoformat(), delta)
```
a1exsh: Why mix format and printf-style?

patroni/ha.py
```diff
@@ -286,31 +291,43 @@ def demote(self, delete_leader=True):
...
                 # The value is very close to now
                 sleep(max(delta, 0))
-                logger.info('Manual scheduled failover at {}'.format(failover.scheduled_at.isoformat()))
+                logger.info('Manual scheduled {0} at %s'.format(action_name), scheduled_at.isoformat())
+                return True
             except TypeError:
-                logger.warning('Incorrect value in of scheduled_at: %s', failover.scheduled_at)
+                logger.warning('Incorrect value in of scheduled_at: %s', scheduled_at)
+                cleanup_fn()
+        return None
```
a1exsh: This is now a tri-state function which can return `True`, `False` or `None`.  Would it make sense to stick to `True/False`?

patroni/ha.py
```diff
@@ -286,31 +291,43 @@ def demote(self, delete_leader=True):
...
+                    return None
 
                 # The value is very close to now
                 sleep(max(delta, 0))
-                logger.info('Manual scheduled failover at {}'.format(failover.scheduled_at.isoformat()))
+                logger.info('Manual scheduled {0} at %s'.format(action_name), scheduled_at.isoformat())
+                return True
             except TypeError:
-                logger.warning('Incorrect value in of scheduled_at: %s', failover.scheduled_at)
+                logger.warning('Incorrect value in of scheduled_at: %s', scheduled_at)
```
a1exsh: "in of"

patroni/ha.py
```diff
@@ -368,20 +385,87 @@ def process_healthy_cluster(self):
...
+            if recent_time and request_time and recent_time != request_time:
+                logger.info("Cancelling scheduled restart: postgres restart has already happened at {0}".
+                            format(recent_time))
+                self.delete_future_restart()
+                return
+
+        if (restart_data and
+           self.should_run_scheduled_action('restart', restart_data['schedule'], self.delete_future_restart)):
+            try:
+                    ret, message = self.restart(restart_data)
```
a1exsh: odd-indented

patroni/ha.py
```diff
@@ -368,20 +385,87 @@ def process_healthy_cluster(self):
...
+                    if ret:
+                        logger.info("Scheduled restart successfull")
```
a1exsh: typo: successful

patroni/api.py
```diff
@@ -112,13 +117,15 @@ def do_GET_config(self):
         else:
             self.send_error(502)
 
-    def _read_json_content(self):
+    def _read_json_content(self, body_is_optional=False):
         if 'content-length' not in self.headers:
-            return self.send_error(411)
+            return self.send_error(411) if body_is_optional else None
```
CyberDem0n: if **not** body_is_optional

tests/test_api.py
```diff
@@ -57,6 +71,8 @@ class MockPatroni(object):
     tags = {}
     version = '0.00'
     noloadbalance = Mock(return_value=False)
+    scheduled_restart = {'schedule': dateutil.parser.parse('2016-08-29 12:45TZ+1'),
+                         'postmaster_start_time': '2016-08-20 12:00TZ+1'}
```
a1exsh: Should have used `postmaster_start_time()` defined above?

patroni/api.py
```diff
@@ -162,18 +169,82 @@ def do_POST_reload(self):
...
     def do_POST_restart(self):
         status_code = 500
         data = 'restart failed'
-        try:
-            status, data = self.server.patroni.ha.restart()
-            status_code = 200 if status else 503
-        except Exception:
-            logger.exception('Exception during restart')
+        request = self._read_json_content(body_is_optional=True)
+        request = request or {}
```
CyberDem0n: _read_json_content sends error to the client and returns None when it fails to extract json from the body

tests/test_api.py
```diff
@@ -167,6 +183,33 @@ def test_do_POST_restart(self):
         with patch.object(MockHa, 'restart', Mock(side_effect=Exception)):
             MockRestApiServer(RestApiHandler, request)
 
+        post = request + '\nContent-Length: '
+        # wrong version
```
a1exsh: wrong role?

tests/test_api.py
```diff
@@ -167,6 +183,33 @@ def test_do_POST_restart(self):
...
+        request = post + '84\n\n{"schedule": "2016-08-20 12:45TZ+1", "role": "unknown", "postgres_version": "9.5.3"}'
+        MockRestApiServer(RestApiHandler, request)
+        # wrong role
```
a1exsh: wrong version?

tests/test_api.py
```diff
@@ -167,6 +183,33 @@ def test_do_POST_restart(self):
...
+        request = post + '85\n\n{"schedule": "2016-08-20 12:45TZ+1", "role": "master", "postgres_version": "9.5.3.1"}'
+        MockRestApiServer(RestApiHandler, request)
+        # unknown filter
+        request = post + '55\n\n{"schedule": "2016-08-29 12:45TZ+1", "batman": "lives"}'
+        MockRestApiServer(RestApiHandler, request)
+        # incorrect schedule
+        request = post + '55\n\n{"schedule": "2016-08-42 12:45TZ+1", "role": "master"}'
+        MockRestApiServer(RestApiHandler, request)
+        # everything fine, but the schedule is missing
+        request = post + '47\n\n{"role": "master", "postgres_version": "9.5.2"}'
```
a1exsh: It would make sense to let the computer calculate content length.

patroni/ha.py
alexeyklyukin: Yeah, from the logic of this function it makes sense to distinguish between the case of "the failover should not be triggered now" aka false and "the failover cannot be triggered altogether", which is None, but since this function does the cleanup by itself in the latter case, I will leave only the True and False.

patroni/api.py
alexeyklyukin: Good catch!

patroni/api.py
alexeyklyukin: Yeah, I will return 404 if the scheduled restart is not there.

patroni/ha.py
alexeyklyukin: The victim of the copy-paste, will fix.

patroni/ha.py
```diff
@@ -286,31 +291,43 @@ def demote(self, delete_leader=True):
...
             # If the value is close to now, we initiate the failover
             now = datetime.datetime.now(pytz.utc)
             try:
-                delta = (failover.scheduled_at - now).total_seconds()
+                delta = (scheduled_at - now).total_seconds()
 
                 if delta > self.patroni.nap_time:
-                    logging.info('Awaiting failover at %s (in %.0f seconds)', failover.scheduled_at.isoformat(), delta)
-                    return
+                    logging.info('Awaiting {0} at %s (in %.0f seconds)'.format(action_name),
```
CyberDem0n: -logging
+logger

don't know why I didn't notice it earlier.

patroni/ha.py
CyberDem0n: Actually I am not only against mixing of `format` and printf-style, but against of using `format` with logging module (debug, info, warning, error, exception, critical).

tests/test_ha.py
```diff
@@ -395,3 +400,11 @@ def test_post_recover(self):
         self.assertEqual(self.ha.post_recover(), 'failed to start postgres')
         self.p.is_running = true
         self.assertIsNone(self.ha.post_recover())
+
+    def test_schedule_future_restart(self):
+        self.ha.patroni.scheduled_restart = {}
+        self.ha.schedule_future_restart({'schedule': '2016-08-30 12:45TZ+1"'})
+        self.ha.schedule_future_restart({'schedule': '2016-08-30 12:45TZ+1"'})
```
slitsche: Why duplicate call? Why no assertion?

alexeyklyukin: Good catch, thank you.

alexeyklyukin: Actually the reason is that the second restart should fail, because the first one is scheduled.

slitsche: Couldn't this be idempotent?

patroni/api.py
```diff
@@ -162,18 +169,86 @@ def do_POST_reload(self):
...
+                if _:
+                    status_code = _
+                    break
+            elif k == 'role':
+                if request[k] not in ('master', 'replica'):
+                    status_code = 400
+                    data = "PostgreSQL role should be either master or replica"
+                    break
+            elif k == 'postgres_version':
+                if not re.match(r'([1-9][0-9]?\.){2}[1-9][0-9]?$', request[k]):
```
CyberDem0n: This regexp will fail with 9.5.0...

It would be better to use something like `r'^[1-9][0-9]?(\.[0-9]+){2}$'` but it also not 100% correct

patroni/api.py
```diff
@@ -162,18 +169,86 @@ def do_POST_reload(self):
...
+                    status_code = 400
+                    data = "PostgreSQL version should be in the first.major.minor format"
+                    break
+            elif k != 'with_pending_restart_flag':
```
CyberDem0n: Why not to use  `restart_pending` (just to be consistent and less typing)?

patroni/ha.py
```diff
@@ -368,20 +387,91 @@ def process_healthy_cluster(self):
...
 
+    def evaluate_scheduled_restart(self):
+        # restart if we need to
+        restart_data = self.future_restart_scheduled()
+        if restart_data:
+            recent_time = self.state_handler.postmaster_start_time()
+            request_time = restart_data['postmaster_start_time']
+            # check if postmaster start time has changed since the last restart
+            if recent_time and request_time and recent_time != request_time:
+                logger.info("Cancelling scheduled restart: postgres restart has already happened at {0}".
```
CyberDem0n: logging module supports printf formatting...
```python
logger.info("Cancelling scheduled restart: postgres restart has already happened at %s", recent_time)
```

patroni/ha.py
```diff
@@ -368,20 +387,91 @@ def process_healthy_cluster(self):
...
+                            format(recent_time))
+                self.delete_future_restart()
+                return
+
+        if (restart_data and
+           self.should_run_scheduled_action('restart', restart_data['schedule'], self.delete_future_restart)):
+            try:
+                ret, message = self.restart(restart_data)
```
CyberDem0n: Restart MUST be done asynchronously...
So far `Ha.restart` method was only called from API, where every request is already performed from a `Thread`, but here it might break ha_loop because restart could be longer than TTL...

patroni/ha.py
```diff
@@ -368,20 +387,91 @@ def process_healthy_cluster(self):
...
+           self.state_handler.postgres_version_to_int(postgres_version) <= int(self.state_handler.server_version)):
+            reason_to_cancel = "postgres version mismatch"
+
+        if pending_restart and not self.state_handler.pending_restart:
+            reason_to_cancel = "pending restart flag is not set"
+
+        if not reason_to_cancel:
+            return True
+        else:
+            logger.info("not proceeding with the restart: {0}".format(reason_to_cancel))
```
CyberDem0n: printf style preferred for logging

patroni/api.py
```diff
@@ -162,18 +169,86 @@ def do_POST_reload(self):
...
+                    status_code = 200 if status else 503
+                except Exception:
+                    logger.exception('Exception during restart')
+                    status_code = 400
+            else:
+                request['postmaster_start_time'] = self.server.patroni.ha.state_handler.postmaster_start_time()
+                if self.server.patroni.ha.schedule_future_restart(request):
+                    data = "Restart scheduled"
+                    status_code = 202
+                else:
```
CyberDem0n: schedule_future_restart can return False when restart_data is not dict

tests/test_api.py
```diff
@@ -57,6 +71,8 @@ class MockPatroni(object):
     tags = {}
     version = '0.00'
     noloadbalance = Mock(return_value=False)
+    scheduled_restart = {'schedule': dateutil.parser.parse('2016-08-29 12:45TZ+1'),
```
CyberDem0n: Very nice time bomb.
Will hit us at the end of summer...

tests/test_ha.py
```diff
@@ -80,6 +81,8 @@ def __init__(self, p, d):
         self.replicatefrom = None
         self.api.connection_string = 'http://127.0.0.1:8008'
         self.clonefrom = None
+        self.scheduled_restart = {'schedule': dateutil.parser.parse('2016-08-29 12:45TZ+1'),
```
CyberDem0n: Time bomb

tests/test_ha.py
```diff
@@ -395,3 +401,42 @@ def test_post_recover(self):
...
+        self.assertFalse(self.ha.schedule_future_restart({'schedule': '2016-08-30 12:45TZ+1"'}))
+
+    def test_delete_future_restarts(self):
+        self.ha.delete_future_restart()
+
+    def test_evaluate_scheduled_restart(self):
+        self.p.postmaster_start_time = Mock(return_value='2016-08-31 12:45TZ+1')
+        with patch.object(self.ha,
+                          'future_restart_scheduled', Mock(return_value={'postmaster_start_time': '2016-08-30 12:45TZ+1',
+                                                                         'schedule': '2016-08-31 12:45TZ+1'})):
```
CyberDem0n: time bomb

tests/test_ha.py
```diff
@@ -395,3 +401,42 @@ def test_post_recover(self):
...
+            self.assertIsNone(self.ha.evaluate_scheduled_restart())
+        with patch.object(self.ha,
+                          'future_restart_scheduled', Mock(return_value={'postmaster_start_time': '2016-08-31 12:45TZ+1',
+                                                                         'schedule': '2016-08-31 12:45TZ+1'})):
```
CyberDem0n: time bomb

patroni/api.py
alexeyklyukin: Good catch. Actually, there is no way for it not to be a dict, because of line 232. I'll remove the check inside the function.

patroni/postgresql.py
```diff
@@ -964,3 +971,37 @@ def basebackup(self, conn_url, env):
...
+    @staticmethod
+    def postgres_version_to_int(pg_version):
+        """ Convert the server_version to integer
+
+        >>> Postgresql.postgres_version_to_int('9.5.3')
+        90503
+        >>> Postgresql.postgres_version_to_int('9.3.13')
+        90313
+        >>> Postgresql.postgres_version_to_int('10.1')
+        100100
```
feikesteenbergen: I believe this would become 100001, see:
http://www.databasesoup.com/2016/05/changing-postgresql-version-numbering.html

alexeyklyukin: Good point, thanks!

feikesteenbergen:
:+1:

alexeyklyukin:
👍

PR #229, opened 2016-06-29
Document that every single zookeeper host:port MUST be quoted

CyberDem0n:
otherwise yaml library can not parse the list.
And make visible yaml exception when trying to parse this list.


feikesteenbergen:
+1

PR #230, opened 2016-06-29
Try to cover as much as possible pg_rewind corner-cases

CyberDem0n:
rewind is not possible when:
1) trying to rewind from themself
2) leader is not reachable
3) leader is_in_recovery

All these cases were leading to removing of data directory...
In all cases except 1) it should "retry" when leader will became
available and not is_in_recovery.


patroni/postgresql.py
```diff
@@ -740,24 +765,30 @@ def follow(self, member, leader, recovery=False):
                 os.unlink(self._recovery_conf)
```
feikesteenbergen: os.unlink will work on files as well as links

alexeyklyukin:
+1

PR #231, opened 2016-06-29
Fix retry logic in etcd.py

CyberDem0n:
Client class takes care about retrying when connection to the etcd node
fails. It calculates amount of retries and timeout depending on etcd
cluster size.

Etcd class should not retry when EtcdConnectionFailed exception is
raised (this case is already handled in the Client).

Besides that adjust retry timeouts in the Client class.


CyberDem0n:
:+1:

PR #232, opened 2016-07-01
Start readonly when holding leader lock

CyberDem0n:
Not starting of postgres was causeing situation when there were no master running...

### DIFF
--- a/patroni/postgresql.py
+++ b/patroni/postgresql.py
@@ -732,6 +732,7 @@ class Postgresql(object):
 
         if leader and leader.name == self.name:
             self._need_rewind = False
+            member = None
             if self.is_running():
                 return
         else:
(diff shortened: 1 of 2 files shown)


patroni/postgresql.py
```diff
@@ -729,14 +729,13 @@ def follow(self, member, leader, recovery=False):
             return True
 
         change_role = self.role == 'master'
-        self._need_rewind = self._need_rewind or change_role and self.can_rewind
+
+        self._need_rewind = (not leader or leader.name != self.name) \
```
feikesteenbergen: Perhaps make it into a bit more complex 

if
elsif
elsif
else

statement

alexeyklyukin:
+1

PR #234, opened 2016-07-01
Implement different connect strategy for zookeeper

CyberDem0n:
Originally it was trying to connect during session_timeout time.
Such strategy doesn't work good during short network hiccups...


feikesteenbergen:
+1

PR #235, opened 2016-07-01
Volume could be passed to the docker only with absolute path

CyberDem0n:
In addition to that add zookeeper support to the docker.


feikesteenbergen:
👍

PR #243, opened 2016-07-13
Make different kazoo timeouts depend on loop_wait

CyberDem0n:
ping timeout ~ 1/2 * loop_wait
connect_timeout ~ 1/2 * loop_wait

Originally these values were calculated from negotiated session timeout
and didn't worked very well, because it was taking significant time to
figure out that connection is dead and reconnect (up to session timeout)
and not giving us time to retry.


patroni/api.py
```diff
@@ -194,7 +194,7 @@ def do_POST_reinitialize(self):
         self._write_response(status_code, data)
 
     def poll_failover_result(self, leader, candidate):
-        timeout = 10 if self.server.patroni.nap_time < 10 else self.server.patroni.nap_time
+        timeout = 10 if self.server.patroni.ha.dcs.loop_wait < 10 else self.server.patroni.ha.dcs.loop_wait
```
alexeyklyukin: timeout = max(10, self.server.patroni.ha.dcs.loop_wait) ?

alexeyklyukin: Also, dcs is the member of Patroni class.

patroni/dcs/__init__.py
```diff
@@ -269,6 +270,18 @@ def set_ttl(self, ttl):
     def set_retry_timeout(self, retry_timeout):
         """Set the new value for retry_timeout"""
 
+    def set_loop_wait(self, loop_wait):
```
alexeyklyukin: The set_loop_wait is not used outside of the file it's defined, so it should be prefixed with an underscore.

patroni/dcs/zookeeper.py
```diff
@@ -211,14 +239,14 @@ def touch_member(self, data, ttl=None):
...
+                self._client.set_async(self.member_path, data).get(timeout=1)
             else:
-                self._client.retry(self._client.create, path, data, makepath=True, ephemeral=True)
+                self._client.create_async(self.member_path, data, makepath=True, ephemeral=True).get(timeout=1)
             self._my_member_data = data
             return True
         except NodeExistsError:
             try:
-                self._client.retry(self._client.set, path, data)
+                self._client.set_async(self.member_path, data).get(timeout=1)
```
alexeyklyukin: Seems that exception handler only belongs to the case of the failed create_async, we won't get this exception in the set_async case

tests/test_zookeeper.py
```diff
@@ -183,8 +190,12 @@ def test_update_leader(self):
     def test_write_leader_optime(self):
         self.zk.last_leader_operation = '0'
         self.zk.write_leader_optime('1')
+        with patch.object(MockKazooClient, 'create_async', Mock()):
```
alexeyklyukin: create_async is defined as create or Mock, what's the purpose of overriding it here with Mock() ?

alexeyklyukin:
👍

alexeyklyukin:
👍

alexeyklyukin:
👍

CyberDem0n:
:+1:

CyberDem0n:
:+1:

alexeyklyukin:
👍

PR #246, opened 2016-07-27
Manage scheduled restarts using patronictl

ikitiki:


patroni/ctl.py
```diff
@@ -119,15 +120,21 @@ def auth_header(config):
...
     url = urlparse(member.api_url)
     logging.debug(url)
     if 'Content-Type' not in headers:
         headers['Content-Type'] = 'application/json'
-    return requests.post('{0}://{1}/{2}'.format(url.scheme, url.netloc, endpoint),
-                         headers=headers,
-                         data=json.dumps(content) if content else None, timeout=60)
+
+    if request_type == 'get':
+        return requests.delete('{0}://{1}/{2}'.format(url.scheme, url.netloc, endpoint), headers=headers, timeout=60)
```
alexeyklyukin: should bet requests.get

patroni/ctl.py
```diff
@@ -119,15 +120,21 @@ def auth_header(config):
...
+    elif request_type == 'post':
+        return requests.post('{0}://{1}/{2}'.format(url.scheme, url.netloc, endpoint),
+                             headers=headers,
+                             data=json.dumps(content) if content else None, timeout=60)
+    elif request_type == 'delete':
+        return requests.delete('{0}://{1}/{2}'.format(url.scheme, url.netloc, endpoint), headers=headers, timeout=60)
```
alexeyklyukin: Building of URL is the same for all types of requests and can be unified.

patroni/ctl.py
```diff
@@ -235,6 +242,39 @@ def get_cursor(cluster, connect_parameters, role='master', member=None):
...
+        output_members(cluster, cluster_name)
+
+    if role:
+        role_names = [m.name for m in get_all_members(cluster, role)]
+        if member_names:
+            member_names = list(set(member_names) & set(role_names))
+        else:
+            member_names = role_names
+
+    if not member_names:
```
alexeyklyukin: If role_members or member_names had been set before being filtered out by lines 251-256, we should not ask the user again to pick up the member.

patroni/ctl.py
```diff
@@ -235,6 +242,39 @@ def get_cursor(cluster, connect_parameters, role='master', member=None):
...
+        member_names = [click.prompt('Which member do you want to {0} [{1}]?'.format(action,
```
alexeyklyukin: if you set force without role or member_names - it will still ask you for the member, but without showing you the member list first.

patroni/ctl.py
```diff
@@ -430,31 +446,85 @@ def ctl_load_config(cluster_name, config_file, dcs):
     return config, dcs, cluster
 
 
+def check_response(response, member_name, action_name, silent_success=False):
+    if response.status_code != 200:
```
alexeyklyukin: API may return you 202 for the scheduled action.
https://github.com/zalando/patroni/blob/master/patroni/api.py

In general, I'd avoid complaining on anything less than 400.

PR #248, opened 2016-07-28
Manage scheduled restarts using patronictl

ikitiki:


patroni/ctl.py
```diff
@@ -646,3 +712,32 @@ def configure(config_file, dcs, namespace):
     config['dcs_api'] = str(dcs)
     config['namespace'] = str(namespace)
     store_config(config, config_file)
+
+
+@ctl.command('flush', help='Flush scheduled events')
+@click.argument('cluster_name')
+@click.argument('member_names', nargs=-1)
+@click.option('--restart', 'p_restart', help='Flush scheduled restart', is_flag=True)
```
alexeyklyukin: The action in flush is mandatory, therefore, it should not be an option.

patroni/ctl.py
```diff
@@ -430,31 +448,85 @@ def ctl_load_config(cluster_name, config_file, dcs):
...
+
+    return None
+
+
 @ctl.command('restart', help='Restart cluster member')
 @click.argument('cluster_name')
 @click.argument('member_names', nargs=-1)
 @click.option('--role', '-r', help='Restart only members with this role', default='any',
               type=click.Choice(['master', 'replica', 'any']))
 @click.option('--any', 'p_any', help='Restart a single member only', is_flag=True)
```
alexeyklyukin: restart_pending option is missing (restart only if restart is required to apply PostgreSQL configuration changes).

patroni/ctl.py
```diff
@@ -436,25 +484,53 @@ def ctl_load_config(cluster_name, config_file, dcs):
...
+        scheduled = click.prompt('When should the restart take place (e.g. 2015-10-01T14:30) ', type=str, default='now')
+
+    scheduled_at = parse_scheduled(scheduled)
+    if scheduled_at:
+        content['schedule'] = scheduled_at.isoformat()
+
+    for mn, member in members.items():
+        if 'schedule' in content:
+            if force and 'scheduled_restart' in patroni_status(member, auth_header(config)):
+                request_patroni('delete', member, 'restart', None, auth_header(config))
```
alexeyklyukin: What would happen if it fails?

patroni/ctl.py
```diff
@@ -646,3 +746,26 @@ def configure(config_file, dcs, namespace):
...
+def flush(cluster_name, member_names, config_file, dcs, force, role, target):
+    config, dcs, cluster = ctl_load_config(cluster_name, config_file, dcs)
+
+    members = get_members(cluster, cluster_name, member_names, role, force, 'flush')
+    for mn, member in members.items():
+        status = patroni_status(member, auth_header(config))
+        if target == 'restart':
+            if 'scheduled_restart' in status:
+                r = request_patroni('delete', member, 'restart', None, auth_header(config))
+                check_response(r, mn, 'flush scheduled restart')
```
alexeyklyukin: Should we also return some code to the user depending on whether the flush failed or not (since there is a force flag, it might could be used in a non-interactive script).

patroni/ctl.py
```diff
@@ -436,25 +484,53 @@ def ctl_load_config(cluster_name, config_file, dcs):
 @click.option('--role', '-r', help='Restart only members with this role', default='any',
               type=click.Choice(['master', 'replica', 'any']))
 @click.option('--any', 'p_any', help='Restart a single member only', is_flag=True)
+@click.option('--scheduled', help='Timestamp of a scheduled restart in unambiguous format (e.g. ISO 8601)',
+              default=None)
+@click.option('--pg-version', 'version', help='Restart if the PostgreSQL version is less than provided (e.g. 9.5.2)',
```
alexeyklyukin: Still no support for 'restart_pending'.

patroni/ctl.py
alexeyklyukin: Scratch that, I've noticed the whole ctl business isn't returning anything, unless an exception occurs.

alexeyklyukin:
👍

patroni/ctl.py
```diff
@@ -119,15 +120,21 @@ def auth_header(config):
...
+    logging.debug(url_parts)
     if 'Content-Type' not in headers:
         headers['Content-Type'] = 'application/json'
-    return requests.post('{0}://{1}/{2}'.format(url.scheme, url.netloc, endpoint),
-                         headers=headers,
-                         data=json.dumps(content) if content else None, timeout=60)
+
+    url = '{0}://{1}/{2}'.format(url_parts.scheme, url_parts.netloc, endpoint)
+
+    if request_type == 'get':
```
CyberDem0n: getattr(requests, request_type)(url, headers=headers, data=json.dumps(content) if content else None, timeout=60)

patroni/ctl.py
```diff
@@ -436,25 +484,58 @@ def ctl_load_config(cluster_name, config_file, dcs):
...
+    if version is None and not force:
+        version = click.prompt('Restart if the PostgreSQL version is less than provided (e.g. 9.5.2) ',
+                               type=str, default='')
+
+    content = {}
+    if pending:
+        content['restart_pending'] = True
+
+    if version:
+        if not re.match(r'[1-9][0-9]?(\.(0|([1-9][0-9]?))){2}$', version):
```
CyberDem0n: I wonder, should we create function which is doing version check, put it into utils and use it in the api.py and ctl.py to avoid code duplication and improve maintainability?

patroni/ctl.py
```diff
@@ -630,7 +719,23 @@ def members(config_file, cluster_names, fmt, watch, w, dcs):
...
-            output_members(dcs.get_cluster(), cluster_name, fmt)
+            extra = dict()
+            cluster = dcs.get_cluster()
+            if extended:
+                extra = {
+                    'Scheduled restart': dict()
+                }
+                for member in cluster.members:
+                    value = ''
+                    status = patroni_status(member, auth_header(config))
```
CyberDem0n: Does it really make sense to get status of every node?
Information about scheduled restarts is available from DCS...
member.data.get('scheduled_restart')

patroni/ctl.py
ikitiki: at least it would be nice to create precompile regex in the api and use it elsewhere as well

patroni/ctl.py
```diff
@@ -119,15 +120,21 @@ def auth_header(config):
         return {'Authorization': 'Basic ' + base64.b64encode(config['restapi']['auth'].encode('utf-8')).decode('utf-8')}
 
 
-def post_patroni(member, endpoint, content, headers=None):
+def request_patroni(request_type, member, endpoint, content=None, headers=None):
```
CyberDem0n: Order of parameters looks confusing...
Either it should be `member, request_type, endpoint` or `request_type, endpoint, member` (In my opinion request_type and endpoint should be next to each other)

patroni/ctl.py
```diff
@@ -646,3 +751,26 @@ def configure(config_file, dcs, namespace):
...
+              type=click.Choice(['master', 'replica', 'any']))
+@option_config_file
+@option_force
+@option_dcs
+def flush(cluster_name, member_names, config_file, dcs, force, role, target):
+    config, dcs, cluster = ctl_load_config(cluster_name, config_file, dcs)
+
+    members = get_members(cluster, cluster_name, member_names, role, force, 'flush')
+    for mn, member in members.items():
+        status = patroni_status(member, auth_header(config))
```
CyberDem0n: The same as line #730

Sure, this information is not written into DCS immediately, but that could be easily fixed...

patroni/ctl.py
```diff
@@ -235,6 +242,41 @@ def get_cursor(cluster, connect_parameters, role='master', member=None):
...
+    for mn in member_names:
+        if mn not in candidates:
+            raise PatroniCtlException('{0} is not a member of cluster'.format(mn))
+
+    if not force:
+        confirm = click.confirm('Are you sure you want to {0} members {1}?'.format(action, ', '.join(member_names)))
+        if not confirm:
+            raise PatroniCtlException('Aborted {0}'.format(action))
+
+    result = {}
```
CyberDem0n: return {n: candidates[n] for n in member_names}

patroni/ctl.py
```diff
@@ -363,7 +405,7 @@ def query_member(cluster, cursor, member, role, command, connect_parameters):
 def remove(config_file, cluster_name, fmt, dcs):
     _, dcs, cluster = ctl_load_config(cluster_name, config_file, dcs)
 
-    output_members(cluster, cluster_name, fmt)
+    output_members(cluster, cluster_name, None, fmt)
```
CyberDem0n: output_members(cluster, cluster_name, fmt=fmt)

patroni/api.py
```diff
@@ -230,22 +229,30 @@ def do_POST_restart(self):
...
-                    data = "Another restart is already scheduled"
-                    status_code = 409
+                try:
+                    if self.server.patroni.ha.schedule_future_restart(request):
+                        data = "Restart scheduled"
+                        status_code = 202
+                    else:
+                        data = "Another restart is already scheduled"
+                        status_code = 409
+                except Exception as e:
```
CyberDem0n: Actually it doesn't make sense to catch exception here, because`touch_member` is already catching everything related...
Sorry, my bad...

patroni/api.py
```diff
@@ -230,22 +229,30 @@ def do_POST_restart(self):
...
-            data = "no restarts are scheduled"
-            code = 404
+        try:
+            if self.server.patroni.ha.delete_future_restart():
+                data = "scheduled restart deleted"
+                code = 200
+            else:
+                data = "no restarts are scheduled"
+                code = 404
+        except Exception as e:
```
CyberDem0n: The same as for line#239

patroni/ctl.py
```diff
@@ -235,6 +238,37 @@ def get_cursor(cluster, connect_parameters, role='master', member=None):
...
+    for mn in member_names:
+        if mn not in candidates:
+            raise PatroniCtlException('{0} is not a member of cluster'.format(mn))
+
+    if not force:
+        confirm = click.confirm('Are you sure you want to {0} members {1}?'.format(action, ', '.join(member_names)))
+        if not confirm:
+            raise PatroniCtlException('Aborted {0}'.format(action))
+
+    return {n: candidates[n] for n in member_names}
```
CyberDem0n: Why do we need `dict` object here?
Every member already has `name` attribute, so it could be just `list`:
return [candidates[n] for n in member_names]

It also will simplify further work with this list in the `restart`, `reinit` and `flush` methods.

tests/test_api.py
```diff
@@ -220,13 +220,20 @@ def make_request(request):
             with patch.object(MockHa, 'restart', Mock(return_value=(retval, "foo"))):
                 request = make_request('{"role": "master", "postgres_version": "9.5.2"}')
                 MockRestApiServer(RestApiHandler, request)
+            with patch.object(MockHa, 'schedule_future_restart', Mock(side_effect=Exception, return_value=retval)):
```
CyberDem0n: I guess this test is obsolete, since we are not handling exceptions in this place anymore.
Plus the same for line#233

patroni/ctl.py
```diff
@@ -630,7 +703,23 @@ def members(config_file, cluster_names, fmt, watch, w, dcs):
         dcs = get_dcs(config, cluster_name)
 
         for _ in watching(w, watch):
-            output_members(dcs.get_cluster(), cluster_name, fmt)
+            extra = dict()
```
CyberDem0n: There is no need to build a `dict` object here and pass it into `output_members` function.
Information about `scheduled_restart` is available for every member and could be used directly from there.

patroni/ctl.py
```diff
@@ -430,31 +440,88 @@ def ctl_load_config(cluster_name, config_file, dcs):
...
+    for member in members:
+        if 'schedule' in content:
+            if force and member.data.get('scheduled_restart'):
+                r = request_patroni(member, 'delete', 'restart', headers=auth_header(config))
+                check_response(r, member.name, 'flush scheduled restart', True)
+
+        r = request_patroni(member, 'post', 'restart', content, auth_header(config))
+        if r.status_code == 200:
+            click.echo('Successful restart on member {0}'.format(member.name))
+        elif 200 < r.status_code < 300:
```
CyberDem0n: POST /restart will return:
202 - if restart was scheduled successfully
409 - if "Another restart is already scheduled"

ikitiki: shall I change `200 < r.status_code < 300` to just `r.status_code == 202` ?

CyberDem0n: Yep.
And don't forget to handle 409 response.

ikitiki: 409 is handled below. r.text will contain something like "Another restart is already scheduled"

ikitiki: ok, I'll add another `elif` clause for 409

CyberDem0n: It is not, there is a difference between "restart failed" and "schedule of restart failed"

CyberDem0n:
:+1:

features/patroni_api.feature
```diff
@@ -54,10 +54,10 @@ Scenario: check API requests for the primary-replica pair
 	And I receive a response role replica
 	When I run patronictl.py reinit batman postgres1 --force
 	Then I receive a response returncode 0
-	And I receive a response output "Succesful reinitialize on member postgres1"
+	And I receive a response output "Success: reinitialize for member postgres1"
 	When I run patronictl.py restart batman postgres0 --force
 	Then I receive a response returncode 0
-	And I receive a response output "Succesful restart on member postgres0"
+	And I receive a response output "Successful restart on member postgres0"
```
alexeyklyukin: We should also change the restart operation to produce Success: and Failed: messages, the same way as reinitialize.

alexeyklyukin:
👍

alexeyklyukin:
👍

CyberDem0n:
:+1:

ikitiki:
👍

PR #252, opened 2016-08-01
Add patronictl scaffold command.

alexeyklyukin:
Creates the cluster structure in DCS as long as the initialize
key does not exist.

The structure consists of the leader and member keys. Both are
intentionally set to never expire, in order to support running the
cluster with the master that doesn't run the Patroni (external master).

Changes to the DCS code as well, in order to support non-expiring
leader and member keys.

Some silly default settings had to be applied if Patroni is unable
to find the configuration file. In particular, the connect address
will point to the localhost. Perhaps we should avoid running with
wihtout the valid configuration altogether, but currently there is
a valid use-case for this behavior, namely the replicas that are
running with the inaccessible master and getting up-to-date with
WAL segments only.


patroni/ctl.py
```diff
@@ -646,3 +646,54 @@ def configure(config_file, dcs, namespace):
     config['dcs_api'] = str(dcs)
     config['namespace'] = str(namespace)
     store_config(config, config_file)
+
+
+def touch_member(config, dcs):
+    ''' Rip of the ha.touch_member without inter-class dependencies '''
+    p = Postgresql(config['postgresql'])
```
CyberDem0n: Do we really need to create `Postgresql` and `RestApiServer` objects here?

alexeyklyukin: I could just extract the code that creates the connection and api strings out of it, but there are dependencies in it, and I wanted to avoid copying too much code (a part of touch_member from HA looked small and easy to comprehend to just copy it here directly).

patroni/ctl.py
```diff
@@ -646,3 +646,54 @@ def configure(config_file, dcs, namespace):
...
+    config, dcs, cluster = ctl_load_config(cluster_name, config_file, dcs)
+    if cluster and cluster.initialize:
+        raise PatroniCtlException("This cluster is already initialized")
+    dcs.initialize(create_new=True, sysid=sysid)
+
+    set_defaults(config, cluster_name)
+
+    # make sure the leader key will never expire
+    if not (touch_member(config, dcs) and dcs.attempt_to_acquire_leader(permanent=True)):
+        dcs.delete_leader()
```
alexeyklyukin: Should remove member as well

patroni/ctl.py
```diff
@@ -646,3 +646,60 @@ def configure(config_file, dcs, namespace):
...
+    try:
+        dcs.touch_member(json.dumps(data, separators=(',', ':')), permanent=True)
+    except DCSError:
+        return False
+    return True
+
+
+def set_defaults(config, cluster_name):
+    ''' fill-in some basic configuration parameters if config file is not set '''
+    config['postgresql']['name'] = config['postgresql'].get('name') or cluster_name
```
CyberDem0n: config['postgresql'].setdefault('name', cluster_name)

and so on

patroni/ctl.py
```diff
@@ -646,3 +646,60 @@ def configure(config_file, dcs, namespace):
...
+
+    try:
+        # make sure the leader keys will never expire
+        if not (touch_member(config, dcs) and dcs.attempt_to_acquire_leader(permanent=True)):
+            # we did initialize this cluster, but failed to write the leader or member keys, wipe it down completely.
+            raise PatroniCtlException("Unable to install permanent leader for cluster {0}".format(cluster_name))
+    except:
+        dcs.delete_cluster()
+        raise
+    click.echo("Cluster {0} has been created successfully".format(cluster_name, sysid))
```
CyberDem0n: There is no placeholder for sysid...

patroni/dcs/consul.py
```diff
@@ -191,16 +191,18 @@ def touch_member(self, data, **kwargs):
             return True
 
         try:
-            self._client.kv.put(self.member_path, data, acquire=self._session)
+            args = {'cas': None} if kwargs.get('permanent') else {}
```
CyberDem0n: Method signature has to be:
```python
    def touch_member(self, data, permanent=False, **kwargs):
```

`cas`? It's for Compare-And-Save operations...
If you want to create permanent member, you should NOT specify `acquire`:
```python
    args = {} if permanent else {'acquire': self._session}
    self._client.kv.put(self.member_path, data, **args)
```

patroni/dcs/consul.py
```diff
@@ -191,16 +191,18 @@ def touch_member(self, data, **kwargs):
...
         except Exception:
             logger.exception('touch_member')
         return False
 
     @catch_consul_errors
-    def attempt_to_acquire_leader(self):
-        ret = self._client.kv.put(self.leader_path, self._name, acquire=self._session)
+    def attempt_to_acquire_leader(self, permanent=False):
+        args = {'cas': None} if permanent else {}
+        ret = self._client.kv.put(self.leader_path, self._name, acquire=self._session, **args)
```
CyberDem0n: The problem same as in `touch_member`

patroni/dcs/etcd.py
```diff
@@ -281,16 +281,20 @@ def _load_cluster(self):
             raise EtcdError('Etcd is not responding properly')
 
     @catch_etcd_errors
-    def touch_member(self, data, ttl=None):
-        return self.retry(self._client.set, self.member_path, data, ttl or self._ttl)
+    def touch_member(self, data, ttl=None, permanent=False):
+        return self.retry(self._client.set, self.member_path, data, (ttl or self._ttl) if not permanent else None)
```
CyberDem0n: Just a nitpicking:
```python
None if permanent else ttl or self._ttl
```

patroni/dcs/etcd.py
```diff
@@ -281,16 +281,20 @@ def _load_cluster(self):
...
         return self.retry(self._client.set, self.leader_path, self._name, self._ttl)
 
-    def attempt_to_acquire_leader(self):
+    def attempt_to_acquire_leader(self, permanent=False):
         try:
-            return bool(self.retry(self._client.write, self.leader_path, self._name, ttl=self._ttl, prevExist=False))
+            return bool(self.retry(self._client.write,
+                                   self.leader_path,
+                                   self._name,
+                                   ttl=self._ttl if permanent else None,
```
CyberDem0n: ttl=None if permanent else self._ttl,

patroni/ctl.py
```diff
@@ -646,3 +646,60 @@ def configure(config_file, dcs, namespace):
...
+
+@ctl.command('scaffold', help='Create a structure for the cluster in DCS')
+@click.argument('cluster_name')
+@click.option('--sysid', '-s', help='System ID of the cluster to put into the initialize key', default="")
+@option_config_file
+@option_dcs
+def scaffold(cluster_name, config_file, dcs, sysid):
+    logging.debug("config_file = %s, cluster_name = %s, dcs = %s, sysid = %s", config_file, cluster_name, dcs, sysid)
+    config, dcs, cluster = ctl_load_config(cluster_name, config_file, dcs)
+    if cluster and cluster.initialize:
```
CyberDem0n: if cluster and cluster.initialize is not None:

patroni/dcs/consul.py
alexeyklyukin: The other notice is valid, but I don't get the kwargs comment. You've put it there, presumably, in order to avoid complaints about unused parameters, so now we can use it as a proper mechanism to pass the optional params. I think mixing the named param and kwargs would be ugly and confusing, since you'd assume {{permanent}} is the first optional param, while in the original function it is the second one (after the {{ttl}}).

patroni/dcs/etcd.py
alexeyklyukin: Good catch!

alexeyklyukin:
👍

patroni/ctl.py
```diff
@@ -674,7 +674,8 @@ def set_defaults(config, cluster_name):
     config['postgresql'].setdefault('scope', cluster_name)
     config['postgresql'].setdefault('listen', '127.0.0.1')
     config['postgresql']['authentication'] = {'replication': None}
-    config['restapi']['listen'] = ':' in config['restapi'].get('listen', ".") or '127.0.0.1:5432'
+    config['restapi']['listen'] = (config['restapi']['listen']
+                                   if ':' in config['restapi'].get('listen', ".") else '127.0.0.1:5432')
```
CyberDem0n: 'listen' is always present in `config['restapi']`, so just ':' in config['restapi']['listen'] will work

patroni/ctl.py
```diff
@@ -646,3 +646,59 @@ def configure(config_file, dcs, namespace):
...
     store_config(config, config_file)
+
+
+def touch_member(config, dcs):
+    ''' Rip-off of the ha.touch_member without inter-class dependencies '''
+    p = Postgresql(config['postgresql'])
+    p.set_state('running')
+    p.set_role('master')
+
+    api = RestApiServer(None, config['restapi'])
```
CyberDem0n: It will start listening on the REST API port, what means it would be not possible to run `scaffold` if there is already Patroni running on the same machine with the same config.

patroni/ctl.py
```diff
@@ -646,3 +646,59 @@ def configure(config_file, dcs, namespace):
...
+    if not dcs.initialize(create_new=True, sysid=sysid):
+        # initialize key already exists, don't touch this cluster
+        raise PatroniCtlException("Initialize key for cluster {0} already exists".format(cluster_name))
+
+    set_defaults(config, cluster_name)
+
+    try:
+        # make sure the leader keys will never expire
+        if not (touch_member(config, dcs) and dcs.attempt_to_acquire_leader(permanent=True)):
+            # we did initialize this cluster, but failed to write the leader or member keys, wipe it down completely.
```
CyberDem0n: Missing `delete_cluster` call?

patroni/ctl.py
```diff
@@ -646,3 +646,59 @@ def configure(config_file, dcs, namespace):
...
+            raise PatroniCtlException("Unable to install permanent leader for cluster {0}".format(cluster_name))
+    except:
```
CyberDem0n: Do we really need try-except-raise here?

tests/test_ctl.py
```diff
@@ -4,14 +4,20 @@
...
 import unittest
 
+from patroni.api import RestApiServer
+from six.moves import BaseHTTPServer
+
 from click.testing import CliRunner
 from mock import patch, Mock
 from patroni.ctl import ctl, members, store_config, load_config, output_members, post_patroni, get_dcs, parse_dcs, \
-    wait_for_leader, get_all_members, get_any_member, get_cursor, query_member, configure, PatroniCtlException
+    wait_for_leader, get_all_members, get_any_member, get_cursor, query_member, configure, touch_member, set_defaults,\
```
CyberDem0n: touch_member and set_defaults are not used here, no need to import them

patroni/ctl.py
```diff
@@ -646,3 +646,59 @@ def configure(config_file, dcs, namespace):
...
+    api = RestApiServer(None, config['restapi'])
+    data = {
+        'conn_url': p.connection_string,
+        'api_url': api.connection_string,
+        'state': p.state,
+        'role': p.role
+    }
+    try:
+        dcs.touch_member(json.dumps(data, separators=(',', ':')), permanent=True)
+    except DCSError:
```
CyberDem0n: `touch_member` if not raising exceptions... 
:returns: `!True` on success otherwise `!False`

patroni/ctl.py
```diff
@@ -646,3 +646,59 @@ def configure(config_file, dcs, namespace):
...
+    return True
+
+
+def set_defaults(config, cluster_name):
+    ''' fill-in some basic configuration parameters if config file is not set '''
+    config['postgresql'].setdefault('name', cluster_name)
+    config['postgresql'].setdefault('scope', cluster_name)
+    config['postgresql'].setdefault('listen', '127.0.0.1')
+    config['postgresql']['authentication'] = {'replication': None}
+    config['restapi']['listen'] = (config['restapi']['listen']
```
CyberDem0n: config['restapi']['listen'] = '127.0.0.1:0'  # shoud solve all problems from line#657

patroni/dcs/__init__.py
```diff
@@ -313,9 +313,10 @@ def update_leader(self):
         for example for etcd `prevValue` parameter must be used."""
 
     @abc.abstractmethod
-    def attempt_to_acquire_leader(self):
+    def attempt_to_acquire_leader(self, permanent=False):
         """Attempt to acquire leader lock
         This method should create `/leader` key with value=`~self._name`
+        :param permanent: if set to `!True`, the leader key will never expie. Used in patronictl for the external master
```
feikesteenbergen: expie <> expire

patroni/ctl.py
alexeyklyukin: Yeah, that's definitely wrong. I'll extract the code we need from RestAPI into the separate function

patroni/dcs/__init__.py
```diff
@@ -343,13 +345,14 @@ def set_config_value(self, value, index=None):
...
     @abc.abstractmethod
-    def touch_member(self, data, ttl=None):
+    def touch_member(self, data, ttl=None, permanent=False):
         """Update member key in DCS.
         This method should create or update key with the name = '/members/' + `~self._name`
         and value = data in a given DCS.
 
         :param data: json serialized information about instance (including connection strings)
         :param ttl: ttl for member key, optional parameter. If it is None `~self.member_ttl will be used`
+        :permanent: if set to `!True`, the member key will never expire. Used in patronictl for the external master.
```
CyberDem0n: :param permanent:

patroni/dcs/zookeeper.py
```diff
@@ -162,8 +162,11 @@ def _create(self, path, value, **kwargs):
         except:
             return False
 
-    def attempt_to_acquire_leader(self):
-        ret = self._create(self.leader_path, self._name, makepath=True, ephemeral=True)
+    def attempt_to_acquire_leader(self, permanent=False):
+        ret = self._create(self.leader_path,
```
CyberDem0n: It could fit on one line

tests/test_ctl.py
```diff
@@ -292,3 +297,27 @@ def test_members(self, mock_get_dcs):
     def test_configure(self):
         result = self.runner.invoke(configure, ['--dcs', 'abc', '-c', 'dummy', '-n', 'bla'])
         assert result.exit_code == 0
+
+    @patch('patroni.ctl.get_dcs')
+    @patch.object(BaseHTTPServer.HTTPServer, '__init__', Mock())
```
CyberDem0n: Obsolete

tests/test_ctl.py
```diff
@@ -4,14 +4,17 @@
 import sys
 import unittest
 
+from six.moves import BaseHTTPServer
```
CyberDem0n: Obsolete

CyberDem0n:
:+1:

alexeyklyukin:
👍

alexeyklyukin:
👍

CyberDem0n:
:+1:

PR #254, opened 2016-08-03
Update etcd topology only after original request succeed

CyberDem0n:
There is no point to try to update topology until original request is
not performed. Also for us it is more important to execute original
request rather then keep topology of etcd cluster in sync.

In addition to that implement the same retry-timeout logic in the
`machines` property which already is used in `api_execute` method.

### DIFF
--- a/tests/test_etcd.py
+++ b/tests/test_etcd.py
@@ -132,6 +132,10 @@ def socket_getaddrinfo(*args):
 def http_request(method, url, **kwargs):
     if url == 'http://localhost:2379/timeout':
         raise ReadTimeoutError(None, None, None)
+    if url == 'http://localhost:2379/v2/machines':
+        ret = MockResponse()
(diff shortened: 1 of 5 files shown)


tests/test_etcd.py
```diff
@@ -132,6 +132,10 @@ def socket_getaddrinfo(*args):
 def http_request(method, url, **kwargs):
     if url == 'http://localhost:2379/timeout':
         raise ReadTimeoutError(None, None, None)
+    if url == 'http://localhost:2379/v2/machines':
+        ret = MockResponse()
+        ret.content = 'http://http://localhost:2379,http://localhost:4001'
```
alexeyklyukin: too many http://

alexeyklyukin:
👍

CyberDem0n:
:+1:

alexeyklyukin:
👍

Issue #85, opened 2015-10-28
The great connection cleanup


jberkus:
I'd like to rationalize how we track, store, and use postgres database connections.  Currently the methods of passing connection information is quite haphazard, resulting in strange and unpredictable issues with hba and passwords.  It also makes it really hard to follow the code, or to make simple fixes like issue #61.

I need to get agreement from all contributors on the way to address this, because otherwise we're headed for merge conflict hell, since touching connections and connection-passing will touch quite a bit of code.

Here is my thinking on how we should do this:
- all connection information should be stored and passed as postgresql URIs.
- we should only be using two types of connections to databases, which should have sensible defaults.

The two types of connections are:
1. connections from the patroni daemon to the local PostgreSQL, which should default to a local socket connection as the supplied superuser.
2. connections from one patroni node to another as the replication user, which connection information is supplied by the key in the DCS.

All other connections should go through the API; there is no reason why a patroni node should need to connect directly to another node for any purpose other than replication (and replication support).

This would also have some other, follow-on changes:
- the URIs we are using will need to be fixed to be valid PostgreSQL URIs.  particularly, the "database" key needs to go (proper key is "dbname").
- the replication user will need perms for setting up replication slots, etc.

So ... feedback/changes/etc.?

alexeyklyukin:
I think we should accept both name=value and URI format when getting connection strings from the user.

As for the internal format, it makes sense to store all connection informations as dicts. This simplifies checking whether two conninfos represent the same connection, as well as modifying them.

As for the primary_conninfo in recovery.conf, we don't have a choice: the URI format was only introduced in 9.5, which is yet to be released. It makes sense to  use the name=value connection strings when we output the value somewhere (i.e. to the recovery.conf, pg_rewind connection string, etc). 

As for the usage of two types of connections, it  won't work: pg_rewind requires a superuser connection to the master host, so you need it as well (there is a discussion to patch pg_rewind to use only replication connection, but this is not going to happen in 9.5).

The idea to go through the API for everything except the local, the replication and pg_rewind connection looks reasonable to me (aren't we doing it already?)

Note that we might be slow on providing feedback this week: the team is on PGConf.EU conference until Monday...

jberkus:
Ah, ok, so there's three types of connections then.  I haven't really looked at the pg_rewind code; I hadn't planned on using pg_rewind short of 9.5.

[...]

PR #255, opened 2016-08-04
Refactor code dealing with database connection string/params

CyberDem0n:
In the original code we were parsing/deparsing url-style connection
strings back and forth. That was not really resource greedy but rather
annoying. Also it was not really obvious how to switch all local
connections to unix-sockets (preferably).

This commit isolates different use-cases of working with connection
strings and minimizes amount of code parsing and deparsing them. Also it
introduces one new helper method in the `Member` object - `conn_kwargs`.
This method can accept as a parameter dict object with credentials
(username and password). As a result it returns dict object which could
be used by `psycopg2.connect` or for building connection urls for
pg_rewind, pg_basebackup or some other replica creation methods.

Params for local connection are builded in the `_local_connect_kwargs`
method and could be changed to unix-socket later easily.

This pull request partially implements https://github.com/zalando/patroni/issues/85
I don't really want to change the way how we keep connection string in DCS, so I tried to minimize parsing/deparsing of them.
The only step we need to do afterwards - implement possibility to connect to the postgres via unix-socket.

### DIFF
--- a/patroni/ctl.py
+++ b/patroni/ctl.py
@@ -21,7 +21,6 @@ from click import ClickException
 from patroni.config import Config
 from patroni.dcs import get_dcs as _get_dcs
 from patroni.exceptions import PatroniException
-from patroni.postgresql import get_conn_kwargs
 from prettytable import PrettyTable
 from six.moves.urllib_parse import urlparse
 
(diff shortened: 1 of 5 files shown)


alexeyklyukin:
👍

CyberDem0n:
:+1:

alexeyklyukin:
👍

PR #256, opened 2016-08-10
Suppress error output on master check

ikitiki:


patroni/ha.py
```diff
@@ -172,7 +172,7 @@ def fetch_node_status(member):
             xlog_location = None if is_master else json['xlog']['replayed_location']
             return (member, True, not is_master, xlog_location, json.get('tags', {}))
         except:
-            logger.exception('request failed: GET %s', member.api_url)
+            logger.error("request failed: GET %s", member.api_url)
```
CyberDem0n: exception was also producing ERROR message, but with the stacktrace.
Annoying part was not stacktrace itself, but ERROR. It should be replaced with warning and probably it also make sense to show exact reason why it has failed:
```python
except Exception as e:
    logger.error("GET %s request failed: %s", member.api_url, e)
```

CyberDem0n:
:+1:

ikitiki:
👍

ikitiki:
:+1:

ikitiki:
:+1:

Issue #247, opened 2016-07-27
Single binary release


alkersan:
Are there any plans to distribute `Patroni` as single binary? For example [docker-compose](https://github.com/docker/compose/blob/master/docker-compose.spec) or [dcos-cli](https://github.com/dcos/dcos-cli/blob/master/cli/binary/binary.spec) provide convenient binary packaging via PyInstaller.
If there is any interest, I could work on pull request.

JamesClonk:
I'd very much interested in having just one single release package, that would be great! :+1:

alkersan:
@JamesClonk, here is my initial [attempt](https://github.com/zalando/patroni/compare/master...alkersan:master). After some trial and error it passes analysis and packaging phases. Didn't check for correct functioning of all DCS yet. It would be great if someone could point to any kind of acceptance test.

CyberDem0n:
@alkersan, please create a pull request.
I've already looked into your changes, and found out that you had problems with dynamic imports of DCS implementations, because modules were searched with `os.listdir.

https://github.com/zalando/patroni/pull/253 should solve this problem.

alexeyklyukin:
I think this is already done at https://github.com/zalando/patroni/pull/290

PR #258, opened 2016-08-15
Binary build with PyInstaller

alkersan:
Proceeding with #247, please review this PR. I've checked the `pkgutil` change (#253), but unfortunately still can't find a way to scan for modules in the bundle.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,6 +6,7 @@ data/*
 .coverage
 .eggs/
 build/
+dist/
 coverage.xml
 junit.xml
 pgpass
--- a/mkbinary.sh
+++ b/mkbinary.sh
@@ -1,0 +1,5 @@
+#!/bin/sh
+set -e
+
+pip install --ignore-installed -r requirements-bin.txt
+pyinstaller --clean --onefile patroni.spec
--- a/patroni.spec
+++ b/patroni.spec
@@ -1,0 +1,29 @@
+# -*- mode: python -*-
+
+block_cipher = None
+
+
+a = Analysis(['patroni/__main__.py', 'patroni/dcs/consul.py', 'patroni/dcs/etcd.py', 'patroni/dcs/exhibitor.py', 'patroni/dcs/zookeeper.py'],
+             pathex=[],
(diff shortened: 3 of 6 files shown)


CyberDem0n:
Yeah, magic with `pkgutil` didn't really worked (unfortunately). But after reading `PyInstaller` documentation and looking on your pull request I've found the way to get list of DCS modules without hard-coding them.
The idea is very simple, `FrozenImporter` already has list of all available resources in the `toc` (table of content). We just need to filter them out by applying simple rule: take all names starting with 'patroni.dcs.'.  This is what I did in the: https://github.com/alkersan/patroni/pull/1/files

Also I've used a call to `patroni.dcs.dcs_modules` from ` patroni.spec` and got rid from hard-coded lists there as well.
Please take a look on it and do some testing, hopefully it should work.

ants:
Getting rid of hard coded lists does not seem too useful if there isn't a mechanism for external implementations. A new DCS implementation could simply add itself to the list of implementations. I find that [setuptools entry points](http://setuptools.readthedocs.io/en/latest/pkg_resources.html#entry-points) work well for this. For an example see [PluginLoader](https://github.com/zzzeek/sqlalchemy/blob/master/lib/sqlalchemy/util/langhelpers.py#L183) and [dialect _auto_fn](https://github.com/zzzeek/sqlalchemy/blob/master/lib/sqlalchemy/dialects/__init__.py) from SQLAlchemy, and a [external dialect implementation declaration](https://github.com/sqlalchemy-redshift/sqlalchemy-redshift/blob/master/setup.py).

CyberDem0n:
> Getting rid of hard coded lists does not seem too useful if there isn't a mechanism for external implementations. A new DCS implementation could simply add itself to the list of implementations. 

In your case it is not a single list but two lists (or even three). I don't think that in nearest future we will have a new DCS implementation, but I still think it is better not to use any hard coded lists if it is possible.

> I find that setuptools entry points work well for this.

Agree, and maybe one day we will make use of it, but this is absolutely another story and separate task.

PR #262, opened 2016-08-17
Always call on_start callback when starting Patroni

CyberDem0n:
When Patroni was "joining" already running postgres it was not calling
callbacks, what in some cases causing issues (callback could be used to
change routing/load-balancer or assign/remove floating (service) ip.

In addition to that we should `start` postgres instead of `restart`-ing
it when doing recovery, because in this case 'on_start' callback should
be called, instead of 'on_restart'


alexeyklyukin:
👍

CyberDem0n:
:+1:

Issue #264, opened 2016-08-18
wal_archive directory out of disk space


ovidiucp:
Is there a way to automatically remove the archived WAL files in the wal_archive directory?

I've run out of disk space on the master, so I temporarily disabled archive_mode.

What is the best way to deal with this problem?

CyberDem0n:
How does your `archive_command` looks like?
It's very strange to keep archived WAL files on the same host where the master runs. If the machine dies you will loose them... Cluster (all database files) should be backed up to some external place and WAL files should be copied (archived) to some external place either.

Actually, the primary task of Patroni - implement automatic failover and some other tasks around it, i.e.: manual failover, cascading replication, provide health-check for haproxy or any other load balancer, provide possibility to reconfigure all nodes of the postgres cluster and so on. Backup&recovery is not something what Patroni could solve for you.

It's not an easy task to implement backup&recovery 100% right. Instead of inventing your own bicycle just use something already exists and proven to work, for example [wal-e](https://github.com/wal-e/wal-e), [pgbarman](http://www.pgbarman.org/) or [pgBackRest ](https://github.com/pgbackrest/pgbackrest). There you can configure your backup&retention policy, i.e. how long your backups and WAL files should be stored. It also will take care about removing old (unneeded) files. But in any case, all these backup&WAL files should not be stored neither on master nor on any replicas.

Actually it is not

ovidiucp:
I agree, using something external makes more sense.

I was using the configuration files provided by patroni (see [postgres0.yml](https://github.com/zalando/patroni/blob/master/postgres0.yml)) with few modifications for my Kubernetes setup. I now changed `archive_mode` to `off`, and changed `archive_command` to `/bin/true`.

Perhaps the default configuration files should comment out those lines to avoid future issues.

CyberDem0n:
Changed sample config files with https://github.com/zalando/patroni/pull/265

PR #265, opened 2016-08-22
Bring all configs to the new format

CyberDem0n:
The v1.0 has been released more than one month ago and the new version
is coming. It doesn't make a lot of sense to keep configuration files in
the old format anymore.
In addition to that I've also commented out all the lines enabling and
configuring "archiving" to avoid incidents like here:
https://github.com/zalando/patroni/issues/264

### DIFF
--- a/postgres2.yml
+++ b/postgres2.yml
@@ -1,106 +1,78 @@
-ttl: &ttl 30
-loop_wait: &loop_wait 10
-scope: &scope batman
+scope: batman
+#namespace: /service/
+name: postgresql2
+
 restapi:
   listen: 127.0.0.1:8010
   connect_address: 127.0.0.1:8010
-  auth: 'username:password'
 #  certfile: /etc/ssl/certs/ssl-cert-snakeoil.pem
 #  keyfile: /etc/ssl/private/ssl-cert-snakeoil.key
+  authentication:
+    username: username
+    password: password
+
 etcd:
-  scope: *scope
-  ttl: *ttl
   host: 127.0.0.1:4001
-  #discovery_srv: my-etcd.domain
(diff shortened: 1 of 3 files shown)


postgres2.yml
```diff
@@ -1,106 +1,78 @@
...
+#namespace: /service/
+name: postgresql2
+
 restapi:
   listen: 127.0.0.1:8010
   connect_address: 127.0.0.1:8010
-  auth: 'username:password'
 #  certfile: /etc/ssl/certs/ssl-cert-snakeoil.pem
 #  keyfile: /etc/ssl/private/ssl-cert-snakeoil.key
+  authentication:
```
alexeyklyukin: Why is it commented on postgres0 and postgres1, but not postgres2?

CyberDem0n: 1. It was the same in the old config files.
2. It doesn't hurt.
3. I slightly increases acceptance-tests coverage

alexeyklyukin: Well, it only works because we are not issuing API requests to postgres2 from the acceptance tests. Anyone extending those tests to check the status of those instance (i.e. adding tag-related tests) will be unpleasantly surprised.

CyberDem0n: I think acceptance tests should not even rely on these "sample" config files but generate them from scratch. With [environment](https://github.com/zalando/patroni/blob/master/docs/ENVIRONMENT.rst) configuration these configs would be really small and in most of the cases even not necessary... Basically config file is needed only when you are bootstrapping the new cluster or you want to configure some "tags"

alexeyklyukin:
👍

CyberDem0n:
:+1:

PR #266, opened 2016-08-23
Do not drop active replication slots.

alexeyklyukin:
Master tried to delete all slots that did not correspond to the
replica registered in Patroni. That produced an error for the slots
that were active, potentially preventing drop and creation of other
slots.

Reported by Murat Kabilov.


CyberDem0n:
:+1:

alexeyklyukin:
👍

Issue #261, opened 2016-08-16
Report the replica states from the monitoring endpoints.


alexeyklyukin:
Right now, the only way to find out if the Patroni cluster has all the replicas streaming from the master is to compare the transaction location between the master and the replicas, which implies the are actually transactions happening. For many staging or just deployed environments that won't hold true.

What we can do is to return the state of pg_stat_replication, namely

```sql
postgres=# select application_name, client_addr, state, sync_state from pg_stat_replication ;
 application_name |  client_addr   |   state   | sync_state
------------------+----------------+-----------+------------
 i_4aeba3f6       | 172.31.156.141 | streaming | async
 i_bc56d001       | 172.31.132.36  | streaming | async
(2 rows)
```

Either the application name or the client address can be used to match the row against the known list of replicas, obtained from elsewhere (i.e. etcd or autoscaling group).

CyberDem0n:
Implemented in https://github.com/zalando/patroni/pull/269

PR #269, opened 2016-08-24
Return replication information on the api

feikesteenbergen:
To enable better monitoring, it is useful to have replication statistics.
Addresses issue #261

### DIFF
--- a/patroni/api.py
+++ b/patroni/api.py
@@ -368,7 +368,11 @@ class RestApiHandler(BaseHTTPRequestHandler):
 
     def get_postgresql_status(self, retry=False):
         try:
-            row = self.query("""SELECT to_char(pg_postmaster_start_time(), 'YYYY-MM-DD HH24:MI:SS.MS TZ'),
+            row = self.query("""WITH replication_info AS (
+                                    SELECT application_name, client_addr, state, sync_state, sync_priority
+                                      FROM pg_stat_replication
+                                )
+                                SELECT to_char(pg_postmaster_start_time(), 'YYYY-MM-DD HH24:MI:SS.MS TZ'),
                                        pg_is_in_recovery(),
                                        CASE WHEN pg_is_in_recovery()
                                             THEN 0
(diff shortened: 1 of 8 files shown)


CyberDem0n:
Do we really need to expose everything (every single connection) from `pg_stat_replication`? What if there is "something" streaming from the node but this something is not a part of the cluster?

patroni/api.py
```diff
@@ -377,12 +381,14 @@ def get_postgresql_status(self, retry=False):
...
                                        to_char(pg_last_xact_replay_timestamp(), 'YYYY-MM-DD HH24:MI:SS.MS TZ'),
-                                       pg_is_in_recovery() AND pg_is_xlog_replay_paused()""", retry=retry)[0]
+                                       pg_is_in_recovery() AND pg_is_xlog_replay_paused(),
+                                       (SELECT json_agg(row_to_json(ri)) FROM replication_info ri)""", retry=retry)[0]
             return {
                 'state': self.server.patroni.postgresql.state,
                 'postmaster_start_time': row[0],
                 'role': 'replica' if row[1] else 'master',
                 'server_version': self.server.patroni.postgresql.server_version,
+                'replication': row[7],
```
alexeyklyukin: I think we should only show this row if it is not empty.

patroni/api.py
```diff
@@ -377,20 +381,26 @@ def get_postgresql_status(self, retry=False):
...
                     'replayed_location': row[4],
                     'replayed_timestamp': row[5],
-                    'paused': row[6]} if row[1] else {
-                    'location': row[2]
-                })
-            }
+                    'paused': row[6]}
+            else:
+                result['xlog'] = {'location': row[2]}
+            if not result['replication']:
```
```python
if row[7]:
    result['replication'] = row[7]
```

alexeyklyukin:
👍

CyberDem0n:
:+1:

patroni/api.py
```diff
@@ -377,8 +381,10 @@ def get_postgresql_status(self, retry=False):
                                        pg_xlog_location_diff(pg_last_xlog_receive_location(), '0/0')::bigint,
                                        pg_xlog_location_diff(pg_last_xlog_replay_location(), '0/0')::bigint,
                                        to_char(pg_last_xact_replay_timestamp(), 'YYYY-MM-DD HH24:MI:SS.MS TZ'),
-                                       pg_is_in_recovery() AND pg_is_xlog_replay_paused()""", retry=retry)[0]
-            return {
+                                       pg_is_in_recovery() AND pg_is_xlog_replay_paused(),
+                                       (SELECT json_agg(row_to_json(ri)) FROM replication_info ri)""", retry=retry)[0]
```
kmoppel: Hey guys!
A quick question - what is actually currently the min. supported Pg version for Patroni? I could  I think these JSON functions came only in 9.3. Would be nice to cite it also on the main README. Cheers

CyberDem0n: Yeah, 9.2 has only rudimentary json support and this query will fail, because `json_agg` was introduced only in 9.3.
We can replace `json_agg` it with `array_to_json(array_agg(row_to_json(ri)))` and make the query 9.2 compatible, but fairly speaking we had never tested it with version < 9.3, because of pg_rewind.

You are right, it is worth to mention in the README

Issue #271, opened 2016-08-24
Failover break


uspen:
On Master kill -9 postgresql main process, and his restarted as Master! And
log:
> WARNING:  recovery command file "recovery.conf" specified neither primary_conninfo nor restore_command

recovery.conf:
>standby_mode = 'on'
recovery_target_timeline = 'latest'

uspen:
* and his restarted as Leader with lock, but because in recovery.conf 'standby_mode = 'on'' his Read Only

CyberDem0n:
Patroni starts postgres as master only in one case, when it is initializing a new cluster.
In any other case it starts it as a read-only and later does a promote.

Why are you killing postgres on the first place?

uspen:
I tested Patroni v1.0 with kill and everthing is ok. And  update  Patroni  from git to newer version. And this case is fail. 
Log:
> авг 24 18:18:14 psql-1 patroni[2369]: 2016-08-24 18:18:14,484 INFO: Lock owner: pgsql1; I am pgsql1
авг 24 18:18:14 psql-1 patroni[2369]: 2016-08-24 18:18:14,524 WARNING: Postgresql is not running.
авг 24 18:18:14 psql-1 patroni[2369]: 2016-08-24 18:18:14,527 INFO: Lock owner: pgsql1; I am pgsql1
авг 24 18:18:14 psql-1 patroni[2369]: 2016-08-24 18:18:14,527 INFO: starting as readonly because i had the session lock
авг 24 18:18:14 psql-1 patroni[2369]: 2016-08-24 18:18:14,528 INFO: Removed /data/pgsql/postmaster.pid
авг 24 18:18:14 psql-1 patroni[2369]: waiting for server to start....LOG:  database system was interrupted; last known up at 2016-08-24 18:17:18 MSK
авг 24 18:18:14 psql-1 patroni[2369]: FATAL:  the database system is starting up
авг 24 18:18:14 psql-1 patroni[2369]: FATAL:  the database system is starting up
авг 24 18:18:15 psql-1 patroni[2369]: FATAL:  the database system is starting up
авг 24 18:18:15 psql-1 patroni[2369]: FATAL:  the database system is starting up
авг 24 18:18:15 psql-1 patroni[2369]: .FATAL:  the database system is starting up
авг 24 18:18:16 psql-1 patroni[2369]: FATAL:  the database system is starting up
авг 24 18:18:16 psql-1 patroni[2369]: FATAL:  the database system is starting up
[...]

PR #272, opened 2016-08-25
close connection on start instead of stop

CyberDem0n:
https://github.com/zalando/patroni/pull/262/commits/6f5c41308f691906738a4eb799977390c1fdf0ae commit along with calling of on_start callback on start of Patroni slightly changed the behavior of the code dealing with "dead" postgres (doing "recovery"). Now it calls `start` method, when doing recovery instead of calling `restart`.

This change caused https://github.com/zalando/patroni/issues/271, because we were closing existing connection from Patroni to postgres in the `stop` method. I think `start` method is a better place for that.

### DIFF
--- a/patroni/postgresql.py
+++ b/patroni/postgresql.py
@@ -503,6 +503,12 @@ class Postgresql(object):
             self._state = value
 
     def start(self, block_callbacks=False):
+        # make sure we close all connections established against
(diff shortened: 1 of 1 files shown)


alexeyklyukin:
👍

CyberDem0n:
:+1:

PR #274, opened 2016-08-26
Feature/disable automatic failover

CyberDem0n:


patroni/api.py
```diff
@@ -239,18 +244,27 @@ def do_POST_restart(self):
 
     @check_auth
     def do_DELETE_restart(self):
-        if self.server.patroni.ha.delete_future_restart():
-            data = "scheduled restart deleted"
-            code = 200
+        if self.server.patroni.ha.is_paused():
```
CyberDem0n: I think removal of "scheduled restart" should be allowed.

patroni/api.py
```diff
@@ -239,18 +244,27 @@ def do_POST_restart(self):
...
+                data = "no restarts are scheduled"
+                code = 404
         self._write_response(code, data)
 
     @check_auth
     def do_POST_reinitialize(self):
         patroni = self.server.patroni
         cluster = patroni.dcs.get_cluster()
+        status_code = 500
+        if self.server.patroni.ha.is_paused():
```
CyberDem0n: We should check that cluster is paused against `cluster` we obtained at line#262, because `Ha` has it's own copy of `cluster` object which might be different.

I think we should create `is_paused` method in the `Cluster` class (patroni/dcs/__init__.py) and `Ha.is_paused` should simply do 'return self.cluster and self.cluster.is_paused()'

patroni/api.py
```diff
@@ -303,14 +317,17 @@ def is_failover_possible(self, cluster, leader, candidate):
...
     def do_POST_failover(self):
         request = self._read_json_content()
+        status_code = 500
         if not request:
             return
 
         leader = request.get('leader')
         candidate = request.get('candidate') or request.get('member')
         scheduled_at = request.get('scheduled_at')
+        if scheduled_at and self.server.patroni.ha.is_paused():
```
CyberDem0n: The same as line#264

patroni/ctl.py
```diff
@@ -499,6 +499,8 @@ def restart(cluster_name, member_names, config_file, dcs, force, role, p_any, sc
 
     scheduled_at = parse_scheduled(scheduled)
     if scheduled_at:
+        if is_paused(cluster):
```
CyberDem0n: We definitely should create `is_paused` method in the `Cluster` class...

patroni/ctl.py
```diff
@@ -589,9 +591,14 @@ def failover(config_file, cluster_name, master, candidate, force, dcs, scheduled
     scheduled_at = parse_scheduled(scheduled)
 
     if scheduled_at:
+        if is_paused(cluster):
+            raise PatroniCtlException("Can't schedule failover in the paused state")
         scheduled_at = scheduled_at.isoformat()
 
-    failover_value = {'leader': master, 'candidate': candidate, 'scheduled_at': scheduled_at}
+    failover_value = {'candidate': candidate, 'scheduled_at': scheduled_at}
+    if master:
```
CyberDem0n: POST-ing {"leader": null} is absolutely the same as not posting "leader" at all. No need to change previous line.

patroni/ctl.py
```diff
@@ -554,17 +556,17 @@ def failover(config_file, cluster_name, master, candidate, force, dcs, scheduled
...
-    if master is None:
+    if master is None and (not is_paused(cluster) or cluster.leader):
         if force:
             master = cluster.leader.member.name
         else:
             master = click.prompt('Master', type=str, default=cluster.leader.member.name)
 
-    if cluster.leader.member.name != master:
-        raise PatroniCtlException('Member {0} is not the leader of cluster {1}'.format(master, cluster_name))
+        if not is_paused(cluster) and cluster.leader.member.name != master:
```
CyberDem0n: Something is wrong here.
 1. Why this block is moved to the right?
 2. It should check that: if the `master` variable is not None and the `cluster` has `leader` then the `leader.name` MUST match with the `master`. If it doesn't match exceptions must be raised.

patroni/__init__.py
```diff
@@ -128,5 +128,8 @@ def main():
         pass
     finally:
         patroni.api.shutdown()
-        patroni.postgresql.stop(checkpoint=False)
-        patroni.dcs.delete_leader()
+        if patroni.ha.is_paused():
+            logger.info('Leader key is not deleted and Postgresql is not stopped due paused state')
```
CyberDem0n: Line is not covered by unit-tests

patroni/ctl.py
```diff
@@ -556,17 +556,17 @@ def failover(config_file, cluster_name, master, candidate, force, dcs, scheduled
 
     config, dcs, cluster = ctl_load_config(cluster_name, config_file, dcs)
 
-    if cluster.leader is None and not is_paused(cluster):
+    if cluster.leader is None and not cluster.is_paused():
         raise PatroniCtlException('This cluster has no master')
 
-    if master is None and (not is_paused(cluster) or cluster.leader):
+    if master is None and (not cluster.is_paused() or cluster.leader):
```
CyberDem0n: `cluster.leader` can't be `None` here if the cluster is not in pause state.

patroni/ctl.py
```diff
@@ -556,17 +556,17 @@ def failover(config_file, cluster_name, master, candidate, force, dcs, scheduled
...
         if force:
             master = cluster.leader.member.name
         else:
             master = click.prompt('Master', type=str, default=cluster.leader.member.name)
 
-        if not is_paused(cluster) and cluster.leader.member.name != master:
-            raise PatroniCtlException('Member {0} is not the leader of cluster {1}'.format(master, cluster_name))
+    if not (master is not None and cluster.leader and cluster.leader.member.name == master):
```
CyberDem0n: Logic is broken, this statement can be rewritten as:
```python
if master is None or not cluster.leader or cluster.leader.member.name != master:
```

So exception will be raised for example if the master is None, what is totally wrong.

patroni/ha.py
```diff
@@ -257,9 +271,15 @@ def manual_failover_process_no_leader(self):
         return self._is_healthiest_node(members, check_replication_lag=False)
 
     def is_healthiest_node(self):
+        if self.is_paused() and self.cluster.failover and not self.cluster.failover.scheduled_at:
```
alexeyklyukin: if self.cluster.failover and not ( self.cluster.failover.scheduled_at and self.is_paused()) covers both line 274 and line 286-287. I think line 274 should be moved below line 284, we might not want to initiate manual failover on nofailover nodes or on the nodes running the leader.

patroni/ha.py
```diff
@@ -319,30 +339,38 @@ def should_run_scheduled_action(self, action_name, scheduled_at, cleanup_fn):
     def process_manual_failover_from_leader(self):
         failover = self.cluster.failover
 
+        if failover.scheduled_at and self.is_paused():
+            return
+
         if (failover.scheduled_at and not
             self.should_run_scheduled_action("failover", failover.scheduled_at, lambda:
```
alexeyklyukin: I'd recommend to add a check for is_paused inside the self.should_run_scheduled_action. It will be triggered then for both scheduled restart and failover (no need do add the explicit checks in the scheduled restart), and will also run the cleanup callback, which will not be run in your case.

patroni/ha.py
CyberDem0n: It can't be moved below 284, because it is a special case when we are doing manual failover in the paused state. It will not even reach the line 284, because either it will return from this method on the line 278 or 281.
But you are right about 'nofailover', this check also should be added here.

patroni/ha.py
```diff
@@ -491,7 +526,10 @@ def reinitialize(self, cluster):
 
     def process_scheduled_action(self):
         if self.reinitialize_scheduled():
-            if self.cluster.is_unlocked():
+            if self.is_paused():
+                logger.warning('Cluster is in a pause state, can not reinitialize')
```
alexeyklyukin: What is the logic behind allowing manual failovers to the specific member and manual restarts, but not reinitializes?

patroni/ha.py
alexeyklyukin: So the decision here is to ignore the master role of a node (without the lock) when determining where the node is healthy enough to become a leader in the case when manual failover is already in progress. Since we don't do any demotes in the paused state except from inside the process_manual_failover_from_leader (which requires having a lock), and we explicitly forbid the Postgres master to acquire a lock during the failover by saying it is not healthy (unless it is a target for promotion), ISTM that a failover will lead to having more than one master under this logic. Is there something I'm missing?

CyberDem0n: > failover will lead to having more than one master under this logic

In theory it could happen, but in the real life somebody should "trigger" manual failover and manually promote one of the nodes nearly at the same time in order to get into such situation.

alexeyklyukin: Ok, on the other hand if there is already a maser running in addition to the leader/master (therefore, at least 2 masters are running), we won't make things worse by promoting to the non-master role, moreover, promoting to the non-master role is what we want, since the second master is likely unfit to support existing replicas following it (it may not archive the WALs properly). Anyway, those decisions should be documented (hopefully those comments may help).

patroni/ha.py
CyberDem0n: Status of restart and manual failover could be seeing immediately, but reinitialize is considered to be kind of "scheduled" action (it is executed asynchronously, from the main loop).

Hmm, this brings the question, why not to trigger reinitialize directly from API? It will remove some duplicated logic from `Ha`.

alexeyklyukin: I think we ran reinitialize as a scheduled action in order to avoid running it at an arbitrary time, where it might interfere with out checks (i.e. recover will try to run a stopped PostgreSQL). On the other hand, it seems to be the only place and we can easily introduce a flag to avoid it doing so. The question is whether the reinitialize thread spawned from the short-lived API thread will block the latter from termination (probably not, we can check it).

Anyway, reinitialize is only scheduled for technical reasons, not because the user wanted it to be scheduled at a certain time, so I don't thing it's a strong reason for blocking it during pause (we already do bootstrap during pause, which is also a 'scheduled' action) .

CyberDem0n: > we already do bootstrap during pause, which is also a 'scheduled' action

Bootstrap is a different story, it does not destroy any data.

alexeyklyukin: But isn't destroying the data what user wants explicitly?

alexeyklyukin: I actually see a use-case for bootstrap when Patroni is disabled, namely, when the master node is upgraded and then restarted to initialize all replicas one by one.

patroni/ctl.py
```diff
@@ -800,3 +805,41 @@ def flush(cluster_name, member_names, config_file, dcs, force, role, target):
                 check_response(r, member.name, 'flush scheduled restart')
             else:
                 click.echo('No scheduled restart for member {0}'.format(member.name))
+
+
+@ctl.command('disable', help='Disable auto failover')
```
alexeyklyukin: Why is it disable and not pause?

ikitiki: the name of the command is taken from the issue description:
>disable $ttl: disables auto-failover for the whole cluster, with optional TTL
>resume: resumes auto-failover for the cluster

patroni/ctl.py
```diff
@@ -800,3 +805,41 @@ def flush(cluster_name, member_names, config_file, dcs, force, role, target):
...
+@click.argument('cluster_name')
+@option_config_file
+@option_dcs
+def disable(config_file, cluster_name, dcs):
```
alexeyklyukin: Should the following 2 functions be merged into one (i.e. toggle_pause(config_file, cluster_name, dcs, pause_state), with pause calling toggle_pause(..., True) and resume calling toggle_pause(..., False). It may simplify the tests as well, less lines to cover :-)

patroni/ctl.py
alexeyklyukin: To me it makes sense to call it pause, since we use 'pause' all over the code (basically, the only place where we call the operation ``disable`` is in ``patronictl``, every other user-visible message uses ``pause``.

Also, we seem to omit TTL from the implementation, is it intentionally?

ikitiki: no, it was not intentional.
We can implement it in the next iteration i.e. pull request

alexeyklyukin:
👍

feikesteenbergen:
👍

CyberDem0n:
:+1:

PR #276, opened 2016-08-29
Describe the decisions made for the pause mode.

alexeyklyukin:


docs/pause.rst
```diff
@@ -0,0 +1,33 @@
...
+
+When Patroni runs in a paused mode, it does not change the state of PostgreSQL, except for the following cases:
+
+- For each node, the member key in DCS is updated with the current information about the cluster. This causes Patroni to run read-only queries on a member node if the member is running.
+
+- For the leader node, Patroni maintains the leader key and promotes the node holding the leader key if it is not the master.
+
+- Manual unscheduled restart and manual failover are allowed. Manual failover is only allowed if the node to failover to is specified. In the paused mode, manual failover does not requre a running master node.
+
+- If 'paralel' masters are detected by Patroni, it will acknowledge them, but will not try to demote those masters not having the leader lock.
```
ikitiki: parallel :)

alexeyklyukin: Thanks, fixed!

docs/pause.rst
```diff
@@ -0,0 +1,33 @@
+Pause/Resume mode for the cluster
+=================================
+
+The goal
+--------
+
+Under certain circumstances Patroni needs to temporary step down from managing the cluster, while still retaining the cluster state in DCS. Possible use cases are uncommon activities on the cluster, such as major version upgrades or corruption recovery. During those activities nodes are often started and stopped for the reason unknown to Patroni, some nodes can be even temporary promoted, violating the assumption of running only one master.
```
ants: Similar functionality is called maintenance mode in Pacemaker. It would be good to mention it for people coming from Linux-HA.

docs/pause.rst
```diff
@@ -0,0 +1,33 @@
...
+
+- For each node, the member key in DCS is updated with the current information about the cluster. This causes Patroni to run read-only queries on a member node if the member is running.
+
+- For the leader node, Patroni maintains the leader key and promotes the node holding the leader key if it is not the master.
+
+- Manual unscheduled restart and manual failover are allowed. Manual failover is only allowed if the node to failover to is specified. In the paused mode, manual failover does not require a running master node.
+
+- If 'parallel' masters are detected by Patroni, it emits a warning, but does not demote the masters without the leader lock.
+
+- If there is no leader lock in the cluster, Patroni promotes a running master. If there is more than one master node, then the first master node to try to acquire the lock wins. There is an exception in this rule: if there is no leader lock because the old master has demoted itself due to the manual promotion, than only the node mentioned in the promotion may take the leader lock.
```
ants: This is confusing. If I'm understanding the feature correctly instead of  "Patroni promotes a running master" it would be  better to say that "Patroni grants the leader lock to a running master".  It would best to also spell out the behavior when there are no masters.

Also maybe add a point specifying what happens when Patroni itself is stopped in paused mode (SIGINT), is PostgreSQL kept running or not.

docs/pause.rst
alexeyklyukin: Good point, thanks!

docs/pause.rst
alexeyklyukin: Agree. I think it makes sense to say "a running master acquires the lock". As for your question, Patroni will not try to stop Postgres, neither it will try to start it if it is stopped manually. 

I'll update the document shortly.

docs/pause.rst
```diff
@@ -0,0 +1,35 @@
...
+
+
+The implementation
+------------------
+
+When Patroni runs in a paused mode, it does not change the state of PostgreSQL, except for the following cases:
+
+- For each node, the member key in DCS is updated with the current information about the cluster. This causes Patroni to run read-only queries on a member node if the member is running.
+
+- For the leader node, Patroni maintains the leader key and promotes the node holding the leader key if it is not the master.
```
CyberDem0n: Actually it is other way around.
If the node is running as the master it will try and maintain leader key.
Promote could happed only if you triggered "manual failover"

docs/pause.rst
```diff
@@ -0,0 +1,35 @@
...
+
+- Manual unscheduled restart and manual failover are allowed. Manual failover is only allowed if the node to failover to is specified. In the paused mode, manual failover does not require a running master node.
```
CyberDem0n: reinitialize is also allowed

alexeyklyukin: It's right at the next line.

docs/pause.rst
alexeyklyukin: It's both, if you have a leader lock, the enforce_master_role will kick in and promote the node. I agree that in reality it's not possible to acquire a master lock, unless you edit DCS manually.

docs/pause.rst
```diff
@@ -0,0 +1,35 @@
...
+- If there is no leader lock in the cluster, the running master acquires the lock. If there is more than one master node, then the first master to acquire the lock wins. If there are no masters altogether, Patroni does not try to promote any replicas. There is an exception in this rule: if there is no leader lock because the old master has demoted itself due to the manual promotion, then only the candidate node mentioned in the promotion request may take the leader lock. When the new leader lock is granted (i.e. after promoting a replica manually), Patroni makes sure the replicas that were streaming from the previous leader will switch to the new one.
+
+- When Postgres is stopped, Patroni does not try to start it. When Patroni is stopped, it does not to stop Postgres instance it is managing.
+
+User guide
+----------
+
+``patronictl`` supports ``pause`` and ``resume`` commands.
+
+One can also issue a ``PATCH`` request to the ``{namespace}/{cluster}/config`` key with ``{'pause': True/False}``
```
CyberDem0n: It has to be json, not python
```json
- {'pause': True/False}
+ {"pause": true/false/null}
```

docs/pause.rst
CyberDem0n: > It's both, if you have a leader lock, the enforce_master_role will kick in and promote the node.

No, it will remove leader key if postgres is not running or not master

alexeyklyukin: You are right, there is a specific code for the paused mode. Will fix the docs.

alexeyklyukin:
👍

CyberDem0n:
:+1:

PR #277, opened 2016-08-29
Replace invalid characters in member names for replication slot names

ants:
PostgreSQL replication slot names only allow names consisting of [a-z0-9_].
Invalid characters cause replication slot creation and standby startup to fail.
This change substitutes the invalid characters with underscores or unicode
codepoints. In case multiple member names map to identical replication slots
master log will contain a corresponding error message.

Motivated by wanting to use hostnames as member names. Hostnames often
contain periods and dashes.

### DIFF
--- a/patroni/postgresql.py
+++ b/patroni/postgresql.py
@@ -1,6 +1,8 @@
+from collections import defaultdict
 import logging
 import os
 import psycopg2
+import re
 import shlex
 import shutil
 import subprocess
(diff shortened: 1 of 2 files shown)


CyberDem0n:
:+1:

alexeyklyukin:
👍

alexeyklyukin:
Merged, thank you!

Issue #81, opened 2015-10-26
Alternative to backup-config files


jberkus:
I wanted to suggest an atlernative to the current backup config files functionality on that branch.

We could support a conf.d directory, which would be outside PGDATA, and thus not wiped out by restores.

I believe this is a much better route than backing up and restoring conf files:
1. we don't have to touch those files, so they could be maintained by a CMF
2. doesn't involve a bunch of copying files
3. makes it easier to extend functionality by simply dropping files in the conf.d directory

... in fact, I plan to implement conf.d support anyway, so why not just go that way?

If there's some reason why you also need pg.conf backup, then can we please make it an _optional_ feature?  For container-based deployments, it makes no sense to be "backing up" pg.conf etc.

alexeyklyukin:
We don't backup postgresql.conf in order preserve it for the future: there are only 2 reasons why are we doing this:

1. wal-e removes all .conf files when it restores the backup.
2. pg_basebackup does not copy symlinks, in order to create the replica out of the master with symlinked .conf files we have to copy the symlinks into normal files inside PGDATA, then produce the base backup and then copy them back.

The first reason is the primary reason Patroni produces the .backups, the second is just a side effect that solves the issue.

I don't have any issues with supporting  conf.d configuration (you didn't specify any details, so I guess it would be implemented by supplying command-line options to pg_ctl in order to specify the location of the configuration files), as long as it is not a default option and Patroni can still locate the recovery.conf in order to rewrite it during promition/demotion.

jberkus:
Ah, OK, I didn't understand this!

1. should then be part of the WAL-E backup module, not part of main Patroni, no?
2. Do people really use symlinked .conf files?

My main point is that backing up the .conf files is currently not optional in the code; it happens whether you actually need it or not.  Since most installations don't need it, it should be optional, and off by default.

For conf.d, we need to support that by appending to the pg.conf file.  You cannot supply a pg_ctl option to add a conf directory; that can only be done in the .conf file.  That's why I haven't implemented it yet; writing to pg.conf is a new thing for us.

alexeyklyukin:
[...]

Issue #53, opened 2015-09-25
add ability to have custom postgresql.conf. pg_hba.conf


jberkus:
Add the following lines to patroni.yml:

postgresql_config_file: /path/to/custom/file
pg_hba_file: /path/to/custom/file

If present, these configuration variables will cause the files named to be copied to the data directory after it is initialized and before PostgreSQL is started up.

A TODO for me.

jberkus:
Yes, I realize that Patroni already has a way to take config settings for postgres in the patroni.yml file.  However, the formatting and organization of these is rather painful if you have a large number of special config settings.

CyberDem0n:
Actually it's already possible to specify locations for config file and hba file with using standard options:
```yaml
postgresql:
  parameters:
    hba_file: /path/to/custom/pg_hba.conf
    config_file: /path/to/custom/postgresql.conf
```
But it wont copy these file to the data directory.

jberkus:
Hmmm.  Good point!  Lemme test if that actually does what I want it to.

LappleApple:
@jberkus Did testing => it did what you wanted it to? Wondering what happens next here cc @feikesteenbergen

zenitraM:
I can't get this way to work properly - it adds `config_file` and `hba_file` entries to postgresql.conf but they get ignored, as those config entries only work as startup parameters to Postgres. Not sure if that was due to the introduction of the dynamic configuration.

zenitraM:
We currently have our postgresql.conf and pg_hba.conf managed outside `PG_DATA` by Chef and are looking how to make it fit with Patroni.

For pg_hba.conf, it seems making `hba_file` parameter be passed on the command line (like it previously did?) would do the job.

But for postgresql.conf, it seems this would fit with the dynamic configuration by being able to make Patroni skip handling of `postgresql.base.conf` and specify a either custom path to a file to include on the real `postgresql.conf` or to a conf.d path (maybe the discussion for that one would go on #81)

If those two make sense to you we can work on implementing those.

jberkus:
Yeah, I was thinking of just supporting conf.d.  Currently you can't do that via Patroni config, because of the special quoting rules for include_directory.  However, if we simply *always* had an include_directory in the container, then the user could drop stuff in it.  Agreed about hba_file.

Since I'm big on "convention over configuration", though, maybe we should make a pattern?  Assume that there's a bind mount at /etc/patroni/pgconfig/.  Any files found in that directory get copied into PGCONFIG at init time.

[...]

PR #281, opened 2016-08-31
Add configuration parameter to specify a path to a custom postgresql.base.conf and disable its backup

zenitraM:
This adds a `custom_conf` parameter on the `postgresql` settings to specify a path for a custom configuration (outside the data path), that replaces postgresql.base.conf and is `include`d on its place on the real postgresql.conf.

Setting this parameter also disables the handling and backup of postgresql.base.conf - it is expected that the user will be responsible of ensuring the external configuration file exists if this parameter is set.

It will also disable pg_hba backup if a custom `hba_file` location is set on `parameters`.

Related to #81, #53.


patroni/postgresql.py
```diff
@@ -100,9 +100,13 @@ def __init__(self, config):
         self._postgresql_conf = os.path.join(self._data_dir, config_base_name + '.conf')
         self._postgresql_base_conf_name = config_base_name + '.base.conf'
         self._postgresql_base_conf = os.path.join(self._data_dir, self._postgresql_base_conf_name)
+        self._postgresql_custom_conf = config.get('custom_conf')
         self._recovery_conf = os.path.join(self._data_dir, 'recovery.conf')
-        self._configuration_to_save = (self._postgresql_conf, self._postgresql_base_conf,
-                                       os.path.join(self._data_dir, 'pg_hba.conf'))
+        self._configuration_to_save = [self._postgresql_conf]
+        if not self._postgresql_custom_conf:
```
CyberDem0n: `postgresql.custom_conf` and `postgresql.parameters.hba_conf` can be set, updated and removed from the configuration in a run-time either by updating of `patroni.yaml` config file manually and sending SIGHUP or by changing value of the cluster `config` in DCS.

`_configuration_to_save` if used to make a copies of the configuration files to be backed up by wal-e (it skips all *.conf files)
In case of external configuration files I think it is still make sense to back them up by copying `custom_conf` and `hba_file` into `data_dir` with the names "postgresql.base.conf.backup" and "pg_hba,conf.backup" accordingly.
 
Such functionality requires some changes in a `save_configuration_files` and `restore_configuration_files` methods and we will need to change the way we keep list of configuration files from `list` or `tuple` to `dict`.

`Postgresql.reload_config` method should update value `_configuration_to_save` if the `custom_conf` or `hba_file` were changed.

patroni/postgresql.py
```diff
@@ -622,12 +626,12 @@ def restart(self):
 
     def _write_postgresql_conf(self):
         # rename the original configuration if it is necessary
-        if not os.path.exists(self._postgresql_base_conf):
+        if not self._postgresql_custom_conf and not os.path.exists(self._postgresql_base_conf):
```
CyberDem0n: What if there is no `postgresql.conf` in the data directory? I think it is exactly your case...

zenitraM: This prevents a `postgresql.base.conf` to be created from an original `postgresql.conf` if there is a custom conf somewhere else, as it will not be used.

The following lines will overwrite the `postgresql.conf` on the data dir anyway.

patroni/postgresql.py
CyberDem0n: > Also, when being restored on another instance that also has custom_conf and hba_file, what do we want to do? overwrite the file at the custom_conf path with the conf that was carried over, or copy it as postgresql.base.conf but keep it unused?.

The second option seems reasonable to be, but taking into account following:
> I find it cleaner to just explain: if you use custom_conf or hba_file, they will be under your control, there will not be any postgresql.base.conf and the configuration files will not be carried over on base_backups, and it's your responsability to ensure the settings on them are consistent and the files exist on all nodes.

I think you are right. We need document this behavior and let user live with it. Actually this is not the only drawback. Also I would like to add: currently Patroni is  "tracking" dynamic changes of postgres parameters and tells you in the API that restart is needed to apply some of the changes. In case of external configuration file it wont tell you this information..

patroni/postgresql.py
CyberDem0n: ok

zenitraM:
I've moved `_configuration_to_save` to its own property method (so I think there's no need to handle anything special on reload now).
 I've also added some mentions to the drawbacks of using its own base, but not sure where in the docs would the behavior change about the `hba_file` parameter fit.

patroni/postgresql.py
```diff
@@ -100,9 +100,8 @@ def __init__(self, config):
         self._postgresql_conf = os.path.join(self._data_dir, config_base_name + '.conf')
         self._postgresql_base_conf_name = config_base_name + '.base.conf'
         self._postgresql_base_conf = os.path.join(self._data_dir, self._postgresql_base_conf_name)
+        self._postgresql_custom_conf = config.get('custom_conf')
```
CyberDem0n: I think you need to convert it into 'property' as well, otherwise it can't be changed dynamically...

zenitraM: :+1:, just changed it to always be queried dynamically

patroni/postgresql.py
```diff
@@ -125,6 +123,15 @@ def __init__(self, config):
             self._write_postgresql_conf()  # we are "joining" already running postgres
 
     @property
+    def _configuration_to_save(self):
+        configuration = [self._postgresql_conf]
+        if not config.get('custom_conf'):
```
CyberDem0n: if not self.config

patroni/postgresql.py
```diff
@@ -622,12 +629,12 @@ def restart(self):
 
     def _write_postgresql_conf(self):
         # rename the original configuration if it is necessary
-        if not os.path.exists(self._postgresql_base_conf):
+        if not config.get('custom_conf') and not os.path.exists(self._postgresql_base_conf):
```
CyberDem0n: if 'custom_conf' not in self.config and not os.path.exists(self._postgresql_base_conf):

patroni/postgresql.py
```diff
@@ -622,12 +629,12 @@ def restart(self):
...
             os.rename(self._postgresql_conf, self._postgresql_base_conf)
 
         with open(self._postgresql_conf, 'w') as f:
             f.write('# Do not edit this file manually!\n# It will be overwritten by Patroni!\n')
-            f.write("include '{0}'\n\n".format(self._postgresql_base_conf_name))
+            f.write("include '{0}'\n\n".format(config.get('custom_conf') or self._postgresql_base_conf_name))
```
CyberDem0n: f.write("include '{0}'\n\n".format(self.config.get('custom_conf') or self._postgresql_base_conf_name))

CyberDem0n:
:+1:

alexeyklyukin:
👍

PR #282, opened 2016-09-01
Avoid retries when syncing replication slots.

alexeyklyukin:
Do not retry postgres queries that fetch, create and drop slots at the end of
the HA cycle. The complete run_cycle routine executes with the async_executor
lock. This lock is also used with scheduling operations like reinit or restart
in different threads. Looks like CPython threading class has fairness issues
when multiple threads try to acquire the same lock and one of them executes
long-running actions while holding it: the others have little chances of
acquiring the lock in order. To get around this issue, the long action (i.e.
retrying the query) is removed.

Investigation by Ants Aasma and Alexander Kukushkin.

### DIFF
--- a/patroni/__init__.py
+++ b/patroni/__init__.py
@@ -84,6 +84,10 @@ class Patroni(object):
         nap_time = self.next_run - current_time
         if nap_time <= 0:
             self.next_run = current_time
+            # Release the GIL so we don't starve anyone waiting on async_executor lock
+            time.sleep(0.001)
+            # Warn user that Patroni is not keeping up
+            logger.warning("Loop time exceeded, rescheduling immediately.")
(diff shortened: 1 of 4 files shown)


CyberDem0n:
:+1:

ants:
To avoid similar future issues we could add the following to schedule_next_run if nap_time <= 0:

    # Release the GIL so we don't starve anyone waiting on async_executor lock
    time.sleep(0.001)
    # Warn user that Patroni is not keeping up
    logger.warning("Loop time exceeded, rescheduling immediately.")

CyberDem0n:
:+1:

alexeyklyukin:
👍

alexeyklyukin:
👍

PR #284, opened 2016-09-02
Callbacks should be loaded not only on init.

alexeyklyukin:
Load callbacks from config (which is properly reloaded) instead of
reading them from init only once.


alexeyklyukin:
👍

CyberDem0n:
:+1:

PR #287, opened 2016-09-02
Try to mitigate EtcdEventIndexCleared exception

CyberDem0n:
This error is send by etcd when Patroni is doing "watch" on leader key
which is never updated after creation and etcd cluster receives a lot of
updates, what cleans history of events.

Instead of doing watch on modifiedIndex + 1 we will do watch on X-Etcd-Index,
which is probably still available...


feikesteenbergen:
👍

alexeyklyukin:
👍

PR #288, opened 2016-09-02
Decode output from wal-e list backup

feikesteenbergen:
When running this script using Python3, the output is bytestring instead of string.
We explicitly decode it to ensure checks further down are ok.

The trigger for this patch is:

```
ERROR: unable to get some of WALE backup parameters: 'expanded_size_bytes'
```


alexeyklyukin:
👍

CyberDem0n:
Tests with python3 are failing

CyberDem0n:
:+1:

feikesteenbergen:
👍

PR #289, opened 2016-09-05
Implement support to initialize an existing Database as Master

RegistryServices:
To avoid dump & load to initialize a patroni cluster and a following production downtime, implement an initialization option to add a single DB Server as Master. To subsequently add slaves with patroni to prevent longer downtimes.


CyberDem0n:
What was that?

feikesteenbergen:
This Pull Request is suggesting that we pull something from master into an already abandoned feature branch. If you want to contribute, could you create a Pull request from your fork/feature to our master branch?

PR #292, opened 2016-09-07
Remove unicode em dashes and single quotes from README

zenitraM:
In Python 3.5, pip 8.1.2 and certain unicode-aware locales/environments, pip install from the git repo fails with this:

```
    STDOUT: Collecting git+https://github.com/zalando/patroni.git@master
      Cloning https://github.com/zalando/patroni.git (to master) to /tmp/pip-2dkxt_f5-build
        Complete output from command python setup.py egg_info:
          File "<string>", line 1, in <module>
          File "/tmp/pip-2dkxt_f5-build/setup.py", line 157, in <module>
            setup_package()
          File "/tmp/pip-2dkxt_f5-build/setup.py", line 142, in setup_package
            long_description=read('README.rst'),
          File "/tmp/pip-2dkxt_f5-build/setup.py", line 112, in read
            return open(os.path.join(__location__, fname)).read()
          File "/opt/python_virtualenvs/patroni/lib/python3.5/encodings/ascii.py", line 26, in decode
            return codecs.ascii_decode(input, self.errors)[0]
        UnicodeDecodeError: 'ascii' codec can't decode byte 0xe2 in position 465: ordinal not in range(128)
```
The reason seem to be using unicode dashes and single quotes on the README (`—, ’` instead of `-, '`).

While it seems to be possible to work around it by playing with LOCALE/LC_CTYPE environment variables during pip install, I think it's worth it to just remove the fancy symbols 😅

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -4,7 +4,7 @@ Patroni: A Template for PostgreSQL HA with ZooKeeper, etcd or Consul
 ------------------------------------------------------------
 There are many ways to run high availability with PostgreSQL; for a list, see the `PostgreSQL Documentation <https://wiki.postgresql.org/wiki/Replication,_Clustering,_and_Connection_Pooling>`__.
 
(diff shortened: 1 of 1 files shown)


CyberDem0n:
Thank you!
Hope it will fix [pypi readme](https://pypi.python.org/pypi/patroni/) as well.

CyberDem0n:
:+1:

feikesteenbergen:
👍

PR #295, opened 2016-09-08
Introduce starting state and master start timeout.

### LINKED ISSUE #283: PostgreSQL taking a long time to start up is considered running and healthy
When PostgreSQL is taking a long time to start up either due to crash recovery or WAL needed to reach consistent state not being available pg_ctl -w will timeout and return with error code 0.  Postgresql.start() will happily consider such server to be running. This is, at best, misleading.

There seems to be at least one race condition here. 
1. Master boots up and starts Patroni
2. In first iteration of run_cycle Ha.recover() is called, this asynchronously starts postgres.
3. Lock ttl expires.
4. The master is now starting up, will not participate in the race due to async action.
5. Standby will consider itself healthiest because master is not reachable (xlog key missing from JSON) acquires leader key, promotes itself.
6. pg_ctl start timeout expires on old master, it sets role to be replica in Postgresql._do_follow need_rewind == False branch.
7. run_cycle calls follow to restart master with new primary_conninfo while in the middle of recovery, because role != master it will not get rewound.

Basically 3 possible outcomes after a crash
1. Master manages to reacquire the lock in handle_long_action_in_progress in second iteration of run_cycle, will continue to be master once recovery completes. Although it is marked as running when pg_ctl timeout expires, despite not yet accepting connections. Somewhat confusing, but does not require administrator intervention.
2. Lock times out after first but before second iteration, leading to rewind not getting called.
3. Lock times out before first iteration, rewind will get called, everything is peachy.

In some cases (huge checkpoint_timeout, availability is priority over durability) it might be preferrable to refrain from acquiring leader lock before crash recovery is completed.

ants:
This is WIP/RFC patch that would fix #283 . I haven't actually tested it outside running the test suite and there a couple of changes yet to be made, but some preliminary feedback on the approach taken would be nice.

Previously pg_ctl waited for a timeout and then happily trodded on considering PostgreSQL to be running. This caused PostgreSQL to show up in listings as running when it was actually not and caused a race condition that resulted in either a failover or a crash recovery or a crash recovery interrupted by failover and a missed rewind.

This change adds a master_start_timeout parameter and introduces a new state for the main run_cycle loop: starting. When master_start_timeout is zero we will fail over as soon as there is a failover candidate. Otherwise PostgreSQL will be started, but once master_start_timeout expires we will stop and release leader lock if failover is possible. Once failover succeeds or fails (no leader and no one to take the role) we continue with normal processing. While we are waiting for the master timeout we handle manual failover requests.


patroni/postgresql.py
```diff
@@ -173,14 +180,32 @@ def pg_ctl(self, cmd, *args, **kwargs):
...
+    def pg_isready(self):
+        """Runs pg_isready to see if PostgreSQL is accepting connections.
+
+        :returns: 'ok' if PostgreSQL is up, 'reject' if starting up, 'no_resopnse' if not up."""
+
+        cmd = [self._pgcommand('pg_isready'),
+               '-h', self._local_address['host'],
+               '-p', self._local_address['port'],
+               '-d', self._database,
+               '-U', self._superuser]
```
CyberDem0n: self._superuser is a dict object:
```python
{'username': 'postgres', 'password': 'blabla'}
```

So far it was not mandatory that `superuser` object should have username and password defined.
In case if there is no username, libpq is taking `PGUSER` from env or just `USER`. That means `-U` argument can be optional... 

Also pg_hba.conf may require password even for local connections. That means we should write password (if it is available) into `pgpass` file and export pgpass filename as `PGPASSFILE` environment variable. We already have `write_pgpass` method doing all this things.

ants: Gotcha, will fix. Pg_isready does not require a password as it only checks
if it gets an authentication packet.

patroni/ha.py
```diff
@@ -579,6 +626,30 @@ def post_recover(self):
...
+            time_left = self.patroni.config['master_start_timeout'] - self.state_handler.time_in_state()
+            if time_left <= 0:
+                if self.is_failover_possible(self.cluster.members):
+                    logger.info("Demoting self because master startup is taking too long")
+                    self.demote_asap()
+                    return 'stopped PostgreSQL because of startup timeout'
+                else:
+                    return 'master start has timed out, but continuing to wait because failover is not possible'
+            else:
+                msg = self.process_manual_failover_from_leader()
```
CyberDem0n: Why do we have it here?
`process_manual_failover_from_leader` should be called only if we have `failover` key in the cluster.
For the freshly created cluster such key doesn't exist.

patroni/ha.py
```diff
@@ -579,6 +626,30 @@ def post_recover(self):
...
+                if msg is not None:
+                    return msg
+
+                return 'PostgreSQL is still starting up, {0:.0f} seconds until timeout'.format(time_left)
+        else:
+            # Use normal processing for standbys
+            logger.info("Still starting up as a standby.")
+            return None
```
CyberDem0n: Shouldn't we simply return from `_run_cycle` in this case either?

patroni/postgresql.py
```diff
@@ -621,12 +651,68 @@ def reload(self):
...
+
     def restart(self):
         self.set_state('restarting')
-        ret = self.stop(block_callbacks=True) and self.start(block_callbacks=True)
-        if ret:
-            self.call_nowait(ACTION_ON_RESTART)
-        else:
+        ret = self.stop(block_callbacks=True) \
+            and self.start(block_callbacks=True) \
+            and self.wait_for_startup(ACTION_ON_RESTART)
```
CyberDem0n: `restart` method currently can be called from two places:
 1. by `POST /restart` REST API call
 2. from `_do_follow`

For the first one it make sense to `wait_for_startup`, in order to give a correct response, but I'm not sure that we should do it for the second case.

patroni/ha.py
ants: I moved the check for failover key into `process_manual_failover_from_leader` to avoid code duplication. The function can now be called whenever we are ready to handle a manual failover and will return a message if action was taken. I'll add documentation to the function.

patroni/ha.py
ants: Deferring to the main cycle handles:

* Retry starting PostgreSQL when it goes down.
* Changing primary_conninfo when leader changes
* Scheduled restarts (not sure if that is desired or not)

The main motivation was actually handling reinit, since I wrote this before 6dc1d9c88eae163ece5292fa4b8696f99bf9e754, but that doesn't seem to be relevant any more.

patroni/ha.py
```diff
@@ -122,6 +122,14 @@ def bootstrap(self):
 
     def recover(self):
         self.recovering = True
+        if self.has_lock():
+            if self.patroni.config['master_start_timeout'] == 0:
```
CyberDem0n: Should we try to update leader lock in other case (before making attempt to start in readonly)?

patroni/ha.py
CyberDem0n: Actually we probably should try to update leader lock here...

patroni/postgresql.py
ants: I re-reviewed the callsites, and I concur that REST API will want to wait, but none of the callers of `Ha.follow()` or `Postgresql.follow()` care except maybe adding context to the displayed state. But for the context it's probably more useful to maintain a log of state transitions in the DCS.

patroni/ha.py
ants: We certainly should. Somehow forgot to add it. The lack of this update is what got me started on this patch in the first place. :)

patroni/postgresql.py
```diff
@@ -173,14 +180,32 @@ def pg_ctl(self, cmd, *args, **kwargs):
         pg_ctl = [self._pgcommand('pg_ctl'), cmd]
         if cmd in ('start', 'stop', 'restart'):
             pg_ctl += ['-w']
-            timeout = self.config.get('pg_ctl_timeout')
+            timeout = kwargs.pop('timeout', None)
+            timeout = self.config.get('pg_ctl_timeout', timeout)
```
CyberDem0n: `pg_ctl_timeout` works for 'start', 'stop' and 'restart' cases.
If this value is not defined in the Patroni configuration file, then the default value from `pg_ctl` is used (60 seconds).

Can we somehow try to simplify everything? I mean we can use `pg_ctl start -w -t %master_start_timeout%` and later do a quick check that it is really accepting connections with `pg_isready`.
But from another side, there might be the case when one of the backends crashed and postmaster terminates all remaining  backends and does crash recovery. Such recovery sometime can last long and we should be able to release leader lock in this case as well.

And one more thing: in some cases even `pg_ctl stop` can "fail" to stop postgres withing a given timeout...

patroni/ha.py
ants: Good catch. Will push fix soon.

patroni/postgresql.py
ants: I didn't want to rely on pg_ctl timeout because I still want to allow for manual failover. For example, master_start_timeout is configured to a large number, but sysadmin decides that in this case it's more important to fail over instead. Maybe that usecase could be taken care of by just stopping Patroni. Not sure if it would simplify anything because the current code also handles PostgreSQL restart_after_crash=true behavior (else branch in Postgresql.check_for_startup).

Merging check_for_startup, post_recover and is_healthy could result in some simplification though.

As for stop timing out, this is why demote_asap does stop('immediate', checkpoint=False). That can still fail to stop within a timeout, but there is little we can do from within Patroni except hope that the load balancer health check will mark us unhealthy and take care of redirecting connections. I do have a patch coming up that uses kernel watchdog support to ensure that the node goes away in case things take too long. I made a note to check how other call sites behave in case of stop time outs.

patroni/ha.py
```diff
@@ -304,20 +317,52 @@ def is_healthiest_node(self):
...
             self.state_handler.follow(node_to_follow, cluster.leader, recovery=True, need_rewind=True)
         else:
             self.state_handler.follow(None, None)
 
+    def demote_asap(self):
+        """We are master, but we want to relinquish control ASAP.
+
+        We don't bother to shut down properly because we will need to rewind anyway."""
+        self.state_handler.stop('immediate', checkpoint=False)
+        # Don't set role to unknown to ensure we get rewound
```
CyberDem0n: Role should be set to 'demoted', otherwise it is still reported as 'master' via API, what stops other nodes from promote, because for them the old master is a kind of still alive.

Setting role to 'demote' will not break `pg_rewind`.

patroni/ha.py
```diff
@@ -122,6 +123,15 @@ def bootstrap(self):
 
     def recover(self):
         self.recovering = True
+        if self.has_lock() and self.update_lock():
```
CyberDem0n: `self.recovering`  should be set to True just before calling `self.follow`.
Otherwise we have following situation:
```
2016-09-13 12:20:53,994 INFO: Master crashed. Failing over.
2016-09-13 12:20:54,004 INFO: Leader key released
2016-09-13 12:20:54,004 INFO: stopped PostgreSQL to fail over after a crash
2016-09-13 12:20:54,239 INFO: Lock owner: None; I am dbnode3
2016-09-13 12:20:54,239 INFO: not healthy enough for leader race
2016-09-13 12:21:04,014 INFO: New leader detected, proceeding with recovery.
localhost:5432 - no response
2016-09-13 12:21:04,257 INFO: Lock owner: dbnode2; I am dbnode3
2016-09-13 12:21:04,258 INFO: failed to start postgres  # But we didn't even tried!
localhost:5432 - no response
2016-09-13 12:21:14,049 WARNING: Postgresql is not running.
```

ants:
Thinking some more about desired behavior when starting up as master times out, I think wait_for_failover is unnecessary and counter productive.

When we time out starting up, immediate stop is desired to allow for quick failover. However, we don't need to wait for failover. We can just initiate start as a standby. If someone else promotes all is well, but if no one does we will have a head start recovering from the crash and will eventually promote ourselves.

Current rewind logic does have a problem with this as need_rewind gets cleared if we start up as a leaderless replica. Going to push a patch for that, but I haven't yet convinced myself that it's completely safe.

patroni/ha.py
```diff
@@ -121,6 +122,15 @@ def bootstrap(self):
             return 'waiting for leader to bootstrap'
 
     def recover(self):
+        if self.has_lock() and self.update_lock():
+            if self.patroni.config['master_start_timeout'] == 0:
+                # We are requested to prefer failing over to restarting master. But see first if there
+                # is anyone to fail over to.
+                if self.is_failover_possible(self.cluster.members):
```
CyberDem0n: So far `is_failover_possible` method was used only for manual failover and it doesn't take into account xlog location of the nodes. It could happen that all the nodes are "lagging" more than `maximum_lag_on_failover` and will consider themselves  as not healthy enough for the leader race.
I think it is a good time to fix "TODO" on the line 236

CyberDem0n:
I've played with your code a little bit and it resulted in https://github.com/zalando/patroni/pull/300/files 
There are 4 separate commits explaining in details what and why I've did it.
Please, pull the latest changes from upstream master.

Tomorrow I can try to implement some acceptance tests (behave) for your functionality.

patroni/api.py
```diff
@@ -398,6 +398,7 @@ def get_postgresql_status(self, retry=False):
                     'received_location': row[3],
                     'replayed_location': row[4],
                     'replayed_timestamp': row[5],
+                    'is_lagging': self.server.patroni.ha.is_lagging(xlog=row[3]),
```
CyberDem0n: Acceptance tests are failing on this line:
```python
  File "/usr/lib/python3.5/socketserver.py", line 625, in process_request_thread                                                        
    self.finish_request(request, client_address)                                                                                        
  File "/usr/lib/python3.5/socketserver.py", line 354, in finish_request
    self.RequestHandlerClass(request, client_address, self)                                                                             
  File "/usr/lib/python3.5/socketserver.py", line 681, in __init__                                                                      
    self.handle()
  File "/usr/lib/python3.5/http/server.py", line 422, in handle                                                                         
    self.handle_one_request()
  File "/usr/lib/python3.5/http/server.py", line 410, in handle_one_request                                                             
    method()
  File "/home/akukushkin/git/ants/patroni/patroni/api.py", line 109, in do_GET_patroni
    response = self.get_postgresql_status(True)                                                                                         
  File "/home/akukushkin/git/ants/patroni/patroni/api.py", line 402, in get_postgresql_status                                           
    'paused': row[6]} if row[1] else {
TypeError: is_lagging() got an unexpected keyword argument 'xlog' 
```

I think 'is_lagging' should not be exposed via API on the first place, because:
 1. It could be calculated on any node from `replayed_location`
 2. It is using "cached" version of the cluster view in order to calculate this boolean value, what can give a wrong result, because at this moment master can already write some other value into `optime/leader`. The result will be `False`, what is wrong...

ants:
Whoops, renamed it at the last minute and unit tests did not catch that one.

You're right about using the cached version of cluster being wrong. How
well is maximum_lag_on_failiver guaranteed to be synchronized across the
cluster? The config system is somewhat convoluted so I was afraid that it
could be overridden in node local configuration.

I think replayed_location is wrong, received seems better because that's
the amount that will get applied on promotion. Replayed could be a
secondary tiebreaker.

Is fetch_node_status refactoring still desired or shall I revert that
completely?

CyberDem0n:
> You're right about using the cached version of cluster being wrong. How well is maximum_lag_on_failiver guaranteed to be synchronized across the cluster? The config system is somewhat convoluted so I was afraid that it could be overridden in node local configuration.

`maximum_lag_on_failiver` can't be overridden in the local configuration. So it's value should be the same everywhere, except the case when it was just changed and the nodes didn't had a chance to read the new config from DCS.

> I think replayed_location is wrong, received seems better because that's the amount that will get applied on promotion. Replayed could be a secondary tiebreaker.

So far we always used replayed location, but may be you are right and we should switch to received location, but we should be consistent and change the query in `Postgresql.xlog_location` accordingly.

> Is fetch_node_status refactoring still desired or shall I revert that completely?

I like this refactoring, it makes code cleaner and easy to read, let's leave it.

ants:
Ok, I'll try to get this done on the plane and push on my layover in
Frankfurt.

CyberDem0n:
It's not that urgent and can wait till Monday.
Thank you!

patroni/ha.py
```diff
@@ -304,19 +347,45 @@ def is_healthiest_node(self):
...
-    def demote(self, delete_leader=True):
-        if delete_leader:
-            self.state_handler.stop()
+    def release_leader_key_voluntarily(self):
+        self.dcs.delete_leader()
+        self.touch_member()
+        self.dcs.reset_cluster()
+        logger.info("Leader key released")
+
+    def demote(self, mode):
```
CyberDem0n: https://github.com/zalando/patroni/pull/316/commits/e538e322bcc530b55b407c2c0d150049e21a519b commit has changed slightly `process_healthy_cluster` behavior. Now it calls `self.demote(delete_leader=False)` immediately after it failed to update the leader lock.

Please update from master and don't forget to fix `process_healthy_cluster`. I think unit tests should catch.

patroni/postgresql.py
```diff
@@ -622,12 +658,84 @@ def reload(self):
...
+
+        Should only be called when state == 'starting'
+
+        :returns: True iff state was changed from 'starting'
+        """
+        ready = self.pg_isready()
+
+        if ready == STATE_REJECT:
+            return False
+        elif ready == STATE_NO_RESPONSE:
```
ants: We can't reliably distinguish between the two cases when using pg\_ctl as it doesn't return the pid that was started. That said, there is very little processing before opening ports. Some system checks, config file loading, pid file creation and `shared_preload_libraries`.  I added some code to do the waiting for port open from within Patroni and increased the timeout. If ports can't be opened within the given timeout I guess it's not too far from the truth to say that start failed.

Next iterations of run_cycle will then either notice that PostgreSQL moved into the "starting" state, see that it successfully started, or retry starting it, which will race with the initial invocation to acquire data dir lock file. So if this is a temporary issue (swap storm, VM move/pause), it will resolve itself eventually. If it isn't there isn't much we can do anyway.

Alternative would be to start postmaster directly. Then we know the pid that we started and can check that it's running even when it hasn't gotten around to creating the pid file yet. I don't think it helps the correctly configured use case too much. But it would make detection of wrong permissions on data dir faster as we notice immediately that the process we started went away without creating a pid file.

patroni/postgresql.py
```diff
@@ -567,10 +581,51 @@ def time_in_state(self):
...
+                continue
+
+            timing_slop = 2
+            is_our_pid_file = (initiated - timing_slop) <= postmaster_start_time <= (completed + timing_slop)
+            if not self.is_pid_running(pid):
+                if is_our_pid_file:
+                    # Start failed
+                    return False
+                else:
+                    # Probably old pid file, need to wait for cleanup
```
CyberDem0n: Patroni removes postmaster.pid before starting postgres

ants: About that, why is that necessary? Doesn't seem like it would be safe. I poked around in git history and that code is relatively ancient with a very uninformative commit message.

patroni/postgresql.py
CyberDem0n: > Alternative would be to start postmaster directly. 

It is rather simple, pg_ctl does something like:
```python
    pid = os.fork()
    if pid == 0:
        os.setsid()  # Creates a session and sets the process group ID (to avoid propagating Ctrl+C to postgres)
        # Environment must be empty, to avoid passwords leak
        os.execle("/bin/sh", "/bin/sh", "-c", 'exec "postgres" -D foobar --parameter="value" < "/dev/null" 2>&1', {})
    return pid
```

patroni/postgresql.py
CyberDem0n: It was inherited from governor, lets remove it

patroni/ha.py
```diff
@@ -434,7 +512,7 @@ def process_healthy_cluster(self):
             else:
                 # Either there is no connection to DCS or someone else acquired the lock
                 logger.error('failed to update leader lock')
-                self.demote(delete_leader=False)
+                self.demote('immediate')
```
CyberDem0n: I think here should be 'offline'

patroni/utils.py
```diff
@@ -302,3 +302,13 @@ def __call__(self, func, *args, **kwargs):
                 else:
                     self.sleep_func(sleeptime)
                 self._cur_delay = min(self._cur_delay * self.backoff, self.max_delay)
+
```
CyberDem0n: flake8 complains: `patroni/utils.py:306:1: E302 expected 2 blank lines, found 1`

patroni/postgresql.py
```diff
@@ -981,7 +1148,7 @@ def clone(self, clone_member):
 
     def bootstrap(self, config):
         """ Initialize a new node from scratch and start it. """
-        if self._initialize(config) and self.start():
+        if self._initialize(config) and self.start() and self.wait_for_startup():
```
CyberDem0n: Here is a small conflict: https://github.com/zalando/patroni/commit/48a6af6994f2bbb8a13e58db6179a514a3937429#diff-d8c1582337c8426ef750420dc6b76949R1008

patroni/postgresql.py
```diff
@@ -655,12 +750,84 @@ def reload(self):
...
-        self.set_state('restarting')
-        ret = self.stop(block_callbacks=True) and self.start(block_callbacks=True)
-        if ret:
-            self.call_nowait(ACTION_ON_RESTART)
+    def check_for_startup(self):
+        """Checks PostgreSQL status and returns if PostgreSQL is in the middle of startup."""
+        if self.is_starting():
+            return not self.check_startup_state_changed()
         else:
+            ready = self.pg_isready()
```
ants: Yeah, that would in principle work. Both `Ha.follow()` and `Ha.enforce_master_role()` call `Postgresql.is_leader()` and from what I can tell all relevant code paths end up there or `Postgresql.stop()` (via `demote()` or `restart()`). `is_leader()` should then catch connection errors and then check pg_isready that the server is indeed starting up, psycopg2 doesn't allow us to distinguish between server down and server starting up without parsing possibly localized error messages.

That said, I don't see a huge issue with checking for pg_isready on each iteration, but I do find loop-to-loop state propagation confusing and hard to understand as it is. Having something  called is_leader() change state to signal some other code about state change starts to get a bit too "spooky action from a distance" for my taste.

I'd actually like to refactor the code a bit to make state modification much more explicit and most of the code purely functional. The main idea is to have the main function take as input cluster state from DCS and current PostgreSQL state and have that output the desired action to take. But this idea is in pretty early phases at this point.

ants: Ah ok, that makes sense. One point of view would be that if the system is so slow that `pg_ctl status` makes ttl run out, then losing the master key is the desired outcome as the system is not really fit to be running a database. The watchdog patch I'm polishing up right now will help with the issue of PostgreSQL still running as master after the lock times out. Probably an additional layer of health checks would be nice too to detect almost-but-not-completely-dead infrastructure. It's hard to do anything useful about not having enough CPU slices to run trivial programs or not enough memory to start PG backends. Except raising our hands up, saying that we can't work like this and shutting everything down in the hope that someone else is in a better position.

Do you want to take a shot at moving pg_isready -> state=starting after a failed is_leader check or should I take it on?

CyberDem0n: > if the system is so slow that pg_ctl status makes ttl run out, then losing the master key is the desired outcome as the system is not really fit to be running a database

You are right, but from another side there is not guaranty that the new master will do better... One missing index can ruin everything :)

> we can't work like this and shutting everything down in the hope that someone else is in a better position.

Usually all database instances have more or less the same capacity and chances that situation will improve are quite small. Moreover, probably it will case not a single failover but bunch of them in a row...

> Do you want to take a shot at moving pg_isready -> state=starting after a failed is_leader check or should I take it on?

I will try to work on it.

CyberDem0n:
It looks like we will have to reimplement in python `pg_ctl start` as well...

```console
2016-10-28 11:17:36,321 WARNING: Timed out waiting for PostgreSQL to start
localhost:5432 - no response
2016-10-28 11:17:46,249 INFO: Lock owner: postgresql1; I am postgresql0
2016-10-28 11:17:46,252 INFO: failed to start postgres
2016-10-28 11:17:56,236 WARNING: Postgresql is not running.
2016-10-28 11:17:56,236 INFO: Lock owner: postgresql1; I am postgresql0
2016-10-28 11:17:56,237 INFO: Lock owner: postgresql1; I am postgresql0
2016-10-28 11:17:56,241 INFO: starting as a secondary
server starting
LOG:  database system was shut down in recovery at 2016-10-28 08:12:46 CEST
LOG:  entering standby mode
FATAL:  requested timeline 2 is not a child of this server's history
DETAIL:  Latest checkpoint is at 0/3000140 on timeline 1, but in the history of the requested timeline, the server forked off from that timeline at 0/3000140.
LOG:  startup process (PID 1915) exited with exit code 1
LOG:  aborting startup due to startup process failure
2016-10-28 11:18:06,236 INFO: Lock owner: postgresql1; I am postgresql0
2016-10-28 11:18:06,239 INFO: changing primary_conninfo and restarting in progress
2016-10-28 11:18:16,236 INFO: Lock owner: postgresql1; I am postgresql0
2016-10-28 11:18:16,240 INFO: changing primary_conninfo and restarting in progress
2016-10-28 11:18:26,236 INFO: Lock owner: postgresql1; I am postgresql0
2016-10-28 11:18:26,241 INFO: changing primary_conninfo and restarting in progress
2016-10-28 11:18:36,236 INFO: Lock owner: postgresql1; I am postgresql0
2016-10-28 11:18:36,239 INFO: changing primary_conninfo and restarting in progress
2016-10-28 11:18:46,237 INFO: Lock owner: postgresql1; I am postgresql0
2016-10-28 11:18:46,240 INFO: changing primary_conninfo and restarting in progress
2016-10-28 11:18:56,237 INFO: Lock owner: postgresql1; I am postgresql0
2016-10-28 11:18:56,239 INFO: changing primary_conninfo and restarting in progress
2016-10-28 11:18:56,316 WARNING: Timed out waiting for PostgreSQL to start
```

What has happened here:
 1. postgres started and failed quickly. Off course it removed postmaster.pid 
 2. `wait_for_port_open` was waiting 60 seconds, because it was not able to figure out that postmaster process is already dead (it doesn't know postmaster pid).

The original pg_ctl wait implementation calls `waitpid((pid_t) pm_pid, &exitstatus, WNOHANG)` every second and compares result with postmaster pid.
And here comes a tricky part...  Currently we are reaping dead processes before and after main HA loop, otherwise (if waitpid was called from sigchild_handler) subprocess was not getting the return code...

CyberDem0n:
@ants, please look on https://github.com/ants/patroni/pull/4

CyberDem0n:
@ants, please take a look on https://github.com/ants/patroni/pull/5 and especially on commit https://github.com/ants/patroni/pull/5/commits/0d7225957ea53816571f6c79fe4f2076e005f0b0

CyberDem0n:
@ants, https://github.com/ants/patroni/pull/6 -- probably the last one

CyberDem0n:
:+1:

alexeyklyukin:
👍

PR #296, opened 2016-09-08
Add post_init configuration parameter on bootstrap

### LINKED ISSUE #291: Callback for custom operations after initial bootstrap
Our usecase involves creating a lot of database clusters that need to have some initial database setup (some specific databases with extensions) created on the cluster (so, only once after the initial initdb bootstrapping) before becoming usable.

So we would like to have a way for Patroni to invoke the running of these tasks after the db is created but before the node is deemed ready. 

I was thinking on adding a new `on_bootstrap` hook that gets executed after the initial bootstrap currently done [here](https://github.com/zalando/patroni/blob/master/patroni/postgresql.py#L973) and blocks until completion. 

Would this make sense?, can you think of any other way to achieve this?

zenitraM:
The script specified on it will be run after initdb, blocking cluster initialization until it finishes.

Fixes #291

### DIFF
--- a/docs/SETTINGS.rst
+++ b/docs/SETTINGS.rst
@@ -32,7 +32,7 @@ Bootstrap configuration
         -  **options**: list of options for CREATE USER statement
             -  **- createrole**
             -  **- createdb**
--  **post_init**: An additional script that will be executed after initializing the cluster. It will be passed a connection string URL to the cluster superuser as a parameter, as well as the location to the **pgpass* file on the `PGPASSENV` variable.
+-  **post_init**: An additional script that will be executed after initializing the cluster. The script receives a connection string URL (with the cluster superuser as a user name). The PGPASSFILE variable is set to the location of pgpass file.
 
 Consul
 ------
--- a/docs/SETTINGS.rst
+++ b/docs/SETTINGS.rst
@@ -32,6 +32,7 @@ Bootstrap configuration
         -  **options**: list of options for CREATE USER statement
             -  **- createrole**
             -  **- createdb**
+-  **post_init**: An additional script that will be executed after initializing the cluster. It will be passed a connection string URL to the cluster superuser as a parameter, as well as the location to the **pgpass* file on the `PGPASSENV` variable.
 
 Consul
 ------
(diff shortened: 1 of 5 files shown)


patroni/postgresql.py
```diff
@@ -389,6 +389,35 @@ def _initialize(self, config):
...
         return ret
 
+    def run_bootstrap_post_init(self, config):
+        """
+        runs a script after initdb is called and waits until completion.
+        passed: cluster name, parameters
+        """
+        if 'post_init' in config:
+            cmd = config['post_init']
+            r = self.get_local_address()
```
CyberDem0n: I would like to replace lines 399-410 with following:
```python
r = self._local_connect_kwargs
if 'user' in r:
    connstring = 'postgres://{user}@{host}:{port}/{database}'.format(**r)
else:
    connstring = 'postgres://{host}:{port}/{database}'.format(**r)

env = self.write_pgpass(r) if 'password' in r else None
```

patroni/postgresql.py
```diff
@@ -389,6 +389,35 @@ def _initialize(self, config):
...
+            if 'username' in self._superuser:
+                r.update({'user': self._superuser['username']})
+                connstring = 'postgres://{user}@{host}:{port}/{database}'.format(**r)
+            else:
+                connstring = 'postgres://{host}:{port}/{database}'.format(**r)
+            if 'password' in self._superuser:
+                r.update({'password': self._superuser['password']})
+                env = self.write_pgpass(r)
+            else:
+                env = os.environ.copy()
```
CyberDem0n: post_init script doesn't need to know all environment of Patroni...

patroni/postgresql.py
```diff
@@ -389,6 +389,35 @@ def _initialize(self, config):
...
+            try:
+                ret = subprocess.call(shlex.split(cmd) + [connstring], env=env)
+            except OSError:
+                logger.exception('post_init script %s failed', cmd)
+                return False
+            if ret != 0:
+                logger.exception('post_init script %s returned non-zero code %d', cmd, ret)
```
CyberDem0n: I think logger.error would be enough in this case

patroni/postgresql.py
zenitraM: :+1: thanks! didn't notice local_connect_kwargs

docs/SETTINGS.rst
```diff
@@ -32,6 +32,7 @@ Bootstrap configuration
         -  **options**: list of options for CREATE USER statement
             -  **- createrole**
             -  **- createdb**
+-  **post_init**: An additional script that will be executed after initializing the cluster. It will be passed a connection string URL to the cluster superuser as a parameter, as well as the location to the **pgpass* file on the `PGPASSENV` variable.
```
zenitraM: Sounds good

CyberDem0n:
:+1:

alexeyklyukin:
👍

PR #304, opened 2016-09-15
Robust BS

maorvnera:


patroni/ha.py
```diff
@@ -613,6 +617,8 @@ def _run_cycle(self):
                 if self.sysid_valid(self.cluster.initialize) and self.cluster.initialize != self.state_handler.sysid:
                     logger.fatal("system ID mismatch, node %s belongs to a different cluster: %s != %s",
                                  self.state_handler.name, self.cluster.initialize, self.state_handler.sysid)
+                    logger.info("Removing Data dir, allowing DB to bootstrap")
```
CyberDem0n: This action is really questionable and dangerous. We should not touch data directory, it might be that it is the last good copy of your cluster!

patroni/ha.py
```diff
@@ -78,12 +78,16 @@ def touch_member(self):
...
+        try:
+            if self.state_handler.clone(clone_member):
+                logger.info('bootstrapped %s', msg)
+                cluster = self.dcs.get_cluster()
+                node_to_follow = self._get_node_to_follow(cluster)
+                self.state_handler.follow(node_to_follow, cluster.leader, True)
+            else:
+                logger.error('failed to bootstrap %s', msg)
+                self.state_handler.remove_data_directory()
+        except:
```
CyberDem0n: Can you please elaborate why do you want to catch all exceptions here?
I see not a single reason to do that, because:
 1. All expected exceptions already handled underneath.
 2. Even if there is something really bad happening it will not disturb the main HA loop, because this method is executed only by `AsyncExecutor.run`, which at the end will catch all remaining things.

patroni/ha.py
maorvnera: I hit a problem during bootstrapping from leader (over ZK)
There was a network reset, and this broke bootstrapping from leader. rendered node in unusable state.
How about moving the data dir to backup and starting fresh?

patroni/ha.py
maorvnera: I can rethrow...
I wan to remove data dir as bootstrapping failed, need to cleanup and start again

patroni/ha.py
CyberDem0n: > I hit a problem during bootstrapping from leader (over ZK)
> There was a network reset, and this broke bootstrapping from leader. rendered node in unusable state.
> How about moving the data dir to backup and starting fresh?

This functionality is there for a long time.  `Postgresql.clone` method returns `False` when something fails, what will cause removal of data directory. `clone` method is not supposed to raise any exceptions, therefore there is no need to catch them.

patroni/ha.py
CyberDem0n: Once again, `clone` method is not supposed to raise any exceptions, therefore there is no need to catch them.

LappleApple:
Hello @maorvnera, do you plan on pursuing this PR any further, or can we close? cc @CyberDem0n

LappleApple:
@CyberDem0n @alexeyklyukin I'm going to close this, OK?

PR #305, opened 2016-09-16
Implement retry and timeout strategy for consul the same way as for etcd

CyberDem0n:
Change HTTPClient implementation from using `requests.session` to
`urllib3.PoolManager`, because reference implementation from python-consul
didn't really worked with timeouts and was blocking HA loop...


alexeyklyukin:
👍

CyberDem0n:
:+1:

PR #307, opened 2016-09-19
Make sure data directory is empty before trying to restore backup

CyberDem0n:
We are doing number of attempts when trying to initialize replica using
different methods. Any of this attemp may create and put something into
data directory, what causes next attempts fail.
In addition to that improve logging when creating replica.

### DIFF
--- a/patroni/postgresql.py
+++ b/patroni/postgresql.py
@@ -450,6 +450,9 @@ class Postgresql(object):
                     # if basebackup succeeds, exit with success
                     break
             else:
+                if not self.data_directory_empty():
+                    self.remove_data_directory()
+
                 cmd = replica_method
                 method_config = {}
                 # user-defined method; check for configuration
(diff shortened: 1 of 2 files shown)


CyberDem0n:
> Perhaps we should unify the retry parameter between all methods (keeping it still per-method), so that we can retry directly from create_replica, not from inside the replica creation method.

You are mixing two different cases of retry:
 1. When custom create_replica method has failed we are retrying with some other method.
 2. We have hard-coded retry mechanism for pg_basebackup.

Before executing any of this methods it better to have data directory empty.

alexeyklyukin:
I don't think those cases are fundamentally different: either cleanup should be executed after each failure, no matter whether this is the last attempt of the chosen method or not, or it should be skipped between the attempts of the same method. I'd chose the first option, and the way you have implemented cleanup inside the basebackup suggests this is your choice as well. However, in this case we should make sure the custom methods run retries between their attempts as well.

alexeyklyukin:
👍

CyberDem0n:
:+1:

Issue #309, opened 2016-09-19
Target non-master server for restore operations


jamessewell:
Hello,

Is it possible to target an up to date standby server for pg_basebackup or pg_rewind operations?

This would become useful in a situation where something like WAL-E wasn't available to retrieve a base-backup and you were in production and didn't want to impact the master unless absolutely needed.

Cheers,
James

jamessewell:
As an addition to this, I suppose what I really mean is:

- Target an up to date standby for restore operations if one exists, else use the master

ants:
I think the (for some reason) undocumented clonefrom tag does what you want. Looking at the code, it randomly picks one of non-leader servers with clonefrom set, if none exist it goes with the leader.

jamessewell:
So it does.

I had a quick look earlier, but thought it was intended for a list of servers to clone from - this is much better!

CyberDem0n:
Yeah, it really should be documented...
So far the only mention of the tags was in the https://github.com/zalando/patroni/issues/48

jamessewell:
Hmmm, the idea is perfect - but this doesn't seem to work.

My three node cluster bootstraps fine with all nodes set to _clonefrom:  false_, but when I set them all to _clonefrom: true_ I get the following on the two non master nodes:

```
2016-09-20 17:36:36,281 ERROR: Trying again in 5 seconds
pg_basebackup: could not connect to server: could not connect to server: Connection refused
        Is the server running on host "192.168.2.247" and accepting
        TCP/IP connections on port 5432?
2016-09-20 17:36:41,295 ERROR: failed to bootstrap from replica 'postgresql-192.168.2.247'
2016-09-20 17:36:41,296 INFO: Removing data directory: /db/9.5/data/pg_data
2016-09-20 17:36:46,275 INFO: trying to bootstrap from replica 'postgresql-192.168.2.245'
pg_basebackup: could not connect to server: could not connect to server: Connection refused
        Is the server running on host "192.168.2.245" and accepting
        TCP/IP connections on port 5432?

```

**.245** is the current node**, .246** is the master and **.246** is another replica - so it looks like it's trying to clone from everyone (including itself) then not trying the master.

Thanks for the help,
James

CyberDem0n:
It is clearly a bug, node itself must be excluded (as well as any other nodes not in the "running" state)

jamessewell:
Sweet, that fix works for me 👍

PR #313, opened 2016-09-20
Exclude unhealthy nodes when choosing where to clone from

CyberDem0n:
Node MUST have tag clonefrom: true, be in the 'running' state and also we should not try to clone from itself.

Fixes https://github.com/zalando/patroni/issues/309

### DIFF
--- a/features/cascading_replication.feature
+++ b/features/cascading_replication.feature
@@ -8,6 +8,7 @@ Scenario: check a base backup and streaming replication from a replica
 	And replication works from postgres0 to postgres1 after 20 seconds
 	And I create label with "postgres0" in postgres0 data directory
 	And I create label with "postgres1" in postgres1 data directory
+	And postgres1 has state=running in dcs after 12 seconds
 	And I configure and start postgres2 with a tag replicatefrom postgres1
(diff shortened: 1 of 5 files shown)


CyberDem0n:
:+1:

features/cascading_replication.feature
```diff
@@ -8,6 +8,7 @@ Scenario: check a base backup and streaming replication from a replica
 	And replication works from postgres0 to postgres1 after 20 seconds
 	And I create label with "postgres0" in postgres0 data directory
 	And I create label with "postgres1" in postgres1 data directory
+	And postgres1 has state=running in dcs after 12 seconds
```
alexeyklyukin: replication works from postgres0 -> postgres1 implies postgres1 has a state running, since replication check actually queries the database. How comes it didn't have state=running in DCS?

CyberDem0n: > How comes it didn't have state=running in DCS?

Because touch_member is not called immediately after starting up postgres.
clone && start takes 2-3 seconds, but loop_wait is 10 seconds, so member would be updated only on the next iteration.

alexeyklyukin:
👍

PR #314, opened 2016-09-20
Add synchronous replication support.

ants:
This is still WIP. Mainly works as intended, but tests and docs are TBD. I mainly want feedback whether this approach is ok.

Adds a new configuration variable synchronous_mode. When enabled Patroni will manage synchronous_standby_names to enable synchronous replication whenever there are healthy standbys available. With synchronous mode enabled Patroni will automatically fail over only to a standby that was synchronously replicating at the time of master failure. This effectively means zero lost user visible transactions.

To enforce the synchronous failover guarantee Patroni stores current synchronous replication state in the DCS, using strict ordering, first enable synchronous replication, then publish the information. Standby can use this to verify that it was indeed a synchronous standby before master failed and is allowed to fail over.

We can't enable multiple standbys as synchronous, allowing PostreSQL to pick one because we can't know which one was actually set to be synchronous on the master when it failed. This means that on standby failure commits will be blocked on the master until next run_cycle iteration. TODO: figure out a way to poke Patroni to run sooner or allow for PostgreSQL to pick one without the possibility of lost transactions.

On graceful shutdown standbys will disable themselves by setting a nosync tag for themselves and waiting for the master to notice and pick another standby. This adds a new mechanism for Ha to publish dynamic tags to the DCS.

When the synchronous standby goes away or disconnects a new one is picked and Patroni switches master over to the new one. If no synchronous standby exists Patroni disables synchronous replication (synchronous_standby_names=''), but not synchronous_mode. In this case, only the node that was previously master is allowed to acquire the leader lock.


jamessewell:
This is something we would be interested in.

So if I got this right then there is only ever a single entry in synchronous_standby_names, and that name is managed by Patroni? That sounds reasonable.

One question - I might have mis-interpreted, but are you proposing that if all standbys leave / fail then synchronous_standby_names is set to ''?

This would be a bad thing for some users, as it negates the data protection guarantee.

ants:
My customers prefer the feature the way it's currently implemented. Losing transactions is viewed as annoyance that is better avoided, but is a risk they are willing to take to stay available. This is coincidentally the same way that Pacemaker's pgsql resource agent does sync replication. I can see that some people might have a different view of the durability-availability tradeoff. It's almost possible to achieve that today - just set synchronous_standby_names manually on each node. The only small issue is that with more than 2 nodes and 2 simultaneous failures it's not possible to be sure that it's actually safe to fail over because the second one of those failures might have been the synchronous standby at the time. I didn't add a mode switch for this because I didn't have a real world usecase to design against. If you say you would need this and are willing to help design the "extra durability" mode, I'm happy to add it.

jamessewell:
I think you need to have an option. Perhaps _strict_synchronous_replication_ or something?

We have customers who see it as you describe, but we also have gambling, finance, and health customers who certainly don't. They value limiting transaction loss above availability and use sync rep to achieve this.

The PostgreSQL synchronous replication feature also works this way, as implemented you are changing the default behavior of the feature (this could obviously be debated at length!).

Having said all that, I do like the way it works currently and I would certainly use the implemented variant in many cases!

patroni/postgresql.py
```diff
@@ -1041,6 +1044,38 @@ def basebackup(self, conn_url, env):
...
+            member = members.get(app_name)
+            if not member:
+                continue
+            if member.tags.get('nosync', False):
+                continue
+            if state != 'streaming':
+                continue
+            if sync_state == 'sync':
+                return app_name, True
+            if app_name == current and state == 'streaming' and sync_state == 'potential':
```
CyberDem0n: There is no guaranty that row with application_name = current will be first in result set...

patroni/postgresql.py
```diff
@@ -1041,6 +1044,38 @@ def basebackup(self, conn_url, env):
...
+        for app_name, state, sync_state in self.query(
+                """SELECT application_name, state, sync_state
+                    FROM pg_stat_replication
+                    ORDER BY sync_state = 'sync' DESC, flush_location DESC"""):
+            member = members.get(app_name)
+            if not member:
+                continue
+            if member.tags.get('nosync', False):
+                continue
+            if state != 'streaming':
```
CyberDem0n: Can we move it into query?
```sql
WHERE state = 'streaming'
```

patroni/__init__.py
```diff
@@ -48,13 +48,19 @@ def load_dynamic_configuration(self):
                 logger.warning('Can not get cluster from dcs')
 
     def get_tags(self):
-        return {tag: value for tag, value in self.config.get('tags', {}).items()
+        tags = {tag: value for tag, value in self.config.get('tags', {}).items()
                 if tag not in ('clonefrom', 'nofailover', 'noloadbalance') or value}
+        tags.update(self.ha.get_dynamic_tags())
```
CyberDem0n: `get_tags` method executed only on patroni start and "reload". Result returned by this method is assigned to the `Patroni.tags` and is used in different places (API and touch_member).
But none of them will see nosync tag if reload didn't happened.

patroni/__init__.py
```diff
@@ -48,13 +48,19 @@ def load_dynamic_configuration(self):
...
+        return tags
 
     @property
     def nofailover(self):
         return bool(self.tags.get('nofailover', False))
 
+    @property
+    def nosync(self):
+        return bool(self.tags.get('nosync', False))
```
CyberDem0n: Does it mean that you introduced a new tag, which will allow to disable possibility to because a synchronous replica for a given node?

patroni/ha.py
```diff
@@ -157,13 +164,103 @@ def follow(self, demote_reason, follow_reason, refresh=True, recovery=False, nee
...
+        publishing it to the DCS. As the window is rather tiny consequences are holding up commits for one cycle
+        period we don't worry about it here."""
+
+        if not self.is_synchronous_mode or self.patroni.nosync:
+            return func()
+
+        self._disable_sync = True
+        try:
+            try:
+                self.touch_member()
```
CyberDem0n: `touch_member` wont see tags.nosync=true ...

patroni/postgresql.py
```diff
@@ -80,6 +80,7 @@ def __init__(self, config):
         self._database = config.get('database', 'postgres')
         self._data_dir = config['data_dir']
         self._pending_restart = False
+        self._sync_standby = None
```
CyberDem0n: is it ever used?

ants: Nope. Leftover from refactoring a previous iteration.

patroni/postgresql.py
ants: I thought about adding some debug logging here so users gain some visibility into why nodes are or are not picked as sync standbys. That said, until that logging actually gets added this can be in the query.

patroni/postgresql.py
ants: `ORDER BY  sync_state = 'sync' DESC` does that, but that's indeed a bit clever. I guess a comment is warranted here.

CyberDem0n: `ORDER BY sync_state = 'sync' DESC` works only for the line 1061 (when there is `sync` replica).

But I'm talking about lines 1063-1066

patroni/__init__.py
ants: Hmm, not sure how to best handle this.

1. Update tags after all callsites that could affect dynamic tags (bug prone)
2. Make patroni.tags a property (lots of useless dict building to check if some flag is set or not)
3. Make `Ha.touch_member` apply dynamic tags. (creates a separation about publicly visible and local tags that might get confusing in the future)
4. Have a separate member attribute for dynamic nosync override and move the `_disable_sync` check into `touch_member`, have `pick_synchronous_standby` check that attribute. (simple and stupid, non-general, makes DCS API wider for a single redundant bit of information)

I'm currently thinking 3, but any other ideas are welcome.

patroni/__init__.py
ants: Right. Nosync means that "don't allow this node to become a sync replica" (is behind WAN for example). It's also used to communicate "I don't want to be a sync standby anymore" to the master to avoid commits hanging until master notices that a standby has gone away.

patroni/__init__.py
CyberDem0n: `patroni.tags` currently used in `Ha` and `Api`.
Pattern in both cases similar:
```python
# Ha
if self.patroni.tags:
   data['tags'] = self.patroni.tags

# Api
response.update({'tags': patroni.tags} if patroni.tags else {}) 
```

Lets just create method or property in Ha which will mix patroni.tags and dynamic tags and return something like: `{'tags': {'tag1': 'value1'}}` if we have some tags to expose and `{}` if we don't.

Later we can use it as:
```python
foo.update(patroni.ha.get_tags())
```

P.S. it make sense to expose dynamic `nosync` tag via API as well

patroni/postgresql.py
ants: Aha, right, that's a bug. I'll figure out a way to code this that makes the priorities explicit.

patroni/__init__.py
ants: Sounds like a plan. I'll try to document this in a way that makes it clear why the tag can locally be False, but advertised as being True. I'm still kind of on the fence if conflating configured values and dynamic state in this way is a good idea. But I guess it isn't too hard to change in the future if it proves to be a bad idea.

patroni/ha.py
```diff
@@ -69,7 +70,7 @@ def touch_member(self):
             'role': self.state_handler.role
         }
         if self.patroni.tags:
-            data['tags'] = self.patroni.tags
+            data['tags'] = self.get_effective_tags()
```
CyberDem0n: If statement is wrong. `patroni.tags` can be empty but 'nosync' set dynamically.

patroni/ha.py
CyberDem0n: Also it doesn't throw an exception but returns False if fails.

patroni/dcs/zookeeper.py
```diff
@@ -307,6 +311,22 @@ def delete_cluster(self):
...
+            return True
+        except NoNodeError:
+            return value == '' or (index is None and self._create(self.sync_path, value))
+        except:
+            logging.exception('set_sync_state_value')
+            return False
+
+    def delete_sync_state(self, index=None):
+        try:
+            return self._client.retry(self._client.delete, self.sync_path, version=index)
```
ants: Ok, no reason why we couldn't use the same approach here. Will fix.

patroni/ha.py
```diff
@@ -25,6 +26,7 @@ def __init__(self, patroni):
         self.old_cluster = None
         self.recovering = False
         self._async_executor = AsyncExecutor()
+        self._disable_sync = False
```
ants: I worked on the premise that setting a boolean flag is atomic. That is probably a mix of laziness and premature optimization. However to get the "right value" a mutex on disable sync will not be sufficient or in any way helpful. Instead touch_member needs to be protected by a mutex to avoid getting the value overwritten by stale one.

However I don't think it's worth the effort in this case - it would be better to add a timeout for while_not_sync_standby (2*loop_wait should be enough). Even if there is a race that makes master not notice in time, at worst there will be a small hickup on commits in the master. A timeout is much more robust solution than a locking scheme.

There is a second cluster wide race there anyway:

* Master picks standby to be sync
* Standby publishes nosync
* Standby checks sync state and thinks it can shut down.
* Master makes standby sync standby and publishes sync state
* Standby shuts down
* Master takes up to loop_wait to notice this.

It would be possible to guarantee no unnecessary sync standby waits with a considerable cost in complexity by adding a sync standby candidate field to sync state, have the master publish that and then run an API call to double check before calling synchronous_standby. Thoughts?

CyberDem0n: > Instead touch_member needs to be protected by a mutex to avoid getting the value overwritten by stale one.

That's actually what I meant.

> it would be better to add a timeout for while_not_sync_standby (2*loop_wait should be enough).

Timeout is also important. But I think it still make sense to "serialize" touch_member calls...

> * Standby publishes nosync
> * Standby checks sync state and thinks it can shut down.

There are 2 seconds between these actions, so chances are very low...

patroni/ha.py
```diff
@@ -157,13 +176,107 @@ def follow(self, demote_reason, follow_reason, refresh=True, recovery=False, nee
...
+            return func()
+
+        with self._member_state_lock:
+            self._disable_sync += 1
+        try:
+            if self.touch_member():
+                # Master should notice the updated value during the next cycle. We will wait double that, if master
+                # hasn't noticed the value by then not disabling sync replication is not likely to matter.
+                for _ in polling_loop(timeout=self.dcs.loop_wait*2, interval=2):
+                    if not self.is_sync_standby(self.dcs.get_cluster()):
```
CyberDem0n: `get_cluster` can raise an exception and `func()` want be called

patroni/dcs/__init__.py
```diff
@@ -458,9 +490,20 @@ def cancel_initialization(self):
         """ Removes the initialize key for a cluster """
 
     @abc.abstractmethod
-    def delete_cluster(self):
+    def delete_cluster(self, index=None):
```
CyberDem0n: Why did you changed interface of `delete_cluster`?

ants: Whoops. That explains the codacy bot complaining about interface mismatch... Will fix.

On that note, should I rebase this series to something cleaner when things start to stabilize?

CyberDem0n: You mean squash all commits into one?

ants: Basically yes. Looking at the diff there isn't much that would make sense as a separate patch besides protecting touch_member in a lock.

CyberDem0n: I am fine with it, but please also fix https://github.com/zalando/patroni/pull/314/files#diff-0996184a693191991bd4f56d522c4f32R50 and https://github.com/zalando/patroni/pull/314/files#diff-8ac0f76acf09bfec37a70fed766b3d27R326

Also `setup.py flake8` complains about `E226 missing whitespace around arithmetic operator`
Probably it is because: https://github.com/zalando/patroni/pull/314/files#diff-b91f3d5bd63fcd17221b267e851608e8R3

Somehow flake8 doesn't like comment lines, probably because they start not from begging of the string. And actually after removing comments result is absolutely the same as not having `ignore` at all.

CyberDem0n:
I've spend some time yesterday and today doing manual and automated testing against not only etcd, but also zookeeper and consul. Testing has revealed a few minor problems (mostly related to DCS) which I've already fixed. Also I've implemented acceptance tests which are enabling synchronous_mode, doing restarts of sync replicas, triggering automatic failover and checking that it really failed over to the sync replica.

Also I've got a question: when we have synchronous_mode enabled but don't have streaming replicas available you are removing "synchronous_standby_names" from postgresql.conf what actually disables synchronous_mode. Is it intentionally? Shouldn't we set `synchronous_standby_names to '*'` in this case?

ants:
Thanks for the acceptance tests. I really need to figure out how that system works to so I can do those myself too. I did some manual system testing on my own with a VMs behind a HAProxy load balancer and a testing tool that shows commit and failover latencies. I'll have to see if the acceptance test framework is suitable for automating this.

Removing `synchronous_standby_names` is intentional. Otherwise a standby that is not managed by Patroni might become synchronous. If this is for example a DR standby behind slow WAN it could cause an availability problem as the master now has to wait for replication to the slow site.

As currently implemented the synchronous mode intends to provide the guarantee that Patroni will not automatically fail over to a node that doesn't contain all commits that have been confirmed to clients (except clients requesting otherwise via synchronous_commit=off/local).

It intentionally does not provide a guarantee of 2 node durability of successful commits while the master is alone in the cluster for availability reasons. However this could be implemented as a separate mode in a reasonably simple manner.

The current synchronous mode also has the issue of introducing `loop_wait` delays to commits on sync standby failure even if an alternate standby is available. One idea how to improve that: given nodes A(master), B, C and `s_s_n = 'B,C'`, we could set sync state to contain both B and C as sync standbys. `is_healthiest_node` would then have to check that either the current node is sync state leader or it is one of the sync standby's and the other standbys are all responding and have xlog position behind us. This means that when both A and C fail, B can't promote because we don't know if it was actually connected as sync standby during the failure. So this approach would improve commit availability on standby failure at the expense of needing manual intervention during multi node failures.

Anything better seems to require patching PostgreSQL or adding replication middleware so we have a cluster wide knowledge of the latest confirmed commit. Middleware would also give the option to change leaders without restarting standby's, but implementing one is definitely a separate project from this feature.

The current scheme mirrors what rep_mode=sync does in Pacemakers pgsql resource agent.

CyberDem0n:
Here are my changes: https://github.com/ants/patroni/pull/1

CyberDem0n:
@ants, any comments on https://github.com/ants/patroni/pull/1 ?

ants:
Sorry, I had another task take priority for a while. Reviewed the changes and they look ok to me. I think the logic could be simplified if we disallow `cluster.sync is None` and always use a null object instead.

Like this:

```diff
-    if self.is_synchronous_mode() and self.cluster.sync and self.cluster.sync.leader:
+    if self.is_synchronous_mode() and self.cluster.sync.leader:
```

It would be nice to have an automated test that we don't fail over if the available standby is not sync. Something like this barring any syntax issues:

```
  Scenario: check no failover when no sync standby
    Given I shut down postgres1
    Then "sync" key in DCS has sync_standby=null after 10 seconds
    When I shut down postgres0
    And I start postgres1
    Then postgres1 is not a leader after 10 seconds
    When I start postgres0
    Then postgres0 is a leader after 10 seconds
```

CyberDem0n:
Something like this: https://github.com/ants/patroni/pull/1/commits/958ab699f4d8f42eaf46eb3b3bd3673c4c9e027b ?

And I will add some more acceptance tests later.

CyberDem0n:
@ants there is one more pull request: https://github.com/ants/patroni/pull/2

Actually it is a bugfix, it was not possible to disable-enable synchronous_mode on the fly, because changing of `/config` key was causing `reload_config` call, rewriting of postgresql.conf and reload (without `synchronous_standby_names`)

CyberDem0n:
:+1:

docs/SETTINGS.rst
```diff
@@ -14,6 +14,7 @@ Bootstrap configuration
     -  **loop\_wait**: the number of seconds the loop will sleep. Default value: 10
     -  **ttl**: the TTL to acquire the leader lock. Think of it as the length of time before initiation of the automatic failover process. Default value: 30
     -  **maximum\_lag\_on\_failover**: the maximum bytes a follower may lag to be able to participate in leader election.
+    -  **synchronous\_mode**: turns on synchronous replication mode. Synchronous mode trades service availability for better durability guarantees. See `replication modes documentation <https://github.com/zalando/patroni/blob/master/docs/replication_modes.rst>`__ for details.
```
alexeyklyukin: I think the availability vs durability tradeoff should be explained in more details, since the whole point of Patroni is providing PostgreSQL HA. What about replacing the sentence starting with "Synchronous mode trades..." with "A replica running in synchronous mode is always in sync with the master. If the synchronous replica is not available, the master does not accept "write" transactions. Therefore, synchronous mode is a tradeoff between write availability and a strict data consistency between the master and the synchronous replica."

ants: A replica does not exactly run in synchronous mode, it's a cluster wide thing. I reworded it to be about lost transactions versus availability for writes and added a sentence about the mechanism. This should give a decent overview for both technically inclined and newbie users.

CyberDem0n:
:+1:

alexeyklyukin:
👍

alexeyklyukin:
Merged, thank you!

Issue #308, opened 2016-09-19
Isolated cluster doesn't demote master


jamessewell:
Hello,

I have a three node Patroni cluster which works fine. I've installed Patroni from pip, and I'm using PostgreSQL 9.5 from PGDG. I'm on Centos 7.2.

I'm simulating a total network outage with iptables, with DROP all INPUT and OUTPUT (keeping connections from my management machine, on lo0, and from each machine to itself). I can still run psql and connect locally to the local network IP.

Patroni seems to get the right idea, and I can see each node gets upset and keeps cycling through the other servers looking for an etcd connection. 

On the master I can see this (it loops):

```
2016-09-20 09:06:58,447 INFO: Reconnection allowed, looking for another server.
2016-09-20 09:06:58,447 INFO: Selected new etcd server http://192.168.2.246:2379
2016-09-20 09:07:00,117 WARNING: Retrying (Retry(total=0, connect=None, read=None, redirect=0)) after connection broken by 'ConnectTimeoutError(<urllib3.connection.HTTPConnection object at 0x1a20fd0>, u'Connection to 192.168.2.246 timed out. (connect timeout=1.66666666667)')': /v2/keys/service/postgresql/leader
2016-09-20 09:07:01,787 ERROR: Request to server http://192.168.2.246:2379 failed: MaxRetryError(u"HTTPConnectionPool(host=u'192.168.2.246', port=2379): Max retries exceeded with url: /v2/keys/service/postgresql/leader (Caused by ConnectTimeoutError(<urllib3.connection.HTTPConnection object at 0x1a20810>, u'Connection to 192.168.2.246 timed out. (connect timeout=1.66666666667)'))",)
2016-09-20 09:07:01,787 INFO: Reconnection allowed, looking for another server.
2016-09-20 09:07:01,787 ERROR: Machines cache is empty, no machines to try.
2016-09-20 09:07:01,787 ERROR: failed to update leader lock
2016-09-20 09:07:01,793 INFO: demoting self because i do not have the lock and i was a leader
2016-09-20 09:07:01,795 WARNING: Loop time exceeded, rescheduling immediately.
```

But PostgreSQL is never dropped from being a master (maybe this message is talking about the cluster state not the intention to demote PostgreSQL?)

```
-bash-4.2$ pg_controldata  | grep state
Database cluster state:               in production

```
```
[root@localhost ansible]# curl http://192.168.2.245:8008/
[...]

PR #316, opened 2016-09-20
Demote immediately if failed to update leader lock

CyberDem0n:
If the Etcd node partitioned from rest of the cluster it is still
possible to read from it (though it returns some stale information),
but it is not possible to write into it.
Previously Patroni was trying to fetch the new cluster view from DCS in
order to figure out is it still the leader or not and Etcd is always
returning stale info where the node still owns the leader key, but with
negative TTL.
This weird bug clearly shows how dangerous premature optimization is.

Fixes https://github.com/zalando/patroni/issues/308

### DIFF
--- a/patroni/ha.py
+++ b/patroni/ha.py
@@ -434,7 +434,8 @@ class Ha(object):
             else:
                 # Either there is no connection to DCS or someone else acquired the lock
                 logger.error('failed to update leader lock')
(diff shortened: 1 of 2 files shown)


alexeyklyukin:
👍

CyberDem0n:
:+1:

PR #320, opened 2016-09-21
Cover case when failover is now and via DCS

Sulion:
In the line 631 call goes to the dcs module. If the `scheduled_at` is `now`, then `parse_scheduled` returned `None` and anyway `failover_value` is passed as an argument, which, as a `dict`, has no `isoformat` method. It then happily fails here:

        if scheduled_at:
            failover_value['scheduled_at'] = scheduled_at.isoformat()

https://github.com/zalando/patroni/blob/master/patroni/dcs/__init__.py#L412

      File "/usr/local/bin/patronictl", line 9, in <module>
        load_entry_point('patroni==1.1', 'console_scripts', 'patronictl')()
      File "/usr/local/lib/python2.7/dist-packages/click/core.py", line 716, in __call__
        return self.main(*args, **kwargs)
      File "/usr/local/lib/python2.7/dist-packages/click/core.py", line 696, in main
        rv = self.invoke(ctx)
      File "/usr/local/lib/python2.7/dist-packages/click/core.py", line 1060, in invoke
        return _process_result(sub_ctx.command.invoke(sub_ctx))
      File "/usr/local/lib/python2.7/dist-packages/click/core.py", line 889, in invoke
        return ctx.invoke(self.callback, **ctx.params)
      File "/usr/local/lib/python2.7/dist-packages/click/core.py", line 534, in invoke
        return callback(*args, **kwargs)
      File "/usr/local/lib/python2.7/dist-packages/patroni/ctl.py", line 631, in failover
        dcs.manual_failover(master, candidate, scheduled_at=failover_value)
      File "/usr/local/lib/python2.7/dist-packages/patroni/dcs/__init__.py", line 397, in manual_failover
        failover_value['scheduled_at'] = scheduled_at.isoformat()

Instead, the code shall return valid date for `now` case and pass a correct value to the `dcs` module.

### DIFF
--- a/patroni/ctl.py
+++ b/patroni/ctl.py
@@ -454,7 +454,7 @@ def parse_scheduled(scheduled):
             raise PatroniCtlException(message.format(scheduled))
         return scheduled_at
 
-    return None
+    return tzlocal.get_localzone().localize(datetime.datetime.now())
 
 
 @ctl.command('restart', help='Restart cluster member')
@@ -628,7 +628,7 @@ def failover(config_file, cluster_name, master, candidate, force, dcs, scheduled
         logging.warning('Failing over to DCS')
         click.echo(timestamp() + ' Could not failover using Patroni api, falling back to DCS')
         click.echo(timestamp() + ' Initializing failover from master {0}'.format(master))
-        dcs.manual_failover(master, candidate, scheduled_at=failover_value)
+        dcs.manual_failover(master, candidate, scheduled_at=parse_scheduled(scheduled))
 
     output_members(cluster, cluster_name)
 
--- a/patroni/ctl.py
+++ b/patroni/ctl.py
@@ -454,7 +454,7 @@ def parse_scheduled(scheduled):
             raise PatroniCtlException(message.format(scheduled))
         return scheduled_at
 
-    return tzlocal.get_localzone().localize(datetime.datetime.now())
+    return None
 
 
 @ctl.command('restart', help='Restart cluster member')
@@ -590,12 +590,13 @@ def failover(config_file, cluster_name, master, candidate, force, dcs, scheduled
 
     scheduled_at = parse_scheduled(scheduled)
 
+    scheduled_at_str = None
     if scheduled_at:
         if cluster.is_paused():
             raise PatroniCtlException("Can't schedule failover in the paused state")
-        scheduled_at = scheduled_at.isoformat()
+        scheduled_at_str = scheduled_at.isoformat()
 
-    failover_value = {'leader': master, 'candidate': candidate, 'scheduled_at': scheduled_at}
+    failover_value = {'leader': master, 'candidate': candidate, 'scheduled_at': scheduled_at_str}
 
     logging.debug(failover_value)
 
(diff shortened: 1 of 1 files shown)


patroni/ctl.py
```diff
@@ -454,7 +454,7 @@ def parse_scheduled(scheduled):
             raise PatroniCtlException(message.format(scheduled))
         return scheduled_at
 
-    return None
```
CyberDem0n: If we want to failover immediately, we should set `scheduled_at` to `None`.

patroni/ctl.py
```diff
@@ -628,7 +628,7 @@ def failover(config_file, cluster_name, master, candidate, force, dcs, scheduled
         logging.warning('Failing over to DCS')
         click.echo(timestamp() + ' Could not failover using Patroni api, falling back to DCS')
         click.echo(timestamp() + ' Initializing failover from master {0}'.format(master))
-        dcs.manual_failover(master, candidate, scheduled_at=failover_value)
```
CyberDem0n: You are right, there is a bug on the line 631, it should look like:
```python
dcs.manual_failover(master, candidate, scheduled_at=scheduled_at)
```

But change on the line 457 is not needed and even dangerous.

Sulion: Type is lost at line [596](https://github.com/zalando/patroni/blob/master/patroni/ctl.py#L596):
    scheduled_at = scheduled_at.isoformat()
It's string from now on, not date. The `isoformat` method in [dcs](https://github.com/zalando/patroni/blob/master/patroni/dcs/__init__.py#L413) will fail all the same.

Sulion: Type is lost at line [596](https://github.com/zalando/patroni/blob/master/patroni/ctl.py#L596):
    scheduled_at = scheduled_at.isoformat()
It's string from now on, not date. The `isoformat` method in [dcs](https://github.com/zalando/patroni/blob/master/patroni/dcs/__init__.py#L413) will fail all the same.

patroni/ctl.py
CyberDem0n: Once again:

If we want to failover immediately, we should set `scheduled_at` to `None`

PERIOD

That is how Patroni works. If you set it to the current time (now) it might be DISCARDED due to time difference between the host where you run patronictl and the host where the leader runs.

patroni/ctl.py
CyberDem0n: You are right. We can't pass string there.
Can you please create a copy of scheduled_at on the line 591 and later use this copy in `dcs.manual_failover` call (line 631)?

patroni/ctl.py
Sulion: Alright, alright. There could still be a solution which will satisfy all the parties involved. One moment, please.

patroni/ctl.py
Sulion: I think, I kinda did: https://github.com/zalando/patroni/pull/320/commits/6321a1827be4e370af6f71a1058625700058a04d

CyberDem0n:
:+1:

alexeyklyukin:
👍

alexeyklyukin:
Merged, thank you!

PR #321, opened 2016-09-26
Call touch_member at the end of HA loop

CyberDem0n:
To make sure that we have up-to-date state of member in DCS after HA loop has changed something.

### DIFF
--- a/patroni/ha.py
+++ b/patroni/ha.py
@@ -165,7 +165,6 @@ class Ha(object):
             return message
         else:
             self.state_handler.promote()
-            self.touch_member()
             return promote_message
 
     @staticmethod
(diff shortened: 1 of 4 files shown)


patroni/ha.py
```diff
@@ -581,10 +579,12 @@ def post_recover(self):
         return None
 
     def _run_cycle(self):
+        dcs_failed = False
         try:
             self.load_cluster_from_dcs()
 
-            self.touch_member()
+            if not self.cluster.has_member(self.state_handler.name) or self.cluster.is_unlocked():
```
alexeyklyukin: Why do we do it also when cluster.is_unlocked?

alexeyklyukin:
👍

CyberDem0n:
:+1:

PR #323, opened 2016-09-26
Write all postgres parameters into postgresql.conf

CyberDem0n:
It will make possible to start postgres easily with pg_ctl (without
Patroni). Previously if you tried to start such postgres manually
it was usually failing because some of the parameters does not much
with values written into xlog (for example max_connections).


patroni/postgresql.py
```diff
@@ -154,7 +155,8 @@ def get_server_parameters(self, config):
         parameters = config['parameters'].copy()
         listen_addresses, port = (config['listen'] + ':5432').split(':')[:2]
         parameters.update({'cluster_name': self.scope, 'listen_addresses': listen_addresses, 'port': port})
-        return parameters
+        major_version = self._major_version or 9.1
```
alexeyklyukin: Should be 9.2, as our code doesn't work with 9.1 (it lacks support for row_to_json and array_to_json at least).

alexeyklyukin:
👍

CyberDem0n:
:+1:

Issue #333, opened 2016-10-06
Can't omit 'users' hash from 'bootstrap' stanza


sm97:
From `SETTINGS.rst`
```
users: Some additional users users which needs to be created after initializing new cluster
```

So doc doesn't insists that `users` hash should be mandatory. But starting Patroni with commented out `users` hash fails with following exception:
```
Traceback (most recent call last):
  File "/opt/python-envs/patroni/bin/patroni", line 9, in <module>
    load_entry_point('patroni==1.0', 'console_scripts', 'patroni')()
  File "/opt/python-envs/patroni/lib/python2.7/site-packages/patroni/__init__.py", line 128, in main
    patroni.run()
  File "/opt/python-envs/patroni/lib/python2.7/site-packages/patroni/__init__.py", line 102, in run
    logger.info(self.ha.run_cycle())
  File "/opt/python-envs/patroni/lib/python2.7/site-packages/patroni/ha.py", line 502, in run_cycle
    return self._run_cycle()
  File "/opt/python-envs/patroni/lib/python2.7/site-packages/patroni/ha.py", line 464, in _run_cycle
    return self.bootstrap()  # new node
  File "/opt/python-envs/patroni/lib/python2.7/site-packages/patroni/ha.py", line 93, in bootstrap
    self.state_handler.bootstrap(self.patroni.config['bootstrap'])
  File "/opt/python-envs/patroni/lib/python2.7/site-packages/patroni/postgresql.py", line 914, in bootstrap
    for name, value in config['users'].items():
KeyError: 'users'
```

CyberDem0n:
Your fix is merged, thank you

PR #334, opened 2016-10-06
Make 'users' hash of 'bootstrap' stanza optional

sm97:
https://github.com/zalando/patroni/issues/333

### DIFF
--- a/patroni/postgresql.py
+++ b/patroni/postgresql.py
@@ -1006,9 +1006,10 @@ $$""".format(name, ' '.join(options)), name, password, password)
     def bootstrap(self, config):
         """ Initialize a new node from scratch and start it. """
         if self._initialize(config) and self.start() and self.run_bootstrap_post_init(config):
-            for name, value in config['users'].items():
-                if name not in (self._superuser.get('username'), self._replication['username']):
-                    self.create_or_update_role(name, value['password'], value.get('options', []))
+            if 'users' in config:
+                for name, value in config['users'].items():
+                    if name not in (self._superuser.get('username'), self._replication['username']):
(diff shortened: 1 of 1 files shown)


patroni/postgresql.py
```diff
@@ -1006,9 +1006,10 @@ def clone(self, clone_member):
     def bootstrap(self, config):
         """ Initialize a new node from scratch and start it. """
         if self._initialize(config) and self.start() and self.run_bootstrap_post_init(config):
-            for name, value in config['users'].items():
-                if name not in (self._superuser.get('username'), self._replication['username']):
-                    self.create_or_update_role(name, value['password'], value.get('options', []))
+            if 'users' in config:
```
alexeyklyukin: It will fail if the users key is empty. I propose the following one-liner instead:
```
diff --git a/patroni/postgresql.py b/patroni/postgresql.py
index 814fb6c..332a99d 100644
--- a/patroni/postgresql.py
+++ b/patroni/postgresql.py
@@ -1006,7 +1006,7 @@ $$""".format(name, ' '.join(options)), name, password, password)
     def bootstrap(self, config):
         if self._initialize(config) and self.start() and self.run_bootstrap_post_init(config):
-            for name, value in config['users'].items():
+            for name, value in (config.get('users') or {}).items():
                 if name not in (self._superuser.get('username'), self._replication['username']):
                     self.create_or_update_role(name, value['password'], value.get('options', []))
             self.create_or_update_role(self._replication['username'], self._replication['password'], ['REPLICATION'])
```

alexeyklyukin:
👍

CyberDem0n:
:+1:

sm97:
Thanks, @alexeyklyukin ! It works like a charm :)

Issue #329, opened 2016-10-06
Old leader thinks he still has the lock after DCS error


zenitraM:
We are using Consul as DCS and had one of the Consul servers flapping, causing a momentary loss of leader and making Consul returning a 500 error.

This is the Patroni log on the previously master node:
```
2016-10-04 20:30:36,486 INFO: demoted self because DCS is not accessible and i was a leader
2016-10-04 20:30:38,309 ERROR: get_cluster
Traceback (most recent call last):
  File "/opt/python_virtualenvs/patroni/lib/python3.5/site-packages/patroni/dcs/consul.py", line 131, in _load_cluster
    _, results = self._client.kv.get(path, recurse=True)
  File "/opt/python_virtualenvs/patroni/lib/python3.5/site-packages/consul/base.py", line 466, in get
    params=params)
  File "/opt/python_virtualenvs/patroni/lib/python3.5/site-packages/patroni/dcs/consul.py", line 51, in get
    return callback(self.response(self.session.get(self.uri(path, params), verify=self.verify, timeout=timeout)))
  File "/opt/python_virtualenvs/patroni/lib/python3.5/site-packages/consul/base.py", line 186, in cb
    CB.__status(response, allow_404=allow_404)
  File "/opt/python_virtualenvs/patroni/lib/python3.5/site-packages/consul/base.py", line 143, in __status
    raise ConsulException("%d %s" % (response.code, response.body))
consul.base.ConsulException: 500 rpc error: No cluster leader
2016-10-04 20:30:38,309 ERROR: Error communicating with DCS
2016-10-04 20:30:38,310 INFO: establishing a new patroni connection to the postgres cluster
2016-10-04 20:30:38,327 INFO: DCS is not accessible
2016-10-04 20:30:48,346 ERROR: touch_member
Traceback (most recent call last):
  File "/opt/python_virtualenvs/patroni/lib/python3.5/site-packages/patroni/dcs/consul.py", line 197, in touch_member
    self._client.kv.put(self.member_path, data, **args)
  File "/opt/python_virtualenvs/patroni/lib/python3.5/site-packages/consul/base.py", line 533, in put
    CB.json(), '/v1/kv/%s' % key, params=params, data=value)
  File "/opt/python_virtualenvs/patroni/lib/python3.5/site-packages/consul/std.py", line 40, in put
    self.session.put(uri, data=data, verify=self.verify)))
[...]

PR #336, opened 2016-10-07
Do session/renew when update_leader is called

CyberDem0n:
should fix https://github.com/zalando/patroni/issues/329

in addition to that make it possible to specify ttl in Consul smaller than 20 (10 seconds).

### DIFF
--- a/features/environment.py
+++ b/features/environment.py
@@ -194,7 +194,7 @@ class AbstractDcsController(AbstractController):
 
     def stop(self, kill=False, timeout=15):
         """ terminate process and wipe out the temp work directory, but only if we actually started it"""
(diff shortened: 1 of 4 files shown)


features/environment.py
```diff
@@ -192,9 +192,9 @@ def __init__(self, context, mktemp=True):
     def _is_accessible(self):
         return self._is_running()
 
-    def stop_and_remove_work_directory(self, timeout=15):
+    def stop(self, kill=False, timeout=15):
```
alexeyklyukin: I wonder if we should also pass kill to the AbastractController.

alexeyklyukin:
👍

alexeyklyukin:
👍

CyberDem0n:
:+1:

PR #340, opened 2016-10-11
Updated README again

LappleApple:
Sorry, this version is better


README.rst
```diff
@@ -1,14 +1,14 @@
 |Build Status| |Coverage Status|
 
-Patroni: A Template for PostgreSQL HA with ZooKeeper, etcd or Consul
+Patroni: Template for PostgreSQL HA with Kubernetes, ZooKeeper, etcd or Consul
```
CyberDem0n: Somebody might think that Kubernetes is some strange animal similar to ZooKeeper. Very confusing :(

alexeyklyukin:
This is wrong:
- Patroni does not use K8S
- Patroni is completely unaware of K8S
- Kubernetes, Zookeeper and Etcd are configuration systems used to store the state of the PostgreSQL cluster. K8S does not belong to this list.

alexeyklyukin:
👍

PR #343, opened 2016-10-20
Add watchdog support on Linux

ants:
Ensures that system gets rebooted before TTL runs out.

Initial version. Open questions:
- Do we want to disable watchdog while we are not master?

TODO:
- merge with master start timeout
- When demoting send keepalives as soon as PostgreSQL terminates client connections to avoid unnecessary reboots.
- When shutting down maintain leader lock until client connections are terminated. Otherwise long shutdown checkpoint causes a reboot.


ants:
Ugh, merging another feature certainly didn't make this PR any clearer. 7b75310 is the relevant main commit.

The keepalive at the end of HA loop turned out be subtly wrong - there are some "should be impossible" cases calling follow() while state_handler.is_leader() where it would be best to postpone keepalive until stop completes. But more importantly handle_long_action_in_progress() should do something sane when update_lock() fails or we lost the lock while running async action (another one of those "should be impossible with watchdog" cases) To fix the follow case and allow early keepalives when stopping I need to reimplent pg_ctl stop in Python, not sure what to do about the async case yet. The interesting case seems to be the following: - master is restarted, DCS lock update fails (or DCS lock times out) while restart is running, we keep happily updating keepalives because running async action as standby is fine, postgres comes up as master while we are not holding leader lock. I'll see if I have better ideas tomorrow.

ants:
Supporting dynamic config changes is one todo I forgot to write up. Seems straightforward enough.

I'm not sure about having mode=disabled, it's certainly safer from unexpected behavior. On the other hand, all the distros I have seen would require you to explicitly assign permissions anyway. Having to change the config adds an extra step. I have a systemd unit here that makes enabling watchdog support as easy as `systemctl enable patroni-watchdog`.

I did some research into how to reliably detect PostgreSQL shut down without having to wait for the shutdown checkpoint to complete. This is mostly relevant for demote('offline') case, as we need to hold off watchdog keepalives until PostgreSQL has completed shutdown process to the point that no backends survive, but having to wait for checkpointer to complete shutdown checkpoint would quite likely result in a reboot due to timeout. Doing an immediate shutdown wouldn't be much of an improvement from a usability standpoint. Postmaster state machine actually has a state for the specific condition we need here, but unfortunately I don't think there is any way to check this state from the outside. I guess polling the postmaster child processes and classifying them based on their command line would work, but seems somewhat fragile.

ants:
Last 6 commits are mostly refactorings. Let me know if you find anything distasteful. I plan to build upon this to complete the watchdog feature.

ants:
Main todos are now done. Async actions are still a problem (see comment in `handle_long_action_in_progress`) but otherwise seems to be taking shape. I don't really love how the keepalives are sprinkled around the code, but I don't have a better idea that would work without major refactoring of the HA loop.

patroni/postgresql.py
```diff
@@ -633,40 +743,167 @@ def checkpoint(self, connect_kwargs=None):
...
+        return True, True
+
+    def _wait_for_postmaster_stop(self, pid):
+        # This wait loop differs subtly from pg_ctl as we check for both the pid file going
+        # away and if the pid is running. This seems safer.
+        while pid == self.get_pid() and self.is_pid_running(pid):
+            time.sleep(STOP_POLLING_INTERVAL)
+
+    def _signal_postmaster_stop(self, mode):
+        pid = self.get_pid()
```
CyberDem0n: `pid` is string, not integer...

ants:
When is it safe to call `reap_children`? Simplest solution seems to be to add a reap_children call in the polling loop, but Postgres.stop is also called from the async thread. That said, async thread also uses subprocess, and there is no interlock with main thread reap_children there, so it has the same type of race condition. So maybe add reap_children to the polling loop and implement some mutual exclusion mechanism between subprocess calls and reap_children?

CyberDem0n:
> When is it safe to call reap_children?

It is safe to call it from the Main thread. Most important there should be no subprocess hanging around, which return code is important to get. I think `is_pid_running` should be relatively good place for it.

```python
@@ -13,7 +13,7 @@ import time
 
 from collections import defaultdict
 from patroni.exceptions import PostgresConnectionException, PostgresException
-from patroni.utils import compare_values, parse_bool, parse_int, Retry, RetryFailedError, polling_loop
+from patroni.utils import compare_values, parse_bool, parse_int, Retry, RetryFailedError, polling_loop, reap_children
 from six import string_types
 from threading import current_thread, Lock, Event
 
@@ -603,11 +603,12 @@ class Postgresql(object):
             logger.warning("Garbage pid in postmaster.pid: {0!r}".format(pid))
             return 0
 
-    @staticmethod
-    def is_pid_running(pid):
+    def is_pid_running(self, pid):
         try:
             if pid < 0:
                 pid = -pid
+            if current_thread().ident == self.__thread_ident:
+                reap_children()
             return pid > 0 and pid != os.getpid() and pid != os.getppid() and (os.kill(pid, 0) or True)
         except Exception:
             return False
```

ants:
I'm not convinced about it being safe to call even from the main thread, but that is a separate issue. This approach works for me. Stop called from asnyc thread might have to wait up to loop_wait if Postgresql crashes during shut down, but that doesn't like it would be a problem even in the unlikely event it did ever happen to someone.

CyberDem0n:
> I'm not convinced about it being safe to call even from the main thread

You are right, it is not safe, threads might loose error exit code as well (they get 0 instead) :(

LappleApple:
@CyberDem0n Hey, this PR from @ants has been with us for some time. Is there a way to resolve soon?

ants:
This is waiting on me to resolve the issues identified in the review. I was
on vacation for a month and am now dispatching some urgent tasks before I
get back to this. I expect to get to this in the next couple of days.

patroni/watchdog/base.py
```diff
@@ -0,0 +1,202 @@
...
+                               "{0} seconds when watchdog times out.".format(self.impl.get_timeout()))
+            self.impl.close()
+        except WatchdogError as e:
+            logger.error("Error while disabling watchdog: %s", e)
+
+    def keepalive(self):
+        try:
+            self.impl.keepalive()
+        except WatchdogError as e:
+            logger.error("Error while disabling watchdog: %s", e)
```
ants: Fixed.

patroni/watchdog/base.py
```diff
@@ -0,0 +1,202 @@
...
+
+    def _get_impl(self):
+        if self.mode not in ['automatic', 'required', 'require']:
```
ants: Missed this spot while refactoring. Fixed.

patroni/watchdog/linux.py
```diff
@@ -0,0 +1,211 @@
...
+    "TEMPPANIC": 0x0004,      # Kernel panic on temperature trip
+}
+
+# Implementation
+
+
+class WatchdogInfo(collections.namedtuple('WatchdogInfo', 'options,version,identity')):
+    """Watchdog descriptor from the kernel"""
+    @property
+    def option_list(self):
```
ants: It was useful for seeing how different watchdog drivers work, but I guess it's not really needed. Removed.

patroni/watchdog/linux.py
```diff
@@ -0,0 +1,211 @@
...
+        return [name for name, bit in WDIOF.items() if self.options & bit]
+
+
+def _property_func(wdiof_name):
```
ants: I guess that was premature optimization. Changed it to getattr. I flipped the if clause because I find negations make logic harder to understand, and having the default error case at the end makes it easier to add more cases (not that any are likely in this case).

ants:
Finally got around to this issue again. I'm on the fence if the feature test implementation is a good idea or not. Seems better than nothing, but it is a bunch of complicated scaffolding to make it work. I guess it's easy enough to rip out later if it proves to be a pain.

We should decide if we want the default mode to be `off` or `automatic`. Auto would make the feature really easy to enable, in my packaging it can be enabled with just a `systemctl enable patroni-watchdog`. But it currently does log a scary message on startup if it can't be enabled:

    2017-04-04 14:41:29,140 WARNING: Could not activate Linux watchdog device: "Can't open watchdog device: [Errno 2] No such file or directory: '/dev/watchdog'"

I don't have any pending changes for this feature. There are a couple of TODO's introduced in the patch, but they are pre-existing issues.

ants:
And I just noticed that the feature tests are completely broken. That's what you get when commiting stuff that has been stashed away for a couple of months without reviewing it properly. Looking into it...

patroni/postgresql.py
```diff
@@ -759,27 +789,111 @@ def checkpoint(self, connect_kwargs=None):
...
+            pass
+
+    @staticmethod
+    def _wait_for_user_backends_to_close(postmaster_pid):
+        # These regexps are cross checked against versions PostgreSQL 9.1 .. 9.6
+        aux_proc_re = re.compile("(?:postgres:)( .*:)? (?:""(?:startup|logger|checkpointer|writer|wal writer|"
+                                 "autovacuum launcher|autovacuum worker|stats collector|wal receiver|archiver|"
+                                 "wal sender) process|bgworker: )")
+
+        postmaster = psutil.Process(postmaster_pid)
```
CyberDem0n: `psutil.Process` will raise `psutil.NoSuchProcess` exception when process is not there.
And I think this is the reason why acceptance tests are failing on travis.

ants: Thanks, that was indeed broken, but not the cause of test failures. I think I got the tests somewhat robust now. Some timing issues, some stupid bugs of my own.

Scaffolding probably needs some improvement to properly handle stuff like start patroni and wait for it to come up. The sleep I have in the third watchdog feature for example is because otherwise API is not yet up.

CyberDem0n: > Thanks, that was indeed broken, but not the cause of test failures. 

Actually it was. According to watchdog logs Patroni shut down was taking ~120 seconds and "magic close" wasn't received from the pipe.

120 seconds comes from 60 seconds * 2, where 2 is a default timeout multiplier when we run tests on travis (sometimes travis is very slow).
features/environment.py sends a SIGTERM to Patroni and waits for 120 seconds until it will shut down normally. After timeout it sends a SIGKILL.

Patroni can't send "magic close" if psutil.Process raise an exception. And Patroni also doesn't shut down normally in this case, because one or more of the threads didn't terminated.
```
features/output/watchdog/watchdog.postgres0.log:2017-04-12 09:57:08,557218: watchdog postgres0 initialized
features/output/watchdog/watchdog.postgres0.log:2017-04-12 09:57:08,574592: starting fifo /home/travis/build/zalando/patroni/data/watchdog.postgres0.fifo
features/output/watchdog/watchdog.postgres0.log:2017-04-12 09:57:08,575271: opening
features/output/watchdog/watchdog.postgres0.log:2017-04-12 09:57:08,897032: Fifo /home/travis/build/zalando/patroni/data/watchdog.postgres0.fifo connected
features/output/watchdog/watchdog.postgres0.log:2017-04-12 09:57:08,897244: timeout=15
features/output/watchdog/watchdog.postgres0.log:2017-04-12 09:57:08,901129: ping after 0.326314926147 seconds
features/output/watchdog/watchdog.postgres0.log:2017-04-12 09:59:11,742916: Pipe closed
features/output/watchdog/watchdog.postgres0.log:2017-04-12 09:59:11,743001: closing
features/output/watchdog/watchdog.postgres0.log:2017-04-12 09:59:11,743034: opening
features/output/watchdog/watchdog.postgres0.log:2017-04-12 09:59:12,740208: Monitor stop
```

patroni/ha.py
```diff
@@ -46,6 +46,50 @@ def failover_limitation(self):
         return None
 
 
+class BackgroundKeepaliveSender(object):
+    """A context manager that sends keepalives every loop_wait seconds in a background thread while the context is
+    running, but only after a safepoint has been reached. After the safepoint PostgreSQL must not be allowed to
```
CyberDem0n: And why we don't want to send keepalives before safepoint has been reached?

Also according to your comment from Oct 24, 2016 this code is mostly needed for `demote('offline')`:
> I did some research into how to reliably detect PostgreSQL shut down without having to wait for the shutdown checkpoint to complete. This is mostly relevant for demote('offline') case, as we need to hold off watchdog keepalives until PostgreSQL has completed shutdown process to the point that no backends survive, but having to wait for checkpointer to complete shutdown checkpoint would quite likely result in a reboot due to timeout.

But it seems that exactly for `demote('offline')` it is not used...

patroni/ha.py
```diff
@@ -196,29 +262,34 @@ def _get_node_to_follow(self, cluster):
...
 
-        if self.is_paused() and not (self.state_handler.need_rewind and self.state_handler.can_rewind):
-            self.state_handler.set_role('master' if is_leader else 'replica')
-            if is_leader:
-                return 'continue to run as master without lock'
-            elif not node_to_follow:
-                return 'no action'
+        if self.is_paused():
+            self.keepalive()
+            if not self.state_handler.need_rewind or self.cluster.is_unlocked():
```
CyberDem0n: I wonder why this line was changed from `not (self.state_handler.need_rewind and self.state_handler.can_rewind)`  to `not self.state_handler.need_rewind or self.cluster.is_unlocked()`?

Original line was introduced in https://github.com/zalando/patroni/commit/82176765e9be148fc8e74d278efd5c49ce52df95#diff-57c53e988b775ee2902488fb33c1f433L211

CyberDem0n: Yeah, `or self.cluster.is_unlocked()` check is definitely useful here, but why `can_rewind` was stripped out?

CyberDem0n:
@ants, can you please have a look on 
https://github.com/zalando/patroni/pull/343/files/4be35ac5988dc0a5426b54b4a719a3acdb32f0c6#diff-57c53e988b775ee2902488fb33c1f433R51

and https://github.com/zalando/patroni/pull/343/files/4be35ac5988dc0a5426b54b4a719a3acdb32f0c6#diff-57c53e988b775ee2902488fb33c1f433R275

Everything else seems to be fine (except conflicts)

ants:
I will take a look as soon as I can. Currently very busy prepping slides
for a two day training on Patroni. Might not get to it before Monday.

patroni/ha.py
ants: That was a bug, fix pushed.

ants: And to answer your questions we don't want to send keepalives before safepoint because shutting down PostgreSQL might not work correctly. Keepalive can be sent when:

1. We have updated our leader lock.
2. PostgreSQL can't send back successful commits before further action is taken - either it's in standby mode, postgres is shut down or postgres is in process of shutting down and all backends have been closed.

CyberDem0n:
:+1:

alexeyklyukin:
👍

CyberDem0n:
I think there are still some open issues, but it would be much easier to fix them by separate pull requests rather than keeping your branch in sync with the master.

Thank you @ants!

ants:
Thanks a lot to you guys too. I will torture the result that got merged
some more to see if I can break it. Do you have some specific issues of
your own in mind besides stuff tagged as FIXME in the code?

CyberDem0n:
As far as I can recall:
1. Your code still does not support dynamic change of `loop_wait` and `ttl`.
2. Do we want to disable watchdog while we are not master?
3. Why do we have so many places where `keepalive` is send? Can it be covered by one call from the `finally` block of `_run_cycle`? Off course places where `BackgroundKeepaliveSender` is used are special, but there are not many of them.

ants:
1. I will create fix for dynamic ttl/loop_wait
2. I thought about that myself. I am not sure if we want to. I view the
watchdog as a feature for people who are concerned about corretness first
and availability second. When we are not master the watchdog can act as a
safeguard layer against misconfiguration - e.g. some other script runs on
the system and makes postgresql master behind Patroni's back, Patroni will
have the server rebooted if the situation can't be fixed quickly enough. On
the other hand, rebooting standbys somebody does a VM migration type of
operation is not exactly user friendly.
3. Main reason for the current way is that I wanted to send the keepalive
ASAP and so I designed it from the bottom up and went through all code
paths to find the earliest safe point to send the keepalive at. With
default settings the timing slack is uncomfortably small considering we are
occasionally waiting in the main HA loop. I agree that the current setup is
quite messy. I'm analyzing the whole setup again to figure out if it could
be simplified somehow. I will post an update with the results.

PR #352, opened 2016-11-03
HA loop decision diagram

### LINKED ISSUE #344: HA decision loop diagram
We would like to create (mainly for our ops documentation / debugging) a diagram or some documentation of how the Patroni HA loop takes decisions.

There was this [old one on Governor](https://github.com/compose/governor/blob/master/postgres-ha.pdf) but it was removed from Patroni.

Before we dive into the code, how current / correct is it?

zenitraM:
This is a first shot at it. I'm missing some things like the behavior changes introduced by pause mode; not sure how deep in detail would we want it to get on those.

Corrections, comments and ideas are welcome :)

Fixes #344


drnic:
Thanks for working on this. I liked showing off the old diagram to people when explaining patronij

CyberDem0n:
@zenitraM Good job!
Actually it has a lot of issues but it is still very useful. I've fixed most of them and result is available here: https://github.com/CartoDB/patroni/compare/docs/ha_loop_diagram...CyberDem0n:docs/ha_loop_diagram

It is still not the full decision loop, because there is no branches for: manual failover, pause (maintenance) mode, synchronous mode and pg_rewind logic (we are going to change it anyway).
Also it doesn't explain how we are making decision which node is the healthiest.  But I think it is already good enough to explain how automatic failover works.

PR #355, opened 2016-11-08
Solve issue of handling sigchld when dunning in a docker

CyberDem0n:
If Patroni was started in a docker with pid=1 it will execute itself
with the same arguments. The original process will take care about init
process duties, i.e. handle sigchld and reap dead orphan processes.
Also it will forward SIGINT, SIGHUP, SIGTERM and some other signals to
the real Patroni process.

### DIFF
--- a/patroni/__init__.py
+++ b/patroni/__init__.py
@@ -157,16 +157,13 @@ def main():
                 if ret == (0, 0):
                     break
                 elif ret[0] != pid:
-                    logging.info('Reaped unknow pid=%s, exit status=%s', *ret)
+                    logging.info('Reaped pid=%s, exit status=%s', *ret)
         except OSError:
             pass
 
     def passtochild(signo, stack_frame):
         if pid:
             os.kill(pid, signo)
-
-        if signo in (signal.SIGTERM, signal.SIGINT, signal.SIGQUIT):
-            sys.exit(0)
 
     signal.signal(signal.SIGCHLD, sigchld_handler)
     signal.signal(signal.SIGHUP, passtochild)
--- a/patroni/__init__.py
+++ b/patroni/__init__.py
@@ -1,16 +1,8 @@
 import logging
+import os
 import signal
 import sys
 import time
-
-from patroni.api import RestApiServer
-from patroni.config import Config
-from patroni.dcs import get_dcs
-from patroni.exceptions import DCSError
-from patroni.ha import Ha
-from patroni.postgresql import Postgresql
-from patroni.utils import reap_children, sigchld_handler
-from patroni.version import __version__
 
 logger = logging.getLogger(__name__)
 
@@ -18,6 +10,13 @@ logger = logging.getLogger(__name__)
 class Patroni(object):
 
     def __init__(self):
+        from patroni.api import RestApiServer
+        from patroni.config import Config
+        from patroni.dcs import get_dcs
+        from patroni.ha import Ha
+        from patroni.postgresql import Postgresql
+        from patroni.version import __version__
+
         self.setup_signal_handlers()
 
         self.version = __version__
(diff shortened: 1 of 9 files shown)


patroni/utils.py
```diff
@@ -197,36 +194,8 @@ def compare_values(vartype, unit, old_value, new_value):
...
-        try:
-            while True:
-                ret = os.waitpid(-1, os.WNOHANG)
-                if ret == (0, 0):
-                    break
-        except OSError:
-            pass
-        finally:
-            __reap_children = False
+def _sleep(interval):
```
CyberDem0n: `time.sleep` should not be interrupted because we are not handling SIGCHLD anymore.

CyberDem0n:
@ants, please take a look on it

patroni/__init__.py
```diff
@@ -145,3 +141,42 @@ def main():
...
+                    logging.info('Reaped unknow pid=%s, exit status=%s', *ret)
+        except OSError:
+            pass
+
+    def passtochild(signo, stack_frame):
+        if pid:
+            os.kill(pid, signo)
+
+        if signo in (signal.SIGTERM, signal.SIGINT, signal.SIGQUIT):
+            sys.exit(0)
```
ants: At least for SIGINT it would make sense to pass it on. Not an expert on containers, but I would guess that otherwise the child processes might get terminated before they can execute a graceful shutdown. Looking at a [project that has a similar goal](https://github.com/Yelp/dumb-init/blob/master/dumb-init.c#L90) doing something about SIGTSTP, SIGTTOU, SIGTTIN might or might not make sense here. Otherwise the code seems to match what dumb-init does.

patroni/utils.py
ants: It could still be interrupted by SIGHUP.

patroni/__init__.py
```diff
@@ -145,3 +141,42 @@ def main():
...
+
+    # Looks like we are in a docker, so we will act like init
+    def sigchld_handler(signo, stack_frame):
+        try:
+            while True:
+                ret = os.waitpid(-1, os.WNOHANG)
+                if ret == (0, 0):
+                    break
+                elif ret[0] != pid:
+                    logging.info('Reaped unknow pid=%s, exit status=%s', *ret)
```
alexeyklyukin: Should be 'unknown', but actually everything except for the child Patroni pid will be considered 'unknown' (including PostgreSQL and archive/restore command), so there is no point in mentioning it (showing the pid and ret status could still be useful for debugging purposes).

patroni/__init__.py
```diff
@@ -145,3 +141,42 @@ def main():
...
+        except OSError:
+            pass
+
+    def passtochild(signo, stack_frame):
+        if pid:
+            os.kill(pid, signo)
+
+        if signo in (signal.SIGTERM, signal.SIGINT, signal.SIGQUIT):
```
alexeyklyukin: Yes, that's what I actually have in mind: instead of calling sys.exit just rely on the part of the code that does wait and then pass the result of the wait in a sys.exit (this might look pointless for the PID 1, but it would allow the docker ps to see the return code).

patroni/utils.py
CyberDem0n: Yep, it could, but it was never handled before. Should we really care?
SIGCHLD was really a problem, because on the master sleep was interrupted all the time by wal-e and ha loop was executed not every `loop_wait` interval, but much more often...

alexeyklyukin: Well, we use small timeouts in order to give other nodes the opportunity to take the leader lock, if Patroni receives SIGHUP at that time it will think by mistake that there is no master to rewind. There is very very small chance of seeing it in the wild though...

patroni/__init__.py
CyberDem0n: It will pass everything listed on the lines 172-177. And @alexeyklyukin  is right about `sys.exit`, we should not call it, but wait until the child process will exit on the line 182. Fix is coming.

alexeyklyukin:
👍

ants:
Looks good to me.

CyberDem0n:
:+1:

PR #356, opened 2016-11-09
Improve leader watch functionality

CyberDem0n:
Previously replicas were always watching for leader key (even if the
postgres was not in the running there). It was not a big issue, but it
was not possible to interrupt such watch in cases if the postgres
started up or stopped successfully. Also it was delaying update_member
call and we had kind of stale information in DCS up to `loop_wait`
seconds. This commit changes such behavior. If the async_executor is
busy by starting/stopping or restarting postgres we will not watch for
leader key but waiting for event from async_executor up to `loop_wait`
seconds. Async executor will fire such event only in case if the
function it was calling returned something what could be evaluated to
boolean True.

Such functionality is really needed to change the way how we are making
decision about necessity of pg_rewind. It will require to have a local
postgres running and for us it is really important to get such
notification as soon as possible.

### DIFF
--- a/patroni/ha.py
+++ b/patroni/ha.py
@@ -25,7 +25,7 @@ class Ha(object):
         self.cluster = None
         self.old_cluster = None
         self.recovering = False
-        self._async_executor = AsyncExecutor(self)
+        self._async_executor = AsyncExecutor(self.wakeup)
 
         # Each member publishes various pieces of information to the DCS using touch_member. This lock protects
         # the state and publishing procedure to have consistent ordering and avoid publishing stale values.
@@ -812,4 +812,7 @@ class Ha(object):
         return self.dcs.watch(leader_index, timeout)
 
     def wakeup(self):
+        """Call of this method will trigger the next run of HA loop if there is
+        no "active" leader watch request in progress.
+        This usually happens on the master or if the node is running async action"""
         self.dcs.event.set()
--- a/patroni/ha.py
+++ b/patroni/ha.py
@@ -666,7 +666,7 @@ class Ha(object):
 
         clone_member = self.cluster.get_clone_member(self.state_handler.name)
         member_role = 'leader' if clone_member == self.cluster.leader else 'replica'
-        self.clone(clone_member, "from {0} '{1}'".format(member_role, clone_member.name))
+        return self.clone(clone_member, "from {0} '{1}'".format(member_role, clone_member.name))
 
     def reinitialize(self):
         with self._async_executor:
--- a/patroni/ha.py
+++ b/patroni/ha.py
@@ -738,11 +738,6 @@ class Ha(object):
                 if msg is not None:
                     return msg
 
-                if not self.has_lock() and len(self.cluster.members) == 1 and \
-                        not (self.is_synchronous_mode() and self.cluster.sync and
-                             self.cluster.sync.matches(self.state_handler.name)):
-                    return 'too few members in cluster after "recovery", postponing leader race'
-
             # is data directory empty?
             if self.state_handler.data_directory_empty():
                 return self.bootstrap()  # new node
--- a/patroni/ha.py
+++ b/patroni/ha.py
@@ -25,7 +25,7 @@ class Ha(object):
         self.cluster = None
         self.old_cluster = None
         self.recovering = False
-        self._async_executor = AsyncExecutor()
+        self._async_executor = AsyncExecutor(self)
 
         # Each member publishes various pieces of information to the DCS using touch_member. This lock protects
         # the state and publishing procedure to have consistent ordering and avoid publishing stale values.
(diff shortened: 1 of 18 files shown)


patroni/ha.py
```diff
@@ -793,3 +799,17 @@ def run_cycle(self):
...
+        if not self._async_executor.busy and cluster and cluster.leader \
+                and cluster.leader.name != self.state_handler.name:
+            leader_index = cluster.leader.index
+        else:
+            leader_index = None
+
+        return self.dcs.watch(leader_index, timeout)
+
+    def wakeup(self):
+        self.dcs.event.set()
```
ants: As far as I can tell `wakeup` only works for when the current node is master or running an async action. That should at least be documented. Ideally we should look into interrupting DCS watches on wakeup. I know I will forget `wakeup()` actually means `wakeup_would_be_nice_thankyou_not_a_problem_if_you_cant_right_now()`.

patroni/ha.py
```diff
@@ -730,13 +730,19 @@ def _run_cycle(self):
...
+                # Check if we tried to recover and failed
                 msg = self.post_recover()
                 if msg is not None:
                     return msg
 
+                if not self.has_lock() and len(self.cluster.members) == 1 and \
+                        not (self.is_synchronous_mode() and self.cluster.sync and
+                             self.cluster.sync.matches(self.state_handler.name)):
+                    return 'too few members in cluster after "recovery", postponing leader race'
+
```
CyberDem0n: I was just trying not to change much existing behavior. I.e. on the first iteration it was starting postgres and only on the next iteration (after loop_wait) it was doing reader race and promotion.

Such delay might be useful only in one case: if the cluster is completely down and you are bringing it back (all nodes more or less at the same time).

CyberDem0n:
>Good idea. The `while_not_sync_standby` code could also use a way to prod the master to release a standby quicker.

We can do this, via REST API, just need to implement endpoint and call it.

> An unfinished idea that popped up while thinking about this - this is a step in the direction of making the HA loop event based. 

Yep, that was my idea as well. But unfortunately we can't make it 100% event based due to the nature how currently `etcd.watch` and `consul.watch` are working. Well, it is possible to make those methods fully asynchronous, but the price would be: double amount of open http connections to the 'etcd'... :(

ants:
Looks good to me.

patroni/ha.py
```diff
@@ -432,7 +432,7 @@ def demote(self, delete_leader=True):
             sleep(2)  # Give a time to somebody to take the leader lock
             cluster = self.dcs.get_cluster()
             node_to_follow = self._get_node_to_follow(cluster)
-            self.state_handler.follow(node_to_follow, cluster.leader, recovery=True, need_rewind=True)
+            return self.state_handler.follow(node_to_follow, cluster.leader, recovery=True, need_rewind=True)
```
alexeyklyukin: I think we also need to add the same to _do_reinitialize, since it is also called asynchronously, so the leader key waiting logic won't apply anymore (not as import for pg_rewind though).

patroni/ha.py
alexeyklyukin: Ah, so while previously we allowed a process to wait for the whole loop_wait after calling recover (which was ultimately calling state_handler.follow asynchronously, therefore, had to wait for at least loop_wait afterwards), with this patch the loop will be terminated as soon as the follow will return True, which could be faster than loop_wait (but not necessary), therefore, starting much faster and giving less time for the other cluster members to start. Ultimately, I think this doesn't really give us any guarantees (it could happen that the first 2 nodes to start would be the "slowest" ones) and makes the whole block harder to comprehend, so I vote for removing it.

patroni/postgresql.py
```diff
@@ -668,7 +668,7 @@ def restart(self):
             self.call_nowait(ACTION_ON_RESTART)
         else:
             self.set_state('restart failed ({0})'.format(self.state))
-        return ret
+        return True
```
alexeyklyukin: I think that would break the code in ha.py(namely in def restart) that checks whether the restart returns False in order to report the issue to the client.

patroni/postgresql.py
```diff
@@ -862,20 +862,22 @@ def _do_follow(self, primary_conninfo, leader, recovery=False):
...
+            if recovery:
+                self.start()
+            else:
+                self.restart()
             self.set_role('replica')
 
         if change_role:
             self.call_nowait(ACTION_ON_ROLE_CHANGE)
-        return ret
+        return True
```
alexeyklyukin: _do_follow result was ignored in the past, but now since we use it to wake-up the process, should we return a real result instead of always True?

patroni/postgresql.py
CyberDem0n: Good catch, will fix it.

alexeyklyukin:
👍

CyberDem0n:
:+1:

PR #357, opened 2016-11-23
Add postgres-9.6 support

CyberDem0n:
starting from 9.6 we need to set wal_level = 'replica' which is alias for
'hot_standby'. It was working before without problems, but if somebody
change wal_level to 'replica', Patroni will expose pending_restart flag,
although restart in this case is not necessary.

Besides that bump versions of consul and etcd to the latest.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,26 +1,41 @@
 ## This Dockerfile is meant to aid in the building and debugging patroni whilst developing on your local machine
 ## It has all the necessary components to play/debug with a single node appliance, running etcd
 FROM ubuntu:16.04
-MAINTAINER Feike Steenbergen <feike.steenbergen@zalando.de>
+MAINTAINER Alexander Kukushkin <alexander.kukushkin@zalando.de>
 
 RUN echo 'APT::Install-Recommends "0";' > /etc/apt/apt.conf.d/01norecommend \
     && echo 'APT::Install-Suggests "0";' >> /etc/apt/apt.conf.d/01norecommend
 
-ENV PGVERSION 9.5
+ENV PGVERSION 9.6
 ENV PATH /usr/lib/postgresql/${PGVERSION}/bin:$PATH
 RUN apt-get update -y \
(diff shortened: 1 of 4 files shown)


Dockerfile
```diff
@@ -1,26 +1,41 @@
...
+
+    ## Make sure we have a en_US.UTF-8 locale available
+    && localedef -i en_US -c -f UTF-8 -A /usr/share/locale/locale.alias en_US.UTF-8 \
+
+    && apt-get update -y \
+    && apt-get install -y postgresql-contrib-${PGVERSION} \
+
+    # Remove the default cluster, which Debian stupidly starts right after installation of the packages
+    && pg_dropcluster --stop ${PGVERSION} main \
     && pip install python-etcd==0.4.3 python-consul==0.6.0 --upgrade \
```
alexeyklyukin: There is python-consul==0.6.1 in pip

patroni/postgresql.py
```diff
@@ -597,8 +599,10 @@ def start(self, block_callbacks=False):
         self._write_postgresql_conf()
         self.resolve_connection_addresses()
 
-        options = ' '.join("--{0}='{1}'".format(p, self._server_parameters[p]) for p, v in self.CMDLINE_OPTIONS.items()
-                           if self._major_version >= v[2])
+        opts = {p: self._server_parameters[p] for p, v in self.CMDLINE_OPTIONS.items() if self._major_version >= v[2]}
+        if self._major_version >= 9.6 and opts['wal_level'] == 'hot_standby':
+            opts['wal_level'] = 'logical'
```
alexeyklyukin: shouldn't it be 'replica' instead of 'logical'?

CyberDem0n: yep...

Dockerfile
CyberDem0n: It's not that important, but will fix it

alexeyklyukin:
👍

CyberDem0n:
:+1:

PR #359, opened 2016-12-01
Fix the WAL-E restore

alexeyklyukin:
Namely:
- broken xlog symlink after the restore
- retry only when there were an exception, not when the decision was made not to use the WAL-E because of the information from the master.

- Challenge the use of WAL-E even when 'no_master' flag is set. This flag in
  fact does not indicate that the master is absent. In order to check the master
  absense the script looks whether the connection string is not empty.

- Retry on a failure to fetch current xlog position from the master. The reason
  it has to be separate from retries in the main loop is that we don't just
  retry the connection attempt, but also make a decision when either it was
  successfull or all attempts are exhausted.

### DIFF
--- a/patroni/scripts/wale_restore.py
+++ b/patroni/scripts/wale_restore.py
@@ -136,7 +136,8 @@ class WALERestore(object):
                     with psycopg2.connect(self.master_connection) as con:
                         con.autocommit = True
                         with con.cursor() as cur:
-                            cur.execute("SELECT pg_xlog_location_diff(pg_current_xlog_location(), %s)", (backup_start_lsn,))
+                            cur.execute("SELECT pg_xlog_location_diff(pg_current_xlog_location(), %s)",
+                                        (backup_start_lsn,))
                             diff_in_bytes = long(cur.fetchone()[0])
                 except psycopg2.Error:
                     logger.exception('could not determine difference with the master location')
@@ -223,6 +224,7 @@ def main():
         time.sleep(RETRY_SLEEP_INTERVAL)
 
     return ret
+
 
 if __name__ == '__main__':
     sys.exit(main())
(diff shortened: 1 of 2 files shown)


patroni/scripts/wale_restore.py
```diff
@@ -134,6 +163,8 @@ def create_replica_with_s3(self):
             logger.error('Error when fetching backup with WAL-E: {0}'.format(e))
             return 1
 
+        self.fix_subdirectory_path_if_broken('pg_xlog' if get_major_version(self.data_dir) < 10.0 else 'pg_wal')
```
a1exsh: Nice precaution ;-)

a1exsh:
:+1:

CyberDem0n:
Where are the unit-tests? :)

alexeyklyukin:
👍

alexeyklyukin:
👍

patroni/scripts/wale_restore.py
```diff
@@ -115,8 +136,8 @@ def should_use_s3_to_create_replica(self):
                         cur.execute("SELECT pg_xlog_location_diff(pg_current_xlog_location(), %s)", (backup_start_lsn,))
                         diff_in_bytes = long(cur.fetchone()[0])
             except psycopg2.Error as e:
-                logger.error('could not determine difference with the master location: %s', e)
-                return False
+                logger.exception('could not determine difference with the master location: %s', e)
```
CyberDem0n: logger.exception will print stacktrace, so there is no need to pass `e` parameter.

And we should try to connect to the master even if `no_master` is set.

patroni/scripts/wale_restore.py
```diff
@@ -83,16 +104,16 @@ def should_use_s3_to_create_replica(self):
 
             backup_info = dict(zip(names, vals))
         except subprocess.CalledProcessError as e:
-            logger.error("could not query wal-e latest backup: {}".format(e))
-            return False
+            logger.exception("could not query wal-e latest backup: {}".format(e))
```
CyberDem0n: It should be just `logger.exception("could not query wal-e latest backup')` because `logger.exception` will print stacktrace

alexeyklyukin:
👍

CyberDem0n:
:+1:

PR #360, opened 2016-12-06
Implement simple asynchronos dns-resolve cache

CyberDem0n:


patroni/dcs/etcd.py
```diff
@@ -51,6 +103,9 @@ def _build_request_parameters(self):
         kwargs['timeout'] = self.read_timeout/float(kwargs['retries'] + 1)/etcd_nodes
         return kwargs
 
+    def set_machines_cache_ttl(self, cache_ttl):
```
alexeyklyukin: I think you should get rid of this method altogether. It's called twice: during the Client init, and when Etcd calls it as a part of set_ttl. There is no need to call it from Etcd, since you can just read your value directly in the Client from config['ttl'], and a single remaining call can be inlined. As a result, you don't need the default cache_ttl value as well.

CyberDem0n: Well, `ttl` may change in runtime

alexeyklyukin:
👍

alexeyklyukin:
👍

CyberDem0n:
:+1:

PR #365, opened 2016-12-16
Fix rewind behavior when paused

ants:
When `can_rewind` is False `need_rewind` will not get reset and when cluster enters paused mode and Postgres is shut down it will get restarted.

Also allow for rewind to work without an explicit superuser config. In some setups the default of OS user with no password auth is enough.


CyberDem0n:
:+1:

PR #366, opened 2016-12-16
Serialize callback execution

CyberDem0n:
If the previous callback is still running - kill it
Also it will fix a problem of zombie processes when executing callbacks from the main thread.


alexeyklyukin:
👍

CyberDem0n:
:+1:

PR #378, opened 2017-01-04
Optimize travis cache usage

CyberDem0n:
Travis cache is a little bit "dump". When it "restores" cache into directory with some existing data it is not cleaning this directory up before restore. It leads to a situation that it is effectively not possible to uninstall some python modules which were preinstalled by travis. Uninstall is needed if you want for example to "upgrade" an existing module.
Originally we were caching `$HOME/virtualenv/python*` directories. And on every single travis run it had to upgrade set of modules and push cache to S3. Off course it requires some time and some IO.

Also it seems that travis creates separate cache file per build job. We will cache etcd, consul or zookeeper binaries together with python virtualenv and save some time on downloading them.

In addition to that we wont cache python3.4 virtualenv for acceptance-tests builds, because we are running them only with 2.7 and 3.5.

At the end, despite on caching of etcd, consul and zookeeper binaries the total size of cache became smaller (200MB vs 220MB) and it should not be updated on every run because usually nothing is changing in there.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -17,32 +17,48 @@ env:
   - DCS="consul" TEST_SUITE="behave"
 cache:
   directories:
-  - $HOME/virtualenv/python2.7.9
-  - $HOME/virtualenv/python3.4.2
-  - $HOME/virtualenv/python3.5.2
+  - $HOME/mycache
+before_cache:
+- mv $HOME/virtualenv/python2.7.9 $HOME/mycache/python2.7.9
+- if [[ $TEST_SUITE != "behave" ]]; then mv $HOME/virtualenv/python3.4.2 $HOME/mycache/python3.4.2; fi
+- mv $HOME/virtualenv/python3.5.2 $HOME/mycache/python3.5.2
+- rm -f $HOME/mycache/python3.5.2/python3.5.2
 install:
   - |
     set -e
 
     if [[ $TEST_SUITE == "behave" ]]; then
         function get_consul() {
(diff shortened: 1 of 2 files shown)


LappleApple:
Hey @roidrage + @danishpkhan: Happy New Year. @CyberDem0n and I wanted to bring this issue onto your radar screen. We know of [this issue](https://github.com/travis-ci/travis-ci/issues/4873) reported a while back, and believe support of postgres 9.5 and 9.6 out of the box would delight many. Just a little ping on this topic. Maybe you have ideas on how we could help you help us. :)

.travis.yml
```diff
@@ -17,32 +17,48 @@ env:
...
   directories:
-  - $HOME/virtualenv/python2.7.9
-  - $HOME/virtualenv/python3.4.2
-  - $HOME/virtualenv/python3.5.2
+  - $HOME/mycache
+before_cache:
+- mv $HOME/virtualenv/python2.7.9 $HOME/mycache/python2.7.9
+- if [[ $TEST_SUITE != "behave" ]]; then mv $HOME/virtualenv/python3.4.2 $HOME/mycache/python3.4.2; fi
+- mv $HOME/virtualenv/python3.5.2 $HOME/mycache/python3.5.2
+- rm -f $HOME/mycache/python3.5.2/python3.5.2
```
CyberDem0n: There is a symlink to itself, and travis cache was complaining about it.

alexeyklyukin:
👍

CyberDem0n:
:+1:

PR #383, opened 2017-01-13
Bugfix/etcd machines

CyberDem0n:
Calling `_load_machines_cache()` method on timeout was causing switch to the new etcd node every 5 minutes...


alexeyklyukin:
👍

CyberDem0n:
:+1:

Issue #384, opened 2017-01-14
[v1.2] How to configure etcd for https endpoint?


drnic:
I'm enthusiastically trying out new support for etcd https endpoints (plus auth support).

But my output is filled with warnings from urllib3:

```
/usr/lib/python3.5/site-packages/urllib3/connectionpool.py:843: InsecureRequestWarning: Unverified HTTPS request is being made. Adding certificate verification is strongly advised. See: https://urllib3.readthedocs.io/en/latest/advanced-usage.html#ssl-warnings
InsecureRequestWarning)
```

I've tried installing `pip3 install certifi` but no change. 

I also tried specifying:

```yaml
etcd:
  cacert: /etc/ssl/certs/ca-certificates.crt
```

but this creates an error:

```
2017-01-14 16:07:43,702 INFO: Selected new etcd server https://54.174.22.61:15978
2017-01-14 16:07:43,886 ERROR: Failed to get list of machines from https://54.174.22.61:15978/v2: SSLError(SSLError(1, '[SSL: CERTIFICATE_VERIFY_FAILED] certificate verify failed (_ssl.c:645)'),)
2017-01-14 16:07:43,886 ERROR: Machines cache is empty, no machines to try.
2017-01-14 16:07:43,886 INFO: waiting on etcd
```

To reproduce the error, try using https://www.compose.com/etcd which provides 30-days free trial of an https etcd cluster with root/password. The configuration I'm using looks like:

```yaml
etcd:
  host: portal-nnn.stark-wayne.composedb.com:15000
  protocol: https
  username: root
  password: password
```

CyberDem0n:
Hi,

I think I know what is wrong...
1) Patroni gets a list of etcd client urls either from etcd server or from config file.
2) It resolves DNS names and connects to a resolved IP
3) Certificate is issued for a specific DNS and off course verification is failing, because https://localhost:2379 and https://127.0.0.1:2379 are not the same urls.

It make sense to resolve http://localhost:2379 into http://127.0.0.1:2379, but for https urls must stay unchanged.
I will prepare a fix on Monday.

drnic:
@CyberDem0n thanks xoxo

CyberDem0n:
@drnic Fix is merged into master. We will release a new minor version soon.
But there is one important thing regarding your configuration:

```yaml
etcd:
  cacert: /etc/ssl/certs/ca-certificates.crt
```
[...]

PR #385, opened 2017-01-16
Don't resolve url hostnames manualy but mokey patch urllib3

CyberDem0n:
Change hostnames by ip addresses was causing certificate verification to
fail. Instead of doing it we will better monkey patch urllib3
functionality which does name resolution. It should work without
problems even for https connection.

Should fix https://github.com/zalando/patroni/issues/384

### DIFF
--- a/patroni/dcs/etcd.py
+++ b/patroni/dcs/etcd.py
@@ -2,6 +2,7 @@ from __future__ import absolute_import
 import etcd
 import logging
 import os
+import urllib3.util.connection
 import random
 import requests
 import socket
@@ -16,7 +17,7 @@ from urllib3.exceptions import HTTPError, ReadTimeoutError
 from requests.exceptions import RequestException
 from six.moves.queue import Queue
 from six.moves.http_client import HTTPException
-from six.moves.urllib_parse import urlparse, urlunparse
+from six.moves.urllib_parse import urlparse
 from threading import Thread
 
 logger = logging.getLogger(__name__)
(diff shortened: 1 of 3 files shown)


patroni/dcs/etcd.py
```diff
@@ -264,8 +248,11 @@ def _get_machines_cache_from_srv(self, srv):
 
     def _get_machines_cache_from_dns(self, host, port):
         """One host might be resolved into multiple ip addresses. We will make list out of it"""
-        ret = ['{0}://{1}:{2}'.format(self.protocol, ip, port) for ip in set(self._dns_resolver.resolve(host))]
-        return list(set(ret)) if ret else ['{0}://{1}:{2}'.format(self.protocol, host, port)]
+        if self.protocol == 'http':
+            ret = ['{0}://{1}:{2}'.format(self.protocol, *r[4]) for r in self._dns_resolver.resolve(host, port)]
```
alexeyklyukin: According to https://docs.python.org/3/library/socket.html#socket.getaddrinfo and https://docs.python.org/2/library/socket.html#socket.getaddrinfo the Python tuple returned for the AF_INET6 contains 4 fields (host, port, flow_info and scope_id).

alexeyklyukin: Apparently, if the flowid is 0, you can ignore it, otherwise, one can prepend it with % to the ip address.

alexeyklyukin: Actually, is this method still correct? Wouldn't  _get_machines_cache_from_dns call result in preliminary resolution of hostnames to ip addresses and passing of ip addresses to connect?

CyberDem0n: This method will generate incorrect urls in case of IPV6 (ipv6 should be taken into square braces) :(

patroni/dcs/etcd.py
```diff
@@ -264,8 +248,16 @@ def _get_machines_cache_from_srv(self, srv):
...
     def _get_machines_cache_from_dns(self, host, port):
         """One host might be resolved into multiple ip addresses. We will make list out of it"""
-        ret = ['{0}://{1}:{2}'.format(self.protocol, ip, port) for ip in set(self._dns_resolver.resolve(host))]
-        return list(set(ret)) if ret else ['{0}://{1}:{2}'.format(self.protocol, host, port)]
+        if self.protocol == 'http':
+            ret = []
+            for af, _, _, _, sa in self._dns_resolver.resolve(host, port):
+                host, port = sa[:2]
+                if af == socket.AF_INET6:
+                    host = '[{0}]'.format(host)
```
alexeyklyukin: I'm not sure IPv6 scopes could be safely ignored. Look at this http://superuser.com/questions/99746/why-is-there-a-percent-sign-in-the-ipv6-address. To me it looks like the local IPv6 address may need a scope to be routed to the correct interface, and omitting it may break  clients connecting to link-local Etcds addresses.

CyberDem0n: underlying code performing http requests will do a second resolve anyway and therefore it should get everything it needs:

```python
In [1]: import socket

In [2]: socket.getaddrinfo('::1', 2379, 0, socket.SOCK_STREAM, socket.IPPROTO_TCP)
Out[2]: [(10, 1, 6, '', ('::1', 2379, 0, 0))]
```

alexeyklyukin:
👍

CyberDem0n:
:+1:

PR #387, opened 2017-01-25
Handle EtcdEventIndexCleared and EtcdWatcherCleared exceptions

CyberDem0n:
If this case it doesn't make sense to retry, because it brings nothing, but produces a lot of exceptions in the log...


alexeyklyukin:
👍

CyberDem0n:
:+1:

PR #395, opened 2017-02-08
 Reassemble postgresql parameters when major version became known

CyberDem0n:
Otherwise we were writing some "unknown" parameters into postgresql.conf
and postgres was refusing to start. Only 9.3 was affected.

In addition to that move rename of wal_level from hot_standby to replica
into get_server_parameters method. Now this rename is handled in a
single place.


alexeyklyukin:
👍

CyberDem0n:
:+1:

PR #396, opened 2017-02-13
Make it possible to compare xlog location against replica

CyberDem0n:
Otherwise it fails with the exception:
```python
2017-02-13 12:33:15,520 ERROR: could not determine difference with the master location
  File "/usr/local/lib/python3.4/dist-packages/patroni/scripts/wale_restore.py", line 137, in should_use_s3_to_create_replica
    (backup_start_lsn,))
psycopg2.OperationalError: recovery is in progress
HINT:  WAL control functions cannot be executed during recovery.
```


a1exsh:
:+1:

alexeyklyukin:
👍

PR #399, opened 2017-02-17
Update the startup script timeouts

jamessewell:
This change increases the Patroni shutdown timeout (after which Patroni is killed) from 10s to 30s.

It is needed when there is a sync replica in synchronous_mode to allow the replica time to be released by the master.

Without this change Patroni will be killed while waiting, and will leave PostgreSQL up rather than stopping it.

### DIFF
--- a/extras/startup-scripts/patroni.service
+++ b/extras/startup-scripts/patroni.service
@@ -21,7 +21,7 @@ ExecStart=/bin/patroni /etc/patroni.yml
 KillMode=process
 
 # Give a reasonable amount of time for the server to start up/shut down
-TimeoutSec=10
+TimeoutSec=30
 
 # Do not restart the service if it crashes, we want to manually inspect database on failure
 Restart=no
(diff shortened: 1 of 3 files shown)


CyberDem0n:
:+1:

alexeyklyukin:
As far as I understand the problem appears on a replica in a ``while_not_sync_standby`` when Patroni receives SIGTERM and waits for the replica to disappear from the sync object in DCS before stopping it. Normally, we wait for the 2 * ``loop_wait`` intervals, so when the ``loop_wait`` is set to 10s (default value) we might wait for at least 20s before bailing out and running stop anyway.

Actually, the value for the ``TimeoutSec`` should depend on the value of the Patroni
s ``loop_wait``. Theoretically, we cannot rely on ``/etc/patroni.yaml``, as one can change the value of ``loop_wait`` dynamically, but practically there are no compelling cases in doing so without reflecting the changes in the ``/etc/patroni.yaml`` as well. Therefore, I'd set this parameter from something like ``grep loop_wait /etc/patroni.yaml|awk '{print $2}'``

alexeyklyukin:
👍

PR #400, opened 2017-02-17
Synchronous mode strict

jamessewell:
This pull request adds a new DCS variable synchronous_mode_strict

If synchronous_mode_strict is true then when the last sync standby disappears in synchronous_mode then synchronous_standby_names is set to 'patroni_dummy_host'.

This prevents writes which can not be sync'ed to a standby - allowing a closer to zero data loss config.

### DIFF
--- a/patroni/config.py
+++ b/patroni/config.py
@@ -43,7 +43,6 @@ class Config(object):
         'maximum_lag_on_failover': 1048576,
         'master_start_timeout': 300,
         'synchronous_mode': False,
-        'synchronous_mode_strict': False,
         'postgresql': {
             'bin_dir': '',
             'use_slots': True,
@@ -177,8 +176,6 @@ class Config(object):
             elif name in config:  # only variables present in __DEFAULT_CONFIG allowed to be overriden from DCS
                 if name == 'synchronous_mode':
                     config[name] = value
-                if name == 'synchronous_mode_strict':
-                    config[name] = value
                 else:
                     config[name] = int(value)
         return config
@@ -302,7 +299,7 @@ class Config(object):
             config['name'] = pg_config['name']
 
         pg_config.update({p: config[p] for p in ('name', 'scope', 'retry_timeout',
-                          'synchronous_mode', 'synchronous_mode_strict', 'maximum_lag_on_failover') if p in config})
+                          'synchronous_mode', 'maximum_lag_on_failover') if p in config})
 
         return config
 
--- a/patroni/config.py
+++ b/patroni/config.py
@@ -43,6 +43,7 @@ class Config(object):
         'maximum_lag_on_failover': 1048576,
         'master_start_timeout': 300,
         'synchronous_mode': False,
+        'synchronous_mode_strict': False,
         'postgresql': {
             'bin_dir': '',
             'use_slots': True,
@@ -176,6 +177,8 @@ class Config(object):
             elif name in config:  # only variables present in __DEFAULT_CONFIG allowed to be overriden from DCS
                 if name == 'synchronous_mode':
                     config[name] = value
+                if name == 'synchronous_mode_strict':
+                    config[name] = value
                 else:
                     config[name] = int(value)
         return config
@@ -299,7 +302,7 @@ class Config(object):
             config['name'] = pg_config['name']
 
         pg_config.update({p: config[p] for p in ('name', 'scope', 'retry_timeout',
-                          'synchronous_mode', 'maximum_lag_on_failover') if p in config})
+                          'synchronous_mode', 'synchronous_mode_strict', 'maximum_lag_on_failover') if p in config})
 
         return config
 
--- a/patroni/config.py
+++ b/patroni/config.py
@@ -302,7 +302,7 @@ class Config(object):
             config['name'] = pg_config['name']
 
         pg_config.update({p: config[p] for p in ('name', 'scope', 'retry_timeout',
-                          'synchronous_mode', 'synchronous_mode_strict', 'maximum_lag_on_failover') if p in config})
+                          'synchronous_mode', 'maximum_lag_on_failover') if p in config})
 
         return config
 
--- a/patroni/ha.py
+++ b/patroni/ha.py
@@ -248,7 +248,7 @@ class Ha(object):
 
                 if self.is_synchronous_mode_strict() and picked is None:
                     picked = 'patroni_dummy_host'
-                    logger.info("No standbys available and synchronous_mode_strict set: blocking writes with dummy hostname"
+                    logger.info("No standbys available: blocking writes with dummy hostname")
 
                 logger.info("Assigning synchronous standby status to %s", picked)
                 self.state_handler.set_synchronous_standby(picked)
--- a/patroni/ha.py
+++ b/patroni/ha.py
@@ -223,9 +223,6 @@ class Ha(object):
     def is_synchronous_mode(self):
         return bool(self.cluster and self.cluster.config and self.cluster.config.data.get('synchronous_mode'))
 
-    def is_synchronous_mode_strict(self):
-        return bool(self.cluster and self.cluster.config and self.cluster.config.data.get('synchronous_mode_strict'))
-
     def process_sync_replication(self):
         """Process synchronous standby beahvior.
 
@@ -245,11 +242,6 @@ class Ha(object):
                     if not self.dcs.write_sync_state(self.state_handler.name, None, index=self.cluster.sync.index):
                         logger.info('Synchronous replication key updated by someone else.')
                         return
-
-                logger.info("Debug: %s", self.is_synchronous_mode_strict())
-                if self.is_synchronous_mode_strict() and picked is None:
-                    picked = 'patroni_dummy_host'
-
                 logger.info("Assigning synchronous standby status to %s", picked)
                 self.state_handler.set_synchronous_standby(picked)
 
--- a/patroni/ha.py
+++ b/patroni/ha.py
@@ -246,9 +246,9 @@ class Ha(object):
                         logger.info('Synchronous replication key updated by someone else.')
                         return
 
-                logger.info("Debug: %s", self.is_synchronous_mode_strict())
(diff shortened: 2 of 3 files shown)


patroni/ha.py
```diff
@@ -242,6 +245,11 @@ def process_sync_replication(self):
                     if not self.dcs.write_sync_state(self.state_handler.name, None, index=self.cluster.sync.index):
                         logger.info('Synchronous replication key updated by someone else.')
                         return
+
+                logger.info("Debug: %s", self.is_synchronous_mode_strict())
```
CyberDem0n: Log message could be more informative.

patroni/ha.py
```diff
@@ -242,6 +245,11 @@ def process_sync_replication(self):
...
+                if self.is_synchronous_mode_strict() and picked is None:
+                    picked = 'patroni_dummy_host'
```
CyberDem0n: I wonder, wouldn't it better to set it  to `'*'`?

patroni/config.py
```diff
@@ -299,7 +302,7 @@ def _build_effective_configuration(self, dynamic_configuration, local_configurat
             config['name'] = pg_config['name']
 
         pg_config.update({p: config[p] for p in ('name', 'scope', 'retry_timeout',
-                          'synchronous_mode', 'maximum_lag_on_failover') if p in config})
+                          'synchronous_mode', 'synchronous_mode_strict', 'maximum_lag_on_failover') if p in config})
```
CyberDem0n: `synchronous_mode_strict` is not used in the `Postgresql` class, therefore we don't need to propagate it.

patroni/ha.py
jamessewell: * would be different - replication to any active host (even though there are none at the moment).

I used 'patroni_dummy_host' to guarantee replication to no hosts. ' ' could be used - but that would be easy to miss when debugging.

At this time we have no hosts - but we also want to assert that if one does return we don't want to use it until such time as Patroni can do a forced setting (otherwise we negate the benefit of synchronous_mode and don't know where our sync writes are going).

patroni/config.py
jamessewell: Done.

patroni/ha.py
jamessewell: Done (this was a debug message I left in by mistake!)

patroni/ha.py
```diff
@@ -242,6 +245,11 @@ def process_sync_replication(self):
                     if not self.dcs.write_sync_state(self.state_handler.name, None, index=self.cluster.sync.index):
                         logger.info('Synchronous replication key updated by someone else.')
                         return
+
+                if self.is_synchronous_mode_strict() and picked is None:
+                    picked = 'patroni_dummy_host'
+                    logger.info("No standbys available and synchronous_mode_strict set: blocking writes with dummy hostname"
```
CyberDem0n: Tests are failing because:
 1. Line is longer than 120 characters
 2. Missing ')' at the end

jamessewell: Fixed

patroni/ha.py
```diff
@@ -242,6 +245,11 @@ def process_sync_replication(self):
                     if not self.dcs.write_sync_state(self.state_handler.name, None, index=self.cluster.sync.index):
                         logger.info('Synchronous replication key updated by someone else.')
                         return
+
+                if self.is_synchronous_mode_strict() and picked is None:
+                    picked = 'patroni_dummy_host'
```
CyberDem0n: I still think it would be better to have a `'*'` there and I will explain why.
First of all let's remember how `synchronous_mode` works:
 1. Patroni picks one of the streaming replicas (closest to the master)
 2. It sets this replica as synchronous in the `postgresql.conf` and runs `pg_ctl reload`
 3. Sleeps for a couple of seconds
 4. Verifies that chosen replica is really became synchronous
 5. Publish this replica into DCS

If the step 4 didn't succeeded it will continue with HA loop and return to the task of choosing and setting synchronous replica only on the next iteration. 

So Patroni doesn't have a goal to block writes on the master until synchronous replica is published in DCS.
During `loop_wait`seconds there is a chance that some other replica will connect, but it will have no chances to become synchronous until the next iteration...

If the `synchronous_standby_names` is set to `'*'`, replica will connect and become synchronous and Patroni will need to go only through steps 1, 2, 5.

jamessewell:
Hello,

Setting sync standby names to '*' means that any replica can be selected to
be the sync standby.

In the case where there were two available this would mean that Patroni
potentially no longer knows which standby is selected (as it hasn't been
enforced with an explicit setting yet) and therefore can't failover with
guaranteed zero data loss at that time.

This patch (for me anyway) is about trying to enforce zero data loss on
failure of the master. Using * would not achieve this.

Now I think about it though I also need to modify the startup code so it
always started up with the dummy host set as well - else writes could be
lost till a sync standby was selected.

Cheers,

CyberDem0n:
We have two different contracts here.
 1. Writes on master should be allowed only if we have a synchronous replica
 2. Patroni first should pick and set one replica as a synchronous and only after write this information to DCS

Patroni is responsible for the first contract only. It can't block writes on the master until information is exposed to DCS.

When we have some replicas streaming from the master - everything is all right, old algorithm works (Patroni should pick one of the replicas, update postgresql.conf and expose it's name to DCS)

Interesting things are happening when we have a corner case: if the synchronous mode is on and there are no replicas connected, Patroni removes name of old synchronous replica from DCS and removes `synchronous_standby_names` from postgresql.conf

For this corner case scenario (when strict mode is on) I propose to set `synchronous_standby_names`  to `'*'` instead of `dummy_host`.

It would be set to `'*'` only until some replica has connected. As soon it happened - Patroni will pick it up and change `synchronous_standby_names` from `'*'` to an actual replica name.

```python
                     if not self.dcs.write_sync_state(self.state_handler.name, None, index=self.cluster.sync.index):
                         logger.info('Synchronous replication key updated by someone else.')
                         return
+
+                if self.is_synchronous_mode_strict() and picked is None:
+                    logger.info("No standbys available: blocking writes")
+                    self.state_handler.set_synchronous_standby('*')
+                    return
```

Something like this...


> Now I think about it though I also need to modify the startup code so it always started up with the dummy host set as well - else writes could be lost till a sync standby was selected.

Indeed, https://github.com/zalando/patroni/blob/master/patroni/postgresql.py#L171 must be changed.
```python
        if config.get('synchronous_mode', False):
            if self._synchronous_standby_names is None:
+                if config.get('synchronous_mode_strict', False):
+                    parameters['synchronous_standby_names'] = '*'
+                else:
+                    parameters.pop('synchronous_standby_names', None)
-                 parameters.pop('synchronous_standby_names', None)
            else:
```

jamessewell:
Hmm.

What would happen in this situation:

1. All standbys die
2. Two standbys return
3. Writes happen
4. Master fails
5. Failover

When using * there is a potential for data loss, when using dummy I don't
think there is?

Cheers,
James Sewell

CyberDem0n:
> What would happen in this situation:
> 1. All standbys die
> 2. Two standbys return
> 3. Writes happen

That's fine. Write can always happen when in DCS there is no information about synchronous replica.
You can't prevent it. First we set synchronous_standby_names and only a few moments later update DCS. This operation can't be atomic.

> 4. Master fails
> 5. Failover 

Failover will not happen, only old leader is allowed to be promoted, because of DCS.

> When using * there is a potential for data loss, when using dummy I don't think there is?

Once again, `'*'`will be used only when we know that there are no replicas available.
At the same time in `sync` key in DCS will contain `{"leader": "leader-name"}`.
That means failover will not happen. Only the old master is allowed to become a leader.

Table of state transitions (hope it will help you to understand how it works):


connected replicas | synchronous_standby_names | DCS | comment
-------------------------|---------------------------------------|--------|--------------
**r2**,r3 | r2 | leader: r1, sync_standby: r2 | two replicas are connected, r2 synchronous and it is also declared as a synchronous in DCS, therefore on master failure it will take over
None | r2 | leader: r1, sync_standby: r2 | both replicas are dead, but in DCS r2 is still declared as synchronous. If the master will fail, only the old master or r2 are allowed to be promoted
None | r2 | leader: r1, sync_standby: None | Patroni removed sync_standby: r2 from DCS. From this moment on master failure only the old master can be promoted.
None | * | leader: r1, sync_standby: None | both replicas are dead. If the master will fail - nobody will take it over except the old master. But if the new replicas will come, one of them will become synchronous. 
r2,**r3** | * | leader: r1, sync_standby: None | two replicas are connected, r3 synchronous, but not in DCS. Writes are possible, but on master death nobody will take over.
r2,**r3** | r3 | leader: r1, sync_standby: None | two replicas are connected, r3 synchronous, but not in DCS. Writes are possible, but on master death nobody will take over.
r2,**r3** | r3 | leader: r1, sync_standby: r3 | two replicas are connected, r3 synchronous and it is also declared as a synchronous in DCS, therefore on master failure it will take over

jamessewell:
OK this table helps thanks - I see your point about failovers.

So in the case of

 two replicas, *, sync_standby:None

On master fail you'd need to go an evaluate both replicas manually to see
which had the most up to date data?

Cheers,
James

CyberDem0n:
> On master fail you'd need to go an evaluate both replicas manually to see which had the most up to date data?

Yes, you need to choose manually, but there is no guaranty that even most up to date replica will be the same as the old master.

And it doesn't depend on the value of `synchronous_standby_names`. Just think about situation when master and synchronous replica have died at the same time and only two asynchronous left. It's absolute equivalent.

alexeyklyukin:
The disadvantage of setting ``synchronous_standby_names`` to ``*``  is that it might result in a  ``nosync`` replica being chosen as a synchronous one. Looks like there is no easy way to prevent this. One idea to avoid this problem is to avoid changing ``synchronous_standby_names`` altogether when it is not possible to pick a new candidate (how we should set it during startup is another question).  We should, of course, keep the call that removes the ``sync_standby`` from the DCS in this case.

Since the whole point of the ``synchronous_mode`` is to failover only to the synchronous replicas, shouldn't we ignore candidates with ``nofailover`` flag set?

PR #403, opened 2017-03-03
Do not pick nofailover replicas as sync candidates.

alexeyklyukin:
Otherwise, we would have a sync replica we would not be able to
use for failover. Implements the suggestion at the end of the discussion at #400

### DIFF
--- a/patroni/postgresql.py
+++ b/patroni/postgresql.py
@@ -1296,7 +1296,7 @@ $$""".format(name, ' '.join(options)), name, password, password)
             member = members.get(app_name)
             if (state != 'streaming' or not member
                     or member.tags.get('nosync', False)
-                    or member.tags.get('nofailover', False)):
+                    or member.nofailover):
                 continue
             if sync_state == 'sync':
                 return app_name, True
--- a/patroni/postgresql.py
+++ b/patroni/postgresql.py
@@ -1294,7 +1294,9 @@ $$""".format(name, ' '.join(options)), name, password, password)
                      FROM pg_stat_replication
                     ORDER BY flush_location DESC"""):
             member = members.get(app_name)
-            if state != 'streaming' or not member or member.tags.get('nosync', False):
+            if (state != 'streaming' or not member
+                    or member.tags.get('nosync', False)
+                    or member.tags.get('nofailover', False)):
                 continue
             if sync_state == 'sync':
                 return app_name, True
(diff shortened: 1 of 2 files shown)


patroni/postgresql.py
```diff
@@ -1294,7 +1294,9 @@ def pick_synchronous_standby(self, cluster):
                      FROM pg_stat_replication
                     ORDER BY flush_location DESC"""):
             member = members.get(app_name)
-            if state != 'streaming' or not member or member.tags.get('nosync', False):
+            if (state != 'streaming' or not member
+                    or member.tags.get('nosync', False)
+                    or member.tags.get('nofailover', False)):
```
CyberDem0n: Just a nitpicking: `Member` object has `nofailover` attribute, so there is no need to get it from tags.
Line could be written as:
```python
if state != 'streaming' or not member or member.nofailover or member.tags.get('nosync', False):
```

Also I'm wondering, is it really necessary to do this? At the end we have a `nosync` tag, which does exactly what we want. Replica wont became a synchronous if `nosync` is set in `patroni.yaml`.

Also by introducing this additional check you are limiting some future functionality. Starting from 9.6 it's possible to have more than 1 synchronous standby (although it's not yet supported by Patroni). So in theory, by playing with nosync and nofailover flags it would be possible to have a cluster let's say with 2 synchronous replicas, but one of them will have a `nofailver` flag set, just because we want to have it always not only read-only, but also synchronous.

alexeyklyukin: > Also I'm wondering, is it really necessary to do this? At the end we have a nosync tag, which does exactly what we want. Replica wont became a synchronous if nosync is set in patroni.yaml.

On the other hand, in most cases people would want a synchronous replica to failover to it when the master fails and to avoid the data loss during the failover. What is the use case of running Patroni in a synchronous mode, which guarantees failing over only to the synchronous replica, and then allowing a replica that is not suitable for the failover as a synchronous one?

With multiple synchronous replicas there could be a valid use case of running a sync replica with ``nofailover``, for instance, using one of the two or more synchronous replicas for  real-time reporting. However, if we speculate how would we implement that with the ``synchronous_mode``, then the schema with a dedicated reporting sync replica would imply pinning that specific replica as a synchronous one, instead of letting  Patroni choose among the list of all replicas that don't have ``nosync`` tag set; if we would invent such a mechanism (i.e. as a dedicated ``always_sync`` tag), we would easily amend the logic proposed in this patch to allow such a replica to become synchronous even when the ``nofailover`` flag is set.

jamessewell:
Does this address the issue of a nofailover replica being picked as the sync replica (by PostgreSQL) when sync_standby_names = *?

PR #414, opened 2017-03-20
Use postgres:9.6 as a base image

CyberDem0n:
* Don't install zookeeper
* Use latest version of etcd
* Expose 5000 and 5001 on corresponding ports

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,33 +1,19 @@
 ## This Dockerfile is meant to aid in the building and debugging patroni whilst developing on your local machine
 ## It has all the necessary components to play/debug with a single node appliance, running etcd
-FROM ubuntu:16.04
+FROM postgres:9.6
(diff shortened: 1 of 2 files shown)


alexeyklyukin:
👍

a1exsh:
I wonder why did we have zookeeper package installed at all: only for the client part?

CyberDem0n:
Because https://github.com/zalando/patroni/blob/master/docker/entrypoint.sh supports zookeeper as well

a1exsh:
@CyberDem0n so now it doesn't? ;)  Honestly, I'm lost.  There are 4 changes in this PR one not related to the other in any way, or I just fail to see a relation.

CyberDem0n:
Zookeeper is not needed for testing/demoing stuff, but stripping it out from entrypoint.sh also doesn't make much sense.

CyberDem0n:
:+1:

PR #417, opened 2017-03-22
Smart pg_rewind

CyberDem0n:
Previously we were running pg_rewind only in limited amount of cases:
 * when we knew postgres was a master (no recovery.conf in data dir)
 * when we were doing a manual switchover to a specific node (no   guaranty that this node is the most up-to-date)
 * when a given node has nofailover tag (it could be ahead of new master)

This approach was kind of working in most of the cases, but sometimes we were executing pg_rewind when it was not necessary and in some other cases we were not executing it although it was needed.

The main idea of this PR is first try to figure out that we really need to run pg_rewind by analyzing timelineid, LSN and history file on master and replica and run it only if it's needed.


patroni/postgresql.py
```diff
@@ -292,6 +293,11 @@ def reload_config(self, config):
     def pending_restart(self):
         return self._pending_restart
 
+    @staticmethod
+    def configuration_allows_rewind(data):
+        return data.get('Current wal_log_hints setting', 'off') == 'on' \
+            or data.get('Data page checksum version', '0') != '0'
```
alexeyklyukin: We should also check that full_page_writes are enabled.

alexeyklyukin: Also, at the moment it checks only Postgres configuration. Is there any reason why doesn't it check Patroni configuration as well?

patroni/postgresql.py
```diff
@@ -931,139 +938,155 @@ def controldata(self):
...
-                        name = name.strip('-')
-                        result[name] = val
-        except IOError:
-            logger.exception('Error when reading postmaster.opts')
-        return result
+            with self._get_replication_connection_cursor(**leader.conn_kwargs()) as cur:
+                cur.execute('IDENTIFY_SYSTEM')
+                master_timeline = cur.fetchone()[1]
+                logger.error('master_timeline=%s', master_timeline)
+                if local_timeline > master_timeline:  # Not always supported by pg_rewind
```
alexeyklyukin: I think we should call get_timeline_lsn here as well, and pass the 'local' boolean flag instead of calling _get_local_timeline_lsn in order to check pg_controldata.

alexeyklyukin: I think we need to check our version, on 9.5 or below we know it is not supported, but how is it event possible when all nodes are managed by Patroni?

patroni/postgresql.py
CyberDem0n: In this regards there is nothing changed comparing to the old version.
We were checking only pg_controldata output.

patroni/postgresql.py
CyberDem0n: I was thinking about it, but: here it reuses the same connection to fetch history file.

CyberDem0n: Do we care?

On 9.5 and earlier pg_rewind will simply fail and that is fine, because afterwards either Patroni will remove data directory and reinitialize it or it will have _rewind_state = 'rewind_failed'.
We can think about exposing this state somewhere in some cases...

patroni/postgresql.py
alexeyklyukin: Yes, my first comment is exactly that we didn't check full_page_writes, but we should.
The second comment dealt with doing various configuration checks in different functions, the PostgreSQL's in configuration_allows_rewind and Patroni's in can_rewind. Wouldn't it make sense to put them all together?

patroni/postgresql.py
```diff
@@ -931,139 +938,152 @@ def controldata(self):
...
-                        name = name.strip('-')
-                        result[name] = val
-        except IOError:
-            logger.exception('Error when reading postmaster.opts')
-        return result
+            with self._get_replication_connection_cursor(**leader.conn_kwargs()) as cur:
+                cur.execute('IDENTIFY_SYSTEM')
+                master_timeline = cur.fetchone()[1]
+                logger.info('master_timeline=%s', master_timeline)
+                if local_timeline > master_timeline:  # Not always supported by pg_rewind
```
alexeyklyukin: Check PostgreSQL >= 9.6. How is it even possible with the cluster managed by Patroni (we are running the former master that must be more than 1 timeline ahead of the recently promoted one, which means it was at least 2 timelines ahead when running as a master before the recent promotion)

patroni/postgresql.py
CyberDem0n: > we didn't check full_page_writes, but we should.

How exactly do you want to check it? Current setting on master? And what if it was enabled just a few moments ago?

alexeyklyukin:
👍

CyberDem0n:
:+1:

PR #418, opened 2017-03-24
Retry when tagging EC2 and EBS with Postgres roles.

alexeyklyukin:
Retry if an error happens when setting Role or Name tags for the
EC2 instances or EBS volumes. The maximum retry interval is 15 seconds.


avaczi:
:+1:

alexeyklyukin:
👍

PR #420, opened 2017-03-28
Release the leader key when the leader restarts with an empty data dir

### LINKED ISSUE #419: No election after loss of leader
I am trying to get patroni setup in Kubernetes (well minikube right now) by [adapting](https://github.com/abkfenris/atomicdb) @jberkus [atomicbd](https://github.com/jberkus/atomicdb/tree/master/patroni_petset).

When I kill the leader pod it will restart fast enough that it reacquires the leader key so their is no failover. It then will try to bootstrap from itself.

```sh
2017-03-24 13:11:11,066 INFO: Selected new etcd server http://172.17.0.6:2379
2017-03-24 13:11:11,066 INFO: Starting new HTTP connection (1): 172.17.0.6
2017-03-24 13:11:11,087 INFO: trying to bootstrap from leader 'patroni_0'
2017-03-24 13:11:11,089 INFO: Lock owner: patroni_0; I am patroni_0
2017-03-24 13:11:11,096 INFO: updated leader lock during bootstrap from leader 'patroni_0'
pg_basebackup: could not connect to server: could not translate host name "patroni-0.patroni" to address: Name or service not known
2017-03-24 13:11:11,129 ERROR: Error when fetching backup: pg_basebackup exited with code=1
2017-03-24 13:11:11,130 WARNING: Trying again in 5 seconds
2017-03-24 13:11:16,090 INFO: Lock owner: patroni_0; I am patroni_0
2017-03-24 13:11:16,097 INFO: updated leader lock during bootstrap from leader 'patroni_0'
pg_basebackup: could not connect to server: could not connect to server: Connection refused
	Is the server running on host "patroni-0.patroni" (172.17.0.7) and accepting
	TCP/IP connections on port 5432?
2017-03-24 13:11:16,146 ERROR: Error when fetching backup: pg_basebackup exited with code=1
2017-03-24 13:11:16,146 ERROR: failed to bootstrap from leader 'patroni_0'
```

The other two pods don't realize that the leader is trying to bootstrap and try to keep following it rather than having an election.

```sh
[patroni-2] 2017-03-24 13:23:56,100 INFO: Lock owner: patroni_0; I am patroni_2
[patroni-2] 2017-03-24 13:23:56,100 INFO: does not have lock
[patroni-1] 2017-03-24 13:23:56,097 INFO: Lock owner: patroni_0; I am patroni_1
[patroni-1] 2017-03-24 13:23:56,098 INFO: does not have lock
[patroni-1] 2017-03-24 13:23:56,108 INFO: no action.  i am a secondary and i am following a leader
[patroni-2] 2017-03-24 13:23:56,110 INFO: no action.  i am a secondary and i am following a leader
[patroni-2] FATAL:  could not connect to the primary server: could not connect to server: Connection refused
[patroni-2] Is the server running on host "patroni-0.patroni" (172.17.0.7) and accepting
[patroni-2] TCP/IP connections on port 5432?
[patroni-2]
[patroni-1] FATAL:  could not connect to the primary server: could not connect to server: Connection refused
[patroni-1] Is the server running on host "patroni-0.patroni" (172.17.0.7) and accepting
[patroni-1] TCP/IP connections on port 5432?
```

As far as I can tell, there is a method ([ha.release_leader_key_voluntarily](https://github.com/zalando/patroni/blob/8fda95780451029cee6b9dc3fdcf9f71ab4480a6/patroni/ha.py#L474)) for the leader to give up the key but it doesn't look like there is a check if a newly created instance is the leader and to release the key.

abkfenris:
On Kubernetes (and maybe other platforms) if the leader instance is killed, it may restart fast enough that the ttl never expires, and an election isn't triggered. 

If the instances are using ephemeral storage, that means the instance comes back, sees that it is the leader, but it has an empty data directory, so it tries to bootstrap off itself. The cluster then gets stuck in a read only state as replicas are available, but the leader isn't able to accept writes as it is still trying to bootstrap.

I've added a check to see if the data directory is empty and if the instance is the leader before it bootstraps. If those a are both true, then it releases the leader key voluntarily so that an election can occur between the healthy replicas.

Closes #419

### DIFF
--- a/patroni/ha.py
+++ b/patroni/ha.py
@@ -883,10 +883,11 @@ class Ha(object):
                 msg = self.post_recover()
                 if msg is not None:
                     return msg
-            
+
             # is the data directory empty and we are the leader?
             if self.state_handler.data_directory_empty() and self.has_lock():
                 self.release_leader_key_voluntarily()
+                return 'released leader key voluntarily as data dir empty and currently leader'
 
             # is data directory empty?
             if self.state_handler.data_directory_empty():
--- a/patroni/ha.py
+++ b/patroni/ha.py
@@ -884,13 +884,13 @@ class Ha(object):
                 if msg is not None:
                     return msg
 
-            # is the data directory empty and we are the leader?
-            if self.state_handler.data_directory_empty() and self.has_lock():
-                self.release_leader_key_voluntarily()
-                return 'released leader key voluntarily as data dir empty and currently leader'
-
             # is data directory empty?
             if self.state_handler.data_directory_empty():
+                # is this instance the leader?
+                if self.has_lock():
+                    self.release_leader_key_voluntarily()
+                    return 'released leader key voluntarily as data dir empty and currently leader'
+
                 return self.bootstrap()  # new node
             # "bootstrap", but data directory is not empty
             elif not self.sysid_valid(self.cluster.initialize) and self.cluster.is_unlocked() and not self.is_paused():
--- a/patroni/ha.py
+++ b/patroni/ha.py
@@ -883,6 +883,10 @@ class Ha(object):
                 msg = self.post_recover()
                 if msg is not None:
                     return msg
+            
+            # is the data directory empty and we are the leader?
+            if self.state_handler.data_directory_empty() and self.has_lock():
+                self.release_leader_key_voluntarily()
 
             # is data directory empty?
             if self.state_handler.data_directory_empty():
--- a/tests/test_ha.py
+++ b/tests/test_ha.py
@@ -771,3 +771,14 @@ class TestHa(unittest.TestCase):
 
     def test_wakup(self):
         self.ha.wakeup()
+
+    def test_leader_with_empty_directory(self):
+        self.ha.cluster = get_cluster_initialized_with_leader()
+        self.ha.has_lock = true
+        self.p.data_directory_empty = true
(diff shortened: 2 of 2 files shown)


patroni/ha.py
```diff
@@ -884,6 +884,11 @@ def _run_cycle(self):
                 if msg is not None:
                     return msg
 
+            # is the data directory empty and we are the leader?
+            if self.state_handler.data_directory_empty() and self.has_lock():
```
CyberDem0n: Why not to move this block no the line 889? It will eliminate double call of `self.state_handler.data_directory_empty()`
Everything else looks good to me.

abkfenris: Do you mean nest an `if self.has_lock()` within 
```python
# is data directory empty?
if self.state_handler.data_directory_empty():
    return self.bootstrap()  # new node
```
instead?

So:
```python
# is data directory empty?
if self.state_handler.data_directory_empty():
    # is this instance the leader?
    if self.has_lock():
        return 'released leader key voluntarily as data dir empty and currently leader'

    return self.bootstrap()  # new node
```

CyberDem0n: Exactly.

CyberDem0n:
:+1:

alexeyklyukin:
👍

CyberDem0n:
Merged, thank you @abkfenris

Issue #425, opened 2017-04-10
Sometimes Patroni fails to drop replication slot and it causes xlog to fill up


CyberDem0n:
We already observed it several times on production.
I think it happens because we are removing only non-active replication slots: https://github.com/zalando/patroni/blob/master/patroni/postgresql.py#L1184

Default value of [wal_sender_timeout](https://www.postgresql.org/docs/9.6/static/runtime-config-replication.html) is 60 seconds while we are trying to remove replication slot after 30 seconds (ttl=30). That means postgres might think that slot is still active and does nothing. Patroni at the same time thinks that it was removed successfully.

CyberDem0n:
Closed by https://github.com/zalando/patroni/issues/427

PR #427, opened 2017-04-12
BUGFIX: pg_drop_replication_slot may not be called if slot is active

CyberDem0n:
Default value of wal_sender_timeout is 60 seconds while we are trying to remove replication slot after 30 seconds (ttl=30). That means postgres might think that slot is still active and does nothing. Patroni at the
same time was thinking that it was removed successfully.

If the drop replication slot query didn't return any single row we must fetch list of existing physical replication slots from postgres on the next iteration of HA loop.

Fixes: issue #425


alexeyklyukin:
👍

CyberDem0n:
:+1:

PR #428, opened 2017-04-12
Add config editing to patronictl

ants:
Current UI to change cluster configuration is somewhat unfriendly, involving a curl command, knowing the REST API endpoint, knowing the specific syntax to call it with and writing a JSON document. I added two commands in this branch to make this a bit easier, `show-config` and `edit-config` (names are merely placeholders, any opinions on better ones?).

* `patronictl show-config clustername` fetches the config from DCS, formats it as YAML and outputs it.

* `patronictl edit-config clustername` fetches the config, formats it as YAML, invokes $EDITOR on it, then shows user the diff and after confirmation applies the changed config to DCS, guarding for concurrent modifications.

* `patronictl edit-config clustername --set synchronous_mode=true --set postgresql.use_slots=true` will set the specific key-value pairs.

There are also some UI capabilities I'm less sure of, but included them here as I already implemented them.

* If output is a tty then the diffs are colored. I'm not sure if this feature is cool enough to pull the weight of adding a dependency on cdiff. Or maybe someone knows of another more task focused diff coloring library?
* `patronictl edit-config clustername --pg work_mem=100MB` - Shorthand for `--set postgresql.parameters.work_mem=100MB`
* `patronictl edit-config clustername --apply changes.yaml` - apply changes from a yaml file.
* `patronictl edit-config clustername --replace new-config.yaml` - replace config with new version.


CyberDem0n:
Looks really cool!
Will  try it tomorrow.

ants:
I wonder if we ever want to support spelling GUC's with dots in them as nested parameters:

    postgresql:
      parameters:
        auto_explain:
          log_min_duration: 1s
          log_analyze: on
          log_buffers: on
          log_timing: off

vs. 

    postgresql:
      parameters:
        auto_explain.log_min_duration: 1s
        auto_explain.log_analyze: on
        auto_explain.log_buffers: on
        auto_explain.log_timing: off

Nested variant looks pretty neat, but the current way is conceptually much simpler. Having nesting will  require some thinking about how to reconcile the two different spellings and will result in at least one of the cases being magically converted to the other. Any opinions one way or the other?

CyberDem0n:
Agree, nested variant looks nice, but what about following corner-case?
```
localhost/postgres=# set work_mem.bla to 'foo';
SET
localhost/postgres=# show work_mem.bla;
 work_mem.bla 
──────────────
 foo
(1 row)
```

With nested variant it would be not possible to set a custom `work_mem` along with `work_mem.bla`.

ants:
I haven't seen any examples of this kind of overlap in the wild, but in theory it's possible. We could invent hacks to get around this, either a magically named key in the nested structure for the scalar value, or fall back to dotted representation when the prefix is already taken by a scalar value. I kind of like the latter variant. The complexity will all be compartmentalized to the UI side, the conversion done when converting to/from JSON. When loading postgresql.parameters we check for names with dots and if they exist and the prefix key does not, we convert it to a dict. When storing back it's just a matter of flattening any dicts out to dotted representation.

    # prettify({"work_mem.bla": "foo"})
    work_mem:
      bla: foo

    # prettify({"work_mem.bla": "foo", "work_mem": "4MB"})
    work_mem: 4MB
    work_mem.bla: foo

CyberDem0n:
I think it's better to be explicit rather than implicit.
Whoever is already using GUC's with dots would be very surprised/confused if we convert them into nested structures.

@alexeyklyukin, what do you think?

alexeyklyukin:
I prefer a simple version with a dot: it's the same syntax as the database uses, and I don't see a reason to invent a new one on the Patroni side.

ants:
I'll do the simple one then.

ants:
Couple more things I'd like to see a second opinion on.

1. Having the command line interpret arguments as YAML values is a potential footgun. `edit-command --pg archive_command=true` will actually set it to `True`. This can be solved by encapsulating the value with quotes, but it could be surprising. I would just store the string, but for values outside `postgresql.parameters` we actually need to get the type correct, otherwise Patroni will start giving nasty exceptions. I think the whole thing would be best solved by making the config system know the types of the datastructure and do some validation. But I think that's something for a separate patch. Maybe just make a special case for postgresql.parameters and handle the rest later?
2. Any opinions for the naming of the commands? Current names are only placeholders. One idea  would be to merge the two commands and have the default just display the config and a `--edit` option to invoke the editor. And I don't have a good idea how to name it in a way that it's clear that `patronictl configuration` is for the patronictl command, but the new thing is for a Patroni config stored in the DCS.

alexeyklyukin:
👍

CyberDem0n:
:+1:

CyberDem0n:
Merged, thank you!

PR #429, opened 2017-04-12
Remove error spinning on etcd failure and reduce log spam

ants:
When all etcd servers refuse connections during watch the call will fail with an exception and will be immediately retried. This creates a huge amount of log spam potentially creating additional issues on top of losing the DCS. This patch takes note if etcd failures are repeating and starting from the second failure will sleep for a second before retrying. It additionally omits the stack trace after the first failure in a streak of failures.

### DIFF
--- a/patroni/dcs/etcd.py
+++ b/patroni/dcs/etcd.py
@@ -299,9 +299,9 @@ class Client(etcd.Client):
 def catch_etcd_errors(func):
     def wrapper(self, *args, **kwargs):
         try:
-            return func(self, *args, **kwargs) is not None
+            retval = func(self, *args, **kwargs) is not None
         except (RetryFailedError, etcd.EtcdException):
-            return False
+            retval = False
         except Exception as e:
             if not self._has_failed:
                 logger.exception("")
@@ -309,8 +309,9 @@ def catch_etcd_errors(func):
                 logger.error(e)
             self._has_failed = True
             raise EtcdError("unexpected error")
-        else:
-            self._has_failed = False
+
+        self._has_failed = False
+        return retval
 
     return wrapper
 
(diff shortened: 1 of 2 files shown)


patroni/dcs/etcd.py
```diff
@@ -297,14 +297,20 @@ def _load_machines_cache(self):
...
-            logger.exception("")
+        except Exception as e:
+            if not self._has_failed:
+                logger.exception("")
+            else:
+                logger.error(e)
+            self._has_failed = True
             raise EtcdError("unexpected error")
+        else:
+            self._has_failed = False
```
CyberDem0n: We return from try block, therefore this branch will never be executed.

patroni/dcs/etcd.py
```diff
@@ -297,14 +297,20 @@ def _load_machines_cache(self):
 
 
 def catch_etcd_errors(func):
-    def wrapper(*args, **kwargs):
+    def wrapper(self, *args, **kwargs):
         try:
-            return func(*args, **kwargs) is not None
+            return func(self, *args, **kwargs) is not None
         except (RetryFailedError, etcd.EtcdException):
             return False
```
CyberDem0n: Shouldn't be here the same logic (setting `_has_failed` to `True`) as for all other cases?

patroni/dcs/etcd.py
ants: Yeah, try-else stubbornly keeps working in differently from what I need of it. Fixed this one.

patroni/dcs/etcd.py
ants: Not sure. This branch doesn't log the error, we might want to still see the stack trace. A better question would be, should this log errors? I think it should, will take a look what the consequences are.

ants:
I was wondering the same thing about the similarity, the logic felt quite duplicative indeed. I will take a look at factoring the code out tomorrow.

CyberDem0n:
:+1:

alexeyklyukin:
👍

PR #433, opened 2017-04-18
Fix race condition when opening connection to cluster

CyberDem0n:
`Postgresql.connection` method could be called from different threads at the same time resulting in more than one connection open but only one used afterwards.


alexeyklyukin:
👍

alexeyklyukin:
Do we possibly have the same issue with the _cursor() method? IIRC the DB query code calls the cursor as well if there is no cursor, and since it may be called concurrently, the first invocation result can be overwritten by a subsequent ones.

CyberDem0n:
_cursor doesn't have this problem, it is called only from main thread

CyberDem0n:
:+1:

PR #435, opened 2017-04-20
Synchronous mode strict

CyberDem0n:
If synchronous_mode_strict is enabled and there are no live standbys connected, Patroni will set synchronous_standby_names to '*' to disable writes on master.


alexeyklyukin:
👍

CyberDem0n:
:+1:

PR #437, opened 2017-04-25
Export locale variables when forking a postmaster.

alexeyklyukin:
Avoid "postmaster became multithreaded during startup" error on OS X
built with --enable-nls (default for petere/homebrew).

The issue is that on OS X the libintl that replaces setlocale() spawns
a thread when it needs to detect the current locale. Setting the LC_
variables prevents this, but we don't propagate them to the fork.

See:
https://www.postgresql.org/message-id/flat/20140902013458.GB906981%40tornado.leadboat.com#20140902013458.GB906981@tornado.leadboat.com
for an additional explanation of the original Postgres behavior.


a1exsh:
:+1:

patroni/postgresql.py
```diff
@@ -717,7 +717,9 @@ def start(self, timeout=None, block_callbacks=False):
         # In order to make everything portable we can't use fork&exec approach here, so  we will call
         # ourselves and pass list of arguments which must be used to start postgres.
         proc = call_self(['pg_ctl_start', self._pgcommand('postgres'), '-D', self._data_dir] + options, close_fds=True,
-                         preexec_fn=os.setsid, stdout=subprocess.PIPE, env={'PATH': os.environ.get('PATH')})
+                         preexec_fn=os.setsid, stdout=subprocess.PIPE,
+                         env={'PATH': os.environ.get('PATH'), 'LC_ALL': os.environ.get('LC_ALL'),
```
CyberDem0n: What if environment not defined?

And in order not to repeat the same consts multiple times I think it make sense to rewrite it like:
```python
env={p: os.environ[p] for p in ('PATH', 'LC_ALL', 'LANG') if p in os.environ}
```

alexeyklyukin: Well, if it's not defined, then .get will get you None, nothing wrong here. I think your suggestion makes sense, on the second look I'm wondering why do we clear the environment altogether?

CyberDem0n:
:+1:

alexeyklyukin:
👍

PR #444, opened 2017-05-19
Implement postgresql-10 support

CyberDem0n:
Mainly it handles rename of xlog to wal.
In the API and inside DCS it is still named xlog (for compatibility).

### DIFF
--- a/patroni/postgresql.py
+++ b/patroni/postgresql.py
@@ -63,20 +63,20 @@ class Postgresql(object):
     #    check_function -- if the new value is not correct must return `!False`
     #    min_version -- major version of PostgreSQL when parameter was introduced
     CMDLINE_OPTIONS = {
-        'listen_addresses': (None, lambda _: False, 9.1),
-        'port': (None, lambda _: False, 9.1),
-        'cluster_name': (None, lambda _: False, 9.5),
-        'wal_level': ('hot_standby', lambda v: v.lower() in ('hot_standby', 'replica', 'logical'), 9.1),
-        'hot_standby': ('on', lambda _: False, 9.1),
-        'max_connections': (100, lambda v: int(v) >= 100, 9.1),
-        'max_wal_senders': (5, lambda v: int(v) >= 5, 9.1),
-        'wal_keep_segments': (8, lambda v: int(v) >= 8, 9.1),
-        'max_prepared_transactions': (0, lambda v: int(v) >= 0, 9.1),
-        'max_locks_per_transaction': (64, lambda v: int(v) >= 64, 9.1),
-        'track_commit_timestamp': ('off', lambda v: parse_bool(v) is not None, 9.5),
-        'max_replication_slots': (5, lambda v: int(v) >= 5, 9.4),
-        'max_worker_processes': (8, lambda v: int(v) >= 8, 9.4),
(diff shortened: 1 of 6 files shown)


patroni/postgresql.py
```diff
@@ -151,23 +151,35 @@ def _configuration_to_save(self):
...
-        return self._use_slots and self._major_version >= 9.4
+        return self._use_slots and self._major_version >= 90400
 
     @property
     def callback(self):
         return self.config.get('callbacks') or {}
 
+    @staticmethod
+    def _wal_name(version):
+        return version >= 100000 and 'wal' or 'xlog'
```
a1exsh: I think the more widely accepted python way to do this is with: `'xxx' if condition else 'yyy'`

patroni/postgresql.py
```diff
@@ -181,10 +193,10 @@ def get_server_parameters(self, config):
                     parameters.pop('synchronous_standby_names', None)
             else:
                 parameters['synchronous_standby_names'] = self._synchronous_standby_names
-        if self._major_version >= 9.6 and parameters['wal_level'] == 'hot_standby':
+        if self._major_version >= 90600 and parameters['wal_level'] == 'hot_standby':
             parameters['wal_level'] = 'replica'
         return {k: v for k, v in parameters.items() if not self._major_version or
-                self._major_version >= self.CMDLINE_OPTIONS.get(k, (0, 1, 9.1))[2]}
+                self._major_version >= self.CMDLINE_OPTIONS.get(k, (0, 1, 90400))[2]}
```
a1exsh: Is the change from 9.1 to 9.4 intentional here?

patroni/postgresql.py
```diff
@@ -63,20 +63,20 @@ class Postgresql(object):
...
-        'max_connections': (100, lambda v: int(v) >= 100, 9.1),
-        'max_wal_senders': (5, lambda v: int(v) >= 5, 9.1),
-        'wal_keep_segments': (8, lambda v: int(v) >= 8, 9.1),
-        'max_prepared_transactions': (0, lambda v: int(v) >= 0, 9.1),
-        'max_locks_per_transaction': (64, lambda v: int(v) >= 64, 9.1),
-        'track_commit_timestamp': ('off', lambda v: parse_bool(v) is not None, 9.5),
-        'max_replication_slots': (5, lambda v: int(v) >= 5, 9.4),
-        'max_worker_processes': (8, lambda v: int(v) >= 8, 9.4),
-        'wal_log_hints': ('on', lambda _: False, 9.4)
+        'listen_addresses': (None, lambda _: False, 90100),
```
a1exsh: Do we really have to change version encoding from float to int here to support v10?

patroni/postgresql.py
Jan-M: Yes, please change. Interesting though that this returns a String

patroni/postgresql.py
CyberDem0n: Yes, inversion is more consistent :)

Jan-M:
:+1:

alexeyklyukin:
👍

CyberDem0n:
:+1:

a1exsh:
:+1:

PR #449, opened 2017-06-02
Watchdog unit-tests, bugfixes and questions

CyberDem0n:
@ants,

I've looked closer on watchdog implementation when trying to increase line coverage and found that some branches can never be reached.  Marked them with with `XXX:`.
Also I've found `return WatchdogError` instead of `raise WatchdogError` in one place.

### DIFF
--- a/features/environment.py
+++ b/features/environment.py
@@ -81,6 +81,7 @@ class AbstractController(object):
     def cancel_background(self):
         pass
 
+
 class PatroniController(AbstractController):
     __PORT = 5440
     PATRONI_CONFIG = '{}.yml'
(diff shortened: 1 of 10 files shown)


ants:
Thanks for the work on tests. The code where you found the non-reachable parts could really use a good beating with the "you ain't gonna need it" simplification stick. I ran out of time today, but I'll take shot at this on Wednesday. Will post my findings on simplifying the amount of keepalives in a second WIP pull request.

alexeyklyukin:
👍

CyberDem0n:
:+1:

PR #450, opened 2017-06-07
BUGFIX: retry on boto exceptions never worked

CyberDem0n:
because `boto.exception` is not an excpetion, but a python module.

+ increase retry timeout to 5 minutes
+ refactor unit-tests to cover the case with retries.


alexeyklyukin:
👍

CyberDem0n:
:+1:

PR #452, opened 2017-06-12
Simplify watchdog code

ants:
I went through all the places that keepalives got sent with an eye for when it would be safe to postpone them. Turns out that having keepalives activated while we are not actively a master is really expensive with regard to complexity. So I went ahead and converted it to a model where we only activate the watchdog while master and not paused. This let me get rid of all kinds of nasty stuff.

I did notice a nasty thing however. When etcd is non-responsive we wait for retry_timeout before giving up. This means that after the last keepalive there is potentially loop_wait+retry_timeout before we know to start demotion. With default settings this is 10+10 = 20s, 5s after we would have been rebooted by the watchdog. Obvious fix would be to change the settings, but I thing the current scheme for keepalive timeout is overly pessimistic. The current keepalive timeout is calculated based on the assumption that it's possible that:

1. update_lock() could be called in quick succession with a trivially small delay.
2. the second call could have up to keepalive timeout seconds elapse between the execution of these two lines:

    ```python
        self.dcs.update_leader()
        self.watchdog.keepalive()
    ```

This is possible in theory but exceedingly unlikely. I'd like to express this as a watchdog configuration parameter that relaxes this. Not sure how it should look like, but I would feel quite fine if instead of `ttl // 2` we use what amounts more or less to `ttl - 5`. This way there would still be a few seconds to demote postgres if etcd is completely unresponsive.

This is still WIP. Besides the timeout calculation I have yet to actually test this code. And I still have to implement the dynamic ttl/loop_wait support. I can work on this in a couple of days.


ants:
Still working on this.

TODO list:
* Investigate if NullWatchdog is an useful coding pattern
* Merge and revise unit and feature tests
* Implement a `safety_margin` config parameter to tune watchdog safety.
* Document the issue and/or emit a warning if loop_wait + retry_timeout > watchdog timeout.

ants:
Should be relatively good now. Tried removing NullWatchdog, but the result looked more complicated to me.

ants:
Behave tests are failing because Patroni doesn't immediately trip up on a hung postmaster, previously were accidentally succeeding because loop_wait + retry_timeout > ttl / 2. Bumping up the timeouts in the test works, but is not exactly elegant. SIGSTOP to postmaster is not the ideal failure case here. I will think about if we can actually recreate realistic test scenarios here.

patroni/ha.py
```diff
@@ -954,6 +900,7 @@ def post_bootstrap(self):
         self._bootstrapping = False
         self.dcs.set_config_value(json.dumps(self.patroni.config.dynamic_configuration, separators=(',', ':')))
         self.dcs.take_leader()
+        self.watchdog.activate()
```
ants: Good point, needs to be checked.

patroni/ha.py
```diff
@@ -393,6 +341,15 @@ def while_not_sync_standby(self, func):
...
     def enforce_master_role(self, message, promote_message):
+        if not self.watchdog.is_running:
+            if not self.watchdog.activate():
+                if self.state_handler.is_leader():
+                    self.demote('immediate')
+                    return 'Demoting self because watchdog could not be activated'
+                else:
+                    self.release_leader_key_voluntarily()
+                    return 'Not promoting self because watchdog could not be actived'
+
```
ants: Shouldn't be a reason to do it, watchdog device open counts as a keepalive.

patroni/ha.py
```diff
@@ -393,6 +341,15 @@ def while_not_sync_standby(self, func):
                 self._disable_sync -= 1
 
     def enforce_master_role(self, message, promote_message):
+        if not self.watchdog.is_running:
+            if not self.watchdog.activate():
+                if self.state_handler.is_leader():
+                    self.demote('immediate')
+                    return 'Demoting self because watchdog could not be activated'
+                else:
+                    self.release_leader_key_voluntarily()
```
ants: demote('immediate') will do the release.

CyberDem0n:
:+1:

alexeyklyukin:
👍

PR #454, opened 2017-06-15
Custom bootstrap

CyberDem0n:
Task of restoring a cluster from backup or cloning existing cluster into a new one was floating around for some time. It was kind of possible to achieve it by doing a lot of manual actions and very error prone. So I come up with the idea of making the way how we bootstrap a new cluster configurable.

In short - we want to run a custom script instead of running initdb.


features/backup_create.sh
```diff
@@ -0,0 +1,30 @@
+#!/bin/bash
+
+PARSED=$(getopt --options v --longoptions datadir:,dbname:,walmethod: --name "$0" -- "$@" 2> /dev/null)
```
alexeyklyukin: This syntax is for GNU getopt, BSD getopt is quite limited compared to the GNU one: https://www.freebsd.org/cgi/man.cgi?getopt(3) (looks like it only supports short one-letter options).

features/environment.py
```diff
@@ -528,6 +540,53 @@ def create_and_set_output_directory(self, feature_name):
...
         self._output_dir = feature_dir
 
+    def clone(self, from_name, cluster_name, to_name):
+        f = self._processes[from_name]
+        custom_config = {
+            'scope': cluster_name,
+            'bootstrap': {
+                'method': 'pg_basebackup',
+                'pg_basebackup': {
+                    'command': self.BACKUP_SCRIPT + ' --walmethod=stream --dbname=' + f.backup_source
```
alexeyklyukin: How does the cloned node finds its data directory? It is started with PGDATA pointing to the backup destination (since there is no --datadir parameter to this script, perhaps, the PGDATA of the clone is used as a destination?)

features/environment.py
```diff
@@ -276,6 +276,15 @@ def terminate_backends(self):
             if 'process' not in p.cmdline()[0]:
                 p.terminate()
 
+    @property
+    def backup_source(self):
+        return 'postgres://{username}:{password}@{host}:{port}/{database}'.format(**self._replication)
+
+    def backup(self, dest='basebackup'):
```
alexeyklyukin: declare 'basebackup' as a "constant" (as it is also used in bootstrap_from_backup).

features/environment.py
CyberDem0n: `--datadir` argument comes from Patroni: https://github.com/zalando/patroni/pull/454/files#diff-d8c1582337c8426ef750420dc6b76949R529

features/environment.py
CyberDem0n: I've been thinking about it, but unfortunately it will require do define a global variable.

patroni/ha.py
```diff
@@ -204,22 +206,11 @@ def bootstrap(self):
...
-                        self.state_handler.stop('immediate')
-                        self.state_handler.move_data_directory()
-                        raise
-                    self.dcs.set_config_value(json.dumps(self.patroni.config.dynamic_configuration,
-                                                         separators=(',', ':')))
-                    self.dcs.take_leader()
-                    self.load_cluster_from_dcs()
-                    return 'initialized a new cluster'
+                self._bootstraping = True
+                self._post_bootstrap_task = CriticalTask()
```
alexeyklyukin: Looks like _post_bootstrap_task is unused.

patroni/postgresql.py
```diff
@@ -521,6 +525,26 @@ def _initialize(self, config):
...
+            logger.info('Running custom bootstrap script: %s', config['command'])
+            if subprocess.call(shlex.split(config['command']) + params) != 0:
+                self.set_state('custom bootstrap failed')
+                return False
+        except Exception:
+            logger.exception('Exception during custom bootstrap')
+            return False
+
+        self._post_restore()
+        self.save_configuration_files()
```
alexeyklyukin: _post_restore will can restore_configration_files()? Why do we need to save them right afterwards?

CyberDem0n: What if we don't have a `.backup` files? We really need to have them, because we are going to change `pg_hba.conf` and should be able to restore the old version.

patroni/postgresql.py
```diff
@@ -1459,21 +1518,49 @@ def clone(self, clone_member):
...
+
+    def post_bootstrap(self, config, task):
+        try:
+            self.create_or_update_role(self._superuser['username'], self._superuser['password'], ['SUPERUSER'])
+
+            # We were doing a custom bootstrap instead of running initdb, therefor we opened trust
+            # access from certain addresses to be able to reach cluster and change password
+            if self._running_custom_bootstrap:
+                self._running_custom_bootstrap = False
+                if not self.config.get('pg_hba'):
```
alexeyklyukin: shouldn't it be 'hba_file' instead of 'pg_hba'? The idea is to avoid removing the custom file, we don't really care about the user-supplied contents of pg_hba, since we will write it again at line 1548.

patroni/ha.py
```diff
@@ -102,6 +102,8 @@ def __init__(self, patroni):
         self.cluster = None
         self.old_cluster = None
         self.recovering = False
+        self._bootstraping = False
```
alexeyklyukin: I think the proper spelling is bootstrapping (2p)

patroni/postgresql.py
```diff
@@ -1459,21 +1518,49 @@ def clone(self, clone_member):
...
         else:
-            raise PostgresException("Could not bootstrap master PostgreSQL")
+            do_initialize = self._initdb
+        return do_initialize(config) and self._configure_server_parameters() and self.start()
+
+    def post_bootstrap(self, config, task):
+        try:
+            self.create_or_update_role(self._superuser['username'], self._superuser['password'], ['SUPERUSER'])
+
+            # We were doing a custom bootstrap instead of running initdb, therefor we opened trust
```
alexeyklyukin: therefore

patroni/postgresql.py
```diff
@@ -1081,8 +1114,23 @@ def _replace_pg_hba(self):
 
         :returns: True if pg_hba.conf was rewritten.
         """
-        if not self.config['parameters'].get('hba_file') and self.config.get('pg_hba'):
-            with open(os.path.join(self._config_dir, 'pg_hba.conf'), 'w') as f:
+
+        # when we are doing custom bootstrap we assume that we don't know superuser passwrod
```
alexeyklyukin: password

patroni/postgresql.py
```diff
@@ -1081,8 +1114,23 @@ def _replace_pg_hba(self):
...
+        # and in order to be able to change it, we are opening trust access from a sertain address
```
alexeyklyukin: certain

patroni/postgresql.py
```diff
@@ -1058,11 +1084,18 @@ def _write_postgresql_conf(self):
...
             f.write("include '{0}'\n\n".format(self.config.get('custom_conf') or self._postgresql_base_conf_name))
-            f.write("data_directory = '{}'\n".format(self._data_dir))
-            f.write("hba_file = '{}'\n".format(os.path.join(self._config_dir, 'pg_hba.conf')))
-            f.write("ident_file = '{}'\n".format(os.path.join(self._config_dir, 'pg_ident.conf')))
+            f.write("data_directory = '{0}'\n".format(self._data_dir))
             for name, value in sorted(self._server_parameters.items()):
-                f.write("{0} = '{1}'\n".format(name, value))
+                if not self._running_custom_bootstrap or name != 'hba_file':
+                    f.write("{0} = '{1}'\n".format(name, value))
+            # when we are doing custom bootstrap we assume that we don't know superuser passwrod
```
alexeyklyukin: password

patroni/postgresql.py
```diff
@@ -1058,11 +1084,18 @@ def _write_postgresql_conf(self):
...
+            # and in order to be able to change it, we are opening trust access from a sertain address
```
alexeyklyukin: certain

patroni/postgresql.py
```diff
@@ -1058,11 +1084,18 @@ def _write_postgresql_conf(self):
...
+            # therefore we need to make sure that hba_file is not overriden
+            # after changing superuser password we will "revert" all these "changes"
+            if self._running_custom_bootstrap or 'hba_file' not in self._server_parameters:
+                f.write("hba_file = '{0}'\n".format(self._pg_hba_conf))
+            if 'ident_file' not in self._server_parameters:
+                f.write("ident_file = '{0}'\n".format(os.path.join(self._config_dir, 'pg_ident.conf')))
```
CyberDem0n: It still needs to be there. Actually we never tried to backup it and postgres doesn't like it, i.e. shows warning.

patroni/postgresql.py
CyberDem0n: May it is confusing , but `hba_file` and `pg_hba` are different things...
It was possible to define a custom `hba_file` even before.
`pg_hba` - is a custom pg_hba configuration which would be enforced by Patroni.

In this case, if we don't have `pg_hba` defined, we need to restore original file.
On the line 1548 it will write into pg_hba.conf only if we have `pg_hba`.

alexeyklyukin: I know the difference. Why do we avoid removing pg_hba.conf if there is a custom pg_hba configuration?

CyberDem0n: You suggest to get rid from if? May be it make sense.

patroni/ha.py
```diff
@@ -937,6 +928,32 @@ def post_recover(self):
...
+    def post_bootstrap(self):
+        # bootstrap has failed if postgres is not running
+        if not self.state_handler.is_running() or self._post_bootstrap_task.result is False:
+            self.cancel_initialization()
+
+        self.keepalive()
+
+        if self._post_bootstrap_task.result is None and self.state_handler.is_leader():
+            self._async_executor.schedule('post_bootstrap')
+            self._async_executor.run_async(self.state_handler.post_bootstrap,
```
alexeyklyukin: Why do we need the ping-pong between HA -> PostgreSQL -> HA, first to run bootstrap in PostgreSQL, then to check that bootstrap didn't return False in HA, just to run post_bootstrap in PostgreSQL? Couldn't we do instead  
``return do_initialize(config) and self._configure_server_parameters() and self.start() and self.bost_bootstrap()`` at the end of ha.bootstrap?

CyberDem0n:
:+1:

Issue #455, opened 2017-06-15
Regression: pg_rewind is not called for Postgres v9.6.3


Vanav:
Commit https://github.com/zalando/patroni/commit/37c1552c0a834c7d40130551212a28ff9e286b04 breaks parsing of pg_controldata output for Postgres v9.6.3:
```
# LANG=C /usr/pgsql-9.6/bin/pg_controldata /var/lib/pgsql/9.6/data | grep -i wal_log_hints
wal_log_hints setting:                on
```
but https://github.com/zalando/patroni/blob/master/patroni/postgresql.py#L327:
```
    def configuration_allows_rewind(data):
        return data.get('Current wal_log_hints setting', 'off') == 'on' \
            or data.get('Data page checksum version', '0') != '0'
```
so this string never matches for recent Postgres versions.
There was fix for this before, but this commit erased it.

PR #456, opened 2017-06-16
BUGFIX: pg_rewind wasn't working when data page checksum is not enabled

CyberDem0n:
pg_controldata output depends on postgres major version and in some cases some of the parameters are prefixed by 'Current ' for old postgres versions.

Bug was introduced by commit 37c1552.
Fixes https://github.com/zalando/patroni/issues/455


alexeyklyukin:
👍

CyberDem0n:
:+1:

Issue #61, opened 2015-10-12
Support socket connections for local patroni instance


jberkus:
Currently, the local patroni instance can only connect using TCP.  We should support local unix socket connections as well; that's both faster and more secure.

This will require some replumbing, which I regard as a cleanup anyway.  Currently, we read config information, use that to create a URL, then parse that URL to get the config information back.  This seems unnecessarily ornate; we should either use URLs in the config in the first place, or we should avoid this whole unnecessary step and use the config fields directly.  For backwards-compatibility reasons, I plan to use the latter path.

I'll also use a file path for the socket interface if that's preferred by the user.  It should  be simple to differentiate this from an IPv4/v6 address.

I plan to do this in the next couple of days, so let me know if you have comments on the design.

LappleApple:
@jberkus Hi, did you finish or are you still working on this one? Wondering if we can resolve this issue. cc @feikesteenbergen

LappleApple:
@alexeyklyukin @feikesteenbergen @CyberDem0n @jberkus One more time: What's the status of this Issue? We could add a "Help Wanted" label if it's unfinished and still in scope.

rsiera:
@LappleApple can I work on this enhancement, or anyone else is involved in task at this moment ?

alexeyklyukin:
@rsiera yes, sure. I'd focus on supporting UNIX sockets instead of hosts in the configuration and would not change the way we work with URLs at the moment, in particular, avoid changing configuration parameters from the dictionary of keys to the URLs. Also, be careful when parsing URLs, as we supply an additional {{application_name}} parameter in dcs/__init__.py, that carries  an api URL.

CyberDem0n:
closed by https://github.com/zalando/patroni/pull/457

PR #457, opened 2017-06-21
Support unix sockets when connecting to a local postgres cluster

CyberDem0n:
This feature is enabled by default. If the `unix_socket_directories` is defined and non empty, Patroni will use the first value from it to connect to the local postgres cluster. Also unix socket will be used
when running `post_bootstrap` (`post_init`) script.

Set `postgresql.use_unix_socket: false` if you want to disable it.

Solves: https://github.com/zalando/patroni/issues/61

In addition to mentioned above, this commit solves couple of bugs:
* manual failover with pg_rewind in a pause state was broken
* psycopg2 (or libpq, I am not really sure what exactly) doesn't mark cursors connection as closed when we use unix socket and there is an `OperationalError` occurs. We will close such connection on our own.


patroni/postgresql.py
```diff
@@ -352,15 +372,21 @@ def sysid(self):
             self._sysid = data.get('Database system identifier', "")
         return self._sysid
 
-    def get_local_address(self):
+    def _get_unix_local_address(self):
+        if 'unix_socket_directories' in self._server_parameters:
```
alexeyklyukin: because of the indentation it took me a few passes to see how this function can return None. I'd write it as:
if 'unix_socket_directories not in self._server_parameters:
    return None
<the rest of the function>

docs/SETTINGS.rst
```diff
@@ -84,6 +84,7 @@ PostgreSQL
...
 -  **bin\_dir**: Path to PostgreSQL binaries. (pg_ctl, pg_rewind, pg_basebackup, postgres) The  default value is an empty string meaning that PATH environment variable will be used to find the executables.
 -  **listen**: IP address + port that Postgres listens to; must be accessible from other nodes in the cluster, if you're using streaming replication. Multiple comma-separated addresses are permitted, as long as the port component is appended after to the last one with a colon, i.e. ``listen: 127.0.0.1,127.0.0.2:5432``. Patroni will use the first address from this list to establish local connections to the PostgreSQL node.
+-  **use\_unix\_socket**: specifies that Patroni should prefer to use unix sockets to connect to the cluster. Default value is ``true``. If ``unix_socket_directories`` is definded, Patroni will use first sutable value from it to connect to the cluster and fallback to tcp if nothing is sutable. If ``unix_socket_directories`` is not specified in ``postgresql.parameters``, Patroni will assume that default value should be used and omit ``host`` from connection parameters.
```
alexeyklyukin: s/sutable/suitable/g

alexeyklyukin:
👍

alexeyklyukin:
👍

CyberDem0n:
:+1:

PR #458, opened 2017-06-22
Manage pg_hba.conf via patroni config or dynamic_configuration

CyberDem0n:
So far Patroni was populating pg_hba.conf only when running bootstrap code and after that it was not very handy to manage it's content, because it was necessary to login to every node, change pg_hba.conf
manually and run pg_ctl reload.

This commit intends to fix it and give Patroni control over pg_hba.conf It is possible to define pg_hba.conf content via `postgresql.pg_hba` in the patroni configuration file or in the `DCS/config` (dynamic configuration).

If the `hba_file` is defined in the `postgresql.parameters`, Patroni will ignore `postgresql.pg_hba`.


alexeyklyukin:
👍

CyberDem0n:
:+1:

Issue #459, opened 2017-06-27
Can't modify option wal_keep_segments without restarting database


Vanav:
It looks like option `wal_keep_segments` is added as command line argument to Postgres, so I can't modify it with `pg_ctl reload`. Is it possible to modify this option without restarting database?

CyberDem0n:
You are right, it is added as command line argument :(
And there is not a single reason to do that.
Fix is coming.

Vanav:
Thank you.

Maybe some other options from this list need to allow for modification too? I just have no need yet:
```
    CMDLINE_OPTIONS = {
        'listen_addresses': (None, lambda _: False, 90100),
        'port': (None, lambda _: False, 90100),
        'cluster_name': (None, lambda _: False, 90500),
        'wal_level': ('hot_standby', lambda v: v.lower() in ('hot_standby', 'replica', 'logical'), 90100),
        'hot_standby': ('on', lambda _: False, 90100),
        'max_connections': (100, lambda v: int(v) >= 100, 90100),
        'max_wal_senders': (5, lambda v: int(v) >= 5, 90100),
        'wal_keep_segments': (8, lambda v: int(v) >= 8, 90100),
        'max_prepared_transactions': (0, lambda v: int(v) >= 0, 90100),
        'max_locks_per_transaction': (64, lambda v: int(v) >= 64, 90100),
        'track_commit_timestamp': ('off', lambda v: parse_bool(v) is not None, 90500),
        'max_replication_slots': (5, lambda v: int(v) >= 5, 90400),
        'max_worker_processes': (8, lambda v: int(v) >= 8, 90400),
        'wal_log_hints': ('on', lambda _: False, 90400)
    }
```

CyberDem0n:
Other options are pretty important: either they must be the same across all cluster nodes (max_connections, max_wal_senders, max_prepared_transactions and so on) or they are needed to make certain features to work (for example wal_log_hints is needed for pg_rewind).

It's important to pass listen_addresses and port  as command line arguments, otherwise it would be not possible to access postgres if somebody decide to run 'alter system ...' and restart postgres.

In any case, rest of parameters needs a restart to change them.

Vanav:
Thank you.

Is there a way to upgrade Patroni without restarting database?

CyberDem0n:
It depends.

If Patroni is not running in a docker with PID=1 you can do following:

1. Bring cluster into a maintenance mode by executing: `patronictl pause <cluster-name>`
[...]

PR #460, opened 2017-06-27
Don't wal_keep_segments as command line argument to postgres

CyberDem0n:
It make it not possible to change it without restart.
Fixes https://github.com/zalando/patroni/issues/459


alexeyklyukin:
👍

CyberDem0n:
:+1:

Issue #461, opened 2017-06-27
ImportError: No module named consul


jouir:
Hello,

Patroni tries to import the consul module even if we don't use this DCS.

Based on the commit `b576e693622c3cf745fd9c444c50550d460546bc` cloned in the `patroni` directory, here is a Dockerfile to reproduce the error:
```
FROM debian:jessie
RUN apt-get update && apt-get upgrade -y
RUN apt-get install -y postgresql-9.4 python-psycopg2 python-dateutil python-yaml python-psutil python-requests
COPY patroni /opt/patroni
```
Then build:
```
$ docker build -t patroni-issue .
```
Then run to produce the error:
```
$ docker run --rm -it patroni-issue
# /opt/patroni/patroni.py /opt/patroni/postgres0.yml
Traceback (most recent call last):
  File "/opt/patroni/patroni.py", line 6, in <module>
    main()
  File "/opt/patroni/patroni/__init__.py", line 172, in main
    return patroni_main()
  File "/opt/patroni/patroni/__init__.py", line 139, in patroni_main
    patroni = Patroni()
  File "/opt/patroni/patroni/__init__.py", line 25, in __init__
    self.dcs = get_dcs(self.config)
  File "/opt/patroni/patroni/dcs/__init__.py", line 54, in get_dcs
    module = importlib.import_module(module_name)
  File "/usr/lib/python2.7/importlib/__init__.py", line 37, in import_module
    __import__(name)
  File "/opt/patroni/patroni/dcs/consul.py", line 8, in <module>
    from consul import ConsulException, NotFound, base
ImportError: No module named consul
```
Is it intended?
Same problem on a brand new debian 8 (without docker).

CyberDem0n:
Hi,

I think you get list of packages from the README file, which is slightly outdated :( 

In order to get it working you have to install some python modules:
``
$ sudo pip install -r requirements.txt
``

If you are looking for a Dockerfile, you can take https://github.com/zalando/patroni/blob/master/Dockerfile or https://github.com/zalando/spilo/blob/master/postgres-appliance/Dockerfile as an example

jouir:
My goal is to install patroni with minimal dependencies as possible. The module `python-consul` has to be installed via pip and this is sadly not an option for some political reasons in our project. Same with `python-etcd`. Let's say we are using zookeeper as a DCS with `python-kazoo` available in the debian repositories by default.

My question is, is it mandatory to install a client library we are not supposed to use?

[...]

PR #463, opened 2017-06-29
Don't fail if one of DCS implementation can't be loaded

CyberDem0n:
It might be that it's not required by configuration.

Fixes https://github.com/zalando/patroni/issues/461


alexeyklyukin:
👍

CyberDem0n:
:+1:

PR #466, opened 2017-07-01
Add config_dir option for configuration files location

### LINKED ISSUE #465: Set different directory for configuration files
Hello,

On debian, the configuration files (`postgresql.conf`, `pg_hba.conf`, etc) are not stored in the data directory. It would be great to be able to configure the location of this separate directory. Patroni could override existing configuration files where they are used to be.

It's quite confusing to see configuration files created in the data directory as a debian sysadmin. We are tempted to ignore them but they are effectively used by Patroni. On the other hand, Patroni ignores the configuration files in `/etc`. Using only one location should avoid confusion.

I will understand if you are not agree with this. I was a CentOS administrator few years ago, configuration files where stored in the data directory and everything was ok. Now that I have to manage plenty of debian systems with PostgreSQL running on it, things have changed. This pull request is reflecting this feeling.

The default is to store configuration files in the data directory. This setting is targeting custom installations like debian and any others moving configuration files out of the data directory.

jouir:
Fixes #465


CyberDem0n:
I don't have anything against possibility to specify `config_dir, but the topic is a little bit more complicated than you think.

1. Patroni MUST have write access to `postgresql.conf` and `pg_hba.conf` and as far as I can recall, entire `/etc/postgresql` is owned by root. It would be great to mention this fact in documentation.
2. There are two methods, `save_configuration_files` and `restore_configuration_files`: https://github.com/zalando/patroni/blob/master/patroni/postgresql.py#L1361 
    These methods should still make/restore a backups of configuration files to/from data directory.
    Why it's important? Because we really want these configuration files to be a backed up together with entire data.
    Why do we have these two methods? Because for example `wal-e` for some reason is skipping all `*.conf` file while backing up cluster.

jouir:
> Patroni MUST have write access to `postgresql.conf` and `pg_hba.conf` and as far as I can recall, entire `/etc/postgresql` is owned by root. It would be great to mention this fact in documentation.

On debian, it's owned by `postgres` so if you run Patroni under the `postgres` user, which I guess is the way we should deploy it, it's ok.
```
# ls -l /etc/postgresql/9.4/main/
total 48
-rw-r--r-- 1 postgres postgres     0 Jul  1 11:37 custom.conf
-rw-r--r-- 1 postgres postgres   315 Jul  1 21:21 environment
-rw-r--r-- 1 postgres postgres   143 Jul  1 21:21 pg_ctl.conf
-rw-r----- 1 postgres postgres  4641 Jul  1 21:21 pg_hba.conf
-rw-r----- 1 postgres postgres  1636 Jul  1 21:21 pg_ident.conf
-rw-r--r-- 1 postgres postgres 21026 Jul  1 21:21 postgresql.conf
-rw-r--r-- 1 postgres postgres   378 Jul  1 21:21 start.conf
```
But, yes, it would be great to mention this fact in the documentation in case of custom deployment.

> There are two methods, `save_configuration_files` and `restore_configuration_files`: https://github.com/zalando/patroni/blob/master/patroni/postgresql.py#L1361
These methods should still make/restore a backups of configuration files to/from data directory.
Why it's important? Because we really want these configuration files to be a backed up together with entire data.
Why do we have these two methods? Because for example `wal-e` for some reason is skipping all `*.conf` file while backing up cluster.

The `*.backup` files are generated by Patroni, we could let them in the data directory. `pg_basebackup` is backuping them:
```
$ pg_lsclusters 
Ver Cluster Port Status        Owner    Data directory                 Log file
9.4 node01  5433 online        postgres /var/lib/postgresql/9.4/node01 /var/log/postgresql/postgresql-9.4-node01.log
9.4 node02  5434 down,recovery postgres /var/lib/postgresql/9.4/node02 /var/log/postgresql/postgresql-9.4-node02.log
$ ls -l /var/lib/postgresql/9.4/node01/{pg_hba,postgresql}.conf.backup
-rw-r----- 1 postgres postgres 412 Jul  1 18:52 /var/lib/postgresql/9.4/node01/pg_hba.conf.backup
-rw-r--r-- 1 postgres postgres 474 Jul  1 18:52 /var/lib/postgresql/9.4/node01/postgresql.conf.backup
$ rm -rf /var/lib/postgresql/9.4/node02
$ pg_basebackup -p 5433 -U repl -R -D /var/lib/postgresql/9.4/node02
Password: 
NOTICE:  pg_stop_backup complete, all required WAL segments have been archived
$ ls -l /var/lib/postgresql/9.4/node02/{pg_hba,postgresql}.conf.backup
-rw-r----- 1 postgres postgres 412 Jul  1 21:17 /var/lib/postgresql/9.4/node02/pg_hba.conf.backup
-rw-r--r-- 1 postgres postgres 474 Jul  1 21:17 /var/lib/postgresql/9.4/node02/postgresql.conf.backup
```
In case of wal-e, that's a good point. I'm not using it so I can't tell.

patroni/postgresql.py
```diff
@@ -1367,16 +1369,16 @@ def save_configuration_files(self):
         try:
             for f in self._configuration_to_save:
                 if os.path.isfile(f):
-                    shutil.copy(f, f + '.backup')
+                    shutil.copy(f, f.replace(self._config_dir, self._data_dir) + '.backup')
```
CyberDem0n: This `replace` scares me.
I think it's better to be explicit, i.e. make `_configuration_to_save` return non-absolute file names ('postgresql.conf', 'pg_hba.conf', etc...) and copy them between `config_dir` and `data_dir`.

Everything else looks good to me.

jouir: You're right. I'm not really satisfied of this replace too. I will follow your advice and push a new branch as soon as possible. Thanks.

CyberDem0n: Actually one more thing, I think we need explicitly write locations of 'pg_hba.conf' and `pg_ident.conf` into `postgresql.conf`
https://github.com/jouir/patroni/blob/aaed71c51d46736aa0a6fc035e705ed73850e1ef/patroni/postgresql.py#L1057

```
hba_file = '/config/dir/pg_hba.conf'
ident_file = '/config/dir/pg_ident.conf'
```

jouir: What if `hba_file` and/or `ident_file` are in the `custom_conf` file? Are we supposed to override those settings by writing them after the `include` or before as a "default" value?

CyberDem0n: If they are custom - _configuration_to_save shouldn't return them

CyberDem0n:
:+1:

alexeyklyukin:
👍

PR #472, opened 2017-07-12
Update python-etcd dependency to the latest version

martyanov:
There are  several issues that were fixed after 0.4.3, currently in use, and in its dependencies. The latest release should be backwards compatible.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -5,7 +5,7 @@ PyYAML
 requests
 six >= 1.7
 kazoo==2.2.1
-python-etcd==0.4.3
+python-etcd==0.4.5
 python-consul==0.7.0
 click>=4.1
 prettytable>=0.7
--- a/requirements.txt
+++ b/requirements.txt
@@ -5,7 +5,7 @@ PyYAML
 requests
 six >= 1.7
 kazoo==2.2.1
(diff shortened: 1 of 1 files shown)


requirements.txt
```diff
@@ -5,7 +5,7 @@ PyYAML
 requests
 six >= 1.7
 kazoo==2.2.1
-python-etcd==0.4.3
+python-etcd==0.4.5
```
CyberDem0n: I am aware of some bug fixes, because actually one of them was reported and fixed by me.
Patroni is not using python-etcd as it is, but inherits from `etcd,Client` and override pretty much all methods and properties used to communicate with etcd nodes, i.e.: `machines` and `api_execute`.
That means non of the fixed bugs are hitting us.

I am not against 0.4.5, but probably it would be better to have something like  `python-etcd>=0.4.3` as a requirement.

martyanov: @CyberDem0n thanks for the clarifications. I think that having `etcd>=0.4.3` is fine. I've updated the pull request.

CyberDem0n:
:+1:

alexeyklyukin:
👍

CyberDem0n:
Merged, thank you!

Issue #481, opened 2017-07-19
pg_hba handling off with new custom bootstrap method


bradnicholson:
There are some issues with how the custom Postgres parameters & Patroni's handling of the pg_hba.conf file is interacting with the new custom bootstrap method. Looking at the following section of the patroni.yaml file where I set shared buffers and a custom hba_file:

```
postgresql:
  listen: 127.0.0.1:5432
  connect_address: 127.0.0.1:5432
  data_dir: data/postgresql0
#  bin_dir:
#  config_dir:
  pgpass: /tmp/pgpass0
  authentication:
    replication:
      username: replicator
      password: rep-pass
    superuser:
      username: postgres
      password: zalando
  parameters:
    unix_socket_directories: '.'
    shared_buffers: 256MB
    hba_file: /tmp/pg_hba.conf
```

If you initialize a new patroni cluster without specifying a custom bootstrap method, the custom hba_file gets used (and shared_buffers gets set to 256MB).

```
grep hba  data/postgresql0/postgresql.conf
hba_file = '/tmp/pg_hba.conf'
```
Now if you add a basic custom bootstrap method as follows:

```
bootstrap:
  # this section will be written into Etcd:/<namespace>/<scope>/config after initializing new cluster
  # and all other cluster members will use it as a `global configuration`
  method: custom_init
  custom_init:
    command: ./custom_bootstrap.sh
```

custom_bootstrap.sh
```
#!/bin/bash

initdb -D data/postgresql0/

``` 
The cluster is initialized, shared_buffers is set to 256MB, but it uses the default pg_hba location.

```
grep hba  data/postgresql0/postgresql.conf
hba_file = 'data/postgresql0/pg_hba.conf'
```

Shared_buffers getting set to 256MB indicates that we are still processing and applying the custom parameters.

Also, this may be intended - the pg_hba rules specified directly via the patroni.yaml file are also not being generated when using the custom bootstrap methods.

Using the default postgreql0.yml file included in master, if I start without a custom bootstrap script the pg_hba.conf adds the new rules as expected:

```
# TYPE  DATABASE        USER            ADDRESS                 METHOD

# "local" is for Unix domain socket connections only
local   all             all                                     trust
# IPv4 local connections:
host    all             all             127.0.0.1/32            trust
# IPv6 local connections:
host    all             all             ::1/128                 trust
# Allow replication connections from localhost, by a user with the
# replication privilege.
#local   replication     postgres                                trust
[...]

PR #482, opened 2017-07-21
Restart postgres after custom bootstrap if hba_file is defined in configuration

CyberDem0n:
In addition to that always use absolute paths to config files.

Fixes https://github.com/zalando/patroni/issues/481


alexeyklyukin:
👍

CyberDem0n:
👍

PR #483, opened 2017-07-22
Block callbacks during bootstrap

CyberDem0n:
It wasn't a big issue when on_start was called during normal boostrap
with initdb, because usually such process is very fast. But situation is
changing when we run custom bootstrap, becuase it might be a long time
between cluster become connectable and end of recovery and promote.

Actually situation was even worse than that, on_start was called with
the `replica` argument and later on_role_changes was never called,
because promote wasn't performed by Patroni.

As a solution for this problem we will block any callbacks during
bootstrap and explicitly call on_start after leader lock was taken.

### DIFF
--- a/patroni/ha.py
+++ b/patroni/ha.py
@@ -102,7 +102,6 @@ class Ha(object):
         self.cluster = None
         self.old_cluster = None
         self.recovering = False
-        self._bootstrapping = False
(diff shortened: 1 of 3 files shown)


alexeyklyukin:
👍

CyberDem0n:
:+1:

PR #486, opened 2017-07-26
Update release notes.

alexeyklyukin:
Draft release notes for Patroni 1.3


docs/releases.rst
```diff
@@ -3,6 +3,133 @@
...
+  The bootstrap command receives the name of the cluster and the path to the data directory. The resulting cluster can
+  be configured to start in recovery, making it possible to bootstrap a new cluster from the point in the timeline
+  of an existing one. Refer to the :ref:`documentaton page<custom_bootstap>` for more detailed description of this feature.
+
+**Smarter pg_rewind support**
+
+-  Decide on whether to run pg_rewind by looking at the timeline differences from the current master (Alexander)
+
+   Previously, Patroni had a fixed set of conditions to trigger pg_rewind, namely when starting a former master, when
+   doing a switchover to the designated node that is not a master or when there is a replica with the nofailover tag,
```
CyberDem0n: > switchover to the designated node that is not a master

sounds very strange, at least "that is not a master" needs to be dropped.
In fact, when we were doing a switchover to the designated node we were running pg_rewind on any other node in the cluster (except former master and new master).

docs/releases.rst
```diff
@@ -3,6 +3,133 @@
...
+   basically those cases when there is a chance that a replica may be ahead of the new master. In some cases, pg_rewind
+   did nothing, in some other ones it was not running when necessary. Instead of relying on this limited list of rules
+   this change makes Patroni compare the master and the replica WAL positions (using the streaming replication protocol)
+   in order to reliably decide if the replica to join is ahead of the new master.
```
CyberDem0n: > is ahead of the new master.

needs to be rewound.

docs/releases.rst
```diff
@@ -3,6 +3,133 @@
...
+   synchronous standby names to point to that replica and shortly afterwards sets it as a synchronous replica in DCS.
+   If the master dies in-between, there will be a synchronous replica retaining the writes, but Patroni will not be able
+   to pick that replica for the failover. In that case, one should perform a manual promotion to such a replica.
+
+**Configuration editing with patronictl**
+
+- Add configruation editing to patronictl (Ants Aasma, Alexander)
+
+  Add the ability to patronictl of editing dynamic cluster configuration stored in DCS. Support either specifying the
+  parameter/values from the command-line, involing the $EDITOR, or applying configuration from the yaml file.
```
a1exsh: typo: "involving".  or better "invoking"?

docs/releases.rst
```diff
@@ -3,6 +3,133 @@
...
+- Support storing of configuration files outside of PostgreSQL data directory (@jouir)
+
+  Add the new configuration ``postgresql`` configuration directive ``config_dif``.
+  It defaults to the data directory and must be writable by Patroni.
+
+**Bug fixes and stability improvements**
+
+- Handle EtcdEventIndexCleared and EtcdWatcherCleared exceptions (Alexander)
+
+  Faster recovery when the watch operation is ended by Etcd by avoiding useless retries (Alexander)
```
a1exsh: List markup missing at start of line: `-`

docs/releases.rst
```diff
@@ -3,6 +3,133 @@
...
+  the normal business while being bootstrapped by a bootstrap method that doesn't require a leader to be present in the
+  cluster.
+
+- Improve WAL-E replica creation method (Joar Wandborg, Alexander).
+
+  - Use csv.DictReader when parsing WAL-E base backup, accepting ISO dates with space-delimiated date and time.
+  - Extra logging when deciding between restoring from S3 or continuing with another replica method.
+  - Allow replica as a node to compare WAL segments against for making decision on whether to proceed with restoring from S3.
+
+- Fix pg_rewind invocation after as manual failover for Patroni is a paused state (Alexander)
```
a1exsh: typo: "in a paused state"

docs/releases.rst
```diff
@@ -391,4 +518,4 @@ This release adds support for *cascading replication* and simplifies Patroni man
 
   The tests can be launched manually using the *behave* command. They are also launched automatically for pull requests and after commits.
 
-  Releases notes for some older versions can be found on `project's github page <https://github.com/zalando/patroni/releases>`__.
\ No newline at end of file
+  Releases notes for some older versions can be found on `project's github page <https://github.com/zalando/patroni/releases>`__.
```
a1exsh: Not sure what changed on this line.  Newline removed or added?  In any case, there's a pre-existing typo: "Release notes".

docs/releases.rst
```diff
@@ -3,6 +3,133 @@
...
+   did nothing, in some other ones it was not running when necessary. Instead of relying on this limited list of rules
+   this change makes Patroni compare the master and the replica WAL positions (using the streaming replication protocol)
+   in order to reliably decide if the replica to join is ahead of the new master.
+
+**Synchronous replication mode strict**
+
+-  Enhance synchronous replication support by adding the strict mode (James Sewell, Alexander)
+
+   When the new configuration option ``synchronous_mode_strict`` is set, Patroni will not let the master accept write
+   requests if there are no replicas attached . This option provides the guarantee that no writes will be lost because
```
CyberDem0n: Way to complicated.

I would describe it differently, something like: If there are no replicas and ``synchronous_mode_strict`` is enabled, Patroni will disable writes on the master nodes. By default this feature is not enabled and running without replicas effectively disabling synchronous replication.

docs/releases.rst
```diff
@@ -3,6 +3,133 @@
...
+- Add configruation editing to patronictl (Ants Aasma, Alexander)
+
+  Add the ability to patronictl of editing dynamic cluster configuration stored in DCS. Support either specifying the
+  parameter/values from the command-line, involing the $EDITOR, or applying configuration from the yaml file.
+
+**Linux watchdog support**
+
+- Implement wachdog support for Linux (Ants)
+
+  Support Linux software watchdog in order to reboot the master node that losses the lock in Etcd. Normally, Patroni will
```
CyberDem0n: > reboot the master node that losses the lock in Etcd.

Watchdog doesn't know anything about Etcd, the main purpose of watchdog is tor reboot the node is Patroni is not running or hanging.

docs/releases.rst
```diff
@@ -3,6 +3,133 @@
...
+
+- Improve WAL-E replica creation method (Joar Wandborg, Alexander).
+
+  - Use csv.DictReader when parsing WAL-E base backup, accepting ISO dates with space-delimiated date and time.
+  - Extra logging when deciding between restoring from S3 or continuing with another replica method.
+  - Allow replica as a node to compare WAL segments against for making decision on whether to proceed with restoring from S3.
+
+- Fix pg_rewind invocation after as manual failover for Patroni is a paused state (Alexander)
+
+- Alleviate the issues with unix socket connections not marked as closed when unix sockets are used (Alexander)
```
CyberDem0n: It could be removed

docs/releases.rst
CyberDem0n: This could be removed

docs/releases.rst
alexeyklyukin: yeah, the idea was to highlight that we were doing pg_rewind on non-master nodes after a manual failover to the specific node, but I messed it up. Thanks!

docs/releases.rst
alexeyklyukin: It's an explanation to the point immediately above. Duplicated a name by mistake.

docs/releases.rst
alexeyklyukin: Hm... perhaps we can formulate it even simpler '... Patroni will not disable synchronous replication in the lack of any replicas, effectively blocking all clients writing data to the master until a synchronous replica appear.' I'm keeping the 'guarantee  that no writes will be lost because of the automatic failover' sentence.

docs/releases.rst
alexeyklyukin: Yeah, I was under impression that it only acts on master nodes, but you are right, when Patroni is not running there is no way to figure out from outside which node has a lock in Etcd, neither it should matter, as promoting the former replica node without running patroni will be equally bad.

docs/releases.rst
alexeyklyukin: hm... I think wen an remove the extra logging, but not being able to use the replica as a source of calculating whether we should go with the S3 or not is an actual bug someone may come across in the older versions (same for pg_rewind).

docs/releases.rst
alexeyklyukin: Or maybe not, as the last sentence really requires some explanation.

ants: The non-strict synchronous mode is "no writes will be lost due to automatic failover". The strict mode is "all writes are two node durable", meaning that a catastrophic loss of any one node (node, including storage, is gone and is not coming back) will not lead to lost writes.

The situation where it matters is when the sync standby goes down for a bit and while the standby is down and a new one hasn't been chosen the master completely explodes.

docs/releases.rst
ants: Actually with #452 I'm making it only act on master nodes, running it on standby systems made the feature much more complex than it needed to be and did not buy much.

It is protecting against Patroni dying due to bugs, OOM killer or sysadmin killing Patroni or a swap storm/VM migration/etc. making it impossible for Patroni to perform a timely demote. I'll try come up with a decent release note while the behave tests are running here.

docs/releases.rst
alexeyklyukin: @ants good point, thank you. I'll take the 'all writes are two nodes durable' into the notes.

docs/releases.rst
ants: Support Linux watchdog devices to reboot the master if Patroni is not able to operate in a timely manner to uphold
    single master guarantee. This protects against Patroni process getting terminated by bugs or external causes.
    (e.g. kernel OOM killer) It also protects against situations where Patroni is not able to disable the master
    before leader key expires. (e.g. when the database is on an unresponsive networked disk)

    When watchdog mechanism is enabled Patroni will set it up so that watchdog will terminate the system before any
    other system becomes eligible for leader. Watchdog enables Patroni to give almost as good guarantees as traditional
    clustering systems provide with a much more complicated STONITH mechanism.

    It is possible to configure the watchdog device (`/dev/watchdog` by default) and the mode (required, automatic, off) 
    from the watchdog section of the Patroni configuration. By default Patroni will try to use the watchdog device
    if the permissions allow, but will start without if the device is missing or not accessible. You can get more
    information from the :ref:`watchdog documentation<watchdog>`.

ants:
I think we should add an "Upgrading" section too. As far as I can tell, it's just a matter of installing new version and restart Patroni on all nodes before enabling new features, but it would be nice to have it documented so people don't have to guess.

docs/releases.rst
```diff
@@ -3,6 +3,126 @@
 Release notes
 =============
 
+Version 1.3
+-----------
+
+Version 1.3 adds custom bootstap possibility, significantly improves support for pg_rewind, enhances the
```
ants: s/bootstap/bootstrap/

docs/releases.rst
```diff
@@ -3,6 +3,126 @@
...
+In addition, this is the first version to work correctly with PostgreSQL 10.
+
+**Custom bootstrap**
+
+- Make the process of bootstrapping the cluster configurable (Alexander Kukushkin)
+
+  Allow custom bootstrap scripts instead of ``initdb`` when initializing the very first node in the cluster.
+  The bootstrap command receives the name of the cluster and the path to the data directory. The resulting cluster can
+  be configured to start in recovery, making it possible to bootstrap a new cluster from the point in the timeline
+  of an existing one. Refer to the :ref:`documentaton page <custom_bootstap>` for more detailed description of this feature.
```
ants: I think the link should be to `<replica_bootstrap>`

docs/releases.rst
```diff
@@ -3,6 +3,126 @@
...
+  Add the ``use_unix_socket`` option to the postgresql section of Patroni configuration. When set to true and the
+  Postgresql ``unix_socket_directories`` option is not empty, enables Patroni to use the first value from it to connect
+  to the local PostgreSQL cluster. If ``unix_socket_directories`` is not defined, Patroni will assume its default value
+  and omit the ``host`` parameter in the PostgreSQL connection string altogether.
+
+- Support change of superuser and replication credentials on reload (Alexander)
+
+- Support storing of configuration files outside of PostgreSQL data directory (@jouir)
+
+  Add the new configuration ``postgresql`` configuration directive ``config_dif``.
```
ants: `config_dir`

docs/releases.rst
ants: I would change the second to last sentence to "The resulting cluster can be configured to perform recovery, making it possible to bootstrap from a backup and do point in time recovery."

alexeyklyukin:
@ants thank you, I've added a very short upgrade notice. I suppose (should check) that we can even upgrade Patroni without disrupting PostgreSQL. it should be as easy as putting Patroni in a pause mode and then restarting it  (and running ``patronictl resume`` afterwards). Not sure if it would play nicely with the watchdog (although Patroni restart is normally very fast.)

ants:
Restarting while paused should work indeed. Pause mode is not a problem for watchdog, current master will disable watchdog when exiting, with #452 watchdog will get disabled when entering paused mode.

alexeyklyukin:
Thank you! I've added a few lines about upgrading.

alexeyklyukin:
👍

alexeyklyukin:
👍

CyberDem0n:
:+1:

Issue #488, opened 2017-07-28
POST failover doesn't work


unixwitch:
Hello,

I set up a new cluster using Patroni 1.3, ZooKeeper 3.4.10 and PostgreSQL 9.6.3 on CentOS 7.  Everything seems to work fine, except failover via the API.

I sent a request like this:

```
[root@postgres-5-2 ~]# curl -d'{"leader": "postgres-5-2", "candidate": "postgres-5-3"}' -H'Content-Type: application/json' http://admin:xxx@postgres-5-2:8008/failover
curl: (52) Empty reply from server
```

Patroni error logs show an exception:

```
Jul 28 12:35:31 postgres-5-2 patroni[1269]: 2017-07-28 12:35:31,258 INFO: received failover request with leader=postgres-5-2 candidate=postgres-5-3 scheduled_at=None
Jul 28 12:35:31 postgres-5-2 patroni[1269]: 2017-07-28 12:35:31,276 INFO: Got response from postgres-5-3 http://postgres-5-3:8008/patroni: b'{"role": "replica", "postmaster_start_time": "2017-07-28 12:31:09.295 UTC", "state": "running", "xlog": {"replayed_timestamp": "2017-07-28 12:31:09.208 UTC", "paused": false, "replayed_location": 369514424, "received_location": 369514424}, "server_version": 90603, "patroni": {"version": "1.3", "scope": "postgres-5"}, "database_system_identifier": "6447546693452231673"}'
Jul 28 12:35:31 postgres-5-2 patroni[1269]: 2017-07-28 12:35:31,367 INFO: Got response from postgres-5-3 http://postgres-5-3:8008/patroni: b'{"role": "replica", "postmaster_start_time": "2017-07-28 12:31:09.295 UTC", "state": "running", "xlog": {"replayed_timestamp": "2017-07-28 12:31:09.208 UTC", "paused": false, "replayed_location": 369514424, "received_location": 369514424}, "server_version": 90603, "patroni": {"version": "1.3", "scope": "postgres-5"}, "database_system_identifier": "6447546693452231673"}'
Jul 28 12:35:34 postgres-5-2 patroni[1269]: Traceback (most recent call last):
Jul 28 12:35:34 postgres-5-2 patroni[1269]: File "/usr/lib64/python3.4/socketserver.py", line 617, in process_request_thread
Jul 28 12:35:34 postgres-5-2 patroni[1269]: self.finish_request(request, client_address)
[...]

PR #489, opened 2017-07-28
Bugfix: failover via API didn't work due to change in _MemberStatus

CyberDem0n:
Originally fetch_nodes_statuses was returning a tuple, later it was
wrapped into namedtuple _MemberStatus and recently _MemberStatus was
extened with watchdog_failed field, but api.py was still relying on
usual tuple and checking failover limitations on it's own instead of
calling `failover_limitation` method.

Fixes: https://github.com/zalando/patroni/issues/488


CyberDem0n:
:+1:

alexeyklyukin:
👍

Issue #491, opened 2017-07-28
patronictl edit-config doesn't work


unixwitch:
Hello,

Thank you for fixing #488 :-)  However, I have another problem:

Using Patroni 1.3 with ZooKeeper 3.4.10, `patronictl edit-config` doesn't seem to work.  Any change (even a trivial one, like changing `ttl` from 30 to 31) gives the same error:

```
[root@postgres-5-2 ~]# patronictl -c /etc/patroni.yaml edit-config postgres-5
--- 
+++ 
@@ -11,4 +11,4 @@
   use_pg_rewind: true
   use_slots: true
 retry_timeout: 10
-ttl: 30
+ttl: 31

Apply these changes? [y/N]: y
2017-07-28 23:27:53,965 - ERROR - Unhandled exception in connection loop
Traceback (most recent call last):
  File "/usr/lib/python3.4/site-packages/kazoo/protocol/connection.py", line 541, in _connect_attempt
    self._send_request(read_timeout, connect_timeout)
  File "/usr/lib/python3.4/site-packages/kazoo/protocol/connection.py", line 445, in _send_request
    self._submit(request, connect_timeout, xid)
  File "/usr/lib/python3.4/site-packages/kazoo/protocol/connection.py", line 279, in _submit
    b += request.serialize()
  File "/usr/lib/python3.4/site-packages/kazoo/protocol/serialization.py", line 195, in serialize
    b.extend(int_struct.pack(self.version))
struct.error: 'i' format requires -2147483648 <= number <= 2147483647
Exception in thread Thread-3:
Traceback (most recent call last):
  File "/usr/lib64/python3.4/threading.py", line 911, in _bootstrap_inner
    self.run()
  File "/usr/lib64/python3.4/threading.py", line 859, in run
    self._target(*self._args, **self._kwargs)
  File "/usr/lib/python3.4/site-packages/kazoo/protocol/connection.py", line 471, in zk_loop
    if retry(self._connect_loop, retry) is STOP_CONNECTING:
  File "/usr/lib/python3.4/site-packages/kazoo/retry.py", line 123, in __call__
    return func(*args, **kwargs)
  File "/usr/lib/python3.4/site-packages/kazoo/protocol/connection.py", line 488, in _connect_loop
    status = self._connect_attempt(host, port, retry)
  File "/usr/lib/python3.4/site-packages/kazoo/protocol/connection.py", line 541, in _connect_attempt
    self._send_request(read_timeout, connect_timeout)
[...]

PR #492, opened 2017-07-29
BUGFIX: patronictl edit-config didn't worked with zookeeper

CyberDem0n:
When updating config key we should use `ClusterConfig.index` instead of
`ClusterConfig.modify_index`. The second one should be used by Patroni
internally to check that key was really changed, because when key is
deleted and recreated it's version always starts from the same value: 0

In addition to that use patronictl instead of http PATCH in some of
acceptance tests to change cluster config.

Fixes https://github.com/zalando/patroni/issues/491


CyberDem0n:
👍

alexeyklyukin:
👍

PR #496, opened 2017-08-01
Fixes #494 - Custom Bootrap Temp hba.conf

### LINKED ISSUE #494: Issues when using custom bootstrap
When trying to perform a restore using wal-e, on a Spilo image, Spilo tries to modify the template1 database. 
Patroni creates a temprary pg_hba file for use during the custom bootstrap, but it only allows to connect to the "postgres" db ([line 1140 or postgresql.py](https://github.com/zalando/patroni/blame/master/patroni/postgresql.py#L1140)).
In order to allow Spilo to create those extensions, instead of setting up "postgres" as the db, it would be handy to set "all" there.

As far as I could see there is no way to pass that as parameter, am I right?

Ideas?
Thanks

francobellagamba:
Fixes #494

### DIFF
--- a/patroni/postgresql.py
+++ b/patroni/postgresql.py
@@ -1137,7 +1137,7 @@ class Postgresql(object):
             with open(self._pg_hba_conf, 'w') as f:
                 f.write(self._CONFIG_WARNING_HEADER)
                 for address, t in addresses.items():
-                    f.write('{0}\t{1}\t{2}\t{3}\ttrust\n'.format(t, self._database,
+                    f.write('{0}\t{1}\t{2}\t{3}\ttrust\n'.format(t, 'all'
                                                                  self._superuser.get('username') or 'all', address))
         elif not self._server_parameters.get('hba_file') and self.config.get('pg_hba'):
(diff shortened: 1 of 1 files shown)


patroni/postgresql.py
```diff
@@ -1137,7 +1137,7 @@ def _replace_pg_hba(self):
             with open(self._pg_hba_conf, 'w') as f:
                 f.write(self._CONFIG_WARNING_HEADER)
                 for address, t in addresses.items():
-                    f.write('{0}\t{1}\t{2}\t{3}\ttrust\n'.format(t, self._database,
+                    f.write('{0}\t{1}\t{2}\t{3}\ttrust\n'.format(t, 'all'
```
CyberDem0n: comma after `'all'` is missing.

francobellagamba: That's why you shouldn't code at 7AM... :stuck_out_tongue:

CyberDem0n:
:+1:

alexeyklyukin:
👍

CyberDem0n:
Merged, thank you!

PR #499, opened 2017-08-04
Small bugfixes

CyberDem0n:
* Create empty pg_ident.conf if it is missing after restoring from backup
* Short after promote synchronous replication was disabled even is synchronous_mode_strict is set


alexeyklyukin:
👍

CyberDem0n:
:+1:

PR #500, opened 2017-08-09
Make Patroni Kubernetes native

CyberDem0n:
In order to be able to find all objects related to our Patroni cluster we use `labels` and `labelSelector`.

patroni.yaml
```yaml
name: batman-0  # MUST match with the Pod name where Patroni is running
scope: &scope batman # Patroni cluster name
kubernetes:
  namespace: default  # Kubernetes namespace where Patroni Pods are running and where Patroni will create different ConfigMap objects
  labels:  # Set of labels which will uniquely identify Patroni cluster.
    application: patroni  # They will be used to find all Pods and ConfigMaps associated with the cluster.
    cluster-name: &scope  # Also Patroni will set these labels when to all ConfigMaps object it creates.
```

Unfortunately Kubernetes API doesn't provide possibility to expire objects, but provides only compare-and-set functionality, therefore we had to implement leader election by periodically updating annotations of `<scope>-leader` ConfigMap object.
Basic idea is taken from https://github.com/kubernetes/client-go/tree/master/tools/leaderelection.
Every node in the cluster periodically checks the annotations of `<scope>-leader` ConfigMap object. If annotations were changed, that means we have a leader, if annotations weren't changed during `ttl` seconds, that means cluster doesn't have a leader.

List of ConfigMaps Patroni is working with:
* `initialize` and `config` keys are stored as annotations of `<scope>-config` ConfigMap
* `leader` and `optime/leader` keys are stored as annotations of `<scope>-leader` `ConfigMap`
* `failover` key is stored as annotations of `<scope>-failover` `ConfigMap`
* `sync` key is stored as annotations of `<scope>-sync` `ConfigMap`

Open questions:
* Should Patroni also label Pods with the 'role'?
* Should Patroni maintain master Endpoint?
* Or we could still use callback for that?


CyberDem0n:
@jberkus, I think you should really try it. It is using Kubernetes API and config maps to store cluster state.

Here you can find  Dockerfile and Kubernetes manifest to deploy it: https://github.com/zalando/patroni/tree/feature/k8s/kubernetes

jberkus:
Will test!

FWIW, it's possible Kube will add leader elections in the future.

jberkus:
*   Should Patroni also label Pods with the 'role'?

Yes.  This is very useful for a quick check on who's the current master via Kubectl.  We'll just need to stress that the information isn't determinative.

*    Should Patroni maintain master Endpoint?
*    Or we could still use callback for that?

Example?  I'm not aware of a way to do this in Kubernetes.

CyberDem0n:
> Yes. This is very useful for a quick check on who's the current master via Kubectl. We'll just need to stress that the information isn't determinative.

>    Should Patroni maintain master Endpoint?
>    Or we could still use callback for that?

> Example? I'm not aware of a way to do this in Kubernetes.

Right now it's all done by a callback, example is here:
https://github.com/zalando/patroni/blob/feature/k8s/kubernetes/entrypoint.sh#L26 -- callbacks
https://github.com/zalando/patroni/blob/feature/k8s/kubernetes/patroni_k8s.yaml#L89 -- endpoint declaration
https://github.com/zalando/patroni/blob/feature/k8s/kubernetes/patroni_k8s.yaml#L98 -- service declaration (it is using the endpoint from a previous line)
https://github.com/zalando/patroni/blob/feature/k8s/kubernetes/callback.py#L41 -- callback, change endpoint
https://github.com/zalando/patroni/blob/feature/k8s/kubernetes/callback.py#L49 -- callback, set and remove role label

jberkus:
Right, I meant, what's the alternative?

jberkus:
BTW, should I be paying attention to codacy-bot?

kubernetes/callback.py
```diff
@@ -0,0 +1,78 @@
...
+
+    action, role, cluster = sys.argv[1:4]
+
+    k8s_config.load_incluster_config()
+    k8s_api = CoreV1Api()
+
+    namespace = os.environ['KUBERNETES_NAMESPACE']
+
+    if role == 'master' and action in ('on_start', 'on_role_change'):
+        patch_master_endpoint(k8s_api, namespace, cluster)
```
ants: Doing this as a callback is fragile, If the API call fails the cluster is effectively unavailable until failover is externally triggered. 

Also, there is no interlock to prevent racy updates. Imaginary scenario:

0. unhealthy cluster
1. n-0 acquires leader lock, starts callback.py
2. n-0 hangs for a period of time (IO system problems, network issues, etc.)
3. n-1 acquires leader lock, starts callback.py
4. n-1 callback.py completes, directing endpoint traffic to n-1
5. n-0 becomes active again, runs the API call and overwrites the endpoint with n-0
6. n-0 demotes itself
7. until next failover traffic is now routed to the standby.

I think this needs integration with the HA loop to fix properly.

kubernetes/callback.py
```diff
@@ -0,0 +1,78 @@
+#!/usr/bin/env python
```
ants: Any ideas how to handle systems where Python2 and Python3 are installed in parallel and we want to install Patroni and dependencies on Python3. Currently I just had the Dockerfile sed in the correct version, but that feels a bit hacky...

patroni/ha.py
```diff
@@ -85,15 +85,13 @@ def acquire_lock(self):
         return self.dcs.attempt_to_acquire_leader()
 
     def update_lock(self, write_leader_optime=False):
-        ret = self.dcs.update_leader()
-        if ret:
-            self.watchdog.keepalive()
```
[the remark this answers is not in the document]
CyberDem0n: Good catch, will fix it tomorrow.

kubernetes/callback.py
CyberDem0n: Not really.
We just call such scripts not as executables, but with specifying interpreter in front: `python3 /callback.py`

kubernetes/callback.py
CyberDem0n: That was just a fast way to copy a script we already have in Spilo: https://github.com/zalando/spilo/blob/master/postgres-appliance/callback_role.py

Agree, it probably should be done from HA loop.

patroni/dcs/kubernetes.py
```diff
@@ -0,0 +1,321 @@
...
+        return ret
+
+    def take_leader(self):
+        return self.attempt_to_acquire_leader()
+
+    def set_failover_value(self, value, index=None):
+        """Unused"""
+
+    def manual_failover(self, leader, candidate, scheduled_at=None, index=None):
+        annotations = {'leader': leader or None, 'member': candidate or None, 'scheduled_at': scheduled_at}
```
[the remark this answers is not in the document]
CyberDem0n: What kubernetes version do you use?
I've been testing it with 1.7. If we pass a null value there it should just remove a key from ConfigMap

CyberDem0n:
> I'm a tiny bit worried about the 3 API calls per HA loop

If node is a leader, it will actually make 4 calls.
My initial idea was to make a proof of concept, which later we can improve. We can easily avoid the 3rd API call updating pods annotations with the latest information and reduce number of writes by scarifying some useful information (replication lag) from patronictl list output. Well, we can still calculate it by querying every patroni instance individually.
Read queries in my opinion should not cause a big problems.

> The clock comparisons for the leader key feel like they might have nasty edge cases. I will try to find some quiet time to meditate on them.

Something nasty may happen only if time on one node runs faster than on another one by ttl/loop_wait times.

kubernetes/callback.py
jberkus: There's always going to be the possibility of code halting between checking role='master' and patch_master_endpoint, no?  Or are you suggesting that the master should have a loop which constantly checks that it still owns the endpoint, so that incoherency can be limited to the loop time?

jberkus: I suppose the demotion code needs to have an endpoint check as well.  Although what would a demoting node do if it found it owned the endpoint?

patroni/dcs/kubernetes.py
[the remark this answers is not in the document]
ants: OpenShift 3.5/Kubernetes 1.5.

I'll investigate closer.

kubernetes/callback.py
ants: Of course. Just specifying it in the config makes much more sense than patching the file itself.

kubernetes/callback.py
ants: Yes I think the loop should check that it still own the endpoint. Also, I think it should release the leader key if it fails to get the endpoint within a reasonable amount of time. I think this might have some overlap with my todo item of implementing basic OCF resource support. Similar considerations would apply with moving a virtual IP around, etc.

patroni/dcs/kubernetes.py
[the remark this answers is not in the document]
ants: On 1.5 patch with a null value removes the key, but create with a null value errors out.
On 1.6 create with a null value will result in a config map with empty string as the value.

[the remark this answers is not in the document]
CyberDem0n: This is very easy to fix. I will just strip all annotations with null values if we are creating a new ConfigMap object

kubernetes/callback.py
CyberDem0n: In order not to have a race conditions master endpoint should be changed only by leader.
Current implementation is using ConfigMaps to store information about leader. The same way we can use Endpoints instead. I.e. Patroni can periodically (every HA loop) update annotations and **subsets** of the leader Endpoint.

Off course such solution has it's own pros and cons.

Pros: when the new master get a leader lock it will atomically change master service

Cons:
 * I have no idea how kubernetes will react on periodic updates of leader `Endpoint`. Every update in theory should cause update of leader `Service`. We should really check that it will not trigger rebuild of iptables rules everywhere.
 * Right now Patroni is does 2 API reads (list_namespaced_config_map and list_namespaced_pod). New implementation will do 3 calls (list_namespaced_endpoint). We kind of can avoid it by converting all ConfigMaps to Endpoints, but I am not sure that it is a good idea.

ants: Directly patching Endpoint has another downside - OpenShift does not let normal users make endpoints that point to pods. As documented [in the final paragraph here](https://docs.openshift.com/container-platform/3.5/architecture/core_concepts/pods_and_services.html). At least for OpenShift I would like to rely on pod annotations and have the platform take care of routing a Service to the master. But even that still has the same issue that for reliability a callback is less than optimal.

patroni/dcs/kubernetes.py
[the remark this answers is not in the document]
ants: This fixed the create case, but there is still something strange going on for the patch case. Isolated test shows that this should be working, but when using Patroni I'm still getting the same error. More investigation needed...

kubernetes/callback.py
jberkus: I'm nervous about the idea of dropping the leader lock if grabbing the endpoint/service fails.  That ups the number of sequential operations which need to succeed to three; that seems prone to race-condition type failures.  

Some unresolved questions:

1. What should be the proper ordering of leader lock, failover/promotion, endpoint?  
2. What's the sequence of undoing these actions if any of them fails?  What happens if the undo fails?
3. What do we want to be the result if  kubernetes is unresponsive/misconfigured and no node can update the endpoint?

That last case is potentially pathological, and argues that updating the endpoint needs to come *before* failover/promotion instead of afterwards, since changing the endpoint is easily reversable and failover is not.  If we keep it afterwards, having RBAC config issues or container networking issues could mean that you have a set of disconnected nodes which can't be ever reconnected as replicas, and among which Patroni can't choose a good new master candidate.

I think this needs a deeper discussion than a line-item comment on a patch review.

CyberDem0n: @ants, @jberkus:

I this this commit https://github.com/zalando/patroni/pull/500/commits/baaf20b8cf381bb04773d2b3bd45a528a6d59047 should fix most of the problems.
Since Patroni is already doing some metadata manipulations on the Pod, why not assign label with the `role` right from HA loop? It makes it possible to use labelSelector without any external callback script.

Unfortunately it is still not solving a biggest problem, when the master Pod hangs for a long time and looses the leader key. In this case it still could happen that two Pods will get role=master...

It looks like the only reliable way to make sure that master Endpoint has only one IP is patching it from Patroni HA loop together (and atomically) with updating leader lock.

jberkus: Let's discuss this on the race conditions issue

unguiculus:
This looks very promising and I would like to test it. I built the Docker image as suggested at https://github.com/zalando/spilo/issues/188#issuecomment-332107742. Pods come up but leader election does not work.

```
patronidemo-0 patronidemo 2017-11-22 14:34:23,645 INFO: Lock owner: None; I am patronidemo-0
patronidemo-0 patronidemo 2017-11-22 14:34:23,645 INFO: waiting for leader to bootstrap
patronidemo-1 patronidemo 2017-11-22 14:34:25,639 INFO: Lock owner: None; I am patronidemo-1
patronidemo-1 patronidemo 2017-11-22 14:34:25,640 INFO: waiting for leader to bootstrap
patronidemo-2 patronidemo 2017-11-22 14:34:26,321 INFO: Lock owner: None; I am patronidemo-2
patronidemo-2 patronidemo 2017-11-22 14:34:26,321 INFO: waiting for leader to bootstrap
```

It this PR currently in a testable state? Anything else I need to do? Do you have plans to merge this anytime soon?

CyberDem0n:
Hi @unguiculus 

> It this PR currently in a testable state? 

Yes, we even run a few clusters on kubernetes by using this branch.

> Pods come up but leader election does not work.
patronidemo-0 patronidemo 2017-11-22 14:34:23,645 INFO: Lock owner: None; I am patronidemo-0
patronidemo-0 patronidemo 2017-11-22 14:34:23,645 INFO: waiting for leader to bootstrap

It would be good to check all logs of the pods, not only latest once.

> Anything else I need to do?

Can you check what endpoints and configmaps exists there and what is inside them?
```bash
$ kubectl get cm
NAME                 DATA      AGE
patronidemo-config   0         6m
patronidemo-leader   0         6m

$ kubectl get cm patronidemo-config -o yaml
apiVersion: v1
kind: ConfigMap
metadata:
  annotations:
    config: '{"postgresql":{"use_pg_rewind":true}}'
    initialize: "6491261098181615669"
  creationTimestamp: 2017-11-22T15:28:19Z
  labels:
    application: patroni
    cluster-name: patronidemo
  name: patronidemo-config
  namespace: default
  resourceVersion: "1190861"
  selfLink: /api/v1/namespaces/default/configmaps/patronidemo-config
  uid: c4a43b24-cf99-11e7-8c50-68f728c1b840

$ kubectl get cm patronidemo-leader -o yaml
apiVersion: v1
kind: ConfigMap
metadata:
  annotations:
    acquireTime: 2017-11-22T15:28:22.489225+00:00
    leader: patronidemo-0
    optime: "67109184"
    renewTime: 2017-11-22T15:36:02.476654+00:00
    transitions: "0"
    ttl: "30"
  creationTimestamp: 2017-11-22T15:28:22Z
  labels:
    application: patroni
    cluster-name: patronidemo
  name: patronidemo-leader
  namespace: default
  resourceVersion: "1191480"
  selfLink: /api/v1/namespaces/default/configmaps/patronidemo-leader
  uid: c67af485-cf99-11e7-8c50-68f728c1b840

```

Also you may try to clean everything up and start from scratch:
```bash
$ kubectl delete -f patroni_k8s.yaml
$ kubectl delete cm patronidemo-leader patronidemo-config


$ kubectl apply -f patroni_k8s.yaml
```

> Do you have plans to merge this anytime soon?

Not really soon, but yes, we do have such plans.
There are couple of unclear moments needs to be addressed and documentation is missing.

unguiculus:
Thanks for the quick response. The configmaps were indeed the problem. After deleting them I managed to get it working. It runs well using an `emptyDir` but switching the a persistent volume (which is commented out in (`patroni_k8s.yaml`) gives me  a permissions issue. I have yet to look into this. 

```
postgres-patroni-k8s-0 postgres-patroni-k8s 2017-11-23 07:23:56,813 INFO: Failed to import patroni.dcs.consul
postgres-patroni-k8s-0 postgres-patroni-k8s 2017-11-23 07:23:56,814 INFO: Failed to import patroni.dcs.etcd
postgres-patroni-k8s-0 postgres-patroni-k8s 2017-11-23 07:23:56,815 INFO: Failed to import patroni.dcs.exhibitor
postgres-patroni-k8s-0 postgres-patroni-k8s 2017-11-23 07:23:56,999 INFO: Lock owner: None; I am postgres-patroni-k8s-0
postgres-patroni-k8s-0 postgres-patroni-k8s 2017-11-23 07:23:57,017 INFO: trying to bootstrap a new cluster
postgres-patroni-k8s-0 postgres-patroni-k8s The files belonging to this database system will be owned by user "postgres".
postgres-patroni-k8s-0 postgres-patroni-k8s This user must also own the server process.
postgres-patroni-k8s-0 postgres-patroni-k8s
postgres-patroni-k8s-0 postgres-patroni-k8s The database cluster will be initialized with locale "en_US.UTF-8".
postgres-patroni-k8s-0 postgres-patroni-k8s The default text search configuration will be set to "english".
postgres-patroni-k8s-0 postgres-patroni-k8s
postgres-patroni-k8s-0 postgres-patroni-k8s Data page checksums are enabled.
postgres-patroni-k8s-0 postgres-patroni-k8s
postgres-patroni-k8s-0 postgres-patroni-k8s initdb: could not create directory "/home/postgres/pgdata/pgroot": Permission denied
postgres-patroni-k8s-0 postgres-patroni-k8s pg_ctl: database system initialization failed
postgres-patroni-k8s-0 postgres-patroni-k8s 2017-11-23 07:23:57,045 INFO: removing initialize key after failed attempt to bootstrap the cluster
postgres-patroni-k8s-0 postgres-patroni-k8s Traceback (most recent call last):
postgres-patroni-k8s-0 postgres-patroni-k8s   File "/usr/local/bin/patroni", line 11, in <module>
postgres-patroni-k8s-0 postgres-patroni-k8s     load_entry_point('patroni==1.3.6', 'console_scripts', 'patroni')()
postgres-patroni-k8s-0 postgres-patroni-k8s   File "/usr/local/lib/python2.7/dist-packages/patroni/__init__.py", line 174, in main
postgres-patroni-k8s-0 postgres-patroni-k8s     return patroni_main()
postgres-patroni-k8s-0 postgres-patroni-k8s   File "/usr/local/lib/python2.7/dist-packages/patroni/__init__.py", line 143, in patroni_main
postgres-patroni-k8s-0 postgres-patroni-k8s     patroni.run()
postgres-patroni-k8s-0 postgres-patroni-k8s   File "/usr/local/lib/python2.7/dist-packages/patroni/__init__.py", line 114, in run
postgres-patroni-k8s-0 postgres-patroni-k8s     logger.info(self.ha.run_cycle())
postgres-patroni-k8s-0 postgres-patroni-k8s   File "/usr/local/lib/python2.7/dist-packages/patroni/ha.py", line 1093, in run_cycle
postgres-patroni-k8s-0 postgres-patroni-k8s     info = self._run_cycle()
postgres-patroni-k8s-0 postgres-patroni-k8s   File "/usr/local/lib/python2.7/dist-packages/patroni/ha.py", line 1017, in _run_cycle
postgres-patroni-k8s-0 postgres-patroni-k8s     return self.post_bootstrap()
postgres-patroni-k8s-0 postgres-patroni-k8s   File "/usr/local/lib/python2.7/dist-packages/patroni/ha.py", line 922, in post_bootstrap
postgres-patroni-k8s-0 postgres-patroni-k8s     self.cancel_initialization()
postgres-patroni-k8s-0 postgres-patroni-k8s   File "/usr/local/lib/python2.7/dist-packages/patroni/ha.py", line 917, in cancel_initialization
postgres-patroni-k8s-0 postgres-patroni-k8s     raise PatroniException('Failed to bootstrap cluster')
postgres-patroni-k8s-0 postgres-patroni-k8s patroni.exceptions.PatroniException: 'Failed to bootstrap cluster'
postgres-patroni-k8s-0 postgres-patroni-k8s creating directory /home/postgres/pgdata/pgroot/data ...
```

I created a Helm chart for this. Basically I started reworking the current Patroni chart and aligned it to chart best practices but then switched over and adapted it to this branch. I decided to add it as a separate chart. The chart contains a post-delete hook that removes the configmaps when the release is deleted.

https://github.com/unguiculus/charts/tree/feature/patroni/incubator/patroni-k8s

I played a bit with labels and am not quite sure now the env var `PATRONI_KUBERNETES_LABELS` is used. It seems I'm not completely free in choosing labels. The label `cluster-name` seems to be required.

Since you mentioned you already run it on a few clusters: Do you publish Docker images from this branch in your registry that are publically available and could be used by the chart? I could even create a PR to already publish the chart in the charts repo so people can try it. Would that make sense at this point?

CyberDem0n:
> It runs well using an emptyDir but switching the a persistent volume (which is commented out in (patroni_k8s.yaml) gives me a permissions issue. 

If you are using docker image build from this branch - it will likely not work with persistent volumes.
Container starts with unprivileged postgres user and can't write into a persistent volume.

Helm chart is using a [spilo](https://github.com/zalando/spilo/tree/master/postgres-appliance) image, which starts as root and changes ownership of PGROOT before starting patroni and bunch of other stuff: https://github.com/zalando/spilo/blob/master/postgres-appliance/launch.sh#L6

Actually I've already did some changes to spilo: https://github.com/zalando/spilo/pull/198 to support kubernetes branch, postgres-10 and to make it much smaller. Latest image is: registry.opensource.zalan.do/acid/spilo-10:1.3-p4
You can try to build it on your own, but before that you have to install docker-squash: 
```bash
$ sudo pip install docker-squash
$ ./build.sh --build-arg DEMO=true --build-arg COMPRESS=true -t spilo .
```

Spilo solves a lot of configuration problems for you. It will create the right patroni.yaml with proper kubernetes settings: https://github.com/zalando/spilo/pull/198/files#diff-9ec119124414a288bc2f1edd468dcce5R330

Also it actually prefers to use kubernetes endpoints for leader elections instead of config maps.


> I played a bit with labels and am not quite sure now the env var PATRONI_KUBERNETES_LABELS is used. It seems I'm not completely free in choosing labels. The label cluster-name seems to be required.

Labels are playing two different roles:
1. They are used to find existing objects related to your postgres cluster: (endpoints or configmaps) + pods
2. These labels will be applied to any object Patroni creates (endpoints or configmaps) in order to be able to find those objects later.

In addition to that there Patroni needs to know which label is used to keep cluster name and role.
Such names are configured via `kubernetes.scope_label` and `kubernetes.role_label`.

If `kubernetes.scope_label` is not set, Patroni will use `cluster-name` for that, but in Spilo docker image we are using `version` label: https://github.com/zalando/spilo/pull/198/files#diff-9ec119124414a288bc2f1edd468dcce5R374
If `kubernetes.role_label` is not set, Patroni will use `role` for that, but in Spilo docker we are using `spilo-role`: https://github.com/zalando/spilo/pull/198/files#diff-9ec119124414a288bc2f1edd468dcce5R373

> Since you mentioned you already run it on a few clusters: Do you publish Docker images from this branch in your registry that are publically available and could be used by the chart? I could even create a PR to already publish the chart in the charts repo so people can try it. Would that make sense at this point?

Image is published (see above), but we would really like to make it more "official", i.e. merge this PR, and later merge Spilo PR.

unguiculus:
Thanks, I appreciate the thorough response. Here's my results so far.

I started out from the refactored Helm chart. Commit https://github.com/unguiculus/charts/commit/e9fc62a7807b997bc6e3d3ced691122dc54c3497 works with Etcd. The service gets the correct endpoint. Postgres can be reached via the service.

The latest commit (https://github.com/kubernetes/charts/commit/7b7821f647364ddedfbb1676a924d74769f90a01) updates the chart for Spilo without DCS. I played around with various label options etc. I managed to create a Postgres cluster with working leader election with `KUBERNETES_USE_CONFIGMAPS=true`. The endpoints is updated but, unfortunately, the service is not updated with the endpoint and Postgres cannot be reached. I must be missing something here.

``` 
k describe ep,svc -l app=patroni
Name:         postgres-patroni
Namespace:    default
Labels:       app=patroni
              application=patroni
              chart=patroni-0.6
              cluster=postgres-patroni
              heritage=Tiller
              release=postgres
Annotations:  <none>
Subsets:
  Addresses:          10.36.5.103
  NotReadyAddresses:  <none>
  Ports:
    Name        Port  Protocol
    ----        ----  --------
    postgresql  5432  TCP

Events:  <none>


Name:              postgres-patroni
Namespace:         default
Labels:            app=patroni
                   application=patroni
                   chart=patroni-0.6
                   cluster=postgres-patroni
                   heritage=Tiller
                   release=postgres
Annotations:       <none>
Selector:          <none>
Type:              ClusterIP
IP:                10.39.246.12
Port:              <unset>  5432/TCP
TargetPort:        5432/TCP
Endpoints:
Session Affinity:  None
Events:            <none>
```

Here's the generated `postgres.yml`:

```yaml
bootstrap:
  dcs:
    loop_wait: 10
    maximum_lag_on_failover: 33554432
    postgresql:
      parameters:
        archive_mode: 'on'
        archive_timeout: 1800s
        autovacuum_analyze_scale_factor: 0.02
        autovacuum_max_workers: 5
        autovacuum_vacuum_scale_factor: 0.05
        checkpoint_completion_target: 0.9
        hot_standby: 'on'
        log_autovacuum_min_duration: 0
        log_checkpoints: 'on'
        log_connections: 'on'
        log_disconnections: 'on'
        log_line_prefix: '%t [%p]: [%l-1] %c %x %d %u %a %h '
        log_lock_waits: 'on'
        log_min_duration_statement: 500
        log_statement: ddl
        log_temp_files: 0
        max_connections: 1000
        max_replication_slots: 5
        max_wal_senders: 5
        tcp_keepalives_idle: 900
        tcp_keepalives_interval: 100
        track_functions: all
        wal_keep_segments: 8
        wal_level: hot_standby
        wal_log_hints: 'on'
      use_pg_rewind: true
      use_slots: true
    retry_timeout: 10
    ttl: 30
  initdb:
  - encoding: UTF8
  - locale: en_US.UTF-8
  - data-checksums
  post_init: /post_init.sh "zalandos"
kubernetes:
  labels:
    app: patroni
    application: patroni
    cluster: postgres-patroni
    release: postgres
  namespace: default
  role_label: spilo-role
  scope_label: cluster
postgresql:
  authentication:
    replication:
      password: pinacolada
      username: standby
    superuser:
      password: tea
      username: postgres
  callbacks:
    on_role_change: python3 /callback_endpoint.py
    on_start: python3 /callback_endpoint.py
    on_stop: python3 /callback_endpoint.py
  connect_address: 10.36.0.101:5432
  data_dir: /home/postgres/pgdata/pgroot/data
  listen: 0.0.0.0:5432
  name: postgres-patroni-0
  parameters:
    archive_command: /bin/true
    bg_mon.listen_address: 0.0.0.0
    extwlist.extensions: btree_gin,btree_gist,hstore,intarray,ltree,pgcrypto,pgq,pg_trgm,postgres_fdw,uuid-ossp,hypopg
    log_destination: csvlog
    log_directory: ../pg_log
    log_file_mode: '0644'
    log_filename: postgresql-%u.log
    log_rotation_age: 1d
    log_truncate_on_rotation: 'on'
    logging_collector: 'on'
    shared_buffers: 7539MB
    shared_preload_libraries: bg_mon,pg_stat_statements,pg_cron,set_user,pgextwlist
    ssl: 'on'
    ssl_cert_file: /home/postgres/server.crt
    ssl_key_file: /home/postgres/server.key
  pg_hba:
  - local   all             all                                   trust
  - host    all             all                127.0.0.1/32       md5
  - host    all             all                ::1/128            md5
  - hostssl replication     standby all                md5
  - hostnossl all           all                all                reject
  - hostssl all             all                all                md5
  use_unix_socket: true
restapi:
  connect_address: 10.36.0.101:8008
  listen: 0.0.0.0:8008
scope: postgres-patroni
```

When I set `KUBERNETES_USE_CONFIGMAPS=false` the logs are flooded with this:

```
postgres-patroni-0 spilo 2017-11-23 16:19:49,289 WARNING: Can not get cluster from dcs
postgres-patroni-0 spilo 2017-11-23 16:19:49,312 ERROR: get_cluster
postgres-patroni-0 spilo Traceback (most recent call last):
postgres-patroni-0 spilo   File "/usr/local/lib/python3.5/dist-packages/patroni/dcs/kubernetes.py", line 138, in _load_cluster
postgres-patroni-0 spilo     response = self.retry(self._api.list_namespaced_kind, self._namespace, label_selector=self._label_selector)
postgres-patroni-0 spilo   File "/usr/local/lib/python3.5/dist-packages/patroni/dcs/kubernetes.py", line 97, in retry
postgres-patroni-0 spilo     return self._retry.copy()(*args, **kwargs)
postgres-patroni-0 spilo   File "/usr/local/lib/python3.5/dist-packages/patroni/utils.py", line 259, in __call__
postgres-patroni-0 spilo     return func(*args, **kwargs)
postgres-patroni-0 spilo   File "/usr/local/lib/python3.5/dist-packages/patroni/dcs/kubernetes.py", line 50, in wrapper
postgres-patroni-0 spilo     return getattr(self._api, func)(*args, **kwargs)
postgres-patroni-0 spilo   File "/usr/local/lib/python3.5/dist-packages/kubernetes/client/apis/core_v1_api.py", line 11763, in list_namespaced_endpoints
postgres-patroni-0 spilo     (data) = self.list_namespaced_endpoints_with_http_info(namespace, **kwargs)
postgres-patroni-0 spilo   File "/usr/local/lib/python3.5/dist-packages/kubernetes/client/apis/core_v1_api.py", line 11866, in list_namespaced_endpoints_with_http_info
postgres-patroni-0 spilo     collection_formats=collection_formats)
postgres-patroni-0 spilo   File "/usr/local/lib/python3.5/dist-packages/kubernetes/client/api_client.py", line 321, in call_api
postgres-patroni-0 spilo     _return_http_data_only, collection_formats, _preload_content, _request_timeout)
postgres-patroni-0 spilo   File "/usr/local/lib/python3.5/dist-packages/kubernetes/client/api_client.py", line 163, in __call_api
postgres-patroni-0 spilo     return_data = self.deserialize(response_data, response_type)
postgres-patroni-0 spilo   File "/usr/local/lib/python3.5/dist-packages/kubernetes/client/api_client.py", line 236, in deserialize
postgres-patroni-0 spilo     return self.__deserialize(data, response_type)
postgres-patroni-0 spilo   File "/usr/local/lib/python3.5/dist-packages/kubernetes/client/api_client.py", line 276, in __deserialize
postgres-patroni-0 spilo     return self.__deserialize_model(data, klass)
postgres-patroni-0 spilo   File "/usr/local/lib/python3.5/dist-packages/kubernetes/client/api_client.py", line 620, in __deserialize_model
postgres-patroni-0 spilo     kwargs[attr] = self.__deserialize(value, attr_type)
postgres-patroni-0 spilo   File "/usr/local/lib/python3.5/dist-packages/kubernetes/client/api_client.py", line 254, in __deserialize
postgres-patroni-0 spilo     for sub_data in data]
postgres-patroni-0 spilo   File "/usr/local/lib/python3.5/dist-packages/kubernetes/client/api_client.py", line 254, in <listcomp>
postgres-patroni-0 spilo     for sub_data in data]
postgres-patroni-0 spilo   File "/usr/local/lib/python3.5/dist-packages/kubernetes/client/api_client.py", line 276, in __deserialize
postgres-patroni-0 spilo     return self.__deserialize_model(data, klass)
postgres-patroni-0 spilo   File "/usr/local/lib/python3.5/dist-packages/kubernetes/client/api_client.py", line 622, in __deserialize_model
postgres-patroni-0 spilo     instance = klass(**kwargs)
postgres-patroni-0 spilo   File "/usr/local/lib/python3.5/dist-packages/kubernetes/client/models/v1_endpoints.py", line 64, in __init__
postgres-patroni-0 spilo     self.subsets = subsets
postgres-patroni-0 spilo   File "/usr/local/lib/python3.5/dist-packages/kubernetes/client/models/v1_endpoints.py", line 156, in subsets
postgres-patroni-0 spilo     raise ValueError("Invalid value for `subsets`, must not be `None`")
postgres-patroni-0 spilo ValueError: Invalid value for `subsets`, must not be `None`
```

I may be hitting this issue: https://github.com/kubernetes-incubator/client-python/issues/376. I'm on K8s 1.8.3 on GKE.

CyberDem0n:
Indeed, this is the problem of kubernetes 4.0.0 module.

I've updated requirements.txt and put fixed version there: kubernetes==3.0.0

unguiculus:
I built a new Spilo image as explained above. I hope this was the right thing to do. The result is this:

```
postgres-patroni-0 spilo 2017-11-23 19:39:18,615 - bootstrapping - INFO - Figuring out my environment (Google? AWS? Openstack? Local?)
postgres-patroni-0 spilo 2017-11-23 19:39:18,618 - bootstrapping - DEBUG - Starting new HTTP connection (1): 169.254.169.254
postgres-patroni-0 spilo 2017-11-23 19:39:18,621 - bootstrapping - DEBUG - http://169.254.169.254:80 "GET / HTTP/1.1" 200 22
postgres-patroni-0 spilo 2017-11-23 19:39:18,624 - bootstrapping - DEBUG - Starting new HTTP connection (1): metadata.google.internal
postgres-patroni-0 spilo 2017-11-23 19:39:18,633 - bootstrapping - DEBUG - http://metadata.google.internal:80 "GET /computeMetadata/v1/instance/zone HTTP/1.1" 200 42
postgres-patroni-0 spilo 2017-11-23 19:39:18,634 - bootstrapping - INFO - Looks like your running google
postgres-patroni-0 spilo 2017-11-23 19:39:18,649 - bootstrapping - INFO - Configuring patronictl
postgres-patroni-0 spilo 2017-11-23 19:39:18,651 - bootstrapping - INFO - Writing to file /home/postgres/.config/patroni/patronictl.yaml
postgres-patroni-0 spilo 2017-11-23 19:39:18,651 - bootstrapping - INFO - Configuring certificate
postgres-patroni-0 spilo 2017-11-23 19:39:18,651 - bootstrapping - INFO - Generating ssl certificate
postgres-patroni-0 spilo 2017-11-23 19:39:18,708 - bootstrapping - DEBUG - b"Generating a 2048 bit RSA private key\n...................+++\n...........+++\nwriting new private key to '/home/postgres/server.key'\n-----\n"
postgres-patroni-0 spilo 2017-11-23 19:39:18,708 - bootstrapping - INFO - Configuring patroni
postgres-patroni-0 spilo 2017-11-23 19:39:18,717 - bootstrapping - INFO - Writing to file /home/postgres/postgres.yml
postgres-patroni-0 spilo 2017-11-23 19:39:18,717 - bootstrapping - INFO - Configuring pam-oauth2
postgres-patroni-0 spilo 2017-11-23 19:39:18,717 - bootstrapping - INFO - No PAM_OAUTH2 configuration was specified, skipping
postgres-patroni-0 spilo 2017-11-23 19:39:19,133 INFO: Failed to import patroni.dcs.consul
postgres-patroni-0 spilo 2017-11-23 19:39:19,134 INFO: Failed to import patroni.dcs.etcd
postgres-patroni-0 spilo 2017-11-23 19:39:19,135 INFO: Failed to import patroni.dcs.exhibitor
postgres-patroni-0 spilo 2017-11-23 19:39:19,339 INFO: Lock owner: None; I am postgres-patroni-0
postgres-patroni-0 spilo 2017-11-23 19:39:19,354 INFO: trying to bootstrap a new cluster
postgres-patroni-0 spilo 2017-11-23 19:39:19,354 ERROR: Exception during execution of long running task bootstrap
postgres-patroni-0 spilo Traceback (most recent call last):
postgres-patroni-0 spilo   File "/usr/local/lib/python3.5/dist-packages/patroni/async_executor.py", line 86, in run
postgres-patroni-0 spilo     wakeup = func(*args) if args else func()
postgres-patroni-0 spilo   File "/usr/local/lib/python3.5/dist-packages/patroni/postgresql.py", line 1471, in bootstrap
postgres-patroni-0 spilo     return do_initialize(config) and self._configure_server_parameters() and self.start()
postgres-patroni-0 spilo   File "/usr/local/lib/python3.5/dist-packages/patroni/postgresql.py", line 523, in _initdb
postgres-patroni-0 spilo     ret = self.pg_ctl('initdb', *options)
postgres-patroni-0 spilo   File "/usr/local/lib/python3.5/dist-packages/patroni/postgresql.py", line 269, in pg_ctl
postgres-patroni-0 spilo     return subprocess.call(pg_ctl + ['-D', self._data_dir] + list(args), **kwargs) == 0
postgres-patroni-0 spilo   File "/usr/lib/python3.5/subprocess.py", line 557, in call
postgres-patroni-0 spilo     with Popen(*popenargs, **kwargs) as p:
postgres-patroni-0 spilo   File "/usr/lib/python3.5/subprocess.py", line 947, in __init__
postgres-patroni-0 spilo     restore_signals, start_new_session)
postgres-patroni-0 spilo   File "/usr/lib/python3.5/subprocess.py", line 1551, in _execute_child
postgres-patroni-0 spilo     raise child_exception_type(errno_num, err_msg)
postgres-patroni-0 spilo FileNotFoundError: [Errno 2] No such file or directory: 'pg_ctl'
postgres-patroni-0 spilo 2017-11-23 19:39:19,377 INFO: removing initialize key after failed attempt to bootstrap the cluster
postgres-patroni-0 spilo Traceback (most recent call last):
postgres-patroni-0 spilo   File "/usr/local/bin/patroni", line 11, in <module>
postgres-patroni-0 spilo     load_entry_point('patroni==1.3.6', 'console_scripts', 'patroni')()
postgres-patroni-0 spilo   File "/usr/local/lib/python3.5/dist-packages/patroni/__init__.py", line 174, in main
postgres-patroni-0 spilo     return patroni_main()
postgres-patroni-0 spilo   File "/usr/local/lib/python3.5/dist-packages/patroni/__init__.py", line 143, in patroni_main
postgres-patroni-0 spilo     patroni.run()
postgres-patroni-0 spilo   File "/usr/local/lib/python3.5/dist-packages/patroni/__init__.py", line 114, in run
postgres-patroni-0 spilo     logger.info(self.ha.run_cycle())
postgres-patroni-0 spilo   File "/usr/local/lib/python3.5/dist-packages/patroni/ha.py", line 1093, in run_cycle
postgres-patroni-0 spilo     info = self._run_cycle()
postgres-patroni-0 spilo   File "/usr/local/lib/python3.5/dist-packages/patroni/ha.py", line 1017, in _run_cycle
postgres-patroni-0 spilo     return self.post_bootstrap()
postgres-patroni-0 spilo   File "/usr/local/lib/python3.5/dist-packages/patroni/ha.py", line 922, in post_bootstrap
postgres-patroni-0 spilo     self.cancel_initialization()
postgres-patroni-0 spilo   File "/usr/local/lib/python3.5/dist-packages/patroni/ha.py", line 917, in cancel_initialization
postgres-patroni-0 spilo     raise PatroniException('Failed to bootstrap cluster')
postgres-patroni-0 spilo patroni.exceptions.PatroniException: 'Failed to bootstrap cluster'
```

Let me know if this is not the right place to further discuss this. I don't want to abuse this PR. Would you be willing to further assist me? I could create an issue or you can find me on Kubernetes Slack.

CyberDem0n:
> I built a new Spilo image as explained above. I hope this was the right thing to do. The result is this:
postgres-patroni-0 spilo FileNotFoundError: [Errno 2] No such file or directory: 'pg_ctl'

Oh, I've told you to build it with `--build-arg DEMO=true` :(

"--build-arg DEMO=true" is used to build an image with postgres 10 only and without a lot of heavy stuff. It is good enough to try it with minikube for example, but it is not for production, because there is wal-e inside.

And you hit a bug that $PATH wasn't propagated to patroni and it failed to run pg_ctl initdb.
Anyway, bug in spilo is fixed and it should work.

Normal spilo image contains postgres 9.3, 9.4, 9.5, 9.6 and 10.
By default it will start postgres 10, but you can change it with environment variable: `SPILO_CONFIGURATION='{postgresql: {bin_dir: /usr/lib/postgresql/9.6/bin}'`
Basically it is possible to supply any part of patroni configuration in the `SPILO_CONFIGURATION` and it would be "merged" into config generated by spilo.

unguiculus:
I built the new image without demo and tried it with Postgres 10 and 9.6. Everything starts up fine with now with both, `KUBERNETES_USE_CONFIGMAPS=true` and `KUBERNETES_USE_CONFIGMAPS=false`. Leader election works but the service is still not updated with the endpoint.

CyberDem0n:
> Leader election works but the service is still not updated with the endpoint.

I think I know what the problem is. At some moment we started creating Services with a "named" port:
```yaml
apiVersion: v1
kind: Service
metadata:
  name: &cluster_name patronidemo
  labels:
    application: spilo
    version: *cluster_name
spec:
  type: ClusterIP
  ports:
  - port: 5432
    targetPort: 5432
    name: postgresql # XXX
```

This commit explain why: https://github.com/zalando/spilo/commit/2be341ab026cef81982ca55b4328ea60aecb13f8#diff-8c54fa1e5677a832585d18f396619701

If name of Service and Endpoint doesn't match - service will not work.
This is configured in spilo: https://github.com/zalando/spilo/pull/198/files#diff-9ec119124414a288bc2f1edd468dcce5R377

I think in your case you've created Service with no name assigned for port=5432

unguiculus:
Thanks, that was the missing piece. It works for both modes (`KUBERNETES_USE_CONFIGMAPS=true|false`). Which mode would be the preferred one? I noticed that it is important to clean up. If you install it and an old endpoint or configmap still exists, the cluster will form correctly, but the service doesn't get the endpoint. Maybe this can be improved. I updated my Helm chart. It now removes both, configmaps and endpoints, in a post-delete hook.

unguiculus:
I also experimented with parallel pod management and rolling updates. Do you see a problem with this?

https://github.com/unguiculus/charts/commit/dd94767d70f94019a3243d7c251f47303a0a59b5

CyberDem0n:
> Which mode would be the preferred one?

Endpoints. Otherwise there is a race condition: https://github.com/zalando/patroni/issues/536

> I noticed that it is important to clean up. If you install it and an old endpoint or configmap still exists, the cluster will form correctly, but the service doesn't get the endpoint. Maybe this can be improved.

I don't think that it will "form correctly". Patroni will notice that there is a `<cluster-name>-config` endpoint or configmap, assume that cluster was already initialized and will not run initdb but try to restore from a backup (with wal-e) or all pods wait until there will be a leader to get basebackup from it.

Database clusters are stateful and usually you don't delete-create them all the time. There is not much we can improve.


> I also experimented with parallel pod management and rolling updates. Do you see a problem with this?

I've never played with rolling upgrades so far, but I think it might be dangerous. It is very important to not terminate the next pod until the previous one become healthy enough (started streaming from master and replication lag close to 0). 
Actually we are not really using helm charts, but [postgres operator](https://github.com/zalando-incubator/postgres-operator) for everything: cluster creation and removal, user management, PV management, rolling upgrades and so on.

unguiculus:
I've seen the operator project and was wondering which way to go. I'm looking for a solution without an additional DCS like Etcd. Is this also possible with the operator?

CyberDem0n:
Postgres-operator doesn't care what DCS is used by Patroni. It just passes some environment variables to Spilo.

jberkus:
So, some testing questions;

1. Does the main spilo image now support running as kube-native?  If so, how do I turn this behavior on?  If not, which image do I need to use?

2. Where do I set KUBERNETES_USE_CONFIGMAP?  is that an ENV in the pod definition?   If so, it doesn't seem to have an effect.

3. @ants did you get some version of this working on OpenShift?

CyberDem0n:
The latest spilo image supporting kube-native is `registry.opensource.zalan.do/acid/spilo-10:1.3-p6`.
It is build from this branch: https://github.com/zalando/spilo/pull/198, which is not yet merged into master.

In order to enable Kubernetes api for leader election you should set `DCS_ENABLE_KUBERNETES_API` env. By default it is using `Endpoints`. It eliminate a lot of problems and race conditions, because `subsets` for the leader endpoint are set at the same time when leader lock is acquired/updated. If you want to use config maps for leader election, you have to set `KUBERNETES_USE_CONFIGMAP` env. Basically all "magic" of generating patroni config yaml is hidden here: https://github.com/zalando/spilo/pull/198/files#diff-9ec119124414a288bc2f1edd468dcce5R399 and here: https://github.com/zalando/spilo/pull/198/files#diff-9ec119124414a288bc2f1edd468dcce5R442

jberkus:
OK, so KUBERNETES_USE_CONFIGMAP is in spilo but not in upstream patroni?

CyberDem0n:
Yes, `KUBERNETES_USE_CONFIGMAP` is in Spilo. Somehow in Spilo it is reverted in comparison to Patroni. Because it Patroni I wanted to make a safe choice.

In Patroni kubernetes configs are different:
* `kubernetes.use_endpoints` in yaml or `PATRONI_KUBERNETES_USE_ENDPOINTS` in env
* `kubernetes.namespace` in yaml or `PATRONI_KUBERNETES_NAMESCAPE` in env -- kubernetes namespace where we are running
* `kubernetes.labels` in yaml or `PATRONI_KUBERNETES_LABELS` in env -- These labels will be used to find existing objects (Endpoints|ConfigMaps + Pods) associated with the current cluster. Also Patroni will set them on every object (Endpoint|ConfigMap) it creates/updates.
* `kubernetes.scope_label` in yaml or `PATRONI_KUBERNETES_SCOPE_LABEL` in env -- name of the label containing cluster name. Default value is `cluster-name`
* `kubernetes.role_label` in yaml or `PATRONI_KUBERNETES_ROLE_LABEL` in env -- name of the label containing role (master or replica). Patroni will set this label on the pod it runs in. Default value is `role`
* `kubernetes.pod_ip` in yaml or `PATRONI_KUBERNETES_POD_IP` in env -- ip of the pod we are running in. It's only necessary when `use_endpoints` is enabled, to write this ip into the leader endpoint subsets.
* `kubernetes.ports` in yaml or `PATRONI_KUBERNETES_PORTS` in env -- if the Service object has the name for the port, the same name must appear in the Endpoint object, otherwise service wont work. Example: `{kund: Service, spec: {ports: [{name: postgresql, port: 5432, targetPort: 5432}]}}`. In this case you have to define `kubernetes.ports`: `{[{"name": "postgresql", "port": 5432}]}` and Patroni will use it for updating subsets of a leader Endpoint. This parameter is not used if `use_endpoints` is not set.

jberkus:
@CyberDem0n I'd like to have it in patroni, so that I can run it on OpenShift (see @ants above).  I can't run Spilo on OpenShift because it violates OpenShift's permissions restrictions.

hjacobs:
This is AFAIK already running in production, can we merge this PR?

ants:
@jberkus - yes I got a version of this working with OpenShift. I actually did use heavily modified Spilo because I wanted the archiving and backup features from there. There is plenty of issues to resolve when building the image, for example initdb will fail because getpwnam() doesn't work because the container user does not exist, everything needs to run as the fake root user, anything that will be modified from the container needs g+rw permissions set on it, setuid will not work at all, so cron daemon needs to be replaced.

I used the config map based approach, the race condition is an acceptable risk for now.

I haven't tried to integrate the latest version, but that is just because I have been busy with other tasks.

alexeyklyukin:
👍

jberkus:
@CyberDem0n 

To date, this branch has done well in all of my testing.  I have yet to hit a specific bug with it.

@ants  

See follow-up issue ...

CyberDem0n:
:+1:

CyberDem0n:
* Use ConfigMaps or Endpoins for leader elections and to keep cluster state
* Label pods with a postgres role
* change behavior of pip install. From now on it will not install all dependencies, you have to specify explicitly DCS you want to use Patroni with: `pip install patroni[etcd,zookeeper,kubernetes]`

PR #502, opened 2017-08-11
"Could not take out TTL lock" message was never logged

CyberDem0n:
This is not a critical bug, because `attempt_to_acquire_leader` method was still returning False in this case.


CyberDem0n:
:+1:

alexeyklyukin:
👍

Issue #504, opened 2017-08-11
Demoted node does not start on fail over when using synchronous mode


akissa:
I am testing 3 nodes in a cluster with  synchronous mode.

```
patronictl -c /etc/patroni/baruwa.yml list baruwa
+---------+------------------------------+--------------+--------+---------+-----------+
| Cluster | Member                       | Host         |  Role  |  State  | Lag in MB |
+---------+------------------------------+--------------+--------+---------+-----------+
| baruwa  | db.home.topdog-software.com  | 192.168.1.15 | Leader | running |       0.0 |
| baruwa  | db2.home.topdog-software.com | 192.168.1.33 |        | running |       0.0 |
| baruwa  | db3.home.topdog-software.com | 192.168.1.34 |        | running |       0.0 |
+---------+------------------------------+--------------+--------+---------+-----------+
```

When i trigger a manual failover using `patronictl -c /etc/patroni/baruwa.yml failover baruwa` The fail over happens and a different host takes over as the leader. However the previous leader is not able to startup postgresql as the `synchronous_standby_names` option is not removed from `postgresql.conf` file.

If i stop the patroni process and start it up again, the `postgresql.conf` file is updated removing the `synchronous_standby_names` option allowing postgresql to startup and rejoin the cluster.

CyberDem0n:
Hi @akissa, thank you for your finding.
It's very interesting use-case we didn't thought about :(
Will prepare a fix soon.

CyberDem0n:
Actually problem is even worse than that.
Synchronous replication didn't worked completely in your case.

https://www.postgresql.org/docs/9.6/static/runtime-config-replication.html#RUNTIME-CONFIG-REPLICATION-MASTER tells:
```
Note: Each standby_name should have the form of a valid SQL identifier, unless it is *.
You can use double-quoting if necessary.
```

Since your member names containing periods and dashes, they must me quoted.

uspen:
Yes, names "Member" with sync standby MUST be SQL syntax. I use
hostname: nodeXX.db.example.com
Member in Patroni: nodexXX

CyberDem0n:
@uspen, yep, but https://github.com/zalando/patroni/pull/505 will fix it and will make it possible to use nodeXX.db.example.com as member name

akissa:
@CyberDem0n #505 does fix the issue for me.

akissa:
@CyberDem0n Thank you.

PR #505, opened 2017-08-12
synchronous_standby_names must be quoted with quote_ident

CyberDem0n:
in addition to that implement additional checks around manual failover and recover when synchronous_mode is enabled

Fixes https://github.com/zalando/patroni/issues/504

### DIFF
--- a/patroni/postgresql.py
+++ b/patroni/postgresql.py
@@ -17,7 +17,7 @@ from contextlib import contextmanager
 from patroni import call_self
 from patroni.callback_executor import CallbackExecutor
 from patroni.exceptions import PostgresConnectionException
-from patroni.utils import compare_values, parse_bool, parse_int, Retry, RetryFailedError, polling_loop, null_context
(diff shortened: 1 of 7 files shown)


patroni/postgresql.py
```diff
@@ -42,6 +42,12 @@
 }
 STOP_POLLING_INTERVAL = 1
 REWIND_STATUS = type('Enum', (), {'INITIAL': 0, 'CHECK': 1, 'NEED': 2, 'NOT_NEED': 3, 'SUCCESS': 4, 'FAILED': 5})
+sync_standby_name_re = re.compile('^[A-Za-z_][A-Za-z_0-9\$]*$')
+
+
+def quote_ident(value):
+    """Very simplified version of quote_ident"""
+    return value if sync_standby_name_re.match(value) else '"' + value + '"'
```
CyberDem0n: 1. It requires psycopg2>=2.7.0
2. This function needs existing connection to postgres
3. We don't need all the "power" of "quote_ident"

CyberDem0n:
:+1:

alexeyklyukin:
👍

PR #507, opened 2017-08-13
Different bugfixes around pause state, mostly related to watchdog

CyberDem0n:
* Do not send keepalives if watchdog is not active
* Avoid activating watchdog in a pause mode
* Set correct postgres state in pause mode
* Don't try to run queries from API if postgres is stopped


ants:
Looks good to me.

CyberDem0n:
:+1:

alexeyklyukin:
👍

PR #513, opened 2017-09-04
Pass the consul token as a header

akissa:
Headers are now the prefered way to pass the token to the consul
API - https://www.consul.io/api/index.html#authentication

### DIFF
--- a/patroni/dcs/consul.py
+++ b/patroni/dcs/consul.py
@@ -83,8 +83,7 @@ class HTTPClient(object):
             else:
                 kwargs['timeout'] = self._read_timeout
             if isinstance(params, dict) and 'token' in params and params['token']:
-                kwargs['headers'] = {'X-Consul-Token': params['token']}
-                del params['token']
+                kwargs['headers'] = {'X-Consul-Token': params.pop('token')}
             return callback(self.response(self.http.request(method.upper(), self.uri(path, params), **kwargs)))
         return wrapper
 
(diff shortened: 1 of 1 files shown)


patroni/dcs/consul.py
```diff
@@ -82,6 +82,9 @@ def wrapper(callback, path, params=None, data=''):
                 kwargs['timeout'] = (float(params['wait'][:-1]) if 'wait' in params else 300) + 1
             else:
                 kwargs['timeout'] = self._read_timeout
+            if isinstance(params, dict) and 'token' in params and params['token']:
+                kwargs['headers'] = {'X-Consul-Token': params['token']}
```
CyberDem0n: lines 85-86 could be rewritten as:
```python
kwargs['headers'] = {'X-Consul-Token': params.pop('token')}
```

My main concern was about breaking compatibility with the old Consul versions, but it seems that 'X-Consul-Token' support was added in version 0.6.0, released on December 3, 2015, so probably it should be fine.

akissa: Using `.pop` could generate an attribute error, as it is possible for `params` to be `None`

CyberDem0n: Previous line guaranties that 'token' is defined and even non empty

akissa: Got you, so the change should be on line 86 then. Will update

CyberDem0n:
:+1:

outime:
:+1:

alexeyklyukin:
👍

CyberDem0n:
Merged, thank you @akissa

Issue #517, opened 2017-09-11
Testsuite does not handle nonexisting home directory during pgpass write


mbanck:
If `$HOME` is set to a non-existing directory (which would e.g. be the case on an official Debian package autobuilder) the following tests fail:

```
=================================== FAILURES ===================================
______________________ TestPostgresql.test_create_replica ______________________
__________________________ TestPostgresql.test_rewind __________________________
_________________ TestPostgresql.test_run_bootstrap_post_init __________________
```

The culprit is each time:

```
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

self = <patroni.postgresql.Postgresql object at 0x7f9fa04497f0>
record = {'connect_timeout': 3, 'database': 'postgres', 'fallback_application_name': 'Patroni', 'host': '127.0.0.2', ...}

    def write_pgpass(self, record):
        if 'user' not in record or 'password' not in record:
            return os.environ.copy()
    
>       with open(self._pgpass, 'w') as f:
E       FileNotFoundError: [Errno 2] No such file or directory: '/sbuild-nonexistent/pgpass'

patroni/postgresql.py:607: FileNotFoundError
```
Now I see that `test_config.py` sets `'PATRONI_POSTGRESQL_PGPASS': '/tmp/pgpass0',`, so this apparently works in some (most?) cases and/or was thought of, but maybe the above three cases need fixing. Unfortunately, it was not trivially obvious to me what needed changing, so filing an issue here.

CyberDem0n:
Do I understand correctly that you want to package Patroni into deb?

mbanck:
Yes, that is also the plan, do you have issues with it or any other useful comments, maybe?

CyberDem0n:
No issues, just thinking how you want to handle not existing packages of some python modules required by Patroni.

For example there is no packages of `cdiff` and `python-consul` for ubuntu16.

`python-etcd` seems to be outdated (0.4.2).

mbanck:
There's `python3-consul` now (at least in Debian), so I've done a python3-only build, which appears to pass all the tests except the ones in this issue.

About `python-etcd`, Debian has 0.4.3 now which seems to be the minimum as defined by `requirements.txt`.

And I just uploaded `cdiff` a couple of minutes ago.

PR #518, opened 2017-09-12
Set pgpass explicitly to /tmp/pgpass0 when running unit-tests

CyberDem0n:
If $HOME is set to a non-existing directory (which would e.g. be the case on an official Debian package autobuilder) some tests were failing.

Fixes https://github.com/zalando/patroni/issues/517


alexeyklyukin:
👍

CyberDem0n:
:+1:

PR #524, opened 2017-09-18
Fix pg_rewind behaviour

CyberDem0n:
When Patroni does calculation whether it should run pg_rewind or not, it relies on pg_controldata output or gets necessary information from replication connection.
In some cases (when for example postgres running as a master was killed), we can't use pg_controldata output immediately, but trying to start postgres. Such start could fail with the following errror:
```
2017-09-18 08:00:38.659 UTC,,,55,,59bf7d26.37,1,,2017-09-18 08:00:38 UTC,,0,LOG,00000,"ending log output to stderr",,"Future log output will go to log destination ""csvlog"".",,,,,,,""
2017-09-18 08:00:38.661 UTC,,,57,,59bf7d26.39,1,,2017-09-18 08:00:38 UTC,,0,LOG,00000,"database system was interrupted; last known up at 2017-09-16 22:35:22 UTC",,,,,,,,,""
2017-09-18 08:00:39.012 UTC,,,57,,59bf7d26.39,2,,2017-09-18 08:00:38 UTC,,0,LOG,00000,"restored log file ""00000006.history"" from archive",,,,,,,,,""
2017-09-18 08:00:39.222 UTC,,,57,,59bf7d26.39,3,,2017-09-18 08:00:38 UTC,,0,LOG,00000,"entering standby mode",,,,,,,,,""
2017-09-18 08:00:39.433 UTC,,,57,,59bf7d26.39,4,,2017-09-18 08:00:38 UTC,,0,LOG,00000,"restored log file ""00000006.history"" from archive",,,,,,,,,""
2017-09-18 08:00:39.437 UTC,,,57,,59bf7d26.39,5,,2017-09-18 08:00:38 UTC,,0,FATAL,XX000,"requested timeline 6 is not a child of this server's history","Latest checkpoint is at 29/1A000178 on timeline 5, but in the history of the requested timeline, the server forked off from that timeline at 29/1A000140.",,,,,,,,""
2017-09-18 08:00:39.439 UTC,,,55,,59bf7d26.37,2,,2017-09-18 08:00:38 UTC,,0,LOG,00000,"startup process (PID 57) exited with exit code 1",,,,,,,,,""
2017-09-18 08:00:39.439 UTC,,,55,,59bf7d26.37,3,,2017-09-18 08:00:38 UTC,,0,LOG,00000,"aborting startup due to startup process failure",,,,,,,,,""
2017-09-18 08:00:39.441 UTC,,,55,,59bf7d26.37,4,,2017-09-18 08:00:38 UTC,,0,LOG,00000,"database system is shut down",,,,,,,,,""
```
After that it's not possible to start such cluster in a normal mode, only single-user is possible.

The second problems is: if postgres was running as master, pg_controldata reports:
```
Database cluster state:               shut down in recovery
Minimum recovery ending location:     0/0
Min recovery ending loc's timeline:   0
```

And this info can't be used for calculations. In this case we should use `Latest checkpoint location` and `Latest checkpoint's TimeLineID`

### DIFF
--- a/patroni/postgresql.py
+++ b/patroni/postgresql.py
@@ -808,6 +808,10 @@ class Postgresql(object):
             if not self.is_pid_running(pid):
                 logger.error('postmaster is not running')
                 self.set_state('start failed')
+                # somehow we observed multiple times that there were no postmaster running, but
+                # pg_controldata was still reporting: `Database cluster state: in archive recovery`
+                # Therefore we will try to give some more time to all processes to finish.
+                time.sleep(1)
                 return False
 
         logger.warning("Timed out waiting for PostgreSQL to start")
@@ -1261,12 +1265,14 @@ class Postgresql(object):
         else:  # otherwise analyze pg_controldata output
             data = self.controldata()
             try:
+                if data.get('Database cluster state') == 'shut down in recovery':
+                    lsn = data.get('Minimum recovery ending location')
+                    timeline = int(data.get("Min recovery ending loc's timeline"))
+                    if lsn == '0/0' or timeline == 0:  # it was a master when it crashed
+                        data['Database cluster state'] = 'shut down'
                 if data.get('Database cluster state') == 'shut down':
                     lsn = data.get('Latest checkpoint location')
                     timeline = int(data.get("Latest checkpoint's TimeLineID"))
-                elif data.get('Database cluster state') == 'shut down in recovery':
-                    lsn = data.get('Minimum recovery ending location')
-                    timeline = int(data.get("Min recovery ending loc's timeline"))
             except (TypeError, ValueError):
                 logger.exception('Failed to get local timeline and lsn from pg_controldata output')
         logger.info('Local timeline=%s lsn=%s', timeline, lsn)
--- a/patroni/postgresql.py
+++ b/patroni/postgresql.py
@@ -808,10 +808,6 @@ class Postgresql(object):
             if not self.is_pid_running(pid):
                 logger.error('postmaster is not running')
                 self.set_state('start failed')
-                # somehow we observed multiple times that there were no postmaster running, but
-                # pg_controldata was still reporting: `Database cluster state: in archive recovery`
-                # Therefore we will try to give some more time to all processes to finish.
-                time.sleep(1)
                 return False
 
         logger.warning("Timed out waiting for PostgreSQL to start")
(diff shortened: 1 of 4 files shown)


a1exsh:
:+1:

patroni/postgresql.py
```diff
@@ -808,6 +808,10 @@ def wait_for_port_open(self, pid, initiated, timeout):
             if not self.is_pid_running(pid):
                 logger.error('postmaster is not running')
                 self.set_state('start failed')
+                # somehow we observed multiple times that there were no postmaster running, but
+                # pg_controldata was still reporting: `Database cluster state: in archive recovery`
+                # Therefore we will try to give some more time to all processes to finish.
+                time.sleep(1)
```
CyberDem0n: Unfortunately it doesn't help.
The problem arises only when there are archive_command and recovery_command used in addition to streaming replication.
Or we can rephrase even more strict, recovery_command is a problem and breaks everything...

patroni/postgresql.py
```diff
@@ -1744,3 +1746,57 @@ def postgres_major_version_to_int(pg_version):
...
+                    logger.exception('Unable to remove %s', path)
+        except OSError:
+            logger.exception('Unable to list %s', status_dir)
+
+    def fix_cluster_state(self):
+        self.cleanup_archive_status()
+
+        # Start in a single user mode and stop to produce a clean shutdown
+        opts = self.read_postmaster_opts()
+        opts.update({'archive_mode': 'on', 'archive_command': 'false'})
```
a1exsh: `/bin/false`?

patroni/postgresql.py
```diff
@@ -1744,3 +1746,57 @@ def postgres_major_version_to_int(pg_version):
...
+        if os.path.isfile(self._recovery_conf) or os.path.islink(self._recovery_conf):
+            os.unlink(self._recovery_conf)
+        return self.single_user_mode(options=opts) == 0 or None
```
a1exsh: Why `or None` here?  Is `True`/`False` not good enough?

a1exsh:
:+1:

patroni/postgresql.py
CyberDem0n: Yes, `/bin/false`, but nobody can guaranty that it is really located inside `/bin`, therefore it's always better to rely on $PATH.

patroni/postgresql.py
CyberDem0n: This method is executed asynchronously.
If it returns something different from None, it will cause main HA loop to wake up immediately.

CyberDem0n:
:+1:

Issue #522, opened 2017-09-15
Consul: lock is lost on 7-10 s network issues


Vanav:
I have Patroni and Consul Server on the same server db1 (192.168.241.17), and the same setup for slave server db2 (192.168.241.18). db1 has a lock. If I have short network issues on db1, then Consul after 7 seconds removes this node from the list, and deletes all the locks, and failover happens.

How can I improve the situation and increase timeout to, say, 30-60 s?

I've [asked question in Consul maillist](https://groups.google.com/forum/#!topic/consul-tool/YMa4J40qgM0), and they suggested to associate Session with a TTL based check, and not to bind to “serfHealth”. So, lock will not be lost, if node db1 will be removed.

Consul on db2:
```
2017/09/13 17:29:15 [ERR] memberlist: Failed fallback ping: read tcp 192.168.241.18:34928->192.168.241.17:8301: i/o timeout
2017/09/13 17:29:15 [INFO] memberlist: Suspect db1.aes128.com has failed, no acks received
2017/09/13 17:29:18 [INFO] memberlist: Marking db1.aes128.com as failed, suspect timeout reached (2 peer confirmations)
2017/09/13 17:29:18 [INFO] serf: EventMemberFailed: db1.aes128.com 192.168.241.17
2017/09/13 17:29:18 [INFO] consul: Removing LAN server db1.aes128.com (Addr: tcp/192.168.241.17:8300) (DC: dc1)
2017/09/13 17:29:19 [ERR] memberlist: Failed fallback ping: read tcp 192.168.241.18:34941->192.168.241.17:8301: i/o timeout
2017/09/13 17:29:19 [INFO] memberlist: Suspect db1.aes128.com has failed, no acks received
2017/09/13 17:29:24 [INFO] serf: EventMemberLeave (forced): db1.aes128.com 192.168.241.17
2017/09/13 17:29:24 [INFO] consul: Removing LAN server db1.aes128.com (Addr: tcp/192.168.241.17:8300) (DC: dc1)
```

Patroni on db2:
```
2017-09-13 17:29:14,488 INFO: Lock owner: db1; I am db2
2017-09-13 17:29:14,488 INFO: does not have lock
2017-09-13 17:29:14,504 INFO: no action.  i am a secondary and i am following a leader
2017-09-13 17:29:20,761 WARNING: request failed: GET http://192.168.241.17:8008/patroni (HTTPConnect
ionPool(host='192.168.241.17', port=8008): Read timed out. (read timeout=2))
[...]

PR #543, opened 2017-10-12
Make it possible to change Consul session checks

CyberDem0n:
If list of checks is not specified, Consul will use "serfHealth" in addition to TTL based created by Patroni.
There are some cases when people want to sacrifice fast detection of network partitioning in favor of ability to tolerate network lags.

Fixes https://github.com/zalando/patroni/issues/522


alexeyklyukin:
Given that the serf check invalidates our promise of holding the leader lock for exactly TTL before releasing it shouldn't we supply an empty list of checks (thus disabling the serf check) by default?

CyberDem0n:
I just don't want to change default behavior. So far only one person complain about it, so I think it should be fine to leave it as it is and give some controls to change it.

alexeyklyukin:
Well, I think we should at least make users aware of that aspect of Consul. Added a sentence to the docs.

alexeyklyukin:
👍

CyberDem0n:
:+1:

PR #549, opened 2017-10-24
Verify process start time when checking if postgres is running.

### LINKED ISSUE #548: is_pid_running() needs to check if the process is actually postgres
Debugged a system where async executor was stuck in `_wait_for_postmaster_stop()`. There was a postmaster.pid left over from before power loss, the pid it contained was the same as the async executor threads PID, but different from Patroni pid. This means that the pid didn't match os.getpid(), but os.kill still succeeded.

I think we should also compare the start time in the pid file with the start time of the running process.

ants:
After a crash that doesn't clean up postmaster.pid there could be a new process with the same pid resulting in a false positive for is_running(), which will lead to all kinds of bad behavior.

Fixes #548 

Not sure how portable the implementation is, needs some research. Tests are also missing, if the general approach seems fine then I will figure out how to test this code.

### DIFF
--- a/patroni/postgresql.py
+++ b/patroni/postgresql.py
@@ -17,7 +17,7 @@ from contextlib import contextmanager
 from patroni import call_self
 from patroni.callback_executor import CallbackExecutor
 from patroni.exceptions import PostgresConnectionException
-from patroni.utils import compare_values, parse_bool, parse_int, Retry, RetryFailedError, polling_loop
+from patroni.utils import compare_values, parse_bool, parse_int, Retry, RetryFailedError, polling_loop, int_or_none
 from six import string_types
 from six.moves.urllib.parse import quote_plus
 from threading import current_thread, Lock
@@ -707,20 +707,9 @@ class Postgresql(object):
         if not (self._version_file_exists() and os.path.isfile(self._postmaster_pid)):
             # XXX: This is dangerous in case somebody deletes the data directory while PostgreSQL is still running.
             return False
+
         pidfile = self.read_pid_file()
-        if 'pid' not in pidfile:
-            return False
-
-        try:
-            proc = psutil.Process(int(pidfile['pid']))
-        except (ValueError, psutil.NoSuchProcess):
-            return False
-
-        # If start time differs, then it's the wrong process. Allow for 3 second difference for rounding errors.
-        if 'start_time' in pidfile and abs(proc.create_time() - int(pidfile['start_time'])) > 3:
-            return False
-
-        return proc.is_running()
+        return self._is_postmaster_pid_running(int_or_none(pidfile.get('pid')), start_time=int_or_none(pidfile.get('start_time')))
 
     def read_pid_file(self):
         """Reads and parses postmaster.pid from the data directory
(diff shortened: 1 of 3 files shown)


patroni/postgresql.py
```diff
@@ -707,7 +707,20 @@ def is_running(self):
         if not (self._version_file_exists() and os.path.isfile(self._postmaster_pid)):
             # XXX: This is dangerous in case somebody deletes the data directory while PostgreSQL is still running.
             return False
-        return self.is_pid_running(self.get_pid())
+        pidfile = self.read_pid_file()
```
alexeyklyukin: is there any reason for not moving the new checks inside `is_pid_running()` instead of adding them to `is_running()`?

mgomezch: I can confirm this is also required in `is_pid_running` to avoid getting stuck in `patroni/postgresql.py#Postgresql._wait_for_postmaster_stop` when some other existing process happens to have the same PID as written in a now defunct postmaster's `postmaster.pid` file.

ants: Not a really good reason. By doing it in is_running() I avoided reading the pid file twice. As far as I can tell the places where is_pid_running() is called currently don't care about this specific failure case. When starting we already know we have the correct pid, when stopping we first check for the correct process via is_running() and when aborting a restart postmaster was supposed to be running before restart was initiated so the pid file should have been correct.

I could rejiggle the code by adding an optional start_time parameter to is_pid_running() if you think that is nicer.

We discussed this with @CyberDem0n at pgconf.eu. I had concerns about the start time check possibly not being robust in face of system clock changes. He proposed that we look at the command name instead. That would fail when somebody runs a postgres instance as the same user (otherwise kill -0 permission check fails), but in a different data directory. I'll test how start time behaves on Linux if system clock is screwed with and will report back.

CyberDem0n: @ants indeed, something feels strange.

#548 tells that async executor stuck in `_wait_for_postmaster_stop`, but this PR does not really change it and `_wait_for_postmaster_stop` still calls `is_pid_running`.

I am pretty sure, it will fix exactly your problem, when you start patroni after power loss, but it will not fix another problem when OS will create a new process with the same PID as postmaster had.

And regarding process name check: we always start postmaster by calling "postgres" binary.
In this case process name always matches with name of the binary (at least on linux).

P.S. can you also have a look on https://github.com/zalando/patroni/pull/553 ?

ants: `_wait_for_postmaster_stop` is called from `_do_stop`, which does `is_running()` check at the beginning, catching the stale pid file case. You're right that there is still a race condition if postgres terminates while we are waiting and kernel reassigns its pid to a process of the same user within the polling interval. I think moving the start time check into `is_pid_running()` is probably a good idea. I will do that.

With the name check I'm now not worried about false negative (postgres running as something else than postgres), but about false positive (postgres that is running is not "our" postgres).

alexeyklyukin: I think it is a good idea to check for the process name in addition to the timestamp, and also do this inside `_is_pid_running`. I'd go even further and rename the function to `_is_postmaster_pid_running`, since it is only used for the checking the postmaster pid so far.  We can indeed pass the process start time from the code that has read the pid originally to that function (so far it seems the only place where we don't immediately pass the pid to the `_is_pid_running` is s `wait_for_port_open`), and assume the postmaster process name remains unchanged, so that we can make it a constant.

CyberDem0n:
BTW, yesterday we hit that problem on Kubernetes.
One very simple additional check can help to solve the problem:
patroni can never be be a parent of postmaster process, therefore in addition to check of start_time we can check that `proc.parent() != os.getpid()`

ants:
I did an experimental refactor of the process handling [here](https://github.com/ants/patroni/pull/9). Moves the process handling stuff to a separate file with a reasonably well defined goal. postgresql.py is pretty big so factoring stuff out seems like a win. Do you think that is worth cleaning up and merging?

CyberDem0n:
I've already did multiple attempts to fix failing reinit due to async executor being busy, but in order to do it properly we should be able to cancel a running task. It is really tough, without refactoring of entire Postgresql class...

So yes, it's definitely worth it!

Can you also have a look on https://github.com/zalando/patroni/pull/553?

ants:
#553 is partly what prompted the refactor. I didn't review it in detail before doing the refactor, but checking it out now it seems like my refactor reduces it to the `state_handler.stop()` call when data dir was found empty.

CyberDem0n:
:+1:

alexeyklyukin:
👍

Issue #542, opened 2017-10-11
Patroni does not perform failover if master lost data directory


ainlolcat:
We are migrating from 1.2 to 1.3.4 and one of our tests fails - we removed data directory and are expecting failover to replica and master reinitialization. But with new version master detects empty dir, releases key and .. wait for leader to bootstrap. Meanwhile replica is considering master as "master" and wait for master too. 
From master: 

> 2017-10-11 11:35:27,431 ERROR: get_postgresql_status
> Traceback (most recent call last):
>   File "/patroni/patroni/api.py", line 405, in get_postgresql_status
>     retry=retry)[0]
>   File "/patroni/patroni/api.py", line 381, in query
>     return retry(self.server.query, sql, *params)
>   File "/patroni/patroni/utils.py", line 259, in __call__
>     return func(*args, **kwargs)
>   File "/patroni/patroni/api.py", line 452, in query
>     raise e
> OperationalError: could not open file "base/12404/2600": No such file or directory
> 
> 2017-10-11 11:35:27,433 INFO: released leader key voluntarily as data dir empty and currently leader
> ...
> 2017-10-11 11:35:37,416 INFO: Lock owner: None; I am node1
> 2017-10-11 11:35:37,420 INFO: waiting for leader to bootstrap

From slave:

> 2017-10-11 14:20:50,223 INFO: Got response from node1 http://10.129.1.63:8008/patroni: {"state": "unknown", "database_system_identifier": "6475561736777527351", "role": "master", "patroni": {"scope": "common", "version": "1.3.4"}}
> 2017-10-11 14:20:50,262 WARNING: Master (node1) is still alive
> 2017-10-11 14:20:50,267 INFO: following a different leader because i am not the healthiest node

I suppose that replica cannot promote itself because broken master declares wrong status. Broken master declares wrong status because it released key but never update `postgresql.role` value.

CyberDem0n:
Good catch, will fix it tomorrow.

ainlolcat:
Thank You. Failover works now but ex-master was not able to start correctly after bootstrap. Looks like this is because postgresql process was not killed before bootstrap and in result we have logs filled with `FATAL:  could not receive data from WAL stream: ERROR:  requested starting point 0/A000000 is ahead of the WAL flush position of this server 0/9002FA8`. Looks like after failover patroni tries to start postgresql several times but each time it fails with `FATAL:  could not create any TCP/IP sockets` because previous process still running. Log looks like this: 

[...]

PR #553, opened 2017-10-31
shutdown postgresql before bootstrap when we lost data directory

ainlolcat:
Tries to kill postgresql before bootstrap to prevent old process from interfering. Fix for second part of https://github.com/zalando/patroni/issues/542

### DIFF
--- a/patroni/postgresql.py
+++ b/patroni/postgresql.py
@@ -760,7 +760,7 @@ class Postgresql(object):
                 self.postmaster = {'pid': pmpid, 'start_time': pmstart}
                 logger.info("Updated postmaster info: {}".format(self.postmaster))
             except ValueError:
-                # Garbage in the pid file
+                logger.exception("Cannot update postmaster info with data duw garbage in pid file: {}".format(pid_data))
                 pass
         else:
             logger.warning("Cannot update postmaster info with data from pid file: {}".format(pid_data))
--- a/patroni/postgresql.py
+++ b/patroni/postgresql.py
@@ -71,6 +71,32 @@ def null_context():
     yield
 
 
+def _update_postmaster_info(func):
+    def wrapper(self):
+        ret = func(self)
+        if isinstance(ret, dict) and ret:
+            if 'pid' in ret and 'start_time' in ret:
+                old_pid = self._postmaster_info.get('pid', 0)
+                old_start_time = self._postmaster_info.get('start_time', 0)
+                try:
+                    pmpid = int(ret['pid'])
+                    pmstart = int(ret['start_time'])
+                    if pmpid != old_pid or pmstart != old_start_time:
+                        self._postmaster_info = {'pid': pmpid, 'start_time': pmstart}
+                        logger.info("Updated postmaster info {}".format(self._postmaster_info))
+                except ValueError:
+                    logger.exception("Cannot update postmaster info with data due garbage in pid file: {}".format(ret))
+            else:
+                logging.warning("Cannot update postmaster info with dict {}. "
+                                "Dict should have pid and start_time keys with values greater then zero.".format(ret))
+        else:
+            logging.debug("Cannot update postmaster info with {}. "
+                          "Wrapped function response is not dict or empty dict.".format(ret))
+        return ret
+
+    return wrapper
+
+
 class Postgresql(object):
 
     # List of parameters which must be always passed to postmaster as command line options
@@ -142,7 +168,7 @@ class Postgresql(object):
         self._pg_hba_conf = os.path.join(self._config_dir, 'pg_hba.conf')
         self._recovery_conf = os.path.join(self._data_dir, 'recovery.conf')
         self._postmaster_pid = os.path.join(self._data_dir, 'postmaster.pid')
-        self.postmaster = {'pid': 0, 'start_time': 0}
+        self._postmaster_info = {'pid': 0, 'start_time': 0}
         self._trigger_file = config.get('recovery_conf', {}).get('trigger_file') or 'promote'
         self._trigger_file = os.path.abspath(os.path.join(self._data_dir, self._trigger_file))
 
@@ -171,7 +197,6 @@ class Postgresql(object):
             self._write_postgresql_conf()  # we are "joining" already running postgres
             if self._replace_pg_hba():
                 self.reload()
-            self.update_postmaster_info()
 
     @property
     def _configuration_to_save(self):
@@ -714,6 +739,7 @@ class Postgresql(object):
     def is_running_with_lost_data_dir(self):
         return self.get_pid_with_lost_data_dir() > 0
 
+    @_update_postmaster_info
     def read_pid_file(self):
         """Reads and parses postmaster.pid from the data directory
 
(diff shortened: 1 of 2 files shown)


patroni/postgresql.py
```diff
@@ -733,6 +736,12 @@ def get_pid(self):
             logger.warning("Garbage pid in postmaster.pid: {0!r}".format(pid))
             return 0
 
+    def get_pid_with_lost_data_dir(self):
+        process = list(filter(lambda p: p.name() == "postgres" and self._data_dir in p.cmdline(), psutil.process_iter()))
```
CyberDem0n: Unfortunately this check is not really safe: `self._data_dir in p.cmdline()`

If you have more than one postgres cluster running on the same host with some strange configuration, it might result that more than one postgres would be killed.

The best we could do - keep the content of `postmaster.pid` in memory (especially lines 1 and 3, containing postmaster pid and start_time), and if data directory has gone away, use this cached content to terminate postmaster.

ants: I was just thinking along similar lines about whether we should create a `psutil.Process` object and cache that. This will do the hard work of checking if we still have the same process running. I think it might even simplify the code a good bit.

ants: Looking at psutil source, Process class just determines identity via `(pid, create_time)`, so rolling our own would also work. But as we need the psutil dependency anyway for `_wait_for_user_backends_to_close()` we might as well use it here too. Seems like a pretty well tested package.

patroni/postgresql.py
```diff
@@ -733,6 +738,37 @@ def get_pid(self):
...
+    def update_postmaster_info(self):
+        pid_data = self.read_pid_file()
+        if len(pid_data) > 5:
+            try:
+                pmpid = int(pid_data['pid'])
+                pmstart = int(pid_data['start_time'])
+                self.postmaster = {'pid': pmpid, 'start_time': pmstart}
+                logger.info("Updated postmaster info: {}".format(self.postmaster))
+            except ValueError:
+                logger.exception("Cannot update postmaster info with data duw garbage in pid file: {}".format(pid_data))
```
hjacobs: "duw"?

patroni/postgresql.py
ainlolcat: Changed logic so it get postmaster from PID file and then during stop compares cached start_time from postmaster and create_time from psutls. Actually they are slightly different so I am checking difference between them.

patroni/postgresql.py
ainlolcat: Sorry. Typo. Fixed.

patroni/postgresql.py
```diff
@@ -733,6 +738,36 @@ def get_pid(self):
             logger.warning("Garbage pid in postmaster.pid: {0!r}".format(pid))
             return 0
 
+    def get_pid_with_lost_data_dir(self):
+        process = list(filter(
+            lambda p: p.pid == self.postmaster["pid"] and p.name() == "postgres" and self._data_dir in p.cmdline(),
```
CyberDem0n: We already know the exact pid, no need to iterate through all processes in the system.

Just replace this line with:
```python
process = psutil.Process(self._postmaster_info['pid'])
```

`self._data_dir in p.cmdline()` -- it will check only exact match, and unfortunately can't handle all cases:
 postgresql.data_dir contains absolute path (/home/postgres/data), but postgres was started with `-D data` arguments and other way around.
I would suggest to use `process.cwd()` instead. This value always match with the real data directory.
Be careful, `cwd` method can throw `psutil.AccessDenied` exception. Also when comparing `process.cwd()` with `self._data_dir`, don't forget that you need to compare absolute paths.

ants: I think it would be best to cache the psutil.Process object once we know it's the correct process (`create_time()` matches start_time from pid file). psutil.Process methods guarantee that we have the correct process for [the methods that interest us](https://github.com/giampaolo/psutil/blob/62c845a7d97eea0c2740b8f488a62b6c08eb4331/psutil/__init__.py#L378-L397). This requires a bit more refactoring, but I suspect that it will give a net win in simplicity.

ainlolcat: We can have another process with same id if previos was killed.  `p.cmdline()` returns array with parameters of command like `['/usr/pgsql-9.6/bin/postgres', '-D', '/var/lib/pgsql/data/postgresql_node1', .... ]` so it will be safe search for `self._data_dir` which should be passed to postgresql, if I understand correctly.

ainlolcat: Also I cannot guarantee if process still running so it's more safe to check if process exists than call `psutil.Process()`

patroni/postgresql.py
```diff
@@ -733,6 +738,36 @@ def get_pid(self):
...
+        logger.debug("Cached postmaster info: {}, possible processes: {}".format(self.postmaster, process))
+        if process:
+            if abs(self.postmaster["start_time"] - process[0].create_time()) < 2:
+                return process[0].pid
+            else:
+                logger.info("Process with pid {} was started at different time {}"
+                            .format(process[0].pid, process[0].create_time()))
+        return 0
+
+    def update_postmaster_info(self):
```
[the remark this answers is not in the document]
ainlolcat: Ok. Sounds good. Will try this.

patroni/postgresql.py
```diff
@@ -71,6 +71,32 @@ def null_context():
...
+                old_pid = self._postmaster_info.get('pid', 0)
+                old_start_time = self._postmaster_info.get('start_time', 0)
+                try:
+                    pmpid = int(ret['pid'])
+                    pmstart = int(ret['start_time'])
+                    if pmpid != old_pid or pmstart != old_start_time:
+                        self._postmaster_info = {'pid': pmpid, 'start_time': pmstart}
+                        logger.info("Updated postmaster info {}".format(self._postmaster_info))
+                except ValueError:
+                    logger.exception("Cannot update postmaster info with data due garbage in pid file: {}".format(ret))
```
CyberDem0n: We don't really want to see an entire stack trace here, warning would be more appropriate.
Also logging functions support "%" placeholders, it is not really nice to mix different types of formatting.
Basically `logger.warning('Cannot update postmaster info with data due garbage in pid file: %s', ret)` would be good enough.

Two other log messages are unnecessary:
line 93 -- this one will never happen
line 90 -- it is a normal situation when only part of data is written into postmaster.pid and it represents different stages of postmaster start.

ainlolcat: If that is the case I suppose that we can remove check `isinstance(ret, dict)` as well because we know return type of `read_pid_file`.

patroni/postgresql.py
ainlolcat: @ants @CyberDem0n should we leave current version or change to `psutil.Process(self._postmaster_info['pid'])` ?

patroni/postgresql.py
CyberDem0n: Sure, `isinstance(ret, dict)` check is not really needed.

I've just posted here very simplified version.

patroni/postgresql.py
```diff
@@ -933,6 +976,12 @@ def stop(self, mode='fast', block_callbacks=False, checkpoint=True, on_safepoint
 
     def _do_stop(self, mode, block_callbacks, checkpoint, on_safepoint):
         if not self.is_running():
+            if self.is_running_with_lost_data_dir():
```
CyberDem0n: I think we can get rid from `is_running_with_lost_data_dir`. It is not doing anything useful. The same check is done below.

patroni/postgresql.py
```diff
@@ -733,6 +757,25 @@ def get_pid(self):
...
+        try:
+            process = psutil.Process(self._postmaster_info['pid'])
+            # check difference instead of values because of rounding issues
+            if abs(self._postmaster_info["start_time"] - process.create_time()) < 2:
+                return process.pid
+            else:
+                logger.warning("Process with pid %s was started at different time %s .",
+                               process.pid, process.create_time())
+        except psutil.NoSuchProcess:
+            logging.warning("Cannot find process %s .", self._postmaster_info['pid'])
```
CyberDem0n: I think here should be an `info` message.
Also please be consistent and use "logger" everywhere (not logging).

CyberDem0n:
LGT. Off course, unit-tests are missing, but I will fix it later.

CyberDem0n:
:+1:

ainlolcat:
I wonder if we can get release with this fix anytime soon?

patroni/postgresql.py
```diff
@@ -733,6 +754,25 @@ def get_pid(self):
             logger.warning("Garbage pid in postmaster.pid: {0!r}".format(pid))
             return 0
 
+    def get_pid_with_lost_data_dir(self):
+        logger.info("Trying to check if process running without directory "
+                    "with cached postmaster info: %s .", self._postmaster_info)
+        try:
+            process = psutil.Process(self._postmaster_info['pid'])
```
alexeyklyukin: Does it make sense to check here that self._postmaster_info['pid'] > 0.

patroni/postgresql.py
```diff
@@ -933,6 +973,11 @@ def stop(self, mode='fast', block_callbacks=False, checkpoint=True, on_safepoint
 
     def _do_stop(self, mode, block_callbacks, checkpoint, on_safepoint):
         if not self.is_running():
+            pid = self.get_pid_with_lost_data_dir()
+            if pid > 0:
+                self.terminate_starting_postmaster(pid)
+                self.clean_postmaster_info()
```
alexeyklyukin: The on_safepoint callback is not called. We could pass a watchdog.disable here from ha.py:1018 and call on_safepoint, eliminating line 1020 and removing a TODO at line 1019.

patroni/postgresql.py
```diff
@@ -71,6 +71,25 @@ def null_context():
     yield
 
 
+def _update_postmaster_info(func):
```
alexeyklyukin: I'd name it postmaster_cached_info.

alexeyklyukin:
👍

ainlolcat:
@alexeyklyukin I am confused by review. Should it be part of new merge request?

alexeyklyukin:
@ainlolcat sorry for not commenting out, this and the surrounding code will be refactored in a https://github.com/ants/patroni/pull/9, no need to touch it in this PR.

PR #554, opened 2017-10-31
Perform crash recovery in a single user mode if postgres died as master

CyberDem0n:
But do it only if pg_rewind is enabled or there is no master at the moment.
Such "crash recovery" procedure was advised by Heikki Linnakangas


alexeyklyukin:
👍

CyberDem0n:
:+1:

Issue #552, opened 2017-10-31
Patroni consul token configuration is not picked


DeathBorn:
Patroni 1.3.5
both python2 and python3
patroni.yml
```
consul:
  url: "http://consul.service.discovery.zone:8500"
  token: "some-token"
```
This token is not picked in https://github.com/zalando/patroni/blob/master/patroni/dcs/consul.py#L86
Judging by https://github.com/cablehead/python-consul/blob/v0.7.2/consul/base.py#L1633
token is not passed here at all. 
(this is a stack trace (I have modified some code in patroni/dcs/consul and consul/base by adding some logging so the actual lines are a bit further) )
```
Oct 31 16:11:25 zhost001 patroni[14900]: Traceback (most recent call last):
Oct 31 16:11:25 zhost001 patroni[14900]: File "/usr/lib/python2.7/site-packages/patroni/dcs/consul.py", line 230, in refresh_session
Oct 31 16:11:25 zhost001 patroni[14900]: return self.retry(self._do_refresh_session)
Oct 31 16:11:25 zhost001 patroni[14900]: File "/usr/lib/python2.7/site-packages/patroni/dcs/consul.py", line 191, in retry
Oct 31 16:11:25 zhost001 patroni[14900]: return self._retry.copy()(*args, **kwargs)
Oct 31 16:11:25 zhost001 patroni[14900]: File "/usr/lib/python2.7/site-packages/patroni/utils.py", line 259, in __call__
Oct 31 16:11:25 zhost001 patroni[14900]: return func(*args, **kwargs)
Oct 31 16:11:25 zhost001 patroni[14900]: File "/usr/lib/python2.7/site-packages/patroni/dcs/consul.py", line 224, in _do_refresh_session
Oct 31 16:11:25 zhost001 patroni[14900]: lock_delay=0.001, behavior='delete')
Oct 31 16:11:25 zhost001 patroni[14900]: File "/usr/lib/python2.7/site-packages/consul/base.py", line 1711, in create
Oct 31 16:11:25 zhost001 patroni[14900]: data=data)
Oct 31 16:11:25 zhost001 patroni[14900]: File "/usr/lib/python2.7/site-packages/patroni/dcs/consul.py", line 101, in wrapper
Oct 31 16:11:25 zhost001 patroni[14900]: return callback(self.response(self.http.request(method.upper(), self.uri(path, params), **kwargs)))
[...]

PR #555, opened 2017-10-31
Always send token in the X-Consul-Token http header

CyberDem0n:
Fixes https://github.com/zalando/patroni/issues/552


DeathBorn:
This works

DeathBorn:
One more thing. Since this is towards consul, there should be another parameter for configuration. That is ```dc```
Please add 'dc' to https://github.com/zalando/patroni/blob/d15e8ffc87d81d9f56777e6cca0594d613e44716/patroni/dcs/consul.py#L149

alexeyklyukin:
👍

CyberDem0n:
:+1:

Issue #557, opened 2017-11-03
Patronictl: list scheduled events (failover)


a1exsh:
It would be nice if `patronictl` had a subcommand to list scheduled failovers.
The `list` subcommand could also have an additional column to indicate the closest scheduled failover (from and/or to member).

uspen:
And pause state :-)

PR #567, opened 2017-11-23
Patronictl extended info

CyberDem0n:
Show information about scheduled failover and maintenance mode when showing list of cluster members.

Fixes https://github.com/zalando/patroni/issues/557

In addition to that fix postgres version check regexp and apply pep8 formatting to the tests.

### DIFF
--- a/patroni/ctl.py
+++ b/patroni/ctl.py
@@ -622,12 +622,15 @@ def output_members(cluster, name, extended=False, fmt='pretty'):
     logging.debug(cluster)
     leader_name = None
     if cluster.leader:
-        leader_name = cluster.leader.member.name
+        leader_name = cluster.leader.name
 
     xlog_location_cluster = cluster.last_leader_operation or 0
 
     # Mainly for consistent pretty printing and watching we sort the output
     cluster.members.sort(key=lambda x: x.name)
+
+    extended = extended or any(m.data.get('scheduled_restart') for m in cluster.members)
+
     for m in cluster.members:
         logging.debug(m)
 
@@ -637,21 +640,13 @@ def output_members(cluster, name, extended=False, fmt='pretty'):
         elif m.name == cluster.sync.sync_standby:
             role = 'Sync standby'
 
-        host = m.conn_kwargs()['host']
-
-        xlog_location = m.data.get('xlog_location') or 0
-        lag = ''
-        if xlog_location_cluster >= xlog_location:
+        xlog_location = m.data.get('xlog_location')
+        lag = 'unknown'
+        if xlog_location is not None and xlog_location_cluster >= xlog_location:
             lag = round((xlog_location_cluster - xlog_location)/1024/1024)
 
-        row = [
-            name,
-            m.name,
-            host,
-            role,
-            m.data.get('state', ''),
-            lag,
-        ]
+        row = [name, m.name, m.conn_kwargs()['host'], role, m.data.get('state', ''), lag]
+
         if extended:
             value = ''
             scheduled_restart = m.data.get('scheduled_restart')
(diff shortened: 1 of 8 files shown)


patroni/utils.py
```diff
@@ -195,7 +195,7 @@ def _sleep(interval):
 
 
 def is_valid_pg_version(version):
-    return re.match(r'[1-9][0-9]?(\.(0|([1-9][0-9]?))){2}$', version)
+    return re.match(r'[1-9][0-9]?(\.(0|([1-9][0-9]?))){1,2}$', version)
```
alexeyklyukin: This is still wrong, although there are little negative consequences of that wrongness. The regex accepts 9.3 and 9.5 as a valid PostgreSQL versions, although it passes it to the postgres_version_to_int that expects major.minor versions (although postgres_version_to_int also doesn't know 9.3 is not a proper syntax for the major.minor version before 10, so it gets passed). From the user perspective, it seems that we accept just the major version, and it actually works fine with 9.x major versions, but then we don't accept 10 or 11, perfectly valid major versions.

What we should do is to accept only full version, i.e. major.minor for 10 and above or 9.major.minor.

Also, the code in the error message on the line 473 still complains about first.major.minor. It should say something like 'PostgreSQL version should be in the 9.major.minor below version 10 or major.minor otherwise.' (the example on the line 464 should also add the new version format, e.g. 9.5.2 or 10.1).

The test in test_ctl.py:240 actually caught the issue correctly, but you've modified it.

patroni/ctl.py
```diff
@@ -637,21 +640,13 @@ def output_members(cluster, name, extended=False, fmt='pretty'):
...
             role = 'Sync standby'
 
-        host = m.conn_kwargs()['host']
-
-        xlog_location = m.data.get('xlog_location') or 0
-        lag = ''
-        if xlog_location_cluster >= xlog_location:
+        xlog_location = m.data.get('xlog_location')
+        lag = 'unknown'
+        if xlog_location is not None and xlog_location_cluster >= xlog_location:
```
alexeyklyukin: On at least one occasion I've seen the second condition (xlog_location_cluster >= xlog_location) as not true consistently for the running test cluster (the cluster has not been receiving any data): https://gist.github.com/alexeyklyukin/6b84df0c129216ba15e43d0421daed20

I cannot reproduce that issue from now on, but I think it's worth complaining about such conditions instead of just hiding them under 'Unknown' (perhaps add it to the service_info line).

tests/test_ctl.py
```diff
@@ -234,7 +237,7 @@ def test_restart_reinit(self, mock_get_dcs):
         assert result.exit_code == 1
 
         # Wrong pg version
-        result = self.runner.invoke(ctl, ['restart', 'alpha', '--any', '--pg-version', '9.1'], input='y')
+        result = self.runner.invoke(ctl, ['restart', 'alpha', '--any', '--pg-version', '9.1.2.3'], input='y')
```
alexeyklyukin: That test has been perfectly valid (see my previous comment).

patroni/ctl.py
```diff
@@ -622,12 +622,15 @@ def output_members(cluster, name, extended=False, fmt='pretty'):
...
     if cluster.leader:
-        leader_name = cluster.leader.member.name
+        leader_name = cluster.leader.name
 
     xlog_location_cluster = cluster.last_leader_operation or 0
 
     # Mainly for consistent pretty printing and watching we sort the output
     cluster.members.sort(key=lambda x: x.name)
+
+    extended = extended or any(m.data.get('scheduled_restart') for m in cluster.members)
```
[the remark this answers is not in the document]
CyberDem0n: Actually it is used.
It is possible to specify `  -e, --extended     Show some extra information` argument, which will force extra column to appear in `patronictl list` output.

alexeyklyukin:
👍

CyberDem0n:
:+1:

patroni/ctl.py
```diff
@@ -680,6 +671,21 @@ def output_members(cluster, name, extended=False, fmt='pretty'):
 
     print_output(columns, rows, alignment, fmt)
 
+    service_info = []
+    if cluster.is_paused():
+        service_info.append('Maintenance mode: on')
+
+    if cluster.failover and cluster.failover.scheduled_at:
```
a1exsh: Hm, does this mean there can be only one failover scheduled at any given moment?

CyberDem0n: Yes

PR #571, opened 2017-12-08
Some improvements to patronictl

ants:
Also adds a check for lock when exiting to avoid an unnecessary exception. Or maybe we should just catch that exception though... Stacktraces confuse users.

### DIFF
--- a/patroni/ctl.py
+++ b/patroni/ctl.py
@@ -97,7 +97,7 @@ def store_config(config, path):
         yaml.dump(config, fd)
 
 
-option_format = click.option('--format', '-f', 'fmt', help='Output format (pretty, json)', default='pretty')
+option_format = click.option('--format', '-f', 'fmt', help='Output format (pretty, json, yaml)', default='pretty')
 option_watchrefresh = click.option('-w', '--watch', type=float, help='Auto update the screen every X seconds')
 option_watch = click.option('-W', is_flag=True, help='Auto update the screen every 2 seconds')
 option_force = click.option('--force', is_flag=True, help='Do not ask for confirmation at any point')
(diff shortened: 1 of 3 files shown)


patroni/ctl.py
```diff
@@ -1035,3 +1046,36 @@ def show_config(obj, cluster_name):
...
+    if not cluster_name:
+        return
+
+    click.echo("")
+    cluster = get_dcs(obj, cluster_name).get_cluster()
+    for m in cluster.members:
+        if m.api_url:
+            if not member_names or m.name in member_names:
+                try:
+                    response = request_patroni(m, 'get', None, auth_header(obj))
```
CyberDem0n: Here you are trying to send `{'Authorization': 'Basic base64encoded'} ` as a payload of a GET request, what doesn't make any sense.
I guess you wanted to execute request `GET http://host:port/patroni`, because it will always return http code = 200
It should be done as `response = request_patroni(m, 'get', 'patroni')`.

The rest looks fine to me.

ants: Indeed. No idea why I was doing that. Will fix.

patroni/ctl.py
```diff
@@ -149,9 +150,12 @@ def print_output(columns, rows=None, alignment=None, fmt='pretty', header=True,
...
         return
 
-    if fmt == 'json':
+    if fmt in ['json', 'yaml']:
         elements = [dict(zip(columns, r)) for r in rows]
-        click.echo(json.dumps(elements))
+        if fmt == 'json':
+            click.echo(json.dumps(elements))
+        elif fmt == 'yaml':
+            click.echo(yaml.safe_dump(elements, encoding=None, allow_unicode=True, width=200))
```
alexeyklyukin: Should we use `default_flow_style=False` for showing the members list?  With a flow style the output is:

```
- {Cluster: batman, Host: 127.0.0.1, Lag in MB: 0, Member: postgresql0, Role: Leader, State: running}
- {Cluster: batman, Host: 127.0.0.1, Lag in MB: 0, Member: postgresql1, Role: '', State: running}

```
not much different from JSON

patroni/ctl.py
```diff
@@ -690,19 +694,26 @@ def output_members(cluster, name, extended=False, fmt='pretty'):
...
+            return
+        else:
+            cluster_names = [obj['scope']]
 
     for cluster_name in cluster_names:
         dcs = get_dcs(obj, cluster_name)
 
         for _ in watching(w, watch):
+            if timestamp:
+                click.echo(datetime.datetime.now().strftime("%Y-%m-%d %H:%M:%S"))
```
alexeyklyukin: Does it make sense to use a timestamp(0) here?

CyberDem0n:
:+1:

alexeyklyukin:
👍

PR #578, opened 2017-12-19
Implement '/sync' endpoint 

### LINKED ISSUE #189: Expose synchronous replication status through API
Hi,

I started a cluster for testing, made of 3 nodes and I think that the current setup does not allow me to scale reads and backups.

**My desired scenario:**
1. Master - writes and reads 
2. Synchronous slave - reads only for offloading of master
3. Potential synchronous slave - backups only

The current haproxy config and patroni API allows for this:

```
backend be_db_rw
    option httpchk GET /master
    http-check expect status 200
    ...

backend be_db_ro
    option httpchk GET /
    http-check expect string replica
    ...
```

But this will result in having one master and 2 read only servers.

I was thinking of adding another key - value to the API response containing the result of:
`postgres=# select client_addr, sync_state from pg_stat_replication;
 client_addr | sync_state
-------------+------------
 10.0.0.1  | sync
 10.0.0.2  | potential`

This is visible only from the current master so I can't simply add it to the API response.
So I was thinking of adding the key client_addr and value sync_state in etcd and after expose them through API. There should be a match of IP so that the value is displayed on the correct server.
This way the haproxy setup for sync and potential slaves will be strait forward.

Maybe I am missing a feature of patroni that allows one to find out which is the sync and potential.
Or maybe there is a better way of doing this.
Ideas?

### LINKED ISSUE #415: Add new healthcheck endpoint to Patroni api
It should return 200 only if the node is running as a synchronous replica (when it's exposed as synchronous in `sync` key in the DCS).

CyberDem0n:
It will respond with http status code 200 only when the node is running as a synchronous replica.

Fixes #189
Fixes #415


patroni/api.py
```diff
@@ -89,6 +89,9 @@ def do_GET(self, write_status_code_only=False):
                 status_code = 503
             elif response['role'] in path:  # response['role'] != 'master'
                 status_code = 503 if patroni.noloadbalance else 200
+            elif path.startswith('/sync') and cluster.is_synchronous_mode() and response.get('role') == 'replica' \
```
alexeyklyukin: I'm not sure this is the right place to plug it:

/sync will return 200 if running on the master with a lock, 503 if it is a master without the lock.
/sync/replica will return 503 even on a sync replica if noloadbalance tag is set for it (which is completely reasonable), or 200 if noloadbalance is not set, even if it's not running in a synchronous mode.

CyberDem0n: > /sync will return 200 if running on the master with a lock, 503 if it is a master without the lock.

It won't, we have a line 84 for that.

> or 200 if noloadbalance is not set, even if it's not running in a synchronous mode.

Why? It check explicitly that `cluster.sync and cluster.sync.sync_standby == patroni.postgresql.name`

alexeyklyukin: > It won't, we have a line 84 for that.

How is so? the user calling /sync will get to the line 84 and if it is a master with a lock will return 200. Otherwise (a master without the lock) it will get to line 89 and return 503.  Both have probably nothing to do with what he asked.

> Why? It check explicitly that cluster.sync and cluster.sync.sync_standby == patroni.postgresql.name

Because of line 90 and 91, that I think will be triggered for the `/sync/replica` since replica is in path.

ants:
On my todo list is a plan to implement "quorum" commit support for the sync commit feature. I haven't yet worked out the details, but the general idea is that you first configure how many nodes you can lose at once without compromising write availability. Patroni would then store a list of synchronous candidates in the sync key and configure synchronous_standby_names with `ANY $n ($candidates)`. When master fails, a candidate considers itself healthy if it can get xlog location from `len(candidates) - n` other nodes. This is safe because we can have a maximum of `n - 1` xlog locations missing and `n` nodes have the latest xlog, so we must see at least one of the nodes that received latest commit.

I think this makes the semantics of the `/sync` endpoint kind of funky. The use cases in the referenced tickets want different things, one wants to know which nodes are definitely sync (for querying) and the other one wants to know which nodes are definitely async. I guess having `/sync` mean definitely sync and `/async` mean definitely not sync would be an ok option. Another idea is a general status query interface that evaluates a boolean expression. Something like `/status?match=role == master or (sync_state == sync and not tags.noloadbalance)`. I'm of two minds on this, the flexibility and conceptual simplicity is great for power users, but for casual users who don't understand all the nuances, prepackaged solutions are nicer. Although an expression interface does not preclude creating higher level concepts like `/status?match=has_latest_data_visible`.

CyberDem0n:
> On my todo list is a plan to implement "quorum" commit support for the sync commit feature. 

I think it is a great idea!

> this makes the semantics of the /sync endpoint kind of funky. 

Yeah, with quorum commit it would be not possible to tell which nodes are synchronous and which are not :(

> Another idea is a general status query interface that evaluates a boolean expression. 

We can have a separate endpoint for that. Shouldn't be so hard to implement.

CyberDem0n:
:+1:

alexeyklyukin:
👍

PR #584, opened 2017-12-21
Minimize amount of SELECTS issued by Patroni on every loop

CyberDem0n:
Every iteration of HA loop Patroni needs to call pg_is_in_recovery() and
calcualte absolute wal_position. It was doing two separate SELECT
statements for that. In case of master it was doing even three queries
(wal_position two times).
We will issue one select for every HA loop and cache the results.

### DIFF
--- a/patroni/postgresql.py
+++ b/patroni/postgresql.py
@@ -36,11 +36,11 @@ STOP_POLLING_INTERVAL = 1
 REWIND_STATUS = type('Enum', (), {'INITIAL': 0, 'CHECK': 1, 'NEED': 2, 'NOT_NEED': 3, 'SUCCESS': 4, 'FAILED': 5})
 sync_standby_name_re = re.compile('^[A-Za-z_][A-Za-z_0-9\$]*$')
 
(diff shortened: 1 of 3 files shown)


patroni/postgresql.py
```diff
@@ -690,13 +698,28 @@ def create_replica(self, clone_member):
...
+        return self._cluster_info_state.get(name)
+
     def is_leader(self):
-        try:
-            return not self._is_leader_retry(self._query, 'SELECT pg_is_in_recovery()').fetchone()[0]
-        except RetryFailedError as e:  # SELECT pg_is_in_recovery() failed two times
-            if not self.is_starting() and self.pg_isready() == STATE_REJECT:
-                self.set_state('starting')
-            raise PostgresConnectionException(str(e))
+        return not self._cluster_info_state_get('is_leader')
```
cbandy: This phrasing is confusing. Perhaps the cache key should be named `in_recovery`?

CyberDem0n: You are right, will fix it.

alexeyklyukin:
👍

Issue #585, opened 2017-12-22
Allow manual failovers when master is not present


alexeyklyukin:
At the moment, when master is not present and the replicas cannot promote themselves for some reason (i.e. being to far from the last known master position) there is no way to force the promotion of one of the replicas. We can introduce the `failover` command that does exactly this (checking that the master is not present), and rename our existing command that does the controlled switchover when the master is present to `switchover`.

This would also address https://github.com/zalando/patroni/issues/520

Issue #520, opened 2017-09-14
Data loss when doing a switchover using async replication?


kagkarlsson:
When running a cluster and using async replication, do Patroni ensure that no data is lost when doing a switchover? 
If yes, how does that work?
If no, how do you suggest a maintenance switchover is done when using async replication?

(Am opening this issue as a `question` because I found no other forum to ask about these things, e.g. IRC, google groups, mailing list..?)

alexeyklyukin:
Normally, the replica that promotes makes sure it restores all available WAL segments before proceeding with promotion (this is done by PostgreSQL), and the master is also archiving all WAL segments and waiting for the streaming replicas to get up-to-date during its shutdown (also done by PostgreSQL). Therefore, as long as this is a controlled switchover and the former master is available during it there should not be any data loss.

It is rather different in the case of the failover, when the former master goes away unexpectedly. There, you cannot guarantee that all data is shipped to the replicas before the promotion, however, Patroni prevents any replica to promote that is more than  `maximum_lag_on_failover` bytes behind the last reported master position.

kagkarlsson:
Thank you for your quick answer!

Am assuming switchover works like this:
1. Issue "normal" shutdown of old master (=> PG allow archiving and replication are to complete before shutdown)
2. Patroni triggers a "manual" promote of the replica
3. Re-initialize old master as replica

Now this is a bit contrived, but what if syncing the replica or archiving is taking a long time to complete. Is there something preventing the regular Patroni loop from doing a "hmm I see the master is down/unavailable, lets promote the replica"?

alexeyklyukin:
No, it doesn't work this way. Patroni on the replica will never make a decision to promote while the leader lock is present. Patroni on the master that is shutting down will not relinquish this lock until it makes sure the master is not running (by checking its PID among other things). 

BTW, the 3rd step is actually not required in most cases, since the former master would start up as a replica and rejoin the new master (in case there are some issues we could use pg_rewind in order to remove the discrepancy).

kagkarlsson:
Ok, thank you for the clarification!

cbandy:
> Therefore, as long as this is a controlled switchover and the former master is available during it there should not be any data loss.

Is `patronictl failover` a controlled switchover?

alexeyklyukin:
it is.

mbanck:
[...]

PR #588, opened 2017-12-28
Rename failover to switchover and make new failover work without leader 

CyberDem0n:
In addition to that implement /switchover endpoint as an alias to /failover endpoint and implement more checks like:
* candidate must be provided for a failover
* switchover can't be scheduled in a pause state
* and so on

Fixes https://github.com/zalando/patroni/issues/585
Fixes https://github.com/zalando/patroni/issues/520

### DIFF
--- a/patroni/api.py
+++ b/patroni/api.py
@@ -337,6 +337,8 @@ class RestApiHandler(BaseHTTPRequestHandler):
 
         if action == 'failover' and not candidate:
             data = 'Failover could be performed only to a specific candidate'
+        elif action == 'switchover' and not leader:
+            data = 'Switchover could be performed only from a specific leader'
 
         if not data and scheduled_at:
             if not leader:
@@ -348,9 +350,6 @@ class RestApiHandler(BaseHTTPRequestHandler):
 
         if not data and cluster.is_paused() and not candidate:
             data = action.title() + ' is possible only to a specific candidate in a paused state'
-
-        if not data and not leader and not candidate:
-            data = 'No values given for required parameters leader and candidate'
 
         if not data and not scheduled_at:
             data = self.is_failover_possible(cluster, leader, candidate, action)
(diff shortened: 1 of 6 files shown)


patroni/api.py
```diff
@@ -325,39 +325,47 @@ def do_POST_failover(self):
...
+        if not data and cluster.is_paused() and not candidate:
+            data = action.title() + ' is possible only to a specific candidate in a paused state'
+
+        if not data and not leader and not candidate:
+            data = 'No values given for required parameters leader and candidate'
+
+        if not data and not scheduled_at:
+            data = self.is_failover_possible(cluster, leader, candidate, action)
+            if data:
+                status_code = 500
```
alexeyklyukin: I'd return 409 (conflict) or 412 (precondition failed) instead of 500 here, as it is not really a server internal error, but rather the client expectation supplied via the query parameters don't match the actual state of the cluster.

alexeyklyukin:
Seems to be there are only 2 differences between the failover and the switchover:

- failover always requires the candidate
- the check for failing over to the synchronous replica is not enforced during the failover if the cluster is running in a synchronous mode.

Otherwise, switchover seems to be just an alias of the failover.

In `patronictl`, there is a check of whether the cluster to switchover has a master, but that check is not duplicated in the API itself, allowing to switchover in the cluster without the master via the REST API, potentially loosing data. I think this check should be present in the API itself.

CyberDem0n:
:+1:

CyberDem0n:
In addition to that implement /switchover endpoint as an alias to /failover endpoint and implement more checks like:
* candidate must be provided for a failover
* switchover can't be scheduled in a pause state
* and so on

Fixes https://github.com/zalando/patroni/issues/585
Fixes https://github.com/zalando/patroni/issues/520

Issue #523, opened 2017-09-15
Add support for multiple Etcd endpoints


davissp14:
Looking at the docs specified here, it appears we can only specify a single Etcd endpoint.
https://patroni.readthedocs.io/en/latest/ENVIRONMENT.html#environment

The limitation here is that if that specific endpoint is not available on start-up, it's obviously not going to be able to auto-discover the other members to establish a connection.  Seems we could avoid this issue by allowing users to manually specify each member endpoint or each proxy endpoint they have available.  Patroni could then cycle through the endpoints until it is able to establish a connection.

alexeyklyukin:
You can point Patroni to either A or SRV records containing multiple endpoints, Patroni will cycle through them to find one it can connect to.

davissp14:
How would you go about specifying multiple internal ips inside a private network?

alexeyklyukin:
Same way. Actually our use-case is mostly DNS-based discovery, we don't use static initial cluster addresses. The latter is error-prone and doesn't play well with dynamic hosts (i.e. in he cloud), since the addresses are not known in advance. YMMV.

davissp14:
Interesting, our initial cluster addresses are known prior to start-up.  In the event a node need to be moved onto a new host, our environments are updated accordingly.  If the endpoints are only used to establish the initial connection and are read from the configuration file, it seems as long as we keep this file is updated when changes are made it shouldn't be an issue, right?

SRV records may be a sane approach, it's not really something we are prepared to use in the near term.  The endpoint cycling on failure sounds like something that's already happening, is multiple endpoint support within the config, something you guys would consider supporting?

CyberDem0n:
@davissp14, Patroni needs to know only one host from etcd cluster.

Every etcd node can report etcd cluster topology via  `/v2/machines` REST API call, and Patroni is using it: https://github.com/zalando/patroni/blob/master/patroni/dcs/etcd.py#L124

If one of the requests to etcd failed, Patroni switches to another etcd node, re-executing request and refreshes topology of etcd cluster.
In addition to that Patroni also doing it (refreshes etcd cluster topology) periodically, every 5 minutes.
If you move one of etcd node to another host and another ip, Patroni will discover it maximum in 5 minutes.

Why did we choose a 5 minutes? Because it nearly impossible to move all etcd nodes to a different ips/hosts.

And at the end we can return to your original question.
[...]

PR #589, opened 2017-12-29
Allow to specify multiple hosts for etcd

CyberDem0n:
This list will be used for initial discovery of etcd cluster members.
If for some reason during work this list of hosts has been exhausted (during work), Patroni will return to initial list.

In addition to that improve ipv6 compatibility by using a special function for splitting host and port.

Fixes https://github.com/zalando/patroni/issues/523

### DIFF
--- a/docs/ENVIRONMENT.rst
+++ b/docs/ENVIRONMENT.rst
@@ -39,6 +39,7 @@ Consul
 Etcd
 ----
 -  **PATRONI\_ETCD\_HOST**: the host:port for the etcd endpoint.
+-  **PATRONI\_ETCD\_HOSTS**: list of etcd endpoints in format host1:port1,host2:port2,etc...
 -  **PATRONI\_ETCD\_URL**: url for the etcd, in format: http(s)://(username:password@)host:port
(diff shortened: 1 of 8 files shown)


alexeyklyukin:
Interesting, the semantics of 'host' and 'hosts' seem to be rather different, namely the hosts members are not attempted to be resolved via DNS, therefore, they are assumed to be IP addresses. Is it intentional?

Perhaps it makes sense to attempt to resolve a single entry of the hosts lists, leaving the old parameter (host) solely for compatibility reasons?

CyberDem0n:
> namely the hosts members are not attempted to be resolved via DNS,

The only reason to resolve host into multiple ip addresses was to increase chances to find accessible member of etcd cluster from the first attempt (for loop over ip addresses)

> therefore, they are assumed to be IP addresses

No, it could be anything, host or IP. If you have more than one value specified in the `hosts` your chances already quite high, doesn't really pay of to resolve it separately.

Host and hosts are used only for initial discovery. As soon as request `GET hostN:port/v2/machines` succeed, Patroni will use the list of members returned by that request.

alexeyklyukin:
👍

CyberDem0n:
:+1:

davissp14:
Nice work!  Very much appreciated!

PR #590, opened 2018-01-04
More improvements in patronictl

CyberDem0n:
Make specifying cluster_name optional for some more commands.
If it is not specified, it's value would be taken from config file.


alexeyklyukin:
👍

CyberDem0n:
:+1:

PR #591, opened 2018-01-04
Expose current timeline in DCS and via API

CyberDem0n:
It is very easy to get current timeline on the master by executing
```sql
SELECT ('x' || SUBSTR(pg_walfile_name(pg_current_wal_lsn()), 1, 8))::bit(32)::int
```

Unfortunately the same method doesn't work when postgres is_in_recovery. Therefore we will use replication connection for that on the replicas. In order to avoid opening and closing replication connection on every HA loop we will cache the result if its value matches with the timeline of the master.

Also this PR introduces a new key in DCS: `/history`. It will contain a json serialized object with timeline history in a format similar to the usual history files. The differences are:
* Second column is the absolute wal position in bytes, instead of LSN
* Optionally there might be a fourth columnt - timestamp, (mtime of history file)

### DIFF
--- a/patroni/ha.py
+++ b/patroni/ha.py
@@ -125,7 +125,7 @@ class Ha(object):
                 data['pending_restart'] = True
             if not self._async_executor.busy and data['state'] in ['running', 'restarting', 'starting']:
                 try:
-                    data['xlog_location'] = self.state_handler.wal_position(retry=False)
+                    data['xlog_location'] = self.state_handler.wal_position()
                 except Exception:
                     pass
             if self.patroni.scheduled_restart:
@@ -989,6 +989,7 @@ class Ha(object):
     def _run_cycle(self):
         dcs_failed = False
         try:
+            self.state_handler.reset_cluster_info_state()
             self.load_cluster_from_dcs()
 
             if self.is_paused():
--- a/patroni/ha.py
+++ b/patroni/ha.py
@@ -83,8 +83,7 @@ class Ha(object):
             self.old_cluster = cluster
         self.cluster = cluster
 
-        if not cluster.is_unlocked():
-            self._leader_timeline = cluster.leader.timeline
+        self._leader_timeline = None if cluster.is_unlocked() else cluster.leader.timeline
 
     def acquire_lock(self):
         return self.dcs.attempt_to_acquire_leader()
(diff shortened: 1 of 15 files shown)


patroni/postgresql.py
```diff
@@ -1252,28 +1279,44 @@ def _check_timeline_and_lsn(self, leader):
...
+
+    def replica_cached_timeline(self, master_timeline):
+        if not self._cached_replica_timeline or not master_timeline or self._cached_replica_timeline != master_timeline:
+            self._cached_replica_timeline = self.get_replica_timeline()
+        return self._cached_replica_timeline
+
+    def get_master_timeline(self):
+        return self._cluster_info_state_get('timeline')
+
+    def get_history(self, timeline=None):
```
alexeyklyukin: the optional second param doesn't make sense here, as the function is called only from one place with both params set (also, if timeline is None, timeline -1 is an error).

patroni/ha.py
```diff
@@ -372,6 +396,9 @@ def enforce_master_role(self, message, promote_message):
                 self.release_leader_key_voluntarily()
                 return 'Not promoting self because watchdog could not be activated'
 
+        if self.state_handler.get_master_timeline() > 1:
```
alexeyklyukin: I wonder if it makes sense to also erase the history stored in the DCS when we detect the timeline 1, as we don't want to carry the leftovers of a pre-pg_upgrade history over to the new cluster.

alexeyklyukin:
Looks good. I think it makes sense to enrich the history further (perhaps in subsequent PRs), for instance, adding the current leader name to the leader_timeline - 1. Also, as it stands, the history is nit exposed in patronictl list -x, but probably it makes sense to do so, at least showing the date of the latest promotion.

I'm not sure I agree with the idea of having a separate replica timeline cache in addition to the cached information stored in the cluster_info_state, why not use the cluster_info_state['cached_timeline'] for that, or even revert back to the idea of having the 'leader' status separate from the timeline, as checking for the timeline = 0 instead of leader = 'f' is potentially more error-prone. OTOH, that's not a big deal.

alexeyklyukin:
👍

CyberDem0n:
:+1:

Issue #349, opened 2016-10-27
patronictl pause does not go into effect inmediately


zenitraM:
Steps to reproduce:
- `patronictl pause` on a leader
- restart inmediately the Patroni service on it (i.e, to upgrade Patroni without causing a Postgres failover)

It seems the pause is not applied until the next time the HA loop (10s?) runs, so restarting Patroni before that happens causes Patroni to kill Postgres. The other nodes see the cluster as paused and will not do anything about it.

I haven't found the `paused` state exposed on the API so we can't query it to see if it was applied properly on all nodes - our workaround for now is sleeping some prudential time after the pause command was issued.

CyberDem0n:
`pause: true` is stored in a `<cluster>/config` and works cluster-wide, but every node reads and applies it only on the next run of ha loop.
Basically we have only two options here:
 - since pause change request if always processed my the master we can "wake-up" the main HA loop and master will recognize change of pause state nearly immediately.
 - we can sleep(loop_wait) seconds in the patronictl after changing of pause state to make sure that every node had enough time to apply changes.

zenitraM:
The first one will solve the master but secondaries might not have been gotten the change yet...

Maybe either the second option or just changing the message to "Pause request sent" to make it more clear that pause/unpause are asynchronous requests might do it

ants:
What about trying to fetch cluster state from the DCS before shut down? That would mean that we would observe relatively recent state when making decisions both in the HA loop or in shutdown.

There is still a race in that the HA loop might be in the middle of running when pause is set. So something this ordering is still possible, even if unlikely:

* Ha loop fetches cluster state
* Set cluster to pause
* Stop Postgresql
* Ha loop discovers that Postgresql is not running and starts it, because it still thinks pause is on.

One idea would be to have `touch_member()` set a timestamp, a `pause --wait` mode would then check the member list for last check-in timestamps, set pause mode and then poll for members to either update their timestamp or go away. This might have some gotchas related to DCS consistency models.

zenitraM:
And what about adding a way to check on a node's API if the currently running HA loop is running on the "paused" state? - or make them post on their DCS check-in their state.

`pause` could then wait until all nodes have explicitly posted that they are paused - same could be applied for `resume`.

PR #592, opened 2018-01-05
Expose pause state of every member to DCS and via REST

CyberDem0n:
and implement patronictl pause|resume --wait on top of that

Fixes https://github.com/zalando/patroni/issues/349


alexeyklyukin:
👍

CyberDem0n:
:+1:

PR #601, opened 2018-01-17
 A few iprovements in patronictl

CyberDem0n:
* make switchover work with old patroni
* exclude leader from candidates when interactively running failover


alexeyklyukin:
👍

CyberDem0n:
:+1:

PR #604, opened 2018-01-19
Allow options to the basebackup built-in method.

alexeyklyukin:
Options should be specified in the basebackup section, which is
optional. Changing some options (i.e. connection arguments) will
break the basebackup procedure. No validation on option names/values
if performed, the proper validation is possible, but it  would require
the major PG verison of the pg_basebackup. Currently no support for
specifying the same option multiple time with different values, i.e
tablespace-mapping.

Add the corresponding documentation paragraph.

In additin, fix the exception condition when the parameter section
for a given replica creation method is empty.


docs/replica_bootstrap.rst
```diff
@@ -94,4 +95,11 @@ A special ``no_master`` parameter, if defined, allows Patroni to call the replic
 running master or replicas. In that case, an empty string will be passed in a connection string. This is useful for
 restoring the formerly running cluster from the binary backup.
 
+A ``basebackup`` method is a special case: it will be used if ``create_replica_method`` is empty, although it is possible
```
CyberDem0n: I think we should also mention here `clonefrom` tag.

Patroni will prefer to use any existing replica tagged with clonefrom to take basebackup from it rather than from leader.

patroni/postgresql.py
```diff
@@ -665,7 +665,9 @@ def create_replica(self, clone_member):
                     break
             # if the method is basebackup, then use the built-in
             if replica_method == "basebackup":
-                ret = self.basebackup(connstring, env)
+                base_backup_options = (self.config[replica_method].copy()
+                                       if self.config.get(replica_method, {}) else {})
+                ret = self.basebackup(connstring, env, base_backup_options)
```
CyberDem0n: It could be just a single line:
```python
                ret = self.basebackup(connstring, env, (self.config.get(replica_method) or {})).copy())
```

Or may be two two line, but in any case `if ... else` construction is not really useful here.

patroni/postgresql.py
```diff
@@ -1614,23 +1616,27 @@ def remove_data_directory(self):
...
         # creates a replica data dir using pg_basebackup.
         # this is the default, built-in create_replica_method
         # tries twice, then returns failure (as 1)
         # uses "stream" as the xlog-method to avoid sync issues
+        # supports additional user-supplied options, those are not validated
         maxfailures = 2
         ret = 1
+        user_options = []
+        for k, v in options.items():
+            if v != '':
```
CyberDem0n: What if `v is None?`

We can also do a list comprehension here:
```python
        user_options = ['--{0}={1}'.format(k, v) for k, v in options.items() if v]
```

docs/replica_bootstrap.rst
alexeyklyukin: Good point

patroni/postgresql.py
alexeyklyukin: Yeah, good point on that as well. I actually wanted to change that anyway in order to support v being a list of options (to handle the `tablespace-mapping` option properly)

docs/replica_bootstrap.rst
```diff
@@ -72,13 +72,14 @@ scripts to clone a new replica. Those are configured in the ``postgresql`` confi
...
+        basebackup:
+            max-rate: '100M'
 
 
 The ``create_replica_method`` defines available replica creation methods and the order of executing them. Patroni will
-stop on the first one that returns 0. The basebackup is the built-in method and doesn't require any configuration. The
-rest of the methods should define a separate section in the configuration file, listing the command to execute and any
-custom parameters that should be passed to that command. All parameters will be passed in a ``--name=value`` format.
-Besides user-defined parameters, Patroni supplies a couple of cluster-specific ones:
+stop on the first one that returns 0. Each method should a separate section in the configuration file, listing the command
```
erthalion: `Each method should *define* a separate section`?

CyberDem0n:
Tests are failing: https://travis-ci.org/zalando/patroni/jobs/371580095#L1209

patroni/postgresql.py
```diff
@@ -660,7 +661,7 @@ def create_replica(self, clone_member):
                     break
             # if the method is basebackup, then use the built-in
             if replica_method == "basebackup":
-                ret = self.basebackup(connstring, env)
+                ret = self.basebackup(connstring, env, copy_list_dict(self.config.get(replica_method) or {}))
```
CyberDem0n: I don't really understand why do you want to make a copy of `config[replica_method]`, `basebackup` doesn't change anything in it...
In case if you really want to protect it from changes, it makes sense to use `copy.deepcopy` instead of reinventing the wheel.

alexeyklyukin: Seems to be a stupid copy/paste thing. Thanks.

CyberDem0n:
:+1:

alexeyklyukin:
👍

CyberDem0n:
:+1:

Issue #609, opened 2018-01-24
System incorrectly rebooted when patroni service stopped with watchdog enabled.


akissa:
A regression has been introduced by commit 5da0e12353a2e2462437ff36c6512d6ae8ea7142

This is because it is possible for an exception to be raised at https://github.com/zalando/patroni/commit/5da0e12353a2e2462437ff36c6512d6ae8ea7142#diff-74fbda32f108239eb16ce7417ef10085R85

The  exception is (`psutil.NoSuchProcess`)(`psutil.NoSuchProcess process no longer exists`).

This means the watchdog is not correctly closed. Which then causes the system to get reset.

It can be replicated easily by enabling the watchdog then stop the patroni service.

ants:
Thanks for the report. I will come up with a fix for this.

PR #610, opened 2018-01-24
Handle exceptions raised from psutil

CyberDem0n:
Process.cmdline can raise `NoSuchProcess` or `AccessDenied`
Process.children can raise `AccessDenied`

Fixes https://github.com/zalando/patroni/issues/609


CyberDem0n:
@ants, what do you think?

ants:
Looks good.

I had an academic question about what is the correct behavior when accessing a childs command line gives AccessDenied, for safety we should then fall back to waiting on the postmaster before triggering safepoint. But I don't see how AccessDenied (EPERM, EACCESS) can even happen when checking a processes commandline so I guess it's not even worth worrying about.

CyberDem0n:
👍

alexeyklyukin:
👍

PR #616, opened 2018-01-27
Add proper rbac to run patroni on k8s

soltysh:
This PR adds 3 resources that will properly setup the RBAC:

1. service account, which is also assigned to the pods of the cluster, so that they use those particular permissions 
2. a role, which holds only the necessary permissions that patroni members need to interact with k8s cluster
3. a rolebinding, which connects two two former things together to use.

The role and rolebinding was created using this tool https://github.com/liggitt/audit2rbac which looks at [audit logs](https://kubernetes.io/docs/tasks/debug-application-cluster/audit/#advanced-audit) provided  by the api server.

### DIFF
--- a/kubernetes/patroni_k8s.yaml
+++ b/kubernetes/patroni_k8s.yaml
@@ -14,6 +14,7 @@ spec:
         application: patroni
         cluster-name: *cluster_name
     spec:
+      serviceAccountName: patronidemo
       containers:
       - name: *cluster_name
         image: patroni  # docker build -t patroni .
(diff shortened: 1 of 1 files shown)


soltysh:
@jberkus @k1-hedayati ptal

kubernetes/patroni_k8s.yaml
```diff
@@ -120,3 +121,60 @@ type: Opaque
...
+  - patch
+  - update
+  - watch
+- apiGroups:
+  - ""
+  resourceNames:
+  - patronidemo
+  resources:
+  - endpoints
+  verbs:
```
CyberDem0n: Patroni also needs to be able to `list` and `watch` Endpoints if `kubernetes.use_endpoint: true`
Also in this case it doesn't need have an access to `ConfigMaps`.

kubernetes/patroni_k8s.yaml
```diff
@@ -120,3 +121,60 @@ type: Opaque
...
+- apiGroups:
+  - ""
+  resources:
+  - pods
+  verbs:
+  - get
+  - list
+  - patch
+  - update
+  - watch
```
CyberDem0n: Patroni never watch on `Pods`

kubernetes/patroni_k8s.yaml
soltysh: Hmmm... The options are:
1. document both approaches with roles specific for each use case
2. create uber role with all in
wdyt?

soltysh: For the demo example I think I'd go with the latter with comment that this is for configmap approach and the other is for endpoints.

kubernetes/patroni_k8s.yaml
soltysh: It's more of an approach we in OpenShift take, that whenever you have list a watch is just there. That's also what's represented in the tool I've mentioned and why this is here.

soltysh: From a security pov having either or both is almost the same.

kubernetes/patroni_k8s.yaml
jberkus: why not do the grants for both things in the default roles/template, and document how to restrict them based on which mode you're running in?

soltysh: Yeah, I'll update it later today. That's I guess the best approach.

soltysh:
I've updated the roles to include the ones needed for endpoints, and added appropriate comment wrt when they are needed. I've also dropped one more limitation, previously there was an exact name for an endpoint, which will only confuse users.

soltysh:
@CyberDem0n ptal

alexeyklyukin:
👍

CyberDem0n:
👍

CyberDem0n:
Merged, thank you @soltysh

PR #622, opened 2018-02-07
Make log level configurable from environment variables

antineutron:
Found this helpful when debugging some kubernetes issues - lets you turn up the logging levels on Patroni itself and the requests library.

### DIFF
--- a/patroni/__init__.py
+++ b/patroni/__init__.py
@@ -135,8 +135,10 @@ class Patroni(object):
 
 def patroni_main():
     logformat = os.environ.get('PATRONI_LOGFORMAT', '%(asctime)s %(levelname)s: %(message)s')
-    logging.basicConfig(format=logformat, level=logging.INFO)
-    logging.getLogger('requests').setLevel(logging.WARNING)
+    loglevel = os.environ.get('PATRONI_LOGLEVEL', 'INFO')
+    requests_loglevel = os.environ.get('PATRONI_REQUESTS_LOGLEVEL', 'WARNING')
+    logging.basicConfig(format=logformat, level=logging.getLevelName(loglevel))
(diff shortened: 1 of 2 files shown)


patroni/__init__.py
```diff
@@ -135,8 +135,10 @@ def shutdown(self):
 
 def patroni_main():
     logformat = os.environ.get('PATRONI_LOGFORMAT', '%(asctime)s %(levelname)s: %(message)s')
-    logging.basicConfig(format=logformat, level=logging.INFO)
-    logging.getLogger('requests').setLevel(logging.WARNING)
+    loglevel = os.environ.get('PATRONI_LOGLEVEL', 'INFO')
+    requests_loglevel = os.environ.get('PATRONI_REQUESTS_LOGLEVEL', 'WARNING')
+    logging.basicConfig(format=logformat, level=logging.getLevelName(loglevel))
```
CyberDem0n:
```
getLevelName(level)
    Return the textual representation of logging level 'level'.

    If the level is one of the predefined levels (CRITICAL, ERROR, WARNING,
    INFO, DEBUG) then you get the corresponding string. If you have
    associated levels with names using addLevelName then the name you have
    associated with 'level' is returned.
    
    If a numeric value corresponding to one of the defined levels is passed
    in, the corresponding string representation is returned.
    
    Otherwise, the string "Level %s" % level is returned.
```

Since we are expecting a string (log level name) here, it doesn't make much sense to call `getLevelName`

antineutron: Hah, not sure why I used that! I'll sort it out when I've got a few minutes

CyberDem0n:
:+1:

Jan-M:
:+1:

CyberDem0n:
Merged, thank you!

Issue #628, opened 2018-02-16
Patroni not trying to reconnect to other etcd members


bradnicholson:
I'm seeing a bit of an odd case under test that I can't quite explain.  Wondering if you have any ideas.

This is with Patroni 1.3.6 with Python 2.7

We have a deployment with 2 Postgres Members and 3 etcd members deployed across three AZ's, lets call them AZ's X, Y, and Z

We have etcd's in each AZ, assume the name matches the AZ
etcdX
ectdY
etcdZ

and Postgres in two AZ's
pgX
pgY

All three etcd members are up and have formed a cluster by the time Patroni starts on both PG members.  We have a connection to etcd hard coded to connect to etcdX on startup as such

```
etcd:
  host: <etcdX_IP>:4001
```

pgX becomes the master, pgY becomes a follower.

To test the failure, we fence off  AZ X at the network layer from AZ's Y & Z. 

pgX times out communicating with etcdX (presumably due to lack of quorum reads).  It attempts to connect to etcdY and etcdZ and fails as expected.  pgX goes read-only as expected.

pgY loses connectivity with etcdX.  It does not try and reconnect to etcdY or etcdZ.  It  simply keeps retrying to connect to etcdX.

I'd expect that pgY should also be trying to reconnect to one of the other etcd members, but it does not.  Any idea why?

We do plan on upgrading to the latest version and giving it the entire list of etcd hosts on startup, but I'm still interested in digging into this failure more.

CyberDem0n:
> pgY loses connectivity with etcdX. It does not try and reconnect to etcdY or etcdZ. It simply keeps retrying to connect to etcdX.

That's very strange. Up on start Patroni discovers topology of etcd cluster by performing request `GET http://ip:4001/v2/machines`. It should get in response three urls separated by comma (http://etcdX:4001,http://etcdY:4001,http://etcdZ:4001).
Can you verify that it really returns it?

In addition to that, to protect from the situation when all nodes except the current one were rotated with change of ip addresses, Patroni periodically (every 5 minutes) tries to rediscover etcd cluster topology (GET /v2/machines).

> I'd expect that pgY should also be trying to reconnect to one of the other etcd members, but it does not. Any idea why?

It should try with other members. But for some reason it didn't got the right topology from etcd.

> We do plan on upgrading to the latest version and giving it the entire list of etcd hosts on startup, but I'm still interested in digging into this failure more.

[...]

PR #630, opened 2018-02-23
Don't rediscover etcd cluster topology when watch timed out

CyberDem0n:
but switch to the next node if it is possible.

Fixes https://github.com/zalando/patroni/issues/628


alexeyklyukin:
👍

CyberDem0n:
:+1:

Issue #631, opened 2018-02-23
When using Custom Bootstrap, Patroni does not write pg_hba values from the patroni conf file


bradnicholson:
When using the custom bootstrap file, Patroni does not use the pg_hba.conf values in the patroni.yml file.

I did a basic A/B test.  First without the custom bootstrap being called.  This test wrote the pg_hba.conf file correctly with the custom values from the patroni config file.

The second test, I enabled a basic custom bootstrap that just calls initdb directly.  I left the rest of the patroni.yml file unchanged.  This time, it did not write the pg_hba values from the patroni config file to the pg_hba.conf file.

CyberDem0n:
Hi @bradnicholson,

can you provide part of the config file with `pg_hba`?

bradnicholson:
```
  initdb:  # Note: It needs to be a list (some options need values, others are switches)
  - encoding: UTF8
  - data-checksums
  pg_hba:  # Add following lines to pg_hba.conf after running 'initdb'
    - local      all             all                                              trust
    - host       all             all                      ::1/128                 trust
    - host       all             all                      0.0.0.0/0               trust
    - host       replication    replication-user      0.0.0.0/0               trust
```

PR #632, opened 2018-02-26
Write bootstrap.pg_hba into a pg_hba.conf after custom bootstrap

CyberDem0n:
Fixes https://github.com/zalando/patroni/issues/631

### DIFF
--- a/patroni/postgresql.py
+++ b/patroni/postgresql.py
@@ -537,12 +537,7 @@ class Postgresql(object):
         ret = self.pg_ctl('initdb', *options)
         if pwfile:
             os.remove(pwfile)
-        if ret:
-            if not self._server_parameters.get('hba_file') and not self.config.get('pg_hba'):
-                self.write_pg_hba(config.get('pg_hba', []))
-            self._major_version = self.get_major_version()
-            self._server_parameters = self.get_server_parameters(self.config)
-        else:
+        if not ret:
             self.set_state('initdb failed')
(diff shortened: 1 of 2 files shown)


patroni/postgresql.py
```diff
@@ -1547,14 +1544,16 @@ def clone(self, clone_member):
...
+        pg_hba = config.get('pg_hba', [])
         method = config.get('method') or 'initdb'
         self._running_custom_bootstrap = method != 'initdb' and method in config and 'command' in config[method]
         if self._running_custom_bootstrap:
             do_initialize = self._custom_bootstrap
             config = config[method]
         else:
             do_initialize = self._initdb
-        return do_initialize(config) and self._configure_server_parameters() and self.start()
+        return do_initialize(config) and self.write_pg_hba(pg_hba) and self.save_configuration_files() \
```
alexeyklyukin: Would it makes sense to look for pg_hba.conf from the config[method] at this point (allowing  to configure per-bootstrap pg_hba.conf) and falling back to the 'generic' pg_hba.conf in the bootstrap only if the method-specific one is not defined?

CyberDem0n: I don't really want to implement anything new here, just want to fix a bug.
Taking into account that we can't use more than one custom bootstrap method it's not really worth it to make it possible to allow it to be configurable per section.

CyberDem0n:
:+1:

Issue #637, opened 2018-03-07
Make Deletion of recovery.conf Optional


bradnicholson:
When bootstrapping a cluster with the custom_replica creation, Patroni automatically deletes the  recovery.conf if it exists.  

When bootstrapping a replica from a pgBackRest restore, we would like to let pgBackRest generate the appropriate recovery.conf and have Patroni leave it in place.

To accomplish that, I'd like to add an optional parameter called `keep_existing_recovery_conf` (as below).  If that parameter is present and set to `True`, skip the deletion of the recovery.conf

```
bootstrap:
    method: <custom_bootstrap_method_name>
    <custom_bootstrap_method_name>:
        command: <path_to_custom_bootstrap_script> [param1 [, ...]]
        keep_existing_recovery_conf: True
```

CyberDem0n:
It's very easy to implement, but should we really to that?
Patroni already gives you all control over recovery.conf, it is possible to configure what it will put inside of it after running custom bootstrap script:

```yaml
bootstrap:
    method: <custom_bootstrap_method_name>
    <custom_bootstrap_method_name>:
        command: <path_to_custom_bootstrap_script> [param1 [, ...]]
        recovery_conf:
            recovery_target_action: promote
            recovery_target_timeline: latest
            restore_command: <method_specific_restore_command>
```

Basically you can build any custom recovery.conf just by adding more options into `bootstrap.<custom_bootstrap_method_name>.recovery_conf`

bradnicholson:
> It's very easy to implement, but should we really to that?

It seems unnatural to me when working with pgBackRest, as well as duplication of logic.  We already have  pgBackRest automatically generating the proper recovery.conf based on the parameters we give that.  Those parameters get passed in dynamically to support a number of different restore options.

Without this feature, we have to duplicate pgBackRest's recovery.conf creation and embed that into the patroni.yml configuration.

bradnicholson:
The other part I don't like about embedding this into the patroni.yml is this. 

Say I'm doing a PITR restore from Cluster A into Cluster B.  I need to give put transient information about cluster A into Cluster B's patroni.yml (in the `restore_command`).   The restore configuration should be temporary during the restore operation. 

That would dictate editing of patroni.yml post restore to remove that data, which is less than ideal.

PR #638, opened 2018-03-08
Make deleting recovery.conf optional.

bradnicholson:
https://github.com/zalando/patroni/issues/637

pgBackRest's restore command generates the appropriate recovery.conf based
on the parameters you provide to pgBackRest.  When calling pgBackRest's restore command
via Patroni's custom bootstrap, it deletes that recovery.conf.  Specifying the recovery.conf
information in the patroni.yml is less than ideal.  It prevent's leveraging pgBackRests
work to ensure recovery.conf files are properly generated.  It also can lead to transient
config data in the patroni.yml under certain restore cases, such as a PITR restore
of Cluster B to  Cluster A, where the restore_commnand in A needs to reference B.

The parameter is optional.  The default behavior is to delete the recovery.conf.

### DIFF
--- a/docs/replica_bootstrap.rst
+++ b/docs/replica_bootstrap.rst
@@ -48,8 +48,8 @@ If a ``recovery_conf`` block is defined in the same section as the custom bootst
 ``recovery.conf`` before starting the newly bootstrapped instance. Typically, such recovery.conf should contain at least
 one of the ``recovery_target_*`` parameters, together with the ``recovery_target_timeline`` set to ``promote``.
 
-If ``keep_existing_recovery_conf`` is defined and set to ``True``, Patroni will not remove the existing recovery.conf file if it exists.
-This is useful when bootstrap from a backup with tools like pgBackRest that generate the appropriate recovery.conf for you.
+If ``keep_existing_recovery_conf`` is defined and set to ``True``, Patroni will not remove the existing ``recovery.conf`` file if it exists.
+This is useful when bootstrapping from a backup with tools like pgBackRest that generate the appropriate ``recovery.conf`` ``for you.
 
  .. note:: Bootstrap methods are neither chained, nor fallen-back to the default one in case the primary one fails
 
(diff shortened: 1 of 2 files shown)


patroni/postgresql.py
```diff
@@ -557,7 +561,8 @@ def _custom_bootstrap(self, config):
         if 'recovery_conf' in config:
             self.write_recovery_conf(config['recovery_conf'])
         elif os.path.isfile(self._recovery_conf) or os.path.islink(self._recovery_conf):
-            os.unlink(self._recovery_conf)
+            if not keep_existing_recovery_conf:
```
CyberDem0n: I think it could be simplified to:
```python
        elif (os.path.isfile(self._recovery_conf) or os.path.islink(self._recovery_conf)) and \
                not config.get('keep_existing_recovery_conf'):
            os.unlink(self._recovery_conf)
```

And lines 547-550 could be dropped.

docs/replica_bootstrap.rst
```diff
@@ -47,6 +48,9 @@ If a ``recovery_conf`` block is defined in the same section as the custom bootst
 ``recovery.conf`` before starting the newly bootstrapped instance. Typically, such recovery.conf should contain at least
 one of the ``recovery_target_*`` parameters, together with the ``recovery_target_timeline`` set to ``promote``.
 
+If ``keep_existing_recovery_conf`` is defined and set to ``True``, Patroni will not remove the existing recovery.conf file if it exists.
+This is useful when bootstrap from a backup with tools like pgBackRest that generate the appropriate recovery.conf for you.
```
avaczi: bootstrap_ping_, and \`\`recovery.conf\`\` (as in the previous paragraphs)

avaczi:
:+1:

CyberDem0n:
:+1:

CyberDem0n:
Merged, thank you!

PR #644, opened 2018-03-22
Modified kubernetes/entrypoint.sh to allow local replication connecti…

jberkus:
…ons, needed to support _get_local_timeline_lsn_from_replication_connection().


jberkus:
@CyberDem0n hmmm, given that this change *isn't* needed for 10, I think I'm going to change it to have the line, but commented out.

Issue #649, opened 2018-03-27
Removing invalid parameter for timezone and datestyle


jouir:
Hello,

When you configure client settings in postgresql.conf like `timezone` and `datestyle` in the postgresql.parameters' section of patroni, you end up with the following warnings at reload:
```
WARNING: Removing invalid parameter `timezone` from postgresql.parameters
WARNING: Removing invalid parameter `datestyle` from postgresql.parameters
```
The message is thrown because they are not present in the `pg_settings` table, because they are client settings but they are also valid postgresql.conf settings.

In practice, those settings are written in local files and PostgreSQL load them perfectly but the message could be misleading. Can we safely ignore the two warnings?

CyberDem0n:
No, you can't ignore it. It actually means those settings are not applied...
The problem is that in pg_settings those parameters are called `TimeZone` and `DateStyle` accordingly.

The easiest fix for you would be to fix a config by changing the name.
Actually there is at least one more parameter, which name is written in CamelCase: https://www.postgresql.org/docs/current/static/runtime-config-client.html#GUC-DATESTYLE

It looks like we should really do case insensitive comparison.

jouir:
It works using `TimeZone` and `DateStyle`. Thanks!

PR #650, opened 2018-03-28
Treat postgres settings parameter names as case insensitive

CyberDem0n:
Because they are indeed case insensitive.
Most of the parameters have snake_case_name, but there are three exceptions from this rule: DateStyle, IntervalStyle and TimeZone.
In fact, if you specify timezone = 'some/tzn' it still works, but Patroni wasn't able to find 'timezone' in pg_settings and stripping this parameter out.

We will use CaseInsensitiveDict to keep postgresql.parameters. This change affects only "final" configuration. That means if you put some"duplicates" (work_mem vs WORK_MEM) into patroni yaml or into cluster config, it would be resolved only at the last stage and for example you will be able to see both values if you use `patronictl edit-config`.

Fixes https://github.com/zalando/patroni/issues/649


alexeyklyukin:
👍

Issue #642, opened 2018-03-21
Error in etcd.py with python 2.7.5


Aytuar:
versions
patroni v1.4.3
python 2.7.5 
CentOS 7.4.1708

```
Traceback (most recent call last):
  File "/opt/patroni/patronictl.py", line 5, in <module>
    ctl(None)
  File "/usr/lib/python2.7/site-packages/click/core.py", line 722, in __call__
    return self.main(*args, **kwargs)
  File "/usr/lib/python2.7/site-packages/click/core.py", line 697, in main
    rv = self.invoke(ctx)
  File "/usr/lib/python2.7/site-packages/click/core.py", line 1066, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
  File "/usr/lib/python2.7/site-packages/click/core.py", line 895, in invoke
    return ctx.invoke(self.callback, **ctx.params)
  File "/usr/lib/python2.7/site-packages/click/core.py", line 535, in invoke
    return callback(*args, **kwargs)
  File "/usr/lib/python2.7/site-packages/click/decorators.py", line 27, in new_func
    return f(get_current_context().obj, *args, **kwargs)
  File "/opt/patroni/patroni/ctl.py", line 748, in members
    dcs = get_dcs(obj, cluster_name)
  File "/opt/patroni/patroni/ctl.py", line 121, in get_dcs
    return _get_dcs(config)
  File "/opt/patroni/patroni/dcs/__init__.py", line 58, in get_dcs
    module = importlib.import_module(module_name)
  File "/usr/lib64/python2.7/importlib/__init__.py", line 37, in import_module
    __import__(name)
  File "/opt/patroni/patroni/dcs/etcd.py", line 81, in <module>
    class Client(etcd.Client):
AttributeError: 'module' object has no attribute 'Client'

```

CyberDem0n:
Hmm,

do you get such error only with patronictl.py or with patroni.py either?

Aytuar:
Yes, the error also occurs separately when importing into the python.
module = importlib.import_module(module_name)
At the same time, there is no error on Python 2.7.14.

Aytuar:
So. I install all packets from requirements.txt in the virtualenv. Here the list of pip freeze.
```
boto==2.48.0
cachetools==2.0.1
cdiff==1.0
certifi==2018.1.18
chardet==3.0.4
click==6.7
dnspython==1.15.0
google-auth==1.4.1
idna==2.6
ipaddress==1.0.19
[...]

PR #651, opened 2018-03-28
Pass dict object to touch_member instead of json encoded string

CyberDem0n:
DCS implementation will take care about encoding it.
Fixes https://github.com/zalando/patroni/issues/642


alexeyklyukin:
👍

CyberDem0n:
:+1:

PR #658, opened 2018-04-09
Fix race condition in poll_failover_result

CyberDem0n:
It didn't affect directly neither failover nor switchover, but in some rare cases it was reporting it as a success too early, when the former leader released the lock: `Failed over to "None" instead of "desired-node"`

In addition to that this commit improves logs and status messages by differentiating between failover and switchover.


alexeyklyukin:
👍

CyberDem0n:
:+1:

Issue #655, opened 2018-04-05
Avoid an accidental replica promotion when attaching Patroni to a running instance


alexeyklyukin:
Patroni can attach itself to an already running PostgreSQL instance. If that is the first instance "seen" in the given cluster, Patroni for that instance will create the initialize key, grab the leader key and, if the instance is running a replica, promote.

Because of this behaviour, when a cluster with a master and one or more replicas gets Patroni for each node, it is imperative to start running Patroni on the master node before getting to the replicas. Otherwise, Patroni will promote the first replica before the former master is even known to the cluster in DCS, resulting in an instant split-brain.) We can do something in Patroni itself to avoid such unwarranted promotions: so far the idea is to stop right away when observing the running PostgreSQL replica during the start of Patroni when the initialize key is missing in DCS. Perhaps we can do even more and do the same if the replica is not running, but has a valid data directory and recovery.conf.

PR #661, opened 2018-04-11
Abort start if attaching to running postgres and cluster not initiazlied

CyberDem0n:
Patroni can attach itself to an already running PostgreSQL instance. If that is the first instance "seen" in the given cluster, Patroni for that instance will create the initialize key, grab the leader key and, if the
instance is running a replica, promote.

Because of this behavior, when a cluster with a master and one or more replicas gets Patroni for each node, it is imperative to start running Patroni on the master node before getting to the replicas.

This commit changes such weird behavior and will abort Patroni start if there is no initialize key in DCS and postgres is running as replica.

Closes https://github.com/zalando/patroni/issues/655

### DIFF
--- a/patroni/ha.py
+++ b/patroni/ha.py
@@ -72,6 +72,10 @@ class Ha(object):
         # standby. Changes protected by _member_state_lock.
         self._disable_sync = 0
 
+        # We need following property to avoid shutdown of postgres when join of Patroni to the postgres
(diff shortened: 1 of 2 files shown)


alexeyklyukin:
👍

CyberDem0n:
:+1:

PR #669, opened 2018-04-23
Do not exit when encountering invalid system ID.

alexeyklyukin:
Do not exit when the cluster system ID is empty or the one that
doesn't pass the validation check. In that case, the cluster most
likely needs a reinit; mention it in the result message. Avoid
terminating Patroni, as otherwise reinit cannot happen.


CyberDem0n:
:+1:

alexeyklyukin:
👍

PR #672, opened 2018-05-01
Quorum commit feature

ants:
I did some work on the subject, did not have enough time to get it finished, but far enough to solicit some feedback.

## Description

The feature generalizes synchronous_mode to multiple nodes being in sync. Trade-off between synchronization overhead and fault tolerance is user selectable via `replication_factor` configuration parameter, specifying how many nodes must contain a commit before it is acknowledged.

Cluster will retain self-healing capability if within one HA loop interval up to `replication_factor - 1` nodes fail.

The old `synchronous_commit` corresponds to `replication_factor = 2`.

On PostgreSQL 10 the quorum commit `ANY k (n)` feature is used. On 9.6 the `k (n)` variant is used picking first nodes. On even older versions maximum `replication_factor` is 2. However, this is still an improvement from status quo as it delegates sync standby selection to PostgreSQL reducing commit latency on sync standby failure.

DCS sync state is changed from `{"leader": .., "sync_standby": ..}` to `{"leader": .., "quorum": .., "members": [..]}`, where leader is kept mainly as an informative field.

## State

The main quorum management engine should be in a reasonable state although I haven't tested it yet.

Open issues:

* Is `replication_factor` a reasonable UI for this? Do we want to keep `synchronous_commit` as the main switch for this functionality.
* Nodes that aren't actually taking part in synchronous replication can still promote themselves if they can verify their wal position with enough nodes that are taking part. Does anyone see a problem with this? It causes minor extra complexity when promoting such node, but on the other hand simplifies health checks.
* I could not figure out a way how to find out from PostgreSQL when `synchronous_standby_names` becomes active. Currently there is a small race condition there. I'm not even sure if the previous check of pg_stat_replication showing walsender as sync was enough as a cursory review of postgres code hints that a walsender might think it is sync while a backend hasn't gotten the message that sync replication was enabled. More study of the code is needed.
* Quorum engine manages transitioning DCS and sync state between "safe" states. If somebody manually modifies state into something we don't think of as valid, there currently isn't any code to automatically recover. Should be simple enough, I just didn't get to it.
* State handling when promoting could be improved. I think we want to keep synchronous replication running with mostly the same settings as before failover, but remove any nodes that we know are down based on DCS information. Speeds up being able to accept commits.
* Testsuite is obviously broken. Needs fixing up of old tests and new tests to exercise corner cases of the new stuff.
* patronictl UI needs updating. How do we want to display the information?
* There is an impedance mismatch between quorum engine and `synchronous_standby_names` in that quorum engine thinks of leader as just one additional synchronous replica. Currently the boundary transition between two worlds is in postgresql.py. Maybe some other place would work better.
* flake8 is probably going to have a fit on this code. A bunch of cleanup is needed.


patroni/postgresql.py
```diff
@@ -1645,52 +1717,77 @@ def basebackup(self, conn_url, env):
...
+        """
+
+        if num == 1 or num is None:
+            # Turn off sync replication
+            assert not sync or sync == set([self.name])
+            ssn = None
+        else:
+            assert self.name in sync
+            sync_standbys = sync.difference([self.name])
+            standby_list = ", ".join(sorted(sync_standbys)) if sync_standbys else "*"
```
CyberDem0n: Lost quote_ident.

patroni/postgresql.py
```diff
@@ -1645,52 +1717,77 @@ def basebackup(self, conn_url, env):
...
+    def current_sync_state(self, cluster):
+        """Returns current synchronous replication state as a dict:
+        * numsync: number of synchronous nodes requested(including leader)
+        * sync: set of nodes potentially being synced to
+        * active: set of nodes that are sync capable
         """
-        current = cluster.sync.sync_standby
-        current = current.lower() if current else current
+        sync_standby_names = self.query("SHOW synchronous_standby_names")
+        result = parse_sync_standby_names(sync_standby_names)
```
CyberDem0n: Why do we need to parse it all the time? Is it some kind of protection from unexpected changes from outside?

patroni/quorum.py
```diff
@@ -0,0 +1,138 @@
...
+                yield self.quorum_update(self.quorum, self.voters - to_remove)
+                yield self.sync_update(self.numsync - len(to_remove), self.sync - to_remove)
+
+        # If any new nodes, join them to quorum
+        to_add = self.active - self.sync
+        if to_add:
+            # First get to requested replication factor
+            increase_numsync_by = self.sync_wanted - self.numsync
+            if increase_numsync_by:
+                add = set(sorted(to_add)[:increase_numsync_by])
```
CyberDem0n: This one I don't really understand.

It is responsible for https://github.com/zalando/patroni/pull/672/files#diff-57e7f41a6358e844d5245ea7b8409311R12
```python
+        self.assertEquals(list(QuorumStateResolver(1, set("a"), 1, set("a"), active=set("abcde"), sync_wanted=3)), [
+            ('sync', 3, set('abc')),
+            ('quorum', 3, set('abcde')),
+            ('sync', 3, set('abcde')),
+        ])
```

Why does it fist adds only 'b' and 'c'? Why not all 4 at the same time? Something like:
```python
+            ('sync', 3, set('abcde')),
+            ('quorum', 3, set('abcde')),
```

patroni/ha.py
```diff
@@ -293,54 +294,103 @@ def is_synchronous_mode(self):
...
+                                                              sync_wanted=sync_wanted):
+                if transition == 'quorum':
+                    logger.info("Setting quorum to %d of %d (%s)", num, len(nodes), ", ".join(sorted(nodes)))
+                    if not self.dcs.write_sync_state(leader=self.state_handler.name, quorum=num, members=list(nodes),
+                                                     index=self.cluster.sync.index):
+                        logger.info('Synchronous replication key updated by someone else.')
                         return
-                    logger.info("Synchronous standby status assigned to %s", picked)
+                elif transition == 'sync':
+                    logger.info("Setting synchronous replication to %d of %d (%s)",
```
CyberDem0n: This message is misleading when `num < min_sync`.

patroni/postgresql.py
ants: Thanks for catching this. I will add it back in.

patroni/postgresql.py
ants: Basically yes. It helps us automatically resolve the situation if due to a bug or administrator intervention our understanding of the state goes out of sync. If we aren't too worried about that we could cache the state and avoid spamming the database with unnecessary queries. What do you think?

patroni/ha.py
ants: Will correct.

patroni/postgresql.py
CyberDem0n: The problem is that it could be changed with 'ALTER SYSTEM' which takes precedence over postgresql.conf :(

patroni/quorum.py
ants: Yes, that would be a correct optimization. Currently the algorithm doesn't take into account that the master is special and always guaranteed to be one of the synced nodes. This may lead to extra steps when adding multiple nodes at the same time. On the plus side this allows for promotion before we update sync state in DCS. I need to think a bit about how the algorithm would look like if we consider that master to be special. If it makes it considerably more complicated than it already is, then it might not be worth the minor improvement in efficiency. I will also write down how and why the QuorumStateResolver works.

patroni/postgresql.py
ants: We could detect that from pg_settings. Now the question is what should the system do in that situation? Override the administrator with `ALTER SYSTEM RESET synchronous_standby_names`, make a quorum state to match the synchronization state, just complain in the logs where nobody is looking anyway, panic with `sys.exit(1)`?

CyberDem0n: Looks like executing `ALTER SYSTEM RESET synchronous_standby_names` is the best we can do.
In addition to that we can write warning into the log with previous value.

patroni/ha.py
```diff
@@ -293,54 +294,103 @@ def is_synchronous_mode(self):
...
+
+            for transition, num, nodes in QuorumStateResolver(quorum=quorum,
+                                                              voters=voters,
+                                                              numsync=sync_state['numsync'],
+                                                              sync=sync_state['sync'],
+                                                              active=sync_state['active'],
+                                                              sync_wanted=sync_wanted):
+                if transition == 'quorum':
+                    logger.info("Setting quorum to %d of %d (%s)", num, len(nodes), ", ".join(sorted(nodes)))
+                    if not self.dcs.write_sync_state(leader=self.state_handler.name, quorum=num, members=list(nodes),
```
anikin-aa: `write_sync_state` - this method should be also changed, no ?
here you are passing `leader`, `quorum`, `membres` and `index` as a parameters, but signature of method wasn't changed in this pull request.

```python
def write_sync_state(self, leader, sync_standby, index=None):
```

anikin-aa:
@ants If there is any chance You are getting back to work on this feature ? 
I am pretty interested in this one, if You are need any help, feel free to contact me.

patroni/postgresql.py
```diff
@@ -1645,52 +1717,77 @@ def basebackup(self, conn_url, env):
...
-        :returns tuple of candidate name or None, and bool showing if the member is the active synchronous standby.
+    def current_sync_state(self, cluster):
+        """Returns current synchronous replication state as a dict:
+        * numsync: number of synchronous nodes requested(including leader)
+        * sync: set of nodes potentially being synced to
+        * active: set of nodes that are sync capable
         """
-        current = cluster.sync.sync_standby
-        current = current.lower() if current else current
+        sync_standby_names = self.query("SHOW synchronous_standby_names")
```
anikin-aa: as I can see here https://github.com/zalando/patroni/blob/master/patroni/postgresql.py#L1509
`self.query` - will return `cursor`, not the result of query execution. And after that You are passing cursor to 
`sync_rep_parser_re.finditer` its not acceptable.

ants:
Thanks for reviewing. I plan to resume work on this in about 1-2 weeks time
and any review effort on the code will be helpful, as will be testing once
the code is in a testable state.

>

anikin-aa:
Guys, any updates on this PR ?

patroni/quorum.py
```diff
@@ -0,0 +1,186 @@
...
+        if self.quorum and not (len(self.voters | self.sync) < self.quorum + self.numsync):
+            raise QuorumError("Quorum and sync not guaranteed to overlap: nodes %d >= quorum %d + sync %d" %
+                              (len(self.voters | self.sync), self.quorum, self.numsync))
+        if not (self.voters <= self.sync or self.sync <= self.voters):
+            raise QuorumError("Mismatched sets: quorum only=%s sync only=%s" %
+                              (self.voters - self.sync, self.sync - self.voters))
+
+    def quorum_update(self, quorum, voters):
+        if quorum < 1:
+            raise QuorumError("Quorum %d < 0 of (%s)" % (quorum, voters))
```
CyberDem0n: Exception message doesn't match the condition: `quorum < 1` and `%d < 0`

patroni/quorum.py
```diff
@@ -0,0 +1,186 @@
...
+        self.quorum = quorum
+        self.voters = set(voters)
+        self.numsync = numsync
+        self.sync = set(sync)
+        self.active = active
+        self.sync_wanted = sync_wanted
+
+    def check_invariants(self):
+        if self.quorum and not (len(self.voters | self.sync) < self.quorum + self.numsync):
+            raise QuorumError("Quorum and sync not guaranteed to overlap: nodes %d >= quorum %d + sync %d" %
```
CyberDem0n: When I removed sync key from the DCS the master crashed:
```
$ etcdctl rm /service/batman/sync
PrevNode.Value: {"leader":"postgresql2","members":["postgresql1","postgresql0","postgresql2"],"quorum":2}

2019-02-10 09:51:33,719 INFO: Lock owner: postgresql2; I am postgresql2
2019-02-10 09:51:33.720 CET [10143] LOG:  received fast shutdown request
2019-02-10 09:51:33.725 CET [10143] LOG:  aborting any active transactions
2019-02-10 09:51:33.725 CET [10178] FATAL:  terminating connection due to administrator command
2019-02-10 09:51:33.730 CET [10143] LOG:  background worker "logical replication launcher" (PID 10193) exited with exit code 1
2019-02-10 09:51:33.731 CET [10146] LOG:  shutting down
2019-02-10 09:51:33.794 CET [10143] LOG:  database system is shut down
2019-02-10 09:51:33,813 INFO: Lock owner: postgresql2; I am postgresql2
  File "./patroni.py", line 6, in <module>
    main()
  File "/home/akukushkin/git/ants/patroni/patroni/__init__.py", line 182, in main
    return patroni_main()
  File "/home/akukushkin/git/ants/patroni/patroni/__init__.py", line 148, in patroni_main
    patroni.run()
  File "/home/akukushkin/git/ants/patroni/patroni/__init__.py", line 119, in run
    logger.info(self.ha.run_cycle())
  File "/home/akukushkin/git/ants/patroni/patroni/ha.py", line 1370, in run_cycle
    info = self._run_cycle()
  File "/home/akukushkin/git/ants/patroni/patroni/ha.py", line 1343, in _run_cycle
    msg = self.process_healthy_cluster()
  File "/home/akukushkin/git/ants/patroni/patroni/ha.py", line 966, in process_healthy_cluster
    'promoted self to leader because i had the session lock'
  File "/home/akukushkin/git/ants/patroni/patroni/ha.py", line 566, in enforce_master_role
    self.process_sync_replication()
  File "/home/akukushkin/git/ants/patroni/patroni/ha.py", line 426, in process_sync_replication
    sync_wanted=sync_wanted):
  File "/home/akukushkin/git/ants/patroni/patroni/quorum.py", line 84, in __iter__
    transitions = list(self._generate_transitions())
  File "/home/akukushkin/git/ants/patroni/patroni/quorum.py", line 95, in _generate_transitions
    self.check_invariants()
  File "/home/akukushkin/git/ants/patroni/patroni/quorum.py", line 62, in check_invariants
    (len(self.voters | self.sync), self.quorum, self.numsync))
patroni.quorum.QuorumError: Quorum and sync not guaranteed to overlap: nodes 3 >= quorum 1 + sync 2
```

patroni/quorum.py
```diff
@@ -0,0 +1,186 @@
...
+                yield self.quorum_update(self.quorum, self.voters - to_remove)
+                yield self.sync_update(self.numsync - len(to_remove), self.sync - to_remove)
+
+        # If any new nodes, join them to quorum
+        to_add = self.active - self.sync
+        if to_add:
+            # First get to requested replication factor
+            logger.debug("Adding nodes: %s", to_add)
+            increase_numsync_by = self.sync_wanted - self.numsync
+            if increase_numsync_by:
```
CyberDem0n: `increase_numsync_by` could become negative.
example:
```python
QuorumStateResolver(quorum=2, voters=set('abcdef'),
                    numsync=5, sync=set('abcdef'),
                    active=set('abcdefg'), sync_wanted=4)
```

patroni/quorum.py
```diff
@@ -0,0 +1,186 @@
...
+                yield self.sync_update(
+                        numsync=min(self.sync_wanted, len(self.sync) - len(remove_from_sync)),
+                        sync=self.sync - remove_from_sync)
+
+        # After handling these two cases quorum and sync must match.
+        assert self.voters == self.sync
+
+        safety_margin = self.quorum + self.numsync - len(self.voters | self.sync)
+        if safety_margin > 1:
+            logger.debug("Case 3: replication factor is bigger than needed")
```
CyberDem0n: This branch is not covered by tests, so I tried to do it:
```python
QuorumStateResolver(quorum=3, voters=set('abcdef'),
                    numsync=5, sync=set('abcdef'),
                    active=set('abcdef'), sync_wanted=3)
```
and got an exception `QuorumError: Quorum and sync not guaranteed to overlap: nodes 6 >= quorum 3 + sync 3` from `sync_update()`

Jan-M:
This PR suddenly is more interesting. Can this Pullrequest be made easier for non support of PG9.6? I expected this to be less complicated given Postgres supports Quorum mode now.

Jan-M:
https://www.postgresql.org/docs/10/runtime-config-replication.html#RUNTIME-CONFIG-REPLICATION-MASTER

tests/test_quorum.py
```diff
@@ -0,0 +1,122 @@
...
+            ('sync', 3, set('abcde')),
+        ])
+
+    def test_1222(self):
+        """2 node cluster"""
+        state = 1, set("ab"), 2, set("ab")
+        # Active set matches state
+        self.assertEqual(list(QuorumStateResolver(*state, active=set("ab"), sync_wanted=3)), [
+        ])
+        # Add node by increasing quorum
```
sdudoladov: I find this description a bit too short , better would be to have sth like
> add a sync standby without increasing the replication factor; must increase quorum to avoid losing commits on failover

tests/test_quorum.py
```diff
@@ -0,0 +1,122 @@
...
+        self.assertEqual(list(QuorumStateResolver(*state, active=set("abc"), sync_wanted=2)), [
+            ('quorum', 2, set('abc')),
+            ('sync', 2, set('abc')),
+        ])
+        # Add node by increasing sync
```
sdudoladov: same here:
> add a sync standby and increase the replication factor in the corner case where before and after adding a sync standby all nodes of a PG cluster must ack the commit

tests/test_quorum.py
```diff
@@ -0,0 +1,122 @@
...
+            ('sync', 3, set('abc')),
+            ('quorum', 1, set('abc')),
+        ])
+        # Add multiple nodes by increasing both sync and quorum
+        self.assertEqual(list(QuorumStateResolver(*state, active=set("abcde"), sync_wanted=3)), [
+            ('sync', 3, set('abc')),
+            ('quorum', 3, set('abcde')),
+            ('sync', 3, set('abcde')),
+        ])
+        # Master is alone
```
sdudoladov: Is that  removing a sync standby from a 2-node cluster ?

patroni/quorum.py
```diff
@@ -0,0 +1,187 @@
...
+        logger.debug("Quorum state: quorum %s, voters %s, numsync %s, sync %s, active %s, sync_wanted %s",
+                     self.quorum, self.voters, self.numsync, self.sync, self.active, self.sync_wanted)
+        self.check_invariants()
+
+        # Handle non steady state cases
+        if self.sync < self.voters:
+            logger.debug("Case 1: synchronous_standby_names subset of DCS state")
+            # Case 1: quorum is superset of sync nodes. In the middle of changing quorum.
+            # Evict from quorum dead nodes that are not being synced.
+            remove_from_quorum = self.voters - (self.sync | self.active)
```
sdudoladov: Can it happen at this point that `self.sync` still contains some node that ceased to be active and  therefore must be evicted from `voters` ? In other words, why do we subtract here the union and not `self.sync & self.active` or simply `self.active` ?

patroni/postgresql/__init__.py
```diff
@@ -804,38 +812,32 @@ def remove_data_directory(self):
...
-        """
-        current = cluster.sync.sync_standby
-        current = current.lower() if current else current
+        active = [self.name]
         members = {m.name.lower(): m for m in cluster.members}
-        candidates = []
-        # Pick candidates based on who has flushed WAL farthest.
-        # TODO: for synchronous_commit = remote_write we actually want to order on write_location
+
         for app_name, state, sync_state in self.query(
```
sdudoladov: `sync_state` is no longer used in the proposed change

patroni/postgresql/__init__.py
```diff
@@ -804,38 +812,32 @@ def remove_data_directory(self):
...
-            if sync_state == 'potential' and app_name == current:
-                # Prefer current even if not the best one any more to avoid indecisivness and spurious swaps.
-                return cluster.sync.sync_standby, False
-            if sync_state in ('async', 'potential'):
-                candidates.append(member.name)
-
-        if candidates:
-            return candidates[0], False
-        return None, False
+            active.append(member.name)
```
sdudoladov: Just to double check: a standby with `pg_stat_replication.sync_state  in ('async', 'potential'):`  is still considered `active` for the purposes of `QuorumStateResolver`?

patroni/quorum.py
```diff
@@ -0,0 +1,187 @@
...
+            if add_to_quorum:
+                yield self.quorum_update(
+                        quorum=self.quorum,
+                        voters=self.voters | add_to_quorum)
+            # Remove from sync nodes that are dead
+            remove_from_sync = self.sync - self.voters
+            if remove_from_sync:
+                yield self.sync_update(
+                        numsync=min(self.sync_wanted, len(self.sync) - len(remove_from_sync)),
+                        sync=self.sync - remove_from_sync)
```
sdudoladov: why not

```python
if remove_from_sync:
      remaining_sync = self.sync - remove_from_sync
      yield self.sync_update(
               numsync=min(self.sync_wanted, len(remaining_sync)),
               sync=remaining_sync)
```

patroni/quorum.py
```diff
@@ -0,0 +1,187 @@
...
+        # After handling these two cases quorum and sync must match.
+        assert self.voters == self.sync
+
+        safety_margin = self.quorum + self.numsync - len(self.voters | self.sync)
+        if safety_margin > 1:
+            logger.debug("Case 3: replication factor is bigger than needed")
+            # Case 3: quorum or replication factor is bigger than needed. In the middle of changing replication factor.
+            if self.numsync > self.sync_wanted:
+                # Reduce replication factor
+                new_sync = clamp(self.sync_wanted, min=len(self.voters) - self.quorum + 1, max=len(self.sync))
```
sdudoladov: `sync` denotes a set everywhere else, so here just for the sake of consistency, `new_sync` should be `new_numsync`

docs/replication_modes.rst
```diff
@@ -40,38 +40,31 @@ Synchronous mode
...
-
-- A node must be marked as the latest leader whenever it can accept write transactions. Patroni crashing or PostgreSQL not shutting down can cause violations of this invariant.
-
-- A node must be set as the synchronous standby in PostgreSQL as long as it is published as the synchronous standby.
-
-- A node that is not the leader or current synchronous standby is not allowed to promote itself automatically.
-
-Patroni will only ever assign one standby to ``synchronous_standby_names`` because with multiple candidates it is not possible to know which node was acting as synchronous during the failure.
+When in synchronous mode Patroni maintains synchronization state in the DCS, containing the latest primary, number of nodes required for quorum and nodes currently eligible to vote on quorum. In steady state the nodes voting on quorum are the leader and all synchronous standbys. This state is updated with strict ordering constraints with regards to node promotion and ``synchronous_standby_names`` to ensure that at all times any subset of voters that can achieve quorum is contained to have at least one node having the latest successful commit.
```
sdudoladov: > In steady state the nodes voting on quorum are the leader and all synchronous standbys.

Judging from [this line](https://github.com/zalando/patroni/pull/672/files#diff-a839e95c7271edf3b62934f9e0e61f16R131), steady state also means updates to quorum and/or replication factor have completed (as opposed to simply having `self.voters == self.sync`).

ksarabu1:
We are looking into implementing Patroni in-house and this feature is one of the shortfall holding us right now. We are very much interested in this enhancement and wondering if we can be any help in getting this PR moving. We are open to collaborate and contribute. Thanks.

ksarabu1:
We utilize mix of synchronous hot standbys (with synchronous_commit set to remote_apply
for read-consistency between master/sync_standby) & async standby
databases where the applications tolerate staleness.

With the Quorum based (ANY) implementation, the databases that receives the change 
synchronously from the list may be random (unless we set sync_num matching to # of replicas).

My suggestion would be to extend current one synchronous standby implementation to 
support precise multiple guaranteed sync standbys based on maximum/minimum replication factor. Synchronous standby members list can be updated by Patroni as the members join/leave to maintain maximum/minimum limit of no of sync standby's.

This functionality can be further extended to make it generic using new parameter(s) for providing 
additional support for Priority (FIRST n) based synchronous replication & Quorum (ANY n) based synchronous replication.

Please let me know your thoughts.

Thanks

Issue #657, opened 2018-04-05
Allow a HA cluster to run in a standby mode 


alexeyklyukin:
Right now, Patroni always promotes one node in a cluster that acquires the leader lock.

We may want to have a standby cluster, running alongside the primary one (i.e. in another datacenter) and consisting of only standby nodes, replicating from the primary one. Such cluster will have no master, there will be one "main" replica replicating from the master in the primary cluster, and all the other cascading replicas replicating from the main one.

We can already achieve this by writing a permanent leader and member keys in DCS, specifying the address of the cluster in the member key. However, the primary leader knows nothing about the standby cluster, it doesn't create a replication slot, and the whole setup requires direct access to DCS from the user implementing it.

What we can do is to create a new key in DCS called standby, holding connection string for the actual master inside. When this key is there, the leader key will be interpreted as a pointer to the main replica, instead of the master, that replica will not promote itself, and the other replicas will replicate from the main one. When the standby key disappears, the cluster stops being a standby, and the replica holding the leader key will promote itself.

From the user point of view, there is not much additional information to be supplied to Patroni. We only need the credentials for the master (standby and superuser names and passwords, they are already required in the current configuration file and obviously should match between the primary and the standby cluster) and the connection string.

jberkus:
Wouldn't it be better to look at this as "support for cascading replication" instead?

Or are you specifically treating the "remote standbys" as a separate cluster in order to deal with the limitations of kubernetes federation?

alexeyklyukin:
@jberkus we already have cascading replication based on per-instance tags. While it is limited to a single fixed node for each cascading destination I don't think it would be terribly useful to make a single origin replica within a single cluster with a primary. 

We didn't have the federation use case in mind, although it may prove useful for that (we don't use Kubernetes federation). We did, however, run an all-replica HA cluster  on AWS at some point in time for one of our biggest databases, replicated from the HA cluster in the DC, as a part of migrating our databases to the cloud.

jberkus:
So a couple things I'm not understanding from your description:

1. How does election of the "relay node" happen?  To make sure that there's always one node replicating from the original DC master?
2. How does the system decide that the original DC is gone?  Would it be possible to automate this via a consensus among three data centers, or are you doing DC failover manually?

CyberDem0n:
> How does election of the "relay node" happen? To make sure that there's always one node replicating from the original DC master?

[...]

PR #679, opened 2018-05-14
[WIP] Standby cluster implementation

erthalion:
First attempt to implement #657 . Created mostly for discussion and transparency, only the main path is implemented without handling of any corner cases and tests so far. To enable you need to bootstrap a new cluster with:

```yaml
standby_cluster:
  conn_url: postgresql://127.0.0.1:1111/postgres
  replication_slot: patroni
```


patroni/ha.py
```diff
@@ -187,6 +195,28 @@ def bootstrap(self):
                 return 'trying to ' + msg
             return 'waiting for leader to bootstrap'
 
+    def bootstrap_standby_leader(self):
+        """ If we found 'standby' key in the configuration, we need to bootstrap
+            not a real master, but a 'standby leader', that will take base backup
+            from a remote master and start follow it.
+        """
+        config_value = json.dumps(
```
CyberDem0n: There should be nither `/config` nor `/leader` key in DCS until bootstrap is done.

patroni/ha.py
```diff
@@ -725,8 +761,23 @@ def process_unhealthy_cluster(self):
...
                         self.dcs.manual_failover('', '')
                 self.load_cluster_from_dcs()
-                return self.enforce_master_role('acquired session lock as a leader',
-                                                'promoted self to leader by acquiring session lock')
+
+                if self.cluster.is_standby_cluster():
+                    # standby leader disappeared, and this is a healthiest
+                    # replica, so it should become a new standby leader.
+                    # This imply that we need to start following a remote master
+                    msg = 'follow remote master {0}'.format(follow_target.conn_url)
```
CyberDem0n: follow_target is not defined here

patroni/ha.py
```diff
@@ -725,8 +761,23 @@ def process_unhealthy_cluster(self):
...
+                    clone_target = self.cluster.get_target_to_follow()
```
CyberDem0n: clone_target is unused

patroni/ha.py
```diff
@@ -725,8 +761,23 @@ def process_unhealthy_cluster(self):
...
+                    follow_target = self.cluster.get_target_to_follow()
+
+                    self._async_executor.schedule('follow_remote_master')
+                    self._async_executor.run_async(self.follow, args=(follow_target, msg))
+                    return 'promoted self to a standby leader because i had the session lock'
```
CyberDem0n: `promoted self to a standby leader by acquiring session lock`

patroni/ha.py
```diff
@@ -758,8 +809,17 @@ def process_healthy_cluster(self):
                 if msg is not None:
                     return msg
 
-                return self.enforce_master_role('no action.  i am the leader with the lock',
-                                                'promoted self to leader because i had the session lock')
+                if self.cluster.is_standby_cluster():
+                    # in case of standby cluster we don't really need to
```
CyberDem0n: What if it was accidentally promoted?
You have to call `self.follow` here.

patroni/postgresql.py
```diff
@@ -1365,7 +1371,9 @@ def follow(self, member, timeout=None):
         if primary_conninfo:
             recovery_params['primary_conninfo'] = primary_conninfo
         if self.use_slots:
-            recovery_params['primary_slot_name'] = slot_name_from_member_name(self.name)
+            required_name = member is not None and member.data.get('replication_slot')
```
CyberDem0n: In case if we are following the remote cluster and there is no `replication_slot` defined in the configuration `primary_slot_name` shouldn't be defined either.

patroni/ha.py
```diff
@@ -1031,6 +1092,9 @@ def set_start_timeout(self, value):
         self._start_timeout = value
 
     def _run_cycle(self):
+        if self._stop is True:
```
CyberDem0n: Doesn't seems to be used.

patroni/ha.py
```diff
@@ -725,8 +761,23 @@ def process_unhealthy_cluster(self):
...
+                if self.cluster.is_standby_cluster():
+                    # standby leader disappeared, and this is a healthiest
+                    # replica, so it should become a new standby leader.
+                    # This imply that we need to start following a remote master
+                    msg = 'follow remote master {0}'.format(follow_target.conn_url)
+                    clone_target = self.cluster.get_target_to_follow()
+                    follow_target = self.cluster.get_target_to_follow()
+
+                    self._async_executor.schedule('follow_remote_master')
+                    self._async_executor.run_async(self.follow, args=(follow_target, msg))
```
CyberDem0n: `follow` itself checks recovery.conf and asynchronously calls `state_handler.follow` when it is necessary, therefore no need to use `_async_executor` here

patroni/config.py
```diff
@@ -312,8 +317,16 @@ def _build_effective_configuration(self, dynamic_configuration, local_configurat
         if 'name' not in config and 'name' in pg_config:
             config['name'] = pg_config['name']
 
-        pg_config.update({p: config[p] for p in ('name', 'scope', 'retry_timeout',
-                          'synchronous_mode', 'maximum_lag_on_failover') if p in config})
+        updated_fields = (
+            'name',
```
CyberDem0n: We actually need only first 3 here...

patroni/dcs/__init__.py
```diff
@@ -426,6 +440,10 @@ def leader_optime_path(self):
     def sync_path(self):
         return self.client_path(self._SYNC)
 
+    @property
+    def standby_path(self):
```
CyberDem0n: There is no standby key in DCS.

patroni/config.py
```diff
@@ -88,6 +89,10 @@ def config_file(self):
     def dynamic_configuration(self):
         return deepcopy(self._dynamic_configuration)
 
+    @property
+    def is_standby_cluster(self):
```
CyberDem0n: There is no need to have `is_standby_cluster` property here.

The only reason why we keep information about `standby_cluster` in the `Config()` is just to be able to dump into `patroni.dynamic.json`. It makes it possible to restore it if somebody accidentally wipes DCS out.

patroni/ha.py
erthalion: This branch is within `if self.acquire_lock()`, isn't it enough?

patroni/config.py
erthalion: There is such need - because we need to get this info directly from config file before bootstrap. It's not really good to read it directly from `_dynamic_configuration`, so I've created a property.

patroni/ha.py
```diff
@@ -252,12 +278,18 @@ def recover(self):
     def _get_node_to_follow(self, cluster):
         # determine the node to follow. If replicatefrom tag is set,
         # try to follow the node mentioned there, otherwise, follow the leader.
-        if not self.patroni.replicatefrom or self.patroni.replicatefrom == self.state_handler.name:
-            node_to_follow = cluster.leader
-        else:
+        is_leader = self.cluster.leader and self.state_handler.name == self.cluster.leader.name
+
+        if self.patroni.replicatefrom and self.patroni.replicatefrom != self.state_handler.name:
```
CyberDem0n: What if node is the leader of standby cluster and replicatefrom tag is defined?
We will get a replication loop :)

patroni/ha.py
```diff
@@ -397,6 +429,15 @@ def update_cluster_history(self):
                         line.append(cluster_history[line[0]][3])
                 self.dcs.set_history_value(json.dumps(history, separators=(',', ':')))
 
+    def enforce_follow_remote_master(self):
+        self.state_handler.set_role('standby_leader')
+        follow_target = self.cluster.get_target_to_follow()
+        message = 'follow remote master {0}'.format(follow_target.conn_url)
+
+        self._async_executor.schedule('follow_remote_master')
+        self._async_executor.run_async(self.follow, args=(follow_target, message))
```
CyberDem0n: It seems that you wanted to call `Postgresql.follow`, but it also has a different signature.

I absolutely agree that we shouldn't touch existing `Ha.follow`, but at the same time `enforce_follow_remote_master` should implement most of the functionality of the `Ha.follow`.

Also `Ha._handle_rewind` should do a rewind from remote master, not from the cluster leader.

patroni/ha.py
```diff
@@ -758,8 +809,17 @@ def process_healthy_cluster(self):
...
                     return msg
 
-                return self.enforce_master_role('no action.  i am the leader with the lock',
-                                                'promoted self to leader because i had the session lock')
+                if self.cluster.is_standby_cluster():
+                    # in case of standby cluster we don't really need to
+                    # enforce anything, since the leader is not a master.
+                    # So just remind the role.
+                    self.enforce_follow_remote_master()
+                    return 'no action. I am the standby leader with the lock'
```
CyberDem0n: It might happen that postgres will be restarted due to change in `recovery.conf`, but it will still report "no action".

patroni/postgresql.py
```diff
@@ -1365,7 +1371,10 @@ def follow(self, member, timeout=None):
...
+            name = required_name or slot_name_from_member_name(self.name)
```
CyberDem0n: Unfortunately it doesn't fix a problem.
Lets assume that we want to follow the remote cluster without using replication slots.
So we have no `replication_slot defined` in the `standby_cluster`, but according to this line it will set `primary_slot_name` to the `slot_name_from_member_name(self.name)`.

Actually the same function should be applied to the `required_name`

patroni/config.py
```diff
@@ -45,6 +45,7 @@ class Config(object):
         'master_start_timeout': 300,
         'synchronous_mode': False,
         'synchronous_mode_strict': False,
+        'standby_cluster': '',
```
CyberDem0n: I've though a little bit about format of `standby_cluster` configuration and would suggest to follow postgres naming conventions here: https://www.postgresql.org/docs/10/static/recovery-config.html
```python
{
    'host': 'localhost',
    'port': 12345,
    'primary_slot_name': 'foo-bar',
    'restore_command': 'true',
    'archive_cleanup_command': ''
    'recovery_min_apply_delay': '10min'
}
```

We can assume that this is a standby cluster when either `host` or `restore_command` are defined.

patroni/postgresql.py
erthalion: Hm...looks like it means that this function should know whether this particular member is a part of standby cluster or not, am I right?

patroni/config.py
```diff
@@ -45,6 +46,14 @@ class Config(object):
         'master_start_timeout': 300,
         'synchronous_mode': False,
         'synchronous_mode_strict': False,
+        'standby_cluster': {
+            'host': 'localhost',
```
erthalion: Empty strings for values

patroni/ha.py
```diff
@@ -1201,3 +1265,21 @@ def wakeup(self):
...
+        config = config or (self.config is not None and self.config.data)
+
+        if config and config.get('standby_cluster'):
+            cluster_params = config.get('standby_cluster')
+            return Member(None, 'remote_master', None, {
+                'conn_kwargs': {
+                    "host": cluster_params.get('host'),
+                    "port": cluster_params.get('port'),
+                },
+                'primary_slot_name': cluster_params['primary_slot_name'],
```
CyberDem0n: What if `primary_slot_name` is not specified?

patroni/ha.py
```diff
@@ -1201,3 +1265,21 @@ def wakeup(self):
...
+                'no_replication_slot': False,
```
CyberDem0n: Must be `True` if there is no `primary_slot_name` defined.

patroni/ha.py
```diff
@@ -1201,3 +1265,21 @@ def wakeup(self):
...
+    def get_target_to_follow(self, config):
+        """ In case of standby cluster this will tel us from which remote
+            master to stream. Config can be both patroni config or
+            cluster.config.data
+        """
+        config = config or (self.config is not None and self.config.data)
+
+        if config and config.get('standby_cluster'):
+            cluster_params = config.get('standby_cluster')
+            return Member(None, 'remote_master', None, {
```
CyberDem0n: What is one of the nodes in the cluster has the name `remote_master`?

features/steps/standby_cluster.py
```diff
@@ -0,0 +1,25 @@
...
+def start_patroni_stanby_cluster(context, name, cluster_name, name2):
+    port = context.pctl._processes[name2]._connkwargs.get('port')
+    return context.pctl.start(name, custom_config={
+        "scope": cluster_name,
+        "bootstrap": {
+            "dcs": {
+                "standby_cluster" :{
+                    "host": "localhost",
+                    "port": port,
+                    "primary_slot_name": "postgresql1",
```
CyberDem0n: Since there is no replication slot on the `postgresql0` it won't be able to connect and stream...

features/standby_cluster.feature
```diff
@@ -0,0 +1,16 @@
...
+    Then postgres1 is a leader of batman1 after 10 seconds
+    And postgres1 role is the standby_leader after 24 seconds
+    And table foo is present on postgres1 after 20 seconds
+    When I start postgres2 in a cluster batman1
+    Then postgres2 role is the replica after 24 seconds
+    And table foo is present on postgres2 after 20 seconds
+
+  Scenario: check failover
+    When I kill postmaster on postgres1
+    Then postgres2 role is the standby_leader after 24 seconds
```
CyberDem0n: At the end of the day it will call: https://github.com/zalando/patroni/blob/master/features/environment.py#L219, which executes `pg_is_in_recovery()` and treats roles as `'primary'` and `not 'primary'`

patroni/common_config.py
```diff
@@ -0,0 +1,9 @@
+def is_standby_cluster(config):
+    """ Check whether or not provided configuration describes a standby cluster.
+        Config can be both patroni config or cluster.config.data
+    """
+    return config is not None and (
```
CyberDem0n:
```python
isinstance(config, dict) and (config.get('host') or config.get('port') or config.get('restore_command'))
```

Because:
```python
In [1]: '' is not None
Out[1]: True
```

patroni/ha.py
```diff
@@ -1201,3 +1266,25 @@ def wakeup(self):
...
+        if config and config.get('standby_cluster'):
+            cluster_params = config.get('standby_cluster')
+            name = self.unique_name_for_remote_master()
+            return Member(None, name, None, {
+                'conn_kwargs': {
+                    "host": cluster_params.get('host'),
+                    "port": cluster_params.get('port'),
+                },
+                'primary_slot_name': cluster_params.get('primary_slot_name'),
+                'no_replication_slot': not hasattr(cluster_params, 'primary_slot_name'),
```
CyberDem0n: `hasattr` will always return False in this case

patroni/ha.py
```diff
@@ -733,8 +775,18 @@ def process_unhealthy_cluster(self):
...
                         self.dcs.manual_failover('', '')
                 self.load_cluster_from_dcs()
-                return self.enforce_master_role('acquired session lock as a leader',
-                                                'promoted self to leader by acquiring session lock')
+
+                if self.cluster.is_standby_cluster():
+                    # standby leader disappeared, and this is a healthiest
+                    # replica, so it should become a new standby leader.
+                    # This imply that we need to start following a remote master
+                    self.enforce_follow_remote_master()
```
alexeyklyukin: I'd pass the message at line 784 to the enforce_follow_remote_master and return its result, which should be either  a message that the actual Postgres master demoted, or that we have promoted ourself to the standby leader and following the remote leader. Right now, the message from line 448 is simply discarded here.

patroni/ha.py
```diff
@@ -766,8 +818,17 @@ def process_healthy_cluster(self):
                 if msg is not None:
                     return msg
 
-                return self.enforce_master_role('no action.  i am the leader with the lock',
-                                                'promoted self to leader because i had the session lock')
+                if self.cluster.is_standby_cluster():
+                    # in case of standby cluster we don't really need to
+                    # enforce anything, since the leader is not a master.
+                    # So just remind the role.
+                    self.enforce_follow_remote_master()
```
alexeyklyukin: same consideration: make enforce_follow_remote_master return you the lock you need.

patroni/ha.py
```diff
@@ -1201,3 +1266,25 @@ def wakeup(self):
         no "active" leader watch request in progress.
         This usually happens on the master or if the node is running async action"""
         self.dcs.event.set()
+
+    def unique_name_for_remote_master(self):
+        return 'remote_master:{}'.format(uuid.uuid1())
+
+    def get_target_to_follow(self, config):
```
alexeyklyukin: Too generic name. It seems to be used only in the standby_cluster scenario and only for figuring out the node that is the remote master. So why not name it get_remote_master?

alexeyklyukin: Besides, it is used not only for finding node to follow, but also for finding the one to clone from.

patroni/ha.py
```diff
@@ -52,6 +54,7 @@ def failover_limitation(self):
 class Ha(object):
 
     def __init__(self, patroni):
+        self._stop = False
```
CyberDem0n: I would still insist on removing it

features/standby_cluster.feature
```diff
@@ -3,10 +3,9 @@ Feature: standby cluster
   Scenario: check replication of a single table in a standby cluster
     Given I start postgres0 without slots sync
     And I create a replication slot postgres1 on postgres0
-    And I create a role replicator on postgres0
     And I start postgres1 in a standby cluster batman1 as a clone of postgres0
     Then postgres1 is a leader of batman1 after 10 seconds
-    And postgres1 role is the standby_leader after 24 seconds
+    And DCS for batman1 has leader=postgres1 after 20 seconds
```
CyberDem0n: https://github.com/zalando/patroni/blob/master/features/custom_bootstrap.feature#L9 -- is already doing exactly what you need

docs/SETTINGS.rst
```diff
@@ -25,6 +25,14 @@ Bootstrap configuration
         -  **use\_slots**: whether or not to use replication_slots. Must be False for PostgreSQL 9.3. You should comment out max_replication_slots before it becomes ineligible for leader status.
         -  **recovery\_conf**: additional configuration settings written to recovery.conf when configuring follower. 
         -  **parameters**: list of configuration settings for Postgres. Many of these are required for replication to work.
+    -  **standby\_cluster**: if this section is defined, we want to bootstrap a standby cluster.
+        -  **host**: an address of remote master
+        -  **port**: a port of remote master
+        -  **primary\_slot\_name**: which slot on the remote master to use for replication
```
CyberDem0n: Would be good to mention that this is an optional parameter.

docs/SETTINGS.rst
```diff
@@ -25,6 +25,14 @@ Bootstrap configuration
...
+        -  **create\_replica\_methods**: an ordered list of methods that can be used to bootstrap standby leader from the remote master, can be different from the list defined in :ref:`postgresql_settings`
+        -  **restore\_command**: command to restore WAL records from the remote master to standby leader, can be different from the list defined in :ref:`postgresql_settings`
+        -  **archive\_cleanup\_command**: cleanup command for standby leader
```
CyberDem0n: `archive_cleanup_command` and `recovery_min_apply_delay` are not yet supported:
* https://github.com/zalando/patroni/pull/679/files#diff-57c53e988b775ee2902488fb33c1f433R1295
* https://github.com/zalando/patroni/pull/679/files#diff-d8c1582337c8426ef750420dc6b76949R1437

docs/replica_bootstrap.rst
```diff
@@ -129,3 +129,36 @@ and
...
+
+Another available option is to run a "standby cluster", that contains only of
+standby nodes replicating from some remote master. This type of clusters has:
+
+* "standby leader", that behaves pretty much like a regular cluster leader,
+  except it replicates from a remote master.
+
+* cascade replicas, that are replicating from standby leader.
+
+Standby leader holds and updates a leader lock in DCS, and if it will
```
bradnicholson: Would read cleaner as:

```
The standby leader holds and updates a leader lock in DCS. If the leader lock expires, cascade replicas will perform an election to choose another leader from the standbys.
```

erthalion: Yes, sounds good, thanks.

docs/replica_bootstrap.rst
```diff
@@ -129,3 +129,36 @@ and
...
+"standby mode", and after it was detached to function as a normal cluster.
+
+To configure such cluster you need to specify the section ``standby_cluster``
+in a patroni configuration:
+
+.. code:: YAML
+
+    bootstrap:
+        dcs:
+            postgresql:
```
CyberDem0n: `standby_cluster` is not inside `postgresql:`

patroni/config.py
```diff
@@ -183,6 +197,13 @@ def _safe_copy_dynamic_configuration(self, dynamic_configuration):
                         config['postgresql'][name].update(self._process_postgresql_parameters(value))
                     elif name not in ('connect_address', 'listen', 'data_dir', 'pgpass', 'authentication'):
                         config['postgresql'][name] = deepcopy(value)
+            elif name == 'standby_cluster':
+                allowed_keys = self.__DEFAULT_CONFIG['standby_cluster'].keys()
+                expected = {
+                    k: v for k, v in (value or {}).items()
+                    if (k in allowed_keys and isinstance(v, six.string_types))
```
redbaron: Before trying it, I was looking at how this standby_cluster feature is implemented and just spotted that this one drops `create_replica_methods` which is a list, is it intentional?

PR #680, opened 2018-05-17
Improve protection of DCS being accidentally wiped

CyberDem0n:
We already have a lot of logic in place to prevent failover in such case and restore all keys, but an accidental removal of `/config` key was effectively switching off pause mode for 1 cycle of HA loop.


alexeyklyukin:
👍

CyberDem0n:
:+1:

PR #686, opened 2018-05-30
Add WorkingDirectory to systemd sample config.

bjoernalbers:
Otherwise `initdb` fails because it tries to create the data directory
in the root directory where the postgres user has no permissions.

### DIFF
--- a/extras/startup-scripts/patroni.service
+++ b/extras/startup-scripts/patroni.service
@@ -11,6 +11,8 @@ Type=simple
 User=postgres
 Group=postgres
 
+WorkingDirectory=~
+
 # Where to send early-startup messages from the server
 # This is normally controlled by the global default set by systemd
 # StandardOutput=syslog


CyberDem0n:
Hi @bjoernalbers 

In general, I have no objections about this changes, but actually, you supposed to specify `data_dir` in the Patroni configuration file: https://github.com/zalando/patroni/blob/master/postgres0.yml#L67

It could be relative or absolute path.

bjoernalbers:
Hello Alexander,

thanks for the fast response.
The sample config `postgres0.yml` includes a relative path.
My idea is that it should work out of the box with systemd, which it currently doesn't.
Besides it it less supprising if the relative path starts from postgres's home dir.

Björh

CyberDem0n:
:+1:

alexeyklyukin:
👍

PR #692, opened 2018-06-05
fix: define portname in service

JohnnyQQQQ:
If there is no portname defined, the endpoint won't be attached to the service. The portname in this case is "postgresql"

### DIFF
--- a/kubernetes/patroni_k8s.yaml
+++ b/kubernetes/patroni_k8s.yaml
@@ -108,6 +108,7 @@ spec:
   ports:
   - port: 5432
     targetPort: 5432
+    name: postgresql
 
 ---
 apiVersion: v1


CyberDem0n:
Interesting. What Kubernetes version do you run?
AFAIK, if there is a portname defined in the Service, the same name must be used in the Endpoint [here](https://github.com/zalando/patroni/blob/master/kubernetes/callback.py#L39)

Or use Endpoints instead of ConfigMaps for leader elections and get rid from callbacks, like [here](https://github.com/zalando/patroni/compare/feature/demo), but don't forget to define [PATRONI_KUBERNETES_PORTS](https://github.com/zalando/patroni/blob/master/docs/ENVIRONMENT.rst#kubernetes) in the manifest.

JohnnyQQQQ:
I also use the Spilo image instead of building my own, could this cause the behavoir?

JohnnyQQQQ:
I tried with the config in the branch where you updated the demo to postgres 10, with this config it works with the name on the service port. I close it for now.

PR #698, opened 2018-06-06
If set, use LD_LIBRARY_PATH when starting postgres

### LINKED ISSUE #697: Could not load library ".../libpqwalreceiver.so"
We compile and install Postgres in a non-standard location (/opt/custom/share/postgresql-9.6.2_7 in the example below) which requires that we specify the `LD_LIBRARY_PATH` env var when starting up postgres and when working with postgres utilities. When starting up Postgres, Patroni currently passes along `PATH`, `LC_ALL` and `LANG` env vars if they are set, but does not do the same with `LD_LIBRARY_PATH`. This results in the following FATAL replication error on the replica node:
```
2018-06-05 21:10:26,675 INFO: Lock owner: postgresdb-01; I am postgresdb-02
2018-06-05 21:10:26,675 INFO: does not have lock
2018-06-05 21:10:26,676 INFO: no action.  i am a secondary and i am following a leader
FATAL:  could not load library "/opt/custom/share/postgresql-9.6.2_7/lib/libpqwalreceiver.so": libpq.so.5: cannot open shared object file: No such file or directory
FATAL:  could not load library "/opt/custom/share/postgresql-9.6.2_7/lib/libpqwalreceiver.so": libpq.so.5: cannot open shared object file: No such file or directory
```

The fix for this is to pass `LD_LIBRARY_PATH` along with the other env vars, which results in replication behaving as expected:
```
2018-06-05 22:03:08,722 INFO: Lock owner: postgresdb-01; I am postgresdb-02
2018-06-05 22:03:08,722 INFO: does not have lock
2018-06-05 22:03:08,727 DEBUG: Sending request(xid=49): SetData(path='/service/postgres_cluster/members/postgresdb-02', data='{"conn_url":"postgres://postgresdb-02:5432/postgres","api_url":"http://postgresdb-02:8008/patroni","timeline":1,"state":"running","role":"replica","xlog_location":184664496}', version=-1)
2018-06-05 22:03:08,737 DEBUG: Received response(xid=49): ZnodeStat(czxid=4294967413, mzxid=4294967497, ctime=1528235848173, mtime=1528236188727, version=28, cversion=0, aversion=0, ephemeralOwner=216194533074731009, dataLength=169, numChildren=0, pzxid=4294967413)
2018-06-05 22:03:08,741 INFO: no action.  i am a secondary and i am following a leader
```

I will submit a pull request shortly that addresses this problem.

superflav:
Fixes #697

### DIFF
--- a/patroni/postmaster.py
+++ b/patroni/postmaster.py
@@ -137,7 +137,7 @@ class PostmasterProcess(psutil.Process):
         # of init process to take care about postmaster.
         # In order to make everything portable we can't use fork&exec approach here, so  we will call
         # ourselves and pass list of arguments which must be used to start postgres.
-        env = {p: os.environ[p] for p in ('PATH', 'LC_ALL', 'LANG') if p in os.environ}
+        env = {p: os.environ[p] for p in ('PATH', 'LD_LIBRARY_PATH', 'LC_ALL', 'LANG') if p in os.environ}
         try:
             proc = PostmasterProcess._from_pidfile(data_dir)
             if proc and not proc._is_postmaster_process():


CyberDem0n:
:+1:

ants:
Do we need something more general for this? I assume there is a similar issue with `ORACLE_HOME` for users of oracle_fdw.

patroni/postmaster.py
```diff
@@ -137,7 +137,7 @@ def start(pgcommand, data_dir, conf, options):
         # of init process to take care about postmaster.
         # In order to make everything portable we can't use fork&exec approach here, so  we will call
         # ourselves and pass list of arguments which must be used to start postgres.
-        env = {p: os.environ[p] for p in ('PATH', 'LC_ALL', 'LANG') if p in os.environ}
+        env = {p: os.environ[p] for p in ('PATH', 'LD_LIBRARY_PATH', 'LC_ALL', 'LANG') if p in os.environ}
```
hjacobs: shouldn't the list of env var names now become a constant/global frozenset?

CyberDem0n:
Good idea @ants 
Maybe you also have a list of other useful environment variables we should have in defaults?
From the top of my head, I can only think of about `LD_PRELOAD`

alexeyklyukin:
👍

PR #707, opened 2018-06-12
Do crash recovery only when we sure that postgres was running as master

CyberDem0n:
pg_controldata reports in this case:
* 'in production'
* 'shutting down'
* 'in crash recovery'


alexeyklyukin:
👍

CyberDem0n:
:+1:

PR #711, opened 2018-06-14
Rest api thread can raise an exception during shutdown

CyberDem0n:
catch it and report

### DIFF
--- a/patroni/__init__.py
+++ b/patroni/__init__.py
@@ -132,7 +132,7 @@ class Patroni(object):
         try:
             self.api.shutdown()
         except Exception:
-            logger.exception('Exception durig RestApi.shutdown')
+            logger.exception('Exception during RestApi.shutdown')
         self.ha.shutdown()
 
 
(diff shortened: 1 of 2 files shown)


alexeyklyukin:
👍

patroni/__init__.py
```diff
@@ -129,7 +129,10 @@ def setup_signal_handlers(self):
         signal.signal(signal.SIGTERM, self.sigterm_handler)
 
     def shutdown(self):
-        self.api.shutdown()
+        try:
+            self.api.shutdown()
+        except Exception:
+            logger.exception('Exception durig RestApi.shutdown')
```
alexeyklyukin: during.

alexeyklyukin: also want to see the specifics on the exception

CyberDem0n: what specifics?
logger.exception will show the full stacktrace

alexeyklyukin:
👍

avaczi:
:+1:

CyberDem0n:
:+1:

PR #713, opened 2018-06-15
Respond 200 to master health-check only if update_lock was successful

CyberDem0n:
If Patroni gets partitioned it starts receiving stale information from DCS.
We can't use this information to determine that we have the leader key.
Instead, we will record in Ha object the actual state of acquire/update lock and report as a leader only if it was successful.

P.S. despite responding with 200 on `GET /master` postgres was still running read-only.

### DIFF
--- a/patroni/ha.py
+++ b/patroni/ha.py
@@ -57,6 +57,8 @@ class Ha(object):
         self.dcs = patroni.dcs
         self.cluster = None
         self.old_cluster = None
+        self._is_leader = False
+        self._is_leader_lock = RLock()
         self._was_paused = False
         self._leader_timeline = None
         self.recovering = False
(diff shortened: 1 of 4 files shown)


patroni/ha.py
```diff
@@ -1147,7 +1161,8 @@ def _run_cycle(self):
             return 'Error communicating with PostgreSQL. Will try again later'
         finally:
             if not dcs_failed:
-                self.touch_member()
+                if not self.touch_member() and not self.is_leader():
```
alexeyklyukin: Hm.. why would we blow up a cached reset_cluster at this moment and not at the point when we actually fail to load the new data at load_cluster_from_dcs (if touch_member fails it should trigger the failed flag and the next load cluster failure will be treated as an exception). Also, if we nevertheless do it here, why won't we do it for the leader according to DCS?

CyberDem0n: If etcd node is partitioned it is still available for non-quorum reads, which we are using for performance, therefore `load_cluster_from_dcs` will never fail.
Writes to the partitioned etcd node will always fail, because they require the quorum.

alexeyklyukin: Good point, but which problem are we trying to solve by erasing the cached state for non-leader members? I assume load_cluster_from_dcs will re-read the same (stale) data once again...

CyberDem0n: You are right, this change should be reverted.

alexeyklyukin:
👍

CyberDem0n:
:+1:

alexeyklyukin:
👍

CyberDem0n:
:+1:

Issue #715, opened 2018-06-18
Add patronictl reload function


dtseiler:
`patronictl` doesn't have an option to reload the config from the YAML file. Currently you either have to send a SIGHUP to the patroni process or use the REST API to send `POST /reload`.

We should add a `patronictl` option to call the reload function via the API, similar to the restart function.

I'm working on a fork now for this.

PR #716, opened 2018-06-19
Adding patronictl reload functionality to reload from yaml config file

dtseiler:
PR to address request in Issue #715

### DIFF
--- a/patroni/ctl.py
+++ b/patroni/ctl.py
@@ -463,7 +463,9 @@ def reload(obj, cluster_name, member_names, force, role):
         if r.status_code == 200:
             click.echo('No changes to apply on member {0}'.format(member.name))
         elif r.status_code == 202:
-            click.echo('Success: reload scheduled on member {0}'.format(member.name))
+            click.echo('Reload request received for member {0} and will be processed within {1} seconds'.format(
+                member.name, cluster.config.data.get('loop_wait'))
+            )
         else:
             click.echo('Failed: reload for member {0}, status code={1}, ({2})'.format(
                 member.name, r.status_code, r.text)
--- a/patroni/ctl.py
+++ b/patroni/ctl.py
@@ -445,6 +445,31 @@ def parse_scheduled(scheduled):
     return None
 
 
+@ctl.command('reload', help='Reload cluster member configuration')
+@click.argument('cluster_name')
+@click.argument('member_names', nargs=-1)
+@click.option('--role', '-r', help='Reload only members with this role', default='any',
+              type=click.Choice(['master', 'replica', 'any']))
+@option_force
+@click.pass_obj
+def reload(obj, cluster_name, member_names, force, role):
+    cluster = get_dcs(obj, cluster_name).get_cluster()
+
+    members = get_members(cluster, cluster_name, member_names, role, force, 'reload')
+
+    content = {}
+    for member in members:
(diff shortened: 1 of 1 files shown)


CyberDem0n:
:+1:

patroni/ctl.py
```diff
@@ -445,6 +445,31 @@ def parse_scheduled(scheduled):
...
+def reload(obj, cluster_name, member_names, force, role):
+    cluster = get_dcs(obj, cluster_name).get_cluster()
+
+    members = get_members(cluster, cluster_name, member_names, role, force, 'reload')
+
+    content = {}
+    for member in members:
+        r = request_patroni(member, 'post', 'reload', content, auth_header(obj))
+        if r.status_code == 200:
+            click.echo('Success: reload on member {0}'.format(member.name))
```
CyberDem0n: It returns 200 if there are no changes to apply from yaml file: https://github.com/zalando/patroni/blob/master/patroni/api.py#L184

patroni/ctl.py
```diff
@@ -445,6 +445,31 @@ def parse_scheduled(scheduled):
...
+        elif r.status_code == 202:
+            click.echo('Success: reload scheduled on member {0}'.format(member.name))
```
CyberDem0n: `POST /reload` and `SIGHUP` are updating postgresql.conf and reloading it not immediately, but within `loop_wait` seconds. Just wondering, does it make sense to change this status message, to avoid confusion with `scheduled restart`, where meaning of `scheduled` is absolutely different?

dtseiler: Yes you are right. I was a little unsure of this myself but figured to start the PR and sort it out then.

So if code is 200, echo something to the effect of no changes to apply? Or just echo nothing?

And if 202, do we not say reload scheduled? Can we say that the reload finished if we don't know that?

patroni/ctl.py
dtseiler: Should we say no changes to apply then?

dtseiler: And would it make sense to say "Success" if there are no changes to apply? It isn't a failure. Is there some kind of "noop" prefix you'd prefer I use?

CyberDem0n: Well, it is definitely not an error, so I think something like "Nothing changed in Patroni config on {0}" should be good enough.

patroni/ctl.py
CyberDem0n: 202 means that there is something has changed in the config and eventually (within `loop_wait`) Patroni will apply them. Actually, it will not necessary case update of postgresql.conf and pg_ctl reload.

dtseiler: > it will not necessary case update of postgresql.conf and pg_ctl reload.

I'm not sure what you're saying here. 

What wording/language do we want to use here to tell the user? Something like "Reload request received and will be processed in loop_wait seconds"?

CyberDem0n: > I'm not sure what you're saying here.

One can update `tags` in patroni.yaml, and this update will not cause pg_ctl reload, because nothing in postgresql.conf was changed.


> "Reload request received and will be processed in loop_wait seconds" 

sounds good.

dtseiler:
@CyberDem0n both changes pushed. Thanks for the feedback.

alexeyklyukin:
👍

CyberDem0n:
:+1:

CyberDem0n:
Merged, thank you!

PR #723, opened 2018-06-26
K8s: Change API request timeout

sapezhko:
Set k8s API request timeout to `retry_timeout / 2.0` instead of hardcoded
connection timeout as `1` and read timeout as `retry_timeout / 3.0`. I think there is no need to distinguish connection timeout from read timeout in k8s dcs requests. Also `retry_timeout / 3.0` in my opinion is too small, especially with `retry_timeout == 10` set by default.

### DIFF
--- a/patroni/dcs/kubernetes.py
+++ b/patroni/dcs/kubernetes.py
@@ -37,7 +37,7 @@ class CoreV1ApiProxy(object):
         self._use_endpoints = use_endpoints
 
     def set_timeout(self, timeout):
-        self._request_timeout = (1, timeout / 3.0)
+        self._request_timeout = timeout / 2.0
 
     def __getattr__(self, func):
         if func.endswith('_kind'):


patroni/dcs/kubernetes.py
```diff
@@ -37,7 +37,7 @@ def __init__(self, use_endpoints=False):
         self._use_endpoints = use_endpoints
 
     def set_timeout(self, timeout):
-        self._request_timeout = (1, timeout / 3.0)
+        self._request_timeout = timeout / 2.0
```
CyberDem0n: There is a very good reason to have `timeout / 3.0` here.
In case is http request to k8s api has failed, it will do two retries.
With retry_timeout=10, in the worst case it will take up to 15 seconds to get response, what is not acceptable

sapezhko: No, `timeout` here is exactly `retry_timeout` from config. So it will wake up to 5 seconds to get response(in case of `timeout / 2.0`)

CyberDem0n: timeout is used here: https://github.com/kubernetes-client/python/blob/master/kubernetes/client/rest.py#L140-L144

In the end, API request is performed by urllib3 pool manager: https://github.com/kubernetes-client/python/blob/master/kubernetes/client/rest.py#L140-L144, which does TWO retries by default

sapezhko: I think it does 3 retries by default.
https://github.com/urllib3/urllib3/blob/master/urllib3/util/retry.py#L411
And in case in case of multiple read timeouts  `Retry` from utils will not work at all(because all retry time will be used by k8s client). I think the problem is that k8s client do retries by default using default settings of urllib3. Maybe need we turn off k8s retries since we have our own? Or maybe configure request timeout explicitly in DCS config.

CyberDem0n: > I think it does 3 retries by default.

That's the point. Every try will take up to specified timeout, therefore we divide it by 3.

> Maybe need we turn off k8s retries since we have our own?

No, they serve a different purpose.

CyberDem0n: We run more than 400 clusters on k8s managed by Patroni, and we spend a lot of time to find good values for different timeouts.

sapezhko: Ok, so why don't store requests timeouts/retries in DCS?

CyberDem0n: I don't really understand your question.

`retry_timeout` is already stored in DCS. If it is not there default value is assumed, which is 10 seconds.

sapezhko: Yes, `retry_timeout` is already stored. But request timeout is not.
```python
(1, timeout / 3.0)
```
It is undocumented magic. So, when you see in logs 
`
2018-06-27 03:11:09,006 WARNING: Retrying (Retry(total=2, connect=None, read=None, redirect=None, status=None)) after connection broken by 'ConnectTimeoutError(<urllib3.connection.VerifiedHTTPSConnection object at 0x7fc4b30c1e48>, 'Connection to 10.3.0.1 timed out. (connect timeout=1)')': /api/v1/namespaces/infra/pods?labelSelector=app%3Dpostgres%2Ccluster-name%3Dpostgres-cluster
`
or
`
2018-06-27 08:45:24,878 WARNING: Retrying (Retry(total=1, connect=None, read=None, redirect=None, status=None)) after connection broken by 'ReadTimeoutError("HTTPSConnectionPool(host='10.3.0.1', port=443): Read timed out. (read timeout=3.3333333333333335)",)': /api/v1/namespaces/infra/pods?labelSelector=app%3Dpostgres%2Ccluster-name%3Dpostgres-cluster
`
you understand nothing.

CyberDem0n: There is no magic here, request (read) timeout depends on and must be calculated from retry_timeout.

sapezhko: Ok

PR #724, opened 2018-06-26
max_worker_processes parameter was introduced only in 9.4

CyberDem0n:
exclude it from the list on 9.3 when building effective configuration

### DIFF
--- a/patroni/postgresql.py
+++ b/patroni/postgresql.py
@@ -835,10 +835,12 @@ class Postgresql(object):
 
         OPTIONS_MAPPING = {
             'max_connections': 'max_connections setting',
-            'max_worker_processes': 'max_worker_processes setting',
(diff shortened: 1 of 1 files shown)


erthalion:
Shouldn't be there any warning if this options is ignored for older versions?

CyberDem0n:
Why? There is just no such option in postgres 9.3...

erthalion:
If I understood correctly, it allows to specify `max_worker_process` in patroni config, am I right? In this case if it was specified, but PG version is old enough, it's going to be silently ignored?

CyberDem0n:
No, you are wrong.
It tries to calculate values of some sensible options from pg_controldata

erthalion:
:+1:

CyberDem0n:
:+1:

PR #730, opened 2018-07-03
Use enforced minimums in postgresX.yml files

dtseiler:
The 3 postgresX.yml files have max_wal_senders and max_replication_slots set as 5. Yes they are commented out, but the postgresql.py file would bump this to the minimum of 10, so I think it would be a better practice to have the example values at least at the minimum.

### DIFF
--- a/postgres0.yml
+++ b/postgres0.yml
@@ -31,8 +31,8 @@ bootstrap:
 #        wal_level: hot_standby
 #        hot_standby: "on"
 #        wal_keep_segments: 8
-#        max_wal_senders: 5
-#        max_replication_slots: 5
+#        max_wal_senders: 10
+#        max_replication_slots: 10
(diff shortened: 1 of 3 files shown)


dtseiler:
Sorry for multiple commits, I was just editing them via github in the browser and it seems to force a commit for each file. I can re-submit as one commit if you prefer.

CyberDem0n:
:+1:

CyberDem0n:
That's fine, usually we are doing "squash and merge" and get only one commit at the end.

alexeyklyukin:
:+1:

dtseiler:
I would have done it via my own local fork but I'm in the middle of debugging a patch for something else in patroni and just thought it would be quicker to edit in-browser.

alexeyklyukin:
@dtseiler merged, thank you!

PR #731, opened 2018-07-09
Log config change or not

dtseiler:
This adds INFO log messages that clearly state if configuration values were seen as changed by Patroni after SIGHUP/reload and warrant reloading (or if nothing was changed an no reloading is necessary).

This ended up being a lot simpler than I had imagined once I found postgresql.py:reload_config().

I add a log line in config.py:reload_local_configuration() since that function will short-circuit the process early if the local config wasn't changed. But the final determination of whether or not values have changed and need reloading is in postgresql.py:reload_config().

### DIFF
--- a/patroni/postgresql.py
+++ b/patroni/postgresql.py
@@ -335,10 +335,12 @@ class Postgresql(object):
                         if new_value is None or not compare_values(r[3], unit, r[1], new_value):
                             if r[4] == 'postmaster':
                                 pending_restart = True
+                                logger.info('Changed %s from %s to %s (restart required)', r[0], r[1], new_value)
                                 if config.get('use_unix_socket') and r[0] == 'unix_socket_directories'\
                                         or r[0] in ('listen_addresses', 'port'):
                                     local_connection_address_changed = True
                             else:
+                                logger.info('Changed %s from %s to %s', r[0], r[1], new_value)
                                 conf_changed = True
                 for param in changes:
                     if param in server_parameters:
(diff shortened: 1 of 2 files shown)


patroni/postgresql.py
```diff
@@ -375,7 +375,10 @@ def reload_config(self, config):
             self._replace_pg_hba()
 
         if conf_changed or hba_changed:
+            logger.info('Configuration items changed, reloading configuration.')
             self.reload()
+        else:
+            logger.info('No configuration items changed, nothing to reload.')
```
CyberDem0n: I would suggest changing the message to `No postgresql configuration items changed, nothing to reload.` to make it look different from `config.py`

patroni/postgresql.py
```diff
@@ -375,7 +375,10 @@ def reload_config(self, config):
             self._replace_pg_hba()
 
         if conf_changed or hba_changed:
+            logger.info('Configuration items changed, reloading configuration.')
             self.reload()
+        else:
```
CyberDem0n:
```diff
- else:
+ elif not pending_restart:
```

Otherwise it will show the message on the next line when you change for example only `max_connections`.

dtseiler:
Great feedback. Do you want me to update my PR with your changes? Or did your changes get included?

CyberDem0n:
Please just include my suggestions to your PR.

CyberDem0n:
:+1:

alexeyklyukin:
@dtseiler Would be nice to report pg_hba.conf changes as well, putiting the log after the check at https://github.com/zalando/patroni/pull/731/files#diff-d8c1582337c8426ef750420dc6b76949L361 if the check returns true.

dtseiler:
Good point, I'll take a stab at it shortly.

dtseiler:
Actually I wonder if this wouldn't be more noisy. Does it return true if things are simply changed orders? The pg_hba isn't a key-value system like the postgresql conf.

alexeyklyukin:
@dtseiler yes, it does return true on the order change, and the order of lines in pg_hba.conf is significant. It is not necessary to output the actual diff.

dtseiler:
If you just wanted to know if the HBA was changed, that should already be done on line 381: https://github.com/zalando/patroni/pull/731/files#diff-d8c1582337c8426ef750420dc6b76949R381

alexeyklyukin:
👍

CyberDem0n:
Merged, thank you!

PR #733, opened 2018-07-11
Added setting state to "stopped" when a member is stopped in Ha.shutdown

tonys66:
This will allow the member to show as "stopped" when viewed through a "patronictl list" command.

### DIFF
--- a/patroni/ha.py
+++ b/patroni/ha.py
@@ -1168,7 +1168,6 @@ class Ha(object):
             if not self.state_handler.is_running():
                 if self.has_lock():
                     self.dcs.delete_leader()
-                self.state_handler.set_state('stopped')
                 self.touch_member()
             else:
                 # XXX: what about when Patroni is started as the wrong user that has access to the watchdog device
--- a/patroni/ha.py
+++ b/patroni/ha.py
@@ -1168,6 +1168,8 @@ class Ha(object):
             if not self.state_handler.is_running():
                 if self.has_lock():
                     self.dcs.delete_leader()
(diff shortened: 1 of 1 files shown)


patroni/ha.py
```diff
@@ -1168,6 +1168,8 @@ def shutdown(self):
             if not self.state_handler.is_running():
                 if self.has_lock():
                     self.dcs.delete_leader()
+                self.state_handler.set_state('stopped')
```
CyberDem0n: The state should already be `stopped` at this moment: https://github.com/zalando/patroni/blob/master/patroni/postgresql.py#L959

tonys66: I did not experience the state changing from running to anything else.  It simply stayed as "running" and eventually, the etcd entry timed 
patroni_stopped_state
out and was deleted.  See attached image for an example.

tonys66: I did not experience the "stopped" status being recorded.  The only status I saw was "running" until the etcd entry timed out.  The attached image shows the output from patronictl as the member was running, two after it stoped, and a final once the etcd entry timed out.

patroni_stopped_state

CyberDem0n: I just wanted to tell that it would be enough to call `touch_member`, without calling `self.state_handler.set_state('stopped')`

tonys66: Ah.... got it.  I'll try that so there is one less call.  If that works, can the change be accepted?

alexeyklyukin:
👍

alexeyklyukin:
@tonys66 merged, thank you!

PR #738, opened 2018-07-17
Improve logging when stale postmaster.pid matches running process

ants:
Currently the informational message logged is beyond confusing. This
improves the logging so there is some indication what this message is
about and that it is somewhat normal.

Also removes double call to _is_postmaster_process(), _from_pidfile()
already checks that.

### DIFF
--- a/patroni/postmaster.py
+++ b/patroni/postmaster.py
@@ -42,7 +42,8 @@ class PostmasterProcess(psutil.Process):
         try:
             start_time = int(self._postmaster_pid.get('start_time', 0))
             if start_time and abs(self.create_time() - start_time) > 3:
-                logger.info('Too much difference between %s and %s', self.create_time(), start_time)
+                logger.info('Process %s is not postmaster, too much difference between PID file start time %s and '
+                            'process start time %s', self.pid, self.create_time(), start_time)
                 return False
         except ValueError:
             logger.warning('Garbage start time value in pid file: %r', self._postmaster_pid.get('start_time'))
@@ -148,6 +149,7 @@ class PostmasterProcess(psutil.Process):
                 # Important!!! Unlink of postmaster.pid isn't an option, because it has a lot of nasty race conditions.
                 # Luckily there is a workaround to this problem, we can pass the pid from postmaster.pid
                 # in the `PG_GRANDPARENT_PID` environment variable and postmaster will ignore it.
+                logger.info("Telling pg_ctl that it is safe to ignore postmaster.pid for process %s", proc.pid)
                 env['PG_GRANDPARENT_PID'] = str(proc.pid)
         except psutil.NoSuchProcess:
             pass


patroni/postmaster.py
```diff
@@ -140,14 +141,15 @@ def start(pgcommand, data_dir, conf, options):
         env = {p: os.environ[p] for p in ('PATH', 'LD_LIBRARY_PATH', 'LC_ALL', 'LANG') if p in os.environ}
         try:
             proc = PostmasterProcess._from_pidfile(data_dir)
-            if proc and not proc._is_postmaster_process():
+            if proc:
```
CyberDem0n: Looks like a leftover from manual testing

alexeyklyukin:
> Also removes double call to _is_postmaster_process(), _from_pidfile()
already checks that.

It doesn't, although its cousin without the underscore (which is `PostmasterProcess.from_pidfile()`) does. We could actually use the `PostmasterProcess.from_pidfile()` in line 143, which would also let us remove the outer try/except block .

alexeyklyukin:
👍

patroni/postmaster.py
```diff
@@ -138,19 +139,17 @@ def start(pgcommand, data_dir, conf, options):
...
-                # file in the data directory. Although Patroni already detected that the running process
-                # corresponding to the postmaster.pid is not a postmaster, the new postmaster might fail
-                # to start, because it thinks that postmaster.pid is already locked.
-                # Important!!! Unlink of postmaster.pid isn't an option, because it has a lot of nasty race conditions.
-                # Luckily there is a workaround to this problem, we can pass the pid from postmaster.pid
-                # in the `PG_GRANDPARENT_PID` environment variable and postmaster will ignore it.
-                env['PG_GRANDPARENT_PID'] = str(proc.pid)
-        except psutil.NoSuchProcess:
-            pass
+        proc = PostmasterProcess.from_pidfile(data_dir)
```
CyberDem0n: Let's imagine that pid from `postmaster.pid` is matching with the pid of one of the Patroni threads. In this case `PostmasterProcess.from_pidfile` will return `None`, because `start_time` from `postmaster.pid` doesn't match with process `create_time`. Patroni will try to start postgres, but postgres will refuse to do it, because `postmaster.pid`
 is already "locked".

The only way to solve such a problem is to pass the `pid` from `postmaster.pid` in the `PG_GRANDPARENT_PID`.

The original code was doing that. The current code will do quite an opposite. It will force the new postgres to ignore existing running postgres process...

alexeyklyukin: good catch, there is a "not is_postmaster_process" in the previous block, however, from_pidfile checks for "proc and is_postmaster_process", so the logic will be indeed reversed.

ants: Yes, you are correct. I'll just back out the logic change. I misread the original code due to seeing a log with the "too much difference" line repeated. Sorry about the noise.

CyberDem0n:
:+1:

alexeyklyukin:
👍

alexeyklyukin:
@ants merged, thank you!

Issue #742, opened 2018-07-19
Patroni cannot start, consul answer timeout


imcitius:
Hello

Just got strange situation with new replica installed in remote datacenter.
Patroni cannot start up with this error:

```
Jul 19 15:09:47 pgsql patroni.py[68505]: 2018-07-19 15:09:47,224 INFO: Lock owner: pgsql-04; I am pgsql-readonly-02
Jul 19 15:09:47 pgsql patroni.py[68505]: 2018-07-19 15:09:47,224 INFO: starting as a secondary
Jul 19 15:09:47 pgsql patroni.py[68505]: 2018-07-19 15:09:47,247 INFO: postmaster pid=68580
Jul 19 15:09:47 pgsql patroni.py[68505]: 2018-07-19 15:09:47,266 INFO: Lock owner: pgsql-04; I am pgsql-readonly-02
Jul 19 15:09:47 pgsql patroni.py[68505]: localhost:5432 - no response
Jul 19 15:09:47 pgsql patroni.py[68505]: 2018-07-19 15:09:47,307 INFO: restarting after failure in progress
Jul 19 15:09:47 pgsql patroni.py[68505]: 2018-07-19 15:09:47.370 UTC 68580 @ from  [vxid: txid:0] [] LOG:  redirecting log output to logging collector process
Jul 19 15:09:47 pgsql patroni.py[68505]: 2018-07-19 15:09:47.370 UTC 68580 @ from  [vxid: txid:0] [] HINT:  Future log output will appear in directory "/var/log/postgresql".
Jul 19 15:09:48 pgsql patroni.py[68505]: localhost:5432 - rejecting connections
Jul 19 15:09:48 pgsql patroni.py[68505]: localhost:5432 - rejecting connections
Jul 19 15:09:49 pgsql patroni.py[68505]: localhost:5432 - accepting connections
Jul 19 15:09:49 pgsql patroni.py[68505]: 2018-07-19 15:09:49,366 INFO: Lock owner: pgsql-04; I am pgsql-readonly-02
Jul 19 15:09:49 pgsql patroni.py[68505]: 2018-07-19 15:09:49,366 INFO: does not have lock
Jul 19 15:09:49 pgsql patroni.py[68505]: 2018-07-19 15:09:49,366 INFO: establishing a new patroni connection to the postgres cluster
Jul 19 15:09:49 pgsql patroni.py[68505]: 2018-07-19 15:09:49,413 INFO: no action.  i am a secondary and i am following a leader
[...]

Issue #734, opened 2018-07-13
'ReadTimeoutError' getting consul leader with index property in secondary


carlosmkb:
In secondary the following error occurs almost every second. If the secondary becomes new primary this erros stops and starts in the other secondary.
```
2018-07-13 09:34:04,711 ERROR: watch
Traceback (most recent call last):
File "/usr/lib/python2.7/site-packages/patroni/dcs/consul.py", line 406, in watch
idx, _ = self._client.kv.get(self.leader_path, index=leader_index, wait=str(timeout) + 's')
File "/usr/lib/python2.7/site-packages/consul/base.py", line 554, in get
params=params)
File "/usr/lib/python2.7/site-packages/patroni/dcs/consul.py", line 102, in wrapper
return callback(self.response(self.http.request(method.upper(), self.uri(path, params), **kwargs)))
File "/usr/lib/python2.7/site-packages/urllib3/request.py", line 68, in request
**urlopen_kw)
File "/usr/lib/python2.7/site-packages/urllib3/request.py", line 89, in request_encode_url
return self.urlopen(method, url, **extra_kw)
File "/usr/lib/python2.7/site-packages/urllib3/poolmanager.py", line 322, in urlopen
response = conn.urlopen(method, u.request_uri, **kw)
File "/usr/lib/python2.7/site-packages/urllib3/connectionpool.py", line 638, in urlopen
_stacktrace=sys.exc_info()[2])
File "/usr/lib/python2.7/site-packages/urllib3/util/retry.py", line 398, in increment
raise MaxRetryError(_pool, url, error or ResponseError(cause))
MaxRetryError: HTTPConnectionPool(host='server.raid.test', port=8500): Max retries exceeded with url: /v1/kv/service/postgres/leader?index=159&wait=3.26965403557s (Caused by ReadTimeoutError("HTTPConnectionPool(host='server.raid.test', port=8500): Read timed out. (read timeout=3.33333333333)",))
2018-07-13 09:34:04,716 INFO: Lock owner: postgres-2; I am postgres-1
2018-07-13 09:34:04,716 INFO: does not have lock
2018-07-13 09:34:04,718 INFO: no action.  i am a secondary and i am following a leader
```

CyberDem0n:
Please try https://github.com/zalando/patroni/pull/744, I hope it should fix your problem

PR #744, opened 2018-07-19
Convert query params from list to dict

CyberDem0n:
Patroni is relying on params to determine timeout and amount of retries when executing api requests to consul.
Starting from v1.1.0 python-consul changed internal API and started using `list` instead of `dict` to pass query parameters. Such change broke "watch" functionality.

Fixes https://github.com/zalando/patroni/issues/742 and https://github.com/zalando/patroni/issues/734


alexeyklyukin:
👍

PR #746, opened 2018-07-20
Add EnvironmentFile directive in sample systemd service file

dtseiler:
Add an EnvironmentFile directive to read in a configuration file with environment variables. The "-" prefix means it can proceed if the file doesn't exist.

This would alllow users to keep sensitive information like the SUPERUSER/REPLICATION passwords in the config file separate from a YAML file that might be deployed from source control.

### DIFF
--- a/extras/startup-scripts/patroni.service
+++ b/extras/startup-scripts/patroni.service
@@ -11,6 +11,9 @@ Type=simple
 User=postgres
 Group=postgres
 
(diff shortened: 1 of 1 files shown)


hjacobs:
LGTM

alexeyklyukin:
👍

CyberDem0n:
:+1:

CyberDem0n:
Merged, thank you!

PR #749, opened 2018-07-23
Grant delete on endpoints and configmaps in RBAC.

alexeyklyukin:
'patronictl remove' deletes the cluster configuration (stored either in
configmaps or endpoints) and cannot be run from the postgres pod w/o
'delete' on those objects being granted to the pod service account.


alexeyklyukin:
👍

CyberDem0n:
:+1:

PR #751, opened 2018-07-23
async is a keyword in python3.7

### LINKED ISSUE #750: python3.7: "async" is a keyword now
Installing patroni with python3.7 fails: (https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=904369)
```
patroni (1.4.4-2) wird eingerichtet ...
  File "/usr/lib/python3/dist-packages/patroni/ha.py", line 610
    'offline':          dict(stop='fast', checkpoint=False, release=False, offline=True, async=False),
                                                                                             ^
SyntaxError: invalid syntax
```
The problem exists also in python3-kubernetes where the parameter was renamed from "async" to "async_req" in 7.0.0a1: https://github.com/kubernetes-client/python/blob/master/CHANGELOG.md

A new release with a fix would be appreciated so we can update the Debian package. Thanks!

ChristophBerg:
Setting up patroni (1.4.4-1) ...
    File "/usr/lib/python3/dist-packages/patroni/ha.py", line 610
      'offline':          dict(stop='fast', checkpoint=False, release=False, offline=True, async=False),
                                                                                               ^
  SyntaxError: invalid syntax

Fix #750 by replacing dict member "async" with "async_req".

### DIFF
--- a/patroni/ha.py
+++ b/patroni/ha.py
@@ -597,10 +597,10 @@ class Ha(object):
                 PostgreSQL as quickly as possible without regard for data durability. May only be called synchronously.
         """
         mode_control = {
-            'offline':          dict(stop='fast', checkpoint=False, release=False, offline=True, async=False),
-            'graceful':         dict(stop='fast', checkpoint=True, release=True, offline=False, async=False),
-            'immediate':        dict(stop='immediate', checkpoint=False, release=True, offline=False, async=True),
-            'immediate-nolock': dict(stop='immediate', checkpoint=False, release=False, offline=False, async=True),
+            'offline':          dict(stop='fast', checkpoint=False, release=False, offline=True, async_req=False),
+            'graceful':         dict(stop='fast', checkpoint=True, release=True, offline=False, async_req=False),
+            'immediate':        dict(stop='immediate', checkpoint=False, release=True, offline=False, async_req=True),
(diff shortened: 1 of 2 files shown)


CyberDem0n:
@ChristophBerg,

could you please also update requirements.txt?

```
-kubernetes>=2.0.0,<=6.0.0,!=4.0.*,!=5.0.*
+kubernetes>=2.0.0,<=7.0.0,!=4.0.*,!=5.0.*
```

I hope the new version of `kubernetes` package will be released soon.

CyberDem0n:
:+1:

hjacobs:
Can we add a test for Python 3.7 in .travis.yml, please?

hjacobs:
We also need to update `setup.py`: https://github.com/zalando/patroni/blob/master/setup.py#L55 (Python 3.6 is also missing there)

CyberDem0n:
Please, don't mess up with .travis.yml, it has some sophisticated logic inside to have a good balance between the number of builds and build time.

I absolutely agree that setup.py should be updated, but we can do it with separate PR.

hjacobs:
@CyberDem0n ok, just saw your comment on #750

hjacobs:
The consul tests are failing (but other tests pass): https://travis-ci.org/zalando/patroni/jobs/407270411
Is this known/ok? @CyberDem0n

hjacobs:
:+1:

PR #752, opened 2018-07-23
Python 3.6 and 3.7 are now supported, too

hjacobs:
Follow-up to #751: update supported Python versions in `setup.py`.


hjacobs:
:+1:

Issue #753, opened 2018-07-25
replica rewind in a paused state


alexeyklyukin:
I've been playing with the cluster that ran out of disk space; during the period before the disk resize replica has managed to create 22 new timelines, failing whenever writing to disk, starting again and repeating the promotion/failure. Somehow it stopped at a certain point; however, this behavior made the replica unrecoverable after we solved the disk space issues, as it fetched the timeline history from the local wal directory and attempted to connect to the recovered original master with the very latest timeline, failing with a message like this:

'highest timeline 2 of the primary is behind recovery timeline 23 on test-cluster.'

Apparently I haven't noticed  initially neither that Patroni had already tried to run pg_rewind and failed, complaining about a missing WAL segment (on timeline 1) that it couldn't fetch from the original archive, nor the fact that higher timeline histories were never archived. Therefore, I decided to stop the replica to run pg_rewind manually by putting patroni into the pause mode and running pg_ctl -my stop on the replica's data directory. To my surprise, Patroni immediately restarted the replica I stopped.

It looks like there is a bug in the logic that decides whether the replica that is not running while the cluster is in the pause state has to start or not. Patroni starts a replica during the pause mode when it suspects the replica has served as the master right before being stopped as the result of being demoted. The reason Patroni starts it is to apply pg_rewind, and it only does it when pg_rewind can be applied; however, we should not attempt to start the cluster once pg_rewind has already finished (whatever the result it). The penultimate lines of the code below (taken from ha.py:1110) looks like a bug to me. I think the condition on rewind_executed should be reversed, namely,  rewind_executed or we don't need it or cannot run it - do not attempt to start the replica.

```
         if not self.state_handler.is_healthy():
                if self.is_paused():
                    if self.has_lock():
                        self.dcs.delete_leader()
                        self.dcs.reset_cluster()
                        return 'removed leader lock because postgres is not running'
                    elif not (self.state_handler.rewind_executed or
                              self.state_handler.need_rewind and self.state_handler.can_rewind):
                        return 'postgres is not running'
```

PR #754, opened 2018-07-25
Fix condition for the replica start due to pg_rewind in paused state.

alexeyklyukin:
Avoid starting the replica that had already executed pg_rewind before.

Described in #753

### DIFF
--- a/patroni/ha.py
+++ b/patroni/ha.py
@@ -1113,8 +1113,8 @@ class Ha(object):
                         self.dcs.delete_leader()
                         self.dcs.reset_cluster()
                         return 'removed leader lock because postgres is not running'
-                    elif not (self.state_handler.rewind_executed or
-                              self.state_handler.need_rewind and self.state_handler.can_rewind):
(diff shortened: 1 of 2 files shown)


patroni/ha.py
```diff
@@ -1113,8 +1113,8 @@ def _run_cycle(self):
                         self.dcs.delete_leader()
                         self.dcs.reset_cluster()
                         return 'removed leader lock because postgres is not running'
-                    elif not (self.state_handler.rewind_executed or
-                              self.state_handler.need_rewind and self.state_handler.can_rewind):
+                    elif (self.state_handler.rewind_executed or
```
CyberDem0n: It feels wrong to me.
We want to start postgres if pg_rewind was executed successfully and don't want to start it if pg_rewind failed.
The new version will not try to start postgres even after successful pg_rewind.

CyberDem0n:
:+1:

alexeyklyukin:
👍

PR #764, opened 2018-08-05
Add "cluster_unlocked" field

erthalion:
Add "cluster_unlocked" field for patroni api to be able to figure out if a
master is there from patroni point of view. It can be useful, when you have an
alert, based on Auto Scaling Groups, and then ASG decided to shutdown the
current master, spin up a new instance but the current master shutdown is
stuck. In this situation the current master is no longer a part of ASG, but
patroni and Postgres are still alive on the instance, which means a new replica
will not be promoted yet - this will lead to a false alert, saying that your
cluster doesn't have any master node.

### DIFF
--- a/patroni/api.py
+++ b/patroni/api.py
@@ -415,10 +415,7 @@ class RestApiHandler(BaseHTTPRequestHandler):
 
     def get_postgresql_status(self, retry=False):
         try:
-            cluster = (
-                self.server.patroni.dcs.cluster or
-                self.server.patroni.dcs.get_cluster()
-            )
+            cluster = self.server.patroni.dcs.cluster
 
             if self.server.patroni.postgresql.state not in ('running', 'restarting', 'starting'):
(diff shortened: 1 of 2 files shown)


patroni/api.py
```diff
@@ -415,6 +415,11 @@ def query(self, sql, *params, **kwargs):
 
     def get_postgresql_status(self, retry=False):
         try:
+            cluster = (
+                self.server.patroni.dcs.cluster or
+                self.server.patroni.dcs.get_cluster()
```
CyberDem0n: I would actually avoid calling `get_cluster` method and always use the cached version.

Also the line https://github.com/zalando/patroni/pull/764/files#diff-790530624eaa21e5fb614fc321aef277R460 could be removed.

erthalion:
:+1:

CyberDem0n:
:+1:

Issue #775, opened 2018-08-14
Getting the  error 403 while try do second switchover


tigranav:
I have setup with PATRONI 1.4.5 with Kubernetes 3.9 as DCS.
When i try do **second** switchover with `patronyctl` i getting error :
```
Switchover failed, details: 503, failed to write switchover key into DCS
```
i can do second switchover only after delete configmap  `xxxxxx-failover`.

I guess couse of error is PATRONI trying create configmap always, instead of  update it.
may be error in https://github.com/zalando/patroni/blob/b282a0f254d69604f4e66218af10c1968eed9953/patroni/dcs/kubernetes.py#L325

CyberDem0n:
In case if config map already exists Patroni does an update.
It seems something is wrong with your config or may be Patroni is not allowed to overwrite config maps.

Can you please share `kubernetes:` section of your patroni.yaml and output of `kubectl get cm xxxxxx-failover`?

tigranav:
from patroni config:
```kubernetes:
  labels:
    pgcluster: common
  namespace: anne-test
  pod_ip: 10.130.0.191
  role_label: pgtype
  scope_label: pgcluster
```
from openshift:
```$ oc get cm common-failover
NAME              DATA      AGE
common-failover   0         3d
$ oc get cm common-failover -o json
{
    "apiVersion": "v1",
    "kind": "ConfigMap",
    "metadata": {
        "creationTimestamp": "2018-08-10T11:11:37Z",
        "labels": {
            "pgcluster": "common"
        },
        "name": "common-failover",
        "namespace": "anne-test",
        "resourceVersion": "52007420",
        "selfLink": "/api/v1/namespaces/anne-test/configmaps/common-failover",
        "uid": "25f9f70b-9c8e-11e8-b884-fa163e5f2c4f"
    }
}
$ oc describe cm/common-failover
Name:           common-failover
Namespace:      anne-test
Labels:         pgcluster=common
Annotations:    <none>

Data
====
```

CyberDem0n:
Can it be that Patroni doesn't have permissions to update config maps?
In some cases (especially on open shift) you have to allow that explicitly: https://github.com/zalando/patroni/blob/master/kubernetes/patroni_k8s.yaml#L126-L186

tigranav:
Patroni can update other configmaps.

CyberDem0n:
`Switchover failed, details: 503, failed to write switchover key into DCS` - this log is from patronictl.
What is in Patroni logs?

[...]

PR #782, opened 2018-08-16
None changed to empty string

anikin-aa:
`None` is change to `null` when dict is converted to json via `json.dumps(data)`. Kubernetes can't handle `null` as a value in annotations.

so, `None` is change to empty string.

https://github.com/zalando/patroni/issues/775

### DIFF
--- a/patroni/dcs/kubernetes.py
+++ b/patroni/dcs/kubernetes.py
@@ -321,7 +321,7 @@ class Kubernetes(AbstractDCS):
         """Unused"""
 
     def manual_failover(self, leader, candidate, scheduled_at=None, index=None):
-        annotations = {'leader': leader or None, 'member': candidate or None, 'scheduled_at': scheduled_at}
+        annotations = {'leader': leader or '', 'member': candidate or '', 'scheduled_at': scheduled_at or ''}
         patch = bool(self.cluster and isinstance(self.cluster.failover, Failover) and self.cluster.failover.index)
         return self.patch_or_create(self.failover_path, annotations, index, bool(index or patch), False)
 


patroni/dcs/kubernetes.py
```diff
@@ -321,7 +321,7 @@ def set_failover_value(self, value, index=None):
         """Unused"""
 
     def manual_failover(self, leader, candidate, scheduled_at=None, index=None):
-        annotations = {'leader': leader or None, 'member': candidate or None, 'scheduled_at': scheduled_at}
+        annotations = {'leader': leader or '', 'member': candidate or '', 'scheduled_at': scheduled_at or ''}
```
CyberDem0n: Kubernetes doesn't like annotations with null values when you create a new object (ConfigMap or Endpoint), therefore we are removing them here: https://github.com/zalando/patroni/pull/782/files#diff-87299faee0431ddce0848d6ff7726df7R264

But in case of PATCH request, null values have a different semantic. If a given annotation has null value in the PATCH request, Kubernetes supposed to remove such annotation from the object.

You are replacing null values with empty string, and as a result annotations are not removed:
```yaml
metadata:
  annotations:
    leader: ""
    member: ""
    scheduled_at: ""
```

From one side Patroni is able to handle it correctly (and acceptance tests prove that), but from another side, this is not how it supposed to work.

Unfortunately I don't have access to openshift clusters and can't investigate further.

anikin-aa: Ok, I got Your point, will investigate a little bit more.

anikin-aa: well, that is issue with kubernetes. just tested on oshit 3.9 (kubernetes 1.9) everything works fine, closing this PR.

CyberDem0n: That's very interesting.
I think I tested it on all versions starting from at least kubernetes 1.7 and it was working fine.
Which version does it fail on?

anikin-aa: server versions
```
openshift v1.5.1
kubernetes v1.5.2+43a9be4
```

Issue #664, opened 2018-04-13
RFC: design for supporting synchronous quorum commit in Patroni


jberkus:
With Postgres 10, support for synchronous quorum commit is a feature, and makes using synchrounous replication to reduce data loss during failover much more practical for Patroni.  Here's a draft of how this would potentially work:

1. Clusters would get a new setting, synchronous_quorum(int), defaulting to 0.
2. If turned on (> 0), all Patroni nodes would set synchronous_standby_names to a quorum of `any $synchronous_quorum ( list, of, all, nodes )`
3. The polling cycle would add an extra check to see if new nodes have been added, and if so, update synchronous_standby_names.

For example, if synchronous_quorum=1 and there are 4 nodes, the setting on each Postgres would be:

```
synchronous_standby_names = 'any 1 ( patroni-0, patroni-1, patroni-2, patroni-3 )'
```

Questions:

* We don't have cluster settings, patroni settings are per-node.  If some nodes had synchronous_quorum turned on, and other nodes did not, it could result in unexpected behavior.  Is this a real problem?
* In the kubernetes-native version, we no longer maintain a list of all nodes in the DCS.  Presumably we would need to add one to the ConfigMap for this to work?  Or is it sufficient to estimate based on the statefulset size?
* How do we want to handle nodes dying or being removed?  Do we want to update the list of replicas?  If so, how do we want to handle this in order to not create a data-losing race condition?  If we don't update automatically, how do we want to enable administrators to do a cleanup?
* Will admins be able to push a config change of synchronous_quorum=0 to a single remaining node, in the event that they're down to one node and rescuing the cluster?

jberkus:
Ants,

That makes sense.  It's also pointless to promote a node if there is no candidate sync rep, since that node will still not be able to accept synchronous writes, so that all works.

What I'm still concerned about is maintaining the list of candidate nodes, particually in a "rolling blackout" situation.  While you've eliminated a race condition by saying only the master can edit the list of nodes, there's still an inherent danger if the master is in the process of adding and removing nodes from the list and then goes dark (this is not a new problem, Cassandra and similar databases have a lot of logic devoted to this issue).

A possible amelioration of uncertainty caused by adding and removing nodes would be just to make it slow; make the timeout on either adding or dropping a node from the list 5 minutes, for example.  The drawback to that is that it would take a lot longer for a cluster to right itself when the nodes come back.

ants:
[...]

PR #787, opened 2018-08-22
WIP Quorum commit feature #2

anikin-aa:
This a PR for quorum commit feature, based on https://github.com/zalando/patroni/pull/672.

But, with minor changes:

* Only one parameter for quorum `synchronous_quorum` - number of members that should be in sync.
* Complicated `QuorumStateResolver` logic removed.

as described here https://github.com/zalando/patroni/issues/664.

Almost all tests are done (need to add new one, also), work is still in progress.

### DIFF
--- a/patroni/ha.py
+++ b/patroni/ha.py
@@ -439,7 +439,7 @@ class Ha(object):
                 # Somebody else updated sync state, it may be due to us losing the lock. To be safe, postpone
                 # promotion until next cycle. TODO: trigger immediate retry of run_cycle
                 return 'Postponing promotion because synchronous replication state was updated by somebody else'
-            self.state_handler.set_synchronous_standby(1 if self.is_synchronous_mode_strict() else None,
+            self.state_handler.set_synchronous_state(1 if self.is_synchronous_mode_strict() else None,
                                                        {} if self.is_synchronous_mode_strict() else None)
             if self.state_handler.role != 'master':
                 self._async_executor.schedule('promote')


CyberDem0n:
It looks like you removed a lot of useful code from the original PR.

It also doesn't follow a specification from [here](https://github.com/zalando/patroni/issues/664#issuecomment-381307309) and [here](https://github.com/zalando/patroni/issues/664#issuecomment-381789441)

anikin-aa:
Yes, I know this. My changes are not so complicated, but they are working. 
May be its acceptable to mark them like `Phase 1` of implementation, or something like that.

CyberDem0n:
Well, let's assume that you have `synchronous_standby_names = 'any 1 ( patroni-0, patroni-1, patroni-2, patroni-3 )'`

That basically means only one node is syncronous, but you don't know which one.
If the primary failed and the only synchronous node also failed, remainig three nodes will successfully elect the new leader and there'll be data loss.

anikin-aa:
Ok, I got You. I've also  done some changes based on @ants code, and its working too. I am interested in such functionality (quorum replication)  in general.
So its better to close this  PR and raise a new one based on @ants code, right ?

Cause I tried to raise PR with my changes to @ants and I have no response from him.

> If the primary failed and the only synchronous node also failed, remainig three nodes will successfully elect the new leader and there'll be data loss.

Anyway, I can't get  how @ants logic will prevent from such situations

CyberDem0n:
> Anyway, I can't get how @ants logic will prevent from such situations

[_is_healthiest_node](https://github.com/zalando/patroni/pull/672/files#diff-57c53e988b775ee2902488fb33c1f433R511) is taking care of it.

PR #788, opened 2018-08-22
Add pgbackrest support

sharmay:
Add following new parameters
args
keep_data
no_params

Here is an example:
```
postgresql:
    create_replica_method:
        - pgBackRest
        - basebackup
    pgBackReset:
        command: /usr/bin/pgbackrest
        args:
            - "--stanza=mydb"
            - "--delta"
            - "restore"
        keep_data: True
        no_params: True
```
Updated documentation.

Please let me know your feedback.

### DIFF
--- a/docs/replica_bootstrap.rst
+++ b/docs/replica_bootstrap.rst
@@ -69,8 +69,21 @@ scripts to clone a new replica. Those are configured in the ``postgresql`` confi
 
     postgresql:
         create_replica_methods:
+            - <method name>
+        <method name>:
+            command: <command name>
+            args:
+                - '<arguments>'
+            keep_data: True
+            no_params: True
+            no_master: 1
+
+.. code:: YAML
+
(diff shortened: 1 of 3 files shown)


docs/replica_bootstrap.rst
```diff
@@ -65,12 +65,25 @@ for outdated backup files. Some people prefer other backup solutions, such as ``
...
+            no_params: True
+            no_master: 1
+
 .. code:: YAML
 
     postgresql:
         create_replica_methods:
             - wal_e
-            - basebackup
+            - pgbackrest
```
CyberDem0n: basebackup is the default name for pg_basebackup. Why did you replace it?

sharmay: Thank you for catching wrong section delete.

sharmay:
Will be resubmitting it soon with all changes.

Issue #785, opened 2018-08-20
switchover show CERTIFICATE_VERIFY_FAILED stack trace


wilfriedroset:
When performing a switchover patronictl fails due to self-sign SSL cert

```
# patronictl -c config.yml switchover
Master [patroni-99-node-1]:
Candidate ['patroni-99-backup-1', 'patroni-99-node-2'] []: patroni-99-node-2
When should the switchover take place (e.g. 2015-10-01T14:30)  [now]:
Current cluster topology
+-----------+---------------------+------------+--------+---------+-----------+
|  Cluster  |        Member       |    Host    |  Role  |  State  | Lag in MB |
+-----------+---------------------+------------+--------+---------+-----------+
| cluster99 | patroni-99-backup-1 | XXXX |        | running |           |
| cluster99 |  patroni-99-node-1  | XXXX | Leader | running |           |
| cluster99 |  patroni-99-node-2  | XXXX |        | running |           |
+-----------+---------------------+------------+--------+---------+-----------+
Are you sure you want to switchover cluster cluster99, demoting current master patroni-99-node-1? [y/N]: y
2018-08-20 15:59:57,292 - ERROR - None
Traceback (most recent call last):
  File "/usr/lib/python3/dist-packages/urllib3/connectionpool.py", line 594, in urlopen
    chunked=chunked)
  File "/usr/lib/python3/dist-packages/urllib3/connectionpool.py", line 350, in _make_request
    self._validate_conn(conn)
  File "/usr/lib/python3/dist-packages/urllib3/connectionpool.py", line 837, in _validate_conn
    conn.connect()
  File "/usr/lib/python3/dist-packages/urllib3/connection.py", line 323, in connect
    ssl_context=context)
  File "/usr/lib/python3/dist-packages/urllib3/util/ssl_.py", line 324, in ssl_wrap_socket
    return context.wrap_socket(sock, server_hostname=server_hostname)
  File "/usr/lib/python3.5/ssl.py", line 385, in wrap_socket
    _context=self)
  File "/usr/lib/python3.5/ssl.py", line 760, in __init__
    self.do_handshake()
  File "/usr/lib/python3.5/ssl.py", line 996, in do_handshake
    self._sslobj.do_handshake()
  File "/usr/lib/python3.5/ssl.py", line 641, in do_handshake
    self._sslobj.do_handshake()
ssl.SSLError: [SSL: CERTIFICATE_VERIFY_FAILED] certificate verify failed (_ssl.c:720)

[...]

PR #790, opened 2018-08-24
Add patronictl -k/--insecure flag and suport for restapi cert

wilfriedroset:
Fixes https://github.com/zalando/patroni/issues/785

### DIFF
--- a/patroni/ctl.py
+++ b/patroni/ctl.py
@@ -103,15 +103,20 @@ option_watch = click.option('-W', is_flag=True, help='Auto update the screen eve
 option_force = click.option('--force', is_flag=True, help='Do not ask for confirmation at any point')
 arg_cluster_name = click.argument('cluster_name', required=False,
                                   default=lambda: click.get_current_context().obj.get('scope'))
+option_insecure = click.option('-k', '--insecure', is_flag=True, help='Allow connections to SSL sites without certs')
 
 
 @click.group()
 @click.option('--config-file', '-c', help='Configuration file', default=CONFIG_FILE_PATH)
 @click.option('--dcs', '-d', help='Use this DCS', envvar='DCS')
+@option_insecure
 @click.pass_context
-def ctl(ctx, config_file, dcs):
+def ctl(ctx, config_file, dcs, insecure):
     logging.basicConfig(format='%(asctime)s - %(levelname)s - %(message)s', level=os.environ.get('LOGLEVEL', 'WARNING'))
+    logging.captureWarnings(True)  # Capture eventual SSL warning
     ctx.obj = load_config(config_file, dcs)
+    # backward compatibility for configuration file where ctl section is not define
+    ctx.obj.setdefault('ctl', {})['insecure'] = ctx.obj.get('ctl', {}).get('insecure') or insecure
 
 
 def get_dcs(config, scope):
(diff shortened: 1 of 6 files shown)


patroni/ctl.py
```diff
@@ -128,17 +130,28 @@ def auth_header(config):
...
     headers = headers or {}
     url_parts = urlparse(member.api_url)
     logging.debug(url_parts)
     if 'Content-Type' not in headers:
         headers['Content-Type'] = 'application/json'
 
     url = '{0}://{1}/{2}'.format(url_parts.scheme, url_parts.netloc, endpoint)
-
+    # Retrieve certfile from configfile in restapi section if any
+    cert = (click.get_current_context().obj.get('restapi', {}).get('certfile'),
```
CyberDem0n: Doc http://docs.python-requests.org/en/master/user/advanced/#client-side-certificates tells that `cert` is used for client-side certificates, but here you are presumably trying to use server-side.

Alright, let's assume that this is a special patronictl config. Even in this case there is a problem: `requests` module expect to get either a `string` (filename containing cert+key) or a tuple. Not sure how it will work if it will get for example `('cert', None)`, because apparently [keyfile](https://patroni.readthedocs.io/en/latest/SETTINGS.html?highlight=keyfile#rest-api) is optional: https://docs.python.org/2/library/ssl.html#ssl.SSLContext.load_cert_chain

wilfriedroset:
I was first going for the configuration file part because I think that we will have to use `-k` a lot. But by doing so we would need to duplicate the restapi certfile/keyfile part to the ctl part or use some kind of interpolation.

patroni/ctl.py
wilfriedroset: Indeed, I will remove this part.

CyberDem0n:
> I was first going for the configuration file part because I think that we will have to use -k a lot.

If you are using self-signed certificates, especially if every node has different certificates, you'll definitely have to use `-k` a lot.
Taking into account that you can run `patronictl` on any host, including your computer, I think it make sense to have a special section for patronictl in the config.

Think about following use-case: certificates on the patroni nodes are signed by the same certificate authority. Patronictl can use `cacert` to verify node certificates. Sure, such cacert can go into `restapi`.
Another use-case would be: we want to use client certificates to authenticate patronictl. It definitely requires a special section (`ctl`) in the config.

> we would need to duplicate the restapi certfile/keyfile part to the ctl part or use some kind of interpolation.

It is very easy to avoid duplication. If `ctl.cacert` is not there, try to use `restapi.cacert` or `restapi.certfile`.

If the `-k` is specified, it should just change the "virtual" config: `ctx.obj['ctl']['insecure'] = True` or `ctx.obj['ctl']['verify'] = False`
You can do it just in one place, in the `def ctl(ctx, config_file, dcs):` function. And later use the result in the `request_patroni`. It will simplify the code, because you don't need to pass around all `insecure` arguments.

wilfriedroset:
What do you think about this version?
`ctx.obj` is updated during `def ctl`.
I've added a graceful support for configuration file which does not contain the `ctl` section.
The context is retrieve using `click.get_current_context()` and we avoid double definition for certfile/cacert using your recommendation.

About travis build I don't understand the failure.
This build pass: https://travis-ci.org/zalando/patroni/builds/421115034?utm_source=github_status&utm_medium=notification
This one doesn't: https://travis-ci.org/zalando/patroni/builds/421125607?utm_source=github_status&utm_medium=notification

The difference between the two is that I've updated the documentation.

CyberDem0n:
I'll look on it a little bit later.

Acceptance tests sometimes are not stable due to the slowness of travis :(
I've re-triggered it, usually it helps.

CyberDem0n:
Merged, thank you!

Issue #769, opened 2018-08-08
Don't show "nofailover" nodes for switchover


dtseiler:
Currently, `patronictl switchover` will list a `nofailover` host as an option to switch to, but then fail if you try to do it saying that there are no candidates.

I think `nofailover` nodes should either be hidden from the candidate list during switchover/failover or let you fail over to them (perhaps with a warning/confirmation). Doesn't make sense to let the user choose them and then fail because they're tagged as `nofailover`.

PR #798, opened 2018-08-29
Don't show "nofailover" nodes for switchover

anikin-aa:
As described here
https://github.com/zalando/patroni/issues/769
exclude members with nofailover tag from `patronictl switchover/failover` output.


CyberDem0n:
Merged, thank you

PR #799, opened 2018-08-30
Support for Windows

pashagolub:
Changes for Windows support:

- On Windows, signal() can only be called with SIGABRT, SIGFPE, SIGILL, SIGINT, SIGSEGV, SIGTERM, or SIGBREAK. A ValueError will be raised in any other case.

- fcntl module from the standard library available only on Linux


patroni/__init__.py
```diff
@@ -1,4 +1,5 @@
 import logging
+import platform
```
CyberDem0n: unused import

patroni/api.py
```diff
@@ -1,11 +1,11 @@
 import base64
-import fcntl
 import json
 import logging
 import psycopg2
 import time
 import dateutil.parser
 import datetime
+import platform
```
```python
-import platform
+import os
```

patroni/__init__.py
pashagolub: Ahh, thanks. My bad!

patroni/postmaster.py
```diff
@@ -154,9 +155,19 @@ def start(pgcommand, data_dir, conf, options):
...
+        # The child process could deadlock before exec is called. If you must use it, keep it trivial! 
+        # Minimize the number of libraries you call into.
+        # If you need to modify the environment for the child use the env parameter rather than doing it in a preexec_fn. 
+        # The start_new_session parameter can take the place of a previously common use of preexec_fn to call 
+        # os.setsid() in the child. 
         proc = call_self(['pg_ctl_start', pgcommand, '-D', data_dir,
-                          '--config-file={}'.format(conf)] + options, close_fds=True,
-                         preexec_fn=os.setsid, stdout=subprocess.PIPE, env=env)
+                          '--config-file={}'.format(conf)] + options,
+                          close_fds=True if os.name != 'nt' else False,
```
CyberDem0n: `close_fds=(os.name != 'nt')`

patroni/postmaster.py
```diff
@@ -154,9 +155,19 @@ def start(pgcommand, data_dir, conf, options):
...
+        # Minimize the number of libraries you call into.
+        # If you need to modify the environment for the child use the env parameter rather than doing it in a preexec_fn. 
+        # The start_new_session parameter can take the place of a previously common use of preexec_fn to call 
+        # os.setsid() in the child. 
         proc = call_self(['pg_ctl_start', pgcommand, '-D', data_dir,
-                          '--config-file={}'.format(conf)] + options, close_fds=True,
-                         preexec_fn=os.setsid, stdout=subprocess.PIPE, env=env)
+                          '--config-file={}'.format(conf)] + options,
+                          close_fds=True if os.name != 'nt' else False,         
+                           _fn=os.setsid if os.name != 'nt' else None,
```
CyberDem0n: `preexec_fn=os.setsid` was here from the `pg_ctl` times, now we can move it to https://github.com/zalando/patroni/blob/master/patroni/__init__.py#L155-L158

```python
def pg_ctl_start(args):
    import subprocess
    if os.name != 'nt':
        os.setsid()
    postmaster = subprocess.Popen(args)
    print(postmaster.pid)
```

CyberDem0n: And remove copy&paste from docs.python.org ;)

patroni/dcs/etcd.py
```diff
@@ -448,7 +448,7 @@ def member(node):
     def _load_cluster(self):
         try:
             result = self.retry(self._client.read, self.client_path(''), recursive=True)
-            nodes = {os.path.relpath(node.key, result.key): node for node in result.leaves}
+            nodes = {os.path.relpath(node.key, result.key).replace('\\', '/'): node for node in result.leaves}
```
CyberDem0n: There is a similar problem with `patroni/dcs/consul.py`

patroni/postmaster.py
```diff
@@ -138,7 +138,8 @@ def start(pgcommand, data_dir, conf, options):
         # of init process to take care about postmaster.
         # In order to make everything portable we can't use fork&exec approach here, so  we will call
         # ourselves and pass list of arguments which must be used to start postgres.
-        env = {p: os.environ[p] for p in ('PATH', 'LD_LIBRARY_PATH', 'LC_ALL', 'LANG') if p in os.environ}
+        # On Windows, in order to run a side-by-side assembly the specified env must include a valid SYSTEMROOT.
+        env = {p: os.environ[p] for p in ('PATH', 'LD_LIBRARY_PATH', 'LC_ALL', 'LANG', 'SYSTEMROOT') if p in os.environ}
```
CyberDem0n: Shouldn't `SYSTEMROOT` also be added to https://github.com/zalando/patroni/blob/master/patroni/postgresql.py#L1225 ?

Please also make sure that you provide the correct env that `pg_controldata` speaks English.

pashagolub: `LC_ALL` seems enough for speaking English

patroni/postgresql.py
```diff
@@ -1224,8 +1224,8 @@ def controldata(self):
         if self._version_file_exists() and self.state != 'creating replica':
             try:
                 data = subprocess.check_output([self._pgcommand('pg_controldata'), self._data_dir],
-                                               env={'LANG': 'C', 'LC_ALL': 'C', 'PATH': os.environ['PATH'],
-                                               'SYSTEMROOT': os.environ['SYSTEMROOT']})
+                                               env={'LANG': 'C', 'LC_ALL': 'C', 'PATH': os.environ.get('PATH'),
+                                               'SYSTEMROOT': os.environ.get('SYSTEMROOT')})
```
CyberDem0n: The value of the environment variable must be a string, but it this case the value of `SYSTEMROOT` would be `None`, therefore it fails on travis.

CyberDem0n:
Merged, thank you!

Issue #656, opened 2018-04-05
Allow 'permanent' replication slots that don't correspond to the cluster members


alexeyklyukin:
When Patroni syncs its replication slots, it removes all slots except for those corresponding to the known members of the cluster.

That may break some useful features that rely on replication slots (i.e. logical decoding via a replication slot, logical replication or even pg_basebackup from the replica in newer versions of PostgreSQL). 

There are at least two ways to resolve this:

- Define some slots in the patroni configuration (in DCS) to be permanent, Patroni will create them and never remove. The downside of this approach is that it requires the user to let Patroni know of such slots explicitly.
- Avoid deleting replication slots that were not created by Patroni, leaving the task of creating actual slots to the user. This is suboptimal since replication slots itself are not replicated and would have to be created anew after each promotion.

Perhaps it would be good to combine both approaches and tag those slots created by Patroni (including permanent one, defined in the configuration) in some way or another, and drop only those slots when necessary. Since there are no comments on replication slots, the obvious way is to use some naming convention with a possible risk of the user naming their slots the same way and confusing Patroni.

bradnicholson:
> Avoid deleting replication slots that were not created by Patroni, leaving the task of creating actual slots to the user. This is suboptimal since replication slots itself are not replicated and would have to be created anew after each promotion.

This seems like a cleaner approach to me though.  This is a limitation in Postgres at the time, and trying to have Patroni work around that seems less than optimal.

When the replication slot gets recreated on another member, it will have lost the place the former slot was on, and the end user will possibly/probably need to take special action to resync anyway.

bradnicholson:
What is the value of persisting a replication slot on failover if we can't pick up from where we left off?

It seems like there is a burden on the application here anyway to handle that case.  Is re-creatin the replication slot as part of that too much burden?

jberkus:
@bradnicholson the purpose of this is to persist replication slots that aren't created by, or maintained by, Patroni.

For example, imagine that you had a replication slot for a logical replication stream that was attached to an external data auditing tool.  Whenever the list of nodes changes, Patroni will delete all slots not associated with an active Patroni node ... including the one for the auditing tool.

CyberDem0n:
Unfortunately there are too many ambiguities in the second approach.
It is not possible to know from the slot name whether it was created by Patroni or not.
[...]

PR #819, opened 2018-10-04
Permanent replication slots

CyberDem0n:
Permanent replication slots are preserved on failover/switchover, that is Patroni on the new primary will create configured replication slots right after doing promote.

Slots could be configured with the help of `patronictl edit-config`.
The initial configuration could be also done in the `bootstrap.dcs`

```yaml
slots:
  permanent_physical_1:
    type: physical
  permanent_logical_1:
    type: logical
    database: foo
    plugin: pgoutput
```

It is the responsibility of the operator to make sure that there are no clashes in names between replication slots automatically created by Patroni for members and permanent replication slots.

Closes https://github.com/zalando/patroni/issues/656

### DIFF
--- a/patroni/dcs/__init__.py
+++ b/patroni/dcs/__init__.py
@@ -469,7 +469,7 @@ class Cluster(namedtuple('Cluster', 'initialize,config,leader,last_leader_operat
 
             if isinstance(value, dict):
                 if 'type' not in value:
-                    value['type'] = 'logical'
+                    value['type'] = 'logical' if value.get('database') and value.get('plugin') else 'physical'
 
                 if value['type'] == 'physical' or value['type'] == 'logical' \
                         and value.get('database') and value.get('plugin'):
@@ -481,10 +481,8 @@ class Cluster(namedtuple('Cluster', 'initialize,config,leader,last_leader_operat
         return slots
 
     def has_permanent_logical_slots(self, name):
-        for slot in self.get_replication_slots(name, 'master').values():
-            if slot['type'] == 'logical':
-                return True
-        return False
+        slots = self.get_replication_slots(name, 'master').values()
+        return any(v for v in slots if v.get("type") == "logical")
 
 
 @six.add_metaclass(abc.ABCMeta)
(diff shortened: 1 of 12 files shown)


patroni/postgresql.py
```diff
@@ -1563,50 +1555,70 @@ def postmaster_start_time(self):
...
             return None
 
+    def drop_replication_slot(self, name):
+        cursor = self._query(('SELECT pg_drop_replication_slot(%s) WHERE EXISTS (SELECT 1 ' +
+                              'FROM pg_replication_slots WHERE slot_name = %s AND NOT active)'), name, name)
+        # In normal situation rowcount should be 1, otherwise either slot doesn't exists or it is still active
+        return cursor.rowcount == 1
+
+    @staticmethod
+    def compare_slots(s1, s2):
```
erthalion: Judging from `load_replication_slots` we can have only:
```python
{ slot1: { "type": "physical" } } 
```
or
```python
{ slot2: { "type": "physical", "plugin": "plugin", "database": db } }
```

Then doesn't it mean that we can just use `s1 == s2` here?

patroni/postgresql.py
```diff
@@ -1563,50 +1555,70 @@ def postmaster_start_time(self):
...
+                                             " WHERE slot_type = 'physical' AND slot_name = %s)").format(
+                                                 immediately_reserve), name, name)
+                            except Exception:
+                                logger.exception("Failed to create physical replication slot '%s'", name)
+                                self._schedule_load_slots = True
+                        elif value['type'] == 'logical' and name not in self._replication_slots:
+                            logical_slots[value['database']][name] = value
+
+                # create new logical slots
+                for database, values in logical_slots.items():
```
erthalion: What prevents us from doing the same logic inside the previous loop and avoid having `logical_slots`?

patroni/dcs/__init__.py
```diff
@@ -397,6 +425,67 @@ def is_paused(self):
...
+                        and value.get('database') and value.get('plugin'):
+                    slots[name] = value
+                    continue
+
+            logger.error("Bad value for slot '%s' in permanent_slots: %s", name, permanent_slots[name])
+
+        return slots
+
+    def has_permanent_logical_slots(self, name):
+        for slot in self.get_replication_slots(name, 'master').values():
```
erthalion: Minor complain, but this is shorter:

```python
slots = self.get_replication_slots(name, 'master').values()
return any(v for v in slots if v.get("type") == "logical")
```

Issue #625, opened 2018-02-13
Feature request: do not delete PGDATA


uspen:
Hello!

It would be nice to have an option for create_replica_metod like "do not delete PGDATA". We use pgbackrest with the option --delta and it's very fast.

alexeyklyukin:
Good idea. PRs are welcome :-)

uspen:
Sory, i dont know python

bradnicholson:
I'm have need of this and am willing to tackle it, but it's not clear to me the best way to do so.  Looking for some pointers if you have any.

I was thinking of adding a `keep_data` parameter to custom replica creation method like this:

```
postgresql:
    create_replica_method:
        - pgBackRest
        - basebackup
    pgBackReset:
        command: pgBackRest_restore
        keep_data: True
```

But the issue I'm seeing the code to delete the data dir is being called prior to any of the replica creation logic here:
https://github.com/zalando/patroni/blob/master/patroni/ha.py#L886

Not to mention the ha object has no knowledge of the Postgres section of the config.

Any ideas?

sharmay:
New to patroni and my first set of changes are available in https://github.com/sharmay/patroni

```
postgresql:
    create_replica_method:
        - pgBackRest
        - basebackup
    pgBackReset:
        command: /usr/bin/pgbacktest --stanza=mydb --delta restore
        keep_data: True
        no_params: True
```

PR will be created after further testing. 

Please let me know your comment.

Give it a try @bradnicholson @uspen

CyberDem0n:
@sharmay there is no need to reinvent the weel.

Just wrap `pgBackRest` into a simple shell script, which will skip all unknown arguments.

Not deleting the PGDATA is a different story, and deserves working on it.

sharmay:
@CyberDem0n will wrap it in a script.

Just for my knowledge what are the benefits of wrapping in a script ?

CyberDem0n:
> what are the benefits of wrapping in a script?

The biggest benefit is: it works without any changes in Patroni :)
If you really want it in the Patroni code (skipping unwanted arguments), please go ahead and implement it.

sharmay:
How about this approach ?
```
postgresql:
    create_replica_method:
        - pgBackRest
        - basebackup
    pgBackReset:
        command: /usr/bin/pgbacktest
        args:
            - "--stanza=mydb"
            - "--delta"
            - "restore"
        keep_data: True
        no_params: True
```
- Add Args section to handle command arguments.
[...]

PR #822, opened 2018-10-06
Add pgbackrest support

sharmay:
Resubmit of PR https://github.com/zalando/patroni/pull/788

pgBackrest can restore in existing $PGDATA folder, this allows speedy
restore as files which have not changed since last backup are skipped,
to support this feature new param keep_data has been introduced. When
keep_data=True, cleanup of $PGDATA will be skipped.

Patroni passes some extra parameters to custom_replica_methods when
calling, this causes an error due to pgbackrest strict parameter
checking. New param no_params=True can be set to skip parameters
passing.

Updated documentation for same.

Fixes https://github.com/zalando/patroni/issues/625


CyberDem0n:
Merged, thank you @sharmay!

PR #835, opened 2018-10-22
Add pre-command to start watchdog

SrikanthDBAdmin:
Add `ExecStartPre` commands to help start watchdog devices.
Commands are commented to make it inline with the YAML: https://github.com/zalando/patroni/blob/master/postgres0.yml#L90-#L93


CyberDem0n:
Merged, thank you!

combrs:
using sudo requires preliminary sudoers configuration. Instead, systemd allow to use "PermissionsStartOnly=true"  for run as root ExecStartPre commands, but ExecStart anyway run as User.

PR #844, opened 2018-10-30
Add pgbouncer + pgbackrest shel script

slardiere:
This script is able to restore data from pgbackrest, for creating
    a new standby


uspen:
It won't work. Because Patroni rewrite recovery.conf. If you get WAL from archive, replica will have very big replication lag. Its worst. If archive not used from replica, then replica will not work

slardiere:
Of course, you need to add `restore_command` into `recovery.conf` : 

~~~
bootstrap:
  dcs:
    postgresql:
      recovery_conf:
        restore_command: pgbackrest --stanza=ats archive-get %f "%p"
~~~

I can provide a full `config.yml`, but i don't know where to put it in the source tree.

extras/confd/templates/pgbouncer.tmpl
```diff
@@ -0,0 +1,17 @@
+[databases]
+{{range gets "/members/*"}}{{$data := json .Value}}{{if eq $data.role "master" }}{{ $hostport := base (replace (index (split $data.conn_url "/") 2) "@" "/" -1)}}{{ $host := base (index (split $hostport ":") 0)}}{{ $port := base (index (split $hostport ":") 1)}}employees = host={{ $host }} port={{ $port }} dbname=employees pool_size=10{{end}}{{end}}
```
CyberDem0n: We should not rely on the role written in the member json.
In order to build the configuration we should do the following:
1. get /leader
2. get /member/$leader

Also I don't like hard-coded database name, "employees", instead we could use pgbouncer [autodb](https://pgbouncer.github.io/config.html#section-databases) feature.
Basically we should specify something like: `* = host={{ $host }} port={{  $port }}`

docs/replica_bootstrap.rst
```diff
@@ -101,7 +101,7 @@ example: pgbackrest
             - pgbackrest
             - basebackup
         pgbackrest:
-            command: /usr/bin/pgbackrest --stanza=mydb --deltarestore
+            command: pgbackrest.sh
```
CyberDem0n: I don't think it is necessary, https://github.com/zalando/patroni/pull/822 adds nice support of pgbackrest.

@sharmay could you please share your opinion on that?

slardiere: Indeed, I've to try this feature. At least, there is a space missing betweer `--delta` and `restore`

extras/confd/templates/pgbouncer.tmpl
slardiere: Ok, I'll correct in this way. The database name is a mistake, indeed, the sample tmpl should use autodb

CyberDem0n:
Merged, thank you!

slardiere:
Thank you !

Issue #572, opened 2017-12-08
get Patroni working on Openshift


jberkus:
@ants 

There are a number of obstacles for Patroni/Spilo working on OpenShift, most particularly file permissions issues and errors associated with the random high userid.  While I have issues open with the openshift team about this, I'd like to also document and/or provide some alternate images and config settings which let patroni run on OpenShift.

Here's what I have so far:

* directory permissions: modify group permissions per the canonical OpenShift postgres image, using their fixpermissions() approach.
* userid calls from Postgres: use nss_wrapper and /tmp/passwd
* python userid calls: ???

The last is what I'm stuck on; python doesn't care about nss_wrapper, and at least one library we're using does a call to getuid().

ibotty:
> userid calls from Postgres

Or `chmod a+w /etc/passwd` as is done in most s2i containers.

davecramer:
@jberkus is this still the status of patroni on openshift ?

jberkus:
it's not, but we haven't put together the templates and the RBAC profile into a coherent package.  Sorry, it's on my TODO list, but I've been travelling a lot for work.

ttiurani:
I'm interested in this as well, so would be happy to hear about and test any configs you have @jberkus

soukron:
Hi @jberkus 
Did you manage to get it working in OpenShift without using a cluster-admin privileged account?

anikin-aa:
Guys, hi ! 
we managed to get it working under restricted SCC. 
> Or chmod a+w /etc/passwd as is done in most s2i containers.
this did the job for our case.

PR #871, opened 2018-11-17
add openshift customizations, templates, and test

stewartshea:
This PR adds an OpenShift working example; 
- It modifies the Dockerfile and entrypoint slightly to allow for OpenShift SCCs to operate correctly
- It adds 2 template examples that can be easily modified by changing parameters

This example is tied to issue #572


openshift/entrypoint.sh
```diff
@@ -0,0 +1,46 @@
...
+restapi:
+  connect_address: '${POD_IP}:8008'
+postgresql:
+  connect_address: '${POD_IP}:5432'
+  authentication:
+    superuser:
+      password: '${PATRONI_SUPERUSER_PASSWORD}'
+    replication:
+      password: '${PATRONI_REPLICATION_PASSWORD}'
+  callbacks:
```
CyberDem0n: We need this callback script only for the case when we run Patroni on kubernetes with Etcd, ZooKeeper or Consul.
The purpose of it - assign the label with the current role on the Pod and update endpoints of the leader (master) service by putting the ip of the leader (master) pod there.

Patroni already takes care about changing pod labels, and, as far as I remember on openshift we can't put ips from the pod network into the endpoint by security reason. Basically all above-mentioned facts make such callback script obsolete.
Originally I ended up on having this script because pod labels update was implemented at the last step and I needed something to test and play with.

openshift/Dockerfile
```diff
@@ -0,0 +1,39 @@
+FROM postgres:10
```
CyberDem0n: It looks like there is no difference between this file in the `kubernetes` directory and here.
Can we just make a symlink?

stewartshea: This file is modified at the bottom to set the permissions appropriately for OpenShift assigned UID/GID to work. 

The updated content is: 
```
# Set directory permissions for OpenShift
RUN chmod a+w /etc/passwd \
    && chgrp -R 0 /home/postgres \
    && chmod -R g+rwX /home/postgres
```

openshift/entrypoint.sh
stewartshea: This was just copied from the kubernetes directory so I can't comment on your callbacks. The important part of the entrypoint.sh modification is on the /etc/passwd file:

```
export USER_ID=$(id -u)
export GROUP_ID=$(id -g)
if [ `id -u` -ge 10000 ]; then
    cat /etc/passwd | sed -e "s/^${USER_ID}:/builder:/" > /tmp/passwd
    echo "postgres:x:${USER_ID}:${GROUP_ID}:PostgreSQL Server:${HOME}:/bin/bash" >> /tmp/passwd
    cat /tmp/passwd > /etc/passwd
    rm /tmp/passwd
fi
```

stewartshea: Regarding your other question, the UID is unique to the project, not just the single pod iteration/run. So each pod in the project has the same UID. In combination with the file system permissions set in the Dockerfile, the files will be accessible within the same project. I haven't tested moving the PV to a separate project, but I'm not sure I see that as a regular use case since that would likely require backup/restore.

stewartshea:
@CyberDem0n let me know if the explanations are sufficient or if there are other clarifications or changes you require. Thanks so much for your time and review!

CyberDem0n:
I would actually like to merge openshift and kubernetes into the same directory. It would be easier to support them in future.
Basically Dockerfile should look like:
```docker
FROM postgres:10
MAINTAINER Alexander Kukushkin <alexander.kukushkin@zalando.de>

RUN export DEBIAN_FRONTEND=noninteractive \
    && echo 'APT::Install-Recommends "0";\nAPT::Install-Suggests "0";' > /etc/apt/apt.conf.d/01norecommend \
    && apt-get update -y \
    && apt-get upgrade -y \
    && apt-cache depends patroni | sed -n -e 's/.* Depends: \(python3-.\+\)$/\1/p' \
            | grep -Ev '^python3-(sphinx|etcd|consul|kazoo|kubernetes)' \
            | xargs apt-get install -y curl jq locales git python3-pip python3-wheel \

    ## Make sure we have a en_US.UTF-8 locale available
    && localedef -i en_US -c -f UTF-8 -A /usr/share/locale/locale.alias en_US.UTF-8 \

    && pip3 install setuptools \
    && pip3 install 'git+https://github.com/zalando/patroni.git#egg=patroni[kubernetes]' \

    && PGHOME=/home/postgres \
    && mkdir -p $PGHOME \
    && chown postgres $PGHOME \
    && sed -i "s|/var/lib/postgresql.*|$PGHOME:/bin/bash|" /etc/passwd \

    # Set permissions for OpenShift
    && chmod 775 $PGHOME \
    && chmod 664 /etc/passwd \

    # Clean up
    && apt-get remove -y git python3-pip python3-wheel \
    && apt-get autoremove -y \
    && apt-get clean -y \
    && rm -rf /var/lib/apt/lists/* /root/.cache

ADD entrypoint.sh /

EXPOSE 5432 8008
ENV LC_ALL=en_US.UTF-8 LANG=en_US.UTF-8
USER postgres
WORKDIR /home/postgres
CMD ["/bin/bash", "/entrypoint.sh"]
```

and entrypoint.sh like:
```bash
#!/bin/bash

if [[ $UID -ge 10000 ]]; then
    GID=$(id -g)
    sed -e "s/^postgres:x:[^:]*:[^:]*:/postgres:x:$UID:$GID:/" /etc/passwd > /tmp/passwd
    cat /tmp/passwd > /etc/passwd
    rm /tmp/passwd
fi

cat > /home/postgres/patroni.yml <<__EOF__
bootstrap:
  dcs:
    postgresql:
      use_pg_rewind: true
  initdb:
  - auth-host: md5
  - auth-local: trust
  - encoding: UTF8
  - locale: en_US.UTF-8
  - data-checksums
  pg_hba:
  - host all all 0.0.0.0/0 md5
  - host replication ${PATRONI_REPLICATION_USERNAME} ${POD_IP}/16    md5
restapi:
  connect_address: '${POD_IP}:8008'
postgresql:
  connect_address: '${POD_IP}:5432'
  authentication:
    superuser:
      password: '${PATRONI_SUPERUSER_PASSWORD}'
    replication:
      password: '${PATRONI_REPLICATION_PASSWORD}'
__EOF__

unset PATRONI_SUPERUSER_PASSWORD PATRONI_REPLICATION_PASSWORD
export KUBERNETES_NAMESPACE=$PATRONI_KUBERNETES_NAMESPACE
export POD_NAME=$PATRONI_NAME

exec /usr/bin/python3 /usr/local/bin/patroni /home/postgres/patroni.yml
```

callback.py could be dropped, we don't need it.

CyberDem0n:
Merged, thank you!

PR #884, opened 2018-11-30
Refactor acceptance tests to improve stability

CyberDem0n:
Hope it will crash less often when executed on travis against k8s


zerg-junior:
:+1:

PR #885, opened 2018-11-30
Fix kubernetes demo files

kostiantyn-nemchenko:
- Update postgres docker image to the latest 11 version.

- Remove empty lines inside the `RUN` command to make the Dockerfile compatible with future docker versions.

- Set the `PATRONI_KUBERNETES_POD_IP` environment variable, which is required when _use_endpoints_ is enabled. Otherwise, the `KeyError` is raised [here](https://github.com/zalando/patroni/blob/e684ca66e5d66a10feef52e3513813fbcaab9581/patroni/dcs/kubernetes.py#L95).

- Set `EDITOR` environment variable to make configuration changes via `patronictl edit-config`.


CyberDem0n:
Merged, thank you!

Issue #890, opened 2018-12-04
Switchover/Failover may not switch timelines.


bradnicholson:
Take the following scenario:

1. A is the leader, B is the follower.   A and B are on Timeline 1 
1. A shuts down. Either a controlled switchover or failover happens and B is promoted. A is still on Timeline 1, B moves to Timeline 2
1. B shuts down, without A ever coming back as a subscriber to B
1. A starts up and becomes leader.  A switches from Timeline 1 to Timeline 2 independent of 

At this point, we have split-brain from a Postgres perspective.  Both have independently promoted themselves from Timeline 1 to Timeline 2 with no knowledge of each other.

The reason we stumbled across this was due to archiving conflicts with pgBackRest (see issue [584](https://github.com/pgbackrest/pgbackrest/issues/584) ).  There is separate underlying issue there where the .partial files being generated on members A & B differ from the archived version (due, I think, to the immediate shutdown Patroni issues - but not conclusive yet.  The data in the wal files is identical, but the sha sums are not).  pgBackRest will catch this because it checksums the files.  wal-e will not.

The pgBackRest folks have some specific recommendations for Patroni [here](https://github.com/pgbackrest/pgbackrest/issues/584#issuecomment-443260844) 
 
>In addition, there really should never be two .partial files on the same timeline. The old primary should be made to follow the new primary and do the timeline switch before it gets promoted again. In that case it would promote to timeline 3 and there's no conflict. What we are seeing here is split-brain and it is not good. Patroni should be tracking the timeline of the primary and refusing to promote another node that has not made it to that timeline yet. If it does then some mitigation may be needed to get things working again.

This seems to be at odds with the goal ensuring a single primary can be running in the event of a network partition though... 🤔

CyberDem0n:
You are right, at the moment there is no protection from such scenario :(
Node A decides to promote because its absolute wal position is within maximum_lag_on_failover and it doesn't take into account that some other node managed to generate a new timeline.
It wouldn't be too hard to implement such check, DCS already has everything we need in the `/history` key.

However, there is one thing I would like you to check for me, did node B managed to archive the history file?

bradnicholson:
> However, there is one thing I would like you to check for me, did node B managed to archive the history file?

Yes, B archives the history file on

CyberDem0n:
Ok, will prepare a PR for that tomorrow.

bradnicholson:
[...]

PR #892, opened 2018-12-05
Consider node not healthy if it is not on the latest timeline

CyberDem0n:
The latest timeline is calculated from the `/history` key in DCS. In case there is no such key or it contains some garbage we consider the node healthy.

Hopefully fixes https://github.com/zalando/patroni/issues/890

### DIFF
--- a/docs/dynamic_configuration.rst
+++ b/docs/dynamic_configuration.rst
@@ -76,6 +76,7 @@ Also, the following Patroni configuration options can be changed only dynamicall
 - loop_wait: 10
 - retry_timeouts: 10
 - maximum_lag_on_failover: 1048576
+- check_timeline: false
 - postgresql.use_slots: true
 
 Upon changing these options, Patroni will read the relevant section of the configuration stored in DCS and change its
--- a/docs/replication_modes.rst
+++ b/docs/replication_modes.rst
@@ -13,6 +13,8 @@ In asynchronous mode the cluster is allowed to lose some committed transactions 
 
 The amount of transactions that can be lost is controlled via ``maximum_lag_on_failover`` parameter. Because the primary transaction log position is not sampled in real time, in reality the amount of lost data on failover is worst case bounded by  ``maximum_lag_on_failover`` bytes of transaction log plus the amount that is written in the last ``ttl`` seconds (``loop_wait``/2 seconds in the average case). However typical steady state replication delay is well under a second.
 
+By default, when running leader elections, Patroni does not take into account the current timeline of replicas, what in some cases could be undesirable behavior. You can prevent the node not having the same timeline as a former master become the new leader by changing the value of ``check_timeline`` parameter to ``true``.
+
 PostgreSQL synchronous replication
 ----------------------------------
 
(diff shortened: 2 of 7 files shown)


bradnicholson:
I took a look and tested it out.  It correctly detects the timeline divergence here:
https://github.com/zalando/patroni/blob/76f4bc286d898e5f810e4747df81f35e362ca2c0/patroni/ha.py#L560

It prevents failing over, but it doesn't log correctly.  It logs:

`INFO: following a different leader because i am not the healthiest node`

There is no other leader running at this time.

CyberDem0n:
You are right, the [message](https://github.com/zalando/patroni/blob/76f4bc286d898e5f810e4747df81f35e362ca2c0/patroni/ha.py#L852) is sometimes misleading, and it logs the same for all cases when Patroni decided that this node isn't healthy enough to become the new primary.
The problem is that we are in a dynamic environment, not having other nodes at the moment doesn't mean that some of them won't become available in a couple of seconds.

I can add more verbose logs in the place where the node tries to determine whether it is the healthiest or not, something similar to what we have here: https://github.com/zalando/patroni/blob/76f4bc286d898e5f810e4747df81f35e362ca2c0/patroni/ha.py#L576-L594

bradnicholson:
I think that would be helpful.  I think the behavior is fine (including the message it is logging), but knowing that it is also failing because of the timeline issue in the case when the other member doesn't come back would be useful.

bradnicholson:
Thanks.  This looks good to me.

jberkus:
@CyberDem0n I'd like the ability to turn off this functionality via a configure switch.

In cases where all nodes on the latest timeline are *gone*, I may want the ability to start up from an older surviving node if I'm willing to live with the data loss.  In fact, I know at least two production clusters where I *do* want that.

jberkus:
... and for that matter, considering the number of ways that this can fail and wedge the cluster, I'd argue that timeline switch detection should never be on by default.

CyberDem0n:
I tend to agree with @jberkus, it should configurable and not be enabled by default.
@jberkus, @bradnicholson, @ants, @sfrost, @dwsteele, what would be the good name for such switch?

bradnicholson:
Favoring availability over data loss is the correct tradeoff with Patroni (with that being controllable via amounts of data loss you will tolerate, or just using sync rep if you want none).  

That said, if that failover can break the consistency of the archive, that seems a bit more nuanced to me.  pgBackRest is catching these issues via checksumming, but other backup tools may not.  

Whether we act or not (or how) on this case is an open question, but silently breaking the archive seems like a really bad behavior.

dwsteele:
Agreed, but I think appending the target timeline to the .partial files and prioritizing .history files will take care of most of the issues that we've seen.

I'm pretty confident that both patches will make their way into Postgres.  If they don't get back-patched then we'll "backpatch" them in pgBackRest.

I don't have any good solution for the situation where two clusters end up on the same timeline.  Prioritizing .history files will help, but it won't eliminate the problem entirely.  The patch under review will also help when it is enabled, but it is not perfect, either.

bradnicholson:
Seems prudent to wait on this until we see what happens with PG.

If they get back patched, the right approach seems to be do nothing with Patroni, except perhaps a mention in the docs that for archiving to be reliable, one must run versions >= x,y,z for PG.

CyberDem0n:
@bradnicholson in the last [commit](https://github.com/zalando/patroni/pull/892/commits/b50d7af5148bb5c8ba23597a49232c39e12146d8) I introduced the new configuration parameter, `check_timeline`, which is set to `false` by default.

I am planning to merge this PR and maybe wrap the new release. What do you think?

PR #893, opened 2018-12-06
Consul service deregister timeout must be in Go time format

Isonami:
According to consul documentation:

> In Consul 0.7 and later, checks that are associated with a service may also contain an optional deregister_critical_service_after field, which is a timeout in the same Go time format as interval and ttl.


CyberDem0n:
Merged, thank you!

PR #896, opened 2018-12-14
Allow postgres stderr to output use the same file handles

snooc:
Starting in Patroni 1.5.0 the "pg_ctl_start" process started to pipe stderr to stdout. In previous versions,  the "pg_ctl_start" process and Postgres both inherited stderr from the main Patroni process

Inheriting stderr from the main Patroni process allows all Postgres logs to be seen along with all patroni logs. This is very useful in a container environment as Patroni and Postgres logs may be consumed using standard tools (docker logs, kubectl, etc).

This PR removes the pipe from stderr to stdout.


CyberDem0n:
Merged, thank you!

Issue #902, opened 2018-12-20
Feature Proposal: env variable to write Patroni logs to a file.


capistrant:
What I propose is an environment variable, perhaps `PATRONI_LOG_DIR` that supplies a path to an existing directory that is writeable by the executing user. When this environment variable is set, Patroni will create two [RotatingFileHandler](https://docs.python.org/3/library/logging.handlers.html?highlight=rotatingfilehandler#logging.handlers.RotatingFileHandler) objects. One for each of root and requests loggers. These handlers would leverage the existing environment variables that control log level and message format.

This would be a non-breaking change since existing users will see everything as business as usual as long as the new environment variable is not set.

Another enhancement that I think would be nice is adding the logger name to the default log line format to provide more context. This would however break any existing users who rely on the default format for things like log parsing, etc.

I look forward to getting some feedback! I can create a PR for this if the community thinks it would be a beneficial feature. I know that my team would certainly like it.

CyberDem0n:
> One for each of root and requests loggers. 

I believe everything should go into the same file. The only reason I set requests logs level to a WARNING by default, is that it is logging too much on INFO.

You are right, the logging system is far from benig ideal, any improvements are welcome.

PR #903, opened 2018-12-21
New Feature: Add ability to configure app logs to be written to a file

capistrant:
Closes https://github.com/zalando/patroni/issues/902

This PR gives users the option to send Patroni application logs to a File instead of Standard Out. There are three environment variables that can be set to enable and configure file logging.
1. `PATRONI_FILE_LOG_DIR`: Path to a directory that is writeable by the executing user. Having this variable set is what activates file logging.
2. `PATRONI_FILE_LOG_NUM`: This is a rolling file logger. This variable dictates how many log files are retained.
3. `PATRONI_FILE_LOG_SIZE`: This variable dictates the size at which the logs will roll.

If `PATRONI_FILE_LOG_DIR` is not set than logging will be the same as what it is in the current trunk.

The loggers initialized throughout the application will inherit from the root logger and thus send their logs to the file as well when the proper env vars are set.

The handlers are explicitly removed from the loggers and closed during shutdown as a precaution.

### DIFF
--- a/patroni/__init__.py
+++ b/patroni/__init__.py
@@ -140,23 +140,21 @@ class Patroni(object):
 
 def patroni_main():
     logdir = os.environ.get('PATRONI_FILE_LOG_DIR', None)
-    logsize = os.environ.get('PATRONI_FILE_LOG_SIZE', 25000000)
-    lognum = os.environ.get('PATRONI_FILE_LOG_NUM', 4)
     logformat = os.environ.get('PATRONI_LOGFORMAT', '%(asctime)s %(levelname)s: %(message)s')
     loglevel = os.environ.get('PATRONI_LOGLEVEL', 'INFO')
-    requests_loglevel = os.environ.get('PATRONI_REQUESTS_LOGLEVEL', 'WARNING')
 
-    logging.getLogger('requests').setLevel(requests_loglevel)
     if not logdir:
         logging.basicConfig(format=logformat, level=loglevel)
     else:
+        logsize = os.environ.get('PATRONI_FILE_LOG_SIZE', 25000000)
+        lognum = os.environ.get('PATRONI_FILE_LOG_NUM', 4)
+
         root_logger = logging.getLogger()
         root_logger.setLevel(loglevel)
         handler = RotatingFileHandler(os.path.join(logdir, 'patroni.log'), mode='a', maxBytes=logsize,
                                       backupCount=lognum)
         handler.setFormatter(logging.Formatter(logformat))
         root_logger.addHandler(handler)
-        logging.getLogger('requests').addHandler(handler)
 
     patroni = Patroni()
     try:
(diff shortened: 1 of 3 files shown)


patroni/__init__.py
```diff
@@ -138,18 +139,41 @@ def shutdown(self):
 
 
 def patroni_main():
+    logdir = os.environ.get('PATRONI_FILE_LOG_DIR', None)
+    logsize = os.environ.get('PATRONI_FILE_LOG_SIZE', 25000000)
+    lognum = os.environ.get('PATRONI_FILE_LOG_NUM', 4)
     logformat = os.environ.get('PATRONI_LOGFORMAT', '%(asctime)s %(levelname)s: %(message)s')
     loglevel = os.environ.get('PATRONI_LOGLEVEL', 'INFO')
     requests_loglevel = os.environ.get('PATRONI_REQUESTS_LOGLEVEL', 'WARNING')
```
CyberDem0n: `requests_loglevel` was introduced a very long time ago, when Patroni was mostly relying on python `requests` module and more important, `urllib3` was part of `requests` module those days.
Since then a lot of things changed, nowadays Patroni is mostly relying on `urllib3` directly, which was also separated from `requests`.
Basically it all means that `logging.getLogger('requests').setLevel(requests_loglevel)` is not doing anymore what it was intended to do and we should get rid of any explicit configuration for `requests` logging completely (`requests` is not logging anything anything).

patroni/__init__.py
```diff
@@ -138,18 +139,41 @@ def shutdown(self):
 
 
 def patroni_main():
+    logdir = os.environ.get('PATRONI_FILE_LOG_DIR', None)
+    logsize = os.environ.get('PATRONI_FILE_LOG_SIZE', 25000000)
```
CyberDem0n: I would suggest moving `logsize` and `lognum` into else branch (just to reduce scope of variables).

patroni/__init__.py
```diff
@@ -138,18 +139,41 @@ def shutdown(self):
...
+        root_logger.addHandler(handler)
+        logging.getLogger('requests').addHandler(handler)
 
     patroni = Patroni()
     try:
         patroni.run()
     except KeyboardInterrupt:
         pass
     finally:
+        root_handlers = logging.getLogger().handlers
```
CyberDem0n: 1. Flushing and closing of handlers should happen after `patroni.shutdown()`
2. There is the `logging.shutdown()` function, which seems already does what you want.

capistrant:
@CyberDem0n Thanks for the feedback! I made some changes based off of your input. That `logging.shutdown()` tip was especially appreciated.

CyberDem0n:
Merged, thank you!

srikanth-medikonda:
Can these environment variables be set from YAML file directly, referring to their respective names: `logdir`, `logsize ` and `lognum`?

PR #904, opened 2018-12-26
possibility to set `logdatefmt` via env.

anikin-aa:
Guys, hi !
I've faced some issue with log date format in case of integration with Graylog. In some cases Graylog can not parse date with spaces.
So, i added new env variable `PATRONI_LOGDATEFMT` for configuration of logdatefmt.

### DIFF
--- a/patroni/__init__.py
+++ b/patroni/__init__.py
@@ -145,8 +145,7 @@ def patroni_main():
     logdatefmt = os.environ.get('PATRONI_LOGDATEFMT', '%Y-%m-%d %H:%M:%S')
 
     if not logdir:
-        logging.basicConfig(format=logformat, level=loglevel,
-                            logdatefmt=logdatefmt)
+        logging.basicConfig(format=logformat, level=loglevel, datefmt=logdatefmt)
     else:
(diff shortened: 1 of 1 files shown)


patroni/__init__.py
```diff
@@ -142,9 +142,11 @@ def patroni_main():
     logdir = os.environ.get('PATRONI_FILE_LOG_DIR', None)
     logformat = os.environ.get('PATRONI_LOGFORMAT', '%(asctime)s %(levelname)s: %(message)s')
     loglevel = os.environ.get('PATRONI_LOGLEVEL', 'INFO')
+    logdatefmt = os.environ.get('PATRONI_LOGDATEFMT', '%Y-%m-%d %H:%M:%S')
 
     if not logdir:
-        logging.basicConfig(format=logformat, level=loglevel)
+        logging.basicConfig(format=logformat, level=loglevel,
+                            logdatefmt=logdatefmt)
```
CyberDem0n: Tests are failing, because the argument name is `datefmt`, not `logdatefmt`.
Also, it makes sense to set `datefmt` for [logging.Formatter](https://github.com/zalando/patroni/pull/904/files#diff-0996184a693191991bd4f56d522c4f32R158)

PR #910, opened 2019-01-04
Explicitly secure rw perms for recovery.conf at creation time

capistrant:
During my testing of base deploys of PG with Patroni, I discovered the recovery.conf file to be world-readable despite containing the replication connection information. I believe it would be  a good idea to explicitly lock this file down at the time of creation since I believe it is a correct assumption that nobody but the user running patroni/postgres should need to read said file.

My initial approach was to use os.open() to set the file mode during creation. However, that approach was not compatible with python 2.7. Per python docs, os.chmod using `stat.S_IWRITE | stat.S_IREAD` should work across all platforms

~~O_CREAT O_TRUNC and O_WRONLY are listed as windows compatible.~~

### DIFF
--- a/patroni/postgresql.py
+++ b/patroni/postgresql.py
@@ -1185,7 +1185,7 @@ class Postgresql(object):
         return not primary_conninfo
 
     def write_recovery_conf(self, recovery_params):
-        with open(self._recovery_conf, 'w') as f:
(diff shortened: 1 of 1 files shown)


CyberDem0n:
Merged, thank you!

PR #924, opened 2019-01-11
Do not check types of standby_cluster configuration

erthalion:
In the original implementation of standby cluster all allowed properties
were checked if they're strings or not. But for `create_replica_methods`
it's wrong since it's a list, so this options is skipped. Fortunately,
if this options is not set, it will be taken from postgresql config section,
but this issue needs to be fixed anyway.

### DIFF
--- a/patroni/config.py
+++ b/patroni/config.py
@@ -48,7 +48,7 @@ class Config(object):
         'synchronous_mode': False,
         'synchronous_mode_strict': False,
         'standby_cluster': {
-            'create_replica_methods': '',
+            'create_replica_methods': [],
             'host': '',
             'port': '',
             'primary_slot_name': '',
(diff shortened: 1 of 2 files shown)


patroni/config.py
```diff
@@ -198,7 +198,9 @@ def _safe_copy_dynamic_configuration(self, dynamic_configuration):
                 allowed_keys = self.__DEFAULT_CONFIG['standby_cluster'].keys()
                 expected = {
                     k: v for k, v in (value or {}).items()
-                    if (k in allowed_keys and isinstance(v, six.string_types))
+                    if (k in allowed_keys and (
+                        isinstance(v, six.string_types) or
```
redbaron: why values are being filtered at all? `self.__DEFAULT_CONFIG['standby_cluster'].keys()` is known  to have valid keys only, right?

PR #927, opened 2019-01-14
Make logging configurable via YAML file

CyberDem0n:
It was already possible to configure logging level and format via environment variables, but such an approach wasn't very convenient and flexible.
This PR intends to solve it. First of all it makes it possible to configure logging parameters via Patroni config file and more important to change them in runtime by updating config and doing reload
or sending `SIGHUP` to the Patroni process.
In addition to that, one can redefine a log level per python module.

Important! Environment configuration names related to logging were renamed and documentation accordingly updated. For compatibility reasons Patroni still accepts `PATRONI_LOGLEVEL` and `PATRONI_FORMAT`, but some other variables related to logging, which were introduced only
recently (between releases), will stop working. I think it is ok, since we didn't release the new version yet and therefore it is very unlikely that somebody is using them except authors of corresponding PRs.

Example of log section in the config file:
```yaml
log:
  dir: /where/to/write/patron/logs  # if not specified, write logs to stderr
  file_size: 50000000  # 50MB
  file_num: 10  # keep history of 10 files
  dateformat: '%Y-%m-%d %H:%M:%S'
  loggers:  # increase log verbosity for etcd.client and urllib3
    etcd.client: DEBUG
    urllib3: DEBUG
```

### DIFF
--- a/docs/ENVIRONMENT.rst
+++ b/docs/ENVIRONMENT.rst
@@ -16,7 +16,7 @@ Global/Universal
 -  **PATRONI\_LOG\_DATEFORMAT**: sets the datetime formatting string. (see the `formatTime() documentation <https://docs.python.org/3.6/library/logging.html#logging.Formatter.formatTime>`_)
 -  **PATRONI\_LOG\_DIR**: Directory to write application logs to. The directory must exist and be writable by the user executing Patroni. If you set this env variable, the application will retain 4 25MB logs by default. You can tune those retention values with `PATRONI_LOG_FILE_NUM` and `PATRONI_LOG_FILE_SIZE` (see below).
 -  **PATRONI\_LOG\_FILE\_NUM**: The number of application logs to retain.
--  **PATRONI\_LOG\_FILE\_SIZE**: Size of log application log file that triggers a log rolling.
+-  **PATRONI\_LOG\_FILE\_SIZE**: Size of patroni.log file (in bytes) that triggers a log rolling.
 -  **PATRONI\_LOG\_LOGGERS**: Redefine logging level per python module. Example ``PATRONI_LOG_LOGGERS="{patroni.postmaster: WARNING, urllib3: DEBUG}"``
 
 Bootstrap configuration
--- a/docs/ENVIRONMENT.rst
+++ b/docs/ENVIRONMENT.rst
@@ -8,13 +8,16 @@ It is possible to override some of the configuration parameters defined in the P
 Global/Universal
 ----------------
 -  **PATRONI\_CONFIGURATION**: it is possible to set the entire configuration for the Patroni via ``PATRONI_CONFIGURATION`` environment variable. In this case any other environment variables will not be considered!
--  **PATRONI\_FILE\_LOG\_DIR**: Directory to write application logs to. The directory must exist and be writable by the user executing Patroni. If you set this env variable, the application will retain 4 25MB logs by default. You can tune those retention values with PATRONI_LOG_FILE_NUM and PATRONI_LOG_FILE_SIZE (see below).
--  **PATRONI\_FILE\_LOG\_NUM**: The number of application logs to retain.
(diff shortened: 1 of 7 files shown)


patroni/config.py
```diff
@@ -220,6 +220,15 @@ def _popenv(name):
             if value:
                 ret[param] = value
 
+        def _fix_log_env(name, oldname):
+            value = _popenv(oldname)
+            name = Config.PATRONI_ENV_PREFIX + 'LOG_' + name.upper()
+            if value and name not in os.environ:
```
zerg-junior: can you pls swap conditions as in 
`name not in os.environ and value :` or put explicit parentheses around `name not ..`?
imo makes it a bit easier to read (no need to remember precedence of logical operations in that case)

tests/test_log.py
```diff
@@ -0,0 +1,38 @@
...
+import sys
+import unittest
+import yaml
+
+from mock import Mock, patch
+from patroni.config import Config
+from patroni.log import PatroniLogger
+
+
+class TestPatroniLogger(unittest.TestCase):
```
zerg-junior: what about the unit test for  backward compatibility with the previous way to configure logging ?

docs/ENVIRONMENT.rst
```diff
@@ -8,13 +8,16 @@ It is possible to override some of the configuration parameters defined in the P
...
+-  **PATRONI\_LOG\_LEVEL**: sets the general logging level. Default value is **INFO** (see `the docs for Python logging <https://docs.python.org/3.6/library/logging.html#levels>`_)
+-  **PATRONI\_LOG\_FORMAT**: sets the log formatting string. Default value is **%(asctime)s %(levelname)s: %(message)s** (see `the LogRecord attributes <https://docs.python.org/3.6/library/logging.html#logrecord-attributes>`_)
+-  **PATRONI\_LOG\_DATEFORMAT**: sets the datetime formatting string. (see the `formatTime() documentation <https://docs.python.org/3.6/library/logging.html#logging.Formatter.formatTime>`_)
+-  **PATRONI\_LOG\_DIR**: Directory to write application logs to. The directory must exist and be writable by the user executing Patroni. If you set this env variable, the application will retain 4 25MB logs by default. You can tune those retention values with `PATRONI_LOG_FILE_NUM` and `PATRONI_LOG_FILE_SIZE` (see below).
+-  **PATRONI\_LOG\_FILE\_NUM**: The number of application logs to retain.
+-  **PATRONI\_LOG\_FILE\_SIZE**: Size of log application log file that triggers a log rolling.
```
zerg-junior: imho worth mentioning a value in bytes is expected

patroni/config.py
CyberDem0n: The code is clean and readable here:
1. `not in` is the standard python operator, it is **not** a combination of two operators
2. My internal branch predictor tells me, that most likely the value is `None` here, and therefore it should be the first in the if statement.

patroni/config.py
```diff
@@ -228,6 +237,22 @@ def _set_section_values(section, params):
...
         _set_section_values('restapi', ['listen', 'connect_address', 'certfile', 'keyfile'])
         _set_section_values('postgresql', ['listen', 'connect_address', 'data_dir', 'pgpass', 'bin_dir'])
+        _set_section_values('log', ['level', 'format', 'dateformat', 'dir', 'file_size', 'file_num', 'loggers'])
+
+        def _parse_dict(value):
+            if not value.strip().startswith('{'):
+                value = '{{{0}}}'.format(value)
+            try:
+                return yaml.safe_load(value)
+            except Exception:
```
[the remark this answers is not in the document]
CyberDem0n: That's exactly what `logger.exception` is doing. It logs the error message and stack trace.

tests/test_log.py
CyberDem0n: It is basically here: https://github.com/zalando/patroni/pull/927/files#diff-423890b7e3c2aa4791f9e203b7f194cbR33

Issue #951, opened 2019-01-26
problem with pg_hba


binooetomo:
Dear All.

Below is patroni generated pg_hba on my master node (192.168.5.81):

----------------
local   all             all                                     trust
host    all             all             127.0.0.1/32            trust
host    all             all             ::1/128                 trust
local   replication     all                                     trust
host    replication     all             127.0.0.1/32            trust
host    replication     all             ::1/128                 trust
host    replication   repl     192.168.5.0/24    md5
host    all    dba    all    md5
----------------

I try to start my second node (192.168.5.82), and got this error message :

------
2019-01-26 12:10:05,392 INFO: Local timeline=None lsn=None
2019-01-26 12:10:05,398 ERROR: Can not fetch local timeline and lsn from replication connection
Traceback (most recent call last):
  File "/var/lib/pgsql/.local/share/virtualenvs/patroni-QGg7MClq/lib/python3.6/site-packages/patroni/postgresql.py", line 1273, in _get_local_timeline_lsn_from_replication_connection
    with self._get_replication_connection_cursor(**self._local_replication_address) as cur:
  File "/usr/lib64/python3.6/contextlib.py", line 81, in __enter__
    return next(self.gen)
  File "/var/lib/pgsql/.local/share/virtualenvs/patroni-QGg7MClq/lib/python3.6/site-packages/patroni/postgresql.py", line 1257, in _get_replication_connection_cursor
    connect_timeout=3, options='-c statement_timeout=2000') as cur:
  File "/usr/lib64/python3.6/contextlib.py", line 81, in __enter__
    return next(self.gen)
  File "/var/lib/pgsql/.local/share/virtualenvs/patroni-QGg7MClq/lib/python3.6/site-packages/patroni/postgresql.py", line 1246, in _get_connection_cursor
    with psycopg2.connect(**kwargs) as conn:
  File "/var/lib/pgsql/.local/share/virtualenvs/patroni-QGg7MClq/lib/python3.6/site-packages/psycopg2/__init__.py", line 130, in connect
    conn = _connect(dsn, connection_factory=connection_factory, **kwasync)
psycopg2.OperationalError: FATAL:  no pg_hba.conf entry for replication connection from host "192.168.5.82", user "repl", SSL off

2019-01-26 12:10:05,402 INFO: no action.  i am a secondary and i am following a leader
^C2019-01-26 12:10:12,971 INFO: Lock owner: patroni-ceph-node1; I am patroni-ceph-node2

------

[...]

Issue #894, opened 2018-12-07
no pg_hba.conf entry for replication


an0ma1ia:
2-node patroni cluster up and running, replication works fine. After first failover I see the following error message on secondary node:

2018-12-07 09:40:37,183 ERROR: Exception when working with master via replication connection
Traceback (most recent call last):
  File "/usr/lib/python3/dist-packages/patroni/postgresql.py", line 1322, in _check_timeline_and_lsn
    with self._get_replication_connection_cursor(**leader.conn_kwargs()) as cur:
  File "/usr/lib/python3.6/contextlib.py", line 81, in __enter__
    return next(self.gen)
  File "/usr/lib/python3/dist-packages/patroni/postgresql.py", line 1248, in _get_replication_connection_cursor
    connect_timeout=3, options='-c statement_timeout=2000') as cur:
  File "/usr/lib/python3.6/contextlib.py", line 81, in __enter__
    return next(self.gen)
  File "/usr/lib/python3/dist-packages/patroni/postgresql.py", line 1238, in _get_connection_cursor
    with psycopg2.connect(**kwargs) as conn:
  File "/usr/lib/python3/dist-packages/psycopg2/__init__.py", line 130, in connect
    conn = _connect(dsn, connection_factory=connection_factory, **kwasync)
psycopg2.OperationalError: FATAL:  no pg_hba.conf entry for replication connection from host "192.168.56.31", user "repl", SSL off

Replication works, all DB changes from primary node I see on the secondary. What causes this error?

an0ma1ia:
All software installed from ubuntu official repo and pgdg:
`$ dpkg -l |grep -e patroni -e postgres |awk '{print $2, $3}'
patroni 1.5.1-2.pgdg18.04+1
pgdg-keyring 2018.2
postgresql-11 11.1-1.pgdg18.04+1
postgresql-11-pgmemcache 2.3.0-6.pgdg18.04+1
postgresql-client-11 11.1-1.pgdg18.04+1
postgresql-client-common 197.pgdg18.04+1
postgresql-common 197.pgdg18.04+1
postgresql-server-dev-11 11.1-1.pgdg18.04+1`

`$ cat /etc/patroni/config.yml
scope: patroni
name: ptr1

restapi:
  listen: 0.0.0.0:8008
  connect_address: 192.168.56.31:8008

etcd:
  hosts: 192.168.56.31:2379,192.168.56.32:2379,192.168.56.1:2379

bootstrap:
  dcs:
[...]

PR #952, opened 2019-01-28
Always use replication=1, otherwise it is considered a logical walsender

CyberDem0n:
Fixes: https://github.com/zalando/patroni/issues/894
Fixes: https://github.com/zalando/patroni/issues/951

### DIFF
--- a/patroni/postgresql.py
+++ b/patroni/postgresql.py
@@ -1250,9 +1250,7 @@ class Postgresql(object):
 
     @contextmanager
     def _get_replication_connection_cursor(self, host='localhost', port=5432, database=None, **kwargs):
-        database = database or self._database
-        replication = 'database' if self._major_version >= 90400 else 1
-        with self._get_connection_cursor(host=host, port=int(port), database=database, replication=replication,
+        with self._get_connection_cursor(host=host, port=int(port), database=database or self._database, replication=1,
                                          user=self._replication['username'], password=self._replication['password'],
                                          connect_timeout=3, options='-c statement_timeout=2000') as cur:
             yield cur


redbaron:
what is the consequence for PG10 users without this patch ?

CyberDem0n:
The same as for pg11, pg_rewind will not work in some cases. But if you haven't seen exceptions similar to the #894 and #951, you are probably fine.

I didn't fugured out yet what exactly triggers the problem, because for example when I am doing a local testing with postgres running on localhost everything seems to be fine with following pg_hba:
```
# TYPE  DATABASE        USER            ADDRESS                 METHOD

# "local" is for Unix domain socket connections only
local   all             all                                     md5
# IPv4 local connections:
host    all             all             127.0.0.1/32            md5
# IPv6 local connections:
host    all             all             ::1/128                 md5
# Allow replication connections from localhost, by a user with the
# replication privilege.
local   replication     all                                     md5
host    replication     all             127.0.0.1/32            md5
host    replication     all             ::1/128                 md5

host replication replicator 127.0.0.1/32 md5
host all all 0.0.0.0/0 md5
```

Superuser name is postgres in my case and replication user is replicator

CyberDem0n:
After a bit more investigation I've figured out why it happens on 10+.

If we are opening replication connection with replication=database option instead of replication=1, postgres will accept it with `host all ...` or `host postgres ...` line in the pg_hba. `host replication ...` will not accept it, because it works exclusively for replication=1 case.

Basically you can mitigate the problem by adding one more line to pg_hba.

mhagander:
I can confirm running into this, and that the workaround with an extra line in pg_hba.conf works fine.

mbanck:
It took me hours the other day to figure out what was wrong with my pg_hba on v11, so +1 for this PR. I've since pulled it into the Debian package.

PR #966, opened 2019-02-07
Change base port for acceptance tests from 5440 to 5360

mbanck:
This allows for enough room in order not to interfere with any possibly running
host instances.  On Debian/Ubuntu, host instances get ports assigned starting
from 5432, so in the (rare but possible) case that 9 instances are running on
the host, the acceptance tests would fail to startup their instances.

Using 5340 as base port will allow for more than 90 instances to be created
during the acceptance tests, which should be plenty considering that currenlty
around 20-25 are created.

### DIFF
--- a/features/environment.py
+++ b/features/environment.py
@@ -84,7 +84,7 @@ class AbstractController(object):
 
 
 class PatroniController(AbstractController):
-    __PORT = 5440
+    __PORT = 5340
     PATRONI_CONFIG = '{}.yml'
     """ starts and stops individual patronis"""
 
--- a/features/environment.py
+++ b/features/environment.py
@@ -84,7 +84,7 @@ class AbstractController(object):
 
 
 class PatroniController(AbstractController):
-    __PORT = 5340
+    __PORT = 5360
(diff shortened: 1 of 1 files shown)


mbanck:
Not sure why the DCS="kubernetes" Travis job failed, is that happening sometimes or is there a possible port clash?

mbanck:
I had a look at https://en.wikipedia.org/wiki/List_of_TCP_and_UDP_port_numbers and while there is nothing very bad in the range 5340-5360, some TCP ports are assigned. So maybe 5360 is a better choice? Nothing is assigned from 5360-5400 really, with the obscure exception of 5394.

Then again, maybe even better would be to just try: the port and switch to the next if that one doesn't work, but I haven't looked how involved that would be?

CyberDem0n:
5394 is unofficially allocated on UDP, therefore I believe starting from 5360 should be fine.

mbanck:
> 5394 is unofficially allocated on UDP, therefore I believe starting from 5360 should be fine.

Ok, done that now.

PR #967, opened 2019-02-07
Copy away output directories of failed acceptance tests.

mbanck:
This is a (crude) patch we carry in the Debian/Ubuntu packaging in order to output the content of failed acceptance tests on `ci.debian.net` as we don't have shell access to the test runners.

This might not be acceptable for the main Patroni project, but I thought I'd give it a shot. Is there maybe some other `behave` option that allows to dump the logs on failure directly that we should use? 

I believe Travis right now just dumps everything on failure, so this could make wading through Travis log easier as well.


mbanck:
@CyberDem0n good point, done that now

PR #968, opened 2019-02-07
Run coverage as python -m coverage

mbanck:
On Debian/Ubuntu, there is no `/usr/bin/coverage` binary, it is called
`/usr/bin/python3-coverage`.  In order to handle both, check which one is
available and use the `COVERAGE_BIN` variable in order to run that.


mbanck:
Hrm, Travis runs python2.7 as well (I was under the impression that Patroni only supports python3 now), where `shutil.which()` is not available.

mbanck:
Well, works for me now, but Quality Review complains, is that a big concern?

PR #972, opened 2019-02-08
Honor remove_data_directory_on_rewind_failure if pg_rewind can't be used

mbanck:
If the checkpoint prior to running `pg_rewind` cannot be issued (e.g. because no password is set for `postgres`), still honor `remove_data_directory_on_rewind_failure` and re-clone from the primary instead of giving up.

This results in the following log:

```
Feb 08 16:30:09 pg2 patroni[8671]: 2019-02-08 16:30:09,991 INFO: running pg_rewind from pg1
Feb 08 16:30:15 pg2 patroni[8671]: 2019-02-08 16:30:15,061 ERROR: Exception during CHECKPOINT
Feb 08 16:30:15 pg2 patroni[8671]: Traceback (most recent call last):
Feb 08 16:30:15 pg2 patroni[8671]:   File "/usr/lib/python3/dist-packages/patroni/postgresql.py", line 948, in checkpoint
Feb 08 16:30:15 pg2 patroni[8671]:     with self._get_connection_cursor(**connect_kwargs) as cur:
Feb 08 16:30:15 pg2 patroni[8671]:   File "/usr/lib/python3.7/contextlib.py", line 112, in __enter__
Feb 08 16:30:15 pg2 patroni[8671]:     return next(self.gen)
Feb 08 16:30:15 pg2 patroni[8671]:   File "/usr/lib/python3/dist-packages/patroni/postgresql.py", line 1250, in _get_connection_cursor
Feb 08 16:30:15 pg2 patroni[8671]:     with psycopg2.connect(**kwargs) as conn:
Feb 08 16:30:15 pg2 patroni[8671]:   File "/usr/lib/python3/dist-packages/psycopg2/__init__.py", line 130, in connect
Feb 08 16:30:15 pg2 patroni[8671]:     conn = _connect(dsn, connection_factory=connection_factory, **kwasync)
Feb 08 16:30:15 pg2 patroni[8671]: psycopg2.OperationalError: fe_sendauth: no password supplied
Feb 08 16:30:15 pg2 patroni[8671]: 2019-02-08 16:30:15,064 WARNING: Can not use pg1 for rewind: not accessible or not healty
Feb 08 16:30:15 pg2 patroni[8671]: 2019-02-08 16:30:15,064 WARNING: remove_data_directory_on_rewind_failure is set. removing...
Feb 08 16:30:15 pg2 patroni[8671]: 2019-02-08 16:30:15,064 INFO: Removing data directory: /var/lib/postgresql/11/test
Feb 08 16:30:15 pg2 patroni[8671]: 2019-02-08 16:30:15,102 INFO: Lock owner: pg1; I am pg2
Feb 08 16:30:15 pg2 patroni[8671]: 2019-02-08 16:30:15,104 INFO: trying to bootstrap from leader 'pg1'
```

instead of:

```
Feb 08 15:55:09 pg3 patroni[9315]: 2019-02-08 15:55:09,632 INFO: running pg_rewind from pg1
Feb 08 15:55:09 pg3 patroni[9315]: 2019-02-08 15:55:09,636 ERROR: Exception during CHECKPOINT
Feb 08 15:55:09 pg3 patroni[9315]: Traceback (most recent call last):
Feb 08 15:55:09 pg3 patroni[9315]:   File "/usr/lib/python3/dist-packages/patroni/postgresql.py", line 948, in checkpoint
Feb 08 15:55:09 pg3 patroni[9315]:     with self._get_connection_cursor(**connect_kwargs) as cur:
Feb 08 15:55:09 pg3 patroni[9315]:   File "/usr/lib/python3.7/contextlib.py", line 112, in __enter__
Feb 08 15:55:09 pg3 patroni[9315]:     return next(self.gen)
Feb 08 15:55:09 pg3 patroni[9315]:   File "/usr/lib/python3/dist-packages/patroni/postgresql.py", line 1250, in _get_connection_cursor
Feb 08 15:55:09 pg3 patroni[9315]:     with psycopg2.connect(**kwargs) as conn:
Feb 08 15:55:09 pg3 patroni[9315]:   File "/usr/lib/python3/dist-packages/psycopg2/__init__.py", line 130, in connect
Feb 08 15:55:09 pg3 patroni[9315]:     conn = _connect(dsn, connection_factory=connection_factory, **kwasync)
Feb 08 15:55:09 pg3 patroni[9315]: psycopg2.OperationalError: fe_sendauth: no password supplied
Feb 08 15:55:09 pg3 patroni[9315]: 2019-02-08 15:55:09,636 WARNING: Can not use pg1 for rewind: not accessible or not healty
```

and a cluster member in an unrecoverable state.

### DIFF
--- a/patroni/postgresql.py
+++ b/patroni/postgresql.py
@@ -1400,7 +1400,16 @@ class Postgresql(object):
         # make it store the new timeline (5540277D.8020309@iki.fi)
         leader_status = self.checkpoint(r)
         if leader_status:
-            return logger.warning('Can not use %s for rewind: %s', leader.name, leader_status)
+            logger.warning('Can not use %s for rewind: %s', leader.name, leader_status)
+            for name in ('remove_data_directory_on_rewind_failure', 'remove_data_directory_on_diverged_timelines'):
+                if self.config.get(name):
+                    logger.warning('%s is set. removing...', name)
+                    self.remove_data_directory()
+                    self._rewind_state = REWIND_STATUS.INITIAL
+                    break
+            else:
+                self._rewind_state = REWIND_STATUS.FAILED
+            return False
 
         if self.pg_rewind(r):
             self._rewind_state = REWIND_STATUS.SUCCESS
--- a/patroni/postgresql.py
+++ b/patroni/postgresql.py
@@ -954,7 +954,7 @@ class Postgresql(object):
                 return cur.execute('CHECKPOINT')
         except psycopg2.Error as e:
             logger.exception('Exception during CHECKPOINT')
-            if str(e).strip()  == 'fe_sendauth: no password supplied':
+            if str(e).strip() == 'fe_sendauth: no password supplied':
                 return 'no password'
             else:
                 return 'not accessible or not healty'
--- a/patroni/postgresql.py
+++ b/patroni/postgresql.py
@@ -952,9 +952,12 @@ class Postgresql(object):
                     if cur.fetchone()[0]:
                         return 'is_in_recovery=true'
                 return cur.execute('CHECKPOINT')
-        except psycopg2.Error:
+        except psycopg2.Error as e:
             logger.exception('Exception during CHECKPOINT')
-            return 'not accessible or not healty'
+            if str(e).strip()  == 'fe_sendauth: no password supplied':
+                return 'no password'
+            else:
+                return 'not accessible or not healty'
 
     def stop(self, mode='fast', block_callbacks=False, checkpoint=None, on_safepoint=None):
         """Stop PostgreSQL
@@ -1402,7 +1405,7 @@ class Postgresql(object):
         if leader_status:
             logger.warning('Can not use %s for rewind: %s', leader.name, leader_status)
             for name in ('remove_data_directory_on_rewind_failure', 'remove_data_directory_on_diverged_timelines'):
-                if self.config.get(name):
+                if self.config.get(name) and leader_status == 'no password':
                     logger.warning('%s is set. removing...', name)
                     self.remove_data_directory()
                     self._rewind_state = REWIND_STATUS.INITIAL


CyberDem0n:
This is a very controversial change which breaks the functionality when superuser credentials are configured correctly.
We don't really want to destroy data directory **before** pg_rewind failed.

mbanck:
> This is a very controversial change which breaks the functionality when superuser credentials are configured correctly.
> We don't really want to destroy data directory **before** pg_rewind failed.

Ok, that's a fair point.

What I would like to do is being able to set `use_pg_rewind: true` and then depending on whether a `postgres` password is set or not, either have `pg_rewind` do its work, or re-clone.

Would you accept a change where the re-clone is guarded against an empty `postgres` password, or is that still not enough to blast away `PGDATA`?

CyberDem0n:
> What I would like to do is being able to set use_pg_rewind: true and then depending on whether a postgres password is set or not, either have pg_rewind do its work, or re-clone.

Unforutunately it is a bit more complicated.
I am 100% sure it is possible to get `pg_rewind` working when there is no superuser credentials specified in the config with the help of trust line in the pg_hba.con or when there is a `~/.pgpass`.

> Would you accept a change where the re-clone is guarded against an empty `postgres` password, or is that still not enough to blast away `PGDATA`?

I beleive we are allowed to re-clone only if authentication fails due to the wrong or empty password. Unfortunately when connection fails psycopg2 doesn't give us access to the error codes but only to error messages, which could be localized https://github.com/psycopg/psycopg2/issues/417 :(

mbanck:
> > What I would like to do is being able to set use_pg_rewind: true and then depending on whether a postgres password is set or not, either have pg_rewind do its work, or re-clone.
> 
> Unforutunately it is a bit more complicated.
> I am 100% sure it is possible to get `pg_rewind` working when there is no superuser credentials specified in the config with the help of trust line in the pg_hba.con or when there is a `~/.pgpass`.

I don't `trust` for TCP/IP connections should be endorsed. But yeah, having it in `~/.pgpass` would be an option, but once that is not managed by Patroni, we are talking pets again, not cattle.
 
> > Would you accept a change where the re-clone is guarded against an empty `postgres` password, or is that still not enough to blast away `PGDATA`?
> 
> I beleive we are allowed to re-clone only if authentication fails due to the wrong or empty password. Unfortunately when connection fails psycopg2 doesn't give us access to the error codes but only to error messages, which could be localized [psycopg/psycopg2#417](https://github.com/psycopg/psycopg2/issues/417) :(

Hrm. If I read that issue correctly then the localization issue is for errors that come from the server ("`
FATAL: password authentication failed for user "x" `" in their case), whereas `fe_sendauth: no password supplied` is a libpq errror, and, from looking at the code, is never translated.

Indeed:

```
>>> import psycopg2
>>> conn = psycopg2.connect("host=10.0.3.156 dbname=test user=postgres")
  File "/usr/lib/python3/dist-packages/psycopg2/__init__.py", line 130, in connect
    conn = _connect(dsn, connection_factory=connection_factory, **kwasync)
psycopg2.OperationalError: fe_sendauth: no password supplied

>>> conn = psycopg2.connect("host=10.0.3.156 dbname=test user=postgres password=foo")
  File "/usr/lib/python3/dist-packages/psycopg2/__init__.py", line 130, in connect
    conn = _connect(dsn, connection_factory=connection_factory, **kwasync)
psycopg2.OperationalError: FATAL:  Passwort-Authentifizierung für Benutzer »postgres« fehlgeschlagen
FATAL:  Passwort-Authentifizierung für Benutzer »postgres« fehlgeschlagen
```

So I think we could compare the exception text to the above `fe_sendauth: no password supplied`, return a different error message (like `no password`) and act on that in `rewind()`, what do you think?

patroni/postgresql.py
```diff
@@ -1400,7 +1403,16 @@ def rewind(self, leader):
         # make it store the new timeline (5540277D.8020309@iki.fi)
         leader_status = self.checkpoint(r)
         if leader_status:
-            return logger.warning('Can not use %s for rewind: %s', leader.name, leader_status)
+            logger.warning('Can not use %s for rewind: %s', leader.name, leader_status)
+            for name in ('remove_data_directory_on_rewind_failure', 'remove_data_directory_on_diverged_timelines'):
+                if self.config.get(name) and leader_status == 'no password':
```
CyberDem0n: check `leader_status == 'no password'` could be done outside the loop

patroni/postgresql.py
```diff
@@ -1400,7 +1403,16 @@ def rewind(self, leader):
         # make it store the new timeline (5540277D.8020309@iki.fi)
         leader_status = self.checkpoint(r)
         if leader_status:
-            return logger.warning('Can not use %s for rewind: %s', leader.name, leader_status)
+            logger.warning('Can not use %s for rewind: %s', leader.name, leader_status)
+            for name in ('remove_data_directory_on_rewind_failure', 'remove_data_directory_on_diverged_timelines'):
```
CyberDem0n: This loop is absolutely the same as a few lines below: https://github.com/zalando/patroni/pull/972/files#diff-d8c1582337c8426ef750420dc6b76949R1424

I think it makes sense to move it into the separate method to avoid code duplication.

mbanck:
> You only partially covered the case when the password is not supplied.
> Why partially? Because there might be `~/.pgpass` and in such case we will get a different error, which might be localized :(

Hrm, I understood your point in the following way:

Even if there is an empty password in the patroni config, the admin could've set a password manually and registered it via `.pgpass`, so wiping `PGDATA` in this case would not be desirable as it is unclear whether that is the admin's intention (`pg_rewind` hasn't failed yet, and it could be just fat-fingering the password in `.pgpass` or something).

If that's not true, then yeah, the patch would be inadequate. But maybe then it'd be indeed too much of a hazard?

CyberDem0n:
> Even if there is an empty password in the patroni config, the admin could've set a password manually and registered it via .pgpass, so wiping PGDATA in this case would not be desirable as it is unclear whether that is the admin's intention (pg_rewind hasn't failed yet, and it could be just fat-fingering the password in .pgpass or something).

Correct.

And actually your patch won't remove the PGDATA in this case, because the exception string will look different from the `fe_sendauth: no password supplied`. Maybe it is fine, not really sure.

CyberDem0n:
Let's try to approach this problem from a different side.

If somebody configured Patroni to use pg_rewind, that also implies that this person should take care of the possibility to connect as superuser between nodes, either via configuring trust in pg_hba.conf between nodes, by putting the superuser credentials into the Patroni config or by writing credentials into `~/.pgpass`.

When Patroni can't access the primary with superuser credentials and gets any kind of authentication error (including `fe_sendauth: no password supplied`), is the clear indicator of the wrong configuration and I think it should not lead to PGDATA removal.

Basically that was our original idea of how `remove_data_directory_on_rewind_failure` option should work.

mbanck:
Ok, so taking another step back, I think there's three possible options the user might want to choose from in case the former primary cannot get back online as-is:

1. Use pg_rewind, potentially wiping out some unreplicated transactions before the failover
2. Re-clone, potentially wiping out some unreplicated transactions before the failover
3. Do nothing, in order to e.g. manually inspect any unreplicated transactions before the failover

I think it is not clear enough from the current options which ones the user needs to set in order to get 1, 2 or 3. 

1 is pretty clear (`use_pg_rewind: true`), but it also needs (i) allowance of remote md5 connections and (ii) an md5 password for postgres.

Option 2 is probably what you get if you set `use_pg_rewind:false` and `remove_data_directory_on_diverged_timelines: true`.

Option 3 is probably `use_pg_rewind:false` and `remove_data_directory_on_diverged_timelines: false`?

Is this documented somewhere in detail? Any other options I missed?

As this is (in my opinion) a critical configuration step for an automated failover system, I think that for a 1.6 or 2.0 release, the configuration (and documentation) could be made clearer in this regard.

CyberDem0n:
> 1 is pretty clear (use_pg_rewind: true), but it also needs (i) allowance of remote md5 connections and (ii) an md5 password for postgres.

Correct. But it is not limited by md5. Any authentication method will work.

> Option 2 is probably what you get if you set use_pg_rewind:false and remove_data_directory_on_diverged_timelines: true.

Correct.

> Option 3 is probably use_pg_rewind:false and remove_data_directory_on_diverged_timelines: false?

Correct. Actually these are default values.

> Is this documented somewhere in detail? Any other options I missed?

Yes, they are [documented](https://patroni.readthedocs.io/en/latest/SETTINGS.html?highlight=use_pg_rewind#postgresql). Not sure if it is detailed enough.


> As this is (in my opinion) a critical configuration step for an automated failover system, I think that for a 1.6 or 2.0 release, the configuration (and documentation) could be made clearer in this regard.

Sure. But unfortunately I am not the best writer :(

CyberDem0n:
Closing this one, because:
1. The new primary is doing forced checkpoint on its own and replicas are waiting for it if they need to rewind
2. On pg11+ it is possible to use pg_rewind without superuser

PR #980, opened 2019-02-20
Update docker-compose demo cluster

CyberDem0n:
1. Multi-stage build with an extensive cleanup of useless files and optional image compression
2. Start three-node etcd cluster
3. Start three-node Patroni cluster
4. One container with haproxy
5. All container names are prefixed with "demo-" and don't have suffixes
6. Decommission dev_patroni_cluster.sh script, docker-compose is now standard de-facto.
7. Provide more examples in the docker/README.md


Dockerfile
```diff
@@ -1,45 +1,152 @@
 ## This Dockerfile is meant to aid in the building and debugging patroni whilst developing on your local machine
 ## It has all the necessary components to play/debug with a single node appliance, running etcd
-FROM postgres:10
-MAINTAINER Alexander Kukushkin <alexander.kukushkin@zalando.de>
+ARG PG_MAJOR=10
```
alexeyklyukin: Let's use 11, should we?

CyberDem0n: It doesn't bring many benefits but unfortunately adds 82MB to uncompressed image and 23MB to compressed version (55MB vs 78MB) due to llvm libraries :(

In any case it is not a big deal to change the default value to 11 and specify the custom version during build: `--build-arg PG_MAJOR=10`

Dockerfile
```diff
@@ -1,45 +1,152 @@
...
+        && echo 'postgres ALL=(ALL) NOPASSWD: /bin/tar xpJf /a.tar.xz -C /, /bin/rm /a.tar.xz, /bin/ln -snf dash /bin/sh' >> /etc/sudoers \
+        && ln -snf busybox /bin/sh \
+        && files="/bin/sh /usr/bin/sudo /usr/lib/sudo/sudoers.so /lib/x86_64-linux-gnu/security/pam_*.so" \
+        && libs="$(ldd $files | awk '{print $3;}' | grep '^/' | sort -u) /lib/x86_64-linux-gnu/ld-linux-x86-64.so.* /lib/x86_64-linux-gnu/libnsl.so.* /lib/x86_64-linux-gnu/libnss_compat.so.*" \
+        && (echo /var/run $files $libs | tr ' ' '\n' && realpath $files $libs) | sort -u | sed 's/^\///' > /exclude \
+        && find /lib/x86_64-linux-gnu/security/ -type f >> /exclude \
+        && find /etc/alternatives -xtype l -delete \
+        && save_dirs="usr lib var bin sbin etc/ssl etc/init.d etc/alternatives etc/apt" \
+        && XZ_OPT=-e9v tar -X /exclude -cpJf a.tar.xz $save_dirs \
+        && /bin/busybox sh -c "(find $save_dirs -not -type d && cat /exclude /exclude && echo exclude) | sort | uniq -u | xargs /bin/busybox rm" \
```
alexeyklyukin: I know this line has been imported from Spilo, but it craves a comment to preclude bug reports  about the duplicate command there, something like 
```
# we call "cat /exclude" to avoid including files from the $save_dirs that are
# also among the exceptions listed in the /exclude, as "uniq -u" eliminates
# all non-unique lines.  By calling "cat /exclude" a second time we guarantee
# that there will be at least two lines for each exception and therefore they
# will be excluded from the output passed to 'rm'.
```

Dockerfile
```diff
@@ -1,45 +1,152 @@
...
+RUN if [ "$COMPRESS" = "true" ]; then \
+        cd / \
+        && set -ex \
+        # Allow certain sudo commands from postgres
+        && echo 'postgres ALL=(ALL) NOPASSWD: /bin/tar xpJf /a.tar.xz -C /, /bin/rm /a.tar.xz, /bin/ln -snf dash /bin/sh' >> /etc/sudoers \
+        && ln -snf busybox /bin/sh \
+        && files="/bin/sh /usr/bin/sudo /usr/lib/sudo/sudoers.so /lib/x86_64-linux-gnu/security/pam_*.so" \
+        && libs="$(ldd $files | awk '{print $3;}' | grep '^/' | sort -u) /lib/x86_64-linux-gnu/ld-linux-x86-64.so.* /lib/x86_64-linux-gnu/libnsl.so.* /lib/x86_64-linux-gnu/libnss_compat.so.*" \
+        && (echo /var/run $files $libs | tr ' ' '\n' && realpath $files $libs) | sort -u | sed 's/^\///' > /exclude \
+        && find /lib/x86_64-linux-gnu/security/ -type f >> /exclude \
```
alexeyklyukin: Looks like a duplicate; there is nothing but pam_ there and those we have already excluded; even if there were anything else, we should have add shared libraries returned from ldd together with those objects.

PR #981, opened 2019-02-20
Use create_replica_methods from standby_cluster for replica bootstrap

CyberDem0n:
It might happen that the standby cluster is configured to be created and replay WAL files from the different source than when it is running not in standby mode.  This is necessary to avoid writing WAL files and backups into the old place after promotion.

The easiest way to achieve such behavior is passing RemoteMember object to `Postgresql.clone` method instead of the usual Member object.


Jan-M:
:+1:

Issue #992, opened 2019-03-05
Setting logging_collector to false breaks archive_command stdout file descriptor.


kupson:
To deliver Postgresql logs into docker/Kubernetes I'm setting `logging_collector` to `false` and `log_destination` to `stderr`.

That causes the default `wal-g` `archive_command` to be killed with SIGPIPE as it's standard output fd (pipe) is closed.

The receiving side is closed after being used to pass the pid of the postmaster in:
https://github.com/zalando/patroni/blob/e670122c805176a3689a7ddc8849a557038bde58/patroni/postmaster.py#L162-L164

Maybe it would be better to use [pass_fds](https://docs.python.org/3.6/library/subprocess.html#popen-constructor) to communicate pid to parent and leave stdout alone?

CyberDem0n:
Hi @kupson,

unfortunately `pass_fds` is not available on python 2.7, but I think it should be possible to get the same results with https://docs.python.org/2.7/library/multiprocessing.html

CyberDem0n:
Despite 2.7 being EOL in 2020 there are still a lot of systems where only 2.7 available :(

redbaron:
whats your view to do bug fixes and/or features to be introduced only in python3.x version of patroni when keeping python 2.7 compatibility is too hacky? for instance this fix can call subprocess with a set of flags supported by the given runtime and therefore bug would be fixed under 3.x python.


using multiprocess for double forking just to pass file descriptor is super ugly

CyberDem0n:
Double forking itself is already ugly enough :)
Actually I am also not sure that `pass_fds` will work on Windows.

So far it wasn't that hard to keep Patroni 2.7/3.X compatible, everything is more or less covered with `six`.

Jan-M:
Interesting discussion for the Python 2.7 vs Python 3.x. At some point we may need to favor 3.x over additional work for 2.7.

PR #994, opened 2019-03-06
Replace self-execute with multiprocessing.Process

CyberDem0n:
In addition to that transfer postmaster pid to Patroni process with the help of multiprocessing.Pipe instead of using stdin-stdout pipes.

Closes https://github.com/zalando/patroni/issues/992

### DIFF
--- a/patroni/postmaster.py
+++ b/patroni/postmaster.py
@@ -19,7 +19,7 @@ def pg_ctl_start(conn, cmdline, env):
     if os.name != 'nt':
         os.setsid()
     try:
-        postmaster = subprocess.Popen(cmdline, env=env)
+        postmaster = subprocess.Popen(cmdline, close_fds=True, env=env)
         conn.send(postmaster.pid)
     except Exception:
         logger.exception('Failed to execute %s', cmdline)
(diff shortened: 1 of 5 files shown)


redbaron:
IMHO it would be cleaner to create multiprocess.Pipe(blocking=False) in
grandparent, then call child via Popen without close_fds and and then child
to call Popen again but with close_fds this time.

multiprocessing is really a wrong tool for demonizing

CyberDem0n:
Could you please be more specific, what is wrong with `multiprocessing`?
How `multiprocessing.Process` is different from the `subprocess.Popen` if our goal is just executing one of **OUR** functions?

With `subprocess.Popen` it was actually way uglier due to the need to execute itself.

patroni/postmaster.py
```diff
@@ -159,10 +170,13 @@ def start(pgcommand, data_dir, conf, options):
             pass
         cmdline = [pgcommand, '-D', data_dir, '--config-file={}'.format(conf)] + options
         logger.debug("Starting postgres: %s", " ".join(cmdline))
-        proc = call_self(['pg_ctl_start'] + cmdline, close_fds=(os.name != 'nt'),
-                         stdout=subprocess.PIPE, env=env)
-        pid = int(proc.stdout.readline().strip())
-        proc.wait()
+        parent_conn, child_conn = multiprocessing.Pipe()
```
redbaron: `multiprocessing.Pipe(False)` would communicate intent of this code better

patroni/postmaster.py
```diff
@@ -16,6 +15,18 @@
 }
 
 
+def pg_ctl_start(conn, cmdline, env):
+    if os.name != 'nt':
+        os.setsid()
+    try:
+        postmaster = subprocess.Popen(cmdline, env=env)
```
redbaron: shouldn't we keep  `close_fds` here

PR #995, opened 2019-03-06
Mock external resources

CyberDem0n:
unit tests should not accidentally hit running Postgres, DCS or filesystem unless we want it explicitly.


tests/test_etcd.py
```diff
@@ -210,8 +210,8 @@ def test_api_execute(self, mock_machines):
         self.client.api_execute('/', 'POST', timeout=0)
         self.client._machines_cache = [self.client._base_uri]
         self.assertRaises(etcd.EtcdWatchTimedOut, self.client.api_execute, '/timeout', 'POST', params={'wait': 'true'})
+        self.assertRaises(etcd.EtcdWatchTimedOut, self.client.api_execute, '/timeout', 'POST', params={'wait': 'true'})
```
Jan-M: ?

tests/test_ha.py
```diff
@@ -322,7 +324,7 @@ def test_promote_without_watchdog(self):
             self.assertEqual(self.ha.run_cycle(), 'Not promoting self because watchdog could not be activated')
 
     def test_leader_with_lock(self):
-        self.ha.cluster = get_cluster_not_initialized_without_leader()
+        self.ha.cluster = get_cluster_initialized_with_leader()
```
Jan-M: :-)

tests/test_etcd.py
Jan-M: Maybe worth a comment?

Issue #988, opened 2019-03-01
Callback executed with wrong arguments when promoted to the standby leader


CyberDem0n:
It happens because the role is explicitly set before calling `follow()`, which resets it back.
There is also a problem with REST API on standby leader, it still shows itself as a `replica`.

PR #998, opened 2019-03-06
Fix callbacks behavior (mostly for standby cluster)

CyberDem0n:
First of all, this patch changes the behavior of `on_start`/`on_restart` callbacks, they will be called only when postgres is started or restarted without role changes. In case if the member is promoted or demoted only the `on_role_change` callback will be executed. `on_role_change` was never called for standby leader, only `on_start`/`on_restart` and with a wrong role argument.
Before that `on_role_change` was never called for standby leader, only `on_start`/`on_restart` and with a wrong role argument.

In addition to that, the REST API will return standby_leader role for the leader of the standby cluster.

Closes https://github.com/zalando/patroni/issues/988


features/steps/standby_cluster.py
```diff
@@ -37,12 +42,18 @@ def start_patroni_stanby_cluster(context, name, cluster_name, name2):
         "scope": cluster_name,
```
avaczi: Would this be the right time to fix the function name `start_patroni_stanby_cluster` (a missing d)?

avaczi:
:+1:

Issue #1013, opened 2019-03-26
Cannot run `python setup.py test` under Windows


pashagolub:
Running `python setup.py test` produces error:

```
===================================================== ERRORS ===================================================== ____________________________________ ERROR collecting tests/test_watchdog.py _____________________________________ 
tests\test_watchdog.py:2: in <module>
    import patroni.watchdog.linux as linuxwd
patroni\watchdog\linux.py:3: in <module>
    import fcntl
E   ModuleNotFoundError: No module named 'fcntl'
____________________________________ ERROR collecting tests/test_watchdog.py _____________________________________ 
ImportError while importing test module 'D:\3_Cybertec\patroni\tests\test_watchdog.py'.
Hint: make sure your test modules/packages have valid Python names.
Traceback:
C:\Users\pasha\AppData\Local\Programs\Python\Python37-32\lib\site-packages\_pytest\python.py:499: in _importtestmodule
    mod = self.fspath.pyimport(ensuresyspath=importmode)
C:\Users\pasha\AppData\Local\Programs\Python\Python37-32\lib\site-packages\py\_path\local.py:701: in pyimport
    __import__(modname)
C:\Users\pasha\AppData\Local\Programs\Python\Python37-32\lib\site-packages\_pytest\assertion\rewrite.py:296: in load_module
    six.exec_(co, mod.__dict__)
tests\test_watchdog.py:2: in <module>
    import patroni.watchdog.linux as linuxwd
patroni\watchdog\linux.py:3: in <module>
    import fcntl
E   ModuleNotFoundError: No module named 'fcntl'
___________________________________ ERROR collecting patroni/watchdog/linux.py ___________________________________ 
patroni\watchdog\linux.py:3: in <module>
    import fcntl
E   ModuleNotFoundError: No module named 'fcntl'
```

PR #1014, opened 2019-03-26
#1013 Fix tests for Windows

pashagolub:
`fcntl` import clause moved from the top and check added for OS name

### DIFF
--- a/tests/test_patroni.py
+++ b/tests/test_patroni.py
@@ -109,11 +109,11 @@ class TestPatroni(unittest.TestCase):
         self.p.ha.dcs.watch = Mock(side_effect=SleepException)
         self.p.api.start = Mock()
         self.p.config._dynamic_configuration = {}
-        # self.assertRaises(SleepException, self.p.run)
-        # with patch('patroni.config.Config.set_dynamic_configuration', Mock(return_value=True)):
-        #     self.assertRaises(SleepException, self.p.run)
-        # with patch('patroni.postgresql.Postgresql.data_directory_empty', Mock(return_value=False)):
-        #     self.assertRaises(SleepException, self.p.run)
+        self.assertRaises(SleepException, self.p.run)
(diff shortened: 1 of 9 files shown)


tests/test_patroni.py
```diff
@@ -108,11 +109,11 @@ def test_run(self):
         self.p.ha.dcs.watch = Mock(side_effect=SleepException)
         self.p.api.start = Mock()
         self.p.config._dynamic_configuration = {}
-        self.assertRaises(SleepException, self.p.run)
-        with patch('patroni.config.Config.set_dynamic_configuration', Mock(return_value=True)):
-            self.assertRaises(SleepException, self.p.run)
-        with patch('patroni.postgresql.Postgresql.data_directory_empty', Mock(return_value=False)):
-            self.assertRaises(SleepException, self.p.run)
+        # self.assertRaises(SleepException, self.p.run)
```
CyberDem0n: Why it is commented out?

pashagolub: Oops! My bad!

pashagolub: Fixed

tests/test_patroni.py
```diff
@@ -87,8 +87,9 @@ def mock_signal(signo, handler):
         ref = {'passtochild': lambda signo, stack_frame: 0}
 
         def mock_sighup(signo, handler):
-            if signo == signal.SIGHUP:
-                ref['passtochild'] = handler
+            if getattr(signal, 'SIGHUP', None) is not None:
```
CyberDem0n:
```python
if hasattr(signal, 'SIGHUP') and signo == signal.SIGHUP:
```

tests/test_postmaster.py
```diff
@@ -1,10 +1,21 @@
 import psutil
 import unittest
+import os
```
CyberDem0n: Unused import

PR #1015, opened 2019-03-26
Enabling kerberos support

vilajit:
If config file has user (super, replication and others) but with no passwords, assume these are for kerberos (gss) access.
User will be created without passwords.
Connect will require the kerberos spn and kerberos key file.

### DIFF
--- a/patroni/postgresql.py
+++ b/patroni/postgresql.py
@@ -113,6 +113,12 @@ class Postgresql(object):
         self._configure_server_parameters()
 
         self._connect_address = config.get('connect_address')
+        self._krbsrvname = config.get('krbsrvname', None)
+
+        # for not so obvious connection attempts that may happen outside of pyscopg2
+        if self._krbsrvname:
+            os.environ['PGKRBSRVNAME'] = self._krbsrvname
+
         self._superuser = config['authentication'].get('superuser', {})
         self.resolve_connection_addresses()
 
@@ -153,7 +159,6 @@ class Postgresql(object):
         self.set_state('stopped')
         self._role_lock = Lock()
         self.set_role(self.get_postgres_role_from_data_directory())
-
         self._state_entry_timestamp = None
 
         self._cluster_info_state = {}
@@ -359,6 +364,11 @@ class Postgresql(object):
         self._pending_restart = pending_restart
         self._server_parameters = server_parameters
         self._connect_address = config.get('connect_address')
+        self._krbsrvname = config.get('krbsrvname', None)
+
+        # for not so obvious connection attempts that may happen outside of pyscopg2
+        if self._krbsrvname:
+            os.environ['PGKRBSRVNAME'] = self._krbsrvname
 
         if not local_connection_address_changed:
             self.resolve_connection_addresses()
@@ -1179,7 +1189,17 @@ class Postgresql(object):
             return None
         r = member.conn_kwargs(self._replication)
         r.update({'application_name': self.name, 'sslmode': 'prefer', 'sslcompression': '1'})
-        keywords = 'user password host port sslmode sslcompression application_name'.split()
+        keywords = 'host port sslmode sslcompression application_name'.split()
+
+        # Only connect with user password if password is supplied in config
+        if r.get('password', None):
+            keywords.extend(['user', 'password'])
+
+        # For kerberos we have to provide the spn
+        if self._krbsrvname:
+            r.update({'krbsrvname': self._krbsrvname})
+            keywords.append('krbsrvname')
+
         return ' '.join('{0}={{{0}}}'.format(kw) for kw in keywords).format(**r)
 
     def check_recovery_conf(self, member):
@@ -1254,10 +1274,15 @@ class Postgresql(object):
 
     @contextmanager
     def _get_replication_connection_cursor(self, host='localhost', port=5432, database=None, **kwargs):
-        with self._get_connection_cursor(host=host, port=int(port), database=database or self._database, replication=1,
-                                         user=self._replication['username'], password=self._replication['password'],
-                                         connect_timeout=3, options='-c statement_timeout=2000') as cur:
-            yield cur
+        if self._replication.get('username', None) and self._replication.get('password', None):
+            with self._get_connection_cursor(host=host, port=int(port), database=database or self._database, replication=1,
+                                             user=self._replication['username'], password=self._replication['password'],
+                                             connect_timeout=3, options='-c statement_timeout=2000') as cur:
+                yield cur
+        else:
+            with self._get_connection_cursor(host=host, port=int(port), database=database or self._database, replication=1,
+                                             connect_timeout=3, options='-c statement_timeout=2000') as cur:
+                yield cur
 
     def check_leader_is_not_in_recovery(self, **kwargs):
         try:
@@ -1529,7 +1554,9 @@ class Postgresql(object):
         if 'NOLOGIN' not in options and 'LOGIN' not in options:
             options.append('LOGIN')
 
-        self.query("""DO $$
+        # options will always be set but user and password may not
+        if name and password:
+            self.query("""DO $$
 BEGIN
     SET local synchronous_commit = 'local';
     PERFORM * FROM pg_authid WHERE rolname = %s;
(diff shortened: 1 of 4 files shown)


patroni/postgresql.py
```diff
@@ -1688,10 +1728,10 @@ def post_bootstrap(self, config, task):
             task.complete(self.run_bootstrap_post_init(config))
             if task.result:
                 self.create_or_update_role(self._replication['username'],
-                                           self._replication['password'], ['REPLICATION'])
+                                           self._replication.get('password', None), ['REPLICATION'])
```
CyberDem0n: `None` in the `dict.get` is redundant

postgres0.yml
```diff
@@ -35,6 +35,9 @@ bootstrap:
       #primary_slot_name: patroni
     postgresql:
       use_pg_rewind: true
+      # Fully qualified kerberos ticket file for the running user
```
CyberDem0n: It should be under the parameters. And I think it would be better to move it off `bootstrap.dcs` to the `postgresql` section (it is highly unlikely but filename could different from host to host).

patroni/postgresql.py
```diff
@@ -1540,6 +1567,19 @@ def create_or_update_role(self, name, password, options):
     END IF;
 END;
 $$""".format(name, ' '.join(options)), name, password, password)
+        else:
+            self.query("""DO $$
```
CyberDem0n: Mostly this is a duplicated code. We can do something like:
```python
params = [name]
if password:
    options.extend(['PASSWORD', '%s'])
    params.extend([password, password])
sql = """DO $$
BEGIN
    SET local synchronous_commit = 'local';
    PERFORM * FROM pg_authid WHERE rolname = %s;
    IF FOUND THEN
        ALTER ROLE "{0}" WITH {1};
    ELSE
        CREATE ROLE "{0}" WITH {1};
    END IF;
END;
$$""".format(name, ' '.join(options))
self.query(sql, *params)
```

patroni/postgresql.py
```diff
@@ -1254,10 +1274,15 @@ def _get_connection_cursor(**kwargs):
 
     @contextmanager
     def _get_replication_connection_cursor(self, host='localhost', port=5432, database=None, **kwargs):
-        with self._get_connection_cursor(host=host, port=int(port), database=database or self._database, replication=1,
-                                         user=self._replication['username'], password=self._replication['password'],
-                                         connect_timeout=3, options='-c statement_timeout=2000') as cur:
-            yield cur
+        if self._replication.get('username', None) and self._replication.get('password', None):
+            with self._get_connection_cursor(host=host, port=int(port), database=database or self._database, replication=1,
```
CyberDem0n: Here we can leave mostly the original code. The only adjustment needed is:
```python
password=self._replication.get('password')
```

If the password is not present `psycopg2.connect` will receive `password=None` and basically ignore it:
```python
c = psycopg2.connect(host='localhost', user='postgres', dbname='postgres', password=None)
```

The above line works without any problems.

patroni/postgresql.py
```diff
@@ -1179,7 +1189,17 @@ def primary_conninfo(self, member):
             return None
         r = member.conn_kwargs(self._replication)
         r.update({'application_name': self.name, 'sslmode': 'prefer', 'sslcompression': '1'})
-        keywords = 'user password host port sslmode sslcompression application_name'.split()
+        keywords = 'host port sslmode sslcompression application_name'.split()
```
CyberDem0n: I don't like the ordering change of parameters. We also have to specify the username if it is present.
Basically I would suggest doing something like:
```python
        r.update(application_name=self.name, sslmode='prefer', sslcompression='1', krbsrvname=self._krbsrvname)
        keywords = 'user password host port sslmode sslcompression application_name krbsrvname'.split()
        return ' '.join('{0}={{{0}}}'.format(kw) for kw in keywords if r.get(kw)).format(**r)
```

patroni/postgresql.py
```diff
@@ -113,6 +113,12 @@ def __init__(self, config):
         self._configure_server_parameters()
 
         self._connect_address = config.get('connect_address')
+        self._krbsrvname = config.get('krbsrvname', None)
```
CyberDem0n: `None` is redundant here.

patroni/postgresql.py
```diff
@@ -359,6 +364,11 @@ def reload_config(self, config):
         self._pending_restart = pending_restart
         self._server_parameters = server_parameters
         self._connect_address = config.get('connect_address')
+        self._krbsrvname = config.get('krbsrvname', None)
```
CyberDem0n: `None` is redundant here

vilajit:
Let me show you my setup. Unfortunately at work I am unable to access github or external email.
Can I call you to discuss the details?

vilajit:
Ok, was able to work on it this AM> incoprporated your suggestions and also added a sample pg_hba.conf entry to show the gss usage. 

You don’t need pg_ident if you don’t care about realm mapping and that is what I do. In my case,  all users are unique across realm so I don’t have to create users with realms or map that in postgres.
More over,  if the  Kerberos user itself is running the postgres process, it doesn’t  have to use -U as  gss will get that from the ticketfile. This is what I do in our env. Please let me know if you have more suggestions.
BTW, please review the second pull request.

Sent from Mail for Windows 10

CyberDem0n:
> if the Kerberos user itself is running the postgres process, it doesn’t have to use -U as gss will get that from the ticketfile.

Not really, libpq gets the username from the OS, not from the ticketfile. It could happen that OS username doesn't match with username in the ticketfile and postgres will refuse such connection with the error: 'provided user name (_postgres_) and authenticated user name (_postgres/myrealm.example@MYREALM.EXAMPLE_) do not match'.
In the above example I was connecting to postgres without specifying `-U` and `postgres` is my current OS user name. `postgres/myrealm.example@MYREALM.EXAMPLE` comes from the tokenfile.
If I specify `-U postgres/myrealm.example@MYREALM.EXAMPLE` it works fine (when the corresponding user exists in the cluster).

In your case OS user name apparently matches with username from tockenfile and therefore it works.
BUT! When you omit username from the `primary_conninfo` it will use superuser for replication connections, what is undesirable.

Basically our task is to make it possible to connect with two different users using the same ticket and it is not possible without touching pg_ident.conf.

vilajit:
I see you point.  Given that patroni has to have both the superuser and replication user (being different in  ideal configuration),  patroni user will have to impersonate the replication user  somehow.

In my case, both these users are so called “proids” and  configured as non-login-able. Meaning it can only run on the server host as a non-supervised user. I have Kerberos tools(privileged access) that I use to impersonate them.

BTW, do you see an issue with both the super and replication user being the same in such a scenario?


Sent from Mail for Windows 10

CyberDem0n:
> Given that patroni has to have both the superuser and replication user (being different in ideal configuration), patroni user will have to impersonate the replication user somehow.

Yep. That was exactly my point.

> do you see an issue with both the super and replication user being the same in such a scenario?

Do you mean that they will be using the same ticket? It is not a very big issue IMHO. At least it is a bit better than using superuser for replication connections.
Actually, nobody prevents you from supplying the same `username` in the `postgresql.superuser` and `postgresql.replication`.

vilajit:
Ok thanks.
Unfortunately, we have our own DNS infra and cannot use your haproxy.  Can any of the existing process in the patroni suite act as a simple half/full tcp pinger?

Our DNS VIP manager supports TCP ping check and can reroute  if a given leg is not pinging. 

Sent from Mail for Windows 10

CyberDem0n:
Is it just checking that some specific port is open or could do something more sophisticated?

vilajit:
A simple port check will do.

Sent from Mail for Windows 10

CyberDem0n:
Right now it is not supported. Shouldn't be hard to implement though.
BTW, one could run HAProxy on the same host with Patroni.

avaczi:
:+1:

CyberDem0n:
:+1:

vilajit:
Alexander,
 We had an edge case with db2 replication that impacted production. At Morgan we are actively working on patroni based replication for our prod rollout and I would really appreciate your feedback on how patroni would have handled this.

We had a scheduled powerdown on one data center and that brought standby down.

Due to n/w overwhelming, master side was not able to talk to the zookeeper (timeout).

So the most logical db2 decision was to terminate the master to avoid split brain.

So both sides were down for access.

If my understanding is right, patroni will flip the master to read only if this happened?

Thank you for your anticipated feedback.

Ajith Vilas

CyberDem0n:
Hi @vilajit,

yes, if Patroni can't talk to zookeeper it restarts master in read-only mode.

Could you please share patroni zookeeper configuration?
How many nodes are listed there?

vilajit:
We will be pairing two at a time. But there will be hundreds of such pairs once this is in full swing.
As for the patroni configuration I am still working on the optimal settings and will share later.
These are going to be cross data centers so if you have any suggestions, please let me know.

vilajit:
Hi Alex, I pushed another commit “zookeeper-kerberos-support for kerberized connection to zookeeper. Please review.
One thing that is pending is to set the proper acl for the nodes that patroni creates (it is currently world r/w/c/d/a) I’ll work on that  if you are ok with this change.

Sent from Mail for Windows 10

Issue #1021, opened 2019-04-04
The release of psycopg2 2.8 breaks pip install of patroni


ericjperry:
With the release of [psycopg2 2.8](https://github.com/psycopg/psycopg2/releases/tag/2_8) the package has been split. If the non binary psycopg2 package is installed, it will attempt to build from source which requires additional dependencies.

Running a command like the following:

```
pip3 --no-cache-dir install patroni[etcd,kubernetes]==1.5.5
```

is broken with

```
Collecting psycopg2>=2.5.4 (from patroni[etcd,kubernetes]==1.5.5)
  Downloading https://files.pythonhosted.org/packages/c7/ca/75236b17f1b951950ffc55d657c5aa408d3d0327a1b6c4c0f7cb16ef7e7b/psycopg2-2.8.tar.gz (367kB)
    Complete output from command python setup.py egg_info:
    running egg_info
    creating pip-egg-info/psycopg2.egg-info
    writing pip-egg-info/psycopg2.egg-info/PKG-INFO
    writing dependency_links to pip-egg-info/psycopg2.egg-info/dependency_links.txt
    writing top-level names to pip-egg-info/psycopg2.egg-info/top_level.txt
    writing manifest file 'pip-egg-info/psycopg2.egg-info/SOURCES.txt'
    
    Error: pg_config executable not found.
    
    pg_config is required to build psycopg2 from source.  Please add the directory
    containing pg_config to the $PATH or specify the full executable path with the
    option:
    
        python setup.py build_ext --pg-config /path/to/pg_config build ...
    
    or with the pg_config option in 'setup.cfg'.
    
    If you prefer to avoid building psycopg2 from source, please install the PyPI
    'psycopg2-binary' package instead.
    
    For further information please check the 'doc/src/install.rst' file (also at
    <http://initd.org/psycopg/docs/install.html>).
```

bersace:
cf. https://github.com/psycopg/psycopg2/issues/674 The warning has been there for long time. ;-)

You, as a user, needs to install build dependencies or use a distro package.

That would be nice if patroni allows to use psycopg2-binary. But there is no OR dependency in Python world.

CyberDem0n:
I know about the warning and I didn't believe till the last moment that they will do this stupid thing.
`psycopg2` and `psycopg2-binary` are two different packages which installed into the same place.
Guess what will happen if you install both of them and then remove one? Right, everything that depends on psycopg2 will break, because files are gone.

[...]

PR #1023, opened 2019-04-07
Remove psycopg2 from requirements

CyberDem0n:
Recently released psycopg2 split into two different packages, psycopg2, and psycopg2-binary which could be installed at the same time into the same place on the filesystem. In order to decrease dependency hell problem, we let a user choose how to install psycopg2. There are a few options available and it is reflected in the documentation.

This PR also changes the following behavior:
* `pip install patroni` will fail if psycopg2 is not installed
* Patroni will check psycopg2 upon start and fail if it can't be found or outdated.

Closes https://github.com/zalando/patroni/issues/1021

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -119,7 +119,7 @@ install:
             fi
             source ~/virtualenv/python${pv}/bin/activate
             # explicitly install all needed python modules to cache them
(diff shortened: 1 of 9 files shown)


avaczi:
:+1:

CyberDem0n:
:+1:

avaczi:
:+1:

PR #1024, opened 2019-04-07
scheduled_at needs a value for manual_failover

sharoonthomas:
The variable scheduled_at may be undefined for a manual
failover where there is an exception raised trying
to switchover with api


CyberDem0n:
Merged, thank you!

PR #1025, opened 2019-04-09
Don't set archive_mode to off during the custom bootstrap

CyberDem0n:
As @ants complained on Slack, it breaks the workflow when the same location is used for bootstrap and further archiving.
For the pg_upgrade we can achieve the same results by modifying postgres config in the upgrade script, which is part of spilo.


avaczi:
:+1:

avaczi:
:+1:

CyberDem0n:
:+1:

PR #1030, opened 2019-04-11
Reload Consul config on SIGHUP

ccakes:
This isn't necessarily complete, I just haven't written much Python and wanted to submit this so devs can point out early if I'm in the wrong place.

Aim is to reload the [Consul-specific config](https://patroni.readthedocs.io/en/latest/SETTINGS.html#consul) on SIGHUP.

I can't think of a realistic use case for changing the Consul server addr or DC as an on-the-fly thing. Seems to me like if you're changing that, it's probably bigger surgery than just regenerating config and sending a reload signal.

Right now this PR only addresses the `token` property but if the devs want me to cover any other uses cases, let me know. I'm also not sure if I need to call `super(Consul, self).reload_config(config)` or do something similar

### DIFF
--- a/patroni/dcs/consul.py
+++ b/patroni/dcs/consul.py
@@ -122,7 +122,7 @@ class ConsulClient(base.Consul):
     def __init__(self, *args, **kwargs):
         self._cert = kwargs.pop('cert', None)
         self._ca_cert = kwargs.pop('ca_cert', None)
-        self._token = kwargs.get('token')
+        self.token = kwargs.get('token')
         super(ConsulClient, self).__init__(*args, **kwargs)
 
     def connect(self, *args, **kwargs):
@@ -136,8 +136,9 @@ class ConsulClient(base.Consul):
         return HTTPClient(**kwargs)
 
     def set_token(self, token):
-        self._token = token
-        self._client.http.token = self._token
+        if token:
+            self.token = token
+            self.http.token = self.token
 
 def catch_consul_errors(func):
     def wrapper(*args, **kwargs):
@@ -236,9 +237,7 @@ class Consul(AbstractDCS):
                 time.sleep(5)
 
     def reload_config(self, config):
-        if config['token']:
-            self._client.set_token(config['token'])
-
+        self._client.set_token(config.get('token'))
         super(Consul, self).reload_config(config)
 
     def set_ttl(self, ttl):
--- a/patroni/dcs/consul.py
+++ b/patroni/dcs/consul.py
@@ -232,6 +232,9 @@ class Consul(AbstractDCS):
                 logger.info('waiting on consul')
                 time.sleep(5)
 
+    def reload_config(self, config):
+        self.token = config['token']
+
     def set_ttl(self, ttl):
         if self._client.http.set_ttl(ttl/2.0):  # Consul multiplies the TTL by 2x
             self._session = None
--- a/patroni/dcs/consul.py
+++ b/patroni/dcs/consul.py
@@ -237,7 +237,10 @@ class Consul(AbstractDCS):
                 time.sleep(5)
 
     def reload_config(self, config):
-        self._client.set_token(config.get('token'))
+        consul = config.get('consul')
+        if consul:
+            self._client.set_token(consul.get('token'))
+
         super(Consul, self).reload_config(config)
 
     def set_ttl(self, ttl):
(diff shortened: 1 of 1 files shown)


patroni/dcs/consul.py
```diff
@@ -232,6 +232,9 @@ def create_session(self):
                 logger.info('waiting on consul')
                 time.sleep(5)
 
+    def reload_config(self, config):
```
CyberDem0n: You still have to call the [parent](https://github.com/zalando/patroni/blob/master/patroni/dcs/__init__.py#L592) method

patroni/dcs/consul.py
```diff
@@ -232,6 +232,9 @@ def create_session(self):
...
+        self.token = config['token']
```
CyberDem0n: There is no `token` property in the `Consul` object, it is defined and used in the [ConsulClient](https://github.com/zalando/patroni/blob/master/patroni/dcs/consul.py#L125-L135) and [HTTPClient](https://github.com/zalando/patroni/blob/master/patroni/dcs/consul.py#L41).
Basically you have to implement `set_token` method in the `ConsulClient`, which will set the new value to `self._token` and take care about updating the token value in the existing `HTTPClient` object.

CyberDem0n: And last, but not least, `'token'` might be not defined in the `config`.

ccakes: Yep, stupid oversight sorry. I've done this and updated the PR

patroni/dcs/consul.py
```diff
@@ -135,6 +135,9 @@ def connect(self, *args, **kwargs):
             kwargs['token'] = self._token
         return HTTPClient(**kwargs)
 
+    def set_token(self, token):
+        self._token = token
```
CyberDem0n: `base.Consul` has `self.token`, which also must be updated. We created a duplicate `self._token` here because the connect method is called earlier than `self.token` is defined.

Or maybe it makes sense to rename `self._token` to `self.token`? IMHO it will not hurt.

patroni/dcs/consul.py
```diff
@@ -135,6 +135,9 @@ def connect(self, *args, **kwargs):
...
+        self._client.http.token = self._token
```
CyberDem0n: According to the https://github.com/cablehead/python-consul/blob/master/consul/base.py#L326 here should be `self.http.token = token`

CyberDem0n: `flake8` doesn't like your formatting: https://travis-ci.org/zalando/patroni/jobs/518699102#L1227

patroni/dcs/consul.py
```diff
@@ -135,6 +135,10 @@ def connect(self, *args, **kwargs):
             kwargs['token'] = self._token
         return HTTPClient(**kwargs)
 
+    def set_token(self, token):
+        if token:
```
CyberDem0n: We just have to set the value, without performing any checks.

patroni/dcs/consul.py
```diff
@@ -122,7 +122,7 @@ class ConsulClient(base.Consul):
     def __init__(self, *args, **kwargs):
         self._cert = kwargs.pop('cert', None)
         self._ca_cert = kwargs.pop('ca_cert', None)
-        self._token = kwargs.get('token')
+        self.token = kwargs.get('token')
```
CyberDem0n: If we rename it here, it also should be renamed everywhere: https://github.com/zalando/patroni/pull/1030/files#diff-01f2f2d2bffe1407b2928d1a79e456b0R135

ccakes: Ah sorry - didn't mean to push yet. Will do 👍

patroni/dcs/consul.py
ccakes: Should be done

patroni/dcs/consul.py
ccakes: Fixed both

patroni/dcs/consul.py
ccakes: Understood & fixed

avaczi:
:+1:

CyberDem0n:
:+1:

CyberDem0n:
:+1:

avaczi:
:+1:

CyberDem0n:
Merged, thank you!

PR #1035, opened 2019-04-12
Make it possible to run pg_rewind without superuser on pg11+

CyberDem0n:
* expose the current patroni version in DCS
* expose `checkpoint_after_promote` flag in DCS as an indicator that pg_rewind could be safely executed
* other nodes will wait until this flag is set instead of connecting as superuser and issuing the CHECKPOINT
* define `postgresql.authention.rewind` with credentials for pg_rewind in patroni configuration files.
* create user for pg_rewind if postgres is 11+
* grant execute on functions required for pg_rewind to rewind user


avaczi:
:+1:

avaczi:
:+1:

CyberDem0n:
:+1:

PR #1036, opened 2019-04-12
Don't wait until the previous callback finish is kill failed

CyberDem0n:
Such wait was happening in the main thread and blocking HA loop.


avaczi:
:+1:

avaczi:
:+1:

CyberDem0n:
:+1:

PR #1037, opened 2019-04-12
Manage pg_ident.conf with Patroni

CyberDem0n:
This functionality works similarly to the `pg_hba`:
If the `postgresql.pg_ident` is defined in the config file or DCS, Patroni will write its value to pg_ident.conf, however, if `postgresql.parameters.ident_file` is defined, Patroni will assume that pg_ident is managed from outside and not update the file.


erthalion:
:+1:

CyberDem0n:
:+1:

PR #1038, opened 2019-04-13
Add read-only and read-write API routes

jouir:
Patroni is great behind a load balancer like HAProxy. API routes are handy to direct the trafic to a valid node depending on its role. We could have one route for writes directed to the primary node and one route for reads directed to the replicas. In a two nodes setup, when we loose a node, there's one host left: the primary. Then, the read traffic will be dropped, even if the primary can handle it. This commit adds read-only and read-write routes. Read-only route enables reads on the primary. Read-write route is an alias for '/master', '/primary' or '/leader' route.

### DIFF
--- a/patroni/api.py
+++ b/patroni/api.py
@@ -85,19 +85,22 @@ class RestApiHandler(BaseHTTPRequestHandler):
         patroni = self.server.patroni
         cluster = patroni.dcs.cluster
 
+        if not cluster and patroni.ha.is_paused():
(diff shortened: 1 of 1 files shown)


Jan-M:
I like this, def. something also to keep in mind for our K8S service selector.

CyberDem0n:
:+1:

Jan-M:
:+1:

CyberDem0n:
Merged, thank you.

PR #1039, opened 2019-04-15
Use SSLContext to wrap REST API socket

jouir:
Security is important and also a requirement for some environments like PCI-DSS for example. This commit adds a way to support an explicit list of SSL/TLS protocols. It also adds a way to support OpenSSL ciphers.

### DIFF
--- a/patroni/api.py
+++ b/patroni/api.py
@@ -542,7 +542,9 @@ class RestApiServer(ThreadingMixIn, HTTPServer, Thread):
         # Sometime it's also needed to pass reference to a 'keyfile'.
         if self.__ssl_options.get('certfile'):
             import ssl
-            self.socket = ssl.wrap_socket(self.socket, server_side=True, **self.__ssl_options)
+            ctx = ssl.create_default_context(ssl.Purpose.CLIENT_AUTH)
+            ctx.load_cert_chain(**self.__ssl_options)
+            self.socket = ctx.wrap_socket(self.socket, server_side=True)
             self.__protocol = 'https'
         return True
 
--- a/tests/test_api.py
+++ b/tests/test_api.py
@@ -141,7 +141,8 @@ class MockRestApiServer(RestApiServer):
         Handler(MockRequest(request), ('0.0.0.0', 8080), self)
 
 
-@patch('ssl.wrap_socket', Mock(return_value=0))
+@patch('ssl.SSLContext.load_cert_chain', Mock())
+@patch('ssl.SSLContext.wrap_socket', Mock(return_value=0))
 @patch.object(BaseHTTPServer.HTTPServer, '__init__', Mock())
 class TestRestApiHandler(unittest.TestCase):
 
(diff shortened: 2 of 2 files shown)


patroni/api.py
```diff
@@ -512,7 +512,21 @@ def check_auth_header(self, auth_header):
...
+                'TLSv1.0': ssl.PROTOCOL_TLSv1,
+                'TLSv1.1': ssl.PROTOCOL_TLSv1_1,
+                'TLSv1.2': ssl.PROTOCOL_TLSv1_2
+            }
+            for version in config['ssl_versions']:
+                if version in protocols:
+                    if 'ssl_version' not in options:
+                        options['ssl_version'] = protocols[version]
+                    else:
+                        options['ssl_version'] |= protocols[version]
```
CyberDem0n: There are a couple of issues with `ssl_version`
1. First of all you can't use bitwise or operator to combine multiple versions
2. The [documentation](https://docs.python.org/2/library/ssl.html#ssl.PROTOCOL_TLSv1) tells that usage of `ssl.PROTOCOL*` is deprecated:
```
Deprecated since version 2.7.13: OpenSSL has deprecated all version specific protocols.
Use the default protocol with flags like OP_NO_SSLv3 instead.
```

jouir: Thanks for your feedback. To follow python recommendations, I've replaced the direct call to `ssl.wrap_socket`, which is deprecated, with an `SSLContext` with secure settings then a call to `SSLContext.wrap_socket`. The REST API is now secure by default.

jouir: But now, tests are failing because `ssl.wrap_socket` was mocked. Do I have to provide tests for the context or can we update the patch decorator? In both cases, I need your help, please. I'm not that comfortable with python tests.

patroni/api.py
```diff
@@ -542,7 +543,9 @@ def __initialize(self, config):
         # Sometime it's also needed to pass reference to a 'keyfile'.
         if self.__ssl_options.get('certfile'):
             import ssl
-            self.socket = ssl.wrap_socket(self.socket, server_side=True, **self.__ssl_options)
+            ctx = ssl.create_default_context(ssl.Purpose.CLIENT_AUTH)
+            ctx.load_cert_chain(certfile=self.__ssl_options['certfile'], keyfile=self.__ssl_options.get('keyfile'))
```
CyberDem0n: If you remove `ciphers` from line 515 you can change this line to:
```python
ctx.load_cert_chain(**self.__ssl_options)
```

If you still want to explicitly specify `ciphers` just get the value from the `config`.

patroni/api.py
```diff
@@ -512,7 +512,8 @@ def check_auth_header(self, auth_header):
 
     @staticmethod
     def __get_ssl_options(config):
-        return {option: config[option] for option in ['certfile', 'keyfile'] if option in config}
+        return {option: config[option] for option in ['certfile', 'keyfile', 'ciphers'] if option in config}
```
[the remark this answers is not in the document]
jouir: By default, the SSL context is secure so no need to set ciphers for now. It's now removed. If someone really need this feature, I can still develop it later.

patroni/api.py
jouir: Updated. Thanks!

CyberDem0n:
:+1:

erthalion:
:+1:

CyberDem0n:
Merged, thank you!

PR #1052, opened 2019-05-07
Make pyscopg2 version parsing more robust

CyberDem0n:
non-released version can have unparsable suffixes, for example 2.8.3.dev0


avaczi:
:+1:

CyberDem0n:
:+1:

Issue #1054, opened 2019-05-08
Failover doesn't occur when consul node under the leader is stopped


maksm90:
Hello everyone!
My problem is similar to #251 but for synchronous replication.

Since patroni bot can communicate with single consul node the autofailover process is sensible to heath of consul node with which patroni leader interacts.

My case is following:

- cluster consists of three nodes

- on each node the consul server agent is deployed

- on two nodes patroni bots (and PostgreSQL instances) are started and communicate with local consul agents

- synchronous replication is established (`synchronous_mode_strict: off`).

After disabling consul agent under the leader patroni bot this one demotes itself (flip to read-only mode) but another patroni node cannot be promoted because it lags in wal position from master - the message **INFO: Wal position of <master_node> is ahead of my wal position** appears in log.

I think this issue is bug because the patroni bot isolated from DCS participates in leader election procedure (shares own wal position when it is not neccessary) and doesn't allow other synchronous replica to promote. 

Patroni version is 1.5.6.

CyberDem0n:
The former master doesn't really participate in the leader race, but it still responds on REST API requests.
For the asynchronous replication,  that would be correct behavior. The node disqualifies itself from the leader race if it knows that some other node is ahead in WAL position.

Synchronous mode is a different beast. We still want to make sure that the old master is not running as a master anymore, but the synchronous replica can safely ignore WAL position check, because we know for sure that it has all **transactions** which were reported to the client as committed.

@ants what do you think?

ants:
To be more accurate, the sync standby will have all transactions that the client has requested sync replication for. But indeed it seems like a good idea to always allow promotion when we know we are in sync. I was not brave enough to go that far with the initial patch, but in hindsight I can't imagine a situation where this safeguard would save us and we have a (semi) realistic scenario where it prevents a valid failover.

Something to think about - maybe we want to do the same thing in async mode too, if we know that the node that is ahead doesn't have access to the DCS. The desired availability/durability tradeoff is less clear in async mode.

chobostar:
> maybe we want to do the same thing in async mode too, if we know that the node that is ahead doesn't have access to the DCS. The desired availability/durability tradeoff is less clear in async mode.

Sounds like a expected behavior for me, exceptionally if it's under `maximum_lag_on_failover`. 

I met the same issue in async mode, Patroni stucks in loop with `INFO: Wal position of <old_master> is ahead of my wal position` and failover doesn't occur

CyberDem0n:
[...]

PR #1059, opened 2019-05-13
Consider sync node as a healthy even when the former leader is ahead

CyberDem0n:
Fixes https://github.com/zalando/patroni/issues/1054


CyberDem0n:
:+1:

PR #1061, opened 2019-05-15
Make compare_values compatible with pg12

CyberDem0n:
* integer variables could be specified as floats: work_mem = '30.1GB'
* memory-based units could be in B and MB
* time-based variables now allow microseconds (us)
* vacuum_cost_delay and autovacuum_vacuum_cost_delay converted to real with base unit ms


CyberDem0n:
:+1:

Issue #1073, opened 2019-05-26
The demo fails (docker-compose up -d) due to failure in parsing postgres0.yml


shaharglazner:
**What**

Running `docker-compose up -d` failing due to syntax error in `postgres0.yml`

**Expected behavior**
7 containers are up & running

- 3 etcd containers
- 3 patroni containers
- 1 HAProxy container 

**Real behavior**
4 containers are up & running

- 3 etcd containers
- 1 HAProxy container

** Working with the previous postgres0.yml works **

**The error**
```
Process Process-1:
Traceback (most recent call last):
  File "/usr/lib/python3.5/multiprocessing/process.py", line 249, in _bootstrap
    self.run()
  File "/usr/lib/python3.5/multiprocessing/process.py", line 93, in run
    self._target(*self._args, **self._kwargs)
  File "/patroni/__init__.py", line 157, in patroni_main
    patroni = Patroni()
  File "/patroni/__init__.py", line 26, in __init__
    self.config = Config()
  File "/patroni/config.py", line 78, in __init__
    self._local_configuration = self._load_config_file()
  File "/patroni/config.py", line 108, in _load_config_file
    config = yaml.safe_load(f)
  File "/usr/lib/python3/dist-packages/yaml/__init__.py", line 94, in safe_load
    return load(stream, SafeLoader)
  File "/usr/lib/python3/dist-packages/yaml/__init__.py", line 72, in load
    return loader.get_single_data()
  File "/usr/lib/python3/dist-packages/yaml/constructor.py", line 35, in get_single_data
    node = self.get_single_node()
  File "/usr/lib/python3/dist-packages/yaml/composer.py", line 36, in get_single_node
    document = self.compose_document()
  File "/usr/lib/python3/dist-packages/yaml/composer.py", line 55, in compose_document
    node = self.compose_node(None, None)
  File "/usr/lib/python3/dist-packages/yaml/composer.py", line 84, in compose_node
    node = self.compose_mapping_node(anchor)
  File "/usr/lib/python3/dist-packages/yaml/composer.py", line 133, in compose_mapping_node
    item_value = self.compose_node(node, item_key)
  File "/usr/lib/python3/dist-packages/yaml/composer.py", line 84, in compose_node
    node = self.compose_mapping_node(anchor)
  File "/usr/lib/python3/dist-packages/yaml/composer.py", line 127, in compose_mapping_node
    while not self.check_event(MappingEndEvent):
[...]

PR #1074, opened 2019-05-26
Fix some small issues in the Dockerfile

CyberDem0n:
* symlink ~/.config/patroni/patronictl.yaml
* comment out rewind section

Fixes https://github.com/zalando/patroni/issues/1073


CyberDem0n:
:+1:

Issue #119, opened 2016-01-19
Create a health check that returns 200 if and only if PostgreSQL is running


alexeyklyukin:
If you put an ELB that incorporates all nodes to scale up reads that you want this health check.

feikesteenbergen:
We currently have:

* /replica
* /master
* /patroni

Should we add the following?

* /postgres

alexeyklyukin:
Another idea would be having a /replica/running and /master/running in order to return only the nodes that actually run PostgreSQL.

LappleApple:
@alexeyklyukin @feikesteenbergen @CyberDem0n Is this something we can open up as a "Help Wanted" task?

alexeyklyukin:
yes

antineutron:
I might be able to pick this up.

henrywangx:
Hi, actually we want the rest API to check running node.
If we redirect all the reading connetction to replica.
There is a situation that when all the replica is down, haproxy can't redirect the reading connection to master.

jberkus:
@alexeyklyukin  is this still a valid task worth having a contributor pick it up?

PR #1079, opened 2019-06-03
Add API route /health (#119)

wilfriedroset:
According to the documented feature I don't think it is necessary: https://github.com/zalando/patroni/blob/master/features/patroni_api.feature#L7

We can simply GET http://127.0.0.1:8008/ and we will receive a return code 200 if the state of postgresql == running.

If feel like it is more explicit to have an API route /health.

### DIFF
--- a/features/patroni_api.feature
+++ b/features/patroni_api.feature
@@ -8,6 +8,8 @@ Scenario: check API requests on a stand-alone server
 	Then I receive a response code 200
 	And I receive a response state running
 	And I receive a response role master
+	When I issue a GET request to http://127.0.0.1:8008/health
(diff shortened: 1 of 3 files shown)


CyberDem0n:
`features/*.feature` don't contain documentation, but behave (integration) tests.
`GET /` is absolutely the same as `GET /master`, it will return 200 only when postgres is running as a leader, so your change makes total sense.

CyberDem0n:
:+1:

Jan-M:
:+1:

CyberDem0n:
Merged, thank you @wilfriedroset

PR #1080, opened 2019-06-06
Implement two-step logging

CyberDem0n:
A few times we observed that Patroni HA loop was blocked for a few minutes due to not being able to write logs to stderr. This is a very rare condition which we hit so far only on k8s. This commit makes Patroni resilient to such kind of problems. All log messages first are written into the in-memory queue and later they are asynchronously flushed into the stderr or file from a separate thread.

The maximum queue size is configurable and the default value is 1000. This should be enough to keep more than one hour of log messages with default settings and when Patroni cluster operates normally (without big issues).

In case if we hit the maximum size of the queue further logs will be discarded until the queue size will be reduced. The number of discarded messages will be reported into the log later.

In addition to that, the number of non-flushed and discarded messages (if there are any), will be reported via Patroni REST API as:
```json
"logger_queue_size": X,
"logger_records_lost": Y`
```

### DIFF
--- a/patroni/api.py
+++ b/patroni/api.py
@@ -75,6 +75,12 @@ class RestApiHandler(BaseHTTPRequestHandler):
             response['watchdog_failed'] = True
         if patroni.ha.is_paused():
             response['pause'] = True
+        qsize = patroni.logger.queue_size
+        if qsize > 2:
+            response['logger_queue_size'] = qsize
+            lost = patroni.logger.records_lost
+            if lost:
+                response['logger_records_lost'] = lost
         self._write_json_response(status_code, response)
 
     def do_GET(self, write_status_code_only=False):
--- a/patroni/api.py
+++ b/patroni/api.py
@@ -76,7 +76,7 @@ class RestApiHandler(BaseHTTPRequestHandler):
         if patroni.ha.is_paused():
             response['pause'] = True
         qsize = patroni.logger.queue_size
-        if qsize > 2:
+        if qsize > patroni.logger.NORMAL_LOG_QUEUE_SIZE:
             response['logger_queue_size'] = qsize
             lost = patroni.logger.records_lost
             if lost:
--- a/patroni/config.py
+++ b/patroni/config.py
@@ -263,7 +263,7 @@ class Config(object):
             ret['restapi']['authentication'] = restapi_auth
 
         authentication = {}
-        for user_type in ('replication', 'superuser', 'rewind'):
+        for user_type in ('replication', 'superuser'):
             entry = _get_auth(user_type)
             if entry:
                 authentication[user_type] = entry
--- a/patroni/config.py
+++ b/patroni/config.py
@@ -232,7 +232,8 @@ class Config(object):
 
         _set_section_values('restapi', ['listen', 'connect_address', 'certfile', 'keyfile'])
         _set_section_values('postgresql', ['listen', 'connect_address', 'config_dir', 'data_dir', 'pgpass', 'bin_dir'])
-        _set_section_values('log', ['level', 'format', 'dateformat', 'dir', 'file_size', 'file_num', 'loggers'])
+        _set_section_values('log', ['level', 'format', 'dateformat', 'max_queue_size',
+                                    'dir', 'file_size', 'file_num', 'loggers'])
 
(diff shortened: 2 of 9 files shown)


patroni/log.py
```diff
@@ -4,63 +4,144 @@
...
+        record.msg = record.message
+        record.args = None
+        record.exc_info = None
+        self.queue.put_nowait(record)
+
+    def _try_to_report_lost_records(self):
+        if self._records_lost:
+            try:
+                record = _LOGGER.makeRecord(_LOGGER.name, logging.WARNING, __file__, 0,
+                                            'QueueHandler has lost %s log records',
```
sdudoladov: imo "logging queue" instead of `QueueHandler` is more understandable. Or do you want to explicitly point to the implementation ?

patroni/log.py
```diff
@@ -4,63 +4,144 @@
...
+
+            record = self._queue_handler.queue.get(True)
+            if record is None:
+                break
+
+            self._log_handler.handle(record)
+            self._queue_handler.queue.task_done()
+
+    def shutdown(self):
+        self._queue_handler.queue.put_nowait(None)
```
sdudoladov: so the exception `queue.Full` is not important during shutdown or ?

patroni/log.py
```diff
@@ -4,63 +4,144 @@
...
-        self.config = None
-        self.handler = None
+        super(PatroniLogger, self).__init__()
+        self.daemon = True
+        self._queue_handler = QueueHandler()
+        self._root_logger = logging.getLogger()
+        self._root_logger.addHandler(self._queue_handler)
+        self._config = None
+        self._log_handler = None
+        self._old_handlers = []
```
sdudoladov: are old log handlers closed separately to avoid extra locking in `reload_config`  or ?

patroni/api.py
```diff
@@ -75,6 +75,12 @@ def _write_status_response(self, status_code, response):
             response['watchdog_failed'] = True
         if patroni.ha.is_paused():
             response['pause'] = True
+        qsize = patroni.logger.queue_size
+        if qsize > 2:
```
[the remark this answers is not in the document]
CyberDem0n: Nope, every HA loop it usually writes into the log only two messages: 'Lock owner: %s; I am %s' and ('no action.  i am the leader with the lock' or 'no action.  i am a secondary and i am following a leader').

Therefore usually we don't expect to be more than 2 entries in the queue.

patroni/log.py
CyberDem0n: We can't handle it anyway.

patroni/log.py
CyberDem0n: Exactly

sdudoladov:
:+1:

patroni/log.py
```diff
@@ -4,63 +4,144 @@
...
         for name, level in loggers.items():
-            logger = self.root_logger.manager.getLogger(name)
+            logger = self._root_logger.manager.getLogger(name)
             logger.setLevel(level)
 
     def reload_config(self, config):
-        if self.config is None or not deep_compare(self.config, config):
-            self.root_logger.setLevel(config.get('level', PatroniLogger.DEFAULT_LEVEL))
+        if self._config is None or not deep_compare(self._config, config):
+            self._queue_handler.queue.maxsize = config.get('max_queue_size', 1000)
```
ants: Not sure if it matters in any way, but it seems prudent to acquire queue.mutex before mutating maxsize.

patroni/config.py
```diff
@@ -262,7 +263,7 @@ def _get_auth(name):
             ret['restapi']['authentication'] = restapi_auth
 
         authentication = {}
-        for user_type in ('replication', 'superuser'):
+        for user_type in ('replication', 'superuser', 'rewind'):
```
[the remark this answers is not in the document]
CyberDem0n: Yes, it was forgotten in the https://github.com/zalando/patroni/commit/bba9066315f8dda23e0a28909790a5a3d48d1842 and https://github.com/zalando/patroni/pull/1084 fixes it.

I will revert this change.

sdudoladov:
:+1:

CyberDem0n:
:+1:

PR #1081, opened 2019-06-06
log exceptions caught in Retry

danielkucera:
Log the final exception when either the number of attempts or the timeout were reached.

### DIFF
--- a/patroni/utils.py
+++ b/patroni/utils.py
@@ -312,17 +312,18 @@ class Retry(object):
                     self._cur_stoptime = time.time() + self.deadline
                 return func(*args, **kwargs)
             except self.retry_exceptions as e:
-                logger.warning('Retry got exception: %s', str(e))
                 # Note: max_tries == -1 means infinite tries.
                 if self._attempts == self.max_tries:
+                    logger.warning('Retry got exception: %s', e)
                     raise RetryFailedError("Too many retry attempts")
                 self._attempts += 1
                 sleeptime = self._cur_delay + (random.randint(0, self.max_jitter) / 100.0)
 
                 if self._cur_stoptime is not None and time.time() + sleeptime >= self._cur_stoptime:
+                    logger.warning('Retry got exception: %s', e)
                     raise RetryFailedError("Exceeded retry deadline")
-                else:
-                    self.sleep_func(sleeptime)
+                logger.debug('Retry got exception: %s', e)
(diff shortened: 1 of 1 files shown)


CyberDem0n:
Hi @danielkucera,

first of all any PR deserves at least a few words about what the problem is it trying to solve.
And the second, IMHO warning severity is a bit too much for expected exceptions. I am not arguing that having more logs is better than having none, but the severity should be adequate. In this specific case the _debug_ would be enough.

danielkucera:
Hi @CyberDem0n , sorry for missing explanation, I was posting in a hurry.
Retry catches all http exceptions so if you make a mistake in config or e.g. consul server suddenly disappears, you will only see an exception about exhausted retries or timeout. Nothing about the real reason - whethere there is a wrong cert, 503, 404 or whatever.
I think that warning is appropriate because you want to know what happened even if you are running in production with disabled debugging. Exceptions should not normally happen here IMHO.

CyberDem0n:
So far you are the first with such a problem, IMHO debug would be more appropriate.
It is possible to change the log level without restart, and moreover, one can set a different log level [per python module](https://patroni.readthedocs.io/en/latest/SETTINGS.html#log):

```yaml
log:
  level: INFO
  loggers:
    patroni.utils: DEBUG
```

patroni/utils.py
```diff
@@ -308,7 +311,8 @@ def __call__(self, func, *args, **kwargs):
                 if self.deadline is not None and self._cur_stoptime is None:
                     self._cur_stoptime = time.time() + self.deadline
                 return func(*args, **kwargs)
-            except self.retry_exceptions:
+            except self.retry_exceptions as e:
+                logger.warning('Retry got exception: %s', str(e))
```
CyberDem0n: `str(e)` us redundant here, just `e` will work absolutely the same.

ants:
I can echo the problem. Having the error output would have made debugging quicker and more approachable in multiple cases. However I was thinking more along the lines of recording the exception and only showing it when retry times out to avoid log spam and, more importantly, warnings about things that are somewhat normal.

danielkucera:
@CyberDem0n , please review the latest version. I tried to make a compromise between our views.

CyberDem0n:
:+1:

Jan-M:
:+1:

CyberDem0n:
Merged, thank you @danielkucera

PR #1084, opened 2019-06-12
Reverse logic around checkpoint_after_promote

CyberDem0n:
It will be set to false in the JSON only until the checkpoint actually happened.

The next improvement of https://github.com/zalando/patroni/commit/bba9066315f8dda23e0a28909790a5a3d48d1842

### DIFF
--- a/patroni/dcs/__init__.py
+++ b/patroni/dcs/__init__.py
@@ -248,7 +248,7 @@ class Leader(namedtuple('Leader', 'index,session,member')):
         if version:
             try:
                 if tuple(map(int, version.split('.'))) >= (1, 5, 6):
-                    return bool(self.member.data.get('checkpoint_after_promote'))
+                    return self.member.data['role'] == 'master' and 'checkpoint_after_promote' not in self.member.data
             except Exception:
                 logger.debug('Failed to parse Patroni version %s', version)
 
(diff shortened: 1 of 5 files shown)


patroni/dcs/__init__.py
```diff
@@ -248,7 +248,7 @@ def checkpoint_after_promote(self):
         if version:
             try:
                 if tuple(map(int, version.split('.'))) >= (1, 5, 6):
-                    return bool(self.member.data.get('checkpoint_after_promote'))
+                    return self.member.data['role'] == 'master' and 'checkpoint_after_promote' not in self.member.data
```
sdudoladov: 1. does  the flag `checkpoint_after_promote` mean everywhere in the codebase that the checkpoint in question is happening right now ?
2. can `self.member.data`  contain `{checkpoint_after_promote : False}`  ?

avaczi:
:+1:

patroni/dcs/__init__.py
CyberDem0n: > can self.member.data contain {checkpoint_after_promote : False} ?

In fact, it will contain only this value. When checkpoint already has happened checkpoint_after_promote will disappear from the json.

Basically replicas only need to know the fact that checkpoint after promotion on the current primary didn't happen yet in order to postpone pg_rewind.

sdudoladov:
:+1:

CyberDem0n:
:+1:

PR #1086, opened 2019-06-13
Small refactoring of postgresql/bootstrap

CyberDem0n:
the main purpose of this PR is simplifying #1068

It is mostly necessary for future support of pg12, where there will be no recovery.conf anymore, but `keep_existing_recovery_conf` parameter still needs to be supported due to backward compatibility.

### DIFF
--- a/patroni/postgresql/bootstrap.py
+++ b/patroni/postgresql/bootstrap.py
@@ -22,6 +22,10 @@ class Bootstrap(object):
     def running_custom_bootstrap(self):
         return self._running_custom_bootstrap
 
+    @property
+    def keep_existing_recovery_conf(self):
+        return self._running_custom_bootstrap and self._keep_existing_recovery_conf
+
     @staticmethod
     def process_user_options(tool, options, not_allowed_options, error_handler):
         user_options = []
@@ -60,7 +64,7 @@ class Bootstrap(object):
         def error_handler(e):
             raise Exception(e)
 
-        options = self.process_user_options('initdb', config.get('initdb') or [], not_allowed_options, error_handler)
+        options = self.process_user_options('initdb', config or [], not_allowed_options, error_handler)
         pwfile = None
 
         if self._postgresql.config.superuser:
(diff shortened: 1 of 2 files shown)


patroni/postgresql/bootstrap.py
```diff
@@ -76,17 +80,21 @@ def error_handler(e):
         ret = self._postgresql.pg_ctl('initdb', *options)
         if pwfile:
             os.remove(pwfile)
-        if not ret:
+        if ret:
```
sdudoladov: it's a refactoring PR anyway, so what about  renaming  `ret` to `initdb_has_succeded` ?

patroni/postgresql/bootstrap.py
```diff
@@ -280,22 +288,21 @@ def clone(self, clone_member):
...
-            self._postgresql.configure_server_parameters()
         return ret
 
     def bootstrap(self, config):
         """ Initialize a new node from scratch and start it. """
         pg_hba = config.get('pg_hba', [])
         method = config.get('method') or 'initdb'
-        self._running_custom_bootstrap = method != 'initdb' and method in config and 'command' in config[method]
-        if self._running_custom_bootstrap:
+        if method != 'initdb' and method in config and 'command' in config[method]:
```
sdudoladov: so if one of the required params for a custom bootstrap method is not specified we fall back to `initdb` ? [docs](https://github.com/zalando/patroni/blob/master/docs/replica_bootstrap.rst#bootstrap) are unclear on that part 
> Each bootstrap method must define at least a name and a command. A special initdb method is available to trigger the default behavior, in which case method parameter can be omitted altogether.

CyberDem0n: If there is no method defined we default to initdb.
If the method doesn't have a configuration section - default to initdb
If the configuration section doesn't have a command defined - default to initdb

sdudoladov:
:+1:

CyberDem0n:
:+1:

PR #1087, opened 2019-06-13
Simplify postgresql.reload_config (and fix some bugs)

CyberDem0n:
`compare_values` function is already smart enough to handle all tricky cases when the unit is not in `B` or `kB`, but also in blocks. Therefore we don't need to query `wal_segment_size` from pg_settings.

This PR is mostly necessary to simplify #1068

### DIFF
--- a/patroni/postgresql/config.py
+++ b/patroni/postgresql/config.py
@@ -302,24 +302,15 @@ class ConfigHandler(object):
             changes = CaseInsensitiveDict({p: v for p, v in server_parameters.items() if '.' not in p})
             changes.update({p: None for p in self._server_parameters.keys() if not ('.' in p or p in changes)})
             if changes:
-                if 'wal_segment_size' not in changes:
-                    changes['wal_segment_size'] = '16384kB'
                 # XXX: query can raise an exception
                 for r in self._postgresql.query(('SELECT name, setting, unit, vartype, context '
                                                  + 'FROM pg_catalog.pg_settings ' +
                                                  ' WHERE pg_catalog.lower(name) IN ('
                                                  + ', '.join(['%s'] * len(changes)) +
-                                                 ') ORDER BY 1 DESC'), *(k.lower() for k in changes.keys())):
-                    if r[4] == 'internal':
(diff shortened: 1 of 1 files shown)


patroni/postgresql/config.py
```diff
@@ -302,24 +302,15 @@ def reload_config(self, config):
...
-                    if r[4] == 'internal':
-                        if r[0] == 'wal_segment_size':
-                            server_parameters.pop(r[0], None)
-                            wal_segment_size = parse_int(r[2], 'kB')
-                            if wal_segment_size is not None:
-                                changes['wal_segment_size'] = '{0}kB'.format(int(r[1]) * wal_segment_size)
-                    elif r[0] in changes:
-                        unit = changes['wal_segment_size'] if r[0] in ('min_wal_size', 'max_wal_size') else r[2]
+                                                 ')'), *(k.lower() for k in changes.keys())):
+                    if r[4] != 'internal' and r[0] in changes:
```
sdudoladov: 1. what about giving it more descriptive names such as `context` instead of `r[4]` ?
2. does it make sense to add a specific warning for the case when one tries to change an `internal` setting ? the general warning `                        logger.warning('Removing invalid parameter `%s` from postgresql.parameters', param)
` does not explain why the parameter change is invalid

CyberDem0n: My intention was to fix a specific problem by touching as minimum code as possible, not refactor the whole method.

sdudoladov:
so `wal_segment_size` was used with Kb as units and not the default blocks as stated in the Postgres docs  to circumvent limitations in `parse_int` ? And does  parsing the values in blocks corresponds to [unitless examples in parse_int() ?](https://github.com/zalando/patroni/blob/638b560cf843506570fd257551e66cbbcd6e06ff/patroni/utils.py#L173)

sdudoladov:
:+1:

CyberDem0n:
The value of `wal_segment_size` in the pg_settings was in `8kb` blocks until postgres 11, now it is in `MB`

CyberDem0n:
:+1:

PR #1089, opened 2019-06-13
Do a smart comparison of actual and desired primary_conninfo

CyberDem0n:
On every replica Patroni periodically opens the recovery.conf file and checks that it contains the correct primary_conninfo.
So far the correctness check was pretty dumb, it was basically doing a full match of strings. That could lead to the restart of the replica when Patroni is "joining" already running postgres process.
Instead of using string comparison we parse the actual primary_conninfo value from the file and check that all parameters are matching with the desired value.

In addition to that we stop reading and parsing recovery.conf of every iteration if the modification time didn't change.

### DIFF
--- a/patroni/postgresql/config.py
+++ b/patroni/postgresql/config.py
@@ -6,17 +6,147 @@ import socket
 import stat
 
 from requests.structures import CaseInsensitiveDict
+from six.moves.urllib_parse import urlparse, parse_qsl, unquote
 
 from ..utils import compare_values, parse_bool, parse_int, split_host_port
 
 logger = logging.getLogger(__name__)
 
 SYNC_STANDBY_NAME_RE = re.compile(r'^[A-Za-z_][A-Za-z_0-9\$]*$')
+PARAMETER_RE = re.compile(r'([a-z_]+)\s*=\s*')
 
 
 def quote_ident(value):
     """Very simplified version of quote_ident"""
     return value if SYNC_STANDBY_NAME_RE.match(value) else '"' + value + '"'
+
+
+def conninfo_uri_parse(dsn):
+    ret = {}
+    r = urlparse(dsn)
+    if r.username:
+        ret['user'] = r.username
+    if r.password:
+        ret['password'] = r.password
+    if r.path[1:]:
+        ret['dbname'] = r.path[1:]
+    hosts = []
+    ports = []
+    for netloc in r.netloc.split('@')[-1].split(','):
+        host = port = None
+        if '[' in netloc and ']' in netloc:
+            host = netloc.split(']')[0][1:]
+        tmp = netloc.split(':', 1)
+        if host is None:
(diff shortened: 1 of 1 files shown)


sdudoladov:
parts of my review for #1068 apply here

patroni/postgresql/config.py
```diff
@@ -186,30 +317,82 @@ def replace_pg_ident(self):
...
+            return None
+        keywords = 'user password host port sslmode application_name krbsrvname'.split()
         return ' '.join('{0}={{{0}}}'.format(kw) for kw in keywords if r.get(kw)).format(**r)
 
     def recovery_conf_exists(self):
         return os.path.exists(self._recovery_conf)
 
-    def check_recovery_conf(self, member):
+    def _read_primary_conninfo(self):
         # TODO: recovery.conf could be stale, would be nice to detect that.
```
sdudoladov: The PR resolves this TODO, does't it ?

patroni/postgresql/config.py
```diff
@@ -186,30 +317,82 @@ def replace_pg_ident(self):
...
+            # - '' (not in config)
+            # - or the actual dsn value
+            self._primary_conninfo = primary_conninfo
+            if primary_conninfo:
+                # We will cache parsed value until the next config change.
+                self._primary_conninfo = parse_dsn(primary_conninfo)
+                # If we failed to parse non-empty connection string this indicates that config if broken.
+                if not self._primary_conninfo:
+                    return False
+                if self._primary_conninfo:
```
sdudoladov: what does successful parsing have to do with setting the default value ?

patroni/postgresql/config.py
```diff
@@ -6,19 +6,148 @@
...
+        i += end + int(is_quoted)
+        ret[param] = value
+    return ret
+
+
+def parse_dsn(value):
+    """
+    Very simple equivalent of `psycopg2.extensions.parse_dsn` introduced in 2.7.0.
+    We are not using psycopg2 function in order to remain compatible with 2.5.4+.
+    >>> r = parse_dsn('postgresql://u%2Fse:pass@:%2f123,[%2Fhost2]/db%2Fsdf?application_name=mya%2Fpp&ssl=true')
```
sdudoladov: are the examples here are picked up by `doctest`   ?

patroni/postgresql/config.py
CyberDem0n: Nope, one can overwrite the recovery.conf and don't restart postgres. There is no way to check the actual value of primary_conninfo before pg12.

patroni/postgresql/config.py
CyberDem0n: These are not examples but actual doctests

sdudoladov:
:+1:

CyberDem0n:
:+1:

Issue #532, opened 2017-09-29
default value for EDITOR env variable?


kostiantyn-nemchenko:
Hi, 

While playing with patronictl options and perusing `invoke_editor()` in `patroni/ctl.py`, I noticed that [`editor_cmd`](https://github.com/zalando/patroni/blob/32b0768631fcd1f8a75b69627e225a19bece4923/patroni/ctl.py#L954) variable doesn't have a default value.
I guess it will be more user-friendly to have a preset value rather than raising `PatroniCtlException` (e.g. Mercurial and IPython projects are using vi as default text editor under Linux/Unix)

Thank you.

ants:
I didn't set it because I was worried newbies would get stuck in vi.

It looks like on debian there is also `/usr/bin/editor` that can be set via `update-alternatives`. Maybe if `EDITOR` is not set we should try `editor` and then `vi`.

PR #1091, opened 2019-06-14
Add fallback value for editor_cmd (#532)

wilfriedroset:
While using patronictl and not having EDITOR environment variable would end up with PatroniCtlException. This can be avoid by having default value:  `/usr/bin/editor` or `vi`(which is common).

`/usr/bin/editor` does not exist on every of so I've added a check before defaulting to this value. Also  `vi` might not be installed where we expect it to be or simply not installed. I've added this check also.

Here is the documentation for [distutils.spawn](https://docs.python.org/2/distutils/apiref.html#module-distutils.spawn).

Here is an example of its behavior
```
# Python 2.7 on debian

$ docker run -it python:2.7 bash
root@fa2e4bf024b3:/# cat /etc/debian_version
9.9

root@fa2e4bf024b3:/# python
Python 2.7.16 (default, Jun 11 2019, 02:02:48)
[GCC 6.3.0 20170516] on linux2
>>> from distutils.spawn import find_executable
>>> find_executable('vi') is None
True

Let's install vim

root@fa2e4bf024b3:/# apt update && apt install -y vim
root@fa2e4bf024b3:/# python
Python 2.7.16 (default, Jun 11 2019, 02:02:48)
[GCC 6.3.0 20170516] on linux2
>>> from distutils.spawn import find_executable
>>> find_executable('vi')
'/usr/bin/vi'
>>> find_executable('vim')
'/usr/bin/vim'
>>>

# Python 2.7 on alpine

$ docker run -it python:2.7-alpine
Python 2.7.16 (default, May 11 2019, 02:15:00)
[GCC 8.3.0] on linux2
>>> from distutils.spawn import find_executable
>>> find_executable('vi')
'/usr/bin/vi'
>>> find_executable('vim')
>>> find_executable('vim') is None
True

# Python 2.7 on centos

$ docker run -it centos:7 bash
[root@23fe5a678cdf /]# python
Python 2.7.5 (default, Oct 30 2018, 23:45:53)
[GCC 4.8.5 20150623 (Red Hat 4.8.5-36)] on linux2
>>> from distutils.spawn import find_executable
>>> find_executable('vi')
'/usr/bin/vi'
>>>

$ docker run -it python:3-alpine
Python 3.7.3 (default, May 11 2019, 02:00:41)
[GCC 8.3.0] on linux
>>> from distutils.spawn import find_executable
>>> find_executable('vi')
'/usr/bin/vi'
>>> find_executable('vim') is None
True
```

Having `vi` as final fallback and can eventually leave developer locked-in.

### DIFF
--- a/patroni/ctl.py
+++ b/patroni/ctl.py
@@ -25,6 +25,7 @@ import yaml
 
 from click import ClickException
 from contextlib import contextmanager
+from distutils.spawn import find_executable
 from patroni.config import Config
 from patroni.dcs import get_dcs as _get_dcs
 from patroni.exceptions import PatroniException
@@ -1088,9 +1089,16 @@ def invoke_editor(before_editing, cluster_name):
     :param before_editing: human representation before editing
     :returns tuple of human readable and parsed datastructure after changes
     """
-    editor_cmd = os.environ.get('EDITOR')
+    if 'EDITOR' in os.environ:
+        editor_cmd = os.environ.get('EDITOR')
+    else:
+        for editor in ('editor', 'vi'):
+            editor_cmd = find_executable(editor)
+            if editor_cmd:
+                logging.debug('Setting fallback editor_cmd=%s', editor)
+                break
     if not editor_cmd:
-        raise PatroniCtlException('EDITOR environment variable is not set')
+        raise PatroniCtlException('EDITOR environment variable is not set. editor or vi are not available')
 
     with temporary_file(contents=before_editing.encode('utf-8'),
                         suffix='.yaml',


kostiantyn-nemchenko:
Fixes #532

patroni/ctl.py
```diff
@@ -1088,9 +1089,18 @@ def invoke_editor(before_editing, cluster_name):
     :param before_editing: human representation before editing
     :returns tuple of human readable and parsed datastructure after changes
     """
-    editor_cmd = os.environ.get('EDITOR')
+    if 'EDITOR' in os.environ:
+        editor_cmd = os.environ.get('EDITOR')
+    elif os.path.exists('/usr/bin/editor'):
```
CyberDem0n: I think it is better to avoid assumptions about the location of binary files. Why not use `find_executable('editor')` here?
We can also make a loop here:
```python
for editor in ('editor', 'vi'):
    editor_cmd = find_executable(editor)
    if editor_cmd:
        logging.debug('Setting fallback editor_cmd=%s', editor_cmd)
        break
```

patroni/ctl.py
```diff
@@ -1088,9 +1089,16 @@ def invoke_editor(before_editing, cluster_name):
...
     :returns tuple of human readable and parsed datastructure after changes
     """
-    editor_cmd = os.environ.get('EDITOR')
+    if 'EDITOR' in os.environ:
+        editor_cmd = os.environ.get('EDITOR')
+    else:
+        for editor in ('editor', 'vi'):
+            editor_cmd = find_executable(editor)
+            if editor_cmd:
+                logging.debug('Setting fallback editor_cmd=%s' % editor)
```
CyberDem0n:
```python
-                 logging.debug('Setting fallback editor_cmd=%s' % editor)
+                 logging.debug('Setting fallback editor_cmd=%s', editor)
```

patroni/ctl.py
```diff
@@ -1088,9 +1089,16 @@ def invoke_editor(before_editing, cluster_name):
...
+        editor_cmd = os.environ.get('EDITOR')
+    else:
+        for editor in ('editor', 'vi'):
+            editor_cmd = find_executable(editor)
+            if editor_cmd:
+                logging.debug('Setting fallback editor_cmd=%s', editor)
+                break
     if not editor_cmd:
-        raise PatroniCtlException('EDITOR environment variable is not set')
+        raise PatroniCtlException('EDITOR environment variable is not set. editor or vi are not available')
```
Jan-M: "editor" could be replaced with editor in message to hint at wrong editor value?

CyberDem0n: "editor" is more or less a standard thing:
```
$ ls -al /usr/bin/editor                                                                                                                                                                                                                                
lrwxrwxrwx 1 root root 24 Aug  8  2018 /usr/bin/editor -> /etc/alternatives/editor

$ ls -al /etc/alternatives/editor
lrwxrwxrwx 1 root root 9 Aug  8  2018 /etc/alternatives/editor -> /bin/nano

$ ls -al /bin/nano 
-rwxr-xr-x 1 root root 245872 Mär  6  2018 /bin/nano
```

Jan-M:
:+1:

CyberDem0n:
:+1:

Issue #736, opened 2018-07-16
1 of 3 nodes failing with "no recovery target specified" 


ktechkez:
Hi, 

I have been struggling for a couple of days now getting a 3 node setup with patroni/postgres running properly. I've been following this guide: https://www.linode.com/docs/databases/postgresql/create-a-highly-available-postgresql-cluster-using-patroni-and-haproxy/. 

The master and a single secondary node start fine every time, but a third node always fails to start postgres with the following: 

`Jul 16 14:30:45 devpostgres03 patroni[10951]: 2018-07-16 14:30:45,213 INFO: Lock owner: devpostgres01; I am devpostgres03
Jul 16 14:30:45 devpostgres03 patroni[10951]: 2018-07-16 14:30:45,228 INFO: Lock owner: devpostgres01; I am devpostgres03
Jul 16 14:30:45 devpostgres03 patroni[10951]: 2018-07-16 14:30:45,255 INFO: Local timeline=14 lsn=0/60005F8
Jul 16 14:30:45 devpostgres03 patroni[10951]: 2018-07-16 14:30:45,269 INFO: master_timeline=22
Jul 16 14:30:45 devpostgres03 patroni[10951]: 2018-07-16 14:30:45,271 INFO: master: history=1        0/19C4218        no recovery target specified
Jul 16 14:30:45 devpostgres03 patroni[10951]: 2        0/19C4638        no recovery target specified
Jul 16 14:30:45 devpostgres03 patroni[10951]: 3        0/40000D0        no recovery target specified
Jul 16 14:30:45 devpostgres03 patroni[10951]: 4        0/4000410        no recovery target specified
Jul 16 14:30:45 devpostgres03 patroni[10951]: 5        0/4000830        no recovery target specified
Jul 16 14:30:45 devpostgres03 patroni[10951]: 6        0/4000C50        no recovery target specified
Jul 16 14:30:45 devpostgres03 patroni[10951]: 7        0/4000F90        no recovery target specified
Jul 16 14:30:45 devpostgres03 patroni[10951]: 8        0/40011F0        no recovery target specified
Jul 16 14:30:45 devpostgres03 patroni[10951]: 9        0/4001370        no recovery target specified
Jul 16 14:30:45 devpostgres03 patroni[10951]: 10        0/40015D0        no recovery target specified
Jul 16 14:30:45 devpostgres03 patroni[10951]: 11        0/6000178        no recovery target specified
[...]

PR #1092, opened 2019-06-14
patronictl checks if --config value exists (#736)

wilfriedroset:
Adding this check allow patronictl ~~to return an error instead of ignoring silently the filename passed~~ to be verbose about non-existing file (or unreadable).
I've put the check in patronictl because it was simplier than modifying the [__init__ of Config object](https://github.com/zalando/patroni/blob/master/patroni/config.py#L76). I can move it if you want.

~~We can also change PatroniCtlException by a logging.warning to not modifying the behavior. Again I can happily edit this PR.~~

### DIFF
--- a/patroni/ctl.py
+++ b/patroni/ctl.py
@@ -68,7 +68,10 @@ def parse_dcs(dcs):
 
 
 def load_config(path, dcs):
-    logging.debug('Loading configuration from file %s', path)
+    if not (os.path.exists(path) and os.access(path, os.R_OK)):
+        logging.debug('Ignoring configuration file "%s". It does not exists or is not readable.', path)
+    else:
+        logging.debug('Loading configuration from file %s', path)
     config = {}
     old_argv = list(sys.argv)
     try:


patroni/ctl.py
```diff
@@ -68,7 +68,10 @@ def parse_dcs(dcs):
 
 
 def load_config(path, dcs):
-    logging.debug('Loading configuration from file %s', path)
+    if not (os.path.exists(path) or os.access(path, os.R_OK)):
```
CyberDem0n: The if statement has a logical error.

patroni/ctl.py
```diff
@@ -68,7 +68,10 @@ def parse_dcs(dcs):
...
+        logging.debug('Ignoring configuration file "{0}" does not exists or is not readable'.format(path))
```
CyberDem0n: `%s` vs `.format`, the first one if preferred for logging module.

patroni/ctl.py
```diff
@@ -68,7 +68,10 @@ def parse_dcs(dcs):
 
 
 def load_config(path, dcs):
-    logging.debug('Loading configuration from file %s', path)
+    if not (os.path.exists(path) and os.access(path, os.R_OK)):
+        logging.debug('Ignoring configuration file "%s" does not exists or is not readable' % path)
```
CyberDem0n:
```python
-logging.debug('Ignoring configuration file "%s" does not exists or is not readable' % path)
+logging.debug('Ignoring configuration file "%s" does not exists or is not readable', path)

File: patroni/ctl.py [hunk 4]
```diff
@@ -68,7 +68,10 @@ def parse_dcs(dcs):
 
 
 def load_config(path, dcs):
-    logging.debug('Loading configuration from file %s', path)
+    if not (os.path.exists(path) and os.access(path, os.R_OK)):
+        logging.debug('Ignoring configuration file "%s" does not exists or is not readable', path)
```
Jan-M: `Ignoring configuration file "%s". It does not exists or is not readable.`

CyberDem0n:
:+1:

Jan-M:
:+1:

PR #1099, opened 2019-07-02
Optional fencing script (pre_promote)

sdudoladov:
Call a fencing script after acquiring the leader lock.  If the script didn't finish successfully - don't promote but remove leader key


patroni/ha.py
```diff
@@ -1163,6 +1173,9 @@ def post_bootstrap(self):
         if not self.watchdog.activate():
             logger.error('Cancelling bootstrap because watchdog activation failed')
             self.cancel_initialization()
+        if not self.call_pre_promote(self.patroni.config):
```
CyberDem0n: We can tell for sure that there is no former leader if we are doing the bootstrap.

patroni/ha.py
```diff
@@ -1402,3 +1415,23 @@ def get_remote_member(self, member=None):
...
         return self.get_remote_member()
+
+    def call_pre_promote(self, config):
+        """
+        Runs a fencing script after the leader lock is acquired but before the replica is promoted.
+        If the script exits with a non-zero code, promotion does not happen and the leader key is removed from DCS.
+        """
+        cmd = config.get('pre_promote')
+        if cmd:
+            self._pre_promote_subprocess = CancellableSubprocess()
```
CyberDem0n: The idea of CancallableSubprocess is that there is only one instance of it, which is also used by [AsyncExecutor](https://github.com/zalando/patroni/blob/master/patroni/ha.py#L77), what makes it possible to easily cancel it.

patroni/ha.py
```diff
@@ -1402,3 +1415,23 @@ def get_remote_member(self, member=None):
 
     def get_remote_master(self):
         return self.get_remote_member()
+
+    def call_pre_promote(self, config):
```
CyberDem0n: I would really recommend to look at [post_bootstrap](https://github.com/zalando/patroni/blob/master/patroni/ha.py#L1151-L1159) method.

The `pre_promote` script should be executed by a function (or method) with the help of AsyncExecutor. It is hard to get the return code from such method, therefore we are using the [CryticalTask](https://github.com/zalando/patroni/blob/master/patroni/async_executor.py#L7) object. Usage example you can find [here](https://github.com/zalando/patroni/blob/master/patroni/postgresql/bootstrap.py#L336).

If the function executed asynchronously returns not None, the HA loop will be immediately [rescheduled](https://github.com/zalando/patroni/blob/master/patroni/async_executor.py#L107-L108).

patroni/ha.py
```diff
@@ -867,6 +872,11 @@ def process_unhealthy_cluster(self):
 
         if self.is_healthiest_node():
             if self.acquire_lock():
+
+                if not self.call_pre_promote(self.patroni.config):
```
CyberDem0n: In case is pre_romote call takes longer than `ttl` the leader lock will expire, therefore it must be executed with AsyncExecutor.

patroni/ha.py
```diff
@@ -1402,3 +1415,23 @@ def get_remote_member(self, member=None):
...
+    def call_pre_promote(self, config):
+        """
+        Runs a fencing script after the leader lock is acquired but before the replica is promoted.
+        If the script exits with a non-zero code, promotion does not happen and the leader key is removed from DCS.
+        """
+        cmd = config.get('pre_promote')
+        if cmd:
+            self._pre_promote_subprocess = CancellableSubprocess()
+            try:
+               ret = self._pre_promote_subprocess.call(shlex.split(cmd))
```
CyberDem0n: It makes sense to add some arguments, like the cluster name (`scope`) and the address and name of the former leader if they are known.

patroni/ha.py
sdudoladov: for sure, but does it mean we want to have a master that may be unfit as far as `pre_promote` is concerned ?

patroni/ha.py
sdudoladov: i will add/test them once the script itself correctly executes

patroni/ha.py
sdudoladov: Is it because having a single process simplifies the `AsyncExecutor` (is > 1 critical simultaneous task actually ever needed ?) or are there OS-level reasons for that ?

patroni/ha.py
```diff
@@ -723,8 +744,10 @@ def _delete_leader(self):
     def release_leader_key_voluntarily(self):
         self._delete_leader()
         self.touch_member()
+        self._released_leader_key_timestamp = time.time()
```
CyberDem0n: Do we want to have a backoff for cases not related to the failure of the `pre_promote` script?

patroni/ha.py
```diff
@@ -1099,7 +1126,10 @@ def handle_long_action_in_progress(self):
                             self.state_handler.terminate_starting_postmaster(postmaster=task.result)
                 self.demote('immediate-nolock')
                 return 'lost leader lock during ' + self._async_executor.scheduled_action
-
+            if self._async_executor.scheduled_action == 'promote' and self.state_handler.pre_promote_script:
```
CyberDem0n: Since the `pre_promote_script` is a property, it's value might change in runtime...
This is especially bad when it was returning something at the moment when the script was started and later it become empty.

patroni/postgresql/__init__.py
```diff
@@ -711,9 +718,42 @@ def _wait_promote(self, wait_seconds):
             if data.get('Database cluster state') == 'in production':
                 return True
 
+    def _call_pre_promote(self, cmd):
+        if cmd:
+            try:
+                ret = self.cancellable.call(shlex.split(cmd))
+            except OSError as err:
```
CyberDem0n: The `cancellable.call()` may raise the `PostgresException`...
In order to be safe, it is better to catch all `Exception` here.

patroni/postgresql/__init__.py
```diff
@@ -711,9 +718,42 @@ def _wait_promote(self, wait_seconds):
...
+                logger.error('pre_promote script %s failed: %s', cmd, err)
+                return False
+
+            if ret != 0:
```
CyberDem0n: return code might also be `None`

patroni/postgresql/__init__.py
```diff
@@ -711,9 +718,42 @@ def _wait_promote(self, wait_seconds):
...
+            logger.exception('pre_promote')
+            self.pre_promote_task.complete(False)
+
+        return self.pre_promote_task.result
+
     def promote(self, wait_seconds, on_success=None, access_is_restricted=False):
         if self.role == 'master':
             return True
+
+        if self.pre_promote_script and not self._pre_promote():
```
CyberDem0n: Here we've got a race condition with https://github.com/zalando/patroni/pull/1099/files#diff-57c53e988b775ee2902488fb33c1f433R1131.
In order to solve it we probably need to use the approch like in the [Postgresql.start()](https://github.com/zalando/patroni/blob/master/patroni/postgresql/__init__.py#L421-L432) method.

patroni/postgresql/__init__.py
```diff
@@ -711,9 +718,42 @@ def _wait_promote(self, wait_seconds):
...
+    def _pre_promote(self):
+        """
+        Runs a fencing script after the leader lock is acquired but before the replica is promoted.
+        If the script exits with a non-zero code, promotion does not happen and the leader key is removed from DCS.
+        """
+
+        self.pre_promote_task = CriticalTask()
+        try:
+            self.pre_promote_task.complete(self._call_pre_promote(self.pre_promote_script))
+        except Exception:
```
sdudoladov: remove

patroni/postgresql/__init__.py
sdudoladov: is it the race condition between cancelling and starting `pre_promote` ? if so, isn't it already handled by the combination of [self._cancellable.cancel()](https://github.com/zalando/patroni/blob/master/patroni/async_executor.py#L127) and [raise PostgresException('cancelled')](https://github.com/zalando/patroni/blob/master/patroni/postgresql/cancellable.py#L92) ?

CyberDem0n: > is it the race condition between canceling and starting pre_promote?

Yes

> isn't it already handled by the combination of self._cancellable.cancel() and raise PostgresException('cancelled') ?

No, it is not. Don't forget about GIL. For example the cancel call might happen after the script was successfully finished but before `self.pg_ctl('promote', '-W')`

sdudoladov:
:+1:

CyberDem0n:
:+1:

PR #1113, opened 2019-07-25
Fix Windows signals incompatibilities

pashagolub:
On Windows only SIGTERM, CTRL_C_EVENT and CTRL_BREAK_EVENT signals are supported. However, sending Ctrl+Break requires already being attached to the same console as the target. Note that the target isn't necessarily a single process that's attached to the console. Sending a control event is implemented by calling WinAPI GenerateConsoleCtrlEvent, which targets process group identifiers (i.e. like Unix killpg), not process identifiers (i.e. not like Unix kill). 

So the only sane solution is to send SIGTERM for all modes supported.

### DIFF
--- a/patroni/postgresql/postmaster.py
+++ b/patroni/postgresql/postmaster.py
@@ -107,7 +107,7 @@ class PostmasterProcess(psutil.Process):
         :returns None if signaled, True if process is already gone, False if error
         """
         if self.is_single_user:
-            logger.warning("Cannot stop server; single-user server is running (PID: {0})".format(self.pid))
+            logger.warning("Cannot stop server; single-user server is running with PID %s)", self.pid)
             return False
         if os.name != "nt":
             try:
@@ -119,12 +119,12 @@ class PostmasterProcess(psutil.Process):
                 logger.warning("Could not send stop signal to PostgreSQL (error: {0})".format(e))
                 return False
         else:
-            if subprocess.call([pg_ctl, 'kill', STOP_SIGNALS[mode], self.pid]) == 0:
+            if subprocess.call([pg_ctl, 'kill', STOP_SIGNALS[mode], str(self.pid)]) == 0:
                 return None
             elif not self.is_running():
                 return True
             else:
-                logger.warning("Could not send stop signal to PostgreSQL with PID {0}".format(self.pid))
+                logger.warning("Could not send stop signal to PostgreSQL with PID %s", self.pid)
                 return False
 
     def wait_for_user_backends_to_close(self):
--- a/patroni/postgresql/postmaster.py
+++ b/patroni/postgresql/postmaster.py
@@ -21,6 +21,7 @@ else:
         'immediate': 'QUIT',
     }
 
+
 def pg_ctl_start(conn, cmdline, env):
     if os.name != 'nt':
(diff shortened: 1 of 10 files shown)


CyberDem0n:
Yeah, Windows is a quite special beast...
Your change doesn't seem right to me. The SIGTERM is used for the smart shutdown, which is never triggered by Patroni. We are using only fast and immediate.
I just checked the source code and figured out that pg_ctl on Windows is using a special PIPE to send kind of unix signals to a postmaster. In order to get everything working correctly on Windows, we have to use `pg_ctl kill    SIGNALNAME PID` instead of just calling `send_signal`, which is in turn is a simple wrapper of `os.kill()`.

patroni/postgresql/postmaster.py
```diff
@@ -102,15 +108,23 @@ def signal_stop(self, mode):
...
+            try:
+                self.send_signal(STOP_SIGNALS[mode])
+                return None
+            except psutil.NoSuchProcess:
+                return True
+            except psutil.AccessDenied as e:
+                logger.warning("Could not send stop signal to PostgreSQL (error: {0})".format(e))
+                return False
+        else:
+            if subprocess.call(['pg_ctl', 'kill', STOP_SIGNALS[mode], self.pid]) == 0:
```
CyberDem0n: Usually `pg_ctl` is not in the default path, you have to use [pg_ctl method](https://github.com/zalando/patroni/blob/master/patroni/postgresql/__init__.py#L160-L166) instead.
Another problem is with passing `self.pid`, which is an integer, while command-line arguments could only be strings.

patroni/postgresql/postmaster.py
```diff
@@ -123,7 +124,7 @@ def signal_stop(self, mode):
             elif not self.is_running():
                 return True
             else:
-                logger.warning("Could not send stop signal %s to postmaster with PID %s", STOP_SIGNALS[mode], self.pid)
+                logger.warning("Could not send stop signal to PostgreSQL with PID {0}".format(self.pid))
```
CyberDem0n: Why this change? `%s` formatting is preferred with the logging module

patroni/postgresql/postmaster.py
pashagolub: But `pg_ctl kill` cannot have `['-D', self._data_dir]`, but they are hardcoded in pg_ctl method. Rewrite it?

CyberDem0n: Oh, You are right.

Than `subprocess.call(['pg_ctl', 'kill', STOP_SIGNALS[mode], self.pid])` should be changed to:
```
subprocess.call([pgcommand('pg_ctl'), 'kill', STOP_SIGNALS[mode], str(self.pid)])
```

Or maybe just pass `pg_ctl` as an argument to the `signal_stop()`. It is kind of ugly, but we already doing it with `PostmasterProcess.start()`

patroni/postgresql/postmaster.py
pashagolub: To be consistent with the similar log message several lines before :)

CyberDem0n: I would actually like to have the other line fixed too, for consistency :)

Issue #1121, opened 2019-07-28
invalid literal for int() with base 10 when using IPv6 addresses


optix2000:
Hi,

I'm working with an IPv6 only Kubernetes cluster and using the spilo container with the latest patron 1.5.6.

However I get the following error when I put in an IPv6 address for `connect_address`.
```
Traceback (most recent call last):
  File "/usr/local/lib/python3.6/dist-packages/patroni/async_executor.py", line 97, in run
    wakeup = func(*args) if args else func()
  File "/usr/local/lib/python3.6/dist-packages/patroni/ha.py", line 220, in clone
    if self.state_handler.bootstrap.clone(clone_member):
  File "/usr/local/lib/python3.6/dist-packages/patroni/postgresql/bootstrap.py", line 288, in clone
    ret = self.create_replica(clone_member) == 0
  File "/usr/local/lib/python3.6/dist-packages/patroni/postgresql/bootstrap.py", line 177, in create_replica
    r = clone_member.conn_kwargs(self._postgresql.config.replication)
  File "/usr/local/lib/python3.6/dist-packages/patroni/dcs/__init__.py", line 232, in conn_kwargs
    return self.member.conn_kwargs(auth)
  File "/usr/local/lib/python3.6/dist-packages/patroni/dcs/__init__.py", line 158, in conn_kwargs
    'port': r.port or 5432,
  File "/usr/lib/python3.6/urllib/parse.py", line 169, in port
    port = int(port, 10)
ValueError: invalid literal for int() with base 10: 'ffff::d0:5432'
```

PR #1122, opened 2019-07-29
IPv6 support

CyberDem0n:
fixes https://github.com/zalando/patroni/issues/1121

### DIFF
--- a/patroni/api.py
+++ b/patroni/api.py
@@ -554,8 +554,6 @@ class RestApiServer(ThreadingMixIn, HTTPServer, Thread):
         dual_stack = self.__has_dual_stack()
         if host == '':
             host = None
-        if host is None and dual_stack:
-            host = '::'
 
         info = socket.getaddrinfo(host, port, socket.AF_UNSPEC, socket.SOCK_STREAM, 0, socket.AI_PASSIVE)
         # in case dual stack is not supported we want IPv4 to be preferred over IPv6
--- a/patroni/api.py
+++ b/patroni/api.py
@@ -7,11 +7,12 @@ import traceback
 import dateutil.parser
 import datetime
 import os
+import socket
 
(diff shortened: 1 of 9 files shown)


patroni/api.py
```diff
@@ -532,8 +533,36 @@ def __get_ssl_options(config):
...
+                if sock:
+                    sock.close()
+        return False
+
+    def __httpserver_init(self, host, port):
+        dual_stack = self.__has_dual_stack()
+        if host == '':
+            host = None
+        if host is None and dual_stack:
+            host = '::'
```
erthalion: Is this required? Python docs says, that if

```
By passing None as the value of host and port, you can
pass NULL to the underlying C API.
```

and in this case:

```
the network address will be set to the loopback interface address
(INADDR_LOOPBACK for IPv4 addresses, IN6ADDR_LOOPBACK_INIT
for IPv6 address)
```

So looks like with `None` result would be the same.

CyberDem0n: You are right. Will remove it.

patroni/utils.py
```diff
@@ -342,3 +343,17 @@ def split_host_port(value, default_port):
     t = value.rsplit(':', 1)
     t.append(default_port)
     return t[0], int(t[1])
+
+
+def uri(proto, netloc, path='', user=None):
+    host, port = netloc if isinstance(netloc, (list, tuple)) else split_host_port(netloc, 0)
+    if host and host[0] != '[' and host[-1] != ']':
+        try:
+            socket.inet_pton(socket.AF_INET6, host)
```
erthalion: Looks like Windows support for `inet_pton` was added only in 3.4, which makes supported versions a bit different.

erthalion:
:+1:

erthalion:
:+1:

CyberDem0n:
:+1:

PR #1134, opened 2019-08-03
Allow for certificate-based authentication from Patroni PostgreSQL accounts

jkatz:
The two principal features this introduces:

1. Provide the Patroni PostgreSQL management accounts (superuser, replication, rewind) to be able to authenticate using certificate-based authentication
1. Allow the user to specify the `sslmode` they wish to connect as. In my tests I used `verify-full` and `verify-ca`

### References
- [PostgreSQL Certificate Based Authentication](https://www.postgresql.org/docs/current/auth-cert.html)
- [libpq connection parameters](https://www.postgresql.org/docs/current/libpq-connect.html) which are used by psycopg2
- [SSL Modes](https://www.postgresql.org/docs/current/libpq-ssl.html)


CyberDem0n:
> I did not run this through the regression test suite as I was unsure how to run the tests.

This is very easy, you need to have all python modules from requirements.txt installed plus `behave coverage flake8 mock pytest-cov pytest`.

There are two different test suits:
1. `python setup.py test` -- will execute unit-tests
2. `behave` -- integration tests

Intergration tests require postgres binaries and etcd be available in the PATH.

`python setup.py flake8` will check the formatting of source code.

docs/ENVIRONMENT.rst
```diff
@@ -26,9 +26,12 @@ It is possible to create new database users right after the successful initializ
 
 -  **PATRONI\_<username>\_PASSWORD='<password>'**
 -  **PATRONI\_<username>\_OPTIONS='list,of,options'**
+-  **PATRONI\_<username>\_DSN='<postgresql://...>'**
```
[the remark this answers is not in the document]
jkatz: Thanks for the explanation. That makes sense.

jkatz:
> > I opted for the dsn parameter to avoid introducing 5 or so new parameters.
> 
> It doesn't seem to make a configuration much simpler but really increases code [complexity](https://github.com/zalando/patroni/pull/1134/files#diff-188997f9c97e1dac6c37391a509d13c9R358-R413).
> Having a separate configuration parameter will just require:
> 
> 1. adding them to [_get_auth](https://github.com/zalando/patroni/blob/master/patroni/config.py#L253-L259) (to handle environment variables)
> 2. filtering out unwanted paramters in the [_build_effective_configuration](https://github.com/zalando/patroni/blob/master/patroni/config.py#L355)
> 
> Something like:
> 
> ```python
> +AUTH_ALLOWED_PARAMETERS = {'username', 'password', 'sslmode', 'sslcert', 'sslkey', 'sslrootcert', 'sslcrl'}
> 
> ...
> 
>     def _get_auth(name):
> -           for param in ('username', 'password'):
> +           for param in AUTH_ALLOWED_PARAMETERS:
>                 value = _popenv(name + '_' + param)
>                 if value:
>                     ret[param] = value
>             return ret
> 
> ...
> 
>     def _build_effective_configuration(self, dynamic_configuration, local_configuration):
> ....
>         # no 'superuser' in the 'postgresql'.'authentication'
>         if 'superuser' not in pg_config['authentication'] and 'pg_rewind' in pg_config:
>             pg_config['authentication']['superuser'] = pg_config['pg_rewind']
> +
> +       for name, value in pg_config['authentication'].items():
> +           pg_config['authentication'][name] = {n: v for n, v in value.items() if n in AUTH_ALLOWED_PARAMETERS}
> ```
> 
> In the `_build_effective_configuration` we could also rename `username` key (if exists) to `user` in order to avoid such "renames" down the code, like [here](https://github.com/zalando/patroni/pull/1134/files#diff-a534a7e0f2c46ba53349f820d9014468R593-R595) for example.

Good call...that seems a lot more reasonable, and should make the code overall a lot less complex. I'll see how quickly I can turn this around.

patroni/postgresql/config.py
```diff
@@ -136,10 +136,12 @@ def parse_dsn(value):
 
     if ret:
         requiressl = ret.pop('requiressl', None)
-        if requiressl == '1':
-            ret['sslmode'] = 'require'
-        elif requiressl is not None:
-            ret['sslmode'] = 'prefer'
+        # allow sslmode to take precedence over requiressl
```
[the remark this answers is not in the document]
jkatz: Cool :smile:

The natural follow-up to having `sslmode` available will be able to pass in the other certificate parameters, but that's on me to clean up the code 😉

jkatz:
@CyberDem0n OK, one month later I finally got back to this :smile:

Per your suggestion, I eliminated the `dsn` parameter and broke it out into the various specific `ssl*` parameters. The implementation is certainly much cleaner.

I followed your `AUTH_ALLOWED_PARAMETERS` suggestion:

https://github.com/zalando/patroni/pull/1134/files#diff-188997f9c97e1dac6c37391a509d13c9R18

though I added some additional parameters due to the `_get_auth` function also being used for the `restapi` and `etcd` configs:

- restapi: https://github.com/zalando/patroni/blob/master/patroni/config.py#L261
- etcd: https://github.com/zalando/patroni/blob/master/patroni/config.py#L303

This also makes it tricky for the "rename username => user" bit ([here](https://github.com/zalando/patroni/pull/1134/files#diff-a534a7e0f2c46ba53349f820d9014468R592))  given both etcd + restapi rely on username. My suggestion would be to leave it alone for now, or work on it in a different commit/PR.

Open to additional feedback. Thanks!

patroni/config.py
```diff
@@ -9,12 +9,31 @@
...
+    # universal
+    'username',
+    'password',
+    # postgresql only
+    'sslmode',
+    'sslcert',
+    'sslkey',
+    'sslrootcert',
+    'sslcrl',
+    # etcd only
```
CyberDem0n: We don't need to put anything related to etcd or restapi here.
The problem is that ENV variable names are not always consistent with structures in YAML file:
```
PATRONI_RESTAPI_USERNAME=user
PATRONI_RESTAPI_PASSWORD=passwd
```
vs
```yaml
restapi:
  authentication:
    username: user
    password: passwd
```

It doesn't break the code only because other ENV variable names are already "consumed" by [_set_section_values()](https://github.com/zalando/patroni/pull/1134/files#diff-188997f9c97e1dac6c37391a509d13c9R252) a few lines earlier than [_get_auth()](https://github.com/zalando/patroni/pull/1134/files#diff-188997f9c97e1dac6c37391a509d13c9L261)

patroni/config.py
```diff
@@ -308,14 +327,24 @@ def _parse_list(value):
...
                 # CREATE USER "<username>" WITH <OPTIONS> PASSWORD '<password>'
-                if name and suffix == 'PASSWORD':
-                    password = os.environ.pop(param)
-                    if password:
-                        users[name] = {'password': password}
-                        options = os.environ.pop(param[:-9] + '_OPTIONS', None)
-                        options = options and _parse_list(options)
-                        if options:
-                            users[name]['options'] = options
+                if name:
```
CyberDem0n: It seems that you forgot to roll back this change.

patroni/config.py
jkatz: Great, thanks for the explanation! I condensed the `tuple` to only the relevant PostgreSQL parameters.

jkatz:
@CyberDem0n Thanks for the feedback! I made another pass and added your suggestions.

avaczi:
:+1:

CyberDem0n:
:+1:

PR #1146, opened 2019-08-14
Get rid from distutils module dependency

CyberDem0n:
We are using only one function from there, `find_executable()` and it is better to implement a similar function in Patroni rather than add `distutils` module into requirements.txt


sdudoladov:
:+1:

CyberDem0n:
:+1:

PR #1148, opened 2019-08-14
Allow lower values for postgresql configuration

### LINKED ISSUE #1142: Reasons of minimum PostgreSQL parameters value (max_connections, ...)
In `postresql/config.py` there is:

```
        'max_connections': (100, lambda v: int(v) >= 100, 90100),
        'max_wal_senders': (10, lambda v: int(v) >= 10, 90100),
        'wal_keep_segments': (8, lambda v: int(v) >= 8, 90100),
        # [...]
```

Is there any reason to have such "high" minimum in the validation method?

We're running small databases for customers where the number of connections for the first plan is 30, and we can't apply such configuration using patroni. (Same for `wal_keep_segments` and `max_wal_senders` which would be running correctly with lower values.

What about keeping the defaults and updating the minimum values?

ie

```
        'max_connections': (100, lambda v: int(v) >= 25, 90100),
        'max_wal_senders': (10, lambda v: int(v) >= 3, 90100),
        'wal_keep_segments': (8, lambda v: int(v) >= 1, 90100),
```

Regards,

Soulou:
* Default values have not been changed
* These minimal values still work properly to boot a (small) cluster

Fixes #1142

### DIFF
--- a/patroni/postgresql/config.py
+++ b/patroni/postgresql/config.py
@@ -176,14 +176,14 @@ class ConfigHandler(object):
         'cluster_name': (None, lambda _: False, 90500),
         'wal_level': ('hot_standby', lambda v: v.lower() in ('hot_standby', 'replica', 'logical'), 90100),
         'hot_standby': ('on', lambda _: False, 90100),
-        'max_connections': (100, lambda v: int(v) >= 100, 90100),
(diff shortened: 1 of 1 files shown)


sdudoladov:
:+1:

CyberDem0n:
:+1:

PR #1150, opened 2019-08-18
Add PATRONICTL_CONFIG_FILE Environment Variable

msvechla:
This PR adds a `PATRONICTL_CONFIG_FILE` environment variable, which allows configuring the `--config-file` flag from the environment.

This is especially helpful when running patroni inside a Kubernetes cluster, e.g. via [postgres-operator](https://github.com/zalando/postgres-operator) and interacting with `patronictl` inside a pod.

As the default location is not working here, setting the correct location as environment variable in the pod will allow us to use `patronictl` without specifying the config file option with every command.

According to https://github.com/pallets/click/pull/874 this change will be non breaking, as first the cli option, then the environment variable, and finally the default will be evaluated.

Let me know if you need anything else and thanks for this awesome project!

### DIFF
--- a/patroni/ctl.py
+++ b/patroni/ctl.py
@@ -111,7 +111,8 @@ option_insecure = click.option('-k', '--insecure', is_flag=True, help='Allow con
 
 
 @click.group()
-@click.option('--config-file', '-c', help='Configuration file', envvar='PATRONICTL_CONFIG_FILE', default=CONFIG_FILE_PATH)
+@click.option('--config-file', '-c', help='Configuration file',
+              envvar='PATRONICTL_CONFIG_FILE', default=CONFIG_FILE_PATH)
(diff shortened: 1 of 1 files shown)


CyberDem0n:
> especially helpful when running patroni inside a Kubernetes cluster, e.g. via postgres-operator and interacting with patronictl inside a pod.

Well, in general I don't object on the change, but there is a much simpler fix to your problem, just run patronictl as a postgres user.
```
# su postgres
$ patronictl
```

sdudoladov:
:+1:

msvechla:
@CyberDem0n thanks for the hint! Nevertheless I think it would make sense to add this option. Can we get this merged?

CyberDem0n:
:+1:

CyberDem0n:
Merged, thank you!

Issue #1128, opened 2019-07-31
OK Status code for GET /replica, when replica is in recovery


jpuris:
Patroni version 1.5.6

**Problem**
`GET /replica` returns status code 200, when replica (hot standby) is not ready to accept connections.
API call returns
```json
{
  "state": "starting",
  "database_system_identifier": "<redacted>",
  "role": "replica",
  "patroni": {
    "scope": "<redacted>",
    "version": "1.5.6"
  }
}
```
This is misleading for out use case, as we rely on status code 200 for this API call, to determine, if replica is up and running, for us to potentially send reads to it.

**Steps to reproduce**
- Have an offline replica that has a lot of replication lag
- Have this replica use WAL archive to recover

Start replica, which is going to be in postgresql recovery state.

Expected behaviour - `GET /replica` returns status code 503

Actual behaviour - `GET /replica` returns status code 200

CyberDem0n:
Yes, this is the known [problem](https://github.com/zalando/patroni/issues/772) and it should not be hard to fix it. Pull requests are always welcomed!

anikin-aa:
Guys, am I correct, that 200 status code should be return only, when `state` == `running` ?

CyberDem0n:
@anikin-aa, yes

jpuris:
Thank you very much, everyone involved! :)

PR #1152, opened 2019-08-20
Fix status code for GET /replica, when replica is starting

anikin-aa:
https://github.com/zalando/patroni/issues/1128

### DIFF
--- a/patroni/api.py
+++ b/patroni/api.py
@@ -101,9 +101,8 @@ class RestApiHandler(BaseHTTPRequestHandler):
         else:
             primary_status_code = 200 if patroni.ha.is_leader() else 503
 
-        replica_status_code = 200 \
-            if not patroni.noloadbalance and response.get('role') == 'replica' and response.get('state') == 'running' \
-            else 503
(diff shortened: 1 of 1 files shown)


patroni/api.py
```diff
@@ -109,7 +109,7 @@ def do_GET(self, write_status_code_only=False):
         elif 'master' in path or 'leader' in path or 'primary' in path or 'read-write' in path:
             status_code = primary_status_code
         elif 'replica' in path:
-            status_code = replica_status_code
+            status_code = 200 if response.get('state') == 'running' else 503
```
CyberDem0n: Your change fixes the problem only partially and unfortunately not 100% correctly.
Instead you should look on the line [104](https://github.com/zalando/patroni/pull/1152/files#diff-790530624eaa21e5fb614fc321aef277R104) and fix it.

anikin-aa: ok, got it, my bad. changes are commited

sdudoladov:
:+1:

CyberDem0n:
:+1:

PR #1155, opened 2019-08-23
Pass statement_timeout=0 in PGOPTIONS when doing pg_rewind

CyberDem0n:
It might happen that statement_timeout on the server is set to some small value and one of the statements executed by pg_rewind is canceled.

I already proposed a patch fixing the pg_rewind itself, but it also would be good to have a workaround in Patroni.


RafiaSabih:
:+1:

sdudoladov:
:+1:

PR #1156, opened 2019-08-26
Kubernetes services for Read and Read-Write

karthiksharmasg:
Adding 2 services depending on pod selector for Read and Read-Write operations.

### DIFF
--- a/kubernetes/patroni_k8s_service_master.yaml
+++ b/kubernetes/patroni_k8s_service_master.yaml
@@ -1,0 +1,17 @@
+apiVersion: v1
+kind: Service
+metadata:
+  name: &cluster_name patronidemo
+  labels:
+    application: patroni
+    cluster-name: *cluster_name
+    
+spec:
+  type: ClusterIP
+  selector:
+    application: patroni
+    role: master
+    cluster-name: patronidemo
+  ports:
+  - port: 5432
+    targetPort: 5432
--- a/kubernetes/patroni_k8s_service_master.yaml
+++ b/kubernetes/patroni_k8s_service_master.yaml
@@ -1,7 +1,7 @@
 apiVersion: v1
 kind: Service
 metadata:
-  name: &cluster_name patronidemo
+  name: &cluster_name patronidemo-master
   labels:
     application: patroni
     cluster-name: *cluster_name
--- a/kubernetes/patroni_k8s_service_replica.yaml
+++ b/kubernetes/patroni_k8s_service_replica.yaml
@@ -1,0 +1,17 @@
+apiVersion: v1
+kind: Service
+metadata:
+  name: &cluster_name patronidemo-replica
+  labels:
+    application: patroni
+    cluster-name: *cluster_name
+    
+spec:
+  type: ClusterIP
+  selector:
+    application: patroni
+    role: replica
+    cluster-name: patronidemo
+  ports:
+  - port: 5432
+    targetPort: 5432


CyberDem0n:
There is already read-write service: https://github.com/zalando/patroni/blob/master/kubernetes/patroni_k8s.yaml#L113-L124

karthiksharmasg:
@CyberDem0n 
The current service will not point to proper pod after the promotion of a node, hence i have added the role as either "master" or "replica".

CyberDem0n:
>The current service will not point to proper pod after the promotion of a node

Really? How do you run it? Do you have PATRONI_KUBERNETES_USE_ENDPOINTS set to true: https://github.com/zalando/patroni/blob/master/kubernetes/patroni_k8s.yaml#L51-L52?

karthiksharmasg:
Yes, I was able to get the complete cluster up, but after new pod took over the master role post deletion of "old master", this server lost the connection.

CyberDem0n:
> but after new pod took over the master role post deletion of "old master", this server lost the connection

I don't really get it. How you explain "lost the connection"? Any application MUST reconnect after failover or switchover. It is not possible to do it transparently.

karthiksharmasg:
Let me try to explain it in a different way, 

lets take this scenario where i have replica set as 3;
patronidemo-0   role=master
patronidemo-1   role=replica
patronidemo-2   role=replica

We have a default service patronidemo which binds to backend pods by application and cluster-name but not with any role

{
    "apiVersion": "v1",
    "kind": "Service",
    "metadata": {
        "creationTimestamp": "2019-08-26T08:29:50Z",
        "labels": {
            "application": "patroni",
            "cluster-name": "patronidemo"
        },
        "name": "patronidemo",
        "namespace": "default",
        "resourceVersion": "16116595",
        "selfLink": "/api/v1/namespaces/default/services/patronidemo",
        "uid": "abbbab49-c7db-11e9-892c-005056ad1ff1"
    },
    "spec": {
        "clusterIP": "xx.xx.xx.xx",
        "ports": [
            {
                "port": 5432,
                "protocol": "TCP",
                "targetPort": 5432
            }
        ],
        "sessionAffinity": "None",
        "type": "ClusterIP"
    },
    "status": {
        "loadBalancer": {}
    }
}

these new services make sure that we have a specific service endpoints that reroutes read and read-write operations to master or a replica nodes.

CyberDem0n:
@karthiksharmasg you are not answering my question.

> We have a default service patronidemo which binds to backend pods by application and cluster-name but not with any role

This service will connect you to the master, because Patroni updates the endpoint and puts IP of the master pod into subsets! I am 100% confident that it works correctly, in the opposite case failover wouldn't work.

karthiksharmasg:
@CyberDem0n 
IF that's the case, then post fail over its not adding the IP of the master pod into subsets, let me validate this part before i ask for the spec changes.

CyberDem0n:
1. There is no such thing as post failover
2. The new leader atomically puts it ip into subsets when grabs the leader lock: https://github.com/zalando/patroni/blob/master/patroni/dcs/kubernetes.py#L342.

karthiksharmasg:
Yes, this i understood, let me check my environment why this is not happening.

karthiksharmasg:
@CyberDem0n 
There was a issue with incorrect label definition in the service and hence the endpoint update was not properly read by service, I will close this PR.

Issue #1173, opened 2019-09-12
Force log output on shutdown  


sdudoladov:
Log lines Patroni outputs during shutdown do not reach the destination.
The shutdown method itself works as expected, and the following log lines are visible when running in a debugger:
```
DEBUG: http://127.0.0.1:2379 "DELETE /v2/keys/service/batman/leader?prevValue=postgresql0 HTTP/1.1" 200 268
DEBUG: Writing {"conn_url":"postgres://127.0.0.1:5432/postgres","api_url":"http://127.0.0.1:8008/patroni","state":"stopped","role":"master","version":"1.6.0"} to key /service/batman/members/postgresql0 ttl=30000 dir=False append=False
DEBUG: Converted retries value: 1 -> Retry(total=1, connect=None, read=None, redirect=0, status=None)
DEBUG: http://127.0.0.1:2379 "PUT /v2/keys/service/batman/members/postgresql0 HTTP/1.1" 200 707
```

These lines are however absent form the log destination.

Just for completeness I suggest Patroni always logs the result of the API shutdown and the final log line indicating the shutdown completed.

PR #1178, opened 2019-09-14
Fix logger shutdown behavior

CyberDem0n:
Since it is based on Thread with daemon set to True, the shutdown of logger was very likely to happen too early, what was causing some lines not to appear at the destination.

Close https://github.com/zalando/patroni/issues/1173


avaczi:
:+1:

CyberDem0n:
:+1:

PR #1184, opened 2019-09-19
Add missing ipv6 support to config

jacksontj:
Continuation of #1122

### DIFF
--- a/patroni/postgresql/config.py
+++ b/patroni/postgresql/config.py
@@ -617,7 +617,12 @@ class ConfigHandler(object):
         self._local_address = local_address
         self.local_replication_address = {'host': tcp_local_address, 'port': port}
 
-        netloc = self._config.get('connect_address') or tcp_local_address + ':' + port
+        if self._config.get('connect_address'):
+            netloc = self._config.get('connect_address')
+        else:
+            if tcp_local_address and ':' in tcp_local_address and tcp_local_address[0] != '[' and tcp_local_address[-1] != ']':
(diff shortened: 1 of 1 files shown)


patroni/postgresql/config.py
```diff
@@ -617,7 +617,12 @@ def resolve_connection_addresses(self):
         self._local_address = local_address
         self.local_replication_address = {'host': tcp_local_address, 'port': port}
 
-        netloc = self._config.get('connect_address') or tcp_local_address + ':' + port
+        if self._config.get('connect_address'):
+            netloc = self._config.get('connect_address')
+        else:
+            if tcp_local_address and ':' in tcp_local_address and tcp_local_address[0] != '[' and tcp_local_address[-1] != ']':
```
CyberDem0n: You are just duplicating the code from the [uri()](https://github.com/zalando/patroni/pull/1122/files#diff-94dca4ff917ed5e452b87407007c08f0R347-R354) function is already handling all possible cases and wraps IPv6 into `[]` if it is necessary.

Have you bumped into some specific issue? What problem you are trying to solve?

jacksontj: I ran into a specific case where the connect_address was of the form `<ipv6_address>:port` (like it was string concat-ed like an ipv6 address. So if this is being wrapped above I must have an issue somewhere else :/

CyberDem0n: Just make sure that you are running 1.6.0

PR #1186, opened 2019-09-21
Explicitly start logger Thread

CyberDem0n:
The PatroniLogger object is instantiated in the Patroni constructor and down the road there might be a fatal error causing Patroni process to exit, but live thread prevents the normal shutdown.
In order to mitigate the issue and don't loose ability to use the logging infrastructure we will switch to QueueLogger only when the thread was explicitly started from the Patroni.run() method.

Continuation of https://github.com/zalando/patroni/pull/1178

### DIFF
--- a/patroni/__init__.py
+++ b/patroni/__init__.py
@@ -114,6 +114,7 @@ class Patroni(object):
 
     def run(self):
         self.api.start()
+        self.logger.start()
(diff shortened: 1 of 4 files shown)


FxKu:
:+1:

CyberDem0n:
:+1:

PR #1189, opened 2019-09-23
Switch to a streaming watcher

CyberDem0n:
Watch requests to K8s API either streaming the data or close connection by timeout. In any case it requires a second connection open, but opening a new connection every 10 seconds is more expensive for both, Patroni and K8s API.

Switching to the streaming model also brings other benefits: we can watch not only on leader object, but also on config and wake up Patroni main thread if the config was changed.

### DIFF
--- a/patroni/dcs/kubernetes.py
+++ b/patroni/dcs/kubernetes.py
@@ -11,8 +11,10 @@ from kubernetes import client as k8s_client, config as k8s_config, watch as k8s_
 from patroni.dcs import AbstractDCS, ClusterConfig, Cluster, Failover, Leader, Member, SyncState, TimelineHistory
 from patroni.exceptions import DCSError
 from patroni.utils import deep_compare, tzutc, Retry, RetryFailedError
+from urllib3 import Timeout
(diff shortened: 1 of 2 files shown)


patroni/dcs/kubernetes.py
```diff
@@ -70,6 +72,48 @@ def wrapper(*args, **kwargs):
...
+
+    def _process_event(self, event):
+        ev_type = event.get('type')
+        metadata = event['raw_object'].get('metadata', {})
+        name = metadata.get('name')
+        annotations_map = {self._dcs.leader_path: self._dcs._LEADER, self._dcs.config_path: self._dcs._CONFIG}
+        value = None if ev_type == 'DELETED' else metadata.get('annotations', {}).get(annotations_map.get(name))
+
+        if name == self._dcs.leader_path:
+            if ev_type != 'ADDED' and self._leader_value and value != self._leader_value:
```
erthalion: I still think a commentary why conditions for leader and config are different.

erthalion:
:+1:

CyberDem0n:
:+1:

PR #1191, opened 2019-09-23
Implement /history and /cluster endpoints

### LINKED ISSUE #586: Store and expose the information about the recent promotion event and current timeline info
The timeline information should be stored separately for each member. The recent promotion time should be stored in Etcd, likely in some new key. One can use those information to make a check that all members are running the same timeline, or, if we store the history of promotions, whether promotions are happening too fast.

### LINKED ISSUE #675: Omit service info from patronictl list when using json format
The service info is not valid JSON and this extra output makes scripting harder and less robust. Maybe we want to also expose that info in some other way that is machine readable.

### LINKED ISSUE #1133: Add 'list' method to API
Currently Patroni cluster can be mostly managed via API, almost all commands from `patronictl` is available. Although not of them documented. But I'm really missing `list` command.  It would be great if it will be implemented in foreseeable future.

I.e. for restart cluster nodes in correct order after changing `max_connections` PostgreSQL option, I need to get all nodes and then restart master first an then slaves or vice versa depending on increasing or decreasing the `max_connections` value.

Currently direct DCS (consul in my case) request can be used as workaround, but due different DCS supported by Patroni it is not versatile and pretty hard to support, while Patroni already have all required code.

CyberDem0n:
The /history endpoint shows the content of the `history` key in DCS
The /cluster endpoint show all cluster members and some service info like pending and scheduled restarts or switchovers.

In addition to that implement `patronictl history`

Close #586
Close #675
Close #1133


Jan-M:
Can patronictl also receive a nice readable history command?

patroni/utils.py
```diff
@@ -352,3 +352,51 @@ def uri(proto, netloc, path='', user=None):
...
+            role = 'standby_leader' if is_standby_cluster(config.get('standby_cluster')) else 'leader'
+        elif m.name == cluster.sync.sync_standby:
+            role = 'sync_standby'
+        else:
+            role = 'replica'
+
+        conn_kwargs = m.conn_kwargs()
+        member = {'name': m.name, 'host': conn_kwargs['host'], 'port': int(conn_kwargs['port']),
+                  'role': role, 'state': m.data.get('state', ''), 'api_url': m.api_url}
+        member.update({n: m.data[n] for n in 'timeline pending_restart scheduled_restart tags'.split() if n in m.data})
```
erthalion: Why `split`, and not having a list right away?

CyberDem0n: Than it won't fit into 120 characters :)

erthalion: Indentation, a separate variable, so many other options.

erthalion:
:+1:

CyberDem0n:
:+1:

Issue #1185, opened 2019-09-20
patroni sets postgresql.conf to 600 even outside PGDATA


mbanck:
Debian (or rather pgapt, via https://redmine.postgresql.org/issues/4717 - might not be public) got an issue that patroni-1.6.0 now sets the `postgresql.conf` to 600 due to the `os.chmod()` in https://github.com/zalando/patroni/commit/0b1b1e3b542714b47634391f7dbb00016d885bff#diff-a534a7e0f2c46ba53349f820d9014468R286

I can verify that starting patroni over a Debian instance (with e.g. `/etc/postgresql/11/test/postgresq.conf`) renders that config file 600, while the same instance created via `pg_createcluster 11 test2` has it as 644:

```
 ls -l /etc/postgresql/11/test{2,}/postgresql.conf
-rw------- 1 postgres postgres   723 Sep 20 16:09 /etc/postgresql/11/test/postgresql.conf
-rw-r--r-- 1 postgres postgres 24070 Sep 20 16:10 /etc/postgresql/11/test2/postgresql.conf
```

That was part of a rather large patch so maybe unintentional? Maybe it was assumed that `postgresql.conf` is in PGDATA so should be 600 anyway? So how about checking `config_dir` in the YAML config to see whether that is set?

CyberDem0n:
Oh...

my idea behind this change was that starting from 12 postgresql.conf will contain primary_conninfo, which is sensitive.

It is not a big deal to avoid setting permission to 600 if the config_dir is different from data_dir, but there is a question though, what are you going to do on 12? How are you going to protect it?
I don't really like the idea of keep primary_conninfo in the postgresql.auto.conf...

mbanck:
Hrm good point, I've passed that on to the Debian postgresql maintainers.

CyberDem0n:
Maybe we can go with 640?

df7cb:
Use ~/.pgpass instead. The location is even configurable if you want to programmatically write the password to a file: `primary_conninfo = 'passfile=/some/where/else'`.

PR #1194, opened 2019-09-26
Use passfile in the primary_conninfo instead of password

CyberDem0n:
Fixed a few minor issues related to the #1134 and #1122
Close https://github.com/zalando/patroni/issues/1185


patroni/postgresql/__init__.py
```diff
@@ -645,12 +631,10 @@ def controldata(self):
...
     def get_replication_connection_cursor(self, host='localhost', port=5432, database=None, **kwargs):
-        replication = self.config.replication
-        extra_kwargs = {k: v for k, v in replication.items()
-                        if k not in ('username', 'password', 'connect_timeout', 'options')}
-        with get_connection_cursor(host=host, port=int(port), database=database or self._database, replication=1,
-                                   user=replication['username'], password=replication.get('password'),
-                                   connect_timeout=3, options='-c statement_timeout=2000', **extra_kwargs) as cur:
+        conn_kwargs = self.config.replication.copy()
+        conn_kwargs.update(host=host, port=int(port), database=database or self._database, connect_timeout=3,
+                           user=conn_kwargs.pop('username', None), replication=1, options='-c statement_timeout=2000')
```
erthalion: Looks like before lack of `username` key will cause an exception, which is not the case now. Is it intentional?

```diff
-user=replication['username']
+user=conn_kwargs.pop('username', None)
```

CyberDem0n: Replication user name is mandatory, so the answer is no.
It should be just `conn_kwargs.pop('username')`.

patroni/postgresql/config.py
```diff
@@ -300,28 +315,26 @@ def write_postgresql_conf(self, configuration=None):
         if 'custom_conf' not in self._config and not os.path.exists(self._postgresql_base_conf):
             os.rename(self._postgresql_conf, self._postgresql_base_conf)
 
-        with open(self._postgresql_conf, 'w') as f:
-            os.chmod(self._postgresql_conf, stat.S_IWRITE | stat.S_IREAD)
-            f.write(self._CONFIG_WARNING_HEADER)
-            f.write("include '{0}'\n\n".format(self._config.get('custom_conf') or self._postgresql_base_conf_name))
+        with ConfigWriter(self._postgresql_conf) as f:
+            include = self._config.get('custom_conf') or self._postgresql_base_conf_name
+            f.writeline("include '{0}'\n".format(ConfigWriter.escape(include)))
```
erthalion: There is already one `\n` inside `writeline` method.

patroni/postgresql/config.py
```diff
@@ -154,6 +154,37 @@ def mtime(filename):
...
+        self.writeline('# Do not edit this file manually!\n# It will be overwritten by Patroni!')
+        return self
+
+    def __exit__(self, exc_type, exc_val, exc_tb):
+        if self._fd:
+            self._fd.close()
+
+    def writeline(self, line):
+        self._fd.write(line)
+        self._fd.write('\n')
```
erthalion: Why not one write with `{line}\n`?

erthalion:
:+1:

patroni/postgresql/config.py
CyberDem0n: Originally there were two: https://github.com/zalando/patroni/pull/1194/files#diff-a534a7e0f2c46ba53349f820d9014468L306, therefore I am adding the second one.

patroni/postgresql/config.py
CyberDem0n: Just avoiding useless string concatenation.

erthalion:
:+1:

CyberDem0n:
:+1:

PR #1197, opened 2019-09-27
Refactor Patroni REST API communication

### LINKED ISSUE #898: InsecureRequestWarning on fetch_node_status when performing failover
Hello!

After testing one of the failover scenarios (power cut on primary node), the log output following warning:
```
/usr/lib/python2.7/site-packages/urllib3/connectionpool.py:847: InsecureRequestWarning: Unverified HTTPS request is being made. Adding certificate verification is strongly advised. See: https://urllib3.readthedocs.io/en/latest/advanced-usage.html#ssl-warnings
2018-12-19 14:06:20,805 INFO: Got response from db-03 https://db-03:8008/patroni: {"database_system_identifier": "6396680538100349567", "postmaster_start_time": "2018-11-23 11:50:01.615 CET", "timeline": 3, "cluster_unlocked": true, "patroni": {"scope": "tags", "version": "1.5.1"}, "state": "running", "role": "replica", "xlog": {"received_location": 207004628544, "replayed_timestamp": "2018-12-19 14:04:28.135 CET", "paused": false, "replayed_location": 207004628544}, "server_version": 90610}
InsecureRequestWarning)
```
The first and third line is the problematic warning. Warning itself seems to be harmless, however it taints the logs and makes the troubleshooting more confusing.

I've looked around the code and I believe this is what is called, when the warning is output.
```
    @staticmethod
    def fetch_node_status(member):
        """This function perform http get request on member.api_url and fetches its status
        :returns: `_MemberStatus` object
        """

        try:
            response = requests.get(member.api_url, timeout=2, verify=False)
            logger.info('Got response from %s %s: %s', member.name, member.api_url, response.content)
            return _MemberStatus.from_api_response(member, response.json())
        except Exception as e:
            logger.warning("Request failed to %s: GET %s (%s)", member.name, member.api_url, e)
        return _MemberStatus.unknown(member)
```

Is there any particular reason `verify=False` is left in instead of relying to the configured cacert entry in `restapi.cacert` ?

If we can not / do not want to use cert verification when checking other nodes cert in same cluster, then would it be possible to suppress this warning or have an option to do so ?

Thanks in advance!

CyberDem0n:
* make it possible to use client certificates with REST API
* define a separate PatroniRequest class which handles all communication
* refactor patronictl to use the new class
* make Ha to use the new class instead of calling requests.get. The old call wasn't taking into account certificates and basic-auth

Close #898


sdudoladov:
what about adding
```
 health-check endpoints on a different port without ssl.
```
and/or explicitly acknowledging potential complications in the docs ?

CyberDem0n:
That would be a different PR.

sdudoladov:
:+1:

CyberDem0n:
:+1:

Issue #1201, opened 2019-10-01
Patroni not marking "Pending restart" flag


RajKiranS:
We are using Patroni v1.6.0. We noticed that Patroni has stopped marking 'Pending restart' flag in this version.

For example, we changed the restore_command and it wasn't marked for 'pending restart' but the config in ZooKeeper was successfully updated and patroni.dynamic.json too.

We updated the config using `patronictl edit-config` option only.

```
---
+++
@@ -12,7 +12,7 @@
     wal_level: hot_standby
     wal_log_hints: true
   recovery_conf:
-    restore_command: get-wal %f %p
+   restore_command: /usr/bin/get-wal %f %p
   use_pg_rewind: true
   use_slots: true
 retry_timeout: 10

Apply these changes? [y/N]: y
Configuration changed
```
```
$ ag restore_command
**patroni.dynamic.json**
1:{"retry_timeout": 10, "postgresql": {"use_slots": true, "parameters": {"hot_standby": true, "wal_keep_segments": 16, "wal_level": "hot_standby", "archive_command": "exit 0", "wal_log_hints": true, "max_wal_senders": 5, "archive_mode": "on", "max_replication_slots": 5}, "use_pg_rewind": true, "recovery_conf": {"restore_command": "/usr/bin/get-wal %f %p"}}, "synchronous_mode": true, "maximum_lag_on_failover": 1048576, "ttl": 10, "master_start_timeout": 10, "loop_wait": 10}

**recovery.conf**
2:restore_command = 'get-wal %f %p'
```
```
+------------+--------------------------------------+--------------------------------------------+--------------+---------+----+-----------+-----------------+-------------------+
|  Cluster   |                Member                |                    Host                    |     Role     |  State  | TL | Lag in MB | **Pending restart** | Scheduled restart |
+------------+--------------------------------------+--------------------------------------------+--------------+---------+----+-----------+-----------------+-------------------+
| abc | xxx | xxx:1234 |    Leader    | running | 7 |           |                 |                   |
| abc | yyy  | yyy:1234  | Sync standby | running | 7 |           |                 |                   |
+------------+--------------------------------------+--------------------------------------------+--------------+---------+----+-----------+-----------------+-------------------+
```

Am I missing something here?

RajKiranS:
FTR: And the PostgeSQL version is 9.6

RajKiranS:
Also, I tested out changing `wal_keep_segments` and that seems to have been smoothly updated.

[...]

PR #1208, opened 2019-10-03
Compare all recovery parameters

CyberDem0n:
Previously check_recovery_conf() function was only checking whether primary_conninfo has changed and never taking into account all other recovery parameters.

Fixes https://github.com/zalando/patroni/issues/1201


CyberDem0n:
:+1:

PR #1215, opened 2019-10-07
Fix race conditions in async actions

CyberDem0n:
Specifically, there was a chance that `patronictl reinit --force` was overwritten by recover and we end up in a situation when Patroni was trying to start the postgres while basebackup still running.


sdudoladov:
:+1:

Issue #1216, opened 2019-10-09
initdb not being run after upgrade to 1.6.0


Lobo75:
I have a testing lab setup with three etdc servers and three postgresql/patroni servers on 1.5.6.  Part of normal testing is to completely reset the environment by removing the /var/lib/etcd/* directories in the cluster of the etdc servers and delete the data directory on the postgresql servers.  Then start patroni.  Normally patroni does the initdb and creates an empty db, life goes on. 

An unanticipated upgrade to patroni 1.6.0 occurred and now the cluster will not rebuild, it appears initdb is not being run (the yml files were not changed).  From the log below looks like patroni is confused as to the cluster status.  From looking at comments in the code it appears as if it's trying to get configuration from etdd but there is none so it fails on a KeyError.

Can you assist in how patroni would be looking for configuration on a cluster that does not exist yet or how the cluster status would be incorrect leading to that message indicating it's trying to start as a secondary?

2019-10-08 14:14:27,170 INFO: Failed to import patroni.dcs.consul
2019-10-08 14:14:27,189 INFO: Selected new etcd server http://10.x.x.x:2379
2019-10-08 14:14:27,203 INFO: No PostgreSQL configuration items changed, nothing to reload.
2019-10-08 14:14:27,216 WARNING: Postgresql is not running.
2019-10-08 14:14:27,216 INFO: Lock owner: None; I am postgresqlL0
2019-10-08 14:14:27,217 INFO: pg_controldata:

2019-10-08 14:14:27,219 INFO: Lock owner: None; I am postgresqlL0
2019-10-08 14:14:27,249 INFO: Lock owner: None; I am postgresqlL0
2019-10-08 14:14:27,255 INFO: starting as a secondary
2019-10-08 14:14:27,258 ERROR: Exception during execution of long running task restarting after failure
Traceback (most recent call last):
  File "/usr/local/lib/python3.6/site-packages/patroni/async_executor.py", line 97, in run
    wakeup = func(*args) if args else func()
  File "/usr/local/lib/python3.6/site-packages/patroni/postgresql/__init__.py", line 728, in follow
    self.start(timeout=timeout, block_callbacks=change_role, role=role)
  File "/usr/local/lib/python3.6/site-packages/patroni/postgresql/__init__.py", line 414, in start
    configuration = self.config.effective_configuration
  File "/usr/local/lib/python3.6/site-packages/patroni/postgresql/config.py", line 731, in effective_configuration
[...]

PR #1217, opened 2019-10-10
Improve workflow when PGDATA is not empty during bootstrap

CyberDem0n:
Recently it has happened two times when people tried to deploy the new cluster but postgres data directory wasn't empty and also wasn't valid. In this case Patroni was still creating initialize key in DCS and trying to start the postgres up.
Now it will complain about non-empty invalid postgres data directory and exit.

Close https://github.com/zalando/patroni/issues/1216

### DIFF
--- a/patroni/postgresql/__init__.py
+++ b/patroni/postgresql/__init__.py
@@ -200,13 +200,13 @@ class Postgresql(object):
 
     @property
     def sysid(self):
-        if not self._sysid and not self.bootstrapping or not self.is_running():
+        if not self._sysid and not self.bootstrapping:
             data = self.controldata()
             self._sysid = data.get('Database system identifier', "")
         return self._sysid
 
     def get_postgres_role_from_data_directory(self):
-        if self.data_directory_empty():
+        if self.data_directory_empty() or not self.controldata():
             return 'uninitialized'
         elif self.config.recovery_conf_exists():
             return 'replica'
@@ -255,7 +255,8 @@ class Postgresql(object):
             raise PostgresConnectionException(str(e))
 
     def data_directory_empty(self):
-        return not os.path.exists(self._data_dir) or os.listdir(self._data_dir) == []
+        return not os.path.exists(self._data_dir) or \
+                all(os.name != 'nt' and (n.startswith('.') or n == 'lost+found') for n in os.listdir(self._data_dir))
 
     def replica_method_options(self, method):
         return deepcopy(self.config.get(method, {}))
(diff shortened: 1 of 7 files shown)


patroni/postgresql/__init__.py
```diff
@@ -200,7 +200,7 @@ def set_pending_restart(self, value):
 
     @property
     def sysid(self):
-        if not self._sysid and not self.bootstrapping:
+        if not self._sysid and not self.bootstrapping or not self.is_running():
```
sdudoladov: Against which situation does the `or` clause protect ?

CyberDem0n: I'll look at it one more time tomorrow morning.

tests/test_ha.py
```diff
@@ -1027,7 +1027,6 @@ def test_update_cluster_history(self):
             self.assertEqual(self.ha.run_cycle(), 'no action.  i am the leader with the lock')
 
     @patch('sys.exit', return_value=1)
-    @patch('requests.get', requests_get)
```
sdudoladov: is it an unrelated change that simply removes a no longer needed mock ?

patroni/postgresql/__init__.py
```diff
@@ -255,7 +255,8 @@ def query(self, sql, *args, **kwargs):
             raise PostgresConnectionException(str(e))
 
     def data_directory_empty(self):
-        return not os.path.exists(self._data_dir) or os.listdir(self._data_dir) == []
+        return not os.path.exists(self._data_dir) or \
+                all(os.name != 'nt' and (n.startswith('.') or n == 'lost+found') for n in os.listdir(self._data_dir))
```
sdudoladov: 1. Moving the `os.name != 'nt'` out of `all()` would make it slightly more readable . I imagine most readers expect only an iterable as an argument passed to `all`.
2. Windows apparently [also has hidden files](https://stackoverflow.com/a/14063074) but I guess it is out of scope

CyberDem0n: 1. If you move `os.name != 'nt'` out of `all()` result would be wrong.
2. src/port/pgcheckdir.c: pg_check_dir() runs exaclty the same check.

tests/test_ha.py
CyberDem0n: Well, actually it is related. Because it tests the [code](https://github.com/zalando/patroni/pull/1217/files#diff-57c53e988b775ee2902488fb33c1f433R1287-R1294) I was changing.

sdudoladov:
:+1:

CyberDem0n:
:+1:

PR #1219, opened 2019-10-10
Build recovery params in a separate method

CyberDem0n:
In addition to that try to protect from the case when some recovery parameters are set in one of included files by explicitly setting their value to an empty string on postgres 12.

Simplifies https://github.com/zalando/patroni/pull/1208


sdudoladov:
:+1:

CyberDem0n:
:+1:

PR #1224, opened 2019-10-11
Only remove PATRONI_ENV_PREFIX'd environment variables

snooc:
I run PG clusters using Patroni that have an ODCB backed Foreign Data Wrapper to connect to about 30 different database technologies. Some drivers require environment variables to be set to work properly (It's pretty much the 1990 Wild Wild West). Since Patroni strips out all env variables except a small hard-coded list, I have to run a custom patched version of Patroni that modifies this list of variables.

I understand that Patroni has taken a vary conservative stance by stripping out all env variables except an explicit list since there may be secrets in some ENV vars, but I urge you to reconsider this and to only strip out known PATRONI_ environment variables. 

Typically when someone comes across Patroni, I think it's a fair assumption that they are already running Postgres directly or using pg_ctl. Neither of these tools strip out any environment variables.

Thanks for the consideration!

### DIFF
--- a/patroni/ctl.py
+++ b/patroni/ctl.py
@@ -23,7 +23,6 @@ import yaml
 
 from click import ClickException
 from contextlib import contextmanager
-from patroni.config import Config
 from patroni.dcs import get_dcs as _get_dcs
 from patroni.exceptions import PatroniException
 from patroni.postgresql import Postgresql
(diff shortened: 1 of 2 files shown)


sdudoladov:
:+1:

CyberDem0n:
:+1:

Issue #1237, opened 2019-10-18
Demotion On Raft Internal Error when Etcd cluster has quorum


bradnicholson:
Patroni version 1.5.6

We are seeing a sporadic issue when performing rolling restarts of our 3 member Etcd clusters.  The Etcd cluster is healthy at the time.  When we shut down one member, we see Patroni issue an immediate shutdown to the primary.  There are two healthy members that have quorum at the time, so we sort of expected it would reconnect to the the other ones:

Stack trace from Patroni:

```
 2019-10-18 15:32:16,239 ERROR:  
 Traceback (most recent call last): 
   File "/usr/local/lib/python3.5/dist-packages/patroni/dcs/etcd.py", line 348, in wrapper 
     retval = func(self, *args, **kwargs) is not None 
   File "/usr/local/lib/python3.5/dist-packages/patroni/dcs/etcd.py", line 547, in _update_leader 
     return self.retry(self._client.test_and_set, self.leader_path, self._name, self._name, self._ttl) 
   File "/usr/local/lib/python3.5/dist-packages/patroni/dcs/etcd.py", line 332, in retry 
     return self._retry.copy()(*args, **kwargs) 
   File "/usr/local/lib/python3.5/dist-packages/patroni/utils.py", line 254, in __call__ 
     return func(*args, **kwargs) 
   File "/usr/local/lib/python3.5/dist-packages/etcd/client.py", line 703, in test_and_set 
     return self.write(key, value, prevValue=prev_value, ttl=ttl) 
   File "/usr/local/lib/python3.5/dist-packages/etcd/client.py", line 500, in write 
     response = self.api_execute(path, method, params=params) 
   File "/usr/local/lib/python3.5/dist-packages/patroni/dcs/etcd.py", line 222, in api_execute 
     return self._handle_server_response(response) 
   File "/usr/local/lib/python3.5/dist-packages/etcd/client.py", line 987, in _handle_server_response 
     etcd.EtcdError.handle(r) 
   File "/usr/local/lib/python3.5/dist-packages/etcd/__init__.py", line 306, in handle 
     raise exc(msg, payload) 
 etcd.EtcdException: Raft Internal Error : etcdserver: server stopped 
 2019-10-18 15:32:16,241 ERROR: failed to update leader lock 
```

Found this searching for this issue as well:  https://github.com/zalando/patroni/issues/659#issuecomment-382701253

CyberDem0n:
[...]

PR #1241, opened 2019-10-19
Retry on raft internal error

CyberDem0n:
Fixes https://github.com/zalando/patroni/issues/1237

### DIFF
--- a/patroni/dcs/etcd.py
+++ b/patroni/dcs/etcd.py
@@ -25,6 +25,10 @@ from threading import Thread
 logger = logging.getLogger(__name__)
 
 
+class EtcdRaftInternal(etcd.EtcdException):
+    """Raft Internal Error"""
+
+
 class EtcdError(DCSError):
     pass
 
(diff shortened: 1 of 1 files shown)


patroni/dcs/etcd.py
```diff
@@ -596,3 +598,6 @@ def watch(self, leader_index, timeout):
             return super(Etcd, self).watch(None, timeout)
         finally:
             self.event.clear()
+
+
+etcd.EtcdError.error_exceptions[300] = EtcdRaftInternal
```
sdudoladov: What does this line achieve ?

CyberDem0n: It adds it [here](https://github.com/jplana/python-etcd/blob/master/src/etcd/__init__.py#L279)

sdudoladov:
:+1:

CyberDem0n:
:+1:

Soulou:
Could you plan a release soon including this fix?

Issue #1238, opened 2019-10-18
missing master pod label


necrolyte2:
There exists a condition where when the callback scripts are executed they can be executed multiple times in succession very close in time to each other. The current callback executor does not guarantee the order of the callbacks being executed and is not thread safe.

Due to this it is possible(and we have seen this quite often in our setup) that the role can be changed to `replica` and then to `master` in close succession which leads to only the `replica` role being set.

PR #1242, opened 2019-10-21
Kill callback child processes when it is necessary

CyberDem0n:
Not doing so makes it hard to implement callbacks in bash and eventually can lead to the situation when two callbacks are running at the same time. In case if we failed to kill the child process we will still wait for it to finish.

The same problem could happen with custom bootstrap, therefore we happen to kill the custom bootstrap process we also kill all child subprocesses.

Closes https://github.com/zalando/patroni/issues/1238

### DIFF
--- a/patroni/postgresql/callback_executor.py
+++ b/patroni/postgresql/callback_executor.py
@@ -1,7 +1,7 @@
 import logging
 
 from patroni.postgresql.cancellable import CancellableExecutor
-from threading import Event, Thread
+from threading import Condition, Thread
 
 logger = logging.getLogger(__name__)
 
@@ -13,20 +13,24 @@ class CallbackExecutor(CancellableExecutor, Thread):
         Thread.__init__(self)
         self.daemon = True
         self._cmd = None
-        self._callback_event = Event()
+        self._condition = Condition()
         self.start()
 
     def call(self, cmd):
         self._kill_process()
-        self._cmd = cmd
-        self._callback_event.set()
+        with self._condition:
+            self._cmd = cmd
+            self._condition.notify()
 
     def run(self):
         while True:
-            self._callback_event.wait()
-            self._callback_event.clear()
+            with self._condition:
+                if self._cmd is None:
+                    self._condition.wait()
+                cmd, self._cmd = self._cmd, None
+
             with self._lock:
-                if not self._start_process(self._cmd, close_fds=True):
+                if not self._start_process(cmd, close_fds=True):
                     continue
             self._process.wait()
             self._kill_children()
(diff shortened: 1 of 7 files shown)


patroni/postgresql/callback_executor.py
```diff
@@ -13,28 +14,59 @@ def __init__(self):
...
+        self._callback_event.set()
+
+    def _cancel(self):
         with self._lock:
-            if self._process and self._process.poll() is None:
+            if self._process and self._process.poll() is None and not self._process_children:
+                try:
+                    self._process_children = self._process.children(recursive=True)
+                except psutil.Error:
+                    pass
```
sdudoladov: please at least log the error

patroni/postgresql/callback_executor.py
```diff
@@ -13,28 +14,59 @@ def __init__(self):
...
                 try:
                     self._process.kill()
-                    logger.warning('Killed the old callback process because it was still running: %s', self._cmd)
-                except OSError:
+                    logger.warning('Killed the old callback process because it was still running: %s',
+                                   self._process_cmd)
+                except psutil.NoSuchProcess:
+                    return
+                except psutil.AccessDenied:
                     logger.exception('Failed to kill the old callback')
```
sdudoladov: sth like 'Failed to kill the old callback (access denied)'  would imo be  clearer in the logs

patroni/postgresql/callback_executor.py
```diff
@@ -13,28 +14,59 @@ def __init__(self):
...
+
+    def _wait(self):
+        self._process.wait()
+
+        waitlist = []
+        with self._lock:
+            for child in self._process_children:
+                if child.is_running():
+                    try:
+                        child.kill()
```
sdudoladov: why directly kill and not terminate followed by a kill later ? that can (at least in theory) enable callback children to terminate gracefully

CyberDem0n: We really want to kill them, as fast as possible.

patroni/postgresql/cancellable.py
```diff
@@ -10,13 +11,66 @@
...
+            self._process = psutil.Popen(cmd, *args, **kwargs)
+        except Exception:
+            return logger.exception('Failed to execute %s',  cmd)
+        return True
+
+    def _kill_process(self):
+        with self._lock:
+            if self._process is not None and self._process.is_running() and not self._process_children:
+                try:
+                    self._process.suspend()  # Suspend the process before getting list of childrens
```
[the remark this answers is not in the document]
CyberDem0n: The callback process might spawn childs. If it is suspended, it won't be able to do that.
Our goal to make sure that we killed not only the callback process, but also all children.

sdudoladov:
:+1:

CyberDem0n:
:+1:

PR #1243, opened 2019-10-23
Fix race condition in postmaster_start_time()

CyberDem0n:
when it is executed not from the main thread we need to create a new cursor object.

### DIFF
--- a/patroni/postgresql/__init__.py
+++ b/patroni/postgresql/__init__.py
@@ -728,9 +728,12 @@ class Postgresql(object):
 
     def postmaster_start_time(self):
         try:
-            cursor = self.query("SELECT pg_catalog.to_char(pg_catalog.pg_postmaster_start_time(),"
-                                " 'YYYY-MM-DD HH24:MI:SS.MS TZ')")
-            return cursor.fetchone()[0]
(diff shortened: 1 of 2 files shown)


patroni/postgresql/__init__.py
```diff
@@ -728,9 +728,12 @@ def timeline_wal_position(self):
 
     def postmaster_start_time(self):
         try:
-            cursor = self.query("SELECT pg_catalog.to_char(pg_catalog.pg_postmaster_start_time(),"
-                                " 'YYYY-MM-DD HH24:MI:SS.MS TZ')")
-            return cursor.fetchone()[0]
+            query = "SELECT pg_catalog.to_char(pg_catalog.pg_postmaster_start_time(), 'YYYY-MM-DD HH24:MI:SS.MS TZ')"
+            if current_thread().ident == self.__thread_ident:
+                return self.query(query).fetchone()[0]
+            with self.connection().cursor() as cursor:
```
sdudoladov: can it be that yet another thread would be using the same cursor from `self._connection` if none of [these conditions ](https://github.com/zalando/patroni/blob/c0cb062118517de0caff92512b8384308fb6f192/patroni/postgresql/connection.py#L29) holds ?

CyberDem0n:
:+1:

Issue #1195, opened 2019-09-27
Patroni starts master demotion before retry_timeout exceeded


psuderevsky:
Patroni version: 1.6.0
PostgreSQL version: 11.5
Configuration
```
ttl=30
retry_timeout=10
loop_wait=10
```

**Test 1 - etcd cluster is unhealthy**
**Action:** `for i in {1..100000}; do date; pkill -KILL etcd ; sleep 0.5; done` 
**Nodes:** all
**Condition:** Leader TTL=23, etcd systemd service restart option set to "on-failure"
**Result:** master started demotion in 2-2 seconds, etcd service stopped trying restarting in 3 seconds
**Service interruption:** until manual etcd start 

**Test 2 - etcd cluster is unhealthy**
**Action:** `pkill -KILL etcd ; sleep 7; systemctl start etcd`
**Nodes:** all
**Condition:** Leader TTL=22, etcd systemd service restart option set to "no"
**Result:** master started demotion in 1-2 seconds, was demoted in 4 seconds, promoted back to master in 11 seconds(4 seconds after etcd start) 
**Service interruption:** 11 seconds

**Test 3 - network interface is down**
**Action:** `ifdown int_name && sleep 8 && ifup int_name`
**Nodes:** all
**Condition:** Leader TTL=22
**Result:** etcd leader has been re-elected after 10 seconds, master started shutting down after 3 seconds (duration = 9 seconds),promoted to master after 12 seconds
**Service interruption:** 12 seconds

**Documentations says:** _retry_timeout: timeout for DCS and PostgreSQL operation retries. DCS or network issues shorter than this will not cause Patroni to demote the leader. Default value: 10_
Actual demotion time with the default settings should be between 10 and 20 seconds.

**In fact:** Demotion performed right after _loop_wait_ exceeded, i.e. when ttl = (ttl - loop_wait).

Test 2 patroni log for example:
```
Sep 27 07:14:47 dbserver3 patroni[308623]: 2019-09-27 07:14:47,558 INFO: Lock owner: pg3; I am pg3
Sep 27 07:14:47 dbserver3 patroni[308623]: 2019-09-27 07:14:47,562 INFO: no action.  i am the leader with the lock
Sep 27 07:14:57 dbserver3 patroni[308623]: 2019-09-27 07:14:57,559 INFO: Lock owner: pg3; I am pg3
[...]

PR #1245, opened 2019-10-23
Don't give up on retry too early

CyberDem0n:
Fixes https://github.com/zalando/patroni/issues/1195


patroni/dcs/etcd.py
```diff
@@ -190,31 +216,51 @@ def api_execute(self, path, method, params=None, timeout=None):
...
+                while not response:
+                    response = self._do_http_request(request_executor, method, self._base_uri + path, **kwargs)
+
+                    if response is False:
+                        if not retry:
+                            raise etcd.EtcdException
+                        some_request_failed = True
+                if some_request_failed:
+                    self._refresh_machines_cache()
+                break
```
sdudoladov: does this `break` mark the happy case when a request retry succeeded ?

patroni/dcs/etcd.py
```diff
@@ -190,31 +216,51 @@ def api_execute(self, path, method, params=None, timeout=None):
...
+            except etcd.EtcdConnectionFailed as e:
+                if isinstance(e, etcd.EtcdWatchTimedOut):
```
sdudoladov: Does `EtcdWatchTimedOut` always imply a etcd problem thus justifying `self._next_server()` or can there the case where the watched key simply does not change and that triggers the timeout ?

patroni/dcs/etcd.py
```diff
@@ -190,31 +216,51 @@ def api_execute(self, path, method, params=None, timeout=None):
...
-                raise
+        while True:
+            try:
+                some_request_failed = False
+                while not response:
+                    response = self._do_http_request(request_executor, method, self._base_uri + path, **kwargs)
+
+                    if response is False:
+                        if not retry:
+                            raise etcd.EtcdException
```
sdudoladov: please add the message to the exception

patroni/dcs/etcd.py
CyberDem0n: Yes, added the `if response` there

patroni/dcs/etcd.py
CyberDem0n: I moved it to the `_do_http_request` and left a comment

CyberDem0n:
1. In case of broken network connectivity Patroni was quickly going over list of etcd nodes and giving up much earlier than retry_timeout.
2. The `EtcdLeaderElectionInProgress` is handled by the retry [object](https://github.com/zalando/patroni/pull/1245/files#diff-b19d6b14b30078b89e33a7666e672cd2L318-L319).

sdudoladov:
:+1:

Issue #1252, opened 2019-10-25
No failover to sync_standby node when consul health check is critical on master


TestDrivenDevil:
Hi everyone, 

First I want to thank all contributors for this great software!

Issue I saw is related to issue [#1054](https://github.com/zalando/patroni/issues/1054). 

I was testing additional health check definitions in the patroni yaml file combined with synchronous_mode enabled and when health check became critical on my master node, no standby node would take over:
- master node is correctly demoted
- sync_standby node as defined in consul's KV store, is just logging "INFO: following a different leader because i am not the healthiest node" without any other info. 
- other standby node is also logging "INFO: following a different leader because i am not the healthiest node" (but that's expected behaviour)

To avoid you the burden of reproducing the problem, I did some investigation and found out following:
- my sync_standby's node name defined in patroni yaml file is camel cased: `name: HACluster5Node3`
- sync_standby value in consul's KV store is lower cased while leader value is just as defined in its patroni yaml file: `{"leader":"HACluster5Node1","sync_standby":"hacluster5node3"}`

If I'm not mistaken only the node defined as sync_standby can be be promoted but I guess check to know who's the sync_standby node is done on node name, which in this case does not match.  To prove if my assumption was right, I changed the name of HACluster5Node3 to lower case in patroni yaml file, restarted and then failover works as expected.

Btw, you can also reproduce the problem by stopping consul agent on the master node instead of defining an additional health check that becomes critical.

PR #1253, opened 2019-10-25
Return the real member name when picking the sync standby

CyberDem0n:
Before we returned it in the lower case, what was preventing such a standby from promoting due to the name comparison mismatch.

Fixes https://github.com/zalando/patroni/issues/1252


avaczi:
:+1:

CyberDem0n:
:+1:

PR #1254, opened 2019-10-25
Avoid calling expensive os.listdir()

CyberDem0n:
When the system is under IO stress, `os.listdir()` could take a few seconds (or even minutes) to execute what is badly affecting the HA loop of Patroni and could even cause the leader key to disappear from DCS due to the lack of updates.

There is a better and less expensive way to check that the PGDATA is not empty. Instead of doing the `os.listdir` we simply check the presence of the `global/pg_control` file in it.


FxKu:
:+1:

avaczi:
:+1:

PR #1260, opened 2019-10-30
Make it possible to apply some recovery params without restart

CyberDem0n:
Starting from PostgreSQL 12 the following recovery parameters could be changed without restart, but Patroni didn't yet support it:
* archive_cleanup_command
* promote_trigger_file
* recovery_end_command
* recovery_min_apply_delay

In future postgres releases this list will be extended and Patroni will support it automatically.


Jan-M:
:+1:

CyberDem0n:
:+1:

Issue #1160, opened 2019-08-28
Postgresql is up and running but patronictl shows an error


anikin-aa:
Hi guys

Faced next issue on one of our environments.

patronictl list shows, that for two members State is `Start failed`, but postgresql itself is up and running, and logs are clear, no errors .

What is the possible RC for this ? I will collect all of the logs, if needed.

CyberDem0n:
1. please run ps and find the time when the postmaster was started and when the checkpointer process was started. Are they the same or checkpointer was started much later?
2. look into patroni and postgres logs. We are interested in timestamps when checkpointer was started. Is there anything interesting?

anikin-aa:
OK, i will comeback when i collect more info

anikin-aa:
```
sh-4.2$ patronictl -c /patroni/pg_node.yml list common
+---------+-------------------------+--------------+--------+--------------+----+-----------+
| Cluster |          Member         |     Host     |  Role  |    State     | TL | Lag in MB |
+---------+-------------------------+--------------+--------+--------------+----+-----------+
|  common | pg-common-node1-1-5lb5j | 10.130.1.220 | Leader | start failed |    |   unknown |
|  common | pg-common-node2-1-ccx4j | 10.130.1.221 |        | start failed |    |   unknown |
+---------+-------------------------+--------------+--------+--------------+----+-----------+
```
ps aux on master:
```
USER        PID %CPU %MEM    VSZ   RSS TTY      STAT START   TIME COMMAND
postgres      1  0.0  0.0  11688  1624 ?        Ss   07:44   0:00 bash /start.sh
postgres     39  0.4  0.2 485440 47412 ?        Sl   07:44   1:24 python ./patroni.py /patroni/pg_node.yml
postgres     62  0.0  0.0  11820  1416 ?        Ss+  08:01   0:00 /bin/sh
postgres     70  0.0  0.1 236116 23652 ?        S    08:15   0:01 /usr/pgsql-11/bin/postgres -D /var/lib/pgsql/data/postgresql_node1 --config-file=/var/lib/pgsql/data/postgresql_node1/postgresql.conf --hot_standby=on --listen_addresses=0.0.0.0 --max_worker_processes=8 --max_locks_per_transaction=64 --wal_level=replica --cluster_name=common--wal_log_hints=on --max_wal_senders=10 --track_commit_timestamp=off --max_prepared_transactions=200 --port=5432 --max_replication_slots=10 --max_connections=200
[...]

PR #1262, opened 2019-11-01
Fix 'start failed' issue

CyberDem0n:
The start of postgres happens in two stages:
1. First Patroni is waiting for postgres port to be open
2. After that, it is waiting for postgres starts to accept connections

There is a default timeout 60 seconds for both stages (in total).

When the port isn't open, pg_isready exits with code=2.
If postgres is rejecting connections due to recovery, exit code=1.

In most cases postgres quickly opens the port and pg_isready starts returning 1, but in rare cases the whole timeout could spend in `1.`
After that, the HA loop is still waiting for postgres to start, but executing only the check from `2.`. Since pg_isready exit code is still = 2, Patroni was falsely assuming that 'start failed' without taking into consideration the fact that the postmaster process is up and running.

Fixes https://github.com/zalando/patroni/issues/1160

### DIFF
--- a/patroni/postgresql/__init__.py
+++ b/patroni/postgresql/__init__.py
@@ -561,10 +561,12 @@ class Postgresql(object):
         if ready == STATE_REJECT:
             return False
         elif ready == STATE_NO_RESPONSE:
-            self.set_state('start failed')
-            self.slots_handler.schedule(False)  # TODO: can remove this?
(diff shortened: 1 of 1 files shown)


sdudoladov:
:+1:

CyberDem0n:
:+1:

PR #1266, opened 2019-11-05
Correctly document PATRONI_KUBERNETES_PORTS

feikesteenbergen:
The previous documentation was wrong and will throw the following error
when used:

        Exception when parsing list {[{"name": "postgresql", "port": 5432}]}

When removing the surrounding braces, the error goes away and the
endpoint is updated with the correct Port name.


CyberDem0n:
:+1:

hjacobs:
:+1:

PR #1284, opened 2019-11-18
Housekeeping

CyberDem0n:
* Implement proper tests for `multiprocessing.set_start_method()`
* Exclude some watchdog code from coverage (it is used only for behave tests)
* properly use os.path.join for windows compatibility
* import DCS modules in `features/environment.py` on demand. It allows to run behave tests against chosen DCS without installing all dependencies.
* remove some unused behave code
* fix some minor issues in the dcs.kubernetes module

### DIFF
--- a/features/environment.py
+++ b/features/environment.py
@@ -1,11 +1,6 @@
 import abc
-import consul
 import datetime
-import etcd
-import kazoo.client
-import kazoo.exceptions
 import os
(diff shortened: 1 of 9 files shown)


sdudoladov:
:+1:

CyberDem0n:
:+1:

PR #1286, opened 2019-11-18
Avoid importing all DCS modules

CyberDem0n:
We will try to import only the module which has a configuration section.
I.e. if there is only zookeeper section in the config, Patroni will try to import only `patroni.dcs.zookeeper` and skip `etcd`, `consul`, and `kubernetes`.
This approach has two benefits:
1. When there are no dependencies installed Patroni was showing INFO messages `Failed to import smth`, which looks scary.
2. It reduces memory usage, because sometimes dependencies are heavy.

### DIFF
--- a/patroni/dcs/__init__.py
+++ b/patroni/dcs/__init__.py
@@ -89,17 +89,17 @@ def get_dcs(config):
                         return item(config[name])
             except ImportError:
                 logger.debug('Failed to import %s', module_name)
-    else:
-        available_implementations = []
-        for module_name in modules:
-            name = module_name.split('.')[-1]
-            try:
-                module = importlib.import_module(module_name)
-                available_implementations.extend(name for key, item in module.__dict__.items() if key.lower() == name
-                                                 and inspect.isclass(item) and issubclass(item, AbstractDCS))
-            except ImportError:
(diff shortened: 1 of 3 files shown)


patroni/dcs/__init__.py
```diff
@@ -73,26 +73,34 @@ def dcs_modules():
...
+    for module_name in modules:
+        name = module_name.split('.')[-1]
+        try:
+            module = importlib.import_module(module_name)
+            available_implementations.extend(name for key, item in module.__dict__.items() if key.lower() == name
+                                             and inspect.isclass(item) and issubclass(item, AbstractDCS))
         except ImportError:
-            if not config.get('patronictl'):
-                logger.info('Failed to import %s', module_name)
+            logger.info('Failed to import %s', module_name)
```
sdudoladov: why is it `info` and not `debug` here ? would not the problem be obvious from the content of `available_implementations` ?

CyberDem0n: 1. for example `zookeper` will not be in the list of available implementations if the `python-kazoo` module is not installed. Debug will hide this fact.
2. Before it was also the `info()`

sdudoladov:
:+1:

CyberDem0n:
:+1:

Issue #1283, opened 2019-11-16
patroni 1.6.1 now only tries to connect to ETCD_ADVERTISE_CLIENT_URLS


mbanck:
My `etcd` section is as follows:

```
etcd: 
  hosts: 
    - 10.0.3.223:2379
```
Since ugprading to 1.6.1, patroni tries once to connect to the remote etcd and then continuously tries to connect to the `ETCD_ADVERTISE_CLIENT_URLS` which was `http://localhost:2379, http://localhost:4001`:

```
2019-11-16 12:31:52,554 INFO: Failed to import patroni.dcs.consul
2019-11-16 12:31:52,563 DEBUG: New etcd client created for http://127.0.0.1:4001
2019-11-16 12:31:52,563 DEBUG: Selection next machine in cache. Available machines: ['http://10.0.3.223:2379']
2019-11-16 12:31:52,564 INFO: Selected new etcd server http://10.0.3.223:2379
2019-11-16 12:31:52,564 DEBUG: Converted retries value: 2 -> Retry(total=2, connect=None, read=None, redirect=0, status=None)
2019-11-16 12:31:52,564 DEBUG: Starting new HTTP connection (1): 10.0.3.223:2379
2019-11-16 12:31:52,565 DEBUG: http://10.0.3.223:2379 "GET /v2/machines HTTP/1.1" 200 44
2019-11-16 12:31:52,565 DEBUG: Retrieved list of machines: ['http://localhost:2379', 'http://localhost:4001']
2019-11-16 12:31:52,565 DEBUG: Selection next machine in cache. Available machines: ['http://localhost:2379', 'http://localhost:4001']
2019-11-16 12:31:52,565 INFO: Selected new etcd server http://localhost:4001
2019-11-16 12:31:52,575 DEBUG: Issuing read for key /postgresql-common/11-test/ with args {'recursive': True, 'retry': <patroni.utils.Retry object at 0x7f6fa4cb1ad0>}
2019-11-16 12:31:52,575 DEBUG: Converted retries value: 1 -> Retry(total=1, connect=None, read=None, redirect=0, status=None)
2019-11-16 12:31:52,575 DEBUG: Starting new HTTP connection (1): localhost:4001
2019-11-16 12:31:52,575 DEBUG: Incremented Retry for (url='/v2/keys/postgresql-common/11-test/?recursive=true'): Retry(total=0, connect=None, read=None, redirect=0, status=None)
[...]

PR #1289, opened 2019-11-20
Added comments to provide example of using multiple etcd endpoints

yanchenko-igor:
Hey, I would like to provide some examples of using multiple etcd endpoints.

### DIFF
--- a/postgres0.yml
+++ b/postgres0.yml
@@ -17,6 +17,7 @@ restapi:
 #   cacert: /etc/ssl/certs/ssl-cacert-snakeoil.pem
 
 etcd:
+  #Provide host to do the initial discovery of the cluster topology:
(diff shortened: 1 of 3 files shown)


CyberDem0n:
If we go this way, it would also make sense to specify there that `host` and `hosts` are just used for the initial discovery of etcd cluster topology. Once discovery is complete Patroni will use the list of advertised `clientURLs`: https://github.com/zalando/patroni/issues/1283

It is possible to change this behavior through by setting `use_proxies: true`.
Documentation: https://github.com/zalando/patroni/blob/master/docs/SETTINGS.rst#etcd

sdudoladov:
:+1:

CyberDem0n:
:+1:

PR #1290, opened 2019-11-20
striping extra spaces if etcd.hosts is written with spaces i.e. "host1:port1, host2:port2"

yanchenko-igor:
patroni fails to connect to the etcd server if config looks like this:
```
etcd:
    hosts: host1:port1, host2:port2
```
it can be easily fixed with striping strings after split.


FxKu:
:+1:

CyberDem0n:
:+1:

PR #1291, opened 2019-11-20
print version if requested

yanchenko-igor:
Would be helpful to print version if option "--version" specified.

### DIFF
--- a/patroni/__init__.py
+++ b/patroni/__init__.py
@@ -5,16 +5,16 @@ import sys
 import time
 import argparse
 
-import patroni.postgresql as postgresql
-import patroni.dcs as dcs
-import patroni.async_executor as async_executor
-import patroni.ha as ha
-import patroni.config as config
-import patroni.watchdog as watchdog
 from patroni.api import RestApiServer
 from patroni.log import PatroniLogger
 from patroni.request import PatroniRequest
 from patroni.version import __version__
+postgresql = __import__('patroni.postgresql', fromlist=[""])
+dcs = __import__('patroni.dcs', fromlist=[""])
+async_executor = __import__('patroni.async_executor', fromlist=[""])
+ha = __import__('patroni.ha', fromlist=[""])
+config = __import__('patroni.config', fromlist=[""])
+watchdog = __import__('patroni.watchdog', fromlist=[""])
 
 logger = logging.getLogger(__name__)
 
@@ -168,10 +168,17 @@ class Patroni(object):
 
 def patroni_main():
     parser = argparse.ArgumentParser()
-    parser.add_argument('--version', action='version',
-                                version='%(prog)s {version}'.format(version=__version__))
-    parser.add_argument("configfile", nargs="?", default="", help="Patroni may also read the configuration from the {0} environment variable"
-            .format(config.Config.PATRONI_CONFIG_VARIABLE))
+    parser.add_argument(
+            '--version', action='version',
+            version='%(prog)s {version}'.format(version=__version__)
+            )
+    parser.add_argument(
+            "configfile",
+            nargs="?",
+            default="",
+            help="Patroni may also read the configuration from the {0} environment variable"
+            .format(config.Config.PATRONI_CONFIG_VARIABLE)
+            )
     args = parser.parse_args()
     conf = config.Config(args.configfile)
     patroni = Patroni(conf)
--- a/patroni/__init__.py
+++ b/patroni/__init__.py
@@ -4,6 +4,17 @@ import signal
 import sys
 import time
 
+import patroni.postgresql as postgresql
+import patroni.dcs as dcs
+import patroni.async_executor as async_executor
+import patroni.ha as ha
+import patroni.config as config
+import patroni.watchdog as watchdog
+from patroni.api import RestApiServer
+from patroni.log import PatroniLogger
+from patroni.request import PatroniRequest
+from patroni.version import __version__
+
 logger = logging.getLogger(__name__)
 
 PATRONI_ENV_PREFIX = 'PATRONI_'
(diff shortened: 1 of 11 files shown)


yanchenko-igor:
> You can get a version number from [patronictl](https://github.com/zalando/patroni/blob/master/patroni/ctl.py#L1166)

So we can safely assume they are the same?

kostiantyn-nemchenko:
> So we can safely assume they are the same?

Yes

yanchenko-igor:
Just a point to consider: postgresql distribution (server and client) include 35 different tools (in my environment at least), each of them except postgresql-check-db-dir can return version.

patroni/__init__.py
```diff
@@ -23,6 +23,9 @@ def __init__(self):
         self.setup_signal_handlers()
 
         self.version = __version__
+        if len(sys.argv) >= 2 and sys.argv[1] and sys.argv[1] == "--version":
```
CyberDem0n: IMO this is too hacky.
Such stuff should be done in the `main()` or in the `patroni_main()`.
Also, the check `and sys.argv[1]` seems to be unnecessary.

yanchenko-igor: Yeah, I was thinking about the same, there is no single entry point for parsing the arguments, I copied this part from Config class, the better solution I was thinking about to rewrite the whole config/environment variables/arguments completely, but unless I have certain goals it seems too big task to solve only version output.

patroni/__init__.py
```diff
@@ -164,7 +167,28 @@ def shutdown(self):
 
 
 def patroni_main():
-    patroni = Patroni()
+    config = __import__('patroni.config', fromlist=[""])
```
CyberDem0n: Why not just `from .config import Config`?

patroni/__init__.py
```diff
@@ -164,7 +167,28 @@ def shutdown(self):
...
+            "configfile",
+            nargs="?",
+            default="",
+            help="Patroni may also read the configuration from the {0} environment variable"
+            .format(config.Config.PATRONI_CONFIG_VARIABLE)
+            )
+    args = parser.parse_args()
+    try:
+        conf = config.Config(args.configfile)
+    except ConfigParseError as e:
```
CyberDem0n: If the config file is broken/malformed there might be some other exceptions, for example:
```python
    self.peek_token().start_mark)
yaml.parser.ParserError: expected '<document start>', but found '<block mapping start>'
  in "postgres1.yml", line 3, column 1
```

patroni/__init__.py
```diff
@@ -164,7 +167,28 @@ def shutdown(self):
...
+        if e.value:
+            print(e.value)
+        parser.print_help()
+        exit(1)
+    patroni = Patroni(conf)
```
CyberDem0n: Patroni could also raise exceptions related to config, for example when initializing DCS module:
```python
  File "/home/akukushkin/git/patroni/patroni/dcs/__init__.py", line 549, in __init__
    self._base_path = re.sub('/+', '/', '/'.join(['', config.get('namespace', 'service'), config['scope']]))
KeyError: 'scope'
```

Looks like the AbstractDCS should also produce some nice readable exceptions.

CyberDem0n: Postgresql and RestApiServer could also raise some exceptions releated to the config.

CyberDem0n: Some stuff could be validated before creation of the Patroni, but unfortunately the config is not complete yet.
The final initialization of the config is done only in the [load_dynamic_configuration()](https://github.com/zalando/patroni/pull/1291/files#diff-0996184a693191991bd4f56d522c4f32R47) method.

tests/test_config.py
```diff
@@ -18,7 +19,7 @@ def setUp(self):
         self.config = Config()
 
     def test_no_config(self):
-        self.assertRaises(SystemExit, Config)
+        self.assertRaises(SystemExit, patroni_main)
```
CyberDem0n: That's weird, we should not test the `patroni_main()` here

patroni/config.py
```diff
@@ -75,26 +75,27 @@ class Config(object):
         }
     }
 
-    def __init__(self):
+    def __init__(self, configfile=""):
```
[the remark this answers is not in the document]
yanchenko-igor: So what do we pass is there is no config file?

patroni/config.py
```diff
@@ -75,26 +75,27 @@ class Config(object):
...
             self._local_configuration = config_env and yaml.safe_load(config_env) or self.__environment_configuration
             if not self._local_configuration:
-                print('Usage: {0} config.yml'.format(sys.argv[0]))
-                print('\tPatroni may also read the configuration from the {0} environment variable'.
-                      format(self.PATRONI_CONFIG_VARIABLE))
-                sys.exit(1)
+                raise ConfigParseError(None)
 
         self.__effective_configuration = self._build_effective_configuration({}, self._local_configuration)
+        if 'postgresql' in self._local_configuration and "data_dir" in self._local_configuration["postgresql"]:
```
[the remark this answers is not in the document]
yanchenko-igor: in that case it shows `pg_basebackup: error: could not create directory "": No such file or directory`

patroni/ctl.py
```diff
@@ -73,16 +73,11 @@ def load_config(path, dcs):
...
-            for p in ('PATRONI_RESTAPI_LISTEN', 'PATRONI_POSTGRESQL_DATA_DIR'):
-                if p not in os.environ:
-                    os.environ[p] = '.'
-        config = Config().copy()
-    finally:
-        sys.argv = old_argv
+    if Config.PATRONI_CONFIG_VARIABLE not in os.environ:
+        for p in ('PATRONI_RESTAPI_LISTEN', 'PATRONI_POSTGRESQL_DATA_DIR'):
+            if p not in os.environ:
+                os.environ[p] = '.'
```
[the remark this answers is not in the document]
yanchenko-igor: removed

patroni/config.py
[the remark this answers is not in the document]
CyberDem0n: We always pass something, even when there is no config (None).

patroni/config.py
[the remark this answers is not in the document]
CyberDem0n: We have to perform a bunch of check later anyway, it could be one of them.

patroni/config.py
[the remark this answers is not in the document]
yanchenko-igor: ok, then I will change everywhere

patroni/__init__.py
yanchenko-igor: Do we want to do it in this PR? we didn't bother to catch those errors before. We can catch errors but it seems out of scope of this task.

patroni/__init__.py
yanchenko-igor: fixed, at some point it didn't work, maybe when there was a circular import

tests/test_config.py
yanchenko-igor: moved to test_patroni.py

FxKu:
:+1:

CyberDem0n:
:+1:

PR #1292, opened 2019-11-21
Produce readable exception if config file is invalid.

yanchenko-igor:
If config file doesn't have "postgresql" section, it generates a KeyError exception which is not readable for people who don't know python.
Also added data/ in .gitignore file.


yanchenko-igor:
merged into #1291

Issue #863, opened 2018-11-13
Create stats_temp_directory during postmaster start


mbanck:
Debian/Ubuntu by default put `stats_temp_directory` under `(/var)/run/postgresql/` which is a ram disk. This directory will be gone between reboots, so Debian's `pg_ctlcluster` creates it before starting postgres.

As patroni is starting postgres, it would be nice if there was a way to similarly tell it to create this directory, as Postgres fails to start if it is no present. It could be done via `RuntimeDirectory=` in patroni's systemd service, but that looks like the wrong place for creating a child's runtime directory.

Does patroni check the PostgreSQL configuration before starting postres in any other way for things to keep in mind? I don't think so, but if it does, it could be hooked in there.

Otherwise, a new option to simply create that/a directory might be ok?

Another option might be to use `on_start` (if I understand that directly) and run `mkdir -p` there, but at least superficial tests did not work for me and there was no log output to figure out what was going wrong. Besides, users might want to put other things in that hook.

CyberDem0n:
This is a very tricky question.

Do we have a guaranty that `(/var)/run/postgresql/` is owned by postgres? If it is not, we wont be able to create a directory there.
Unfortunately we can't use `on_start` hook, it is executed only when postgres already accepting connections.
ATM Patroni is not really checking all postgres parameters on start, it does it only if something has changed in runtime by querieing `pg_settings`. Verifying postgres configuration is the important thing which should actually be implemented, but we are not yet there.

The easiest thing to get temp dir would be using the `RuntimeDirectory=` in the patroni systemd.
We could also think about implementing a new patroni functionality, like creating `stats_temp_directory` if it is defined in the patroni config, but it won't cover the case when the parameter is defined in the postgres config. Parsing postgresql.conf and pg_hba.conf was actually the thing I was insistently avoiding.

mbanck:
I am using `RuntimeDirectory=` now, but one issue with that is that `RuntimeDirectory` only recently added support for subdirectories (as is required here); notably this does not work on Debian stable.

If you agree to support this in the Patroni yaml config, I think a reasonable thing to do might be to try creating the directory, but to only log a warning if it fails. Or to put it another way: I think if the user requests the creation of that subdirectory, they should be responsible for the postgres user to be able to create it.

PR #1293, opened 2019-11-21
make sure that unix_socket_directories exist, otherwise server won't start

yanchenko-igor:
If configuration has unix_socket_directories which don't exist server doesn't start, I would give it a try to create those directories.

### DIFF
--- a/patroni/postgresql/__init__.py
+++ b/patroni/postgresql/__init__.py
@@ -60,6 +60,16 @@ class Postgresql(object):
         self._pending_restart = False
         self._connection = Connection()
         self.config = ConfigHandler(self, config)
+        if "unix_socket_directories" in self.config._server_parameters:
+            for d in self.config._server_parameters["unix_socket_directories"].split(","):
+                d = os.path.join(self._data_dir, d.strip())
+                if not d.startswith(self._data_dir) or not self.data_directory_empty():
+                    directory_exists_or_create(d,
+                            "'{}' is defined in unix_socket_directories, but it is not a directory")
+        if "stats_temp_directory" in self.config._server_parameters:
+            d = os.path.join(self._data_dir, self.config._server_parameters["stats_temp_directory"])
+            if not d.startswith(self._data_dir) or not self.data_directory_empty():
+                directory_exists_or_create(d, "'{}' is defined in stats_temp_directory, but it is not a directory")
 
         self._bin_dir = config.get('bin_dir') or ''
         self.bootstrap = Bootstrap(self)
--- a/patroni/postgresql/__init__.py
+++ b/patroni/postgresql/__init__.py
@@ -407,6 +407,12 @@ class Postgresql(object):
         self._pending_restart = False
 
         configuration = self.config.effective_configuration
+        if "unix_socket_directories" in configuration:
+            for d in configuration["unix_socket_directories"].split(","):
+                if not d.startswith('/'):
+                    d = os.path.join(self._data_dir, d)
+                if not os.path.isdir(d):
+                    os.makedirs(d)
         self.config.write_postgresql_conf(configuration)
         self.config.resolve_connection_addresses()
         self.config.replace_pg_hba()
--- a/patroni/postgresql/__init__.py
+++ b/patroni/postgresql/__init__.py
@@ -17,7 +17,7 @@ from patroni.postgresql.misc import parse_history, postgres_major_version_to_int
 from patroni.postgresql.postmaster import PostmasterProcess
 from patroni.postgresql.slots import SlotsHandler
 from patroni.exceptions import PostgresConnectionException
-from patroni.utils import Retry, RetryFailedError, polling_loop, directory_exists_or_create
+from patroni.utils import Retry, RetryFailedError, polling_loop
 from threading import current_thread, Lock
 
 
@@ -60,16 +60,7 @@ class Postgresql(object):
         self._pending_restart = False
         self._connection = Connection()
         self.config = ConfigHandler(self, config)
-        if "unix_socket_directories" in self.config._server_parameters:
-            for d in self.config._server_parameters["unix_socket_directories"].split(","):
-                d = os.path.join(self._data_dir, d.strip())
-                if not d.startswith(self._data_dir) or not self.data_directory_empty():
-                    directory_exists_or_create(d,
-                            "'{}' is defined in unix_socket_directories, but it is not a directory")
-        if "stats_temp_directory" in self.config._server_parameters:
-            d = os.path.join(self._data_dir, self.config._server_parameters["stats_temp_directory"])
-            if not d.startswith(self._data_dir) or not self.data_directory_empty():
-                directory_exists_or_create(d, "'{}' is defined in stats_temp_directory, but it is not a directory")
+        self.config.check_directories()
 
         self._bin_dir = config.get('bin_dir') or ''
         self.bootstrap = Bootstrap(self)
@@ -417,13 +408,7 @@ class Postgresql(object):
         self._pending_restart = False
 
         configuration = self.config.effective_configuration
-        if "unix_socket_directories" in configuration:
-            for d in configuration["unix_socket_directories"].split(","):
-                d = os.path.join(self._data_dir, d.strip())
-                directory_exists_or_create(d, "'{}' is defined in unix_socket_directories, but it is not a directory")
-        if "stats_temp_directory" in configuration:
-            d = os.path.join(self._data_dir, configuration["stats_temp_directory"])
-            directory_exists_or_create(d, "'{}' is defined in stats_temp_directory, but it is not a directory")
+        self.config.check_directories()
         self.config.write_postgresql_conf(configuration)
         self.config.resolve_connection_addresses()
         self.config.replace_pg_hba()
--- a/patroni/postgresql/__init__.py
+++ b/patroni/postgresql/__init__.py
@@ -414,7 +414,18 @@ class Postgresql(object):
                     try:
                         os.makedirs(d)
                     except FileExistsError:
-                        raise PatroniException(f"'{d}' is defined in unix_socket_directories, but it is not a directory")
+                        raise PatroniException(
+                                "'%s' is defined in unix_socket_directories, but it is not a directory".format(d)
+                                )
+        if "stats_temp_directory" in configuration:
+            d = os.path.join(self._data_dir, configuration["stats_temp_directory"])
+            if not os.path.isdir(d):
+                try:
+                    os.makedirs(d)
+                except FileExistsError:
+                    raise PatroniException(
+                            "'%s' is defined in stats_temp_directory, but it is not a directory".format(d)
+                            )
         self.config.write_postgresql_conf(configuration)
         self.config.resolve_connection_addresses()
         self.config.replace_pg_hba()
--- a/patroni/postgresql/__init__.py
+++ b/patroni/postgresql/__init__.py
@@ -16,8 +16,8 @@ from patroni.postgresql.connection import Connection, get_connection_cursor
 from patroni.postgresql.misc import parse_history, postgres_major_version_to_int
 from patroni.postgresql.postmaster import PostmasterProcess
 from patroni.postgresql.slots import SlotsHandler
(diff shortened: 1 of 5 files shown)


patroni/postgresql/__init__.py
```diff
@@ -407,6 +407,12 @@ def start(self, timeout=None, task=None, block_callbacks=False, role=None):
         self._pending_restart = False
 
         configuration = self.config.effective_configuration
+        if "unix_socket_directories" in configuration:
+            for d in configuration["unix_socket_directories"].split(","):
+                if not d.startswith('/'):
```
CyberDem0n: 1. this is not compatible with windows
2. we have to `d.strip()`
2. `os.path.join()` is actually smart enough to do it unconditionally:
```python
>>> import os
>>> os.path.join('/foo', '/bar')
'/bar'
```

patroni/postgresql/__init__.py
```diff
@@ -407,6 +407,12 @@ def start(self, timeout=None, task=None, block_callbacks=False, role=None):
...
+                    d = os.path.join(self._data_dir, d)
+                if not os.path.isdir(d):
+                    os.makedirs(d)
```
CyberDem0n: `os.makedirs()` will fail if there is already something...

yanchenko-igor: Well, if there is already something but directory, the server won't start anyway, and I don't think we should resolve that automatically, the only thing we can do about it, make an additional check and give the user a hint.

patroni/postgresql/__init__.py
yanchenko-igor: fixed

patroni/postgresql/__init__.py
yanchenko-igor: added a try/except

patroni/postgresql/__init__.py
```diff
@@ -407,6 +407,14 @@ def start(self, timeout=None, task=None, block_callbacks=False, role=None):
...
 
         configuration = self.config.effective_configuration
+        if "unix_socket_directories" in configuration:
+            for d in configuration["unix_socket_directories"].split(","):
+                d = os.path.join(self._data_dir, d.strip())
+                if not os.path.isdir(d):
+                    try:
+                        os.makedirs(d)
+                    except FileExistsError:
+                        raise PatroniException(f"'{d}' is defined in unix_socket_directories, but it is not a directory")
```
CyberDem0n: No f strings please, we need to be compatible with python 2.7

patroni/postgresql/__init__.py
```diff
@@ -407,6 +407,14 @@ def start(self, timeout=None, task=None, block_callbacks=False, role=None):
         self._pending_restart = False
 
         configuration = self.config.effective_configuration
+        if "unix_socket_directories" in configuration:
```
CyberDem0n: Actually we should do the same for [stats_temp_directory](https://github.com/zalando/patroni/issues/863)

patroni/postgresql/__init__.py
```diff
@@ -407,6 +407,25 @@ def start(self, timeout=None, task=None, block_callbacks=False, role=None):
...
+            for d in configuration["unix_socket_directories"].split(","):
+                d = os.path.join(self._data_dir, d.strip())
+                if not os.path.isdir(d):
+                    try:
+                        os.makedirs(d)
+                    except FileExistsError:
+                        raise PatroniException(
+                                "'%s' is defined in unix_socket_directories, but it is not a directory".format(d)
+                                )
```
CyberDem0n: Lines 410-419 and 420-428 are very similar.
It would make sense to make the separate method that tries to create the directory and use it from both places.

patroni/postgresql/__init__.py
```diff
@@ -407,6 +407,25 @@ def start(self, timeout=None, task=None, block_callbacks=False, role=None):
         self._pending_restart = False
 
         configuration = self.config.effective_configuration
+        if "unix_socket_directories" in configuration:
+            for d in configuration["unix_socket_directories"].split(","):
+                d = os.path.join(self._data_dir, d.strip())
+                if not os.path.isdir(d):
+                    try:
+                        os.makedirs(d)
+                    except FileExistsError:
```
CyberDem0n: `FileExistsError` doesn't exist in python 2.7

patroni/postgresql/__init__.py
```diff
@@ -407,6 +407,25 @@ def start(self, timeout=None, task=None, block_callbacks=False, role=None):
...
+                        raise PatroniException(
+                                "'%s' is defined in unix_socket_directories, but it is not a directory".format(d)
```
CyberDem0n:
```diff
-"'%s' is defined 
+"'{0}' is defined 
```

patroni/postgresql/__init__.py
yanchenko-igor: Fixed

patroni/postgresql/__init__.py
yanchenko-igor: Fixed

patroni/postgresql/__init__.py
yanchenko-igor: Fixed

patroni/utils.py
```diff
@@ -416,3 +417,22 @@ def cluster_as_json(cluster):
         if cluster.failover.candidate:
             ret['scheduled_switchover']['to'] = cluster.failover.candidate
     return ret
+
+
+def directory_exists_or_create(d, msg="{} {}"):
+    try:
+        if not os.path.exists(d):
+            os.makedirs(d)
+        if os.path.isdir(d):
```
CyberDem0n: `elif`.
If the os.makedirs()` didn't fail there is no need to create files

patroni/utils.py
```diff
@@ -416,3 +417,22 @@ def cluster_as_json(cluster):
...
+            test_file = os.path.join(d, ".writable")
+            try:
+                test_writable = open(test_file, "w")
```
CyberDem0n: I would better use here:
```python
fd, tmpfile = tempfile.mkstemp(dir=d)
os.close(fd)
os.remove(tmpfile)
```

In case of problems it raises `OSError`.

patroni/utils.py
```diff
@@ -416,3 +417,22 @@ def cluster_as_json(cluster):
         if cluster.failover.candidate:
             ret['scheduled_switchover']['to'] = cluster.failover.candidate
     return ret
+
+
+def directory_exists_or_create(d, msg="{} {}"):
```
CyberDem0n: The function name doesn't match with things it is doing.

patroni/postgresql/config.py
```diff
@@ -347,6 +348,17 @@ def setup_server_parameters(self):
         self._server_parameters = self.get_server_parameters(self._config)
         self._adjust_recovery_parameters()
 
+    def check_directories(self):
+        if "unix_socket_directories" in self._server_parameters:
+            for d in self._server_parameters["unix_socket_directories"].split(","):
+                d = os.path.join(self._postgresql._data_dir, d.strip())
+                if not d.startswith(self._postgresql._data_dir) or not self._postgresql.data_directory_empty():
```
CyberDem0n: `startswith` will not work exactly as you expect
example:
```python
data_dir = '/foo/bar'
d = '/foo/bar_buz'

Moreover, things would become even worse when `data_dir` is relative, but `d` is absolute.

File: patroni/utils.py [hunk 12]
```diff
@@ -416,3 +418,25 @@ def cluster_as_json(cluster):
...
+def validate_directory(d, msg="{} {}"):
+    try:
+        if not os.path.exists(d):
+            os.makedirs(d)
+        elif os.path.isdir(d):
+            try:
+                fd, tmpfile = tempfile.mkstemp(dir=d)
+                os.close(fd)
+                os.remove(tmpfile)
+            except IOError:
```
CyberDem0n: `tempfile.mkstemp` is throwing OSError

patroni/utils.py
```diff
@@ -416,3 +418,25 @@ def cluster_as_json(cluster):
         if cluster.failover.candidate:
             ret['scheduled_switchover']['to'] = cluster.failover.candidate
     return ret
+
+def merge_paths(d1, d2):
+    real_d1 = os.path.realpath(d1)
+    real_d2 = os.path.realpath(os.path.join(real_d1, d2))
+    return (real_d2, (os.path.commonpath([real_d1, real_d2]) == real_d1))
```
CyberDem0n: `os.path.commonpath()` doesn't exist on 2.7

yanchenko-igor: fixed

patroni/utils.py
yanchenko-igor: fixed

patroni/utils.py
```diff
@@ -416,3 +418,27 @@ def cluster_as_json(cluster):
         if cluster.failover.candidate:
             ret['scheduled_switchover']['to'] = cluster.failover.candidate
     return ret
+
+
+def merge_paths(d1, d2):
```
CyberDem0n: Let's give it a name `is_subpath(d1, d2)` and make it do just one thing and return `bool`.

patroni/utils.py
```diff
@@ -416,3 +418,27 @@ def cluster_as_json(cluster):
...
+                fd, tmpfile = tempfile.mkstemp(dir=d)
+                os.close(fd)
+                os.remove(tmpfile)
+            except OSError:
+                raise OSError("{} is not writable.".format(d))
+        else:
+            raise OSError("{} is not a directory".format(d))
+    except OSError as e:
+        logger.error(e)
+        raise PatroniException(msg.format(d, e))
```
CyberDem0n: The message formatting will look ugly, because in some cases the `d` will be presented there twice.

patroni/utils.py
```diff
@@ -416,3 +418,27 @@ def cluster_as_json(cluster):
...
+    try:
+        if not os.path.exists(d):
+            os.makedirs(d)
+        elif os.path.isdir(d):
+            try:
+                fd, tmpfile = tempfile.mkstemp(dir=d)
+                os.close(fd)
+                os.remove(tmpfile)
+            except OSError:
+                raise OSError("{} is not writable.".format(d))
```
CyberDem0n: Why not just immediately raise the `PatroniException` with the correctly formatted message and without the directory name redundancy?

patroni/utils.py
```diff
@@ -416,3 +418,27 @@ def cluster_as_json(cluster):
...
+def merge_paths(d1, d2):
+    real_d1 = os.path.realpath(d1)
+    real_d2 = os.path.realpath(os.path.join(real_d1, d2))
+    return (real_d2, (os.path.commonprefix([real_d1+os.path.sep, real_d2+os.path.sep]) == real_d1+os.path.sep))
+
+
+def validate_directory(d, msg="{} {}"):
+    try:
+        if not os.path.exists(d):
+            os.makedirs(d)
```
CyberDem0n: We can just wrap it with try...except here, without wrapping the whole body.

patroni/postgresql/config.py
```diff
@@ -347,6 +348,17 @@ def setup_server_parameters(self):
...
+    def check_directories(self):
+        if "unix_socket_directories" in self._server_parameters:
+            for d in self._server_parameters["unix_socket_directories"].split(","):
+                d, is_subdir = merge_paths(self._postgresql._data_dir, d.strip())
+                if not is_subdir or not self._postgresql.data_directory_empty():
+                    validate_directory(d, "'{}' is defined in unix_socket_directories, {}")
+        if "stats_temp_directory" in self._server_parameters:
+            d, is_subdir = merge_paths(self._postgresql._data_dir, self._server_parameters["stats_temp_directory"])
+            if not is_subdir or not self._postgresql.data_directory_empty():
+                validate_directory(d, "'{}' is defined in stats_temp_directory, {}")
```
CyberDem0n: These lines are basically duplicating [354-356](https://github.com/zalando/patroni/pull/1293/files#diff-a534a7e0f2c46ba53349f820d9014468R354-R356).

FxKu:
:+1:

CyberDem0n:
:+1:

Issue #391, opened 2017-02-02
cmdline flag to generate a default config.yaml


kmoppel:
Hey!
Not a big problem of course but currently when you do "pip install patroni" you still need to go to Github and save a copy of for https://github.com/zalando/patroni/blob/master/postgres0.yml to get started. I know there are env variables but there are quite some to remember. So would be very convenient if running something like "patroni --initconf postgres0.yml" would fill the postgres0.yml with all the config settings available (or would just download https://github.com/zalando/patroni/blob/master/postgres0.yml) and rename. What do you think?
Cheers

PR #1294, opened 2019-11-21
Show warning if default configuration does not exist

yanchenko-igor:
When default configuration doesn't exist patronictl returns confusing messages, which don't help to understand what is the issue, a little hint if default configuration does not exist will be very helpful.

### DIFF
--- a/patroni/ctl.py
+++ b/patroni/ctl.py
@@ -70,6 +70,10 @@ def load_config(path, dcs):
 
     if not (os.path.exists(path) and os.access(path, os.R_OK)):
         logging.debug('Ignoring configuration file "%s". It does not exists or is not readable.', path)
+        if path == CONFIG_FILE_PATH:
+            ctx = click.get_current_context()
+            logging.warning('Default configuration file "%s" does not exist or is not readable, '\
+                    'try running "%s configure"', path, ctx.command_path)
     else:
         logging.debug('Loading configuration from file %s', path)
     config = {}
--- a/patroni/ctl.py
+++ b/patroni/ctl.py
@@ -96,12 +96,12 @@ def load_config(path, dcs):
     return config
 
 
-def store_config(config, path):
-    dir_path = os.path.dirname(path)
-    if dir_path and not os.path.isdir(dir_path):
-        os.makedirs(dir_path)
-    with open(path, 'w') as fd:
-        yaml.dump(config, fd)
+def store_config(path):
+    if CONFIG_DIR_PATH and not os.path.isdir(CONFIG_DIR_PATH):
+        os.makedirs(CONFIG_DIR_PATH)
+    if os.path.exists(CONFIG_FILE_PATH) or os.path.islink(CONFIG_FILE_PATH):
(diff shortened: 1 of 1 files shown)


patroni/ctl.py
```diff
@@ -70,6 +70,10 @@ def load_config(path, dcs):
 
     if not (os.path.exists(path) and os.access(path, os.R_OK)):
         logging.debug('Ignoring configuration file "%s". It does not exists or is not readable.', path)
+        if path == CONFIG_FILE_PATH:
+            ctx = click.get_current_context()
+            logging.warning('Default configuration file "%s" does not exist or is not readable, '\
+                    'try running "%s configure"', path, ctx.command_path)
```
CyberDem0n: Meh, I would actually recommend to use the same config file as for patroni...

yanchenko-igor: Do you mean in the message or by default?

CyberDem0n: It is a long story.
Originally patronictl had its own config file and didn't support the config file used by patroni.
Later it was changed, but the support of the dedicated file wasn't removed.
Maybe it is better to either "deprecate" the 'configure' option or make it create the symlink from the patroni.yaml to `~/.config/patroni/patronictl.yaml`

yanchenko-igor: I believe 'configure' is useful, it the easiest way to make a default config, I rewrote configure to create a symlink to patroni config

CyberDem0n: Actually... Maybe we should rethink the workflow.
There is an very old issue opened by your collegue, https://github.com/zalando/patroni/issues/391
Have you seen it?

yanchenko-igor: I'm not sure I understand, if we already have a running cluster and we want to manage it with patronictl, what would be the workflow?

markwort: If I wanted to run patronictl to monitor several clusters, without having a copy of a patroni config file of any member of each of the clusters, then I'd create a patronictl config file like this:

/very/short/config,yml
```yaml
namespace: /service/nonstandard/
etcd:
  hosts: 10.77.0.111:2379,10.77.0.112:2379,10.77.0.113:2379
  # below user in etcd has a role that allows him only to read from everything in /service/nonstandard
  username: patronictl_monitoring
  password: snakeoil
```
The file could also just contain the details on how to connect to any other DCS.
That (and possibly namespace and scope) should be the only configuration items of the patroni YAML file that are relevant to patronictl anyway.

Then I'd call `patronictl -c /very/short/config,yml list $clustername` for each clustername that I know.


A nice feature, further down the line, would be to have patronictl detect clusters in the given namespace on its own and print the member list for all of them...

markwort: So I'd change the suggestion message to something like:
>please provide the patronictl configuration file using the --config-file or -c flags . You can simply use the patroni config file that you used to start one of the patroni cluster's members, or write a custom patronictl config file to be able to monitor and manipulate several patroni clusters with the same patronictl config file.

PR #1295, opened 2019-11-21
Move multiprocessing.set_start_method() back to main

CyberDem0n:
It is not possible to call it from forked process


sdudoladov:
:+1:

CyberDem0n:
:+1:

PR #1296, opened 2019-11-21
Get rid from requests module

CyberDem0n:
It wasn't used for anything critical anyway, so it doesn't make a lot of sense to keep it as an explicit dependency.

### DIFF
--- a/patroni/postgresql/config.py
+++ b/patroni/postgresql/config.py
@@ -6,7 +6,7 @@ import socket
 import stat
 import time
 
-from requests.structures import CaseInsensitiveDict
+from collections import MutableMapping, OrderedDict
 from six.moves.urllib_parse import urlparse, parse_qsl, unquote
 
 from ..dcs import slot_name_from_member_name, RemoteMember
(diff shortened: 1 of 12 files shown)


patroni/postgresql/config.py
```diff
@@ -250,6 +250,31 @@ def write_param(self, param, value):
...
+        self._store[key.lower()] = (key, value)
+
+    def __getitem__(self, key):
+        return self._store[key.lower()][1]
+
+    def __delitem__(self, key):
+        del self._store[key.lower()]
+
+    def __iter__(self):
+        return (casedkey for casedkey, mappedvalue in self._store.values())
```
sdudoladov: `cased_key` / `mapped_value` (i.e snake_case)

sdudoladov:
:+1:

patroni/postgresql/config.py
CyberDem0n: This is copy&paste from the `requests` module.

CyberDem0n:
The first reason - `requests` is relatively heavy, it takes 11MB in memory on top of urllib3.
The second reason - yes, it causes some issues when the new `urllib3` is released.

Unfortunately `requests` is also dependancy for the official `kubernetes` module and for `python-consul`. With `kubernetes` is solvable. The K8s API we use is very stable, calling REST API endpoints is not a big deal, so I just implemented the part we need and we will get rid of another dependency.

CyberDem0n:
:+1:

PR #1279, opened 2019-11-14
Use "spawn" multiprocessing start method to workaround https://bugs.python.org/issue6721

maciejkowalczyk9ld:
Fixes #1278

PR #1299, opened 2019-11-26
Avoid spawning semaphore tracker process

CyberDem0n:
We are not using semaphores, therefore we don't need to track them.

### DIFF
--- a/patroni/__init__.py
+++ b/patroni/__init__.py
@@ -198,11 +198,6 @@ def check_psycopg2():
 
 
 def main():
-    import multiprocessing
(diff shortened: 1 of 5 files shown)


CyberDem0n:
Hi @maciejkowalczyk9ld,

the unfortunate side-effect of #1279 is a `multiprocessing.semaphore_tracker` tracker subprocess, which is spawned on the first attempt to use `multiprocessing.Process`.
I did a bit hacky attempt to get rid of it, because we are not using semaphores. Could you please give it a test?

FxKu:
:+1:

CyberDem0n:
:+1:

PR #1300, opened 2019-11-26
Make the error more helpful if restapi cannot bind

yanchenko-igor:
Giving user a hint if we couldn't start restapi service.

### DIFF
--- a/patroni/api.py
+++ b/patroni/api.py
@@ -10,7 +10,7 @@ import os
 import six
 import socket
 
-from patroni.exceptions import PostgresConnectionException, PostgresException, PatroniException
+from patroni.exceptions import PostgresConnectionException, PostgresException
 from patroni.postgresql.misc import postgres_version_to_int
 from patroni.utils import deep_compare, parse_bool, patch_config, Retry, \
     RetryFailedError, parse_int, split_host_port, tzutc, uri, cluster_as_json
--- a/patroni/api.py
+++ b/patroni/api.py
@@ -10,8 +10,8 @@ import os
 import six
 import socket
 
-from patroni.postgresql import PostgresConnectionException
-from patroni.postgresql.misc import postgres_version_to_int, PostgresException
+from patroni.exceptions import PostgresConnectionException, PostgresException, PatroniException
+from patroni.postgresql.misc import postgres_version_to_int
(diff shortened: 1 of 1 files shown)


patroni/api.py
```diff
@@ -556,7 +556,12 @@ def __httpserver_init(self, host, port):
         info.sort(key=lambda x: x[0] == socket.AF_INET, reverse=not dual_stack)
 
         self.address_family = info[0][0]
-        HTTPServer.__init__(self, info[0][-1][:2], RestApiHandler)
+        try:
+            HTTPServer.__init__(self, info[0][-1][:2], RestApiHandler)
+        except OSError:
+            raise PatroniException(
```
CyberDem0n: 1. Now we are hiding the real error, `Address already in use` or something like this. It makes it much harder to figure out what it wrong.
2. It would be nice to add the test

patroni/api.py
```diff
@@ -556,7 +556,13 @@ def __httpserver_init(self, host, port):
         info.sort(key=lambda x: x[0] == socket.AF_INET, reverse=not dual_stack)
 
         self.address_family = info[0][0]
-        HTTPServer.__init__(self, info[0][-1][:2], RestApiHandler)
+        try:
+            HTTPServer.__init__(self, info[0][-1][:2], RestApiHandler)
+        except OSError:
+            logger.error(
+                    "Couldn't start a service on '{}:{}', please check your `restapi.listen` configuration"
```
CyberDem0n: `%s` vs .`format` in logging

FxKu:
:+1:

patroni/api.py
```diff
@@ -556,7 +556,12 @@ def __httpserver_init(self, host, port):
         info.sort(key=lambda x: x[0] == socket.AF_INET, reverse=not dual_stack)
 
         self.address_family = info[0][0]
-        HTTPServer.__init__(self, info[0][-1][:2], RestApiHandler)
+        try:
+            HTTPServer.__init__(self, info[0][-1][:2], RestApiHandler)
+        except OSError:
```
CyberDem0n: I think it should be the `socket.error`, otherwise python2 will miss it.

yanchenko-igor: Yes, you are right, fixed

FxKu:
:+1:

CyberDem0n:
:+1:

PR #1301, opened 2019-11-26
Use passfile in primary_conninfo only on 10+

CyberDem0n:
somehow passfile happened to work on ubuntu with older postgres versions, but it is not always the case for other distros.

### DIFF
--- a/patroni/postgresql/config.py
+++ b/patroni/postgresql/config.py
@@ -490,8 +490,8 @@ class ConfigHandler(object):
 
     def format_dsn(self, params, include_dbname=False):
         # A list of keywords that can be found in a conninfo string. Follows what is acceptable by libpq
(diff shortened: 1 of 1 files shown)


sdudoladov:
:+1:

CyberDem0n:
:+1:

johannes-celix:
👍
@CyberDem0n would you be so kind an trigger a release so I can update patroni via pip?
I would need that fix as soon as possible for a Production Cluster.

Thanks,
Johannes

CyberDem0n:
@johannes-celix,
sorry, not earlier than next week, there are a few more issues to be fixed.

johannes-celix:
@CyberDem0n
Thanks, i would appreciate it!

PR #1302, opened 2019-11-26
Inherit CaseInsensitiveDict from urllib3 HTTPHeaderDict

CyberDem0n:
It might look like a hack, but the API is stable enough and didn't change in the past 3+ years.


FxKu:
:+1:

CyberDem0n:
:+1:

PR #1304, opened 2019-11-26
Make sure postgresql.pgpass is a file or it does not exist

yanchenko-igor:


yanchenko-igor:
merged into #1291

yanchenko-igor:
In #1291 we decided to do it separately, Reopening.

PR #1305, opened 2019-11-26
Don't make user to choose from an empty list

yanchenko-igor:
If user provides a wrong cluster name, it returns an empty list of members, then user gets to choose a member from an empty list.

### DIFF
--- a/patroni/ctl.py
+++ b/patroni/ctl.py
@@ -247,10 +247,12 @@ def get_cursor(cluster, connect_parameters, role='master', member=None):
     return None
 
 
-def get_members(cluster, cluster_name, member_names, role, force, action, scheduled_at=None):
+def get_members(cluster, cluster_name, member_names, role, force, action):
     candidates = {m.name: m for m in cluster.members}
 
     if not force or role:
+        if not member_names and not candidates:
+            raise PatroniCtlException('{0} cluster doesn\'t have any members'.format(cluster_name))
         output_members(cluster, cluster_name)
 
     if role:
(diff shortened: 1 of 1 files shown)


patroni/ctl.py
```diff
@@ -263,6 +263,8 @@ def get_members(cluster, cluster_name, member_names, role, force, action, schedu
             member_names = role_names
 
     if not member_names and not force:
+        if not candidates:
```
CyberDem0n: At this moment it is already too late:
```
./patronictl.py -c postgres1.yml restart bla
When should the restart take place (e.g. 2019-11-27T17:07)  [now]: 
+---------+--------+------+------+-------+----+-----------+
| Cluster | Member | Host | Role | State | TL | Lag in MB |
+---------+--------+------+------+-------+----+-----------+
+---------+--------+------+------+-------+----+-----------+
Error: bla cluster doesn't have any members
```

yanchenko-igor: At least user doesn't need to choose member

patroni/ctl.py
```diff
@@ -247,10 +247,12 @@ def get_cursor(cluster, connect_parameters, role='master', member=None):
     return None
 
 
-def get_members(cluster, cluster_name, member_names, role, force, action, scheduled_at=None):
+def get_members(cluster, cluster_name, member_names, role, force, action):
```
CyberDem0n: Let's add one more argument here, `ask_confirmation=True`.
It allows us to make conditional call of the `confirm_members_action()` directly from `get_members()` in most of the cases (only `restart()` will call it explicitly).

yanchenko-igor: done

patroni/ctl.py
```diff
@@ -270,21 +272,23 @@ def get_members(cluster, cluster_name, member_names, role, force, action, schedu
         if member_name not in candidates:
             raise PatroniCtlException('{0} is not a member of cluster'.format(member_name))
 
+    return [candidates[n] for n in member_names]
+
+
+def confirm_scheduled(members, force, action, scheduled_at=None):
```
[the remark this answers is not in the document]
yanchenko-igor: fixed

FxKu:
:+1:

CyberDem0n:
:+1:

Issue #1265, opened 2019-11-04
Suppress full stack traces durring errors and show them in DEBUG level.


yanchenko-igor:
Hello, I would like to suppress the full stack traces when errors happens, and keep just error messages whenever error happens, basically I see a solution to change 
```logger.exception(name)```
 to 
```
logger.error(name)
logger.debug(name, exc_info=True)
```
But I would like to avoid changing the code which cannot be accepted by upstream. Do you see any solution which could be taken back to upstream?

PR #1311, opened 2019-11-29
monkey patch logging to disable tracebacks if log.traceback_level == debug

yanchenko-igor:
#1265 I came to solution how to disable tracebacks by managing if from config. What do you think about it?

### DIFF
--- a/docs/SETTINGS.rst
+++ b/docs/SETTINGS.rst
@@ -46,6 +46,7 @@ Global/Universal
 Log
 ---
 -  **level**: sets the general logging level. Default value is **INFO** (see `the docs for Python logging <https://docs.python.org/3.6/library/logging.html#levels>`_)
+-  **traceback_level**: sets the level where tracebacks will be visible. Default value is **ERROR**. Set it to **DEBUG** if you want to see tracebacks only if you enable **log.level=DEBUG**.
 -  **format**: sets the log formatting string. Default value is **%(asctime)s %(levelname)s: %(message)s** (see `the LogRecord attributes <https://docs.python.org/3.6/library/logging.html#logrecord-attributes>`_)
 -  **dateformat**: sets the datetime formatting string. (see the `formatTime() documentation <https://docs.python.org/3.6/library/logging.html#logging.Formatter.formatTime>`_)
 -  **max\_queue\_size**: Patroni is using two-step logging. Log records are written into the in-memory queue and there is a separate thread which pulls them from the queue and writes to stderr or file. The maximum size of the internal queue is limited by default by **1000** records, which is enough to keep logs for the past 1h20m.
--- a/docs/SETTINGS.rst
+++ b/docs/SETTINGS.rst
@@ -46,7 +46,7 @@ Global/Universal
 Log
 ---
 -  **level**: sets the general logging level. Default value is **INFO** (see `the docs for Python logging <https://docs.python.org/3.6/library/logging.html#levels>`_)
--  **traceback_level**: sets the level where tracebacks will be visible. Default value is **ERROR**. Set it to **DEBUG** if you want to see tracebacks only if you enable **log.level=DEBUG**.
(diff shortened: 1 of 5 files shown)


patroni/log.py
```diff
@@ -61,6 +67,7 @@ def emit(self, record):
 class PatroniLogger(Thread):
 
     DEFAULT_LEVEL = 'INFO'
+    TRACEBACK_LEVEL = 'ERROR'
```
CyberDem0n: `DEFAULT_TRACEBACK_LEVEL`

patroni/log.py
```diff
@@ -11,6 +11,12 @@
 _LOGGER = logging.getLogger(__name__)
 
 
+def debug_exception(logger_obj, msg, *args, **kwargs):
+    kwargs.pop("exc_info", False)
+    logger_obj.error(msg, *args, exc_info=False, **kwargs)
+    logger_obj.debug(msg, *args, exc_info=True, **kwargs)
```
CyberDem0n: Basically that means we will see tracebacks only if log level is set to debug, which is too verbose.

patroni/log.py
```diff
@@ -99,6 +106,8 @@ def reload_config(self, config):
                 self._queue_handler.queue.maxsize = config.get('max_queue_size', self.DEFAULT_MAX_QUEUE_SIZE)
 
             self._root_logger.setLevel(config.get('level', PatroniLogger.DEFAULT_LEVEL))
+            if config.get('traceback_level', PatroniLogger.TRACEBACK_LEVEL).lower() == 'debug':
```
CyberDem0n: The `traceback_level` could be changed in the runtime.

patroni/log.py
yanchenko-igor: Only if we set log.traceback_level == debug, if we don't change log.traceback_level nothing changes.

patroni/log.py
yanchenko-igor: fixed

patroni/log.py
yanchenko-igor: made it switch back to original behavior if configuration is changed back.

patroni/log.py
```diff
@@ -99,6 +111,10 @@ def reload_config(self, config):
                 self._queue_handler.queue.maxsize = config.get('max_queue_size', self.DEFAULT_MAX_QUEUE_SIZE)
 
             self._root_logger.setLevel(config.get('level', PatroniLogger.DEFAULT_LEVEL))
+            if config.get('traceback_level', PatroniLogger.DEFAULT_TRACEBACK_LEVEL).lower() == 'debug':
+                logging.Logger.exception = debug_exception
+            elif config.get('traceback_level', PatroniLogger.DEFAULT_TRACEBACK_LEVEL).lower() == 'error':
```
CyberDem0n: I think here it should be just `else`.

CyberDem0n:
It would be also good to:
1. add [support](https://github.com/zalando/patroni/blob/master/patroni/config.py#L245-L246) of `PATRONI_LOG_TRACEBACK_LEVEL`
2. The documentation
3. Tests

docs/SETTINGS.rst
```diff
@@ -46,6 +46,7 @@ Global/Universal
 Log
 ---
 -  **level**: sets the general logging level. Default value is **INFO** (see `the docs for Python logging <https://docs.python.org/3.6/library/logging.html#levels>`_)
+-  **traceback_level**: sets the level where tracebacks will be visible. Default value is **ERROR**. Set it to **DEBUG** if you want to see tracebacks only if you enable **log.level=DEBUG**.
```
CyberDem0n: We need the same here: https://github.com/zalando/patroni/blob/master/docs/ENVIRONMENT.rst#log

FxKu:
:+1:

CyberDem0n:
:+1:

PR #1314, opened 2019-12-03
Config validator implemented

yanchenko-igor:
Config validating mode added.

### DIFF
--- a/patroni/validator.py
+++ b/patroni/validator.py
@@ -70,7 +70,7 @@ def is_ipv4_address(ip):
     try:
         socket.inet_aton(ip)
     except Exception:
-        raise ConfigParseError("Is not valid ipv4 address")
+        raise ConfigParseError("Is not a valid ipv4 address")
     return True
 
 
@@ -78,7 +78,7 @@ def is_ipv6_address(ip):
     try:
         socket.inet_pton(socket.AF_INET6, ip)
     except Exception:
-        raise ConfigParseError("Is not valid ipv6 address")
+        raise ConfigParseError("Is not a valid ipv6 address")
     return True
 
 
--- a/patroni/validator.py
+++ b/patroni/validator.py
@@ -136,7 +136,7 @@ class Schema(object):
 
     def __call__(self, data):
         for i in self.validate(data):
-            if not i:
+            if not i.status:
                 print(i)
 
     def validate(self, data):
--- a/patroni/validator.py
+++ b/patroni/validator.py
@@ -9,6 +9,7 @@ from patroni.utils import split_host_port, data_directory_is_empty
 from patroni.ctl import find_executable
 from patroni.dcs import dcs_modules
 from patroni.exceptions import ConfigParseError
+from six import string_types
 
 logger = logging.getLogger(__name__)
 
@@ -20,7 +21,7 @@ def data_directory_empty(data_dir):
 
 
 def validate_connect_address(address):
-    if not isinstance(address, str):
+    if not isinstance(address, string_types):
         raise ConfigParseError("is not a string")
     else:
         for s in ["127.0.0.1", "0.0.0.0", "*", "::1"]:
@@ -49,7 +50,7 @@ def validate_host_port(host_port, listen=False, connect=False):
 
 
 def comma_separated_host_port(string):
-    if not isinstance(string, str):
+    if not isinstance(string, string_types):
         raise ConfigParseError("is not a string")
     assert all([validate_host_port(s.strip()) for s in string.split(",")]), "didn't pass the validation"
     return True
@@ -153,8 +154,8 @@ class Schema(object):
 
     def validate(self, data):
         self.data = data
-        if isinstance(self.validator, str):
-            yield Result(isinstance(self.data, str), "is not a string", data=self.data)
+        if isinstance(self.validator, string_types):
+            yield Result(isinstance(self.data, string_types), "is not a string", data=self.data)
         elif issubclass(type(self.validator), type):
             yield Result(isinstance(self.data, self.validator),
                          "is not {}".format(_get_type_name(self.validator)), data=self.data)
@@ -318,7 +319,7 @@ schema = Schema({
     "data_dir": validate_data_dir,
     "bin_dir": validate_bin_dir,
     "parameters": {
-      Optional("unix_socket_directories"): lambda s: assert_(all([isinstance(s, str), len(s)]))
+      Optional("unix_socket_directories"): lambda s: assert_(all([isinstance(s, string_types), len(s)]))
     },
     Optional("pg_hba"): [str],
     Optional("pg_ident"): [str],
--- a/patroni/validator.py
+++ b/patroni/validator.py
@@ -141,7 +141,7 @@ class Schema(object):
                 yield i
         elif issubclass(type(self.validator), type):
             yield Result(isinstance(self.data, self.validator),
-                         "is not a(n) {}".format(_get_type_name(self.validator)), data=self.data)
+                         "is not {}".format(_get_type_name(self.validator)), data=self.data)
         elif callable(self.validator):
             try:
                 self.validator(data)
@@ -159,8 +159,11 @@ class Schema(object):
 
     def iter(self):
         if isinstance(self.validator, dict):
-            for i in self.iter_dict():
-                yield i
+            if not isinstance(self.data, dict):
+                yield Result(False, "is not a dictionary.")
+            else:
+                for i in self.iter_dict():
+                    yield i
         elif isinstance(self.validator, list):
             if len(self.validator) == 0:
                 yield Result(isinstance(self.data, list), "is not a list", data=self.data)
@@ -190,25 +193,18 @@ class Schema(object):
 
     def iter_dict(self):
         for key in self.validator.keys():
-            if isinstance(self.data, dict):
-                orcase = False
-                if isinstance(key, Or) and isinstance(self.validator[key], Case):
-                    orcase = True
-                for d in self._data_key(key):
-                    if d not in self.data and not isinstance(key, Optional):
-                        yield Result(False, "is not defined.", path=d)
-                    elif d not in self.data and isinstance(key, Optional):
-                        continue
-                    else:
-                        validator = self.validator[key]
-                        if orcase:
-                            validator = self.validator[key]._schema[d]
-                        for v in Schema(validator).validate(self.data[d]):
-                            yield Result(v.status, v.error,
-                                         path=(d + ("." + v.path if v.path else "")), data=v.data)
-            else:
-                yield Result(False, "is not a dictionary.")
-                return
+            for d in self._data_key(key):
+                if d not in self.data and not isinstance(key, Optional):
+                    yield Result(False, "is not defined.", path=d)
+                elif d not in self.data and isinstance(key, Optional):
+                    continue
+                else:
+                    validator = self.validator[key]
+                    if isinstance(key, Or) and isinstance(self.validator[key], Case):
+                        validator = self.validator[key]._schema[d]
+                    for v in Schema(validator).validate(self.data[d]):
+                        yield Result(v.status, v.error,
+                                     path=(d + ("." + v.path if v.path else "")), data=v.data)
 
     def _data_key(self, key):
         if isinstance(self.data, dict) and isinstance(key, str):
@@ -225,23 +221,23 @@ class Schema(object):
 
 def _get_type_name(python_type):
     if python_type == str:
-        return "string"
+        return "a string"
     elif python_type == int:
-        return "integer"
+        return "an integer"
     elif python_type == float:
-        return "number"
+        return "a number"
     elif python_type == bool:
-        return "boolean"
+        return "a boolean"
     elif python_type == list:
-        return "array"
+        return "an array"
     elif python_type == dict:
-        return "dictionary"
+        return "a dictionary"
     return python_type.__name__
 
 
 def assert_(condition, message="Wrong value"):
     if message:
-        assert condition,message
+        assert condition, message
     else:
         assert condition
 
@@ -294,7 +290,7 @@ schema = Schema({
           Optional("scope_label"): str,
           Optional("role_label"): str,
           Optional("use_endpoints"): bool,
-          Optional("ports"): [{"name":str,"port":int}],
+          Optional("ports"): [{"name": str, "port": int}],
           },
       }),
   "postgresql": {
@@ -308,7 +304,7 @@ schema = Schema({
     "data_dir": validate_data_dir,
     "bin_dir": validate_bin_dir,
     "parameters": {
-      Optional("unix_socket_directories"): lambda s: asserct_(all([isinstance(s, str), len(s)]))
+      Optional("unix_socket_directories"): lambda s: assert_(all([isinstance(s, str), len(s)]))
     },
     Optional("pg_hba"): [str],
     Optional("pg_ident"): [str],
--- a/patroni/validator.py
+++ b/patroni/validator.py
@@ -254,19 +254,8 @@ class Schema(object):
 
 
 def _get_type_name(python_type):
-    if python_type == str:
-        return "a string"
-    elif python_type == int:
-        return "an integer"
-    elif python_type == float:
-        return "a number"
-    elif python_type == bool:
-        return "a boolean"
-    elif python_type == list:
-        return "an array"
-    elif python_type == dict:
-        return "a dictionary"
-    return python_type.__name__
+    return {str: 'a string', int: 'and integer', float: 'a number', bool: 'a boolean',                                          
+            list: 'an array', dict: 'a dictionary'}.get(python_type, python_type.__name__)
 
 
 def assert_(condition, message="Wrong value"):
--- a/patroni/validator.py
+++ b/patroni/validator.py
@@ -47,8 +47,7 @@ def validate_host_port(host_port, listen=False, connect=False):
 def comma_separated_host_port(string):
     if not isinstance(string, str):
         raise ConfigParseError("is not a string")
-    if not all([validate_host_port(s.strip()) for s in string.split(",")]):
-        raise ConfigParseError("didn't pass the validation")
+    assert all([validate_host_port(s.strip()) for s in string.split(",")]), "didn't pass the validation"
     return True
 
 
--- a/patroni/validator.py
+++ b/patroni/validator.py
@@ -165,7 +165,9 @@ class Schema(object):
             if len(self.validator) == 0:
                 yield Result(isinstance(self.data, list), "is not a list", data=self.data)
             elif len(self.validator) == 1:
-                if isinstance(self.data, list):
+                if not isinstance(self.data, list):
+                    yield Result(False, "is not a list", data=self.data)
+                else:
                     for key, value in enumerate(self.data):
                         for v in Schema(self.validator[0]).validate(value):
                             yield Result(v.status, v.error,
@@ -237,6 +239,13 @@ def _get_type_name(python_type):
     return python_type.__name__
 
 
+def assert_(condition, message="Wrong value"):
+    if message:
+        assert condition,message
+    else:
+        assert condition
+
+
 userattributes = {"username": "", Optional("password"): ""}
 available_dcs = [m.split(".")[-1] for m in dcs_modules()]
 
@@ -273,7 +282,7 @@ schema = Schema({
          },
       "exhibitor": {
           "hosts": str,
-          "port": lambda i: int(i) <= 65535,
+          "port": lambda i: assert_(int(i) <= 65535),
           Optional("pool_interval"): int
           },
       "zookeeper": {
@@ -285,6 +294,7 @@ schema = Schema({
           Optional("scope_label"): str,
           Optional("role_label"): str,
           Optional("use_endpoints"): bool,
+          Optional("ports"): [{"name":str,"port":int}],
           },
       }),
   "postgresql": {
@@ -298,7 +308,21 @@ schema = Schema({
     "data_dir": validate_data_dir,
     "bin_dir": validate_bin_dir,
     "parameters": {
-      Optional("unix_socket_directories"): lambda s: all([isinstance(s, str), len(s)])
-    }
+      Optional("unix_socket_directories"): lambda s: asserct_(all([isinstance(s, str), len(s)]))
+    },
+    Optional("pg_hba"): [str],
+    Optional("pg_ident"): [str],
+    Optional("pg_ctl_timeout"): int,
+    Optional("use_pg_rewind"): bool
+  },
+  Optional("watchdog"): {
+    Optional("mode"): lambda m: assert_(m in ["off", "automatic", "required"]),
+    Optional("device"): str
+  },
+  Optional("tags"): {
+    Optional("nofailover"): bool,
+    Optional("clonefrom"): bool,
+    Optional("noloadbalance"): bool,
+    Optional("nosync"): bool
   }
 })
--- a/patroni/validator.py
+++ b/patroni/validator.py
@@ -254,7 +254,7 @@ class Schema(object):
 
 
 def _get_type_name(python_type):
-    return {str: 'a string', int: 'and integer', float: 'a number', bool: 'a boolean',                                          
+    return {str: 'a string', int: 'and integer', float: 'a number', bool: 'a boolean',
             list: 'an array', dict: 'a dictionary'}.get(python_type, python_type.__name__)
 
 
--- a/patroni/validator.py
+++ b/patroni/validator.py
@@ -103,22 +103,25 @@ def config_validator(config):
         if not find_executable(program, bin_dir):
             logger.warning("Program '%s' not found.", program)
     if "bootstrap" in config and "initdb" in config["bootstrap"]:
-        xlogdir = config["bootstrap"]["initdb"].get(
-            "xlogdir", os.path.join(data_dir, "pg_xlog")
-        )
-        waldir = config["bootstrap"]["initdb"].get(
-            "waldir", os.path.join(data_dir, "pg_wal")
-        )
-        log_directory = config["bootstrap"]["initdb"].get(
-            "log_directory", os.path.join(data_dir, "pg_log")
-        )
-        if waldir == log_directory:
+        initdb = {
+            'xlogdir': os.path.join(data_dir, "pg_xlog"),
+            'waldir': os.path.join(data_dir, "pg_wal"),
+            'log_directory': os.path.join(data_dir, "pg_log")}
+        for item in config["bootstrap"]["initdb"]:
+            if isinstance(item, dict):
+                initdb.update(item)
+
+        if initdb["waldir"] == initdb["log_directory"]:
             logger.warning(
-                "waldir(%s) and log_directory (%s) are pointing to the same path", waldir, log_directory
+                "waldir(%s) and log_directory (%s) are pointing to the same path",
+                initdb["waldir"],
+                initdb["log_directory"]
             )
-        if xlogdir == log_directory:
+        if initdb["xlogdir"] == initdb["log_directory"]:
             logger.warning(
-                "xlogdir(%s) and log_directory (%s) are pointing to the same path", xlogdir, log_directory
+                "xlogdir(%s) and log_directory (%s) are pointing to the same path",
+                initdb["xlogdir"],
+                initdb["log_directory"]
             )
 
     if fatal:
--- a/patroni/validator.py
+++ b/patroni/validator.py
@@ -68,9 +68,10 @@ def is_ipv6_address(ip):
     assert socket.inet_pton(socket.AF_INET6, ip)
     return True
 
+
 def get_major_version():
     version = subprocess.check_output(['postgres', '--version']).decode()
-    version = re.match('^[^\s]+ [^\s]+ (\d+)(\.(\d+))?', version)
+    version = re.match(r'^[^\s]+ [^\s]+ (\d+)(\.(\d+))?', version)
     return '.'.join([version.group(1), version.group(3)]) if int(version.group(1)) < 10 else version.group(1)
 
 
@@ -86,7 +87,8 @@ def validate_data_dir(data_dir):
             with open(os.path.join(data_dir, "PG_VERSION"), "r") as version:
                 pgversion = int(version.read())
             if str(pgversion) != get_major_version():
-                raise ConfigParseError("data_dir directory postgresql version doesn't match with 'postgres --version' output")
+                raise ConfigParseError("data_dir directory postgresql version doesn't match"
+                                       "with 'postgres --version' output")
             waldir = ("pg_wal" if pgversion >= 10 else "pg_xlog")
             if not os.path.isdir(os.path.join(data_dir, waldir)):
                 raise ConfigParseError("data dir for the cluster is not empty, but doesn't contain"
@@ -214,7 +216,7 @@ class Schema(object):
                 results += filter(lambda x: not x.status, r)
             else:
                 results += r
-        if not any([r.status for r in results]):
+        if not any([x.status for x in results]):
             for v in results:
                 yield Result(v.status, v.error, path=v.path, data=v.data)
 
--- a/patroni/validator.py
+++ b/patroni/validator.py
@@ -1,5 +1,4 @@
 #!/usr/bin/env python3
-import logging
 import os
 import socket
 import re
@@ -10,8 +9,6 @@ from patroni.ctl import find_executable
 from patroni.dcs import dcs_modules
 from patroni.exceptions import ConfigParseError
 from six import string_types
-
-logger = logging.getLogger(__name__)
 
 
 def data_directory_empty(data_dir):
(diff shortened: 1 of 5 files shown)


patroni/validator.py
```diff
@@ -0,0 +1,114 @@
...
+
+def data_directory_empty(data_dir):
+    if os.path.isfile(os.path.join(data_dir, "global", "pg_control")):
+        return False
+    if not os.path.exists(data_dir):
+        return True
+    return all(
+        os.name != "nt" and (n.startswith(".") or n == "lost+found")
+        for n in os.listdir(data_dir)
+    )
```
CyberDem0n: It pretty much duplicates `Postgresql.data_directory_empty()`

patroni/validator.py
```diff
@@ -0,0 +1,114 @@
...
+
+def which(program):
+    for path in os.environ["PATH"].split(os.pathsep):
+        exe_file = os.path.join(path, program)
+        if os.path.isfile(exe_file) and os.access(exe_file, os.X_OK):
+            return exe_file
+    return None
```
CyberDem0n: It duplicates `patroni.ctl.find_executable`.

patroni/validator.py
```diff
@@ -0,0 +1,114 @@
...
+
+def config_validator(config):
+    if not "name" in config:
+        logger.warning("name is not defined")
```
CyberDem0n: This is actually a fatal.

patroni/validator.py
```diff
@@ -0,0 +1,114 @@
...
+    if not "scope" in config:
+        logger.warning("scope is not defined")
```
CyberDem0n: Fatal

patroni/validator.py
```diff
@@ -0,0 +1,114 @@
...
+    if not "postgresql" in config:
+        logger.warning("postgresql section is missing")
```
CyberDem0n: Fatal

patroni/validator.py
```diff
@@ -0,0 +1,114 @@
...
+        if not "authentication" in config["postgresql"]:
+            logger.warning("postgresql.authentication is nor defined")
+        else:
+            if not "replication" in config["postgresql"]["authentication"]:
+                logger.warning("postgresql.authentication.replication is not defined")
+            if not "superuser" in config["postgresql"]["authentication"]:
+                logger.warning("postgresql.authentication.superuser is not defined")
+
+        if not "bin_dir" in config["postgresql"]:
+            logger.info("postgresql.bin_dir is not defined")
```
CyberDem0n: In case is bin_dir isn't defind we should still search for binaries.

patroni/validator.py
```diff
@@ -0,0 +1,114 @@
...
+        else:
+            if os.path.exists(config["postgresql"]["bin_dir"]):
+                for program in ["pg_ctl", "initdb", "pg_controldata", "pg_basebackup"]:
```
CyberDem0n: The list is missing the most important binary: `postgres` :)

patroni/validator.py
```diff
@@ -0,0 +1,114 @@
...
+                    exe_file = os.path.join(config["postgresql"]["bin_dir"], program)
+                    if not (os.path.isfile(exe_file) and os.access(exe_file, os.X_OK)):
+                        logger.warning("Program '%s' not found.", program)
+            else:
+                logger.warning("Directory '%s' does not exist.", config['postgresql']['bin_dir'])
+                for program in ["pg_ctl", "initdb", "pg_controldata", "pg_basebackup"]:
+                    if not which(program):
+                        logger.warning("Program '%s' not found.", program)
```
CyberDem0n: There is quite a few duplicated code.
It could be simplified as:
```python
path = config['postgresql'].get('bin_dir') or None
for ...
  if not find_executable(program, path):
    logger.fatal...
```

Off course we also have to check that the `bin_dir` is a string if it is defined.

patroni/validator.py
```diff
@@ -0,0 +1,114 @@
...
+        if not data_directory_empty(config["postgresql"]["data_dir"]):
+            if not os.path.exists(
+                os.path.join(config["postgresql"]["data_dir"], "PG_VERSION")
+            ):
+                logger.warning(
+                    "data_dir doesn't look like a valid data directory make sure you provide a valid path in configuration"
+                )
+            elif not os.path.isdir(
+                os.path.join(config["postgresql"]["data_dir"], "pg_wal")
```
CyberDem0n: This is version dependent, pg_xlog or pg_wal for 10+

patroni/validator.py
```diff
@@ -0,0 +1,117 @@
...
+        else:
+            if not data_directory_empty(config["postgresql"]["data_dir"]):
+                if not os.path.exists(
+                    os.path.join(config["postgresql"]["data_dir"], "PG_VERSION")
+                ):
+                    logger.warning(
+                        "data_dir doesn't look like a valid data directory make sure you provide a valid path in configuration"
+                    )
+                elif not os.path.isdir(
+                    os.path.join(config["postgresql"]["data_dir"], "pg_wal")
```
CyberDem0n: It is a version dependent, either pg_xlog, or pg_wal for 10+

patroni/validator.py
```diff
@@ -0,0 +1,117 @@
...
+                ):
+                    logger.warning(
+                        'data dir for the cluster is not empty, but doesn\'t contain "pg_wal" directory'
+                    )
+            data_dir = config["postgresql"]["data_dir"]
+            if "parameters" not in config["postgresql"]:
+                logger.info("postgresql.parameters is not defined")
+            else:
+                xlogdir = config["postgresql"]["parameters"].get(
+                    "xlogdir", os.path.join(data_dir, "pg_xlog")
```
CyberDem0n: There are no such parameters as waldir or xlogdir.

pg_log and pg_wal are always located in the PGDATA, but they could be symlinks.

patroni/validator.py
```diff
@@ -0,0 +1,117 @@
...
+                    )
+                if xlogdir == log_directory:
+                    logger.warning(
+                        "xlogdir(%s) and log_directory (%s) are pointing to the same path", xlogdir, log_directory
+                    )
+
+        if "listen" not in config["postgresql"]:
+            logger.warning("postgresql.listen is not defined")
+        else:
+            host, port = split_host_port(config["postgresql"]["listen"], 5432)
```
CyberDem0n: The `listen` could be in the format: `localhost,10.0.0.1:5433`

CyberDem0n: There is also an optional `connect_address`.
If it is not defined, the value from `listen` will be used.
It is very important to check that the `host` in the `connect_adress` isn't a `localhost`, `127.0.0.1`, `0.0.0.0`, `::1`, `*` or something similar.

patroni/validator.py
yanchenko-igor: yeah)

patroni/validator.py
yanchenko-igor: I'm checking both cases separately

patroni/validator.py
yanchenko-igor: https://www.postgresql.org/docs/10/app-initdb.html
https://www.postgresql.org/docs/9.6/app-initdb.html

CyberDem0n: You are referring to options of `initdb`, which are not postgres configuration parameters.
There is a way to specify them though, but in a different place:
```yaml
bootstrap:
  initdb:
  - data-checksums
  - waldir: /foo
```

It just tells to initdb that we want the waldir to be located at `/foo` and it will create the symlink `$PGDATA/pg_wal -> /foo`

patroni/validator.py
```diff
@@ -0,0 +1,228 @@
...
+                    logger.warning("postgresql.authentication.superuser is not a dictionary")
+                else:
+                    if "username" not in config["postgresql"]["authentication"]["superuser"]:
+                        logger.warning("postgresql.authentication.superuser.username is not defined")
+                    elif not isinstance(config["postgresql"]["authentication"]["superuser"]["username"], str):
+                        logger.warning("postgresql.authentication.superuser.username is not a string")
+                    if "password" not in config["postgresql"]["authentication"]["superuser"]:
+                        logger.warning("postgresql.authentication.superuser.password is not defined")
+                    elif not isinstance(config["postgresql"]["authentication"]["superuser"]["password"], str):
+                        logger.warning("postgresql.authentication.superuser.password is not a string")
```
CyberDem0n: Why not do a couple of loops instead?
There is also auth parameters section for `rewind`.
Lack of password is actually fine, people might use different authentication methods like kerberos or client certificates.

patroni/validator.py
```diff
@@ -0,0 +1,228 @@
...
+    for program in ["pg_ctl", "initdb", "pg_controldata", "pg_basebackup", "postgres"]:
+        if not find_executable(program, bin_dir):
+            logger.warning("Program '%s' not found.", program)
+    if "bootstrap" not in config:
+        logger.warning("bootstrap section is missing")
+    elif not isinstance(config["bootstrap"], dict):
+        logger.warning("bootstrap is not a dictionary")
+    else:
+        if "dcs" not in config["bootstrap"]:
+            logger.warning("bootstrap.dcs is not defined")
```
CyberDem0n: This is absolutely fine, and moreover is the whole bootstrap is used only onces, when we bootstrap the new cluster.

patroni/validator.py
```diff
@@ -0,0 +1,228 @@
...
+        elif not isinstance(config["bootstrap"]["dcs"], dict):
+            logger.warning("bootstrap.dcs is not a dictionary")
+        if "initdb" in config["bootstrap"]:
+            if not isinstance(config["bootstrap"]["initdb"], list):
+                logger.warning("bootstrap.initdb is not a list")
+            initdb = {
+                'xlogdir': os.path.join(data_dir, "pg_xlog"),
+                'waldir': os.path.join(data_dir, "pg_wal"),
+                'log_directory': os.path.join(data_dir, "pg_log")}
```
CyberDem0n: `log_directory` is actually configurable via postgresql.conf and the default name depends on the major version.

patroni/validator.py
```diff
@@ -0,0 +1,228 @@
...
+                    initdb["waldir"],
+                    initdb["log_directory"]
+                )
+            if initdb["xlogdir"] == initdb["log_directory"]:
+                logger.warning(
+                    "xlogdir(%s) and log_directory (%s) are pointing to the same path",
+                    initdb["xlogdir"],
+                    initdb["log_directory"]
+                )
+    if not any([dcs in config for dcs in ["consul", "etcd", "exhibitor", "kubernetes", "zookeeper"]]):
```
CyberDem0n: You can get the list by calling `patroni.dcs.dcs_modules()`, but you'll have to extract the last part of each item: `name.split('.')[-1]`

patroni/validator.py
```diff
@@ -0,0 +1,228 @@
...
+        logger.error("Neither of distributed configuration store is configured, "
+                     "please configure one of these: consul, etcd, exhibitor, kubernetes, zookeeper")
+        fatal = 1
+    elif "etcd" in config:
+        if not isinstance(config["etcd"], dict):
+            logger.warning("etcd is not a dictionary")
+        elif not any([i in config["etcd"] for i in ["srv", "hosts", "host", "url"]]):
```
CyberDem0n: there is also the `proxy`

patroni/validator.py
```diff
@@ -0,0 +1,228 @@
...
+        return False
+    if not os.path.exists(data_dir):
+        return True
+    return all(
+        os.name != "nt" and (n.startswith(".") or n == "lost+found")
+        for n in os.listdir(data_dir)
+    )
+
+
+def config_validator(config):
```
CyberDem0n: The whole function is huge, it would make sense to break it into multiple smaller functions.
It will simplify future maintenance and testing.

patroni/validator.py
```diff
@@ -0,0 +1,230 @@
...
+              "hosts": Or(str, list),
+              "srv": str,
+              "url": str,
+              "proxy": str})
+         },
+      "exhibitor": {
+          "hosts": str,
+          },
+      "zookeeper": {
+          "hosts": [validate_host_port],
```
CyberDem0n: It will try to connect and show the warning?

CyberDem0n: It also could be a comma-separated string.

patroni/validator.py
```diff
@@ -0,0 +1,230 @@
...
+      "etcd": {
+          Or("host", "hosts", "srv", "url", "proxy"): Case({
+              "host": str,
+              "hosts": Or(str, list),
+              "srv": str,
+              "url": str,
+              "proxy": str})
+         },
+      "exhibitor": {
+          "hosts": str,
```
CyberDem0n: `hosts` must be a list.
`port` is also mandatory (int or str)
`poll_interval` - int, optional

patroni/validator.py
```diff
@@ -0,0 +1,230 @@
...
+
+
+userattributes = {"username": "", Optional("password"): ""}
+available_dcs = [m.split(".")[-1] for m in dcs_modules()]
+
+schema = Schema({
+  "name": str,
+  "scope": str,
+  "restapi": {
+    "listen": validate_host_port
```
CyberDem0n: `connect_address` - we should give a huge warning if it is not accessible from outside (i.e. 127.0.0.1, 0.0.0.0.0 and so on)

patroni/validator.py
```diff
@@ -0,0 +1,230 @@
...
+          },
+      "zookeeper": {
+          "hosts": [validate_host_port],
+          },
+      "kubernetes": {
+          Optional("namespace"): str,
+          },
+      }),
+  "postgresql": {
+    "listen": validate_host_port,
```
CyberDem0n: `connect_address` - we should give a huge warning if it is not accessible from outside (i.e. 127.0.0.1, 0.0.0.0.0 and so on)

patroni/validator.py
```diff
@@ -0,0 +1,230 @@
...
+    "authentication": {
+      "replication": userattributes,
```
CyberDem0n: For the `replication` the username is mandatory.

patroni/validator.py
```diff
@@ -0,0 +1,230 @@
...
+      "superuser": userattributes,
+      "rewind":  userattributes
+    },
+    "data_dir": validate_data_dir,
+    "bin_dir": validate_bin_dir,
+    "parameters": {
+      "unix_socket_directories": lambda s: all([str(s), len(s)])
```
CyberDem0n: `unix_socket_directories` is actually optional.

patroni/validator.py
```diff
@@ -0,0 +1,230 @@
...
+              "url": str,
+              "proxy": str})
+         },
+      "exhibitor": {
+          "hosts": str,
+          },
+      "zookeeper": {
+          "hosts": [validate_host_port],
+          },
+      "kubernetes": {
```
CyberDem0n: There are a bit more parameters for kubernetes: https://github.com/zalando/patroni/blob/master/docs/SETTINGS.rst#kubernetes

Most of them are optional, but the structure is not really obvious.

patroni/validator.py
```diff
@@ -0,0 +1,230 @@
...
+                subschema = schema._schema[a]
+                self.validator(a, data[a], subschema)
+        if counter < 1:
+            logger.warning("neither of %s is defined", ", ".join([self.get_fullpath(a) for a in key.args]))
+
+    def get_fullpath(self, path):
+        return self._path + ("." if self._path else "") + str(path)
+
+
+def _get_type_name(python_type):
```
CyberDem0n: Looks like it is unused.

patroni/validator.py
```diff
@@ -0,0 +1,230 @@
...
+    "dcs": {
+        Optional("ttl"): int,
+        Optional("loop_wait"): int,
+        Optional("retry_timeout"): int,
+        Optional("maximum_lag_on_failover"): int
+        },
+    "pg_hba": [str],
+    "initdb": [Or(str, dict)]
+  },
+  Or(*available_dcs): Case({
```
CyberDem0n: there is also `consul`

patroni/validator.py
yanchenko-igor: Sure, as soon as I come up with a solution.

patroni/validator.py
yanchenko-igor: > It will try to connect and show the warning?

yes, I will fix that

patroni/validator.py
yanchenko-igor: https://github.com/zalando/patroni/pull/1314/files#diff-05020fb6b44e1d780d1bb03b5b0e257eR143

patroni/validator.py
yanchenko-igor: should make username for the rest of them optional?

patroni/validator.py
yanchenko-igor: https://github.com/zalando/patroni/blob/master/docs/SETTINGS.rst#exhibitor says: 
> initial list of Exhibitor (ZooKeeper) nodes in format: 'host1,host2,etc...'. 
As I understood it's a list in string format.

patroni/validator.py
yanchenko-igor: Rewritten into Schema class and separate validators

patroni/validator.py
yanchenko-igor: added

patroni/validator.py
yanchenko-igor: fixed

patroni/validator.py
```diff
@@ -0,0 +1,325 @@
...
+
+
+def validate_bin_dir(bin_dir):
+    if not bin_dir:
+        raise ConfigParseError("is an empty string")
+    elif not os.path.exists(bin_dir):
+        raise ConfigParseError("Directory '{}' does not exist.".format(bin_dir))
+    elif not os.path.isdir(bin_dir):
+        raise ConfigParseError("'{}' is not a directory.".format(bin_dir))
+    for program in ["pg_ctl", "initdb", "pg_controldata", "pg_basebackup", "postgres"]:
```
CyberDem0n: It would be also great to call the `postgres --version`, parse the output and make sure that it matches with `PG_VERSION`.

The above code parses major version from the output and is able to correctly handle case when the version is in format: `13devel`.
```python
def get_major_version():
    version = subprocess.check_output(['postgres', '--version']).decode()
    version = re.match('^[^\s]+ [^\s]+ (\d+)(\.(\d+))?', version)
    return '.'.join([version.group(1), version.group(3)]) if int(version.group(1)) < 10 else version.group(1)
```

tests/test_validator.py
```diff
@@ -0,0 +1,215 @@
...
+import os
+import socket
+import copy
+from mock import Mock, patch, mock_open
+from patroni.validator import schema
+import sys
+if not sys.version_info.major == 3:
+    from StringIO import StringIO
+else:
+    from io import StringIO
```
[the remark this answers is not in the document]
yanchenko-igor: It expects strings instead of unicode objects.

patroni/validator.py
```diff
@@ -0,0 +1,325 @@
...
+    return True
+
+
+def validate_host_port(host_port, listen=False, connect=False):
+    try:
+        host, port = split_host_port(host_port, None)
+    except (ValueError, TypeError):
+        raise ConfigParseError("contains a wrong value")
+    else:
+        with socket.socket(socket.AF_INET, socket.SOCK_STREAM) as s:
```
CyberDem0n: 1. `socket.socket()` doesn't support `with` statement in python 2.7
2. do we really need to execute this code if neither listen or connect are set?

patroni/validator.py
```diff
@@ -0,0 +1,325 @@
...
+                for i in key.args:
+                    if i in self.data:
+                        yield i
+            else:
+                for i in key.args:
+                    yield i
+
+
+def _get_type_name(python_type):
+    if python_type == str:
```
CyberDem0n: In the python 2.7 highly likely most of the strings actually would be `unicode`, so we have to use `six.string_types` for all checks.

tests/test_validator.py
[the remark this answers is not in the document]
CyberDem0n: Ok, but we could still use `from six import StringIO` :)

patroni/validator.py
yanchenko-igor: new commit with fix is on the way :)

yanchenko-igor: 2. it finds host resolve issues actually.

patroni/validator.py
yanchenko-igor: it's used against strings in our schema, "" is a sting in python 2.7, so it is not an issue.

CyberDem0n: I was mostly meaning a lot of checks `isinstance(xxx, str)`.

Loading yaml in python2 seems to be not a problem, it returns string, but json return unicode:

```python
$ python2
Python 2.7.15+ (default, Oct  7 2019, 17:39:04) 
[GCC 7.4.0] on linux2
>>> import yaml
>>> with open('postgres0.yml') as fd:
...     yaml.safe_load(fd)
... 
{'postgresql': {'authentication': {'replication': {'username': 'replicator', 'password': 'rep-pass'}, 'superuser': {'username': 'postgres', 'password': 'zalando'}, 'rewind': {'username': 'rewind_user', 'password': 'rewind_password'}}, 'data_dir': 'data/postgresql0', 'parameters': {'unix_socket_directories': '.'}, 'connect_address': '127.0.0.1:5432', 'pgpass': '/tmp/pgpass0', 'listen': '127.0.0.1:5432'}, 'name': 'postgresql0', 'etcd': {'host': '127.0.0.1:2379'}, 'tags': {'nofailover': False, 'noloadbalance': False, 'clonefrom': False, 'nosync': False}, 'restapi': {'connect_address': '127.0.0.1:8008', 'listen': '127.0.0.1:8008'}, 'bootstrap': {'pg_hba': ['host replication replicator 127.0.0.1/32 md5', 'host all all 0.0.0.0/0 md5'], 'users': {'admin': {'password': 'admin', 'options': ['createrole', 'createdb']}}, 'initdb': [{'encoding': 'UTF8'}, 'data-checksums'], 'dcs': {'ttl': 30, 'maximum_lag_on_failover': 1048576, 'retry_timeout': 10, 'postgresql': {'use_pg_rewind': True, 'parameters': None}, 'loop_wait': 10}}, 'scope': 'batman'}
>>> import json
>>> json.loads('{"foo": "bar"}')
{u'foo': u'bar'}
```

Since we also need to validate the config coming from DCS it could potentially become a problem.

patroni/validator.py
```diff
@@ -0,0 +1,338 @@
...
+        raise ConfigParseError("is an empty string")
+    elif os.path.exists(data_dir) and not os.path.isdir(data_dir):
+        raise ConfigParseError("is not a directory")
+    elif not data_directory_empty(data_dir):
+        if not os.path.exists(os.path.join(data_dir, "PG_VERSION")):
+            raise ConfigParseError("doesn't look like a valid data directory")
+        else:
+            with open(os.path.join(data_dir, "PG_VERSION"), "r") as version:
+                pgversion = int(version.read())
+            if str(pgversion) != get_major_version():
```
CyberDem0n: It could be that the postgres isn't available in the default PATH, we have to take postgresql.bin_dir into account.

patroni/validator.py
yanchenko-igor: understood, fixed

patroni/validator.py
```diff
@@ -0,0 +1,339 @@
...
+      "replication": userattributes,
+      "superuser": userattributes,
+      "rewind":  userattributes
+    },
+    "data_dir": validate_data_dir,
+    "bin_dir": validate_bin_dir,
+    "parameters": {
+      Optional("unix_socket_directories"): lambda s: assert_(all([isinstance(s, string_types), len(s)]))
+    },
+    Optional("pg_hba"): [str],
```
CyberDem0n: Look, I slightly changed the test:
```diff
--- a/tests/test_validator.py
+++ b/tests/test_validator.py
@@ -59,7 +59,7 @@ config = {
         "parameters": {
             "unix_socket_directories": "."
         },
-        "pg_hba": ["string"],
+        "pg_hba": [u'string'],
         "pg_ident": ["string"],
         "pg_ctl_timeout": 1000,
         "use_pg_rewind": False
```

and such change breaks validator. As I already mentioned we can get the unicode string instead of str when parsing JSON object from DCS.
The following change fixes the issue just for pg_hba case:
```diff
--- a/patroni/validator.py
+++ b/patroni/validator.py
@@ -321,7 +321,7 @@ schema = Schema({
     "parameters": {
       Optional("unix_socket_directories"): lambda s: assert_(all([isinstance(s, string_types), len(s)]))
     },
-    Optional("pg_hba"): [str],
+    Optional("pg_hba"): [string_types],
     Optional("pg_ident"): [str],
     Optional("pg_ctl_timeout"): int,
     Optional("use_pg_rewind"): bool
```

We should do the same for all remaining cases.

patroni/validator.py
```diff
@@ -0,0 +1,358 @@
...
+    Optional("pg_ident"): [str],
+    Optional("pg_ctl_timeout"): int,
+    Optional("use_pg_rewind"): bool
+  },
+  Optional("watchdog"): {
+    Optional("mode"): lambda m: assert_(m in ["off", "automatic", "required"]),
+    Optional("device"): str
+  },
+  Optional("tags"): {
```
CyberDem0n: There is also the [replicatefrom](https://patroni.readthedocs.io/en/latest/SETTINGS.html#settings) tag,

patroni/validator.py
```diff
@@ -317,7 +336,8 @@ def assert_(condition, message="Wrong value"):
       "rewind":  userattributes
     },
     "data_dir": validate_data_dir,
-    "bin_dir": validate_bin_dir,
+    "bin_dir": Directory(contains=["pg_ctl"],
```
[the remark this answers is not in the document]
yanchenko-igor: no good reason actually, just to have better coverage with tests

patroni/validator.py
```diff
@@ -0,0 +1,358 @@
...
+        if not self.status:
+            self.error = error
+        else:
+            self.error = None
+
+    def __repr__(self):
+        return self.path + (" " + str(self.data) + " " + self._error if self.error else "")
+
+    def __bool__(self):
+        return self.status
```
CyberDem0n: Seems to be unused, because the [status](https://github.com/zalando/patroni/pull/1314/files#diff-05020fb6b44e1d780d1bb03b5b0e257eR164) is checked explicitly.

patroni/validator.py
```diff
@@ -0,0 +1,358 @@
...
+    try:
+        host, port = split_host_port(host_port, None)
+    except (ValueError, TypeError):
+        raise ConfigParseError("contains a wrong value")
+    else:
+        s = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
+        try:
+            if s.connect_ex((host, port)) == 0 and listen:
+                ConfigParseError("Port {} is already in use.".format(port))
+            elif connect:
```
CyberDem0n: Can we safely assume that connect = True if listen = False?

patroni/validator.py
```diff
@@ -0,0 +1,358 @@
...
+  },
+  Or(*available_dcs): Case({
+      "consul": {
+          Or("host", "url"): Case({
+              "host": str,
+              "url": str})
+          },
+      "etcd": {
+          Or("host", "hosts", "srv", "url", "proxy"): Case({
+              "host": str,
```
CyberDem0n:
```diff
-               "host": str, 
+               "host": validate_host_port, 
```

patroni/validator.py
```diff
@@ -0,0 +1,358 @@
...
+        Optional("retry_timeout"): int,
+        Optional("maximum_lag_on_failover"): int
+        },
+    "pg_hba": [str],
+    "initdb": [Or(str, dict)]
+  },
+  Or(*available_dcs): Case({
+      "consul": {
+          Or("host", "url"): Case({
+              "host": str,
```
CyberDem0n:
```diff
-               "host": str, 
+               "host": validate_host_port, 
```

patroni/validator.py
CyberDem0n: No, it is actually the real list: https://github.com/zalando/patroni/blob/master/patroni/dcs/exhibitor.py#L49

patroni/validator.py
```diff
@@ -0,0 +1,358 @@
...
+          "hosts": str,
+          "port": lambda i: assert_(int(i) <= 65535),
+          Optional("pool_interval"): int
+          },
+      "zookeeper": {
+          "hosts": Or(comma_separated_host_port, [validate_host_port]),
+          },
+      "kubernetes": {
+          Optional("namespace"): str,
+          Optional("labels"): {},
```
CyberDem0n: labels is not an optional parameter, it must be present and must be a dict.

patroni/validator.py
yanchenko-igor: Yes, without checking status failed in python2.7

CyberDem0n: I mean the whole `__bool__()` method is unused and could be dropped.

yanchenko-igor: I understood that and removed the method

yanchenko-igor:
> I am also a bit worried that coverage shows different results on python2: https://travis-ci.org/zalando/patroni/jobs/625542854#L1239

Fixed

patroni/validator.py
```diff
@@ -0,0 +1,347 @@
...
+        return False
+    return data_directory_is_empty(data_dir)
+
+
+def validate_connect_address(address):
+    if not isinstance(address, string_types):
+        raise ConfigParseError("is not a string")
+    else:
+        for s in ["127.0.0.1", "0.0.0.0", "*", "::1"]:
+            if s in address:
```
CyberDem0n: The address is in format `host:port`, for we have to split first and check as:
```python
host in {"localhost", "127.0.0.1", "0.0.0.0", "*", "::1"}
```

patroni/validator.py
```diff
@@ -0,0 +1,347 @@
...
+      "zookeeper": {
+          "hosts": Or(comma_separated_host_port, [validate_host_port]),
+          },
+      "kubernetes": {
+          "labels": {},
+          Optional("namespace"): str,
+          Optional("scope_label"): str,
+          Optional("role_label"): str,
+          Optional("use_endpoints"): bool,
+          Optional("pod_ip"): Or(is_ipv4_address, is_ipv6_address),
```
CyberDem0n: The following snippet 
```yaml
kubernetes:
  pod_ip: blabla
```

makes the code to generate:
```
kubernetes.pod_ip blabla didn't pass validation: illegal IP address string passed to inet_aton
kubernetes.pod_ip blabla didn't pass validation: illegal IP address string passed to inet_pton
```

Very cryptic and not human-friendly at all :(

patroni/validator.py
```diff
@@ -0,0 +1,347 @@
...
+          Optional("ports"): [{"name": str, "port": int}],
+          },
+      }),
+  "postgresql": {
+    "listen": validate_host_port_listen,
```
CyberDem0n:
```yaml
postgresql:
  listen: '127.0.0.1,::1:5432'
```

is producing error:
```
postgresql.listen 127.0.0.1,::1:5432 didn't pass validation: gaierror(-2, 'Name or service not known')
```

But it is absolutely valid configuration.


Than I tried `listen: '::1:5432'`, and it is producing
```
postgresql.listen ::1:5432 didn't pass validation: gaierror(-9, 'Address family for hostname not supported')
```

patroni/validator.py
```diff
@@ -0,0 +1,347 @@
...
+  Or(*available_dcs): Case({
+      "consul": {
+          Or("host", "url"): Case({
+              "host": validate_host_port,
+              "url": str})
+          },
+      "etcd": {
+          Or("host", "hosts", "srv", "url", "proxy"): Case({
+              "host": validate_host_port,
+              "hosts": Or(comma_separated_host_port, [validate_host_port]),
```
CyberDem0n:
```yaml
etcd:
    hosts: '127.0.0.1:8500,127.0.0.1:2379'
```

producing:
```
etcd.hosts 127.0.0.1:8500,127.0.0.1:2379 didn't pass validation: '127.0.0.1:2379 is not reachable'
etcd.hosts 127.0.0.1:8500,127.0.0.1:2379 is not a list
```

With the first line I agree, but why there is the second line?

patroni/validator.py
```diff
@@ -0,0 +1,347 @@
...
+  "postgresql": {
+    "listen": validate_host_port_listen,
+    "connect_address": validate_connect_address,
+    "authentication": {
+      "replication": userattributes,
+      "superuser": userattributes,
+      "rewind":  userattributes
+    },
+    "data_dir": validate_data_dir,
+    "bin_dir": Directory(contains=["pg_ctl"], contains_executable=["pg_ctl", "initdb", "pg_controldata",
```
CyberDem0n: `bin_dir` is optional.

patroni/validator.py
```diff
@@ -0,0 +1,347 @@
...
+                                                                   "pg_basebackup", "postgres", "pg_isready"]),
+    "parameters": {
```
CyberDem0n: `parameters` is optional

patroni/validator.py
yanchenko-igor: There is "or" check, we expect either string "host1:port,host2:port" or list ["host1:port", "host2:port"] if neither of them worked we show both errors.

patroni/validator.py
yanchenko-igor: Fixed

patroni/validator.py
CyberDem0n: I understand that there is `Or`, but since we know for sure that the value is a string the second line doesn't make any sense.

yanchenko-igor: Fixed

patroni/validator.py
```diff
@@ -0,0 +1,379 @@
...
+
+def validate_host_port_listen_multiple_hosts(host_port):
+    return validate_host_port(host_port, listen=True, multiple_hosts=True)
+
+
+def is_ipv4_address(ip):
+    try:
+        socket.inet_aton(ip)
+    except Exception:
+        raise ConfigParseError("Is not valid ipv4 address")
```
avaczi: 'Is no _a_ valid ipv4 address' - but more importantly, don't you want to include the bogus value, just like elsewhere?

yanchenko-igor:
```% env PATRONI_CONFIGURATION='{"kubernetes":{"pod_ip":"bogus_value"}}'  python ./patroni.py --validate-config 
name  is not defined.
scope  is not defined.
restapi  is not defined.
kubernetes.labels  is not defined.
kubernetes.pod_ip bogus_value didn't pass validation: 'Is not valid ipv4 address'
kubernetes.pod_ip bogus_value didn't pass validation: 'Is not valid ipv6 address'
postgresql  is not defined.
```

FxKu:
:+1:

CyberDem0n:
:+1:

PR #1324, opened 2019-12-06
Make `dest` argument default value of `backup()` cross platform

### LINKED ISSUE #1325: `dest` argument default value is platform dependant
We should use `os.path.join` to handle path separators

pashagolub:
Fixes #1325


CyberDem0n:
:+1:

FxKu:
:+1:

Issue #1231, opened 2019-10-16
Patroni 1.6 the replica restarts cyclically 


uspen:
Hello!

Clean OS, cluster. After initializing replica restarts cyclically. But if master shut down, replica promoted and working good. (this not logged in attach)
And with patroni 1.5.6 all good.

At first it happened in a test environment like a production with CentOS 7. I tested patroni 1.5.6 and 1.6. Just downgrade patroni version and all working.
Then I set everything cleanly with my hands in CentOS 8 and patroni 1.6

```
CentOS Linux release 8.0.1905 (Core)
patroni version 1.6.0
postgresql-server-10.6 (from stock centos repo, pgdg not working)
python3-psycopg2-2.8.3-2.rhel8.x86_64
```
```
patronictl -c /etc/patroni.yml list shard01
+---------+--------+----------------+--------+---------+----+-----------+
| Cluster | Member |      Host      |  Role  |  State  | TL | Lag in MB |
+---------+--------+----------------+--------+---------+----+-----------+
| shard01 | node1  | 192.168.56.101 | Leader | running |  2 |         0 |
| shard01 | node2  | 192.168.56.102 |        | running |    |   unknown |
+---------+--------+----------------+--------+---------+----+-----------+
```

1.1 [master_patroni.txt](https://github.com/zalando/patroni/files/3735424/master_patroni.txt) - All good
1.2 [master_pg.txt](https://github.com/zalando/patroni/files/3735426/master_pg.txt) - All good

2.1 [replica_patroni.txt](https://github.com/zalando/patroni/files/3735427/replica_patroni.txt) - bad
2.2 [replica_pg.txt](https://github.com/zalando/patroni/files/3735428/replica_pg.txt) - bad
2.3 [master_pg_replica.txt](https://github.com/zalando/patroni/files/3735425/master_pg_replica.txt) - All good

[patroni.yml.txt](https://github.com/zalando/patroni/files/3738052/patroni.yml.txt)

CyberDem0n:
Does the replication password have some special characters?

CyberDem0n:
Reopening, it still needs to be fixed.

PR #1330, opened 2019-12-11
Handle case when replication password is set to null

CyberDem0n:
Fixes https://github.com/zalando/patroni/issues/1231


FxKu:
:+1:

CyberDem0n:
:+1:

Issue #665, opened 2018-04-17
Postgres state monitoring


ainlolcat:
We have custom scripts which every 30 second checks if postgres available for RW operations and performs swichover if postgres cannot serve RW requests and disk is almost full. Patroni does not check if master available for RW operations and keeps master the same even though postgres cannot perform single insert. 

So I wondering if patroni can check if postges is available for RW operations and hande some scenario (like full disk) or allow users to specify callbalck if some events occures.  

In first option patroni can check if postgres available for RW operation, consider other parameters ( someone put postgres in recovery mode manually for backup or something like that) and perform switchover on healthy replica if possible. 

In second option patroni can check if postgres availiable for RW operation and call callback with paramtetrs like connection to DB, occurrence count, state transition (RW>RO/RO>RW), uptime.

ainlolcat:
Yes. Not everyone likes someone messing around with database with RW requests. We have custom tests for simple RW check - insert/update/select/delete in special table which pretty small but still exists. So I prefer second option there user can specify additional healthcheck/reaction if required. It will become user headache to define cases which requires failover and cases which is just maintenance. Maybe some of such checks and reactions can be included in archive with patroni but disabled by default.

ants:
Can you describe some specific failure scenarios where master is not responding correctly and failing over would be helpful to the situation? In the full disk example, in a typical deployment all nodes have the same amount of disk, so failing over is unlikely to improve availability. More likely you will then have a full disk on all nodes and if implemented correctly a cascade of failovers across all nodes, if not, a neverending cycle of failovers.

Partial hardware failures (I/O errors or slowness, etc.) could and I think should be monitored and handled outside of Patroni. Probably by removing the whole node from the cluster. To turn a node readonly an external agent could already turn on nofailover tag of Patroni and trigger a failover.

ainlolcat:
In some cases (logging configuration error, core dumps, parts of some backup, another application error, stuck wals because of abandoned replication slot, etc) we can see full disk only on master. We do check if slave has enough free space to prevent from failover from full master to full slave.

CyberDem0n:
In general, I agree with @ants, I don't really like an idea of turning Patroni into a monitoring agent which also perform some reactive feedback actions.
It is basically your responsibility to set up an extensive monitoring system, which will do lots of checks, like CPU Utilization, free space on different partitions, hardware failures, number of open and active connections and many many more.

jberkus:
[...]

PR #1331, opened 2019-12-12
Test I/O before feeding the (watch)dog

jouir:
RDBMS, PostgreSQL included, are highly dependent of disk i/o to perform correctly. PostgreSQL doesn't work on read-only file systems. There is no point of letting a primary node alive when it can't write to its own data directory.

This commit implements a test file write in the data directory before feeding the watchdog. The primary host will reboot when there is a read-only file system or an i/o freeze longer than the watchdog timeout. Patroni will elect a new leader and recover the service automatically.

### DIFF
--- a/patroni/watchdog/base.py
+++ b/patroni/watchdog/base.py
@@ -1,6 +1,7 @@
 import abc
 import logging
 import platform
+import os
 import six
 import sys
 from threading import RLock
@@ -47,6 +48,7 @@ class WatchdogConfig(object):
         self.driver = config['watchdog'].get('driver', 'default')
         self.driver_config = dict((k, v) for k, v in config['watchdog'].items()
                                   if k not in ['mode', 'safety_margin', 'driver'])
+        self.test_file = os.path.join(config['postgresql']['data_dir'], 'watchdog')
 
     def __eq__(self, other):
         return isinstance(other, WatchdogConfig) and \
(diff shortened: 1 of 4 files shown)


jouir:
> Do you have some specific failure case you are protecting against with this?

We experience RAID card issues where any I/O operation is frozen for 5 to 10 minutes. No read. No write. Every query is stuck until the RAID card reset. For us, that means downtime and we can't afford that amount of downtime. During that window, there is no failover.

If you have MegaRAID card like us, you can fake a freeze with storcli (180 seconds in the example):

```
storcli /c0 start diag duration=180
```

> I think watchdog keepalive is the wrong place to do it. [...]

I'm agree with you. I will rework this branch to include your remarks as soon as possible. Thanks for the input! 👍

CyberDem0n:
> No read. No write. Every query is stuck until the RAID card reset.

I see. Recently we also experienced something similar when NVMe volume on i3 EC2 instance has failed.
But, Patroni was actually stuck in HA loop (because it is already trying to access the PGDATA in order to check that it is not disappeared). The leader lock expired, but failover didn't happen because the REST API was still working. The watchdog would not be pinged either and would have rebooted the machine, but we don't use it.
Meanwhile, in the REST API health-check Patroni runs a query, something like `SELECT pg_is_in_recovery(), pg_current_wal_lsn()`. Despite the PGDATA being inaccessible the monitoring query perfectly works.

I tend to agree with @ants that stepping down in such a situation shouldn't be the default behavior.

jouir:
I've worked this week on implementing healthchecks, as suggested here and in #665, but it requires more work than I thought. I'm closing this pull request because it's now off-topic. I will re-open one when it will be ready. Thanks again for your review!

Issue #1333, opened 2019-12-12
Patroni 1.6.1 continuously restarting PostgreSQL 12 replica


caseyallenshobe:
The following configuration was working perfectly with Patroni 1.6.0, as installed from apt.postgresql.org repo, with PostgreSQL 12 as well as older PostgreSQL releases.

At some point the package on that repo was replaced by a 1.6.1 one, and once I upgraded to that, the standby nodes will not come online.  Patroni ends up looping forever restarting PostgreSQL about once every second.

I tried to use a manual install of Patroni using pip, and that did not work either.  I then tried a manual install of version 1.6.0 using pip, and that is working, so it seem to be a problem introduced in 1.6.1.

Attached is my Patroni YML, the Patroni startup log, and what PostgreSQL logs each time it is started.
[patroni.zip](https://github.com/zalando/patroni/files/3957346/patroni.zip)

caseyallenshobe:
Working with Alexander Kukushkin on Slack, this can be worked around by putting the restore_command into postgresql.parameters.restore_command in Patroni's YML.

CyberDem0n:
Or in other words archive_cleanup_command, promote_trigger_file, recovery_end_command, recovery_min_apply_delay, and restore_command should not be used for comparison if they are not defined in the patroni config but defined in the configs other than postgresql.auto.conf and postgresql.conf

PR #1335, opened 2019-12-13
Allow certain recovery parameters be defined in the custom_conf

CyberDem0n:
Fixes https://github.com/zalando/patroni/issues/1333

### DIFF
--- a/patroni/postgresql/config.py
+++ b/patroni/postgresql/config.py
@@ -578,7 +578,7 @@ class ConfigHandler(object):
 
         try:
             values = self._get_pg_settings(self._recovery_parameters_to_compare).values()
-            values = {p[0]: [p[1], p[4] == 'postmaster'] for p in values}
+            values = {p[0]: [p[1], p[4] == 'postmaster', p[5]] for p in values}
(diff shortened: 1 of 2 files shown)


patroni/postgresql/config.py
```diff
@@ -687,6 +687,12 @@ def record_missmatch(mtype):
 
         wanted_recovery_params = self.build_recovery_params(member)
         for param, value in self._current_recovery_params.items():
+            # Skip certain parameters defined in the included postgres config files
+            # if we know that they are not specified in the patroni configuration.
```
sdudoladov: but changes in these params may still require reload / restart. Are they skipped simply because they are not a part of Patroni-managed configuration ?

CyberDem0n: They are skipped only when 1. Values are defined neither in patroni.yaml nor in DCS and 2. current values are coming not from expected places, i.e. `postgresql.conf` or `postgresql.auto.conf`.

That means user is managing them on they own via the `custom_conf` and therefore must do reload/restart.

avaczi:
:+1:

CyberDem0n:
:+1:

PR #1337, opened 2019-12-17
Make sure postgresql.pgpass is a file or it does not exist

yanchenko-igor:
We decided to not merge this change into #1291

### DIFF
--- a/patroni/postgresql/config.py
+++ b/patroni/postgresql/config.py
@@ -335,12 +335,10 @@ class ConfigHandler(object):
         self._standby_signal = os.path.join(postgresql.data_dir, 'standby.signal')
         self._auto_conf = os.path.join(postgresql.data_dir, 'postgresql.auto.conf')
         self._auto_conf_mtime = None
-        self._pgpass = config.get('pgpass') or os.path.join(os.path.expanduser('~'), 'pgpass')
+        self._pgpass = os.path.abspath(config.get('pgpass') or os.path.join(os.path.expanduser('~'), 'pgpass'))
         if os.path.exists(self._pgpass) and not os.path.isfile(self._pgpass):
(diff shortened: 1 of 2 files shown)


patroni/postgresql/config.py
```diff
@@ -335,6 +336,9 @@ def __init__(self, postgresql, config):
         self._auto_conf = os.path.join(postgresql.data_dir, 'postgresql.auto.conf')
         self._auto_conf_mtime = None
         self._pgpass = config.get('pgpass') or os.path.join(os.path.expanduser('~'), 'pgpass')
+        if os.path.exists(self._pgpass) and not os.path.isfile(self._pgpass):
```
CyberDem0n: It doesn't help much for the case when pgpass can't be created

yanchenko-igor: Should we try to create it from __init__ method?

CyberDem0n: I think we should try to open it in append mode.
If the file can't be created or isn't writable open will fail.

yanchenko-igor: If we create file anyway, should we write credentials or keep an empty file?

patroni/postgresql/config.py
```diff
@@ -713,6 +719,22 @@ def escape(value):
             record = {n: escape(record.get(n, '*')) for n in ('host', 'port', 'user', 'password')}
             return '{host}:{port}:*:{user}:{password}'.format(**record)
 
+    def write_empty_pgpass(self):
```
CyberDem0n: The name doesn't sound right, because we check that the file is writable.

patroni/postgresql/config.py
yanchenko-igor: I added the directory of pgpassfile to check_directory which makes sure it's writable.

avaczi:
:+1:

CyberDem0n:
:+1:

PR #1338, opened 2019-12-17
startup script changes

yanchenko-igor:
Some improvements to startup  script.


kien-truong:
Just a recommendation: the current settings `KillMode=process` does not cleanly stop patroni service if the children processes hang during shutdown. It should be changed to `KillMode=mixed`, which makes systemd kill the children processes as well, after the time out is reached.

CyberDem0n:
@mbanck, @df7cb maybe you can give some input on it?

df7cb:
Sounds plausible.

PR #1339, opened 2019-12-17
wait option for patronictl reinit implemented

yanchenko-igor:
Wait to finish reinit if --wait option is used.

### DIFF
--- a/patroni/api.py
+++ b/patroni/api.py
@@ -75,7 +75,6 @@ class RestApiHandler(BaseHTTPRequestHandler):
             lost = patroni.logger.records_lost
             if lost:
                 response['logger_records_lost'] = lost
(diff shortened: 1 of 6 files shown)


patroni/api.py
```diff
@@ -75,6 +75,7 @@ def _write_status_response(self, status_code, response):
             lost = patroni.logger.records_lost
             if lost:
                 response['logger_records_lost'] = lost
+        response['action'] = self.server.patroni.ha.action
```
CyberDem0n: I don't think we really need to introduce one more field in the response.
During reinitialize the state field is set to `creating replica`.

FxKu:
:+1:

CyberDem0n:
:+1:

PR #1341, opened 2019-12-20
Use restore_command from standby_cluster config on cascading replicas

CyberDem0n:
The standby_leader was already doing it from the beginning feature existed. Not doing the same on replicas might prevent them from catching up with standby leader due to WALs being recycled.

In addition to that apply the same strategy to archive_cleanup_command.


sdudoladov:
:+1:

CyberDem0n:
:+1:

PR #1342, opened 2019-12-20
implemented stop signal using pg_ctl for non posix systems

yanchenko-igor:
Using pg_ctl to send stop signal for non posix os.

### DIFF
--- a/patroni/postgresql/__init__.py
+++ b/patroni/postgresql/__init__.py
@@ -508,7 +508,7 @@ class Postgresql(object):
             self.set_state('stopping')
 
         # Send signal to postmaster to stop
-        success = postmaster.signal_stop(mode)
+        success = postmaster.signal_stop(mode, self._bin_dir)
         if success is not None:
             if success and on_safepoint:
                 on_safepoint()
@@ -526,10 +526,10 @@ class Postgresql(object):
         return True, True
 
     @staticmethod
-    def terminate_starting_postmaster(postmaster):
+    def terminate_starting_postmaster(postmaster, bin_dir):
         """Terminates a postmaster that has not yet opened ports or possibly even written a pid file. Blocks
         until the process goes away."""
-        postmaster.signal_stop('immediate')
+        postmaster.signal_stop('immediate', bin_dir)
         postmaster.wait()
 
     def _wait_for_connection_close(self, postmaster):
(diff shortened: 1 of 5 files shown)


patroni/postgresql/__init__.py
```diff
@@ -526,10 +526,10 @@ def _do_stop(self, mode, block_callbacks, checkpoint, on_safepoint):
         return True, True
 
     @staticmethod
-    def terminate_starting_postmaster(postmaster):
+    def terminate_starting_postmaster(postmaster, bin_dir):
```
CyberDem0n: Ouch...
Lets better remove `@staticmethod` decorator...

patroni/postgresql/__init__.py
```diff
@@ -508,7 +508,7 @@ def _do_stop(self, mode, block_callbacks, checkpoint, on_safepoint):
             self.set_state('stopping')
 
         # Send signal to postmaster to stop
-        success = postmaster.signal_stop(mode)
+        success = postmaster.signal_stop(mode, self._bin_dir)
```
CyberDem0n: I think it would be better to pass `self.pgcommand('pg_ctl')` instead of _bin_dir.

patroni/postgresql/postmaster.py
```diff
@@ -114,7 +114,10 @@ def signal_stop(self, mode):
             logger.warning("Cannot stop server; single-user server is running (PID: {0})".format(self.pid))
             return False
         try:
-            self.send_signal(STOP_SIGNALS[mode])
+            if os.name != 'posix':
+                return self.pg_ctl_stop(mode, bin_dir)
```
CyberDem0n: It would be better to move a call of `self.pg_ctl_kill` outside of try...catch block.

patroni/postgresql/postmaster.py
```diff
@@ -123,6 +126,18 @@ def signal_stop(self, mode):
 
         return None
 
+    def pg_ctl_stop(self, mode, bin_dir):
```
CyberDem0n: pg_ctl_kill

patroni/postgresql/postmaster.py
```diff
@@ -123,6 +126,18 @@ def signal_stop(self, mode):
...
+        pg_ctl = os.path.join(bin_dir, "pg_ctl")
+        SIGNALNAME = {"smart": "TERM", "fast": "INT", "immediate": "QUIT"}[mode]
+        status = subprocess.call([pg_ctl, "kill", SIGNALNAME, str(self.pid)])
```
CyberDem0n: `subprocess.call` may raise an exception

FxKu:
:+1:

CyberDem0n:
:+1:

PR #1349, opened 2020-01-02
Special handling of parameters with period

CyberDem0n:
It might be that they are defined by the extension and therefore the unit is not necessarily is the string.
It also could be that change of the value requires a restart (for example pg_stat_statements.max).


FxKu:
:+1:

CyberDem0n:
:+1:

Issue #1344, opened 2019-12-23
Patroni hangs after communication loss with DCS (Consul)


TestDrivenDevil:
Hi,

I'm doing network loss tests on a patroni cluster (1 master, 2 replicas)  and I see following behavior:  Each time patroni loses communication towards DCS, once communication is up again, the REST API of patroni is not responding anymore (it blocks on calling it). It's like patroni process is hanging. When I stop the service, it takes some time to stop and I see following lines in /var/log/messages, telling me the process is hanging.

```
systemd: patroni.service stop-sigterm timed out. Killing.
systemd: patroni.service: main process exited, code=killed, status=9/KILL
systemd: Unit patroni.service entered failed state.
systemd: patroni.service failed.
```

Looking for patroni logging I see following lines in /var/log/messages:

```
patroni: 2019-12-23 11:32:25,907 WARNING: Retry got exception: HTTPConnectionPool(host='192.168.1.10', port=8500): Max retries exceeded with url: /v1/session/renew/72d086a6-2fd5-ede2-fa31-529ea97bf1bc (Caused by ReadTimeoutError("HTTPConnectionPool(host='10.50.202.85', port=8500): Read timed out. (read timeout=3.33333333333)",))
patroni: 2019-12-23 11:32:25,907 ERROR: refresh_session
patroni: Traceback (most recent call last):
patroni: File "/usr/lib/python2.7/site-packages/patroni/dcs/consul.py", line 294, in refresh_session
patroni: return self.retry(self._do_refresh_session)
patroni: File "/usr/lib/python2.7/site-packages/patroni/dcs/consul.py", line 232, in retry
patroni: return self._retry.copy()(*args, **kwargs)
patroni: File "/usr/lib/python2.7/site-packages/patroni/utils.py", line 339, in __call__
patroni: raise RetryFailedError("Exceeded retry deadline")
patroni: RetryFailedError: 'Exceeded retry deadline'
python: detected unhandled Python exception in '/usr/bin/patroni'
patroni: Traceback (most recent call last):
patroni: File "/usr/bin/patroni", line 11, in <module>
patroni: load_entry_point('patroni==1.6.3', 'console_scripts', 'patroni')()
patroni: File "/usr/lib/python2.7/site-packages/patroni/__init__.py", line 221, in main
patroni: return patroni_main()
patroni: File "/usr/lib/python2.7/site-packages/patroni/__init__.py", line 189, in patroni_main
patroni: patroni.shutdown()
[...]

PR #1351, opened 2020-01-06
Handle exception from Ha.shutdown

CyberDem0n:
During the shutdown Patroni is trying to update its status in the DCS.
If the DCS is inaccessible an exception might be raised. Lack of exception handling prevents logger thread from stopping.

Fixes https://github.com/zalando/patroni/issues/1344

### DIFF
--- a/patroni/__init__.py
+++ b/patroni/__init__.py
@@ -159,7 +159,10 @@ class Patroni(object):
             self.api.shutdown()
         except Exception:
             logger.exception('Exception during RestApi.shutdown')
-        self.ha.shutdown()
+        try:
+            self.ha.shutdown()
+        except Exception:
+            logger.exception('Exception during Ha.shutdown')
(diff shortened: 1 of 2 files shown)


patroni/__init__.py
```diff
@@ -159,7 +159,10 @@ def shutdown(self):
             self.api.shutdown()
         except Exception:
             logger.exception('Exception during RestApi.shutdown')
-        self.ha.shutdown()
+        try:
+            self.ha.shutdown()
+        except Exception:
+            logger.exception('Exception during Ha.shutdown')
```
avaczi: Shouldn't we log the actual exception?

CyberDem0n: It will log the entire stack-trace.

FxKu:
:+1:

avaczi:
:+1:

Issue #1353, opened 2020-01-07
Randomly getting AttributeError exception with patronictl


stedounet:
### Version info

PostgreSQL 11.5
Patroni 1.6.0
Ubuntu 16 (container environnment)

### Problem

Hi

I sometime encounter this type of "error" while using patronictl : 

```
user@patroni_node_1:~$ patronictl -d etcd://127.0.0.1:2379 list patroni-cluster
+-----------------+-------------------+------------+--------+---------+----+-----------+
|     Cluster     |       Member      |    Host    |  Role  |  State  | TL | Lag in MB |
+-----------------+-------------------+------------+--------+---------+----+-----------+
| patroni-cluster | percona_patroni_1 | 172.17.0.2 |        | stopped |    |   unknown |
| patroni-cluster | percona_patroni_2 | 172.17.0.3 | Leader | running |  2 |       0.0 |
+-----------------+-------------------+------------+--------+---------+----+-----------+
Exception in thread Thread-1 (most likely raised during interpreter shutdown):
Traceback (most recent call last):
  File "/usr/lib/python2.7/threading.py", line 801, in __bootstrap_inner
  File "/opt/patroni/local/lib/python2.7/site-packages/patroni/dcs/etcd.py", line 46, in run
  File "/opt/patroni/local/lib/python2.7/site-packages/patroni/dcs/etcd.py", line 72, in _do_resolve
<type 'exceptions.AttributeError'>: 'NoneType' object has no attribute 'gaierror'
```

Though I do have what I want I wonder if I should worry about it.

I would be very interested to hear any potential impacts, or how to debug this block of code efficiently

CyberDem0n:
What member lists (clientUrls) does your etcd cluster exposes?
Does DNS work normally?
BTW, you are using python 2.7, which is EOL for the whole week.

stedounet:
Sorry, missed your reply

It exposes 172.17.0.2, 172.17.0.3, 172.17.0.4.
It's a container minimal test env, I try to keep things as small as possible to understand what I need to get it working properly and I do not have any DNS currently, is it an issue for patroni ?

Yes it is 2.7 , this specific error was from a Patroni installed via Percona packages, but I got it with a "proper" Patroni installed with pip3 too

PR #1355, opened 2020-01-09
Catch all exceptions when calling socket.getaddrinfo

CyberDem0n:
it serves two purposes:
1. We don't want accidentally break the thread
2. During the shutdown socket.gaierror become unresolvable and nasty exceptions are raised

Close: https://github.com/zalando/patroni/issues/1353


avaczi:
:+1:

CyberDem0n:
:+1:

PR #1361, opened 2020-01-13
Fix permissions for openshift block devices

pauvos:
OpenShift enforces securityContext.fsGroups for block devices and sets group stickybits for volumeMounts.

This leads to patroni pods failing to start after the first restart:
> 2020-01-13 14:46:13.695 UTC [143] FATAL:  data directory "/home/postgres/pgdata/pgroot/data" has invalid permissions
2020-01-13 14:46:13.695 UTC [143] DETAIL:  Permissions should be u=rwx (0700) or u=rwx,g=rx (0750).

A initContainer which fixes the OpenShift tampering solves the issue. I stole the solution from the stable postgres helm chart:
https://github.com/helm/charts/pull/14540/files

Tested on OpenShift v3.11

Note: This error does not occur when using shared filesystems (like NFS)

### DIFF
--- a/kubernetes/openshift-example/templates/template_patroni_persistent.yaml
+++ b/kubernetes/openshift-example/templates/template_patroni_persistent.yaml
@@ -106,6 +106,20 @@ objects:
           application: ${APPLICATION_NAME}
           cluster-name: ${PATRONI_CLUSTER_NAME}
       spec:
+        initContainers:
+        - command:
+          - chmod
+          - '0700'
+          - /home/postgres/pgdata/pgroot/data
+          image: docker-registry.default.svc:5000/${NAMESPACE}/patroni:latest
+          imagePullPolicy: IfNotPresent
+          name: fix-perms
+          resources: {}
+          terminationMessagePath: /dev/termination-log
+          terminationMessagePolicy: File
+          volumeMounts:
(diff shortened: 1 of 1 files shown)


kubernetes/openshift-example/templates/template_patroni_persistent.yaml
```diff
@@ -106,6 +106,20 @@ objects:
           application: ${APPLICATION_NAME}
           cluster-name: ${PATRONI_CLUSTER_NAME}
       spec:
+        initContainers:
+        - command:
+          - sh
+          - -c 
+          - "mkdir -p /home/postgres/pgdata/pgroot/data && chmod 0700 /home/postgres/pgdata/pgroot/data"
```
CyberDem0n: I would suggest to create/assign permissions only for `/home/postgres/pgdata/pgroot`

pauvos: I tried but it keeps complaining about permissions of the ./data directory. :-/

`2020-01-20 10:35:25.972 UTC [47] FATAL:  data directory "/home/postgres/pgdata/pgroot/data" has invalid permissions`

avaczi:
:+1:

CyberDem0n:
:+1:

PR #1366, opened 2020-01-21
Configure keepalive for connections to K8s API

CyberDem0n:
In case if we got nothing from the socket after the TTL seconds it should be considered dead.

### DIFF
--- a/patroni/dcs/kubernetes.py
+++ b/patroni/dcs/kubernetes.py
@@ -41,8 +41,22 @@ class CoreV1ApiProxy(object):
         self._request_timeout = None
         self._use_endpoints = use_endpoints
 
-    def set_timeout(self, timeout):
-        self._request_timeout = (1, timeout / 3.0)
+    def configure_timeouts(self, loop_wait, retry_timeout, ttl):
+        # Normally every loop_wait seconds we should have receive something from the socket.
+        # If we didn't received anything after the loop_wait + retry_timeout it is a time
+        # to start worrying (send keepalive messages). Finally, the connection should be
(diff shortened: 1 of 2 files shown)


patroni/dcs/kubernetes.py
```diff
@@ -41,8 +41,22 @@ def __init__(self, use_endpoints=False):
         self._request_timeout = None
         self._use_endpoints = use_endpoints
 
-    def set_timeout(self, timeout):
-        self._request_timeout = (1, timeout / 3.0)
+    def configure_timeouts(self, loop_wait, retry_timeout, ttl):
+        # Normally every loop_wait seconds we should have receive something from the socket.
+        # If we didn't received anything after the loop_wait + retry_timeout it is a time
+        # to start worrying (send keepalive messages). Finally, the connection should be
+        # considered as dead if we received nothing from the socket after the ttl seconds.
```
FxKu: Maybe shorten a bit here:
`# If not, start worrying (sending keepalive messages) and consider connection dead after ttl seconds.`

CyberDem0n: The sentence you suggest doesn't precisely describe what and why we are doing.

Jan-M:
I don't see where Patroni considers it dead or takes any action? Is nothing needed but configuring the timeouts?

CyberDem0n:
The OS will close such socket, Patroni will get an exception and retry in the loop: https://github.com/zalando/patroni/pull/1366/files#diff-87299faee0431ddce0848d6ff7726df7R183-R189

FxKu:
:+1:

CyberDem0n:
:+1:

Issue #1365, opened 2020-01-20
Password logging on user creation with passwordcheck


benjaminjb:
I'm using a postgres image with the passwordcheck library preloaded; and with a password generator to create a random password for the superuser and replication user for patroni. 

One time the password generator created a password that would not pass the passwordcheck requirements, and PSQL printed out the error (reasonable) and the statement that failed, producing something like:

`ERROR:  password is too short` 
`STATEMENT:  CREATE ROLE "hello" WITH LOGIN CREATEROLE CREATEDB PASSWORD 'world'`

I noticed that `create_or_update_role` does a SQL query to create/alter roles (see https://github.com/zalando/patroni/blob/master/patroni/postgresql/bootstrap.py#L303-L313) and I was wondering if there was ever any discussion about suppressing the error logs just for that query to prevent logging passwords. (Even if the password is rejected, I feel it's better to avoid printing any passwords.)

For reference on PSQL printing passwords, see https://www.postgresql.org/message-id/CAHJZqBDe1-oiYtdh5pcWLAhr3jsDMFQg4miv3PYADH3YWKqZ2w%40mail.gmail.com

CyberDem0n:
Indeed, we need to `set log_statement = none` before creating users.

PR #1370, opened 2020-01-22
Avoid logging passwords on user creation

CyberDem0n:
Fixes https://github.com/zalando/patroni/issues/1365

### DIFF
--- a/patroni/postgresql/bootstrap.py
+++ b/patroni/postgresql/bootstrap.py
@@ -311,11 +311,13 @@ BEGIN
     END IF;
 END;$$""".format(name, ' '.join(options))
         self._postgresql.query('SET log_statement TO none')
+        self._postgresql.query('SET log_min_duration_statement TO -1')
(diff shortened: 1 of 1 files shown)


patroni/postgresql/bootstrap.py
```diff
@@ -310,7 +310,13 @@ def create_or_update_role(self, name, password, options):
         CREATE ROLE "{0}" WITH {1};
     END IF;
 END;$$""".format(name, ' '.join(options))
-        self._postgresql.query(sql, *params)
+        self._postgresql.query('SET log_statement TO none')
+        self._postgresql.query("SET log_min_error_statement TO 'log'")
```
bradnicholson: If someone set's `log_min_duration` low enough or to 0, it will still log, will it not?

CyberDem0n: Good question, we need to check it

RafiaSabih:
:+1:

CyberDem0n:
:+1:

Issue #1385, opened 2020-02-08
regression tests fail if $HOME-directory is unavailable


mbanck:
```
____________________________ TestCtl.test_scaffold _____________________________

self = <tests.test_ctl.TestCtl testMethod=test_scaffold>
mock_get_dcs = <MagicMock name='get_dcs' id='139897113160400'>

    @patch('patroni.ctl.get_dcs')
    def test_scaffold(self, mock_get_dcs):
        mock_get_dcs.return_value = self.e
        mock_get_dcs.return_value.get_cluster = get_cluster_not_initialized_without_leader
        mock_get_dcs.return_value.initialize = Mock(return_value=True)
        mock_get_dcs.return_value.touch_member = Mock(return_value=True)
        mock_get_dcs.return_value.attempt_to_acquire_leader = Mock(return_value=True)
        mock_get_dcs.return_value.delete_cluster = Mock()
    
        with patch.object(self.e, 'initialize', return_value=False):
            result = self.runner.invoke(ctl, ['scaffold', 'alpha'])
            assert result.exception
    
        with patch.object(mock_get_dcs.return_value, 'touch_member', Mock(return_value=False)):
            result = self.runner.invoke(ctl, ['scaffold', 'alpha'])
            assert result.exception
    
        result = self.runner.invoke(ctl, ['scaffold', 'alpha'])
>       assert result.exit_code == 0
E       AssertionError: assert 1 == 0
E         -1
E         +0

tests/test_ctl.py:403: AssertionError
----------------------------- Captured stderr call -----------------------------
2020-02-08 18:58:35,372 ERROR: [Errno 13] Permission denied: '/home/mba'
2020-02-08 18:58:35,375 ERROR: [Errno 13] Permission denied: '/home/mba'
------------------------------ Captured log call -------------------------------
ERROR    patroni.utils:utils.py:434 [Errno 13] Permission denied: '/home/mba'
ERROR    patroni.utils:utils.py:434 [Errno 13] Permission denied: '/home/mba'
[...]
(unstable-amd64-sbuild)root@kohn:/tmp/autopkgtest.h72Lqd/build.Qtn/src# ls /home/mba
ls: cannot access '/home/mba': No such file or directory
```

`mba` is my user name - it is common that Debian build environments have non-existing home directories. I have not changed my setup, so this seems to be a regression in 1.6.4. I've just retested with 1.6.3 and it works fine there, but I have not diagnosed it further yet.

PR #1386, opened 2020-02-08
Set postgresql.pgpass to ./pgpass

mbanck:
This avoids test failures if $HOME is not available (fixes: #1385).

### DIFF
--- a/tests/test_ctl.py
+++ b/tests/test_ctl.py
@@ -32,7 +32,7 @@ def test_rw_config():
 
 
 @patch('patroni.ctl.load_config',
-       Mock(return_value={'scope': 'alpha', 'postgresql': {'data_dir': '.', 'pgpass': '/tmp/pgpass', 'parameters': {}, 'retry_timeout': 5},
+       Mock(return_value={'scope': 'alpha', 'postgresql': {'data_dir': '.', 'pgpass': './pgpass', 'parameters': {}, 'retry_timeout': 5},
                           'restapi': {'listen': '::', 'certfile': 'a'}, 'etcd': {'host': 'localhost:2379'}}))
 class TestCtl(unittest.TestCase):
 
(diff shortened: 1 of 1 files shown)


mbanck:
This was the wrong fix, I will have to look further.

CyberDem0n:
As a quick fix it looks ok (except that it breaks tests on windows), but the issue is a bit bigger. 
The test is failing due to the https://github.com/zalando/patroni/pull/1337. It tries to check that directory where `postgresql.pgpass` is located is writable, which is by default the home directory. Initializing Postgresql object from patronictl is a bit stupid, and should not be done at all. I will 

Also recently I came to the conclusion that just trying to create the random file in the directory where pgpass is supposed to be located is not the best idea either, because it might be that only the exact file is writable.

mbanck:
> As a quick fix it looks ok (except that it breaks tests on windows), 

I've moved it to `.` (`PGDATA`) instead of `/tmp` now, which I guess was the Windows issue?

> The test is failing due to the #1337. It tries to check that directory where `postgresql.pgpass` is located is writable, which is by default the home directory. 

I thought about it for a while and decided the logic in patroni is pretty ok - either you set the `pgpass` file somewhere where you want it (and can write to, of course), or it should use the home directory, as libpq does. In practise, not having a `$HOME` should make Patroni fail pretty early then and can be fixed or worked-around.

In the tests nothing should be written outside `.` and `/tmp` though in my opinion.

> Also recently I came to the conclusion that just trying to create the random file in the directory where pgpass is supposed to be located is not the best idea either, because it might be that only the exact file is writable.

The following python should check whether one can write to a directory: `os.access('/foo/bar', os.W_OK | os.X_OK) `. One still needs to check that the file is writable if it is already present though I think.

avaczi:
:+1:

CyberDem0n:
:+1:

PR #1390, opened 2020-02-13
Get rid of dependency on tzlocal module

CyberDem0n:
It was used only to add the local timezone to the datetime specified in the patronictl for scheduled switchover or restart.
The `dateutil.tz.tzlocal()` does the same job equally well.


avaczi:
:+1:

avaczi:
:+1:

CyberDem0n:
:+1:

PR #1392, opened 2020-02-14
Don't create permanent physical slot with name of the primary

### LINKED ISSUE #280: Replication slots don't prevent recycling of WAL when standby goes offline
When standby goes offline for a while the member key will expire and Patroni will happily remove its replication slot.  Having an archive works around this, but it would be nice to make simple deployments more resilient too. Probably needs some kind of persistent flag, or offline_ttl for cluster members so Patroni knows they will be coming back and doesn't remove their replication slots. And a corresponding administration command to kick them out when they actually need to go away.

CyberDem0n:
It is a regular issue that primary is recycling WALs when one of the replicas is down for a long time. So far there were only two solutions for such a problem and both of them are not perfect:
1. Increase `wal_keep_segments`, but it is hard to guess the good value.
2. Use continuous archiving and PITR, but it is not always possible.

This PR is introducing the way to solve the problem for static clusters, with a fixed number of nodes and names that never change. You just need to list the names of all nodes in the `slots` so the primary will not remove the slot when the node is down (not registered in DCS).
Of course, the primary will not create the permanent slot which is matching its own name.

Usage example: let's assume you have a cluster with nodes named *abc1*, *abc2*, and *abc3*.
You have to run `patronictl edit-config` and put the following snippet into the configuration:
```yaml
slots:
  abc1:
    type: physical
  abc2:
    type: physical
  abc3:
    type: physical
```

If the node *abc2* is the primary, it will always create slots for *abc1* and *abc3* even if they are not running, but will not create slot *abc2*.
Other nodes will behave the same.

Close #280

### DIFF
--- a/docs/SETTINGS.rst
+++ b/docs/SETTINGS.rst
@@ -20,7 +20,7 @@ Dynamic configuration is stored in the DCS (Distributed Configuration Store) and
 -  **synchronous\_mode\_strict**: prevents disabling synchronous replication if no synchronous replicas are available, blocking all client writes to the master. See :ref:`replication modes documentation <replication_modes>` for details.
 -  **postgresql**:
     -  **use\_pg\_rewind**: whether or not to use pg_rewind. Defaults to `false`.
-    -  **use\_slots**: whether or not to use replication_slots. Defaults to `true` on PostgreSQL 9.4+.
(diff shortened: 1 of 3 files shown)


sdudoladov:
> not remove the slot when the node is down (not registered in DCS)

imo sounds like a better name for the PR

sdudoladov:
:+1:

CyberDem0n:
:+1:

PR #1397, opened 2020-02-18
[doc] add PATRONICTL_CONFIG_FILE env var

daamien:
Hey guys,

A tiny patch to document the new variable introduced by PR #1150

### DIFF
--- a/docs/ENVIRONMENT.rst
+++ b/docs/ENVIRONMENT.rst
@@ -8,7 +8,6 @@ It is possible to override some of the configuration parameters defined in the P
 Global/Universal
 ----------------
 -  **PATRONI\_CONFIGURATION**: it is possible to set the entire configuration for the Patroni via ``PATRONI_CONFIGURATION`` environment variable. In this case any other environment variables will not be considered!
--  **PATRONI\_CONFIG\_FILE**: location of the configuration file. 
 -  **PATRONI\_NAME**: name of the node where the current instance of Patroni is running. Must be unique for the cluster.
 -  **PATRONI\_NAMESPACE**: path within the configuration store where Patroni will keep information about the cluster. Default value: "/service"
 -  **PATRONI\_SCOPE**: cluster name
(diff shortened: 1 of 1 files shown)


docs/ENVIRONMENT.rst
```diff
@@ -8,6 +8,7 @@ It is possible to override some of the configuration parameters defined in the P
 Global/Universal
 ----------------
 -  **PATRONI\_CONFIGURATION**: it is possible to set the entire configuration for the Patroni via ``PATRONI_CONFIGURATION`` environment variable. In this case any other environment variables will not be considered!
+-  **PATRONI\_CONFIG\_FILE**: location of the configuration file.
```
CyberDem0n: Hi @daamien 

It looks like typo, the name should be `PATRONICTL\_CONFIG\_FILE`, and the better place for it is the 
[CTL](https://github.com/zalando/patroni/pull/1397/files#diff-1eb5a66e4f8c5c7a581dfee209c23d29R132) section.

daamien: My bad ! I wrote this a bit too quickly. Here's a fix.

By the way, I noticed that the others PATRONI_CTL env variables in the  CTL section
( `PATRONI_CTL_INSECURE`, `PATRONI_CTL_CACERT`, etc. ) have inconsistant names ( PATRONICTL vs PATRONICTL) and theyr are absent in the code base. Maybe these variables are deprecated ?

CyberDem0n: Yes, there is some little inconsistency, and no, these variables are not deprecated.

`PATRONICTL_CONFIG_FILE` describes where the config file is located.

Patronictl uses the normal Patroni config file, where one can define the [ctl](https://patroni.readthedocs.io/en/latest/SETTINGS.html#ctl) section.

On top of that, there is a way to override any of the configuration settings with ENV variables.
Most of the variable names are following the pattern `PATRONI_<SECTION>_<PARAMETER>`.
The name of the section is `ctl`, parameter names are `insecure`, 'cacert', and so on.

FxKu:
:+1:

CyberDem0n:
:+1:

CyberDem0n:
Merged, thank you @daamien

daamien:
Thanks for the explanation !

PR #1399, opened 2020-02-19
Disable SSL verification for Consul

jouir:
Consul client uses urllib3 with a verify=True by default. When
SSL verification is disabled with verify=False, we can see
CERTIFICATE_VERIFY_FAILED exceptions. With urllib3 1.19.1-1 on
Debian Stretch, the "cert_reqs" argument  must be explicitaly set
to ssl.CERT_NONE to effectively disable SSL verification.


FxKu:
:+1:

CyberDem0n:
:+1:

CyberDem0n:
Merged, thank you @jouir

PR #1414, opened 2020-02-23
Note about possible data loss while canceling postgres backends.

michail-nikolaev:
Note about possible data loss while canceling postgres backends.

Related to zalando#1412

### DIFF
--- a/docs/replication_modes.rst
+++ b/docs/replication_modes.rst
@@ -57,6 +57,8 @@ You can ensure that a standby never becomes the synchronous standby by setting `
 
 Synchronous mode can be switched on and off via Patroni REST interface. See :ref:`dynamic configuration <dynamic_configuration>` for instructions.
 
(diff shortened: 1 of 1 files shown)


docs/replication_modes.rst
```diff
@@ -57,6 +57,8 @@ You can ensure that a standby never becomes the synchronous standby by setting `
 
 Synchronous mode can be switched on and off via Patroni REST interface. See :ref:`dynamic configuration <dynamic_configuration>` for instructions.
 
+Note: because of the way synchronous replication implemented in PostgreSQL it is still possible to lost transactions even using ``synchronous_mode_strict``. If PostgreSQL backend is canceled while waiting for aknowdledge of synchronous replication (as a result of packet cancellation due to client timeout or backend failure) transaction changes become visible for other backends. Such changes are non-replicated yet and may be lost in case of stanby promotion.
```
FxKu:
```suggestion
Note: Because of the way synchronous replication is implemented in PostgreSQL it is still possible to lose transactions even when using ``synchronous_mode_strict``. If the PostgreSQL backend is cancelled while waiting to acknowledge replication (as a result of packet cancellation due to client timeout or backend failure) transaction changes become visible for other backends. Such changes are not yet replicated and may be lost in case of standby promotion.
```

FxKu:
:+1:

CyberDem0n:
:+1:

PR #1422, opened 2020-02-27
Avoid opening replication connection on every cycle of HA loop

CyberDem0n:
Bug was introduces in the https://github.com/zalando/patroni/pull/1332


FxKu:
:+1:

CyberDem0n:
:+1:

PR #1424, opened 2020-02-27
Apply 1 second backoff if LIST failed

CyberDem0n:
It is mostly necessary to avoid flooding logs, but also help to prevent starvation of the main thread.


FxKu:
:+1:

CyberDem0n:
:+1:

PR #1431, opened 2020-03-10
Retry if the retry-after http header is set

CyberDem0n:
If the K8s API is overwhelmed with requests it might ask to retry.


avaczi:
:+1:

CyberDem0n:
:+1:

Issue #1434, opened 2020-03-11
Consul configuration section documentation doesn't match application behavior


awiller42:
In this section of the documentation
https://patroni.readthedocs.io/en/latest/SETTINGS.html#consul
one can find the following description for the `host` key:
`host: the host:port for the Consul endpoint, in format: http(s)://host:port`

When the URL is entered in the configuration as described, e.g. `host: http://127.0.0.1:8500`, patroni fails to start with an internal exception from urllib3.

By inserting a `print(self.uri(path, params))` before this line
https://github.com/zalando/patroni/blob/master/patroni/dcs/consul.py#L117
I found out that the URL gets assembled as follows: `http://[http://127.0.0.1]:8500`

This doesn't seem right. By changing the config key to `host: 127.0.0.1:8500` - removing the protocol prefix - I was able to solve this issue.

PR #1438, opened 2020-03-11
Correct documentation for consul.host, and use periods on bullet points consistently

caseyallenshobe:
Address concern in #1434

### DIFF
--- a/docs/SETTINGS.rst
+++ b/docs/SETTINGS.rst
@@ -87,8 +87,8 @@ Consul
 ------
 Most of the parameters are optional, but you have to specify one of the **host** or **url**
 
--  **host**: the host:port for the Consul endpoint.
--  **url**: url for the Consul endpoint, in format: http(s)://host:port.
+-  **host**: the host:port for the Consul local agent.
+-  **url**: url for the Consul local agent, in format: http(s)://host:port.
 -  **port**: (optional) Consul port.
 -  **scheme**: (optional) **http** or **https**, defaults to **http**.
 -  **token**: (optional) ACL token.
--- a/docs/SETTINGS.rst
+++ b/docs/SETTINGS.rst
@@ -87,20 +87,20 @@ Consul
 ------
 Most of the parameters are optional, but you have to specify one of the **host** or **url**
 
--  **host**: the host:port for the Consul endpoint, in format: http(s)://host:port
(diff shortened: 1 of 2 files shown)


caseyallenshobe:
First attempt at contributing something, please let me know if I did anything wrong.

docs/SETTINGS.rst
```diff
@@ -87,20 +87,20 @@ Consul
...
 
--  **host**: the host:port for the Consul endpoint, in format: http(s)://host:port
--  **url**: url for the Consul endpoint
--  **port**: (optional) Consul port
--  **scheme**: (optional) **http** or **https**, defaults to **http**
--  **token**: (optional) ACL token
--  **verify**: (optional) whether to verify the SSL certificate for HTTPS requests
+-  **host**: the host:port for the Consul endpoint.
+-  **url**: url for the Consul endpoint.
+-  **port**: (optional) Consul port.
```
CyberDem0n: I think for consistency with https://github.com/zalando/patroni/blob/7176a4155a61f46c6eb44c071c3e483eda09093f/docs/ENVIRONMENT.rst#consul it makes sense to move the `, in format: http(s)://host:port` to the line with `url`.

Since you already on it lets also replace `Consul` with `Consul local agent` on the lines with `host`, `port`, and `url` in this file and in the ENVIRONMENT.rst

caseyallenshobe: Ok, done, but does it have to be a local agent?  If so, why is host a parameter rather than always using localhost?  I assumed it could utilize a different host.  Also, isn't url duplication of schema, host, and port?

CyberDem0n: > does it have to be a local agent?

I think that in most cases yes.

> If so, why is host a parameter rather than always using localhost?

What if agents listen on a host different from localhost for some reason?

> I assumed it could utilize a different host.

With an agent running on a different host there is way more uncertainty. What if the different host is down?

> Also, isn't url duplication of schema, host, and port?

Yes, it duplicates them, but someone might find the url more convenient to use, especially when using environment variables to configure Patroni. It is enough to define just one variable instead of three.

FxKu:
:+1:

CyberDem0n:
:+1:

CyberDem0n:
Merged, thank you @caseyallenshobe

PR #1444, opened 2020-03-13
Feature: Tags/noslotdrop: retain inactive replication slots

ksarabu1:
Patroni drops unused & undefined (in slots section) replication slots by default including replica related replication slot when the replica is taken offline for planned maintenance. This affects the replica when it rejoins the cluster and if the required wal logs are already removed from the disk. 

A new tag **_noslotdrop_** (True|False, defaults to False) is being added as part of this change. When this tag is set to True (on master), Patroni will not drop the unused and undefined replication slots when the replicas are taken offline. This helps to retain wal logs on master when replicas are taken offline for maintenance. Please note that this may lead to a large number of wal logs accumulating in the file system and cause the file system to become full if the replicas are taken offline for an extended period.

## Tags     [Tags](https://patroni.readthedocs.io/en/latest/SETTINGS.html#tags)

- **noslotdrop**: ``true`` or ``false``. If set to ``true`` the node will retain the inactive replication slots. Defaults to ``false``. Patroni drops physical replication slots by default, when the replica is taken offline for maintenance. When set to ``true``, it retains the slot.

### DIFF
--- a/docs/SETTINGS.rst
+++ b/docs/SETTINGS.rst
@@ -240,3 +240,4 @@ Tags
 - **noloadbalance**: ``true`` or ``false``. If set to ``true`` the node will return HTTP Status Code 503 for the ``GET /replica`` REST API health-check and therefore will be excluded from the load-balancing. Defaults to ``false``.
 - **replicatefrom**: The IP address/hostname of another replica. Used to support cascading replication.
 - **nosync**: ``true`` or ``false``. If set to ``true`` the node will never be selected as a synchronous replica.
+- **noslotdrop**: ``true`` or ``false``. If set to ``true`` the node will retain the inactive replication slots. Defaults to ``false``. Patroni drops related physical replication slots by default, when the standby is taken offline for maintenance. When set to ``true``, it retains the slot.
--- a/patroni/__init__.py
+++ b/patroni/__init__.py
@@ -61,7 +61,7 @@ class Patroni(object):
 
     def get_tags(self):
         return {tag: value for tag, value in self.config.get('tags', {}).items()
-                if tag not in ('clonefrom', 'nofailover', 'noloadbalance', 'nosync') or value}
+                if tag not in ('clonefrom', 'nofailover', 'noloadbalance', 'nosync', 'noslotdrop') or value}
 
     @property
     def nofailover(self):
--- a/patroni/dcs/__init__.py
+++ b/patroni/dcs/__init__.py
@@ -196,6 +196,10 @@ class Member(namedtuple('Member', 'index,name,session,data')):
         return self.tags.get('clonefrom', False) and bool(self.conn_url)
 
     @property
+    def noslotdrop(self):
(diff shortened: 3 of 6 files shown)


CyberDem0n:
Have you seen https://github.com/zalando/patroni/pull/1392?
One can just define necessary permanent slots in the configuration with names matching node names and these slots will never be removed.

CyberDem0n:
Your approach will not work anyway, if the node is gone information about it i(including tags) is also gone from dcs, therefore slot will be removed.

ksarabu1:
> Have you seen #1392?
> One can just define necessary permanent slots in the configuration with names matching node names and these slots will never be removed.
> Your approach will not work anyway, if the node is gone information about it i(including tags) is also gone from dcs, therefore slot will be removed.

We discussed with you regarding the tag based approach back in Feb during conference call. Adding the hard-coded slot names as per the PR 1392 will work for maintaining the replication slots during the maintenance , but in large scale & dynamic environments it adds overhead of maintaining the configuration files.
We are trying to handle the replication slot issue by adding a tag thereby minimizing the changes to the configuration files.
If Tag is set on Master node for e.g., will not drop the replication slots that are being consumed by replica when it is taken offline for maintenance

CyberDem0n:
> large scale & dynamic environments it adds overhead of maintaining the configuration files.

Not really. Slots are configured in dcs, no need to change config files at all.

Your implementation has following problems:
1. You always check that noslotdrop is set for the current node by checking the member tags in dcs while this information could be just obtained from the local config and shouldn't be necessary exposed to dcs.
2. If the tag is set and node become the replica due to failover/switchover - it will not remove slots.
3. You will have to somehow monitor that slots are used and still remove them after some time or postgres will accumulate a lot of WAL files.

ksarabu1:
> Your implementation has following problems:
> 
> 1. You always check that noslotdrop is set for the current node by checking the member tags in dcs while this information could be just obtained from the local config and shouldn't be necessary exposed to dcs.
> 2. If the tag is set and node become the replica due to failover/switchover - it will not remove slots.
> 3. You will have to somehow monitor that slots are used and still remove them after some time or postgres will accumulate a lot of WAL files.

Thanks Alex for the feedback. 

(1) Sure. I will update the code.
(2) Correct. We initially updated the code for retaining slots only for master. I will push the change that keeps the slots only on master.
(3) As mentioned in our conf call, we already have monitoring in place for this.

ksarabu1:
Thanks for your help. 

We have agreed to use existing/implemented feature #1392 to predefine needed slots in dcs/config for now. Closing the PR.

CyberDem0n:
Sorry, I forgot to follow up no this today.
The main problem of your approach that for the end user it becomes too complicated to resolve further consequences (going to every node and removing unused slots) after enabling it. Nodes start running out of free space and people start opening issues here.

At the same time I really liked that PR have had not only feature, but also unit-tests!

Issue #1371, opened 2020-01-22
Failover when the postmaster is hung


kmsarabu:
We are looking into replacing Repmgr with Patroni and testing various failover options.

It seems that the patroni currently keeping persistent connection to the master database. For any reason if postmaster is hung, and that basically blocks all new connection requests to the master database. It appears that patroni was unable to identify such scenario and not triggering failover. 

We have also tested the same scenario with repmgr. repmgr was able to identify such case and failed over the database. Repmgr makes use of parameters reconnect_interval and reconnect_attempts instead of keeping a persistent connection & failover the database if the process unable connect to master in provided threshold.

Test case:

[host1] : ps -eaf |egrep postgres|egrep config |egrep listen_address
postgres 193623      1  0 17:13 ?        00:00:00 /postgres/product/server/10.3.2/bin/postgres -D /postgres/data001/pgrept1/sysdata --config-file=/postgres/data001/pgrept1/sysdata/postgresql.conf --listen_addresses=host1 --port=4345 --cluster_name=pgrept1 --wal_level=logical --hot_standby=on --max_connections=1000 --max_wal_senders=10 --max_prepared_transactions=0 --max_locks_per_transaction=64 --track_commit_timestamp=off --max_replication_slots=10 --max_worker_processes=8 --wal_log_hints=on

[host1] : kill -STOP 193623

[host1]  : psql -h `hostname` -p 4355 -U krishna
<HUNG>

From the patroni.log:
2020-01-22 13:35:35,082 INFO: Lock owner: host1; I am host1
2020-01-22 13:35:35,090 INFO: no action.  i am the leader with the lock
...
2020-01-22 13:37:05,082 INFO: Lock owner: host1; I am host1
2020-01-22 13:37:05,091 INFO: no action.  i am the leader with the lock
....

CyberDem0n:
Fairly speaking I have never ever seen a situation when the postmaster hung so that is not able to respond at all, but I have seen a lot of situations when all connection slots are occupied and therefore a connection is rejected.
Opening a new connection is a very expensive operation. If the system is under stress it could easily take seconds. Doing a failover in such case IMO is a bit insane, because:
1. first of all you'll have to fence the old primary. Depending on how you are doing fencing it might result in some data loss.
2. Clients which are currently connected would not be happy.
3. Highly likely that the new primary would become stressed even harder than the old one.

kmsarabu:
Thanks for quick response and looking into this issue.

[...]

PR #1445, opened 2020-03-16
Feature: Postgres stop timeout

ksarabu1:
## Feature: Postgres stop timeout

Switchover/Failover operation hangs on signal_stop (or checkpoint) call when postmaster doesn't respond or  hangs for some reason(Issue described in [1371](https://github.com/zalando/patroni/issues/1371)). This is leading to service loss for an extended period of time until the hung postmaster starts responding or it is killed by some other actor.

### master_stop_timeout

The number of seconds Patroni is allowed to wait when stopping Postgres and effective only when synchronous_mode is enabled. When set to > 0 and the synchronous_mode is enabled, Patroni sends SIGKILL to the postmaster if the stop operation is running for more than the value set by master_stop_timeout. Set the value according to your durability/availability tradeoff. If the parameter is not set or set <= 0, master_stop_timeout does not apply.


patroni/ha.py
```diff
@@ -94,6 +94,11 @@ def check_mode(self, mode):
         else:
             return self.patroni.config.check_mode(mode)
 
+    def master_stop_timeout(self):
+        """ Master stop timeout """
+        ret = parse_int(self.patroni.config.dynamic_configuration.get('master_stop_timeout'))
+        return ret if ret and ret > 0 and self.is_synchronous_mode() else None
+
```
CyberDem0n: Please take a look how `master_start_timeout` works, you need to "declare" in the [config](https://github.com/zalando/patroni/blob/master/patroni/config.py#L61) and later you can access it as `self.patroni.config['master_stop_timeout']`

patroni/ha.py
```diff
@@ -1066,7 +1072,8 @@ def restart(self, restart_data, run_async=False):
         self.set_start_timeout(timeout)
 
         # For non async cases we want to wait for restart to complete or timeout before returning.
-        do_restart = functools.partial(self.state_handler.restart, timeout, self._async_executor.critical_task)
+        do_restart = functools.partial(self.state_handler.restart, timeout, self._async_executor.critical_task,
```
CyberDem0n: I don't think that we need to apply the stop timeout to the restart procedure.
If there is explicit restart command for a specific node that means we this node to retain its role. In case if nodes is the master not doing a clean shutdown will increase the downtime.

patroni/postgresql/__init__.py
```diff
@@ -462,24 +465,33 @@ def start(self, timeout=None, task=None, block_callbacks=False, role=None):
...
         try:
-            with get_connection_cursor(**connect_kwargs) as cur:
-                cur.execute("SET statement_timeout = 0")
-                if check_not_is_in_recovery:
-                    cur.execute('SELECT pg_catalog.pg_is_in_recovery()')
-                    if cur.fetchone()[0]:
-                        return 'is_in_recovery=true'
-                return cur.execute('CHECKPOINT')
-        except psycopg2.Error:
+            return self.stop_retry(self.wait_for_checkpoint, check_not_is_in_recovery, connect_kwargs) if timeout \
```
CyberDem0n: While checkpoint is performed postgres continue working normally.
The only thing we need to apply here - `connect_timeout` and if it failed to connect proceed with the shutdown, we don't need to retry the CHECKPOINT.

patroni/postgresql/postmaster.py
```diff
@@ -113,10 +114,18 @@ def signal_stop(self, mode, pg_ctl='pg_ctl'):
...
         try:
+            children = self.children()
             self.send_signal(getattr(signal, 'SIG' + STOP_SIGNALS[mode]))
+            if mode == 'abort':
+                _, alive = psutil.wait_procs(children, timeout=1)
+                for child in alive:
+                    try:
+                        child.send_signal(getattr(signal, 'SIG' + STOP_SIGNALS[mode]))
+                    except psutil.NoSuchProcess:
+                        pass
```
CyberDem0n: I think it would be better to extract handling of `abort` into a separate method.
This is a very special operation and we don't even need to use the pg_ctl for that. Essentially we want to hardly kill postmaster and all ancestors.
In order to make sure that no new child processes are created by postmaster I would also `suspend` it. We already doing something similar when [terminating](https://github.com/zalando/patroni/blob/master/patroni/postgresql/cancellable.py#L34-L52) old callbacks

ksarabu1: Thankyou. I will update the code and the unit tests.

patroni/postgresql/postmaster.py
```diff
@@ -105,6 +106,34 @@ def from_pid(cls, pid):
...
+    def signal_kill(self):
+        """to suspend and kill postmaster and all children
+
+        :returns None if postmaster and children are killed, True if process is already gone, False if error
+        """
+        if self.is_single_user:
+            logger.warning("Cannot stop server; single-user server is running (PID: {0})".format(self.pid))
+            return False
+        plock = Lock()
+        with plock:
```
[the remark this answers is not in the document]
ksarabu1: Basically want to protect it from concurrent calls. Agree, local variable makes no difference, will update the code to use global variable.

patroni/postgresql/postmaster.py
```diff
@@ -105,6 +106,34 @@ def from_pid(cls, pid):
...
+            try:
+                self.suspend()
+                children = self.children()
+                self.kill()
+            except psutil.NoSuchProcess:
+                return True
```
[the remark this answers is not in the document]
ksarabu1: Got it, we are looking at a condition where Postmaster is killed, but the children are still alive. Sure, will fix the code to handle that.

patroni/postgresql/postmaster.py
```diff
@@ -105,6 +106,34 @@ def from_pid(cls, pid):
...
+            except psutil.AccessDenied as e:
+                logger.warning("Could not kill PostgreSQL (error: {0})".format(e))
+                return False
+            _, alive = psutil.wait_procs(children, timeout=3)
```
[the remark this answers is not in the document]
ksarabu1: Sure, will update the code to come out immediately.

patroni/postgresql/postmaster.py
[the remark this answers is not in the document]
CyberDem0n: It is already protected, just remove it completely.

patroni/postgresql/postmaster.py
```diff
@@ -105,6 +106,34 @@ def from_pid(cls, pid):
...
             return None
 
+    def signal_kill(self):
+        """to suspend and kill postmaster and all children
+
+        :returns None if postmaster and children are killed, True if process is already gone, False if error
+        """
+        if self.is_single_user:
+            logger.warning("Cannot stop server; single-user server is running (PID: {0})".format(self.pid))
+            return False
```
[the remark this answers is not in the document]
ksarabu1: Will update.

patroni/postgresql/__init__.py
```diff
@@ -531,10 +534,34 @@ def _do_stop(self, mode, block_callbacks, checkpoint, on_safepoint):
...
+
         postmaster.wait()
 
         return True, True
 
+    def terminate_postmaster(self, postmaster, mode, stop_timeout):
+        if mode in ['fast', 'smart']:
+            try:
+                success = postmaster.signal_stop('immediate', self.pgcommand('pg_ctl'))
+                if success is not None:
```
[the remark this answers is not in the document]
ksarabu1: Thanks. Fixing the code.

patroni/postgresql/__init__.py
```diff
@@ -531,10 +534,34 @@ def _do_stop(self, mode, block_callbacks, checkpoint, on_safepoint):
             postmaster.wait_for_user_backends_to_close()
             on_safepoint()
 
+        if stop_timeout:
```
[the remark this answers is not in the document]
ksarabu1: Agree, no need for redundant checks. will fix the code.

patroni/postgresql/postmaster.py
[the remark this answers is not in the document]
ksarabu1: Fixed.

patroni/postgresql/postmaster.py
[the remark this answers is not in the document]
ksarabu1: Fixed.

patroni/postgresql/postmaster.py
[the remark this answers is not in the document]
ksarabu1: Fixed.

patroni/postgresql/postmaster.py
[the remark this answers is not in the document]
ksarabu1: Fixed.

patroni/postgresql/__init__.py
```diff
@@ -531,10 +534,34 @@ def _do_stop(self, mode, block_callbacks, checkpoint, on_safepoint):
...
+            try:
+                success = postmaster.signal_stop('immediate', self.pgcommand('pg_ctl'))
+                if success is not None:
+                    return True
+                postmaster.wait(timeout=stop_timeout)
+                return True
+            except TimeoutExpired:
+                pass
+        logger.warning("Sending SIGKILL to Postmaster and its children")
+        return True if postmaster.signal_kill() else False
```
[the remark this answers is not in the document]
ksarabu1: Fixed.

patroni/postgresql/__init__.py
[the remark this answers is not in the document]
ksarabu1: Fixed.

patroni/postgresql/__init__.py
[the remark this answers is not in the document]
ksarabu1: Fixed.

-

patroni/postgresql/__init__.py
```diff
@@ -610,7 +631,7 @@ def wait_for_startup(self, timeout=None):
 
         return self.state == 'running'
 
-    def restart(self, timeout=None, task=None, block_callbacks=False, role=None):
+    def restart(self, timeout=None, task=None, block_callbacks=False, role=None, stop_timeout=None):
```
CyberDem0n: The `restart()` is never called with `stop_timeout` argument.

patroni/ha.py
```diff
@@ -1083,7 +1089,7 @@ def restart(self, restart_data, run_async=False):
                 return (False, 'restart failed')
 
     def _do_reinitialize(self, cluster):
-        self.state_handler.stop('immediate')
+        self.state_handler.stop('immediate', stop_timeout=self.master_stop_timeout())
```
CyberDem0n: This code-path is never executed when postgres is running as master, but IMO it still makes sense to pass non-zero value here. It would theoretically help with the case when postgres doesn't respond to immediate shutdown.

patroni/ha.py
```diff
@@ -1156,7 +1162,7 @@ def post_recover(self):
     def cancel_initialization(self):
         logger.info('removing initialize key after failed attempt to bootstrap the cluster')
         self.dcs.cancel_initialization()
-        self.state_handler.stop('immediate')
+        self.state_handler.stop('immediate', stop_timeout=self.master_stop_timeout())
```
CyberDem0n: Here we just want to make sure that postgres isn't running. IMO it would still make sense to pass non-zero value here. It would theoretically help with the case when postgres doesn't respond to immediate shutdown.

patroni/ha.py
```diff
@@ -1279,7 +1285,7 @@ def _run_cycle(self):
             # is data directory empty?
             if self.state_handler.data_directory_empty():
                 self.state_handler.set_role('uninitialized')
-                self.state_handler.stop('immediate')
+                self.state_handler.stop('immediate', stop_timeout=self.master_stop_timeout())
```
CyberDem0n: If the data directory isn't present or empty we want to make sure that postgres isn't running. It would theoretically help with the case when postgres is alive, but doesn't respond to immediate shutdown.

patroni/postgresql/postmaster.py
```diff
@@ -105,6 +105,29 @@ def from_pid(cls, pid):
...
+            self.suspend()
+            children = self.children()
+            self.kill()
+        except psutil.NoSuchProcess:
+            pass
+        except psutil.AccessDenied as e:
+            logger.warning("Could not kill PostgreSQL (error: {0})".format(e))
+            return False
+        _, alive = psutil.wait_procs(children, timeout=0)
+        for child in alive:
```
CyberDem0n:
```suggestion
        for child in children:
```

patroni/postgresql/postmaster.py
```diff
@@ -105,6 +105,29 @@ def from_pid(cls, pid):
...
+            try:
+                child.kill()
+            except psutil.NoSuchProcess:
+                continue
+
```
CyberDem0n: I wonder, should we wait for processes before returning from the method?
```python
psutil.wait_procs(children + [self])
```

patroni/postgresql/__init__.py
ksarabu1: Yes, missed out on removing the reference after code update. will fix it.

patroni/ha.py
ksarabu1: Agree, Do you have any number in mind? I am leaning towards 1 minute timeout provided that the call is intended for reinitialize. Thanks.

patroni/postgresql/postmaster.py
ksarabu1: It probably helps in case of Zombie state.

patroni/ha.py
ksarabu1: Make sense. Should we give precedence to the stop timeout set in config and override it only if its null?

patroni/ha.py
ksarabu1: Agree, especially in case where the data dir is empty and the master is hung. Should we give precedence to the stop timeout set in config and override it only if its null?

CyberDem0n: The same as https://github.com/zalando/patroni/pull/1445/files#diff-57c53e988b775ee2902488fb33c1f433R1092
```suggestion
                self.state_handler.stop('immediate', stop_timeout=self.patroni.config['retry_timeout'])
```

patroni/postgresql/postmaster.py
```diff
@@ -105,6 +105,28 @@ def from_pid(cls, pid):
...
             return None
 
+    def signal_kill(self):
+        """to suspend and kill postmaster and all children
+
+        :returns True if postmaster and children are killed, False if error
+        """
+        try:
+            self.suspend()
+            children = self.children()
```
ksarabu1:
```suggestion
            children = self.children(recursive=True)
```

patroni/postgresql/postmaster.py
```diff
@@ -105,6 +105,28 @@ def from_pid(cls, pid):
...
+            pass
+        except psutil.AccessDenied as e:
+            logger.warning("Could not kill PostgreSQL (error: {0})".format(e))
+            return False
+        for child in children:
+            try:
+                child.kill()
+            except psutil.NoSuchProcess:
+                continue
+
```
ksarabu1:
```suggestion
        _, _ = psutil.wait_procs(children.append(self))
```

patroni/postgresql/__init__.py
ksarabu1:
```suggestion
    def restart(self, timeout=None, task=None, block_callbacks=False, role=None):
```

patroni/postgresql/__init__.py
```diff
@@ -621,7 +642,7 @@ def restart(self, timeout=None, task=None, block_callbacks=False, role=None):
         self.set_state('restarting')
         if not block_callbacks:
             self.__cb_pending = ACTION_ON_RESTART
-        ret = self.stop(block_callbacks=True) and self.start(timeout, task, True, role)
+        ret = self.stop(block_callbacks=True, stop_timeout=stop_timeout) and self.start(timeout, task, True, role)
```
ksarabu1:
```suggestion
        ret = self.stop(block_callbacks=True) and self.start(timeout, task, True, role)
```

patroni/postgresql/postmaster.py
ksarabu1: Please review the following changes. Thanks.

patroni/postgresql/postmaster.py
CyberDem0n: 1. The `list.append()` method doesn't return anything
2. We can discards results of `psutil.wait_procs()`

patroni/postgresql/postmaster.py
CyberDem0n: :+1:

patroni/config.py
```diff
@@ -59,6 +59,7 @@ class Config(object):
         'maximum_lag_on_failover': 1048576,
         'check_timeline': False,
         'master_start_timeout': 300,
+        'master_stop_timeout': '',
```
CyberDem0n, line 62:
```suggestion
        'master_stop_timeout': 0,
```

patroni/ha.py
```diff
@@ -94,6 +94,11 @@ def check_mode(self, mode):
         else:
             return self.patroni.config.check_mode(mode)
 
+    def master_stop_timeout(self):
+        """ Master stop timeout """
+        ret = parse_int(self.patroni.config['master_stop_timeout'])
+        return ret if ret and ret > 0 and self.is_synchronous_mode() else None
```
ksarabu1, line 100:
```suggestion
        return ret if ret > 0 and self.is_synchronous_mode() else None
```

CyberDem0n: From the config key we can get anything, not necessarily integer. In this case parse_int() returns None.
The old version could deal with it.

patroni/postgresql/postmaster.py
ksarabu1: Fixed

patroni/ha.py
ksarabu1: :+1:

sdudoladov:
:+1:

CyberDem0n:
:+1:

PR #1452, opened 2020-03-19
WIP: cluster-health API path and patronictl command

mbanck:
This PR tries to improve monitoring by defining an overall cluster-health, similar to e.g. `etcdctl cluster-health`. This considers a cluster healthy if: (i) there is a leader, (ii)  every follower is replicating from it on the same timeline and (iii) the replication lag is smaller than `maximum_lag_on_failover`.

Major changes are:

*  Implement global cluster health

This adds an API path `/cluster_health` that returns 200 if the overall cluster
is considered healthy and 503 if not. Also, a `cluster-health' command is
added that exposes this in patronictl. If the cluster is healthy,
patronictl exits with exit code 0. If a leader exists, but the cluster is
otherwise unhealthy, the exit status 1. If not leader exists, the exit status is 2.

* Make replica role more explicit.

Only set it to `replica` if node is in archive recovery, set it to empty
otherwise. In `cluster_as_json()`, return the actual role and not just `replica`
for all non-leader nodes. Finally, explicitly print replica nodes as `Follower`
in `patronictl list` in order to distinguish them from nodes which are not
replicating.

Examples:

```

# patronictl -c /etc/patroni/11-test.yml list
+---------+--------+-----------------+----------+---------+----+-----------+
| Cluster | Member |       Host      |   Role   |  State  | TL | Lag in MB |
+---------+--------+-----------------+----------+---------+----+-----------+
| 11-test |  pg1   | 192.168.122.206 |  Leader  | running |  3 |           |
| 11-test |  pg2   |  192.168.122.71 | Follower | running |  3 |         0 |
| 11-test |  pg3   | 192.168.122.225 | Follower | running |  3 |         0 |
+---------+--------+-----------------+----------+---------+----+-----------+
# patronictl -c /etc/patroni/11-test.yml cluster-health
cluster is healthy

# patronictl -c /etc/patroni/11-test.yml list
+---------+--------+-----------------+---------------+---------+----+-----------+
| Cluster | Member |       Host      |      Role     |  State  | TL | Lag in MB |
+---------+--------+-----------------+---------------+---------+----+-----------+
| 11-test |  pg1   | 192.168.122.206 | Uninitialized | stopped |    |   unknown |
| 11-test |  pg2   |  192.168.122.71 |     Leader    | running | 14 |           |
| 11-test |  pg3   | 192.168.122.225 | Uninitialized | stopped |    |   unknown |
+---------+--------+-----------------+---------------+---------+----+-----------+
# patronictl -c /etc/patroni/11-test.yml cluster-health
cluster has leader but is not healthy
# echo $?
1
```


patroni/utils.py
```diff
@@ -448,3 +449,40 @@ def data_directory_is_empty(data_dir):
...
+                logger.warning('cluster is not healthy: leader does not have role master or is not in state running')
+                return False
+            # check replication data from leader and make sure there are as many rows as replicas
+            try:
+                response = patroni.request(m, timeout=2, retries=0)
+                data = json.loads(response.data.decode('utf-8'))
+            except Exception as e:
+                logger.warning("Request failed to %s: GET %s (%s)", m.name, m.api_url, e)
+            if not 'replication' in data or len(data['replication']) + 1 != len(cluster.members):
+                logger.warning('cluster is not healthy: not all members take part in replication')
```
CyberDem0n: Well, some members might do cascading replication:
master <- cascading_replica <- replica2 {tags: {replicatefrom: cascading_replica}}

patroni/utils.py
```diff
@@ -448,3 +449,40 @@ def data_directory_is_empty(data_dir):
...
+                return False
+    if cluster.config:
+        maximum_lag_on_failover = cluster.config.data.get('maximum_lag_on_failover', 0)
+    for m in cluster.members:
+        if m.name == leader_name:
+            continue
+        if m.data.get('timeline', '') != leader_tl or int(m.data.get('lag', 0)) > maximum_lag_on_failover:
+            logger.warning('cluster is not healthy: timeline mismatch in member %s', m.name)
```
CyberDem0n: The warning is misleading, while in fact it could be on the same timeline but replication is lagging.
It is also possible that the replica is tagged with `nofailover` and  `recovery_min_apply_delay` is set to some high values.

patroni/utils.py
```diff
@@ -448,3 +449,40 @@ def data_directory_is_empty(data_dir):
...
+            return False
+        if m.data.get('role', '') != 'replica' or m.data.get('state', '') != 'running':
```
CyberDem0n: The role could be `sync_standby`.

patroni/ctl.py
```diff
@@ -1220,6 +1220,23 @@ def history(obj, cluster_name, fmt):
...
 
 
+@ctl.command('cluster-health', help="Show the overall health of the cluster")
+@arg_cluster_name
+@click.pass_obj
+def cluster_health(obj, cluster_name):
+    cluster = get_dcs(obj, cluster_name).get_cluster()
+    if cluster.leader:
+        member = cluster.leader.member
+        r = request_patroni(member, 'get', 'cluster_health')
```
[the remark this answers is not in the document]
mbanck: I don't see `request_patroni()` be wrapped in `try/except` elsewhere, or do you mean something else?

patroni/utils.py
mbanck: About the former, I would say the cluster isn't very healthy in this case, but the warning message could be improved to add the max log thing.

About the latter, I will need to think a bit more about this; maybe in this case the user cannot expect this feature to work correctly

patroni/ctl.py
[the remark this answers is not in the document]
CyberDem0n: Sometimes it is wrapped: https://github.com/zalando/patroni/blob/93efa91bbdc5a50cad4f99c4e99286340ab94ce8/patroni/ctl.py#L1018-L1023

You only check the response code, but if a request to the leader API fails (network issues) the exception will be raised and we need to handle it.

mbanck:
Hrm, that merge looks weird

mbanck:
> Hrm, that merge looks weird

That looks better after a force-push

patroni/utils.py
mbanck: The second paragraph is addressed by #2089/#2108

PR #1461, opened 2020-03-25
Update Dockerfile

0m1xa:
On postgres:12 find command without this pattern deletes file i18n_ctype which is needed by localdef. And localedef exit with code !=0


FxKu:
:+1:

CyberDem0n:
:+1:

CyberDem0n:
Merged, thank you @0m1xa

PR #1466, opened 2020-03-27
Better patronictl help text for the "flush" subcommand

kmoppel:
Right now it's not really clear from --help what it does, flushing in software
context usually means persisting...so actually it's the opposite, so make
the intention more explicit.


FxKu:
:+1:

CyberDem0n:
:+1:

PR #1467, opened 2020-03-27
Patronictl - fail if a config file is specified explicitly but not found

kmoppel:
For some reason patronictl currently does not complain when an explicitly provided config file does not exists. Is that on purpose? 
Countless times I've written something like that:
```
$ python3 patronictl.py -c postgresql0.yml list
2020-03-27 13:28:15,773 - WARNING - Listing members: No cluster names were provided
```
With the patched version it would bail out if non-default file location does not exist:
```
$ python3 patronictl.py -c postgresql0.yml list
Error: Provided config file postgresql0.yml not existing or no read rights. Check the -c/--config-file parameter
```

### DIFF
--- a/patroni/ctl.py
+++ b/patroni/ctl.py
@@ -69,7 +69,11 @@ def load_config(path, dcs):
     from patroni.config import Config
 
     if not (os.path.exists(path) and os.access(path, os.R_OK)):
-        logging.debug('Ignoring configuration file "%s". It does not exists or is not readable.', path)
+        if path != CONFIG_FILE_PATH:    # bail if non-default config location specified but file not found / readable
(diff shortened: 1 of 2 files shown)


CyberDem0n:
Hi @kmoppel 
you need to update [tests](https://travis-ci.org/github/zalando/patroni/jobs/667659794#L1286) and probably fix formatting (python setup.py flake8)

kmoppel:
@CyberDem0n Hey, gotcha - should be Ok now. First time messing around with Patroni code :)

FxKu:
:+1:

CyberDem0n:
:+1:

PR #1478, opened 2020-04-02
WIP: Note that restart is required for Consul reconfiguration

NikolayS:
TODO

- [ ] check (or add) words about how to restart properly
- [ ] what is the exact list of params changing which requires a restart?

### DIFF
--- a/docs/SETTINGS.rst
+++ b/docs/SETTINGS.rst
@@ -85,7 +85,9 @@ Bootstrap configuration
 
 Consul
 ------
-Most of the parameters are optional, but you have to specify one of the **host** or **url**
+Most of the parameters are optional, but you have to specify one of the **host** or **url**.
+
+*Restart is required:* when changing these parameters, sending SIGHUP to the Patroni patroni process is not enough, you need to restart it.
 
 -  **host**: the host:port for the Consul local agent.
 -  **url**: url for the Consul local agent, in format: http(s)://host:port.


docs/SETTINGS.rst
```diff
@@ -85,7 +85,9 @@ Bootstrap configuration
 
 Consul
 ------
-Most of the parameters are optional, but you have to specify one of the **host** or **url**
+Most of the parameters are optional, but you have to specify one of the **host** or **url**.
+
+*Restart is required:* when changing these parameters, sending SIGHUP to the Patroni patroni process is not enough, you need to restart it.
```
NikolayS, line 90:
```suggestion
*Restart is required:* when changing these parameters, sending SIGHUP to the Patroni process is not enough, you need to restart it.
```

docs/SETTINGS.rst
```diff
@@ -85,7 +85,9 @@ Bootstrap configuration
 
 Consul
 ------
-Most of the parameters are optional, but you have to specify one of the **host** or **url**
+Most of the parameters are optional, but you have to specify one of the **host** or **url**.
+
+*Restart is required:* when changing these parameters, sending SIGHUP to the Patroni process is not enough, you need to restart it.
```
CyberDem0n, line 90: It is possible to change token and consistency without restart.

NikolayS:
@CyberDem0n adjusted the text and marked `token` and `consistency` as not requiring a restart to apply a change.

What about etcd, zookeeper, and k8s parameter? Is it the same, a restart is needed when we need to change them?

CyberDem0n:
For consul there is also `dc` that could be changed without a restart (not sure if somebody even using it).
The kubernetes configuration supposed to be static due to its nature and therefore restart is always required.
For etcd there is nothing yet what could be changed without a restart, but in the future `username`  and `password` will be supported.
For zookeeper there is nothing to change without a restart.

Issue #1463, opened 2020-03-26
patronictl list not work after recovery from "no space left"


qingguee:
Hello,
We meet "no space left" issue in our WAL file folder, after some clean up, we have enough space.
But "patronictl list" not work.

```
db-pg-0:/ # patronictl list
Traceback (most recent call last):
  File "/usr/local/bin/patronictl", line 11, in <module>
    load_entry_point('patroni==1.6.3', 'console_scripts', 'patronictl')()
  File "/usr/local/lib/python2.7/site-packages/click/core.py", line 829, in __call__
    return self.main(*args, **kwargs)
  File "/usr/local/lib/python2.7/site-packages/click/core.py", line 782, in main
    rv = self.invoke(ctx)
  File "/usr/local/lib/python2.7/site-packages/click/core.py", line 1259, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
  File "/usr/local/lib/python2.7/site-packages/click/core.py", line 1066, in invoke
    return ctx.invoke(self.callback, **ctx.params)
  File "/usr/local/lib/python2.7/site-packages/click/core.py", line 610, in invoke
    return callback(*args, **kwargs)
  File "/usr/local/lib/python2.7/site-packages/click/decorators.py", line 33, in new_func
    return f(get_current_context().obj, *args, **kwargs)
  File "/usr/local/lib/python2.7/site-packages/patroni/ctl.py", line 781, in members
    output_members(cluster, cluster_name, extended, fmt)
  File "/usr/local/lib/python2.7/site-packages/patroni/ctl.py", line 707, in output_members
    cluster = cluster_as_json(cluster)
  File "/usr/local/lib/python2.7/site-packages/patroni/utils.py", line 391, in cluster_as_json
    conn_kwargs = m.conn_kwargs()
  File "/usr/local/lib/python2.7/site-packages/patroni/dcs/__init__.py", line 160, in conn_kwargs
    r = urlparse(self.conn_url)
  File "/usr/lib64/python2.7/urlparse.py", line 143, in urlparse
    tuple = urlsplit(url, scheme, allow_fragments)
  File "/usr/lib64/python2.7/urlparse.py", line 201, in urlsplit
    i = url.find(':')
AttributeError: 'NoneType' object has no attribute 'find'
```
- Patroni config
```
bootstrap:
  dcs:
    synchronous_mode: true
    postgresql:
      use_pg_rewind: true
  initdb:
  - auth-host: md5
[...]

Issue #1100, opened 2019-07-03
Patronictl Error


anikin-aa:
Hi, we have faced next error during patronictl call
```
Traceback (most recent call last):
  File "/patroni/patronictl.py", line 5, in <module>
    ctl(None)
  File "/usr/lib/python2.7/site-packages/click/core.py", line 764, in __call__
    return self.main(*args, **kwargs)
  File "/usr/lib/python2.7/site-packages/click/core.py", line 717, in main
    rv = self.invoke(ctx)
  File "/usr/lib/python2.7/site-packages/click/core.py", line 1137, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
  File "/usr/lib/python2.7/site-packages/click/core.py", line 956, in invoke
    return ctx.invoke(self.callback, **ctx.params)
  File "/usr/lib/python2.7/site-packages/click/core.py", line 555, in invoke
    return callback(*args, **kwargs)
  File "/usr/lib/python2.7/site-packages/click/decorators.py", line 27, in new_func
    return f(get_current_context().obj, *args, **kwargs)
  File "/patroni/patroni/ctl.py", line 814, in members
    output_members(cluster, cluster_name, extended, fmt)
  File "/patroni/patroni/ctl.py", line 723, in output_members
    append_port = any(str(m.conn_kwargs()['port']) != '5432' for m in cluster.members) or\
  File "/patroni/patroni/ctl.py", line 723, in <genexpr>
    append_port = any(str(m.conn_kwargs()['port']) != '5432' for m in cluster.members) or\
  File "/patroni/patroni/dcs/__init__.py", line 155, in conn_kwargs
    r = urlparse(self.conn_url)
  File "/usr/lib64/python2.7/urlparse.py", line 142, in urlparse
    tuple = urlsplit(url, scheme, allow_fragments)
  File "/usr/lib64/python2.7/urlparse.py", line 196, in urlsplit
    i = url.find(':')
AttributeError: 'NoneType' object has no attribute 'find'
```

Could you please help with this ?
From trace I can see, that patroni is trying to parse conn_url from cluster member. 
What is the potential problem can be ?
All of the pods got parsable conn_url in annontation.

We are running patroni in OpenShift, using ConfigMaps as a DCS.
Patroni ver 1.5.6

Thanks.

CyberDem0n:
Could you please change the `log.level` to `DEBUG` and post here the logs?

anikin-aa:
[...]

PR #1482, opened 2020-04-07
Handle cases when conn_url is not defined

CyberDem0n:
On K8s when one of the Patroni pods in starting there is valid annotation yet, which could cause failure in patronictl.
In addition to that handle cases if port isn't specified in the standby_cluster configuration.

Close https://github.com/zalando/patroni/issues/1100
Close https://github.com/zalando/patroni/issues/1463


sdudoladov:
:+1:

CyberDem0n:
:+1:

PR #1490, opened 2020-04-14
To issue checkpoint after the promotion

rjayakri:
After the failover, the old primary waits on checkpoint to happen on the new primary before starting the rewind process.
This change request is to do a checkpoint on the primary right after the promotion to prevent the message like
2020-04-12 12:33:08,427 INFO 140054969964288 rewind.py execute 171: Waiting for checkpoint on <new_primary> before rewind


patroni/postgresql/__init__.py
```diff
@@ -738,6 +738,8 @@ def promote(self, wait_seconds, on_success=None, access_is_restricted=False):
             if not access_is_restricted:
                 self.call_nowait(ACTION_ON_ROLE_CHANGE)
             ret = self._wait_promote(wait_seconds)
+            if ret:
```
CyberDem0n, line 741: Unfortunately such an approach fixes the problem only partially. In case if it took longer than wait_seconds for postgres to finish recovery the checkpoint will not be executed explicitly. There is alo another issue, which is even more problematic. Promotion is executed in a thread and the main thread is waiting for promotion thread to finish. When you issue a checkpoint here, you are potentially delaying the moment when patroni starts reporting this node as a real primary, what is undesirable.

The correct way of doing checkpoint after promotion is starting the new thread from HA loop.

rjayakri: Thanks for the suggestion. Will update the code with async call from ha loop following the promote.

rjayakri: Thanks for the suggestion. We will update the code with async call from ha loop following the promote.

rjayakri:
Closing this pull request. Issue with my github caused multiple commits to this PR.
Will open a new PR including the above suggestions.

Issue #1487, opened 2020-04-11
timeline history consuming space in dcs


ksarabu1:
Do you foresee any issues with trimming the data in dcs to last 20 (or configurable) timeline history points?

```
zk: xxxxxxx:xxxx: get /......../xyz/history
....
dataLength = 31117
....
```

CyberDem0n:
I have an idea to make it configurable how many lines to keep there.

ksarabu1:
Do you see any advantage in keeping this configurable over a hard limit of 20? I have made changes to the code using hard limit and also an additional issue addressing replica reinitialize  when user-defined tablespaces are present. I can submit PR if hard-limit is acceptable. Otherwise, I will use configurable and submit changes.
Thanks.

CyberDem0n:
I just envision that after making the hard limit people will start asking why is it so small or large.

ksarabu1:
Agree. But, I think too many parameters would unnecessarily complicate and trying to see if we can avoid introducing new parameter for such a change where it won't make much of a difference for a custom-value over hard-limit?
Thanks.

CyberDem0n:
I agree that 20 is a reasonable default. It should be ok for most of setups, but in case if someone is unhappy there'll be a way to change it.

I'm actually talking about global configuration, similar to maximum_lag_on_failover.

ksarabu1:
Thanks. Will update the code to to use dcs param for time history points to retain in dcs & submit changes for both the issues.

PR #1491, opened 2020-04-14
max timelines in dcs

ksarabu1:
### timeline history consuming space in dcs

This is to address the issue from #[1487](https://github.com/zalando/patroni/issues/1487). A new global parameter max_timelines_history for keeping maximum timelines history in DCS. Default: 0. When set to 0, it keeps all timelines history in dcs.


ksarabu1:
Looks like its the issue with test case. [Switchover](https://patroni.readthedocs.io/en/latest/pause.html#the-implementation) supposed to go through in paused mode when leader and candidate info is provided.

```
curl -u xyz:xyzxyz -X POST http://pgtest132008/switchover --data '{"leader": "pgtest1"}'
Switchover is possible only to a specific candidate in a paused state
```

patroni/postgresql/misc.py
```diff
@@ -58,8 +58,12 @@ def parse_lsn(lsn):
...
 
 
-def parse_history(data):
-    for line in data.split('\n'):
+def parse_history(data, max_timelines=0):
+    timelines = None
+    if max_timelines is not None and abs(max_timelines) > 0:
+        # twice the requested as the timeline history has empty lines
+        timelines = -abs(max_timelines*2)
+    for line in data.split('\n')[timelines:]:
```
CyberDem0n, line 66: 1. I wouldn't rely on empty lines, because there is no guaranty that it will not change in the future.
2. I also think that passing the `max_timelines` argument all the way down to this function is unnecessary. Do we really want to save a few cpu cycles on parsing the complete history file? It would be good enough to just change the https://github.com/zalando/patroni/pull/1491/files#diff-57c53e988b775ee2902488fb33c1f433R525 to
```python
               for line in history[-max_timelines:]:
```

patroni/postgresql/__init__.py
```diff
@@ -801,6 +801,15 @@ def remove_data_directory(self):
...
 
+                pg_tblsp_path = os.path.join(self._data_dir, 'pg_tblspc')
+                if os.path.exists(pg_tblsp_path):
+                    for tspdn in [os.path.join(pg_tblsp_path, dn) for dn in os.listdir(pg_tblsp_path)
+                                  if parse_int(dn) and os.path.islink(os.path.join(pg_tblsp_path, dn))]:
+                        pg_tsp_path = os.path.realpath(tspdn)
+                        if pg_tsp_path is not None and os.path.exists(pg_tsp_path):
+                            logger.info('Removing user defined tablespace directory: %s', pg_tsp_path)
+                            shutil.rmtree(pg_tsp_path, ignore_errors=True)
+
```
CyberDem0n, line 812: This change is irrelevant to max timelines, please move it into a separate PR.
Also using the list comprehension is a bit inappropriate in the for loop, because later you are doing additional checks anyway.

patroni/postgresql/misc.py
ksarabu1: 1. Agree. We can move that logic to get_history to trim timelines.
2. Will update.

patroni/postgresql/__init__.py
ksarabu1: Sure. Will submit new PR for this change.

patroni/dcs/__init__.py
```diff
@@ -535,6 +535,10 @@ def timeline(self):
                 return 1
         return 0
 
+    @property
+    def max_timelines_history(self):
+        return self.config.data.get('max_timelines_history', 0)
+
```
CyberDem0n, line 541: The `max_timelines_history` is a configuration parameter, so it is more appropriate to put it into the [ClusterConfig](https://github.com/zalando/patroni/pull/1491/files#diff-0964cdd11e7e37271af75805b15fbcd2R319) class.

patroni/ha.py
```diff
@@ -514,12 +514,13 @@ def while_not_sync_standby(self, func):
     def update_cluster_history(self):
         master_timeline = self.state_handler.get_master_timeline()
         cluster_history = self.cluster.history and self.cluster.history.lines
+
```
CyberDem0n, line 517: Unrelated change.

patroni/dcs/__init__.py
ksarabu1: Thanks. Updating the code.

patroni/ha.py
ksarabu1: Thanks. Removed.

patroni/postgresql/__init__.py
```diff
@@ -679,7 +679,7 @@ def replica_cached_timeline(self, master_timeline):
     def get_master_timeline(self):
         return self._cluster_info_state_get('timeline')
 
-    def get_history(self, timeline):
+    def get_history(self, timeline, max_timelines=0):
```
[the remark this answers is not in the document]
ksarabu1: Thanks. Removed.

docs/SETTINGS.rst
```diff
@@ -15,6 +15,7 @@ Dynamic configuration is stored in the DCS (Distributed Configuration Store) and
 -  **ttl**: the TTL to acquire the leader lock (in seconds). Think of it as the length of time before initiation of the automatic failover process. Default value: 30
 -  **retry\_timeout**: timeout for DCS and PostgreSQL operation retries (in seconds). DCS or network issues shorter than this will not cause Patroni to demote the leader. Default value: 10
 -  **maximum\_lag\_on\_failover**: the maximum bytes a follower may lag to be able to participate in leader election.
+-  **max\_timelines\_history**: maximum timelines history should be kept in DCS. Default value: 0. When set to 0, it keeps all timelines history in dcs.
```
avaczi, line 18: I suggest changing the wording to 
'...: maximum number of timeline history items kept in DCS.  Default value: 0. When set to 0, it keeps the full history in DCS.'

Jan-M:
Looks good and also appreciated making this configurable. Difficult to anticipate what other users expect as a good default.

Jan-M:
:+1:

CyberDem0n:
:+1:

PR #1492, opened 2020-04-15
Simplify Dockerfile to avoid the unnecessary problems caused by compression

cobolbaby:
If COMPRESS is specified during construction, an error will be reported during startup.

```shell
# prod@tao-infra-01 /opt/patroni/bin $ docker logs -f postgres_pg01.xxx
# decompressing image...
# sudo: unable to resolve host pg01
# sudo: unable to resolve host pg01
# hostname: unrecognized option '--ip-address'
# BusyBox v1.22.1 (Debian 1:1.22.0-19+b3) multi-call binary.

# Usage: hostname [OPTIONS] [HOSTNAME | -F FILE]

# Get or set hostname or DNS domain name

# 	-s	Short
# 	-i	Addresses for the hostname
# 	-d	DNS domain name
# 	-f	Fully qualified domain name
# 	-F FILE	Use FILE's content as hostname

# python3: error while loading shared libraries: libexpat.so.1: cannot open shared object file: No such file or directory
```

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,18 +1,13 @@
 ## This Dockerfile is meant to aid in the building and debugging patroni whilst developing on your local machine
 ## It has all the necessary components to play/debug with a single node appliance, running etcd
 ARG PG_MAJOR=10
-ARG COMPRESS=false
 ARG PGHOME=/home/postgres
 ARG PGDATA=$PGHOME/data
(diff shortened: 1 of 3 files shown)


CyberDem0n:
Compression is optional, nobody forcing you to use it.

cobolbaby:
Yeah well that might be my OCD :)

PR #1494, opened 2020-04-15
Reinitialize Bug fix with user defined tablespace

ksarabu1:
### reinit not working when user-defined tablespaces exist

During reinit, Patroni removing pgdata and leaving user-defined tablespace directory. This is causing Patroni to loop in reinit.


CyberDem0n:
:+1:

FxKu:
:+1:

PR #1497, opened 2020-04-16
To issue checkpoint after the promotion

rjayakri:
After the failover, the old primary waits on checkpoint to happen on the new primary before starting the rewind process.
This change request is to do a checkpoint on the primary right after the promotion to prevent the message like
2020-04-12 12:33:08,427 INFO 140054969964288 rewind.py execute 171: Waiting for checkpoint on <new_primary> before rewind

Reference to  PR https://github.com/zalando/patroni/pull/1490


CyberDem0n:
@rjayakri it turned out to be more complex than anticipated: https://github.com/zalando/patroni/pull/1498

PR #1498, opened 2020-04-17
Issue CHEKPOINT explicitely after promote happened

CyberDem0n:
It is safe to call pg_rewind on the replica only when pg_control on the primary contains information about the latest timeline. Postgres is usually doing immediate checkpoint right after promote and in most cases it works just fine. Unfortunately we regularly receive complaints that it takes to long (minutes) until the checkpoint is done and replicas can't perform rewind. At the same time doing the checkpoint manually immediately helped. So Patroni starts doing the same. When the promotion happened and postgres is not running in recovery, we explicitly issue the checkpoint.

We are intentionally not using the AsyncExecutor here, because we want the HA loop continues doing its normal flow.

### DIFF
--- a/patroni/ha.py
+++ b/patroni/ha.py
@@ -944,7 +944,7 @@ class Ha(object):
                     return msg
 
                 # check if the node is ready to be used by pg_rewind
-                self._rewind.check_for_checkpoint_after_promote()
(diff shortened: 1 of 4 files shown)


Jan-M:
:+1:

CyberDem0n:
:+1:

PR #1499, opened 2020-04-17
Etcd smart refresh members

CyberDem0n:
In dynamic environments it is common that during the rolling upgrade etcd nodes are changing their IP addresses. If the etcd node where Patroni is currently connected to is upgraded last, it could happen that the cached topology doesn't contain any live node anymore and therefore request can't be retried and totally fails, usually resulting in demoting of the primary.

In order to partially overcome the problem, Patroni is already doing a periodic (every 5 minutes) rediscovery of the etcd cluster topology, but in case of very fast node rotation there was still a possibility to hit the issue.

This PR is an attempt to address the problem. If the list of nodes exhausted, Patroni will try to perform initial discovery via an external mechanism, like resolving A or SRV dns records and if the new list is different from the original, Patroni will use it as the new etcd cluster topology.

In order to deal with tcp issues the connect_timeout is set to max(read_timeout/2, 1). It will make list of members exhaust faster, but leaves the time to perform topology rediscovery and another attempt.

The third issue addressed by this PR - it could happen that dns names of etcd nodes didn't change, but ip addresses are new, therefore we clean up the internal dns cache when doing topology rediscovery.

Besides that, this commit makes `_machines_cache` property pretty much static, it will be updated only when the topology has changed and helps to avoid concurrency issues.

### DIFF
--- a/patroni/dcs/etcd.py
+++ b/patroni/dcs/etcd.py
@@ -194,14 +194,14 @@ class Client(etcd.Client):
                 response.data.decode('utf-8')
                 self._check_cluster_id(response)
                 if some_request_failed:
-                    self._base_uri = base_uri
+                    self.set_base_uri(base_uri)
                     self._refresh_machines_cache()
                 return response
             except (HTTPError, HTTPException, socket.error, socket.timeout) as e:
                 # switch to the next etcd node because we don't know exactly what happened,
                 # whether the key didn't received an update or there is a network problem.
                 if not retry and i + 1 < len(machines_cache):
-                    self._base_uri = machines_cache[i + 1]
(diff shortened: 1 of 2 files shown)


Jan-M:
:+1:

patroni/dcs/etcd.py
```diff
@@ -155,58 +165,58 @@ def machines(self):
         Also this method implements the same timeout-retry logic as `api_execute`, because
         the original method was retrying 2 times with the `read_timeout` on each node."""
 
-        kwargs = self._build_request_parameters()
+        machines_cache = self.machines_cache
```
avaczi, line 168: Is `self.machines_cache` a typo?

avaczi:
:+1:

PR #1501, opened 2020-04-17
Skip missing values from pg_controldata

feikesteenbergen:
Skip missing values from pg_controldata

When calling controldata(), it may return an empty dictionary, which in
turn caused the following error to occur:

    effective_configuration
        cvalue = parse_int(data[cname])
    KeyError: 'max_wal_senders setting'

Instead of causing a crash of this part, we now log the error and
continue.

This is the full output of the error:
```
2020-04-17 14:31:54,791 ERROR: Exception during execution of long running task restarting after failure
  File "/usr/lib/python3/dist-packages/patroni/async_executor.py", line 97, in run
    wakeup = func(*args) if args else func()
  File "/usr/lib/python3/dist-packages/patroni/postgresql/__init__.py", line 707, in follow
    self.start(timeout=timeout, block_callbacks=change_role, role=role)
  File "/usr/lib/python3/dist-packages/patroni/postgresql/__init__.py", line 409, in start
    configuration = self.config.effective_configuration
  File "/usr/lib/python3/dist-packages/patroni/postgresql/config.py", line 983, in effective_configuration
    cvalue = parse_int(data[cname])
KeyError: 'max_wal_senders setting'
```

### DIFF
--- a/patroni/postgresql/__init__.py
+++ b/patroni/postgresql/__init__.py
@@ -664,7 +664,7 @@ class Postgresql(object):
                 data = subprocess.check_output([self.pgcommand('pg_controldata'), self._data_dir], env=env)
                 if data:
                     data = data.decode('utf-8').splitlines()
-                    # pg_controldata output depends on major verion. Some of parameters are prefixed by 'Current '
+                    # pg_controldata output depends on major version. Some of parameters are prefixed by 'Current '
                     result = {l.split(':')[0].replace('Current ', '', 1): l.split(':', 1)[1].strip() for l in data
                               if l and ':' in l}
             except subprocess.CalledProcessError:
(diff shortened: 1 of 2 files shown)


FxKu:
:+1:

Jan-M:
What's the impact of just skipping this? Old config values? Wrong config values?

CyberDem0n:
Values are missing because someone is trying to use binaries which version doesn't match with PGDATA.
Patroni will try to start postgres anyway, postgres will complain that major version doesn't match and abort with the error.

Jan-M:
:+1:

feikesteenbergen:
I'll dive a bit deeper, because, yes, this is a container with multiple binaries, but the PATH and PGDATA both point to PostgreSQL 12.

feikesteenbergen:
Hmm, this may actually show a more interesting issue, it seems `initdb` was run not from `bin_dir` but from the regular `$PATH` variable. Will investigate even more.

PR #1514, opened 2020-04-28
Separate received and replayed location

CyberDem0n:
When making a decision whether the running replica is able to stream from the new primary or must be rewound we should use replayed location, therefore we extract received and replayed independently.

Reuse the part of the query that extracts the timeline and locations in the REST API.


erthalion:
:+1:

CyberDem0n:
:+1:

PR #1515, opened 2020-04-28
Sync HA loops of all pods in one cluster

CyberDem0n:
There is no expire mechanism available on K8s, therefore we implement soft leader lock, i.e. every pod is "watching" for changes of the leader object and when there are no changes during the TTL it starts leader race.

Before we switched to LIST+WATCH approach in #1189 and #1276, we only watched for the leader object and every time it was updated, the main thread of the HA loop was waking up. As a result, all replica pods were synchronized, and starting the leader race more or less at the same time.

The new approach made all pods "unsynchronized" and the biggest downside of it - it takes `ttl + loop_wait` in the worst case to detect the leader failure.

This commit makes all pods in one cluster to sync HA loops again based on updates of the leader object.

### DIFF
--- a/patroni/dcs/kubernetes.py
+++ b/patroni/dcs/kubernetes.py
@@ -187,7 +187,7 @@ class ObjectCache(Thread):
                     if value_changed:
                         logger.debug('%s changed from %s to %s', name, old_value, new_value)
 
-                    # We don't want to wakeup HA loop if we run as leader and received leader object update event!
+                    # Do not wake up HA loop if we run as leader and received leader object update event
                     if value_changed or name == self._dcs.leader_path and self._name != new_value:
                         self._dcs.event.set()
         finally:
--- a/patroni/dcs/kubernetes.py
+++ b/patroni/dcs/kubernetes.py
@@ -98,7 +98,7 @@ def catch_kubernetes_errors(func):
 
 class ObjectCache(Thread):
 
-    def __init__(self, dcs, func, retry, condition):
+    def __init__(self, dcs, func, retry, condition, name=None):
         Thread.__init__(self)
         self.daemon = True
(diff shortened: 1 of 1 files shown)


patroni/dcs/kubernetes.py
```diff
@@ -180,9 +181,14 @@ def _build_cache(self):
...
 
-                    if old_value != new_value and \
-                            (name != self._dcs.config_path or old_value is not None and new_value is not None):
+                    value_changed = old_value != new_value and \
+                        (name != self._dcs.config_path or old_value is not None and new_value is not None)
+
+                    if value_changed:
                         logger.debug('%s changed from %s to %s', name, old_value, new_value)
+
+                    # We don't want to wakeup HA loop if we run as leader and received leader object update event!
```
FxKu, line 190:
```suggestion
                    # Do not wake up HA loop if we run as leader and received leader object update event
```

patroni/dcs/kubernetes.py
```diff
@@ -576,6 +582,6 @@ def watch(self, leader_index, timeout):
             return True
 
         try:
-            return super(Kubernetes, self).watch(None, timeout)
+            return super(Kubernetes, self).watch(None, timeout + 0.5)
```
FxKu, line 585: the 0.5 looks a bit random, but ok

CyberDem0n: It is [consistently](https://github.com/zalando/patroni/blob/master/patroni/dcs/etcd.py#L661) "random" :)

FxKu: then maybe create a constant or config for this offset :)

FxKu:
:+1:

CyberDem0n:
:+1:

PR #1522, opened 2020-05-06
Detect a new timeline in the standby cluster

CyberDem0n:
The standby cluster doesn't know about leader elections in the main cluster and therefore the usual mechanisms of detecting divergences don't work. For example, it could happen that the standby cluster is ahead of the new primary of the main cluster and must be rewound.
There is a way to know that the new timeline has been created by checking the presence of a history file in pg_wal. If the new file is there, we will start usual procedures of making sure that we can
continue streaming or will run the pg_rewind.

### DIFF
--- a/patroni/ha.py
+++ b/patroni/ha.py
@@ -398,6 +398,10 @@ class Ha(object):
             self.demote('immediate-nolock')
             return demote_reason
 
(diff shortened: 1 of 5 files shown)


avaczi:
:+1:

CyberDem0n:
:+1:

PR #1524, opened 2020-05-06
Wake up the main thread after checkpoint is done

CyberDem0n:
Replicas are waiting for checkpoint indication via member key of the leader in DCS. The key is normally updated only one time per HA loop.
Without waking the main thread up replicas will have to wait up to `loop_wait` seconds longer than necessary.


FxKu:
:+1:

CyberDem0n:
:+1:

PR #1525, opened 2020-05-06
Use pg_rewind with --restore-target-wal on 13 if possible

CyberDem0n:
On PostgreSQL 13 check if restore_command is configured and tell pg_rewind to use it.


FxKu:
:+1:

CyberDem0n:
:+1:

PR #1526, opened 2020-05-06
Shorten and beautify history log output

CyberDem0n:
when Patroni is trying to figure out the necessity of pg_rewind it could write the content history file from the primary into the log. The history file is growing with every failover/switchover and eventually starts taking too many lines in the log, most of them are not so much useful.
Instead of showing the raw data, we will show only 3 lines before the current replica timeline and 2 lines after.

### DIFF
--- a/tests/test_rewind.py
+++ b/tests/test_rewind.py
@@ -82,6 +82,13 @@ class TestRewind(BaseTestPostgresql):
         with patch.object(Postgresql, 'is_running', Mock(return_value=True)):
             self.r.execute(self.leader)
 
+    @patch('patroni.postgresql.rewind.logger.info')
+    def test__log_master_history(self, mock_logger):
+        history = [[n, n, ''] for n in range(1, 10)]
+        self.r._log_master_history(history, 1)
+        expected = '\n'.join(['{0}\t0/{0}\t'.format(n) for n in range(1, 4)] + ['...', '9\t0/9\t'])
+        self.assertEqual(mock_logger.call_args[0][1], expected)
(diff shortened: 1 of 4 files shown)


tests/test_rewind.py
```diff
@@ -98,7 +105,7 @@ def test__check_timeline_and_lsn(self, mock_check_leader_is_not_in_recovery):
         with patch('psycopg2.connect', Mock(side_effect=Exception)):
             self.assertFalse(self.r.rewind_or_reinitialize_needed_and_possible(self.leader))
         self.r.trigger_check_diverged_lsn()
-        with patch.object(MockCursor, 'fetchone', Mock(side_effect=[('', 2, '0/0'), ('', b'3\t0/40159C0\tn\n')])):
+        with patch.object(MockCursor, 'fetchone', Mock(side_effect=[('', 3, '0/0'), ('', b'3\t0/40159C0\tn\n')])):
```
FxKu, line 108:
```suggestion
        with patch.object(MockCursor, 'fetchone', Mock(side_effect=[('', 3, '0/0'), ('', b'4\t0/40159C0\tn\n')])):
```
not sure what the 3 stands for, but as some other numbers are raised, I thought, I'd question this one.

CyberDem0n: This is just a specially crafted mock data, the purpose of it to bring the system into a state that it will go through certain code-paths.

FxKu:
:+1:

CyberDem0n:
:+1:

PR #1527, opened 2020-05-06
Update optime/leader with checkpoint location after clean shut down

CyberDem0n:
Potentially this information could be used in order to make sure that there is no data loss on switchover.


FxKu:
:+1:

CyberDem0n:
:+1:

Issue #1529, opened 2020-05-06
Failure to retry when Consul query failed


caseyallenshobe:
We had a situation where in a development environment, our consul nodes as well as all Patroni nodes were shut down for some time, and restarted all at once.  Most Patroni nodes came back fine on their own, but one ended up dead with a bunch of tracebacks in the log:

```
2020-05-03 01:37:00,570 INFO: Lock owner: None; I am d-gp2-dbpgmgsearch1-1
2020-05-03 01:37:00,594 INFO: Lock owner: None; I am d-gp2-dbpgmgsearch1-1
2020-05-03 01:37:00,605 INFO: starting as a secondary
2020-05-03 01:37:01,513 INFO: postmaster pid=31187
2020-05-03 01:37:01,662 WARNING: Postgresql is not running.
2020-05-03 01:37:01,662 INFO: Lock owner: None; I am d-gp2-dbpgmgsearch1-1
2020-05-03 01:37:01,681 INFO: pg_controldata:
  pg_control version number: 1201
  Catalog version number: 201909212
  Database system identifier: 6776700553226514039
  Database cluster state: in production
  pg_control last modified: Sun May  3 01:07:01 2020
  Latest checkpoint location: C0B/F8771F68
  Latest checkpoint's REDO location: C0B/55890FD0
  Latest checkpoint's REDO WAL file: 0000000300000C0B00000055
  Latest checkpoint's TimeLineID: 3
  Latest checkpoint's PrevTimeLineID: 3
  Latest checkpoint's full_page_writes: on
  Latest checkpoint's NextXID: 0:90035284
  Latest checkpoint's NextOID: 3723327
  Latest checkpoint's NextMultiXactId: 85296716
  Latest checkpoint's NextMultiOffset: 272389864
  Latest checkpoint's oldestXID: 480
  Latest checkpoint's oldestXID's DB: 1
  Latest checkpoint's oldestActiveXID: 90035284
  Latest checkpoint's oldestMultiXid: 1
  Latest checkpoint's oldestMulti's DB: 1
  Latest checkpoint's oldestCommitTsXid: 0
  Latest checkpoint's newestCommitTsXid: 0
  Time of latest checkpoint: Sun May  3 00:40:01 2020
  Fake LSN counter for unlogged rels: 0/3E8
  Minimum recovery ending location: 0/0
  Min recovery ending loc's timeline: 0
  Backup start location: 0/0
  Backup end location: 0/0
  End-of-backup record required: no
[...]

PR #1531, opened 2020-05-07
Properly handle the exception raised from refresh_session

CyberDem0n:
The `touch_member()` could be called from the finally block of the `_run_cycle()`. In case if it raised an exception the whole Patroni process was crashing.
In order to avoid future crashes we wrap `_run_cycle()` into the try..except block and ask a user to report a BUG.

Close https://github.com/zalando/patroni/issues/1529

### DIFF
--- a/patroni/ha.py
+++ b/patroni/ha.py
@@ -1367,8 +1367,12 @@ class Ha(object):
 
     def run_cycle(self):
         with self._async_executor:
-            info = self._run_cycle()
-            return (self.is_paused() and 'PAUSE: ' or '') + info
+            try:
+                info = self._run_cycle()
+                return (self.is_paused() and 'PAUSE: ' or '') + info
+            except Exception:
+                logger.exception('Unexpected exception')
+                return 'Unexpected exception raised, please report it as a BUG'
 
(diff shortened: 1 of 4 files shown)


patroni/ha.py
```diff
@@ -1367,8 +1367,12 @@ def _run_cycle(self):
...
     def run_cycle(self):
         with self._async_executor:
-            info = self._run_cycle()
-            return (self.is_paused() and 'PAUSE: ' or '') + info
+            try:
+                info = self._run_cycle()
+                return (self.is_paused() and 'PAUSE: ' or '') + info
+            except Exception:
+                logger.exception('Unexpected exception')
+                return 'Unexpected exception raised, please report it as a BUG'
```
FxKu, line 1375: We should tell what went wrong, no?
If a user someday files a bug report with `Unexpected exception`, would we immediately know where to look at? :smile:

CyberDem0n: `logger.exception()` producing the stack trace

avaczi:
:+1:

CyberDem0n:
:+1:

Issue #1532, opened 2020-05-08
IPv6 listen address not handled properly


mkowalski:
The following stack trace appears, when trying to use IPv6-only environment, i.e. configuring `listen: '[2a0f:a580:0:420::2]:5432'`

```
patroni              | Traceback (most recent call last):                      
patroni              |   File "patroni.py", line 6, in <module>                                                             
patroni              |     main()                                                     
patroni              |   File "/patroni/__init__.py", line 224, in main                           
patroni              |     return patroni_main()                            
patroni              |   File "/patroni/__init__.py", line 186, in patroni_main
patroni              |     patroni = Patroni(conf)    
patroni              |   File "/patroni/__init__.py", line 36, in __init__
patroni              |     self.api = RestApiServer(self, self.config['restapi'])
patroni              |   File "/patroni/api.py", line 492, in __init__
patroni              |     self.reload_config(config)                                      
patroni              |   File "/patroni/api.py", line 612, in reload_config
patroni              |     self.__initialize(config['listen'], ssl_options)
patroni              |   File "/patroni/api.py", line 580, in __initialize
patroni              |     self.__httpserver_init(host, port)
patroni              |   File "/patroni/api.py", line 554, in __httpserver_init
patroni              |     info = socket.getaddrinfo(host, port, socket.AF_UNSPEC, socket.SOCK_STREAM, 0, socket.AI_PASSIVE)
patroni              |   File "/usr/lib/python3.7/socket.py", line 748, in getaddrinfo
patroni              |     for res in _socket.getaddrinfo(host, port, family, type, proto, flags):
patroni              | socket.gaierror: [Errno -2] Name or service not known
```

After quick debugging it seems that the function initializing the server has a flaw... In https://github.com/zalando/patroni/blob/master/patroni/api.py#L580 we are passing a result of `split_host_port(listen, None)` however

```
>>> split_host_port('[2a0f:a580:0:420::2]:5432',5432)
('[2a0f:a580:0:420::2]', 5432)
```
(note the brackets)

If we try to call
```
socket.getaddrinfo(host, port, socket.AF_UNSPEC, socket.SOCK_STREAM, 0, socket.AI_PASSIVE)
```
with `host` containing the brackets, the call will fail.

A solution could be to add an additional logic to the `split_host_port` function so that returned tuple always contains host in correct format, i.e. with no brackets. Given this could have a potential impact on other parts of the codebase, at the moment I'm not 100% sure how safe that is.

[...]

PR #1533, opened 2020-05-08
Handle IPv6 addresses in split_host_port

mkowalski:
This PR makes split_host_port return IPv6 address without enclosing
brackets. This is due to the fact that e.g. socket.* functions expect
host not to contain them when being called with IPv6.

Close: #1532

### DIFF
--- a/patroni/utils.py
+++ b/patroni/utils.py
@@ -357,6 +357,8 @@ def polling_loop(timeout, interval=1):
 
 def split_host_port(value, default_port):
     t = value.rsplit(':', 1)
+    if ':' in t[0]:
+        t[0] = t[0].strip('[]')
     t.append(default_port)
     return t[0], int(t[1])
 


mkowalski:
I have briefly checked in how many places `split_host_port` is being used and given that the change impacts more than 1 usage, another option would be to add
```
if host.startswith('['):
  host = host.strip('[]')
```
just above the https://github.com/zalando/patroni/blob/master/patroni/api.py#L580

As I'm not sure what's preferred by maintainers, let me know if you'd prefer the other approach.

patroni/utils.py
```diff
@@ -357,6 +357,8 @@ def polling_loop(timeout, interval=1):
 
 def split_host_port(value, default_port):
     t = value.rsplit(':', 1)
+    if ':' in t[0]:
```
Jan-M, line 360: Whats the value of this `if`

Jan-M:
Would it make sense to check this already early on in config, so not all places potentially need to deal with the `[]`

patroni/utils.py
CyberDem0n: It is about IPv6 format, for example `::1`.

CyberDem0n:
@Jan-M  All places already calling `split_host_port()` function.

Jan-M:
:+1:

CyberDem0n:
:+1:

CyberDem0n:
Thank you @mkowalski!

PR #1542, opened 2020-05-12
Don't try calling a non existing leader in patronictl pause

CyberDem0n:
While pausing a cluster without a leader on K8s patronictl was showing warnings that member "None" could not be accessed.


FxKu:
:+1:

CyberDem0n:
:+1:

Issue #1545, opened 2020-05-13
REST API is trivial to accidentally DoS when TLS enabled


bjh21:
If I make a long-lived TCP connection to Patroni's REST API port, this causes any other attempts at connecting to it to time out until I close that connection. So if in one window I run:
```
# nc localhost 8008
```
then in another:
```
# curl --insecure https://localhost:8008/
curl: (7) Failed to connect to localhost port 8008: Connection timed out
```
This means that it's trivial to accidentally break the API endpoint on a Patroni node simply by opening a TCP connection to it and then forgetting to do anything. I discovered this because a nearby OpenVAS connected to my API port and thus broke my monitoring.

I think the REST API endpoint should either be able to cope with a reasonable number of simultaneous connections or should time out idle connections quickly. Alternatively, a documented way to get the cluster state from etcd might avoid some need for the REST API.

I've observed this problem on version 1.6.5-2.pgdg18.04+1 of the PGDG package of Patroni for Ubuntu, but I have no reason to think it's specific to the packaged version.

CyberDem0n:
> So if in one window I run:
>  nc localhost 8008
> then in another:
>  curl --insecure https://localhost:8008/
> curl: (7) Failed to connect to localhost port 8008: Connection timed out

I can't reproduce it. Every incoming request is processed by a new thread, they are not waiting until the previous disconnects.

bjh21:
That's very curious.  I shall investigate further.

bjh21:
It appears that the problem only occurs when:
* I have TLS enabled on the API server, and
* The incoming connection stops before completing the TLS handshake.

So the problem doesn't occur if TLS is disabled (as it presumably was in your test).  It also doesn't occur if I connect using `openssl s_client` in place of `nc`: once the TLS handshake is complete, the new thread is properly created.

I'm sorry I didn't explicitly mention the use of TLS in original report.

PR #1546, opened 2020-05-14
Switch API socket to TLS after connection

bjh21:
The API endpoint is a threaded HTTP server, so it creates a new thread to handle each request.  However, because it wraps its listening socket in SSL, the `ssl` module insists on completing the TLS handshake before it returns from `accept()`, so the entire TLS handshake happens in the main thread.  This means that if the TLS handshake stalls for any reason, the API endpoint becomes unresponsive (issue #1545).

This commit moves the `wrap_socket()` call so that it happens only after each connection has been accepted, meaning that it happens in the per-connection thread and can't affect the main one.

My understanding of Python's threading is limited, so I'm not sure about the thread-safety of my changes. I suspect there may be a problem when reconfiguring to turn on TLS: `RestApiServer.__initialize` updates `__protocol` before configuring TLS, so if a connection thread calls `setup` in that gap, it might try to use `__ctx` before it exists. If this is a problem, it should be fairly easy to fix.

### DIFF
--- a/patroni/api.py
+++ b/patroni/api.py
@@ -395,7 +395,7 @@ class RestApiHandler(BaseHTTPRequestHandler):
 
     def setup(self):
         self.request = self.server.wrap_socket(self.request)
-        super().setup()
+        super(RestApiHandler, self).setup()
 
     def parse_request(self):
         """Override parse_request method to enrich basic functionality of `BaseHTTPRequestHandler` class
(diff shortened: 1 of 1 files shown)


CyberDem0n:
@bjh21 please take a look: https://github.com/zalando/patroni/pull/1547

bjh21:
Excellent!  I can stop discovering how much of Python 2 I've forgotten.

Issue #1127, opened 2019-07-31
Patroni API becomes unresponsive (timeout) for two clusters at the same time


jpuris:
**Some details on environment**
- Patroni version 1.5.6
- Patroni uses 5 node etcd cluster

**Problem**
We are running multiple clusters and API requests started to timeout on all nodes across 2 clusters seperate. All clusters are using the same etcd cluster.

This is happening for the third time now and happens roughly at 1-3 weeks interval.

**Recovery**
Restart patroni service on all nodes in all effected clusters.

**etcd state during incident**
All seems to be in order and all monitored metrics are normal. Logs are also clean.

**Whats working during incident**
- patronictl list
- postgresql nodes and replication
- patroni logs show absolutely no problems

**Whats not working during incident**
- patroni API is unresponsive
- patronictl switchover command times out and needs to be cancelled
```
Are you sure you want to switchover cluster my_cluster, demoting current master my_master_node? [y/N]: y
^C
Aborted!
```

One of the affected clusters had been recovered with the described recovery method, however the other (dev) cluster patroni API's are still in limbo.

Patroni config file from one of the nodes.
```yml
scope: <redacted>
name: node-1

watchdog:
    mode: off

restapi:
    listen: 0.0.0.0:8008
    connect_address: <redacted>:8008
    certfile: /etc/postgresql/ssl/self-signed-node-1.pem
    keyfile: /etc/postgresql/ssl/self-signed-node-1-key.pem
    cacert: /etc/postgresql/ssl/self-signed-ca.pem

etcd:
    hosts: <redacted>:2379,<redacted>:2379,<redacted>:2379,<redacted>:2379,<redacted>:2379
    protocol: https
    cacert: /etc/postgresql/ssl/self-signed-ca.pem
    cert: /etc/postgresql/ssl/self-signed-node-1.pem
    key: /etc/postgresql/ssl/self-signed-node-1-key.pem

bootstrap:
    dcs:
        ttl: 43200
        loop_wait: 10
        retry_timeout: 10
        maximum_lag_on_failover: 1048576
        postgresql:
            use_pg_rewind: true

    pg_ident:
    - auth_map postgres postgres
    - auth_map <redacted> <redacted>

    pg_hba:
    # LOCAL
    - local all all ident map=auth_map
    # REPLICATION
    - hostssl replication repl_user 127.0.0.1/32 md5
    - hostssl replication repl_user <redacted>/32 md5
    - hostssl replication repl_user <redacted>/32 md5
[...]

Issue #857, opened 2018-11-08
Connection reset by peer


wilfriedroset:
Hi,

We have encountered today a similar bug to #393. Patroni API stop responding for approximately 10min, hence no probes and therefore no more queries. All traffic flow restart just after the time of the following traceback:

```
Nov 08 15:50:17 patroni-node-1 patroni[12390]: Traceback (most recent call last):
Nov 08 15:50:17 patroni-node-1 patroni[12390]:   File "/usr/lib/python3.5/socketserver.py", line 625, in process_request_thread
Nov 08 15:50:17 patroni-node-1 patroni[12390]:     self.finish_request(request, client_address)
Nov 08 15:50:17 patroni-node-1 patroni[12390]:   File "/usr/lib/python3.5/socketserver.py", line 354, in finish_request
Nov 08 15:50:17 patroni-node-1 patroni[12390]:     self.RequestHandlerClass(request, client_address, self)
Nov 08 15:50:17 patroni-node-1 patroni[12390]:   File "/usr/lib/python3.5/socketserver.py", line 681, in __init__
Nov 08 15:50:17 patroni-node-1 patroni[12390]:     self.handle()
Nov 08 15:50:17 patroni-node-1 patroni[12390]:   File "/usr/lib/python3.5/http/server.py", line 422, in handle
Nov 08 15:50:17 patroni-node-1 patroni[12390]:     self.handle_one_request()
Nov 08 15:50:17 patroni-node-1 patroni[12390]:   File "/usr/lib/python3.5/http/server.py", line 400, in handle_one_request
Nov 08 15:50:17 patroni-node-1 patroni[12390]:     if not self.parse_request():
Nov 08 15:50:17 patroni-node-1 patroni[12390]:   File "/usr/lib/python3/dist-packages/patroni/api.py", line 389, in parse_request
Nov 08 15:50:17 patroni-node-1 patroni[12390]:     ret = BaseHTTPRequestHandler.parse_request(self)
Nov 08 15:50:17 patroni-node-1 patroni[12390]:   File "/usr/lib/python3.5/http/server.py", line 327, in parse_request
Nov 08 15:50:17 patroni-node-1 patroni[12390]:     "Bad request syntax (%r)" % requestline)
Nov 08 15:50:17 patroni-node-1 patroni[12390]:   File "/usr/lib/python3.5/http/server.py", line 477, in send_error
Nov 08 15:50:17 patroni-node-1 patroni[12390]:     self.wfile.write(body)
[...]

PR #1547, opened 2020-05-14
Defer TLS handshake until thread has started

CyberDem0n:
The `SSLSocket` is immediately doing the handshake on accept. Effectively it blocks the whole API thread if the client-side doesn't send any data.
In order to solve the issue we defer the handshake until a thread serving request has started.

The solution is a bit hacky, but thread-safe.

Close https://github.com/zalando/patroni/issues/1545

### DIFF
--- a/patroni/api.py
+++ b/patroni/api.py
@@ -599,6 +599,23 @@ class RestApiServer(ThreadingMixIn, HTTPServer, Thread):
         if reloading_config:
             self.start()
 
+    def process_request_thread(self, request, client_address):
+        if isinstance(request, tuple):
+            sock, newsock = request
+            try:
+                request = sock.context.wrap_socket(newsock, do_handshake_on_connect=sock.do_handshake_on_connect,
+                                                   suppress_ragged_eofs=sock.suppress_ragged_eofs, server_side=True)
+            except socket.error:
+                return
+        super(RestApiServer, self).process_request_thread(request, client_address)
+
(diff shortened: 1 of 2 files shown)


CyberDem0n:
I have a feeling that it should solve https://github.com/zalando/patroni/issues/1127 and https://github.com/zalando/patroni/issues/857

bjh21:
I've applied this patch on my test system, and it seems to work in minimal testing.

bjh21:
Slightly more testing: I scanned my test cluster with OpenVAS last night and the monitoring worked correctly throughout.

CyberDem0n:
@wilfriedroset, @jpuris, @psuderevsky, @jouir please take a look, highly likely it will solve problems of API freezes when TLS is configured. It would be great if you put it to the test.

psuderevsky:
@CyberDem0n Hi. I encountered this issue only once in a year and actual steps to reproduce are unknown, so it seems that only time will tell, unfortunately.

KostyaEsmukov:
I had 3 freezes in last 2 months on a cluster with TLS enabled so I'm really looking forward for #1127/#857 to be solved.

I have just applied this patch on the problematic cluster. Will let you know if the issue persists.

FxKu:
:+1:

CyberDem0n:
:+1:

CyberDem0n:
:+1:

PR #1552, opened 2020-05-18
Don't fire on_reload when promoting to standby_leader on 13+

CyberDem0n:
PostgreSQL 13 finally introduced the possibility to change primary_conninfo without a restart. Just doing reload is enough, but in case if the role is changing from the replica to the standby_leader we want to call only on_role_change callback and skip on_reload, because they duplicate each other.


erthalion:
:+1:

CyberDem0n:
:+1:

Issue #1544, opened 2020-05-12
Patroni does not restore hba_file with custom bootstrap


pursultani:
With custom bootstrap method, when `hba_file` is explicitly set to `${postgresql.data_dir}/pg_hba.conf`, the [logic for the temporary pg_hba](https://github.com/zalando/patroni/blob/v1.6.5/patroni/postgresql/config.py#L458) overrides it and after bootstrapping it goes missing. 

```yaml
postgresql:
  bin_dir: /opt/gitlab/embedded/bin
  data_dir: /var/opt/gitlab/postgresql/data
  config_dir: /var/opt/gitlab/postgresql/data
  listen: 0.0.0.0:5432
  connect_address: 192.168.64.15:5432
  use_unix_socket: true
  parameters:
    unix_socket_directories: /var/opt/gitlab/postgresql
    ident_file: /var/opt/gitlab/postgresql/data/pg_ident.conf
    hba_file: /var/opt/gitlab/postgresql/data/pg_hba.conf
  ...
bootstrap:
  method: gitlab_bootstrap
  gitlab_bootstrap:
    command: /var/opt/gitlab/patroni/bootstrap
  post_bootstrap: /var/opt/gitlab/patroni/post-bootstrap
```

```
2020-05-09_21:40:00.20427 2020-05-10 07:40:00,202 INFO: running post_bootstrap
....
2020-05-09_21:40:00.64002 LOG:  received SIGHUP, reloading configuration files
2020-05-09_21:40:00.64162 LOG:  could not open configuration file "/var/opt/gitlab/postgresql/data/pg_hba.conf": No such file or directory
2020-05-09_21:40:00.64193 LOG:  pg_hba.conf was not reloaded
2020-05-09_21:40:00.64220 server signaled
```

This is of course redundant but one wouldn't expect it to happen.

CyberDem0n:
If `hba_file` and `ident_file` are specified it more or less mean that you take responsibility to manage them.
I never thought about the case when someone will set them to the same files in the PGDATA.

pursultani:
> If hba_file and ident_file are specified it more or less mean that you take responsibility to manage them.

That is correct, but in this particular case Patroni interferes with `pg_hba.conf` even though temporarily. The interesting thing is that if I remove `hba_file` from the config above, everything works fine. If I am not mistaken, those parameters are [set by default](https://github.com/zalando/patroni/blob/v1.6.5/patroni/postgresql/config.py#L812).

[...]

PR #1555, opened 2020-05-18
Ignore hba_file and ident_file when they match with defaults

CyberDem0n:
It is possible to specify custom hba_file and ident_file in the postgresql configuration parameters and Patroni is considering that these files are managed externally. It could happen that locations of these files matching with default locations of pg_hba,conf and pg_ident.conf. In this case we will ignore custom values and fallback to the default workflow, i.e. Patroni will overwrite them.

Close: https://github.com/zalando/patroni/issues/1544


FxKu:
:+1:

CyberDem0n:
:+1:

PR #1559, opened 2020-05-25
Enforce synchronous_commit=local for post_init script

CyberDem0n:
Patroni was already doing that before creating users for a long time, but the post_init was an oversight. It will help to all utilities relying on libpq and reduce the end-user confusion.


FxKu:
:+1:

CyberDem0n:
:+1:

PR #1561, opened 2020-05-27
Try to fetch missing WAL if pg_rewind complains about it

CyberDem0n:
It could happen that the WAL segment required for pg_rewind doesn't exist in the pg_wal anymore and therefore pg_rewind can't find the checkpoint location before the diverging point.
Starting from PostgreSQL 13 pg_rewind could use restore_command for fetching missing WALs, but we can do better than that.
On older PostgreSQL versions Patroni will parse the stdout and stderr of failed rewind, try to fetch the missing WAL, and repeat the pg_rewind attempt.


Jan-M:
:+1:

CyberDem0n:
:+1:

Issue #1493, opened 2020-04-15
switchover leads to rewind/reclone in third node if remove_data_directory_on_diverged_timelines is true


mbanck:
If I do a manual switchover, the third node (i.e. neither the old nor the new leader) gets rewound/recloned (depending on whether `use_pg_rewind` is set or not) if `remove_data_directory_on_diverged_timelines` is set to true. If the latter is not the case, the third node just closes the connection and happily follows the new one (I can post logs for that, if needed). If 
`remove_data_directory_on_diverged_timelines` is set to true, I get (with some additional debugging `info` added):

```
Apr 15 12:15:16 pg2 patroni[11343]: 2020-04-15 12:15:16,011 INFO: Lock owner: pg3; I am pg2
Apr 15 12:15:16 pg2 patroni[11343]: 2020-04-15 12:15:16,011 INFO: does not have lock
Apr 15 12:15:16 pg2 patroni[11343]: 2020-04-15 12:15:16,017 INFO: no action.  i am a secondary and i am following a leader
Apr 15 12:15:16 pg2 patroni[11343]: 2020-04-15 12:15:16,346 INFO: Got response from pg1 http://192.168.122.163:8008/patroni: {"state": "running", "postmaster_start_time": "2020-04-15 12:13:11.090 UTC", "role": "
Apr 15 12:15:16 pg2 patroni[11343]: 2020-04-15 12:15:16,347 INFO: manual failover: to pg1, i am pg2
Apr 15 12:15:18 pg2 patroni[11343]: 2020-04-15 12:15:18,369 INFO: Local timeline=5 lsn=0/C000178
Apr 15 12:15:18 pg2 patroni[11343]: 2020-04-15 12:15:18,379 INFO: master_timeline=6
Apr 15 12:15:18 pg2 patroni[11343]: 2020-04-15 12:15:18,379 INFO: master: history=1        0/4000178        no recovery target specified
Apr 15 12:15:18 pg2 patroni[11343]: 2        0/6000178        no recovery target specified
Apr 15 12:15:18 pg2 patroni[11343]: 3        0/8000258        no recovery target specified
Apr 15 12:15:18 pg2 patroni[11343]: 4        0/A000258        no recovery target specified
Apr 15 12:15:18 pg2 patroni[11343]: 5        0/C000178        no recovery target specified
Apr 15 12:15:18 pg2 patroni[11343]: 2020-04-15 12:15:18,379 INFO: need_rewind: None
[...]

PR #1563, opened 2020-05-27
Improve timeline divergence check

CyberDem0n:
We don't need to rewind when:
1. replayed location for the former replica is not ahead of switchpoint
2. end of checkpoint record for the former primary is the same as switchpoint

In order to get the end of checkpoint record we use the `pg_waldump` and parse its output.

Close https://github.com/zalando/patroni/issues/1493


erthalion:
:+1:

erthalion:
Looks a bit scary though with pg_waldump.

CyberDem0n:
:+1:

PR #1573, opened 2020-06-03
Bugfix: GET /cluster was showing stale member info in zookeeper

CyberDem0n:
Zookpeeper implementation heavily relies on cached version of the cluster view in order to minimize the number of requests. Having stale members information is fine for Patroni workflow because it basically relies only on member names and tags.

The `GET /cluster` is a different case. Being exposed outside it might be used for monitoring purposes and therefore we should show the up-to-date members information.


FxKu:
:+1:

CyberDem0n:
:+1:

PR #1576, opened 2020-06-04
patronictl add ability to print ASCII topology

MaxFedotov:
Hi guys,

As we started using Patroni, we find out that it's missing the ability to print current topology of the cluster, which is very important info if you use cascading replicas and want to understand how cluster looks like currently.

This PR adds `topology` command to patronictl, which is able to print ASCII topology of specified cluster like this
```bash
[root@postgresql1 patroni]# patronictl -c /etc/patroni/patroni.yaml topology
Cluster: patroni14_test_mf_dba (6826270184262086410)
postgresql1    	|10.144.173.75:5432|leader |running|tl:20|
+ postgresql2 	|10.144.173.73:5432|replica|running|tl:20|lag:0.0 MB|
  + postgresql3	|10.144.173.74:5432|replica|running|tl:20|lag:0.0 MB|
+ postgresql4 	|10.144.173.76:5432|replica|running|tl:20|lag:0.0 MB|
```

### DIFF
--- a/patroni/ctl.py
+++ b/patroni/ctl.py
@@ -170,7 +170,7 @@ def print_output(columns, rows, alignment=None, fmt='pretty', header=None, delim
         elements = [{k: v for k, v in zip(columns, r) if not header or str(v)} for r in rows]
         func = json.dumps if fmt == 'json' else format_config_for_editing
         click.echo(func(elements))
-    elif fmt in {'pretty', 'tsv'}:
+    elif fmt in {'pretty', 'tsv', 'topology'}:
         list_cluster = bool(header and columns and columns[0] == 'Cluster')
         if list_cluster and 'Tags' in columns:  # we want to format member tags as YAML
             i = columns.index('Tags')
@@ -187,11 +187,20 @@ def print_output(columns, rows, alignment=None, fmt='pretty', header=None, delim
         else:
             hrules = ALL if any(any(isinstance(c, six.string_types) and '\n' in c for c in r) for r in rows) else FRAME
             table = PatronictlPrettyTable(header, columns, hrules=hrules)
+            if fmt == 'topology':
+                table.header = False
+                table.horizontal_char = table.junction_char = table.vertical_char = " "
+                table.vrules = table.hrules = table.padding_width = 0
+            else:
+                table = PatronictlPrettyTable(header, columns, hrules=hrules)
             for k, v in (alignment or {}).items():
                 table.align[k] = v
             for r in rows:
                 table.add_row(r)
-            click.echo(table)
+            if fmt == 'topology':
+                click.echo(table.__str__().strip().replace(" |", "|"))
+            else:
+                click.echo(table)
 
 
 def watching(w, watch, max_count=None, clear=True):
@@ -746,19 +755,10 @@ def switchover(obj, cluster_name, master, candidate, force, scheduled):
     _do_failover_or_switchover(obj, 'switchover', cluster_name, master, candidate, force, scheduled)
 
 
-def output_members(cluster, name, extended=False, fmt='pretty'):
-    rows = []
-    logging.debug(cluster)
-    initialize = {None: 'uninitialized', '': 'initializing'}.get(cluster.initialize, cluster.initialize)
+def _prepare_cluster_info(cluster, name):
     cluster = cluster_as_json(cluster)
-
-    columns = ['Cluster', 'Member', 'Host', 'Role', 'State', 'TL', 'Lag in MB']
-    for c in ('Pending restart', 'Scheduled restart', 'Tags'):
-        if extended or any(m.get(c.lower().replace(' ', '_')) for m in cluster['members']):
-            columns.append(c)
-
-    # Show Host as 'host:port' if somebody is running on non-standard port or two nodes are running on the same host
     members = [m for m in cluster['members'] if 'host' in m]
+    # Show Host as 'host:port' if somebody is running on non-standard port or two nodes are running on the same host
     append_port = any('port' in m and m['port'] != 5432 for m in members) or\
         len(set(m['host'] for m in members)) < len(members)
 
@@ -767,7 +767,7 @@ def output_members(cluster, name, extended=False, fmt='pretty'):
 
         lag = m.get('lag', '')
         m.update(cluster=name, member=m['name'], host=m.get('host', ''), tl=m.get('timeline', ''),
-                 role='' if m['role'] == 'replica' else m['role'].replace('_', ' ').title(),
+                 role=m['role'].replace('_', ' ').title(),
                  lag_in_mb=round(lag/1024/1024) if isinstance(lag, six.integer_types) else lag,
                  pending_restart='*' if m.get('pending_restart') else '')
 
@@ -779,7 +779,21 @@ def output_members(cluster, name, extended=False, fmt='pretty'):
             if 'postgres_version' in m['scheduled_restart']:
                 value += ' if version < {0}'.format(m['scheduled_restart']['postgres_version'])
             m['scheduled_restart'] = value
+    return cluster
 
+
+def output_members(cluster, name, extended=False, fmt='pretty'):
+    rows = []
+    logging.debug(cluster)
+    initialize = {None: 'uninitialized', '': 'initializing'}.get(cluster.initialize, cluster.initialize)
+    cluster = _prepare_cluster_info(cluster, name)
+
+    columns = ['Cluster', 'Member', 'Host', 'Role', 'State', 'TL', 'Lag in MB']
+    for c in ('Pending restart', 'Scheduled restart', 'Tags'):
+        if extended or any(m.get(c.lower().replace(' ', '_')) for m in cluster['members']):
+            columns.append(c)
+
+    for m in cluster['members']:
         rows.append([m.get(n.lower().replace(' ', '_'), '') for n in columns])
 
     print_output(columns, rows, {'Lag in MB': 'r', 'TL': 'r', 'Tags': 'l'},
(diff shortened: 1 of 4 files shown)


patroni/ctl.py
```diff
@@ -828,6 +828,80 @@ def members(obj, cluster_names, fmt, watch, w, extended, ts):
             cluster = dcs.get_cluster()
             output_members(cluster, cluster_name, extended, fmt)
 
+def output_topology(cluster, name):
```
CyberDem0n, line 832: Why not to use existing function [print_output()](https://github.com/zalando/patroni/blob/6a505b837da98f275dfb10e5cf36a914f3a54e6b/patroni/ctl.py#L168)?
It would save from from quite a lot of ident calculations.

patroni/ha.py
```diff
@@ -175,12 +175,14 @@ def get_effective_tags(self):
...
         with self._member_state_lock:
+            primary_conninfo = self.state_handler.config._current_recovery_params.get('primary_conninfo', {})[0] if self.state_handler.config._current_recovery_params is not None and self.state_handler.role != 'master' else None
             data = {
                 'conn_url': self.state_handler.connection_string,
                 'api_url': self.patroni.api.connection_string,
                 'state': self.state_handler.state,
                 'role': self.state_handler.role,
-                'version': self.patroni.version
+                'version': self.patroni.version,
+                'master_host': primary_conninfo.get('host', None) if primary_conninfo is not None else None
```
CyberDem0n, line 187: Thyamlere is no need to add a new field to the structure, [tags](https://github.com/zalando/patroni/pull/1576/files#diff-57c53e988b775ee2902488fb33c1f433R193-R195) already has everything you need, i.e.
```json
{
  "tags": {
    "replicatefrom": "foo"
  }
}
```

If there is a member `foo` in the cluster, the given member will be replicating from it (or at least try to do so).

MaxFedotov: Yes, but tags are optional and the can be absent. So it this case you won't be able to build topology. Or I am not correct?

CyberDem0n: It is not possible to configure cascading replication without using tags.

patroni/ctl.py
MaxFedotov: These ident calculations are the source of this PR :) As they provide pretty and clear ASCII printing and view of cluster topology. I can move them to `print_output()` function and add some type of custom format in order to support this pretty printing. Will it be OK?

MaxFedotov: Ok, will take a look on it and fix it. Thanks!

patroni/ctl.py
```diff
@@ -828,6 +828,80 @@ def members(obj, cluster_names, fmt, watch, w, extended, ts):
...
+    filler = ' '
+    initialize = {None: 'uninitialized', '': 'initializing'}.get(cluster.initialize, cluster.initialize)
+    cluster = cluster_as_json(cluster)
+    members = {member['host']:member for member in [m for m in cluster['members'] if 'host' in m]}
+    for _, data in members.items():
+        if data['master_host'] is None:
+            masters.append(data)
+        else:
+            if data['master_host'] not in topology:
+                topology[data['master_host']] = []
```
drednout, line 845: `topology.setdefault(data['master_host'], [])` construction may be used as simplified version of 

`
if data['master_host'] not in topology:
     topology[data['master_host']] = []
`

patroni/ha.py
CyberDem0n: Effectively the field you have added is just duplicating the `tags.replicatefrom`.
Absence of such tag means that the node is replicating from the leader.
When I told `it is trying to do so` I only meant to say that this node could be not in a good shape, for example postgres is failing to start. The node which holds the `replicatefrom` tag will stream from the primary only in one case, if the cascading node is not registered in DCS, therefore it is always safe to rely on tags.

If in the future there will be the other way of configuring cascading replication we will adjust the code which prints the topology accordingly, this is not a big deal, but now it is not necessary.

MaxFedotov: I do not think it's duplicating. `replicatefrom` - is desired state, but this field is the current state.
If we will switch on using tags, we should also switch on using `role` field in dcs in order to determine who is current leader.

But let's assume the situation, where the current leader lost connection to dcs.
In this case there will be some time, when patroni will not report a leader, like 
```bash
+ Cluster: patroni14_test_mf_dba (6826270184262086410) --------+----+-----------+--------------------------------------------+
|            Member           |      Host     | Role |  State  | TL | Lag in MB | Tags                                       |
+-----------------------------+---------------+------+---------+----+-----------+--------------------------------------------+
| postgresql3 | 10.144.173.73 |      | running | 23 |       0.0 | replicatefrom: :undef                      |
+-----------------------------+---------------+------+---------+----+-----------+--------------------------------------------+
| postgresql4 | 10.144.173.74 |      | running | 23 |       0.0 | clonefrom: true                            |
|                             |               |      |         |    |           | replicatefrom: postgresql3 |
+-----------------------------+---------------+------+---------+----+-----------+--------------------------------------------+
```
but in actual topology, this cluster will look like 
```bash
Cluster: patroni14_test_mf_dba (6826270184262086410)
postgresql3  	|10.144.173.73:5432|replica|running|tl:23|
+ postgresql4	|10.144.173.74:5432|replica|running|tl:23|lag:0.0 MB|
```
so it's obviously has a master, although it might be in read-only mode.
and in case, if there was no cascade and tags are empty, we will see situation from the side of postgres like this (two standalone not replicated servers)
```bash
Cluster: patroni14_test_mf_dba (6826270184262086410)
postgresql3  	|10.144.173.73:5432|replica|running|tl:23|
postgresql4	|10.144.173.74:5432|replica|running|tl:23|lag:0.0 MB|
```

But In dcs they will both have `role:replica`.

if we will rely on `role` field in order to determine leader - it will make the logic of determining current master very complicated

CyberDem0n: > so it's obviously has a master, although it might be in read-only mode.

Nope. The this topology example is incorrect. Absence of leader doesn't make any other node a new leader automatically., therefore there should be a `+ ` before postgresql3.

> if we will rely on role field in order to determine leader - it will make the logic of determining current master

That's the point. Only the node owning the leader key in the DCS is the true leader and you should never rely on the `role` field for that.

> I do not think it's duplicating. replicatefrom - is desired state, but this field is the current state.

This is sort of correct, but you can always derive one from another. If there are any minor discrepancies between these two states, they will go away in `loop_wait` seconds maximum.

After all, the member key in DCS is updated only once per HA loop, therefore it is not possible to avoid minor discrepancies.

MaxFedotov: I think we are mixing point of views :)
You are speaking from the patroni side, how it sees the cluster, while I am from the postgresql side.
> Nope. The this topology example is incorrect. Absence of leader doesn't make any other node a new leader automatically., therefore there should be a `+ ` before postgresql3.

It's taken from the real cluster, where I've emulated connection issue between patroni on leader and dcs with large enough retry_timeout. While leader is retrying and postgresql is alive on it, during retry timeout there will be a situation, when leader key will be missing from dcs. So from the patroni side, this cluster do not have a leader. But, if we look into postgres, there obviously will be one master (`postgresql3`) and replica (`postgresql4`), as it's shown on topology. 

Moreover, the absence of leader key in such situations makes building topology graph very complicated, as there will be no vertex, from which we can start traversing it.

I understand that scenario like this with dcs failures are rare enough, but I think that presenting a true topology in such cases is very important in order to make decisions on real state of things.

CyberDem0n: > You are speaking from the patroni side, how it sees the cluster, while I am from the postgresql side.

There is no point of view from the postgres side, the DCS is the only source of trust!

> It's taken from the real cluster, where I've emulated connection issue between patroni on leader and dcs with large enough retry_timeout.

So you violated the requirement to set `ttl <= loop_wait + 2 * retry_timeout `?
The master should be demoted in the worst case when the leader key expires, or better a bit earlier.

> So from the patroni side, this cluster do not have a leader. But, if we look into postgres, there obviously will be one master (postgresql3) and replica (postgresql4), as it's shown on topology.

Nope. The postgresql3 is the replica, and currently there is absolutely no guaranty that it will become the new leader.

> Moreover, the absence of leader key in such situations makes building topology graph very complicated, as there will be no vertex, from which we can start traversing it.

Just replace the `leader` with `None`. Problem solved.

MaxFedotov: Yes, I understand your concerns and that this situation is a bit fictional and that is a misconfiguration, which I made for a purpose.
But let's think a bit further - as `replicatefrom` is user-configurable, there is always an opportunity of mistakes or typos. And the data I'm using for `master_host` is just the same methods as Patroni uses in order to fill `recovery.conf`. So if, for example, user will make a mistake in configuration, he will be able to see real topology and fix it

CyberDem0n: > replicatefrom is user-configurable, there is always an opportunity of mistakes or typos. And the data I'm using for master_host is just the same methods as Patroni uses in order to fill recovery.conf. So if, for example, user will make a mistake in configuration, he will be able to see real topology and fix it.

In case of any typos or missing intermediate node it will be replicating from the leader:
https://github.com/zalando/patroni/blob/6406b39b77269252fc0b0149ca4472d0bab0f06c/patroni/ha.py#L366-L367
https://github.com/zalando/patroni/blob/6406b39b77269252fc0b0149ca4472d0bab0f06c/patroni/dcs/__init__.py#L445-L446

And I don't see any reason to not use the same logic in patronictl.

> may be it will be a good idea to add tags to output of topology, so such misconfigurations/typos be visible to user?

Probably, but currently tags are formatted to multiple lines, so drawing ASCII tree will become harder, unless we apply special formatting to tags (default_flow_style=True)

MaxFedotov: I understand your concerns, that adding to structure in DCS a new field, that will be used only in one function can look like something wasteful, so will try to derive it from available info. 

But I got some question regarding finding who is a leader now.
https://github.com/zalando/patroni/blob/e95e54b94e08a9bdf624ba6cd126c26c6c95ff94/patroni/ctl.py#L765-L772

When we are using `list` command, we derive leader from member role. 
But you said that 

> That's the point. Only the node owning the leader key in the DCS is the true leader and you should never rely on the role field for that.

So what is the right way to find out who is a leader?

And one more question, why do patronictl set `role='' if m['role'] == 'replica'`?

CyberDem0n: The "real" role is calculated here: https://github.com/zalando/patroni/blob/798c46bc0376f9b401bc25e45ef3be9745cc58f8/patroni/utils.py#L381-L393

And it guarantees that there is not more than one `leader` or `standby_leader` among members.

CyberDem0n: > And one more question, why do patronictl set role='' if m['role'] == 'replica'?

Because it was always shown as an empty string: https://github.com/zalando/patroni/blob/dcb5113f9d937db498847cea2f0ad92b301d43cc/patroni/ctl.py#L328-L346

CyberDem0n: Another reason to not introduce a new field - the new patronictl will work with old clusters.
It is quite common that people run patronictl remotely, not from the host where Patroin is running.

MaxFedotov: > Because it was always shown as an empty string:
Can I fix it and return `replica` and not an empty string? I just want to reuse the same code, that is used in `output_members`?

CyberDem0n: I don't have any objections on that.

MaxFedotov: done, now use `replicatefrom` in order to find leader

patroni/ctl.py
MaxFedotov: fixed

patroni/ctl.py
MaxFedotov: done

MaxFedotov:
done, now use `print_output()` for printing topology, `replicatefrom` tag to find a leader and added tests.

CyberDem0n:
@MaxFedotov 
What about having something like following?
```
+ Cluster: batman (6834835313225022118) -----+---------+----+-----------+------------------------------------------------+
| Member          |      Host      |   Role  |  State  | TL | Lag in MB | Tags                                           |
+-----------------+----------------+---------+---------+----+-----------+------------------------------------------------+
| postgresql0     | localhost:5432 |  Leader | running |  2 |           |                                                |
| + postgresql1   | localhost:5433 | Replica | running |  2 |       0.0 |                                                |
|   + postgresql2 | localhost:5434 | Replica | running |  2 |       0.0 | {nofailover: true, replicatefrom: postgresql1} |
+-----------------+----------------+---------+---------+----+-----------+------------------------------------------------+
```

So it will look more or less the same as any other ASCII table produced by patronictl?
It would also simplify quite a few things and the only you'll have to do - reorder members and prepend some of member names with ` + ` prefixes.

CyberDem0n:
@MaxFedotov something like https://gist.github.com/CyberDem0n/d4932760a3ff71e75ce582378a528795

MaxFedotov:
Yeap, that looks much better, applied review notes, thank you!
Btw, maybe do not create additional `topology` command, but change default look of `list`, so it prints topology sorted output by default?

CyberDem0n:
> Btw, maybe do not create additional topology command, but change default look of list, so it prints topology sorted output by default?

I think it would be better to leave defaults as it is, to not confuse people.

Please also add the unit-test calling the `topology()` function.

MaxFedotov:
added tests for topology

Jan-M:
:+1:

CyberDem0n:
:+1:

PR #1586, opened 2020-06-11
Get rid of kubernetes python module

CyberDem0n:
The official python kubernetes client contains a lot of auto-generated code and therefore very heavy, but we need only a little fraction of it.
The naive implementation, that covers all API methods we use, takes about 250 LoC, and about half of it is responsible for the handling of configuration files.

Disadvantage: If somebody was using the `patronictl` outside of the pod (on his machine), it might not work anymore (depending on the environment).


FxKu:
👍

CyberDem0n:
:+1:

PR #1590, opened 2020-06-15
Create readiness and liveness endpoints

CyberDem0n:
They could be useful to eliminate "unhealthy" pods from subsets addresses when the K8s service with label selectors are used.
Real-life example: the node where the primary was running has failed and being shutdown and Patroni can't update (remove) the role label.
Therefore on OpenShift the leader service will have two pods assigned, one of them is a failed primary.
With the readiness probe defined, the failed primary pod will be excluded from the list.


erthalion:
:+1:

CyberDem0n:
:+1:

PR #1591, opened 2020-06-15
Populate references and nodename in subsets addresses

CyberDem0n:
It makes subsets to exactly look like they were populated by the service with label selector and would help with https://github.com/zalando/postgres-operator/issues/340#issuecomment-587001109

Unit-tests are refactored to minimize amount of mocks.


erthalion:
:+1:

CyberDem0n:
:+1:

PR #1592, opened 2020-06-15
Don't let Patroni mistake logical replication slots for physical replicas

bmdavi3:
In testing a PostgreSQL plugin that uses logical slots alongside Patroni, we found that the logical slots were mistakenly showing in the Patroni REST API output as though they were physical Patroni-managed replicas. This wasn't causing an obvious problem in Patroni's functioning, however the
output was still incorrect. Although we did not observe such, we believe that this may also impact failovers if Patroni were to try to failover to the (erroneous) logical replication slot.

We tracked this behavior to the SQL statement seen here. This adds logic
to select only physical slots when considering Patroni replicas.

@devmage

### DIFF
--- a/patroni/api.py
+++ b/patroni/api.py
@@ -439,8 +439,8 @@ class RestApiHandler(BaseHTTPRequestHandler):
                     "FROM (SELECT (SELECT rolname FROM pg_authid WHERE oid = usesysid) AS usename,"
                     " application_name, client_addr, w.state, sync_state, sync_priority"
                     " FROM pg_catalog.pg_stat_get_wal_senders() w, pg_catalog.pg_stat_get_activity(pid)"
-                    " INNER JOIN pg_replication_slots AS rs ON rs.active_pid = pid"
-                    " WHERE application_name slot_type = 'physical') AS ri")
+                    " LEFT JOIN pg_replication_slots AS rs ON rs.active_pid = pid"
+                    " WHERE slot_type IS NULL OR slot_type = 'physical') AS ri")
 
             row = self.query(stmt.format(postgresql.wal_name, postgresql.lsn_name), retry=retry)[0]
 
--- a/patroni/api.py
+++ b/patroni/api.py
@@ -438,7 +438,9 @@ class RestApiHandler(BaseHTTPRequestHandler):
                     " pg_catalog.array_to_json(pg_catalog.array_agg(pg_catalog.row_to_json(ri))) "
                     "FROM (SELECT (SELECT rolname FROM pg_authid WHERE oid = usesysid) AS usename,"
                     " application_name, client_addr, w.state, sync_state, sync_priority"
-                    " FROM pg_catalog.pg_stat_get_wal_senders() w, pg_catalog.pg_stat_get_activity(pid)) AS ri")
+                    " FROM pg_catalog.pg_stat_get_wal_senders() w, pg_catalog.pg_stat_get_activity(pid)"
+                    " INNER JOIN pg_replication_slots AS rs ON rs.active_pid = pid"
+                    " WHERE application_name slot_type = 'physical') AS ri")
 
             row = self.query(stmt.format(postgresql.wal_name, postgresql.lsn_name), retry=retry)[0]
 


CyberDem0n:
> logical slots were mistakenly showing in the Patroni REST API output

Actually, this is intentional.

> we believe that this may also impact failovers if Patroni were to try to failover to the (erroneous) logical replication slot

It will not:
1. Every node participates in the leader race independently.
2. Slots or replication status don't play a role here.
3. It is even possible to disable replication slots completely and everything will still work as expected.

devmage:
Hey @CyberDem0n , appreciate the reply. I'm working with @bmdavi3 on this.

> > logical slots were mistakenly showing in the Patroni REST API output
> 
> Actually, this is intentional.

Would you be willing to explain this in a couple sentences? We would really find that helpful. It isn't obvious to us from the Patroni API output why this would be an intentional behavior.

> > we believe that this may also impact failovers if Patroni were to try to failover to the (erroneous) logical replication slot
> 
> It will not:
> 
> 1. Every node participates in the leader race independently.
> 2. Slots or replication status don't play a role here.
> 3. It is even possible to disable replication slots completely and everything will still work as expected.

Thanks for this information! Good to know. 🙂

CyberDem0n:
Until PostgreSQL 9.6 there was no way to get information about wal receiver process on the replica, so we added to the API results of SELECT from`pg_stat_replication` view so we can monitor that replicas are actually streaming.
It turned out to be quite handy, because the API is also showing that for example there is a pg_basebackup streaming from the given host.

Also, it is possible to disable replication slots and rely exclusively on archive_command/restore_command. I know a few such setups. Doing JOIN with pg_replication_slots will make the resultset empty even if all replicas are successfully streaming.

Actually, there is a way to exclude logical streams without joining with any other views or calling other functions, it is enough to check that `datid <> 0`, but personally I consider showing something for logical replication as a bonus which we get for free.

devmage:
> Until PostgreSQL 9.6 there was no way to get information about wal receiver process on the replica, so we added to the API results of SELECT from`pg_stat_replication` view so we can monitor that replicas are actually streaming.
> It turned out to be quite handy, because the API is also showing that for example there is a pg_basebackup streaming from the given host.

Understood, thank you. So this was added as a convenience mechanism for monitoring, before there was a simple way to monitor streaming for replicas. That makes a lot of sense, I definitely see the utility. 🙂 

> Also, it is possible to disable replication slots and rely exclusively on archive_command/restore_command. I know a few such setups. Doing JOIN with pg_replication_slots will make the resultset empty even if all replicas are successfully streaming.
> Actually, there is a way to exclude logical streams without joining with any other views or calling other functions, it is enough to check that `datid <> 0`

Thanks for these ideas! I like the `datid` approach, I think we would also prefer that over joining another view, although I worry a bit that relying on the value of `datid` feels like relying on an implementation detail. @bmdavi3 also remarks that we might be able to switch the code we show here from an `INNER JOIN` to a `LEFT JOIN`, to continue showing information when replication slots aren't in use. We'll test that on our side to confirm.

> but personally I consider showing something for logical replication as a bonus which we get for free.

My objection, and the genesis of our PR, is that `/cluster` is described as showing "the current cluster topology and state", whereas an arbitrary logical replication slot isn't part of that topology. The erroneous entry we're seeing in `members` isn't a cluster member, it doesn't represent a node that should be considered by an operator. In other words, showing logical slots in the output of `/cluster` doesn't seem to match its stated intent. Maybe adjusting the docs to reflect this behavior would help clarify that intent, like if the docs also explained this logical slot behavior.

All this said though, we can just ignore the logical slot entry on our side. 🤷 

Since you mention that this is a holdover from pre-9.6, maybe this is a conversation better left for February 2021 when PostgreSQL 9.5 is slated for its last update.

patroni/api.py
```diff
@@ -439,8 +439,8 @@ def get_postgresql_status(self, retry=False):
                     "FROM (SELECT (SELECT rolname FROM pg_authid WHERE oid = usesysid) AS usename,"
                     " application_name, client_addr, w.state, sync_state, sync_priority"
                     " FROM pg_catalog.pg_stat_get_wal_senders() w, pg_catalog.pg_stat_get_activity(pid)"
-                    " INNER JOIN pg_replication_slots AS rs ON rs.active_pid = pid"
-                    " WHERE application_name slot_type = 'physical') AS ri")
+                    " LEFT JOIN pg_replication_slots AS rs ON rs.active_pid = pid"
+                    " WHERE slot_type IS NULL OR slot_type = 'physical') AS ri")
```
devmage, line 443: Responding to:
> Doing JOIN with pg_replication_slots will make the resultset empty even if all replicas are successfully streaming.

We found that this change limits results to the physical slots that we would expect, plus WAL senders that are not associated with a replication slot.

To confirm this in manual testing, we adjusted the join condition to `rs.active_pid = 0`, which still showed all the slots on our host. So, this query should only exclude WAL senders associated with a logical (or non-physical) replication slot.

CyberDem0n:
> My objection, and the genesis of our PR, is that /cluster is described as showing "the current cluster topology and state", whereas an arbitrary logical replication slot isn't part of that topology. The erroneous entry we're seeing in members isn't a cluster member, it doesn't represent a node that should be considered by an operator. In other words, showing logical slots in the output of /cluster doesn't seem to match its stated intent. 

Well, nobody promised you that it should match.
The `/cluster` endpoint shows only members of the cluster. This is more or less the equivalent of `patronictl list`
The `/patroni` endpoint shows the state of the **current member**. The only intention of the `replication` field in the JSON output is to show that someone or somebody is replicating from the node. Please don't try to find any hidden meaning here. 

I am pretty sure that someone might rely on this information for monitoring, especially in case of logical decoding (wal2json, debezium, and so on), and removing it is undesirable.

devmage:
Ok! Thanks for the conversation in here then. You make good points about other people relying on WAL sender information for monitoring, and we wouldn't want to upset their use-cases without clearly communicating that beforehand.

In the future, perhaps it would be useful to amend the docs around `/cluster` to explain that `members` shows info for _all_ WAL senders, and is not just a list of Patroni cluster members. That's where we were getting confused here. I don't believe we were using `/patroni` for reference, only `/cluster` and `/health` on our end.

We will close this PR. Thanks again! 👍

devmage:
I'm so sorry! On further review, we were confusing ourselves between the `/cluster` and `/health` endpoints from the beginning.

The (to us) unexpected replication slot was showing up, arguably correctly, in the `replication` section in `/health`, not in `members` in `/cluster`.

That's our mistake, and apologies for taking up time on that non-issue. Thanks again for your patience with us here. 😅

PR #1594, opened 2020-06-16
Multi Sync Standby Support

ksarabu1:
Credit to @ants.

As per comment in PR [672](https://github.com/zalando/patroni/pull/672), this is to extend current one synchronous standby functionality to support precise number of synchronous standby databases. 

The new parameter synchronous_node_count is used by Patroni to manage number of synchronous standby databases. It is set to 1 by default. It has no effect when synchronous_mode is set to off. When enabled, Patroni manages precise number of synchronous standby databases based on parameter synchronous_node_count and adjusts the state in DCS & synchronous_standby_names as members join and leave.

This functionality can be further extended to support Priority (FIRST n) based synchronous replication & Quorum (ANY n) based synchronous replication in future.

### Synchronous Replication Factor

The parameter ``synchronous_node_count`` is used by Patroni to manage number of synchronous standby databases. It is set to 1 by default. It has no effect when ``synchronous_mode`` is set to off. When enabled, Patroni manages precise number of synchronous standby databases based on parameter ``synchronous_node_count`` and adjusts the state in DCS & synchronous_standby_names as members join and leave.

### DIFF
--- a/docs/replication_modes.rst
+++ b/docs/replication_modes.rst
@@ -27,7 +27,7 @@ To enable a simple synchronous replication test, add the following lines to the 
 .. code:: YAML
 
         synchronous_commit: "on"
-        synchronous_standby_names: "*"
+        synchronous_node_count: 1
 
 When using PostgreSQL synchronous replication, use at least three Postgres data nodes to ensure write availability if one host fails.
 
@@ -59,11 +59,14 @@ Synchronous mode can be switched on and off via Patroni REST interface. See :ref
 
 Note: Because of the way synchronous replication is implemented in PostgreSQL it is still possible to lose transactions even when using ``synchronous_mode_strict``. If the PostgreSQL backend is cancelled while waiting to acknowledge replication (as a result of packet cancellation due to client timeout or backend failure) transaction changes become visible for other backends. Such changes are not yet replicated and may be lost in case of standby promotion.
 
+Synchronous Replication Factor
+------------------------------
+The parameter ``synchronous_node_count`` is used by Patroni to manage number of synchronous standby databases. It is set to 1 by default. It has no effect when ``synchronous_mode`` is set to off. When enabled, Patroni manages precise number of synchronous standby databases based on parameter ``synchronous_node_count`` and adjusts the state in DCS & synchronous_standby_names as members join and leave.
 
 Synchronous mode implementation
 -------------------------------
 
-When in synchronous mode Patroni maintains synchronization state in the DCS, containing the latest primary and current synchronous standby. This state is updated with strict ordering constraints to ensure the following invariants:
+When in synchronous mode Patroni maintains synchronization state in the DCS, containing the latest primary and current synchronous standby databases. This state is updated with strict ordering constraints to ensure the following invariants:
 
 - A node must be marked as the latest leader whenever it can accept write transactions. Patroni crashing or PostgreSQL not shutting down can cause violations of this invariant.
 
@@ -71,9 +74,9 @@ When in synchronous mode Patroni maintains synchronization state in the DCS, con
 
 - A node that is not the leader or current synchronous standby is not allowed to promote itself automatically.
 
-Patroni will only ever assign one standby to ``synchronous_standby_names`` because with multiple candidates it is not possible to know which node was acting as synchronous during the failure.
+Patroni will only assign one or more synchronous standby nodes based on ``synchronous_node_count`` parameter to ``synchronous_standby_names``.
 
-On each HA loop iteration Patroni re-evaluates synchronous standby choice. If the current synchronous standby is connected and has not requested its synchronous status to be removed it remains picked. Otherwise the cluster member available for sync that is furthest ahead in replication is picked.
+On each HA loop iteration Patroni re-evaluates synchronous standby nodes choice. If the current list of synchronous standby nodes are connected and has not requested its synchronous status to be removed it remains picked. Otherwise the cluster member available for sync that is furthest ahead in replication is picked.
 
 
 .. [1] The data is still there, but recovering it requires a manual recovery effort by data recovery specialists. When Patroni is allowed to rewind with ``use_pg_rewind`` the forked timeline will be automatically erased to rejoin the failed primary with the cluster.
--- a/patroni/api.py
+++ b/patroni/api.py
@@ -119,7 +119,7 @@ class RestApiHandler(BaseHTTPRequestHandler):
             status_code = 200 if response.get('state') == 'running' else 503
         elif cluster:  # dcs is available
             is_synchronous = cluster.is_synchronous_mode() and cluster.sync \
-                    and cluster.sync.sync_standby == patroni.postgresql.name
+                    and cluster.sync.matches(patroni.postgresql.name)
             if path in ('/sync', '/synchronous') and is_synchronous:
                 status_code = replica_status_code
             elif path in ('/async', '/asynchronous') and not is_synchronous:
@@ -331,13 +331,13 @@ class RestApiHandler(BaseHTTPRequestHandler):
         if leader and (not cluster.leader or cluster.leader.name != leader):
             return 'leader name does not match'
         if candidate:
-            if action == 'switchover' and cluster.is_synchronous_mode() and cluster.sync.sync_standby != candidate:
+            if action == 'switchover' and cluster.is_synchronous_mode() and not cluster.sync.matches(candidate):
                 return 'candidate name does not match with sync_standby'
             members = [m for m in cluster.members if m.name == candidate]
             if not members:
                 return 'candidate does not exists'
         elif cluster.is_synchronous_mode():
-            members = [m for m in cluster.members if m.name == cluster.sync.sync_standby]
+            members = [m for m in cluster.members if cluster.sync.matches(m.name)]
             if not members:
                 return action + ' is not possible: can not find sync_standby'
         else:
--- a/patroni/dcs/__init__.py
+++ b/patroni/dcs/__init__.py
@@ -381,7 +381,7 @@ class SyncState(namedtuple('SyncState', 'index,leader,sync_standby')):
                 data = {}
         else:
             data = {}
-        return SyncState(index, data.get('leader'), data.get('sync_standby'))
+        return SyncState(index, data.get('leader'), data.get('sync_standby', []))
 
     def matches(self, name):
         """
@@ -399,7 +399,10 @@ class SyncState(namedtuple('SyncState', 'index,leader,sync_standby')):
         >>> SyncState(1, None, None).matches('foo')
         False
         """
-        return name is not None and name in (self.leader, self.sync_standby)
+        member_list = [self.leader]
+        if self.sync_standby:
+            member_list += self.sync_standby
+        return name is not None and name in member_list
 
 
 class TimelineHistory(namedtuple('TimelineHistory', 'index,value,lines')):
--- a/patroni/ha.py
+++ b/patroni/ha.py
@@ -437,7 +437,7 @@ class Ha(object):
     def is_synchronous_mode_strict(self):
         return self.check_mode('synchronous_mode_strict')
 
-    def process_sync_replication(self):
+    def process_sync_replication(self, sync_node_count=1):
         """Process synchronous standby beahvior.
 
         Synchronous standbys are registered in two places postgresql.conf and DCS. The order of updating them must
@@ -447,27 +447,33 @@ class Ha(object):
         promoting standbys that were guaranteed to be replicating synchronously.
         """
         if self.is_synchronous_mode():
-            current = self.cluster.sync.leader and self.cluster.sync.sync_standby
-            picked, allow_promote = self.state_handler.pick_synchronous_standby(self.cluster)
-            if picked != current:
-                # We need to revoke privilege from current before replacing it in the config
-                if current:
-                    logger.info("Removing synchronous privilege from %s", current)
-                    if not self.dcs.write_sync_state(self.state_handler.name, None, index=self.cluster.sync.index):
+            current = self.cluster.sync.leader and self.cluster.sync.sync_standby or []
+            picked, allow_promote = self.state_handler.pick_synchronous_standby(self.cluster, sync_node_count)
+            if set(picked) != set(current):
+                if picked:
+                    logger.info("Updating synchronous privilege from {} to {}".format(','.join(current),
+                                                                                      ','.join(picked)))
+                    if not self.dcs.write_sync_state(self.state_handler.name, picked, index=self.cluster.sync.index):
                         logger.info('Synchronous replication key updated by someone else.')
                         return
 
-                if self.is_synchronous_mode_strict() and picked is None:
-                    picked = '*'
+                if current and not picked:
+                    logger.info("Removing synchronous privilege from {}".format(','.join(current)))
+                    if not self.dcs.write_sync_state(self.state_handler.name, [], index=self.cluster.sync.index):
+                        logger.info('Synchronous replication key updated by someone else.')
+                        return
+
+                if self.is_synchronous_mode_strict() and not picked:
+                    picked = ['*']
                     logger.warning("No standbys available!")
 
                 logger.info("Assigning synchronous standby status to %s", picked)
                 self.state_handler.config.set_synchronous_standby(picked)
 
-                if picked and picked != '*' and not allow_promote:
+                if picked and picked[0] != '*' and not allow_promote:
                     # Wait for PostgreSQL to enable synchronous mode and see if we can immediately set sync_standby
                     time.sleep(2)
-                    picked, allow_promote = self.state_handler.pick_synchronous_standby(self.cluster)
+                    picked, allow_promote = self.state_handler.pick_synchronous_standby(self.cluster, sync_node_count)
                 if allow_promote:
                     try:
                         cluster = self.dcs.get_cluster()
@@ -483,11 +489,11 @@ class Ha(object):
         else:
             if self.cluster.sync.leader and self.dcs.delete_sync_state(index=self.cluster.sync.index):
                 logger.info("Disabled synchronous replication")
-            self.state_handler.config.set_synchronous_standby(None)
+            self.state_handler.config.set_synchronous_standby([])
 
     def is_sync_standby(self, cluster):
         return cluster.leader and cluster.sync.leader == cluster.leader.name \
-            and cluster.sync.sync_standby == self.state_handler.name
+            and cluster.sync.matches(self.state_handler.name)
 
     def while_not_sync_standby(self, func):
         """Runs specified action while trying to make sure that the node is not assigned synchronous standby status.
@@ -562,21 +568,21 @@ class Ha(object):
             # Inform the state handler about its master role.
             # It may be unaware of it if postgres is promoted manually.
             self.state_handler.set_role('master')
-            self.process_sync_replication()
+            self.process_sync_replication(self.patroni.config['synchronous_node_count'])
             self.update_cluster_history()
             return message
         elif self.state_handler.role == 'master':
-            self.process_sync_replication()
+            self.process_sync_replication(self.patroni.config['synchronous_node_count'])
             return message
         else:
             if self.is_synchronous_mode():
                 # Just set ourselves as the authoritative source of truth for now. We don't want to wait for standbys
                 # to connect. We will try finding a synchronous standby in the next cycle.
-                if not self.dcs.write_sync_state(self.state_handler.name, None, index=self.cluster.sync.index):
+                if not self.dcs.write_sync_state(self.state_handler.name, [], index=self.cluster.sync.index):
                     # Somebody else updated sync state, it may be due to us losing the lock. To be safe, postpone
                     # promotion until next cycle. TODO: trigger immediate retry of run_cycle
                     return 'Postponing promotion because synchronous replication state was updated by somebody else'
-                self.state_handler.config.set_synchronous_standby('*' if self.is_synchronous_mode_strict() else None)
+                self.state_handler.config.set_synchronous_standby(['*'] if self.is_synchronous_mode_strict() else [])
             if self.state_handler.role != 'master':
                 self.set_leader_access_is_restricted(self.cluster.has_permanent_logical_slots(self.state_handler.name))
 
@@ -811,7 +817,7 @@ class Ha(object):
             self._async_executor.try_run_async('starting after demotion', self.state_handler.follow, (node_to_follow,))
         else:
             if self.is_synchronous_mode():
-                self.state_handler.config.set_synchronous_standby(None)
+                self.state_handler.config.set_synchronous_standby([])
             if self._rewind.rewind_or_reinitialize_needed_and_possible(leader):
                 return False  # do not start postgres, but run pg_rewind on the next iteration
             self.state_handler.follow(node_to_follow)
@@ -872,8 +878,8 @@ class Ha(object):
                 else:
                     if self.is_synchronous_mode():
                         if failover.candidate and not self.cluster.sync.matches(failover.candidate):
-                            logger.warning('Failover candidate=%s does not match with sync_standby=%s',
-                                           failover.candidate, self.cluster.sync.sync_standby)
+                            logger.warning('Failover candidate=%s does not match with sync_standbys=%s',
+                                           failover.candidate, ','.join(self.cluster.sync.sync_standby))
                             members = []
                         else:
                             members = [m for m in self.cluster.members if self.cluster.sync.matches(m.name)]
--- a/patroni/postgresql/__init__.py
+++ b/patroni/postgresql/__init__.py
@@ -889,7 +889,7 @@ class Postgresql(object):
             logger.exception('Could not remove data directory %s', self._data_dir)
             self.move_data_directory()
 
-    def pick_synchronous_standby(self, cluster):
+    def pick_synchronous_standby(self, cluster, sync_node_count=1):
         """Finds the best candidate to be the synchronous standby.
 
         Current synchronous standby is always preferred, unless it has disconnected or does not want to be a
@@ -897,30 +897,41 @@ class Postgresql(object):
 
         :returns tuple of candidate name or None, and bool showing if the member is the active synchronous standby.
         """
-        current = cluster.sync.sync_standby
-        current = current.lower() if current else current
+        current = [x.lower() for x in cluster.sync.sync_standby] if cluster.sync.sync_standby else []
         members = {m.name.lower(): m for m in cluster.members}
         candidates = []
         # Pick candidates based on who has flushed WAL farthest.
         # TODO: for synchronous_commit = remote_write we actually want to order on write_location
+        try:
+            query = "SELECT setting from pg_settings where name = 'synchronous_commit'"
+            sync_commit_par = self.query(query).fetchone()[0]
+        except AttributeError:
+            sync_commit_par = None
+        sort_lsn_col = "flush_{}".format(self.lsn_name)
+        if sync_commit_par == 'remote_apply':
+            sort_lsn_col = "replay_{}".format(self.lsn_name)
+        elif sync_commit_par == 'remote_write':
+            sort_lsn_col = "write_{}".format(self.lsn_name)
         for app_name, state, sync_state in self.query(
                 "SELECT pg_catalog.lower(application_name), state, sync_state"
                 " FROM pg_catalog.pg_stat_replication"
-                " ORDER BY flush_{0} DESC".format(self.lsn_name)):
+                " WHERE state = 'streaming'"
+                " ORDER BY (case when sync_state='sync' then 1 "
+                "  when sync_state='potential' then 2 else 3 end), {0} DESC".format(sort_lsn_col)):
             member = members.get(app_name)
-            if state != 'streaming' or not member or member.tags.get('nosync', False):
+            if not member or member.tags.get('nosync', False):
                 continue
-            if sync_state == 'sync':
-                return member.name, True
-            if sync_state == 'potential' and app_name == current:
-                # Prefer current even if not the best one any more to avoid indecisivness and spurious swaps.
-                return cluster.sync.sync_standby, False
-            if sync_state in ('async', 'potential'):
-                candidates.append(member.name)
+            candidates.append(member.name)
+            if len(candidates) >= sync_node_count:
+                break
 
-        if candidates:
-            return candidates[0], False
-        return None, False
+        if self._major_version < 96000 and candidates:
+            candidates = [candidates[0]]
+
+        if current and candidates and set(current) == set(candidates):
+            return candidates, True
+
+        return candidates, False
 
     def schedule_sanity_checks_after_pause(self):
         """
--- a/patroni/postgresql/config.py
+++ b/patroni/postgresql/config.py
@@ -824,12 +824,19 @@ class ConfigHandler(object):
         parameters.update(cluster_name=self._postgresql.scope, listen_addresses=listen_addresses, port=str(port))
         if config.get('synchronous_mode', False):
             if self._synchronous_standby_names is None:
-                if config.get('synchronous_mode_strict', False):
+                if config.get('synchronous_mode_strict', False) and \
+                   hasattr(self._postgresql, 'role') and \
+                   self._postgresql.role == 'master':
                     parameters['synchronous_standby_names'] = '*'
                 else:
                     parameters.pop('synchronous_standby_names', None)
             else:
+                # TODO identify change in synchronous_node_count and compare it with current num_sync,
+                # adjust synchronous_standby_names accordingly. Leaving to existing value for now for
+                # Patroni to reset the value during next run cycle
                 parameters['synchronous_standby_names'] = self._synchronous_standby_names
+        else:
+            parameters.pop('synchronous_standby_names', None)
         if self._postgresql.major_version >= 90600 and parameters['wal_level'] == 'hot_standby':
             parameters['wal_level'] = 'replica'
         ret = CaseInsensitiveDict({k: v for k, v in parameters.items() if not self._postgresql.major_version or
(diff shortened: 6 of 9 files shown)


patroni/dcs/__init__.py
```diff
@@ -381,13 +381,13 @@ def from_node(index, value):
                 data = {}
         else:
             data = {}
-        return SyncState(index, data.get('leader'), data.get('sync_standby'))
+        return SyncState(index, data.get('leader'), data.get('sync_standby', []))
```
CyberDem0n, line 384: What about backward compatibility?
When replicas are upgraded they should still be able to read the data in the old format.

patroni/ha.py
```diff
@@ -447,27 +447,33 @@ def process_sync_replication(self):
...
-                if current:
-                    logger.info("Removing synchronous privilege from %s", current)
-                    if not self.dcs.write_sync_state(self.state_handler.name, None, index=self.cluster.sync.index):
+            current = self.cluster.sync.leader and self.cluster.sync.sync_standby or []
+            picked, allow_promote = self.state_handler.pick_synchronous_standby(self.cluster, sync_node_count)
+            if set(picked) != set(current):
+                if picked:
+                    logger.info("Updating synchronous privilege from {} to {}".format(','.join(current),
+                                                                                      ','.join(picked)))
+                    if not self.dcs.write_sync_state(self.state_handler.name, picked, index=self.cluster.sync.index):
```
CyberDem0n, line 456: This is where the dragon sits...

Let's imagine that the `current` is `(n1, n2)` and the `picked` is `(n3, n4)`.
By publishing `(n3, n4)` to DCS we are giving them a green light for the leader race.
BUT! In the `synchronous_standby_names` they are not yet declared as synchronous, therefore we can't guaranty no data loss during failover.

Order matters a lot here:
1. First you should "remove" from DCS `current` names that are not in `picked`.
2. Update `synchronous_standby_names` by putting `picked` names there.
3. Wait for some of `picked` nodes to become really synchronous.
4. Publish into DCS `picked` nodes that are already synchronous.

Basically, the [quorum state resolver](https://github.com/zalando/patroni/pull/672/files#diff-a839e95c7271edf3b62934f9e0e61f16R18) is the attempt to implement such an algorithm.

ksarabu1: Agree. Will fix the code.

patroni/dcs/__init__.py
ksarabu1: We could possibly use a new key instead to store synchronous standbys list and leave the existing key to keep only one node info to provide backward compatibility for its peers during transition to new version. Please let me know if this is acceptable so I can fix the code accordingly. Thanks.

ksarabu1:
Updated code as per feedback. Please review. I will fix/add unit tests if the proposed change look good and acceptable. Thanks.

ksarabu1:
New changes use comma delim string for sync_standby_list in DCS. This is to address CI failure on using unsupported list in k8s annotations.

patroni/ha.py
```diff
@@ -437,57 +437,65 @@ def is_synchronous_mode(self):
...
         if self.is_synchronous_mode():
-            current = self.cluster.sync.leader and self.cluster.sync.sync_standby
-            picked, allow_promote = self.state_handler.pick_synchronous_standby(self.cluster)
-            if picked != current:
-                # We need to revoke privilege from current before replacing it in the config
-                if current:
-                    logger.info("Removing synchronous privilege from %s", current)
-                    if not self.dcs.write_sync_state(self.state_handler.name, None, index=self.cluster.sync.index):
+            current = self.cluster.sync.leader and self.cluster.sync.members or []
+            picked, allow_promote = self.state_handler.pick_synchronous_standby(self.cluster, sync_node_count)
```
CyberDem0n, line 451: IMO allow_promote needs rethinking.
With the only possible sync standby it was an indicator that the specific node is already `sync` in pg_stat_replication.
That also means we are [allowed to publish](https://github.com/zalando/patroni/pull/1594/files#diff-57c53e988b775ee2902488fb33c1f433L471-L482) it into the sync key in DCS.

Since the `picked` became the list, the `allow_promote` should be too and it should contain names of nodes from the `picked` that are already synchronous. Only nodes from the allow_promote list are allowed to be published into DCS.

patroni/postgresql/config.py
```diff
@@ -1007,16 +1013,20 @@ def reload_config(self, config, sighup=False):
...
+    def set_synchronous_standby(self, sync_members):
         """Sets a node to be synchronous standby and if changed does a reload for PostgreSQL."""
-        if name and name != '*':
-            name = quote_ident(name)
-        if name != self._synchronous_standby_names:
-            if name is None:
+        if sync_members and set(sync_members) != set(['*']):
+            sync_members = [quote_ident(x) for x in sync_members]
+        if self._postgresql.major_version >= 90600:
+            sync_param = sync_members and '{} ({})'.format(len(sync_members), ','.join(sync_members)) or None
```
CyberDem0n, line 1021: The automatic reduction of `synchronous_node_count` is fine only for non-strict synchronous mode.

Let's say we have `synchronous_mode_strict = True`, `synchronous_node_count = 2` and `sync_members = ['*']`.
In this case we want to put `2 (*)` into `synchronous_standby_names`.

patroni/dcs/__init__.py
```diff
@@ -381,25 +381,38 @@ def from_node(index, value):
...
+        elif 'sync_standby' in data.keys():
+            sync_list = data.get('sync_standby')
+        else:
+            sync_list = None
+        return SyncState(index, data.get('leader'), sync_list)
+
+    @property
+    def members(self):
+        """ Returns sync_standby_list in list """
+        return self.sync_standby_list and self.sync_standby_list.split(',') or []
```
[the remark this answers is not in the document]
ksarabu1: Agree. Will update the changes back to using sync_standby. Thanks.

patroni/ha.py
```diff
@@ -437,57 +437,65 @@ def is_synchronous_mode(self):
...
+            if set(picked) != set(current):
+                # update synchronous standby list in dcs temporarily to point to common nodes in current and picked
+                sync_common = list(set(current).intersection(set(picked)))
+                if set(sync_common) != set(current):
+                    logger.info("Updating synchronous state temporarily in dcs from {} to {}".format(current,
+                                                                                                     sync_common))
```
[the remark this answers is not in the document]
ksarabu1: will fix.

patroni/postgresql/config.py
```diff
@@ -920,7 +927,6 @@ def _handle_wal_buffers(old_values, changes):
     def reload_config(self, config, sighup=False):
         self._superuser = config['authentication'].get('superuser', {})
         server_parameters = self.get_server_parameters(config)
-
```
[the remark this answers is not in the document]
ksarabu1: will fix.

patroni/postgresql/__init__.py
```diff
@@ -889,38 +889,51 @@ def remove_data_directory(self):
...
+        elif sync_commit_par == 'remote_write':
+            sort_lsn_col = "write_{}".format(self.lsn_name)
+        last_sync_state = None
         for app_name, state, sync_state in self.query(
                 "SELECT pg_catalog.lower(application_name), state, sync_state"
                 " FROM pg_catalog.pg_stat_replication"
-                " ORDER BY flush_{0} DESC".format(self.lsn_name)):
+                " WHERE state = 'streaming'"
+                " ORDER BY (case when sync_state='sync' then 1 "
+                "  when sync_state='potential' then 2 else 3 end), {0} DESC".format(sort_lsn_col)):
```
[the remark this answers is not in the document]
ksarabu1: Agree, will fix the sync_state sorting. As per sort_lsn_col alone, I have tried using that initially, but I think it was unnecessarily causing patroni keep updating the list in each iteration and is the reason for adding sync_state in order by. Will check it again and make changes accordingly.

features/steps/cascading_replication.py
```diff
@@ -21,6 +21,27 @@ def write_label(context, content, name):
     context.pctl.write_label(name, content)
 
 
+@step('"{name}" sync key in DCS has {key:w}={value:w} after {time_limit:d} seconds')
```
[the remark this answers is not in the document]
ksarabu1: will fix

patroni/ha.py
ksarabu1: It make sense. Will update the code.

patroni/postgresql/config.py
ksarabu1: Intention was to make the numsync dynamically adjusted by Patroni so it won't affect leader as replica nodes join and leave. We could add additional parameter to keep min_synchronous_node_count and act accordingly?

CyberDem0n: Hmm, probably you are right. So far semantic of `synchronous_mode_strict` was that we require at least one sync node and suggested implementation doesn't change it.
Ok, let's leave it as it is for now, and later we could agree on configuration options names, i.e. whether we should use `synchronous_node_count` and `synchronous_node_count` or maybe better `replication_factor` and `minimum_replication_factor`, like it was proposed in the #672.

ksarabu1: Agree, It make sense to have configurable parameter for minimum synchronous stanbys instead of hard limit of 1. I will update the code accordingly.

features/multi_sync_replication.feature
```diff
@@ -0,0 +1,38 @@
+Feature: multi sync replication
```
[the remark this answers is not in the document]
ksarabu1: Sure, will merge the tests into basic replication.

patroni/postgresql/config.py
```diff
@@ -824,12 +824,19 @@ def get_server_parameters(self, config):
...
                     parameters['synchronous_standby_names'] = '*'
                 else:
                     parameters.pop('synchronous_standby_names', None)
             else:
+                # TODO identify change in synchronous_node_count and compare it with current num_sync,
+                # adjust synchronous_standby_names accordingly. Leaving to existing value for now for
+                # Patroni to reset the value during next run cycle
                 parameters['synchronous_standby_names'] = self._synchronous_standby_names
+        else:
+            parameters.pop('synchronous_standby_names', None)
```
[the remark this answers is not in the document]
ksarabu1: During the tests, noticed that Patroni updating synchronous_standby_names on standby databases when synchronous_mode_strict was turned on. This change is to avoid changes on standby nodes.

patroni/postgresql/config.py
```diff
@@ -824,12 +824,17 @@ def get_server_parameters(self, config):
         parameters.update(cluster_name=self._postgresql.scope, listen_addresses=listen_addresses, port=str(port))
         if config.get('synchronous_mode', False):
             if self._synchronous_standby_names is None:
-                if config.get('synchronous_mode_strict', False):
+                if config.get('synchronous_mode_strict', False) and \
+                   hasattr(self._postgresql, 'role') and \
+                   self._postgresql.role == 'master':
```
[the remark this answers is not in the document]
ksarabu1: Thanks. Will revert the behavior back.

patroni/postgresql/config.py
```diff
@@ -824,12 +824,17 @@ def get_server_parameters(self, config):
...
                     parameters['synchronous_standby_names'] = '*'
                 else:
                     parameters.pop('synchronous_standby_names', None)
             else:
                 parameters['synchronous_standby_names'] = self._synchronous_standby_names
+        else:
+            if 'synchronous_standby_names' in parameters.keys():
+                parameters.pop('synchronous_standby_names', None)
```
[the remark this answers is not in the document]
ksarabu1: Thanks. Will revert the change.

patroni/postgresql/__init__.py
```diff
@@ -889,38 +889,44 @@ def remove_data_directory(self):
             logger.exception('Could not remove data directory %s', self._data_dir)
             self.move_data_directory()
 
-    def pick_synchronous_standby(self, cluster):
+    def _get_synchronous_commit_param(self):
+        query = "SELECT setting from pg_settings where name = 'synchronous_commit'"
```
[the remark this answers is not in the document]
ksarabu1: will fix it.

patroni/postgresql/__init__.py
```diff
@@ -889,38 +889,44 @@ def remove_data_directory(self):
...
+        # Pick candidates based on who has higher replay/remote_write/flush lsn.
+        sync_commit_par = self._get_synchronous_commit_param()
+        sort_lsn_col = {'remote_apply': 'replay', 'remote_write': 'write'}.get(sync_commit_par, 'flush')
+        sort_lsn_col = '{0}_{1}'.format(sort_lsn_col, self.lsn_name)
         for app_name, state, sync_state in self.query(
                 "SELECT pg_catalog.lower(application_name), state, sync_state"
                 " FROM pg_catalog.pg_stat_replication"
-                " ORDER BY flush_{0} DESC".format(self.lsn_name)):
+                " WHERE state = 'streaming'"
+                " ORDER BY sync_state DESC, {0} DESC".format(sort_lsn_col)):
```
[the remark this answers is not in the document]
ksarabu1: Sure.  The sort option ``ORDER BY sync_state DESC`` is with the intent to keep the Query result predicable.  Especially in cases when a synchronous standby member is slowed down OR async node receiving changes faster than the sync member (very rare but possible). Such cases would trigger sync standby members swapping frequently and we want to avoid that by keeping the query result predictable.

patroni/postgresql/__init__.py
```diff
@@ -889,38 +889,44 @@ def remove_data_directory(self):
...
             member = members.get(app_name)
-            if state != 'streaming' or not member or member.tags.get('nosync', False):
+            if not member or member.tags.get('nosync', False):
                 continue
+            candidates.append(member.name)
             if sync_state == 'sync':
-                return member.name, True
-            if sync_state == 'potential' and app_name == current:
-                # Prefer current even if not the best one any more to avoid indecisivness and spurious swaps.
-                return cluster.sync.sync_standby, False
```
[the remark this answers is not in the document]
ksarabu1: Please excuse my ignorance. I thought that the `ORDER BY sync_state DESC` should keep the order consistent and there shouldn't cause any spurious swaps.

patroni/postgresql/__init__.py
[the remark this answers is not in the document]
CyberDem0n: I mean it deserves a comment in the code, because we are relying on very specific naming of sync_state: https://github.com/zalando/patroni/pull/1594#discussion_r443129801

patroni/postgresql/__init__.py
```diff
@@ -889,38 +889,42 @@ def remove_data_directory(self):
...
-        if candidates:
-            return candidates[0], False
-        return None, False
+                sync_nodes.append(member.name)
+            if len(candidates) >= sync_node_count:
+                break
+
+        if self._major_version < 90600:
+            candidates = candidates and [candidates[0]] or []
+            sync_nodes = sync_nodes and [sync_nodes[0]] or []
```
CyberDem0n, line 925: We can move this if statement to the beginning of the method and make it like
```python
         if self._major_version < 90600:
            sync_node_count = 1
```

tests/test_postgresql.py
```diff
@@ -635,18 +636,18 @@ def value_in_conf():
                         return line.strip()
 
         mock_reload = self.p.reload = Mock()
-        self.p.config.set_synchronous_standby('n1')
-        self.assertEqual(value_in_conf(), "synchronous_standby_names = 'n1'")
+        self.p.config.set_synchronous_standby(['n1'])
+        self.assertEqual(value_in_conf(), "synchronous_standby_names = '1 (n1)'")
```
CyberDem0n, line 640:
```suggestion
        self.assertEqual(value_in_conf(), "synchronous_standby_names = 'n1'")
```

tests/test_postgresql.py
```diff
@@ -635,18 +636,18 @@ def value_in_conf():
...
         mock_reload.assert_called()
 
         mock_reload.reset_mock()
-        self.p.config.set_synchronous_standby('n1')
+        self.p.config.set_synchronous_standby(['n1'])
         mock_reload.assert_not_called()
-        self.assertEqual(value_in_conf(), "synchronous_standby_names = 'n1'")
+        self.assertEqual(value_in_conf(), "synchronous_standby_names = '1 (n1)'")
```
CyberDem0n, line 646:
```suggestion
        self.assertEqual(value_in_conf(), "synchronous_standby_names = 'n1'")
```

tests/test_postgresql.py
```diff
@@ -635,18 +636,18 @@ def value_in_conf():
...
 
-        self.p.config.set_synchronous_standby('n2')
+        self.p.config.set_synchronous_standby(['n2'])
```
CyberDem0n, line 648:
```suggestion
        self.p.config.set_synchronous_standby(['n1', 'n2'])
```

tests/test_postgresql.py
```diff
@@ -635,18 +636,18 @@ def value_in_conf():
...
         mock_reload.assert_called()
-        self.assertEqual(value_in_conf(), "synchronous_standby_names = 'n2'")
+        self.assertEqual(value_in_conf(), "synchronous_standby_names = '1 (n2)'")
```
CyberDem0n, line 650:
```suggestion
        self.assertEqual(value_in_conf(), "synchronous_standby_names = '1 (n1,n2)'")
```

tests/test_postgresql.py
```diff
@@ -635,18 +636,18 @@ def value_in_conf():
...
 
         mock_reload.reset_mock()
         self.p.config.set_synchronous_standby(None)
```
CyberDem0n, line 653:
```suggestion
        self.p.config.set_synchronous_standby([])
```

patroni/api.py
```diff
@@ -119,7 +119,7 @@ def do_GET(self, write_status_code_only=False):
             status_code = 200 if response.get('state') == 'running' else 503
         elif cluster:  # dcs is available
             is_synchronous = cluster.is_synchronous_mode() and cluster.sync \
-                    and cluster.sync.sync_standby == patroni.postgresql.name
+                    and cluster.sync.matches(patroni.postgresql.name)
```
CyberDem0n, line 122:
```suggestion
                    and patroni.postgresql.name in cluster.sync.members
```

patroni/api.py
```diff
@@ -331,13 +331,13 @@ def is_failover_possible(self, cluster, leader, candidate, action):
         if leader and (not cluster.leader or cluster.leader.name != leader):
             return 'leader name does not match'
         if candidate:
-            if action == 'switchover' and cluster.is_synchronous_mode() and cluster.sync.sync_standby != candidate:
+            if action == 'switchover' and cluster.is_synchronous_mode() and not cluster.sync.matches(candidate):
```
CyberDem0n, line 334:
```suggestion
            if action == 'switchover' and cluster.is_synchronous_mode() and candidate not in cluster.sync.members:
```

patroni/api.py
```diff
@@ -331,13 +331,13 @@ def is_failover_possible(self, cluster, leader, candidate, action):
...
                 return 'candidate name does not match with sync_standby'
             members = [m for m in cluster.members if m.name == candidate]
             if not members:
                 return 'candidate does not exists'
         elif cluster.is_synchronous_mode():
-            members = [m for m in cluster.members if m.name == cluster.sync.sync_standby]
+            members = [m for m in cluster.members if cluster.sync.matches(m.name)]
```
CyberDem0n, line 340:
```suggestion
            members = [m for m in cluster.members if m.name in cluster.sync.members]
```

patroni/utils.py
```diff
@@ -387,7 +387,7 @@ def cluster_as_json(cluster):
         if m.name == leader_name:
             config = cluster.config.data if cluster.config and cluster.config.modify_index else {}
             role = 'standby_leader' if is_standby_cluster(config.get('standby_cluster')) else 'leader'
-        elif m.name == cluster.sync.sync_standby:
+        elif cluster.sync.matches(m.name):
```
CyberDem0n, line 390:
```suggestion
        elif m.name in cluster.sync.members:
```

patroni/ha.py
```diff
@@ -476,18 +482,18 @@ def process_sync_replication(self):
...
         else:
             if self.cluster.sync.leader and self.dcs.delete_sync_state(index=self.cluster.sync.index):
                 logger.info("Disabled synchronous replication")
-            self.state_handler.config.set_synchronous_standby(None)
+            self.state_handler.config.set_synchronous_standby([])
 
     def is_sync_standby(self, cluster):
         return cluster.leader and cluster.sync.leader == cluster.leader.name \
-            and cluster.sync.sync_standby == self.state_handler.name
+            and cluster.sync.matches(self.state_handler.name)
```
CyberDem0n, line 496:
```suggestion
            and self.state_handler.name in cluster.sync.members
```

patroni/ha.py
```diff
@@ -447,27 +447,33 @@ def process_sync_replication(self):
...
 
-                if picked and picked != '*' and not allow_promote:
-                    # Wait for PostgreSQL to enable synchronous mode and see if we can immediately set sync_standby
+                if picked and picked[0] != '*' and set(allow_promote) != set(picked):
+                    # Wait for PostgreSQL to enable synchronous mode and see if we can immediately
+                    # set sync_standby
                     time.sleep(2)
-                    picked, allow_promote = self.state_handler.pick_synchronous_standby(self.cluster)
+                    _, allow_promote = self.state_handler.pick_synchronous_standby(self.cluster,
+                                                                                   sync_node_count)
```
CyberDem0n, line 476: I am thinking about little optimization here. As long as we have at least one active sync node we probably don't need to time.sleep(2), but just let it resolve on the next HA loop.

patroni/ha.py
```diff
@@ -562,11 +568,11 @@ def enforce_master_role(self, message, promote_message):
             # Inform the state handler about its master role.
             # It may be unaware of it if postgres is promoted manually.
             self.state_handler.set_role('master')
-            self.process_sync_replication()
+            self.process_sync_replication(self.patroni.config['synchronous_node_count'])
             self.update_cluster_history()
             return message
         elif self.state_handler.role == 'master':
-            self.process_sync_replication()
+            self.process_sync_replication(self.patroni.config['synchronous_node_count'])
```
CyberDem0n, line 575: We could do just `sync_node_count = self.patroni.config['synchronous_node_count']` in the `process_sync_replication()` method.

patroni/postgresql/__init__.py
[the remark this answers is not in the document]
ksarabu1: fixed

docs/replication_modes.rst
```diff
@@ -27,7 +27,7 @@ To enable a simple synchronous replication test, add the following lines to the
 .. code:: YAML
 
         synchronous_commit: "on"
-        synchronous_standby_names: "*"
+        synchronous_node_count: 1
```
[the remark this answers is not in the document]
ksarabu1: reverted the change.

patroni/postgresql/__init__.py
ksarabu1: fixed

patroni/ha.py
ksarabu1: fixed

patroni/ha.py
ksarabu1: It makes sense.  fixing the code.

patroni/ha.py
```diff
@@ -447,27 +447,34 @@ def process_sync_replication(self):
...
+                if self.is_synchronous_mode_strict() and not picked:
+                    picked = ['*']
                     logger.warning("No standbys available!")
 
                 logger.info("Assigning synchronous standby status to %s", picked)
                 self.state_handler.config.set_synchronous_standby(picked)
 
-                if picked and picked != '*' and not allow_promote:
-                    # Wait for PostgreSQL to enable synchronous mode and see if we can immediately set sync_standby
+                if picked and not allow_promote and picked[0] != '*' and set(allow_promote) != set(picked):
```
CyberDem0n, line 472: I mean if the `allow_promote` or `sync_common` (which effectively should be the same) are not empty, we can just return from this method.
Anything else should stay the same.

patroni/ha.py
```diff
@@ -447,27 +447,34 @@ def process_sync_replication(self):
...
+                    # Wait for PostgreSQL to enable synchronous mode and see if we can immediately
+                    # set sync_standby
                     time.sleep(2)
-                    picked, allow_promote = self.state_handler.pick_synchronous_standby(self.cluster)
+                    _, allow_promote = self.state_handler.pick_synchronous_standby(self.cluster,
+                                                                                   sync_node_count)
                 if allow_promote:
```
CyberDem0n, line 478: If the `allow_promote` didn't change compared to the previous value (`sync_common`), we don't want to update sync_state in DCS.

patroni/ha.py
ksarabu1: got it. will fix it.

patroni/ha.py
ksarabu1: will fix the code.

patroni/ha.py
ksarabu1: I am thinking more of comparing `allow_promote` with `picked` and return if they are not same. This should optimize the call and let Patroni handle list in DCS during next iteration? Please let me know your thoughts.

```
                if (picked and picked[0] != '*' and set(allow_promote) != set(picked)):
                    logger.info("Skipping DCS Update for Patroni to take care in next iteration")
                    return
```

CyberDem0n: We don't want to return if `allow_promote` is not empty and `picked` is empty, because it will leave the system in the state where only the former primary would be allowed to promote.

ksarabu1: I think this situation would never arise as the query in `[pick_synchronous_standby](https://github.com/ksarabu1/patroni/blob/feature/multisync/patroni/postgresql/__init__.py#L925-L927)` always return empty `allow_promote` if `picked` is empty.

CyberDem0n: Sorry, I meant the opposite.
We don't want to return if `picked` is not empty and `allow_promote` is empty.

ksarabu1: Thanks. The following should mitigate that condition as well. Please let me know.

```
                if (picked and allow_promote and picked[0] != '*' and set(allow_promote) != set(picked)):
                    logger.info("Skipping DCS Update for Patroni to take care in next iteration")
                    return
```

ksarabu1: Reverting the logic back to previous [commit](https://github.com/zalando/patroni/commit/6a1523278e00597842aa804f2cdd18a130c54f13#diff-57c53e988b775ee2902488fb33c1f433L471) for now as it was tested and worked in all test cases.

tests/test_postgresql.py
```diff
@@ -593,6 +593,7 @@ def time_in_state(*args):
             self.p._state = 'starting'
             self.assertIsNone(self.p.wait_for_startup())
 
+    @patch.object(Postgresql, '_get_synchronous_commit_param', Mock(return_value='remote_apply'))
```
CyberDem0n, line 596: Lets also change this test to:
```python
    def test_pick_sync_standby(self):                                                                                                   
        cluster = Cluster(True, None, self.leader, 0, [self.me, self.other, self.leadermem], None,                                      
                          SyncState(0, self.me.name, self.leadermem.name), None)                                                        
        mock_cursor = Mock()                                                                                                            
        mock_cursor.fetchone.return_value = ('remote_apply',)                                                                           
                                                                                                                                        
        with patch.object(Postgresql, "query", side_effect=[                                                                            
                    mock_cursor,                                                                                                        
                    [(self.leadermem.name, 'streaming', 'sync'),                                                                        
                     (self.me.name, 'streaming', 'async'),                                                                              
                     (self.other.name, 'streaming', 'async')]                                                                           
                ]):                                                                                                                     
            self.assertEqual(self.p.pick_synchronous_standby(cluster), ([self.leadermem.name], [self.leadermem.name]))                  
                                                                                                                                        
        with patch.object(Postgresql, "query", side_effect=[                                                                            
                    mock_cursor,                                                                                                        
                    [(self.leadermem.name, 'streaming', 'potential'),                                                                   
                     (self.me.name, 'streaming', 'async'),                                                                              
                     (self.other.name, 'streaming', 'async')]                                                                           
                ]):                                                                                                                     
            self.assertEqual(self.p.pick_synchronous_standby(cluster), ([self.leadermem.name], []))                                     
                                                                                                                                        
        with patch.object(Postgresql, "query", side_effect=[                                                                            
                    mock_cursor,                                                                                                        
                    [(self.me.name, 'streaming', 'async'),                                                                              
                     (self.other.name, 'streaming', 'async')]                                                                           
                ]):                                                                                                                     
            self.assertEqual(self.p.pick_synchronous_standby(cluster), ([self.me.name], []))                                            
                                                                                                                                        
        with patch.object(Postgresql, "query", side_effect=[                                                                            
                    mock_cursor,                                                                                                        
                    [('missing', 'streaming', 'sync'),                                                                                  
                     (self.me.name, 'streaming', 'async'),                                                                              
                     (self.other.name, 'streaming', 'async')]                                                                           
                ]):                                                                                                                     
            self.assertEqual(self.p.pick_synchronous_standby(cluster), ([self.me.name], []))                                            
                                                                                                                                        
        with patch.object(Postgresql, "query", side_effect=[mock_cursor, []]):                                                          
            self.p._major_version = 90400                                                                                               
            self.assertEqual(self.p.pick_synchronous_standby(cluster), ([], []))                                                        
```

I just want that `pick_synchronous_standby()` also calls `_get_synchronous_commit_param()`method.

patroni/ha.py
```diff
@@ -447,47 +447,54 @@ def process_sync_replication(self):
...
                     logger.warning("No standbys available!")
 
                 logger.info("Assigning synchronous standby status to %s", picked)
                 self.state_handler.config.set_synchronous_standby(picked)
 
-                if picked and picked != '*' and not allow_promote:
-                    # Wait for PostgreSQL to enable synchronous mode and see if we can immediately set sync_standby
+                if picked and picked[0] != '*' and set(allow_promote) != set(picked):
+                    # Wait for PostgreSQL to enable synchronous mode and see if we can immediately
+                    # set sync_standby
```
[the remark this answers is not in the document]
ksarabu1: Thank you. We could possibly workaround this by comparing current set Vs picked & update the database ONLY if they don't match? 

```
          if set(current) != set(picked): 
                logger.info("Assigning synchronous standby status to %s", picked)
                self.state_handler.config.set_synchronous_standby(picked)
```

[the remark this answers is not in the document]
CyberDem0n: I thought about it, unfortunately with sync_mode_strict it will not work as expected.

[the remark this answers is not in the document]
ksarabu1: Got it. Thanks.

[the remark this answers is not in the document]
ksarabu1: Tested the following to workaround redundant db call. Please review.

```
                if (set(allow_promote) != set(picked)) or (set(allow_promote) == set(picked) and len(allow_promote) != sync_node_count):
                    logger.info("Assigning synchronous standby status to %s", picked)
                    self.state_handler.config.set_synchronous_standby(picked)
```

tests/test_postgresql.py
ksarabu1: fixed

avaczi:
Did some tests so far.  Here is a short description of it and a list of what was tested.

After merging the changes from #1641, locally, I run `docker build -t patroni . && docker-compose up` to start things up.  The different aspects are checked in the output, in the contents of DCS (via `patronictl edit-config`) and executing DML and statements on the leader.

Test cases:
- [ ] cluster starts up with leader and two async replicas when no configuration is applied
- [ ] cluster starts with leader and two synchronous replicas when `synchronous_mode = true` and `synchronous_node_count = 2`
- [ ] DML is blocked when a sync replica is not available (using `docker pause`), succeeds when the replica is back
- [ ] changing `synchronous_node_count` on the fly (`patronictl edit-config`) is reflected in the sync replica count and `SHOW synchronous_standby_names;`, pausing the the async replica doesn't block DML
- [ ] `synchronous_mode = off` and `synchronous_node_count = 2` still results in two async replicas

Jan-M:
What does this PR do if sync number is set to 2 in a 3 nodes cluster and 1 node leaves? Will it halt or reduce to 1?

CyberDem0n:
> Will it halt or reduce to 1?

It will reduce to 1.

avaczi:
Further test cases:
- [ ] after the crash of a sync standby when there are two, the cluster stays writable (i.e. it does not halt because of too few sync standbys)
- [ ] in a cluster with more nodes (4), it is possible to have 3 sync standbys

ksarabu1:
> * [ ]  after the crash of a sync standby when there are two, the cluster stays writable (i.e. it does not halt because of too few sync standbys)
Thanks for your help. Wondering if the test was run with `synchronous_mode_strict` enabled? Otherwise, Patroni downgrades the synchronous_standby_names.

avaczi:
@ksarabu1 at the time, those test were not done (they are only when the tickmark is present), but based on your question, I added more test cases.

avaczi:
One thing I noticed and not sure if it's indended to be like that: when there are N nodes with N-1 being synhronous standbys, and one of them crashes, there is a few log lines
```
demo-patroni3 | 2020-08-13 06:00:55,276 INFO: Assigning synchronous standby status to ['patroni2']
demo-patroni3 | server signaled
demo-patroni3 | 2020-08-13 06:00:55.286 UTC [25] LOG:  received SIGHUP, reloading configuration files
demo-patroni3 | 2020-08-13 06:00:55.287 UTC [25] LOG:  parameter "synchronous_standby_names" changed to "patroni2"
demo-patroni3 | 2020-08-13 06:00:55.297 UTC [34] LOG:  standby "patroni2" is now a synchronous standby with priority 1
```

avaczi:
:+1:

CyberDem0n:
:+1:

CyberDem0n:
Merged, thank you @ksarabu1!

PR #1599, opened 2020-06-23
replica & async rest API health check enhancement

ksarabu1:
**Description**: rest API health check enhancement on endpoints /replica & /async or /asynchronous to support additional query parameter `lag`. 

- ``GET /replica?lag=<max-lag>``: replica check endpoint. In addition to checks from ``replica``, it also checks replication latency and returns status code **200** only when the latency is below specified value (in bytes). The key leader_optime from DCS is used for Leader wal position and to compute latency on replica for performance reasons. Please note that the value in leader_optime might be couple of seconds old (based on loop_wait).

- ``GET /asynchronous?lag=<max-lag>`` or ``GET /async&lag=<max-lag>``: asynchronous standby check endpoint. In addition to checks from ``asynchronous`` or ``async``, it also checks replication latency and returns status code **200** only when the latency is below specified value (in bytes). The key leader_optime from DCS is used for Leader wal position and compute latency on replica for performance reasons. Please note that the value in leader_optime might be a couple of seconds old (based on loop_wait).

### DIFF
--- a/docs/rest_api.rst
+++ b/docs/rest_api.rst
@@ -19,6 +19,8 @@ For all health check ``GET`` requests Patroni returns a JSON document with the s
 
 - ``GET /replica``: replica health check endpoint. It returns HTTP status code **200** only when the Patroni node is in the state ``running``, the role is ``replica`` and ``noloadbalance`` tag is not set.
 
+- ``GET /replica&lag=<bytes>``: replica check endpoint. In addition to checks from ``replica``, it also checks replication latency and returns status code **200** only when it is below specified value. The key leader_optime from DCS is used for Leader wal position and compute latency on replica for performance reasons. Please note that the value in leader_optime might be couple of seconds old (based on loop_wait).
+
 - ``GET /read-only``: like the above endpoint, but also includes the primary.
 
 - ``GET /standby-leader``: returns HTTP status code **200** only when the Patroni node is running as the leader in a :ref:`standby cluster <standby_cluster>`.
@@ -26,6 +28,8 @@ For all health check ``GET`` requests Patroni returns a JSON document with the s
 - ``GET /synchronous`` or ``GET /sync``: returns HTTP status code **200** only when the Patroni node is running as a synchronous standby.
 
 - ``GET /asynchronous`` or ``GET /async``: returns HTTP status code **200** only when the Patroni node is running as an asynchronous standby.
+
+- ``GET /asynchronous&lag=<bytes>`` or ``GET /async&lag=<bytes>``: asynchronous standby check endpoint. In addition to checks from ``asynchronous`` or ``async``, it also checks replication latency and returns status code **200** only when it is below specified value. The key leader_optime from DCS is used for Leader wal position and compute latency on replica for performance reasons. Please note that the value in leader_optime might be couple of seconds old (based on loop_wait).
 
 - ``GET /health``: returns HTTP status code **200** only when PostgreSQL is up and running.
 
--- a/docs/rest_api.rst
+++ b/docs/rest_api.rst
@@ -19,7 +19,7 @@ For all health check ``GET`` requests Patroni returns a JSON document with the s
 
 - ``GET /replica``: replica health check endpoint. It returns HTTP status code **200** only when the Patroni node is in the state ``running``, the role is ``replica`` and ``noloadbalance`` tag is not set.
 
-- ``GET /replica&lag=<bytes>``: replica check endpoint. In addition to checks from ``replica``, it also checks replication latency and returns status code **200** only when it is below specified value. The key leader_optime from DCS is used for Leader wal position and compute latency on replica for performance reasons. Please note that the value in leader_optime might be couple of seconds old (based on loop_wait).
+- ``GET /replica&lag=<bytes>``: replica check endpoint. In addition to checks from ``replica``, it also checks replication latency and returns status code **200** only when it is below specified value. The key cluster.last_leader_operation from DCS is used for Leader wal position and compute latency on replica for performance reasons.
 
 - ``GET /read-only``: like the above endpoint, but also includes the primary.
 
@@ -29,7 +29,7 @@ For all health check ``GET`` requests Patroni returns a JSON document with the s
 
 - ``GET /asynchronous`` or ``GET /async``: returns HTTP status code **200** only when the Patroni node is running as an asynchronous standby.
 
-- ``GET /asynchronous&lag=<bytes>`` or ``GET /async&lag=<bytes>``: asynchronous standby check endpoint. In addition to checks from ``asynchronous`` or ``async``, it also checks replication latency and returns status code **200** only when it is below specified value. The key leader_optime from DCS is used for Leader wal position and compute latency on replica for performance reasons. Please note that the value in leader_optime might be couple of seconds old (based on loop_wait).
+- ``GET /asynchronous&lag=<bytes>`` or ``GET /async&lag=<bytes>``: asynchronous standby check endpoint. In addition to checks from ``asynchronous`` or ``async``, it also checks replication latency and returns status code **200** only when it is below specified value. The key cluster.last_leader_operation from DCS is used for Leader wal position and compute latency on replica for performance reasons.
 
 - ``GET /health``: returns HTTP status code **200** only when PostgreSQL is up and running.
 
--- a/docs/rest_api.rst
+++ b/docs/rest_api.rst
@@ -19,12 +19,12 @@ For all health check ``GET`` requests Patroni returns a JSON document with the s
 
 - ``GET /replica``: replica health check endpoint. It returns HTTP status code **200** only when the Patroni node is in the state ``running``, the role is ``replica`` and ``noloadbalance`` tag is not set.
 
-- ``GET /replica&lag=<max-lag>``: replica check endpoint. In addition to checks from ``replica``, it also checks replication latency and returns status code **200** only when it is below specified value. The key cluster.last_leader_operation from DCS is used for Leader wal position and compute latency on replica for performance reasons. max-lag can be specified in bytes (integer) or in human readable values, for e.g. 16kB, 64MB, 1GB.
+- ``GET /replica?lag=<max-lag>``: replica check endpoint. In addition to checks from ``replica``, it also checks replication latency and returns status code **200** only when it is below specified value. The key cluster.last_leader_operation from DCS is used for Leader wal position and compute latency on replica for performance reasons. max-lag can be specified in bytes (integer) or in human readable values, for e.g. 16kB, 64MB, 1GB.
 
-  - ``GET /replica&lag=1048576``
-  - ``GET /replica&lag=1024kB``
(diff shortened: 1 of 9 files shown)


patroni/api.py
```diff
@@ -89,8 +90,17 @@ def do_GET(self, write_status_code_only=False):
         patroni = self.server.patroni
         cluster = patroni.dcs.cluster
 
-        replica_status_code = 200 if not patroni.noloadbalance and \
-            response.get('role') == 'replica' and response.get('state') == 'running' else 503
+        replica_status_code = 503
+        if response.get('role') == 'replica' and response.get('state') == 'running' \
+           and not patroni.noloadbalance:
+            max_replica_lag = self.path_query.get('lag') and parse_int(self.path_query.get('lag')[0])
+            leader_optime = parse_int(patroni.dcs.leader_optime)
```
CyberDem0n, line 97: We should really avoid calling the DCS from REST API.
There is already `cluster.last_leader_operation` that contains the value from `/optime/leader`, the only minor problem that this value sometimes is outdated, but https://github.com/zalando/patroni/pull/1600 solves it.

ksarabu1: Agree. I have tried initially with cluster.last_leader_operation and also cluster.leader.member.data.xlog_location. Fixing last_leader_operation greatly simplifies the effort and improves performance. We can fix the code to use cluster.last_leader_operation. Thanks.

docs/rest_api.rst
```diff
@@ -19,6 +19,8 @@ For all health check ``GET`` requests Patroni returns a JSON document with the s
 
 - ``GET /replica``: replica health check endpoint. It returns HTTP status code **200** only when the Patroni node is in the state ``running``, the role is ``replica`` and ``noloadbalance`` tag is not set.
 
+- ``GET /replica&lag=<bytes>``: replica check endpoint. In addition to checks from ``replica``, it also checks replication latency and returns status code **200** only when it is below specified value. The key cluster.last_leader_operation from DCS is used for Leader wal position and compute latency on replica for performance reasons.
```
CyberDem0n, line 22: I think it has to be `GET /replica?lag=<max-lag>`.

patroni/api.py
```diff
@@ -89,8 +90,17 @@ def do_GET(self, write_status_code_only=False):
         patroni = self.server.patroni
         cluster = patroni.dcs.cluster
 
-        replica_status_code = 200 if not patroni.noloadbalance and \
-            response.get('role') == 'replica' and response.get('state') == 'running' else 503
+        replica_status_code = 503
+        if response.get('role') == 'replica' and response.get('state') == 'running' \
+           and not patroni.noloadbalance:
+            max_replica_lag = self.path_query.get('lag') and parse_int(self.path_query.get('lag')[0]) or 0
+            leader_optime = cluster and cluster.last_leader_operation and parse_int(cluster.last_leader_operation) or 0
```
CyberDem0n, line 97: `last_leader_operation` supposed to be int, no need to cast it.

patroni/api.py
```diff
@@ -89,8 +90,17 @@ def do_GET(self, write_status_code_only=False):
...
+            if max_replica_lag and leader_optime:
```
CyberDem0n, line 98: Since the `leader_optime` is cached, the "calculated" lag naturally quite often would be negative.
In theory someone might want to use `?lag=-100` in the health-check. I agree, that having the negative number makes little sense, but having `?lag=0` would be absolutely valid case.

Also I have a suggestion that we should always take lag into account and in case if it is not specified (or not valid) in the health-check assume that `lag = sys.maxsize`.

patroni/api.py
```diff
@@ -89,8 +90,17 @@ def do_GET(self, write_status_code_only=False):
         patroni = self.server.patroni
         cluster = patroni.dcs.cluster
 
-        replica_status_code = 200 if not patroni.noloadbalance and \
-            response.get('role') == 'replica' and response.get('state') == 'running' else 503
+        replica_status_code = 503
+        if response.get('role') == 'replica' and response.get('state') == 'running' \
+           and not patroni.noloadbalance:
+            max_replica_lag = self.path_query.get('lag') and parse_int(self.path_query.get('lag')[0]) or 0
```
CyberDem0n, line 96: The `parse_int()` function optionally accepts unit as a second argument. It allows doing very neat tricks like specifying lag not only in bytes, but in the arbitrary unit:
```python
In [2]: parse_int('16MB', 'B')
Out[2]: 16777216
```
If the value doesn't have the unit it would still work as previously:
```python
In [3]: parse_int('16777216', 'B')
Out[3]: 16777216
```

docs/rest_api.rst
CyberDem0n: If you implement suggestion from https://github.com/zalando/patroni/pull/1599/files#diff-790530624eaa21e5fb614fc321aef277R96 the `max-lag` could be specified as `16777216`, `16777216B`, `16MB`, `1GB`, and so on.

patroni/api.py
ksarabu1: Thanks for pointing it out. Will update.

patroni/api.py
ksarabu1: True. We should allow lag=0 or -ve lag health checks. Will update the code.

Not sure if we should make the lag health-check default. Agree, we could use default/max-lag if incorrect value is specified? Please let me know your thoughts. Thanks.

patroni/api.py
ksarabu1: will fix it.

patroni/api.py
```diff
@@ -89,8 +91,20 @@ def do_GET(self, write_status_code_only=False):
         patroni = self.server.patroni
         cluster = patroni.dcs.cluster
 
-        replica_status_code = 200 if not patroni.noloadbalance and \
-            response.get('role') == 'replica' and response.get('state') == 'running' else 503
+        replica_status_code = 503
+        if response.get('role') == 'replica' and response.get('state') == 'running' \
+           and not patroni.noloadbalance:
+            max_replica_lag = self.path_query.get('lag') and parse_int(self.path_query.get('lag')[0], 'B') \
+                                                         or sys.maxsize
```
CyberDem0n, line 98: It will make `max_replica_lag = sys.maxsize` if `?lag=0`.

patroni/api.py
```diff
@@ -89,8 +91,20 @@ def do_GET(self, write_status_code_only=False):
...
+            leader_optime = cluster and cluster.last_leader_operation or 0
+            replayed_location = response.get('xlog').get('replayed_location') or 0
+            current_lag = leader_optime - replayed_location
+            if current_lag >= 0:
+                max_replica_lag = max(max_replica_lag, 0)
```
CyberDem0n, line 103: That will set a negative `max_replica_lag` to 0. What will it bring us?

patroni/api.py
```diff
@@ -89,8 +91,20 @@ def do_GET(self, write_status_code_only=False):
...
 
-        replica_status_code = 200 if not patroni.noloadbalance and \
-            response.get('role') == 'replica' and response.get('state') == 'running' else 503
+        replica_status_code = 503
+        if response.get('role') == 'replica' and response.get('state') == 'running' \
+           and not patroni.noloadbalance:
+            max_replica_lag = self.path_query.get('lag') and parse_int(self.path_query.get('lag')[0], 'B') \
+                                                         or sys.maxsize
+            leader_optime = cluster and cluster.last_leader_operation or 0
+            replayed_location = response.get('xlog').get('replayed_location') or 0
```
CyberDem0n, line 100: `xlog` key in dict could be undefined.

tests/test_api.py
```diff
@@ -161,7 +161,10 @@ class TestRestApiHandler(unittest.TestCase):
     _authorization = '\nAuthorization: Basic dGVzdDp0ZXN0'
 
     def test_do_GET(self):
+        MockPatroni.dcs.cluster.last_leader_operation = 20
         MockRestApiServer(RestApiHandler, 'GET /replica')
+        MockRestApiServer(RestApiHandler, 'GET /replica?lag=10MB')
```
CyberDem0n, line 166:
```suggestion
        MockRestApiServer(RestApiHandler, 'GET /replica?lag=1M')
        MockRestApiServer(RestApiHandler, 'GET /replica?lag=10MB')
```

patroni/api.py
```diff
@@ -89,8 +91,20 @@ def do_GET(self, write_status_code_only=False):
...
+            current_lag = leader_optime - replayed_location
+            if current_lag >= 0:
+                max_replica_lag = max(max_replica_lag, 0)
+            if leader_optime:
+                replica_status_code = 200 if max_replica_lag >= current_lag else 503
+            else:
+                replica_status_code = 200
```
CyberDem0n, line 107:
```suggestion
        leader_optime = cluster and cluster.last_leader_operation or 0                                                                  
        replayed_location = response.get('xlog', {}).get('replayed_location', 0)                                                        
        max_replica_lag = parse_int(self.path_query.get('lag', [sys.maxsize])[0], 'B')                                                  
        if max_replica_lag is None:
            max_replica_lag = sys.maxsize 
        is_lagging = leader_optime and leader_optime > replayed_location + max_replica_lag                                              
        
        replica_status_code = 200 if not patroni.noloadbalance and not is_lagging and \
            response.get('role') == 'replica' and response.get('state') == 'running' else 503                                           
```

docs/rest_api.rst
```diff
@@ -19,6 +19,13 @@ For all health check ``GET`` requests Patroni returns a JSON document with the s
 
 - ``GET /replica``: replica health check endpoint. It returns HTTP status code **200** only when the Patroni node is in the state ``running``, the role is ``replica`` and ``noloadbalance`` tag is not set.
 
+- ``GET /replica&lag=<max-lag>``: replica check endpoint. In addition to checks from ``replica``, it also checks replication latency and returns status code **200** only when it is below specified value. The key cluster.last_leader_operation from DCS is used for Leader wal position and compute latency on replica for performance reasons. max-lag can be specified in bytes (integer) or in human readable values, for e.g. 16kB, 64MB, 1GB.
+
+  - ``GET /replica&lag=1048576``
+  - ``GET /replica&lag=1024kB``
+  - ``GET /replica&lag=10MB``
+  - ``GET /replica&lag=1GB``
```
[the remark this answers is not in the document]
ksarabu1: oops! will fix it.

patroni/api.py
ksarabu1: true. will fix it

patroni/api.py
ksarabu1: will fix it. assumed that it always exists.

patroni/api.py
ksarabu1: you suggested change includes the fix. thanks

patroni/api.py
ksarabu1: you suggested commit already includes the fix. thanks

FxKu:
:+1:

CyberDem0n:
:+1:

CyberDem0n:
Merged, thank you @ksarabu1!

PR #1604, opened 2020-07-07
PATRONI_ETCD_HOSTS should be wrapped with double quotes

sieginglion:
Fixes #
PATRONI_ETCD_HOSTS should be wrapped with double quotes


docker/patroni.env
```diff
@@ -1,5 +1,5 @@
 PATRONI_SCOPE=demo
-PATRONI_ETCD_HOSTS='etcd1:2379','etcd2:2379','etcd3:2379'
+PATRONI_ETCD_HOSTS="'etcd1:2379','etcd2:2379','etcd3:2379'"
```
CyberDem0n, line 2: It doesn't cause any problems without double-quotes for me, but with quotes it actually fails:
```python
$ docker logs demo-patroni1 
Process Process-1:
  File "/usr/lib/python3.5/multiprocessing/process.py", line 249, in _bootstrap
    self.run()
  File "/usr/lib/python3.5/multiprocessing/process.py", line 93, in run
    self._target(*self._args, **self._kwargs)
  File "/patroni/__init__.py", line 197, in patroni_main
    patroni = Patroni(conf)
  File "/patroni/__init__.py", line 32, in __init__
    self.dcs = get_dcs(self.config)
  File "/patroni/dcs/__init__.py", line 92, in get_dcs
    return item(config[name])
  File "/patroni/dcs/etcd.py", line 399, in __init__
    self._client = self.get_etcd_client(config)
  File "/patroni/dcs/etcd.py", line 454, in get_etcd_client
    config['hosts'].append(uri(protocol, split_host_port(value.strip(), default_port)))
  File "/patroni/utils.py", line 363, in split_host_port
    return t[0], int(t[1])
ValueError: invalid literal for int() with base 10: "2379'"
```

PR #1609, opened 2020-07-08
Update postgres0.yml

krssanthosh:
#        hot_standby: "on" should not be under parameters.

### DIFF
--- a/postgres0.yml
+++ b/postgres0.yml
@@ -40,6 +40,7 @@ bootstrap:
     maximum_lag_on_failover: 1048576
 #    master_start_timeout: 300
 #    synchronous_mode: false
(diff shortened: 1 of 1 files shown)


CyberDem0n:
What made you think so?

krssanthosh:
I think, my configuration did not work. After removing it from parameters and adding it under bootstrad/dcs, it worked for me, only to realize later that it works with parameters also. Unfortunately while setting sycn replication cluster, my linux crashed and could not verify. Please ignore this PR and if you have any genuine reference document about the possible combinations of patroni configurations for a synchronous Postgres cluster with hot standby, it would indeed be a great help to me.

PR #1614, opened 2020-07-17
Make it possible to bypass kubernetes service

CyberDem0n:
When running on K8s Patroni is communicating with API via the `kubernetes` service, which is address is exposed via the
`KUBERNETES_SERVICE_HOST` environment variable. Like any other service, the `kubernetes` service is handled by `kube-proxy`, that depending on configuration is either relying on userspace program or `iptables` for traffic routing.

During K8s upgrade, when master nodes are replaced, it is possible that `kube-proxy` doesn't update the service configuration in time and as a result Patroni fails to update the leader lock and demotes postgres.

In order to improve the user experience and get more control on the problem we make it possible to bypass the `kubernetes` service and connect directly to API nodes.
The strategy is very simple:
1. Resolve list IPs of API nodes from the kubernetes endpoint on every iteration of HA loop.
2. Stick to one of these IPs for API requests
3. Switch to a different IP if connected to IP is not from the list
4. If the request fails, switch to another IP and retry

Such a strategy is already used for Etcd and proven to work quite well.

In order to enable the feature, you need either to set to `true` `kubernetes.bypass_api_service` in the Patroni configuration file or `PATRONI_KUBERNETES_BYPASS_API_SERVICE` environment variable.


docs/ENVIRONMENT.rst
```diff
@@ -79,6 +79,7 @@ Exhibitor
 
 Kubernetes
 ----------
+-  **PATRONI\_KUBERNETES\_BYPASS\_API\_SERVICE**: (optional) When communicating with Kubernetes API Patroni is usually relying on `kubernetes` service, which address is exposed via the `KUBERNETES_SERVICE_HOST` environment variable. If the `PATRONI_KUBERNETES_BYPASS_API_SERVICE` parameter is set to ``true``, Patroni will resolve the list of API nodes behind the service and connect directly to them.
```
avaczi, line 82:
```suggestion
-  **PATRONI\_KUBERNETES\_BYPASS\_API\_SERVICE**: (optional) When communicating with the Kubernetes API, Patroni is usually relying on the `kubernetes` service, the address of which is exposed in the pods via the `KUBERNETES_SERVICE_HOST` environment variable. If `PATRONI_KUBERNETES_BYPASS_API_SERVICE` is set to ``true``, Patroni will resolve the list of API nodes behind the service and connect directly to them.
```

docs/SETTINGS.rst
```diff
@@ -136,6 +136,7 @@ Exhibitor
 
 Kubernetes
 ----------
+-  **bypass\_api\_service**: (optional) When communicating with Kubernetes API Patroni is usually relying on `kubernetes` service, which address is exposed via the `KUBERNETES_SERVICE_HOST` environment variable. If the `bypass_api_service` parameter is set to ``true``, Patroni will resolve the list of API nodes behind the service and connect directly to them.
```
avaczi, line 139:
```suggestion
-  **bypass\_api\_service**: (optional) When communicating with the Kubernetes API, Patroni is usually relying on the `kubernetes` service, the address of which is exposed in the pods via the `KUBERNETES_SERVICE_HOST` environment variable. If `bypass\_api\_service` is set to ``true``, Patroni will resolve the list of API nodes behind the service and connect directly to them.
```

kubernetes/openshift-example/templates/template_patroni_persistent.yaml
```diff
@@ -130,6 +130,8 @@ objects:
               fieldRef:
                 apiVersion: v1
                 fieldPath: metadata.namespace
+          - name: PATRONI_KUBERNETES_BYPASS_API_SERVICE
+            value 'true'
```
FxKu, line 134:
```suggestion
            value: 'true'
```

FxKu:
:+1:

CyberDem0n:
:+1:

Issue #1617, opened 2020-07-21
standby_cluster is considered as replica


bukem:
Hi,
I have a cluster with Leader and 2 replicas. I want to add standby node to this deployment, so I prepared the following config before starting patroni on the 4th node:
```
scope: testcluster
name: stdb-node-345032702

restapi:
  listen: 0.0.0.0:8008
  connect_address: 10.38.48.138:8008

consul:
  register_service: true
  host: 127.0.0.1:8500
  checks: []

bootstrap:
  dcs:
    standby_cluster:
      host: master.testcluster.service.dc.consul
      post: 5432

  users:
    replicator:
      password: passwordhere
      options:
        - replication

postgresql:
...

log:
    format: '{"time": "%(asctime)s", "name": "%(name)s", "process": %(process)d, "thread": %(thread)d, "level": "%(levelname)s", "message": "%(message)s"}'

```

But when I start patroni, the node joins as `replica`, while I expect it to be `standby_leader`:
```
Jul 21 09:58:33 stdb-node-345032702 patroni[29177]: {"time": "2020-07-21 09:58:33,886", "name": "patroni.dcs.consul", "process": 29177, "thread": 139955463599936, "level": "INFO", "message": "Register service testcluster, params {'service_id': 'testcluster/stdb-node-345032702', 'address': '10.38.48.226', 'port': 5432, 'check': {'http': 'http://10.38.48.138:8008/replica', 'interval': '5s', 'DeregisterCriticalServiceAfter': '150.0s'}, 'tags': ['replica']}"}
```

Config from leader node:
```
scope: testcluster
name: patroni-cluster-345032702

restapi:
  listen: 0.0.0.0:8008
  connect_address: 10.38.48.226:8008

consul:
  register_service: true
  host: 127.0.0.1:8500
  checks: []

bootstrap:
  dcs:
    ttl: 30
    loop_wait: 10
    retry_timeout: 10
    maximum_lag_on_failover: 1048576
    synchronous_mode: true
    postgresql:
      use_pg_rewind: true
      use_slots: false
      parameters:
        archive_mode: "on"
...
        work_mem: "8MB"
      recovery_conf:
        restore_command: /bin/true
  initdb:
    - encoding: UTF8
    - data-checksums
    - locale: en_US.UTF-8
    - lc-collate: en_US.UTF-8
    - lc-ctype: en_US.UTF-8

  users:
    replicator:
      password: passwordhere
      options:
        - replication

postgresql:
...

watchdog:
[...]

PR #1623, opened 2020-07-22
A few little bug-fixes

CyberDem0n:
1. Put the `*` character into pgpass if the actual value is empty
2. Re-raise fatal exception from the HA loop (we need to exit if for example cluster initialization failed)

Close https://github.com/zalando/patroni/issues/1617


FxKu:
:+1:

CyberDem0n:
:+1:

PR #1625, opened 2020-07-23
Add consul service_tags configuration field

Legogris:
This is useful for dynamic service discovery, for example by load balancers.

### DIFF
--- a/docs/SETTINGS.rst
+++ b/docs/SETTINGS.rst
@@ -103,6 +103,7 @@ Most of the parameters are optional, but you have to specify one of the **host**
 -  **consistency**: (optional) Select consul consistency mode. Possible values are ``default``, ``consistent``, or ``stale`` (more details in `consul API reference <https://www.consul.io/api/features/consistency.html/>`__)
 -  **checks**: (optional) list of Consul health checks used for the session. By default an empty list is used.
 -  **register\_service**: (optional) whether or not to register a service with the name defined by the scope parameter and the tag master, replica or standby-leader depending on the node's role. Defaults to **false**.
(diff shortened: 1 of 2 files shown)


docs/SETTINGS.rst
```diff
@@ -103,6 +103,7 @@ Most of the parameters are optional, but you have to specify one of the **host**
 -  **consistency**: (optional) Select consul consistency mode. Possible values are ``default``, ``consistent``, or ``stale`` (more details in `consul API reference <https://www.consul.io/api/features/consistency.html/>`__)
 -  **checks**: (optional) list of Consul health checks used for the session. By default an empty list is used.
 -  **register\_service**: (optional) whether or not to register a service with the name defined by the scope parameter and the tag master, replica or standby-leader depending on the node's role. Defaults to **false**.
+-  **service\_tags**: (optional) additional static tags to add to the Consul service apart from the role (``master``/``replica``/``standby-leader``).  By default an empty list is used.
```
CyberDem0n, line 106: IIRC if there is more than one tag it will become impossible to rely on DNS to find the primary/replicas.

Legogris: I don't think so actually - at least it works fine for me for other services. Also note that this is optional, and defaulting to an empty list,

Legogris:
Just pushed update for failing CI (didn't realize this was targeting *both* python2 and python3)

Legogris:
FYI been running with this for 3-4 days with no issues so far

avaczi:
:+1:

CyberDem0n:
:+1:

CyberDem0n:
Merged, thank you @Legogris!

PR #1631, opened 2020-07-30
Move WAL and tablespaces after a failed init

feikesteenbergen:
For init processes that use a symlinked WAL directory, or use custom
scripts that create new tablespaces, these directories should also be
renamed after a failed init attempt, as currently the following errors
occur if the first init attempt failed, but a second one might succeed:

      fixing permissions on existing directory /var/lib/postgresql/data ... ok
      initdb: error: directory "/var/lib/postgresql/wal/pg_wal" exists but is not empty
      [...]
      File "/usr/lib/python3/dist-packages/patroni/ha.py", line 1173, in post_bootstrap
        self.cancel_initialization()
      File "/usr/lib/python3/dist-packages/patroni/ha.py", line 1168, in cancel_initialization
        raise PatroniException('Failed to bootstrap cluster')
      patroni.exceptions.PatroniException: 'Failed to bootstrap cluster'

In the remove_data_directory function the same happens for removing the
data directory, it seems the same kind of thing should also happen when
moving a data directory.

### DIFF
--- a/patroni/postgresql/__init__.py
+++ b/patroni/postgresql/__init__.py
@@ -850,10 +850,51 @@ class Postgresql(object):
         self.config.setup_server_parameters()
         return True
 
+    def pg_wal_realpath(self):
+        """Returns a dict containing the symlink (key) and target (value) for the wal directory"""
+        links = {}
+        for pg_wal_dir in ('pg_xlog', 'pg_wal'):
+            pg_wal_path = os.path.join(self._data_dir, pg_wal_dir)
+            if os.path.exists(pg_wal_path) and os.path.islink(pg_wal_path):
+                pg_wal_realpath = os.path.realpath(pg_wal_path)
+                links[pg_wal_path] = pg_wal_realpath
+        return links
+
+    def pg_tblspc_realpaths(self):
+        """Returns a dict containing the symlink (key) and target (values) for the tablespaces"""
+        links = {}
+        pg_tblsp_dir = os.path.join(self._data_dir, 'pg_tblspc')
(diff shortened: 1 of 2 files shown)


feikesteenbergen:
I still need to test this, however it would be interesting to get some feedback on whether or not this should be in here.

patroni/postgresql/__init__.py
```diff
@@ -853,7 +853,28 @@ def configure_server_parameters(self):
...
+                postfix = time.strftime('%Y-%m-%d-%H-%M-%S')
+                # let's see if pg_xlog|pg_wal is a symlink, in this case we
+                # should move the target
+                for pg_wal_dir in ('pg_xlog', 'pg_wal'):
+                    pg_wal_path = os.path.join(self._data_dir, pg_wal_dir)
+                    if os.path.exists(pg_wal_path) and os.path.islink(pg_wal_path):
+                        pg_wal_realpath = os.path.realpath(pg_wal_path)
+                        logger.info('renaming WAL directory: %s', pg_wal_realpath)
+                        new_name = '{0}_{1}'.format(pg_wal_realpath, postfix)
+                        os.rename(pg_wal_realpath, new_name)
```
CyberDem0n, line 865: I think it would be great not only to rename, but update the source of symlink.
The same applies to tablespaces.

RafiaSabih:
:+1:

CyberDem0n:
:+1:

CyberDem0n:
Merged, thank you @feikesteenbergen!

PR #1639, opened 2020-08-07
Explicitly disallow patching non-existent config

CyberDem0n:
For DCS other than `kubernetes` it was failing with exception due to the `cluster.config` being `None`, but on Kubernetes it was happily creating the config annotation and preventing writing bootstrap configuration after the bootstrap finished.


CyberDem0n:
:+1:

CyberDem0n:
:+1:

avaczi:
:+1:

PR #1641, opened 2020-08-11
Refactor docker-compose.yml for better compatibility with new version

CyberDem0n:
The newest versions of docker-compose want to have some values double-quoted in the env file while old versions failing to process such files.
The solution is simple, move some of the parameters to the `docker-compose.yml` and rely on anchors for inheritance.
Since the main idea behind env files was to keep "secret" information off the main YAML we also get rid of any non-secret stuff, mainly located in the etcd.env.


avaczi:
:+1:

CyberDem0n:
:+1:

PR #1654, opened 2020-08-17
Compatibility with PostgreSQL 13

CyberDem0n:
So far Patroni was enforcing the same value of `wal_keep_segments` on all nodes in the cluster. If the parameter was missing from the global configuration it was using the default value `8`.
In pg13 beta3 the `wal_keep_segments` was renamed to the `wal_keep_size` and it broke Patroni.

If `wal_keep_segments` happened to be present in the configuration for pg13, Paroni will recalculate the value to `wal_keep_size` assuming that the `wal_segment_size` is 16MB. Sure, it is possible to get the real value of `wal_segment_size` from pg_control, but since we are dealing with the case of misconfiguration it is not worse time spend on it.

### DIFF
--- a/docs/dynamic_configuration.rst
+++ b/docs/dynamic_configuration.rst
@@ -35,6 +35,7 @@ For the parameters below, PostgreSQL does not require equal values among the mas
 - max_wal_senders: 5
 - max_replication_slots: 5
 - wal_keep_segments: 8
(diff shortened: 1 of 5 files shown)


Jan-M:
:+1:

CyberDem0n:
:+1:

PR #1659, opened 2020-08-20
BUGFIX: pause on K8s

CyberDem0n:
On K8s the `Cluster.leader` is a valid object even if the cluster has no leader because we need to know the `resourceVersion` for future CAS operation. Such a non-empty object broke HA loop and made other nodes to think that the leader is there.

The right way to identify the missing leader which reliably works across all DCS is checking that the leader's name is empty.


avaczi:
:+1:

CyberDem0n:
:+1:

Issue #1669, opened 2020-08-27
local patroni config with a directory of yml files rather than a single file


fvannee:
What's your opinion on adding a feature where instead of specifying a single local configuration file, you can specify a directory containing multiple .yml files, which will all be loaded and combined. 
The main advantage this brings is to allow some common configuration files to be identical across all hosts in a cluster, while overriding specific settings in a separate file. Currently, it's often the case that 95% of the `patroni.yml` file is identical across hosts in a cluster, but there are a few things that may differ. It would be great in deployments if we could make this clear on a file-level with a layered structure.

For example, start Patroni by passing `config` directory. This `config` directory has files:
- `config/00-base.yml`
- `config/01-specific.yml`

In my example, everything in `00-base.yml` would be identical for all hosts in a cluster. `01-specific.yml` could contain some specific stuff like `tags` or the `raft.self_addr` attribute.
My proposal would be to modify `_load_config_file` to loop over all yml files in the directory in alphabetical order and run `patch_config` each time updating the current configuration. There are a couple of other places that need updating, but this is the rough idea.

I'd be happy to give this a try with a PR if it's a feature you'd be willing to accept.

CyberDem0n:
IMO this is more a task of the configuration management system.

fvannee:
In general I agree, but sometimes you have to deal with whatever configuration management system that's already in use, and it would be nice if Patroni could cater to some of these cases too.

CyberDem0n:
Since if you volunteer to implement the feature I am fine with it.

PR #1671, opened 2020-08-28
Add support for configuration directories (#1669)

fvannee:
It is now also possible to point the configuration path to a directory instead of a file.
Patroni will find all yml files in the directory and apply them in sorted order

### DIFF
--- a/patroni/config.py
+++ b/patroni/config.py
@@ -92,7 +92,7 @@ class Config(object):
         self.__environment_configuration = self._build_environment_configuration()
 
         # Patroni reads the configuration from the command-line argument if it exists, otherwise from the environment
-        self._config_file = configfile and os.path.isfile(configfile) and configfile
+        self._config_file = configfile and os.path.exists(configfile) and configfile
         if self._config_file:
             self._local_configuration = self._load_config_file()
         else:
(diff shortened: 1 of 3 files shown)


patroni/config.py
```diff
@@ -120,12 +120,32 @@ def dynamic_configuration(self):
...
+    def _load_config_path(self, path):
+        """
+        If path is a file, loads the yml file pointed to by path.
+        If path is a directory, loads all yml files in that directory in alphabetical order
+        """
+        if os.path.isfile(path):
+            files = [path]
+        elif os.path.isdir(path):
+            files = [f for f in sorted(os.listdir(path))
+                     if (f.endswith('.yml') or f.endswith('.yaml')) and os.path.isfile(f)]
```
CyberDem0n, line 132: `os.listdir` returns names that are relative to the `path`.

patroni/config.py
```diff
@@ -120,12 +120,32 @@ def dynamic_configuration(self):
...
+        else:
+            logger.error('config path %s is neither directory nor file', path)
+            raise ConfigParseError('invalid config path')
```
CyberDem0n, line 135: Could you please also add the test-case for these lines?

CyberDem0n (changes requested):

patroni/config.py
fvannee: Technically this is unreachable because of the `os.path.exists(configfile)` check in `__init__`. I don't think there are any other possibilities for an existing path. I kept it there for safety in case something changes to the code in the future that makes it suddenly reachable. Would you rather that I remove the whole else part?

CyberDem0n: It is perfectly reachable:
```python
In [1]: import os

In [2]: os.path.isfile('data/postgresql0/.s.PGSQL.5432')
Out[2]: False

In [3]: os.path.isdir('data/postgresql0/.s.PGSQL.5432')
Out[3]: False
```

`data/postgresql0/.s.PGSQL.5432` is neither file nor directory, but unix-socket.

CyberDem0n (changes requested):
Now unit-tests are failing.

patroni/config.py
fvannee: Ah, I didn't know that. Ok, I've added a test case that asserts it throws when an existing path is neither a file nor a directory

fvannee:
> Now unit-tests are failing.

Apologies for the messy PR.. It was a bit too hastily created. I've fixed it.

CyberDem0n (approved):

Jan-M:
:+1:

CyberDem0n:
:+1:

CyberDem0n:
Merged, thank you @fvannee

PR #1674, opened 2020-08-31
Advanced validation of PostgreSQL parameters

CyberDem0n:
So far Patroni was performing a comparison of the old value (in the `pg_settings`) with the new value (from Patroni configuration or from DCS) in order to figure out if reload or restart is required when the parameter has been changed. If the given parameter was missing in the `pg_settings` Patroni was ignoring it and not writing into the `postgresql.conf`.

In case if Postgres is not running, no validation has been performed and parameters and values were written into the config as it is.

It is not a very common mistake, but people tend to mistype parameter names or values.
Also, it happens that some parameters are removed in specific Postgres versions and some new are added (e.g. `checkpoint_segments` replaced with `min_wal_size` and `max_wal_size` in 9.5 or` wal_keep_segments` was replaced with `wal_keep_size` in 13).

Writing nonexistent parameters or invalid values into the `postgresql.conf` makes postgres unstartable.
This change doesn't solve the issue 100%, but at least approaching this goal very close.

### DIFF
--- a/patroni/postgresql/config.py
+++ b/patroni/postgresql/config.py
@@ -6,11 +6,12 @@ import socket
 import stat
 import time
 
-from patroni.exceptions import PatroniFatalException
 from six.moves.urllib_parse import urlparse, parse_qsl, unquote
-from urllib3.response import HTTPHeaderDict
 
+from .validator import CaseInsensitiveDict, recovery_parameters,\
+        transform_postgresql_parameter_value, transform_recovery_parameter_value
 from ..dcs import slot_name_from_member_name, RemoteMember
+from ..exceptions import PatroniFatalException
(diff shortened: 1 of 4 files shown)


bradnicholson:
Is this going to be compatible with optional paramaters, such as `pg_stat_statements.*` parameters?

CyberDem0n:
This implementation ignores any parameter with the period in its name: https://github.com/zalando/patroni/pull/1674/files#diff-70e16ad3b2b67d9c00e0da1966cb8e30R484
In future it shouldn't be hard to extend the list and support contribs from the core.

erthalion:
:+1:

erthalion:
:+1:

CyberDem0n:
:+1:

Issue #1684, opened 2020-09-09
Unexpected exception when Patroni disconnected from DCS


le-duane:
Thanks so much for the latest release! When testing it out in our environment, I encountered the following error.

**Describe the bug**
When disconnected from DCS, Patroni hits an unexpected exception and does not seem to demote itself after exceeding retry_timeout. This did not result in split-brain, but did leave the former primary ahead of the synchronous replica when it promoted itself.

**To Reproduce**
Null route kube API service
`ip route add blackhole 172.20.0.1` on primary 

**Expected behavior**
Expected the primary to "demoted self because failed to update leader lock in DCS"

**Environment**
- Patroni version: 2.0.0
- PostgreSQL version: 11.9
- DCS (and its version): Kubernetes v1.15.11

**Patroni configuration file**
[dev-patroni-config.txt](https://github.com/zalando/patroni/files/5193580/dev-patroni-config.txt)

**patronictl show-config**
```
loop_wait: 10
master_start_timeout: 300
master_stop_timeout: 135
maximum_lag_on_failover: 0
postgresql:
  parameters:
    max_connections: 90
    max_locks_per_transaction: 64
    max_prepared_transactions: 0
    max_replication_slots: 10
    max_wal_senders: 10
    max_worker_processes: 8
    track_commit_timestamp: 'off'
    wal_keep_segments: 200
    wal_level: logical
    wal_log_hints: 'off'
  use_pg_rewind: false
  use_slots: false
retry_timeout: 20
synchronous_mode: true
ttl: 50
```
**Patroni logs?**
[primary.log](https://github.com/zalando/patroni/files/5193566/primary.log)
[standby.log](https://github.com/zalando/patroni/files/5193574/standby.log)

CyberDem0n:
Yes, it is clearly a bug:
```
2020-09-09 06:51:43,288 INFO: Unexpected exception raised, please report it as a BUG
```

Will prepare a PR fixing it ASAP.

CyberDem0n:
@le-duane please check. It should be enough only to change the https://github.com/zalando/patroni/pull/1685/files#diff-87299faee0431ddce0848d6ff7726df7R904

CyberDem0n:
BTW, Patroni doesn't allow to disable `wal_log_hints`, they are required for pg_rewind to work.
If you have enabled data_checksums (which is highly recommended), they produce the same effect on WAL as wal_log_hints.

[...]

PR #1685, opened 2020-09-09
Fix bug in kubernetes.update_leader

CyberDem0n:
Unhandled exception prevented demoting the primary.
In addition to that wrap the update_leader call in the HA loop into try..except block and implement a test case.

Fixes https://github.com/zalando/patroni/issues/1684


avaczi:
:+1:

CyberDem0n:
:+1:

PR #1687, opened 2020-09-10
Change the behavior in pause

CyberDem0n:
1. Don't call bootstrap if PGDATA is missing/empty, because it might be for purpose, and someone/something working on it.
2. Consider postgres running as a leader in pause not healthy if pg_control sysid doesn't match with the /initialize key (empty initialize key will allow the "race" and the leader will "restore" initialize key).
3. Don't exit on sysid mismatch in pause, only log a warning.
4. Cover corner cases when Patroni started in pause with empty PGDATA and it was restored by somebody else
5. Empty string is a valid `recovery_target`.

### DIFF
--- a/patroni/ha.py
+++ b/patroni/ha.py
@@ -762,9 +762,9 @@ class Ha(object):
             return False
 
         if self.state_handler.is_leader():
-            # in pause leader is the healthiest only when sysid matches with initialize!
-            return not self.is_paused() or self.sysid_valid(self.cluster.initialize)\
-                    and self.state_handler.sysid == self.cluster.initialize
+            # in pause leader is the healthiest only when no initialize or sysid matches with initialize!
+            return not self.is_paused() or not self.cluster.initialize\
+                    or self.state_handler.sysid == self.cluster.initialize
 
         if self.is_paused():
             return False
--- a/patroni/ha.py
+++ b/patroni/ha.py
@@ -761,8 +761,10 @@ class Ha(object):
         if self.state_handler.is_starting():  # postgresql still starting up is unhealthy
             return False
 
-        if self.state_handler.is_leader():  # leader is always the healthiest
-            return True
+        if self.state_handler.is_leader():
+            # in pause leader is the healthiest only when sysid matches with initialize!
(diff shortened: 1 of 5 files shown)


patroni/ha.py
```diff
@@ -1359,9 +1363,16 @@ def _run_cycle(self):
 
                 if self.sysid_valid(self.cluster.initialize):
                     if self.cluster.initialize != data_sysid:
-                        logger.fatal("system ID mismatch, node %s belongs to a different cluster: %s != %s",
-                                     self.state_handler.name, self.cluster.initialize, data_sysid)
-                        sys.exit(1)
+                        if self.is_paused():
+                            logger.warning('system ID has changed while in paused mode. Patroni will exit when resuming'
+                                           ' unless system ID is reset: %s != %s', self.cluster.initialize, data_sysid)
```
sdudoladov, line 1368: Should system id or  the `/initialize` key be reset ? Is resetting the `/initialize` in DCS to the new system ID sufficient to prevent Patroni from exiting on resume ?

sdudoladov (approved):
> Don't exit on sysid mismatch in pause, only log a warning.

... and release the leader lock  judging from the code

> Empty string is a valid recovery_target.

I assume this change affects only the validator, doesn't it ?

sdudoladov:
:+1:

patroni/ha.py
```diff
@@ -1348,6 +1350,8 @@ def _run_cycle(self):
                     self.release_leader_key_voluntarily()
                     return 'released leader key voluntarily as data dir empty and currently leader'
 
+                if self.is_paused():
```
avaczi, line 1353:
```suggestion
                # don't try to bootstrap in pause mode
                if self.is_paused():
```

patroni/ha.py
avaczi: I think when there is another node, resetting the `/initialize` key would break that one

CyberDem0n: Patroni should not reset or change the valid initialize key. It is used as a safety measure to prevent the wrong node from joining the current cluster.

CyberDem0n:
:+1:

Issue #1717, opened 2020-09-28
regression test failure in 2.0.0


mbanck:
**Describe the bug**
We're seeing this regression test failure on the apt.postgresql.org autobuilders for 2.0.0:

```
11:27:58 ==================================== ERRORS ====================================
11:27:58 ______________________ ERROR collecting tests/test_ha.py _______________________
11:27:58 /usr/lib/python3.7/doctest.py:932: in find
11:27:58     self._find(tests, obj, name, module, source_lines, globs, {})
11:27:58 /usr/lib/python3.7/doctest.py:991: in _find
11:27:58     if ((inspect.isroutine(inspect.unwrap(val))
11:27:58 /usr/lib/python3.7/inspect.py:515: in unwrap
11:27:58     raise ValueError('wrapper loop when unwrapping {!r}'.format(f))
11:27:58 E   ValueError: wrapper loop when unwrapping call
```

This happens on Debian buster (`python3_3.7.3-1`) and Ubuntu bionic (`python3_3.6.7-1~18.04`), but not Debian unstable (`python3_3.8.2-3`).

**Expected behavior**

No regression test failures.

**Environment**
- Patroni version: 2.0.0
- PostgreSQL version: all support PGDG versions
- DCS (and its version): n/a

Issue #1709, opened 2020-09-21
The 'ipaddress' distribution was not found and is required by urllib3


seenutn:
This is not an patroni issue but I encountered while running patroni 2 on centos 8.2 and hence posting it here.

Here is the error I am encountering:

[root@Maindb-rep ~]# patroni --version
Traceback (most recent call last):
  File "/usr/bin/patroni", line 6, in <module>
    from pkg_resources import load_entry_point
  File "/usr/lib/python3.6/site-packages/pkg_resources/__init__.py", line 3095, in <module>
    @_call_aside
  File "/usr/lib/python3.6/site-packages/pkg_resources/__init__.py", line 3079, in _call_aside
    f(*args, **kwargs)
  File "/usr/lib/python3.6/site-packages/pkg_resources/__init__.py", line 3108, in _initialize_master_working_set
    working_set = WorkingSet._build_master()
  File "/usr/lib/python3.6/site-packages/pkg_resources/__init__.py", line 570, in _build_master
    ws.require(__requires__)
  File "/usr/lib/python3.6/site-packages/pkg_resources/__init__.py", line 888, in require
    needed = self.resolve(parse_requirements(requirements))
  File "/usr/lib/python3.6/site-packages/pkg_resources/__init__.py", line 774, in resolve
    raise DistributionNotFound(req, requirers)
pkg_resources.DistributionNotFound: The 'ipaddress' distribution was not found and is required by urllib3

[root@Maindb-rep ~]# python3 --version
Python 3.6.8

[root@Maindb-rep ~]# locate ipaddress.py
/usr/lib/python3.6/site-packages/pip/_vendor/ipaddress.py
/usr/lib64/python3.6/ipaddress.py
/usr/lib64/python3.6/site-packages/psycopg2/_ipaddress.py
[root@Maindb-rep ~]#

How can I fix this error?

Regards,
Seenu.

CyberDem0n:
> How can I fix this error?

How was the Patroni installed? Judging from ` File "/usr/bin/patroni", line 6, in ` you haven't used pip.

seenutn:
Using yum.  

[root@Maindb-rep ~]# rpm -qfi /usr/bin/patroni
Name        : patroni
Version     : 2.0.0
Release     : 2.rhel8
Architecture: x86_64
Install Date: Sunday 20 September 2020 10:38:40 PM IST
Group       : Unspecified
Size        : 2090358
License     : MIT
[...]

PR #1718, opened 2020-09-28
Workaround unittest bug and fix requirements

CyberDem0n:
* unittest bug: https://bugs.python.org/issue25532
* `urllib3[secure]` wrongly depends on `ipaddress` for python3, while in fact we don't need all dependencies of the `secure` extra, but only `ipaddress` for the `kubernetes` on python2.7 

Close https://github.com/zalando/patroni/issues/1717
Close https://github.com/zalando/patroni/issues/1709


Jan-M:
:+1:

CyberDem0n:
:+1:

PR #1726, opened 2020-10-05
Don't interrupt restart or promote if lost leader lock in pause

CyberDem0n:
In pause it is allowed to run postgres as master without lock.


sdudoladov:
:+1:

CyberDem0n:
:+1:

PR #1736, opened 2020-10-20
[fix] add database name into connection string for rewinding

mizhka:
Hi,

During configuration of standby cluster, we faced issue with error "psycopg2.OperationalError: FATAL:  database "patroni" does not exist". It's reproduced if username differs from database name (ex. user for rewind - patroni, database - postgres). Root cause is missing / empty database parameter in connection string of remote leader. The fix is simple: just to add "database" in connection parameters (parameter is present in DCS, but not used).

Here is stack trace from try block:
```
Stack (most recent call last):
  File "/usr/lib64/python3.6/threading.py", line 884, in _bootstrap
    self._bootstrap_inner()
  File "/usr/lib64/python3.6/threading.py", line 916, in _bootstrap_inner
    self.run()
  File "/usr/lib64/python3.6/threading.py", line 864, in run
    self._target(*self._args, **self._kwargs)
  File "/usr/lib/python3.6/site-packages/patroni/async_executor.py", line 97, in run
    wakeup = func(*args) if args else func()
  File "/usr/lib/python3.6/site-packages/patroni/postgresql/rewind.py", line 344, in execute
    leader_status = self._postgresql.checkpoint(leader.conn_kwargs(self._postgresql.config.superuser))
  File "/usr/lib/python3.6/site-packages/patroni/postgresql/__init__.py", line 531, in checkpoint
    logger.exception('Exception during CHECKPOINT', stack_info=True, exc_info=True)
  File "/usr/lib/python3.6/site-packages/patroni/log.py", line 25, in error_exception
    logger_obj.error(msg, *args, exc_info=exc_info, **kwargs)
```

Standby cluster configuration is typical:
```
bootstrap: 
  dcs:
    standby_cluster:
      host: <IPADDRESS>
      port: 5432
      primary_slot_name: patroni
      create_replica_methods:
      - basebackup
```

But for authentication we use users different from "postgres" (database name):
```
  authentication:
    replication:
      username: replicator
      password: replicator
    superuser:
      username: patroni
      password: patroni
    rewind:
      username: patroni
      password: patroni
```

### DIFF
--- a/patroni/ha.py
+++ b/patroni/ha.py
@@ -1509,7 +1509,7 @@ class Ha(object):
             data = {k: v for k, v in cluster_params.items() if k in RemoteMember.allowed_keys()}
             data['no_replication_slot'] = 'primary_slot_name' not in cluster_params
             conn_kwargs = member.conn_kwargs() if member else \
-                {k: cluster_params[k] for k in ('host', 'port') if k in cluster_params}
+                {k: cluster_params[k] for k in ('host', 'port', 'database') if k in cluster_params}
             if conn_kwargs:
                 data['conn_kwargs'] = conn_kwargs
 


CyberDem0n:
The bug indeed exists, but the fix doesn't look correct to me.

mizhka:
> The bug indeed exists, but the fix doesn't look correct to me.

Could be. :)

Thank you for quick reply!

My understanding: before pg_rewind, patroni tries to do manual checkpoint, so it should connect to source database to trigger SQL statement for checkpoint. The information about source DB location is stored only in DCS, so it should fetch it from DCS (aka standby_config). Now it fetches only information about host and port. Connection string with no database is incorrect, so error occurs.
Is it correct? 

Could you please share your idea about way to fix it?

CyberDem0n:
The problem is related only to pg_rewind of the standby cluster and localized in the rewind.py.
In one place we already enrech conn_kwargs with database name is it is missing: https://github.com/zalando/patroni/blob/e8e87bf0a14b5a85e3ea0b405d2f77003e22f7ad/patroni/postgresql/rewind.py#L62-L64
There is a few more places where something similar should be done. Basically every call of `leader.conn_kwargs` might return a dict without a database key set.
Therefore I suggest to create a new method, `_conn_kwargs(self, member, auth)`, which will call the `member.conn_kwargs(auth)` and in case is database key is missing add it similarly to the `check_leader_is_not_in_recovery()`.
All calls of `leader.conn_kwargs()` should be changed to this new method. Old "hack" from check_leader_is_not_in_recovery() should be removed.

PR #1737, opened 2020-10-20
[cosmetic] log lines from stdout/stderr in case of pg_basebackup error

mizhka:
Hi,

During creating replica via pg_basebackup method, if pg_basebackup fails, Patroni logs only exit code (ex. exit code 1). 
This fix enhances logging in case of pg_basebackup error. 

Before: 
```
2020-10-20 13:30:42,332 ERROR: Error when fetching backup: pg_basebackup exited with code=1
2020-10-20 13:30:42,332 WARNING: Trying again in 5 seconds
```

After:
```
2020-10-20 13:38:17,231 ERROR: Error when fetching backup: pg_basebackup exited with code=1
2020-10-20 13:38:17,232 INFO:  stdout=
2020-10-20 13:38:17,232 INFO:  stderr=pg_basebackup: error: could not connect to server: could not connect to server: No route to host
        Is the server running on host "192.168.x.y" and accepting
        TCP/IP connections on port 5432?
```

In particular example, pg_basebackup can't connect to leader due to firewall misconfiguration. But it's hard to identify without stderr.

Thanks!

### DIFF
--- a/patroni/postgresql/bootstrap.py
+++ b/patroni/postgresql/bootstrap.py
@@ -248,14 +248,17 @@ class Bootstrap(object):
             if not self._postgresql.data_directory_empty():
                 self._postgresql.remove_data_directory()
             try:
+                results = {}
                 ret = self._postgresql.cancellable.call([self._postgresql.pgcommand('pg_basebackup'),
                                                          '--pgdata=' + self._postgresql.data_dir, '-X', 'stream',
-                                                         '--dbname=' + conn_url] + user_options, env=env)
+                                                         '--dbname=' + conn_url] + user_options, env=env, communicate=results)
                 if ret == 0:
                     break
                 else:
                     logger.error('Error when fetching backup: pg_basebackup exited with code=%s', ret)
-
+                    logger.info(' stdout=%s', results['stdout'].decode('utf-8'))
+                    logger.info(' stderr=%s', results['stderr'].decode('utf-8'))
+                    
             except Exception as e:
                 logger.error('Error when fetching backup with pg_basebackup: %s', e)
 


CyberDem0n (commented):
IMO, this change is questionable and I am unsure that it will improve the user experience.

As I understand you haven't found the pg_basebackup output in the Patroni log. How logs are configured? Do you use a rotating handler (Patroni itself writes logs to files in a specific directory)? If yes, then in this specific case the output is not lost, it just didn't get into the Patroni log. You probably should also capture stdout and stderr of Patroni, pg_basebackup will write to the same streams and therefore you will see its output.

When the database size is a few TB it could take hours for pg_basebackup to finish.
Patroni allows configuring custom options for pg_basebackup, like for example `--progress`. If we start capturing stdout and stderr, it will become impossible to follow the basebackup progress in logs.

mizhka:
Thank you, Alexander! 

Absolutely correct, this patch shouldn't touch stdout of pg_basebackup. The intention was to gather stderr due to happened error.

Patroni is configured to rotate logs because it's long-life daemon and should care about disk usage by logs. As well Patroni is managed by systemd/service/upstart/openrc/yet-another-service-mgr, so it's hard to find output file location. It will be useful to output actual error message into log files.

Please share your opinion about following proposal:
 - to keep stdout of new process "AS-IS", i.e. to inherit it from Patroni process.
 - to redirect stderr of new process to Patroni. If error occured, it will be logged via python logging. 

If it's acceptable, I will update patch. 

Thanks!

CyberDem0n:
pg_basebackup logs the progress to stderr, not stdout.

CyberDem0n:
> As well Patroni is managed by systemd/service/upstart/openrc/yet-another-service-mgr, so it's hard to find output file location

I am pretty sure it is possible to configure services to set logs to syslog or something similar, which in turn could be configured to write logs to a dedicated file.
As the last step, there is a logrotate, and the cronjob that will take care of log rotation, archiving, and retention.

With such a scheme you will get everything in a single place.

PR #1742, opened 2020-10-29
Feature: ability to ignore externally managed replication slots

jcoleman:
There are sometimes good reasons to manage replication slots externally
to Patroni. For example, a consumer may wish to manage its own slots (so
that it can more easily track when a failover has a occurred and whether
it is ahead of or behind the WAL position on the new primary).
Additionally tooling like pglogical actually replicates slots to all
replicas so that the current position can be maintained on failover
targets (this also aids consumers by supplying primitives so that they
can verify data hasn't been lost or a split brain occurred relative to
the physical cluster).

To support these use cases this new feature allows configuring Patroni
to entirely ignore sets of slots specified by any subset of name,
database, slot type, and plugin.

### DIFF
--- a/docs/SETTINGS.rst
+++ b/docs/SETTINGS.rst
@@ -38,6 +38,32 @@ Dynamic configuration is stored in the DCS (Distributed Configuration Store) and
         -  **type**: slot type. Could be ``physical`` or ``logical``. If the slot is logical, you have to additionally define ``database`` and ``plugin``.
         -  **database**: the database name where logical slots should be created.
         -  **plugin**: the plugin name for the logical slot.
+-  **ignore_slots**: list of sets of replication slot properties for which Patroni should ignore matching slots. This configuration/feature/etc. is useful when some replication slots are managed outside of Patroni. Any subset of matching properties will cause a slot to be ignored.
+    -  **name**: the name of the replication slot.
+    -  **type**: slot type. Can be ``physical`` or ``logical``. If the slot is logical, you may additionally define ``database`` and/or ``plugin``.
+    -  **database**: the database name (when matching a ``logical`` slot).
+    -  **plugin**: the logical decoding plugin (when matching a ``logical`` slot).
+
+Note: **slots** is a hashmap while **ignore_slots** is an array. For example:
+
+.. code:: YAML
+
+        slots:
+          permanent_logical_slot_name:
+            type: logical
+            database: my_db
+            plugin: test_decoding
+          permanent_physical_slot_name:
+            type: physical
+          ...
+        ignore_slots:
+          - name: ignored_logical_slot_name
+            type: logical
+            database: my_db
+            plugin: test_decoding
+          - name: ignored_physical_slot_name
+            type: physical
+          ...
 
 Global/Universal
 ----------------
--- a/features/ignored_slots.feature
+++ b/features/ignored_slots.feature
@@ -44,6 +44,7 @@ Feature: ignored slots
     When I start postgres1
     Then postgres1 role is the secondary after 20 seconds
     And "members/postgres1" key in DCS has role=replica after 3 seconds
+    And I sleep for 2 seconds  # give Patroni time to sync replication slots
     And postgres1 has a logical replication slot named unmanaged_slot_0 with the test_decoding plugin
     And postgres1 has a logical replication slot named unmanaged_slot_1 with the test_decoding plugin
     And postgres1 has a logical replication slot named unmanaged_slot_2 with the test_decoding plugin
--- a/features/ignored_slots.feature
+++ b/features/ignored_slots.feature
@@ -1,0 +1,59 @@
+Feature: ignored slots
+  Scenario: check ignored slots aren't removed on failover/switchover
+    Given I start postgres1
+    Then postgres1 is a leader after 10 seconds
+    And there is a non empty initialize key in DCS after 15 seconds
+    When I issue a PATCH request to http://127.0.0.1:8009/config with {"loop_wait": 2, "ignore_slots": [{"name": "unmanaged_slot_0", "database": "postgres", "plugin": "test_decoding", "type": "logical"}, {"name": "unmanaged_slot_1", "database": "postgres", "plugin": "test_decoding"}, {"name": "unmanaged_slot_2", "database": "postgres"}, {"name": "unmanaged_slot_3"}], "postgresql": {"parameters": {"wal_level": "logical"}}}
+    Then I receive a response code 200
+    And Response on GET http://127.0.0.1:8009/config contains ignore_slots after 10 seconds
+    # Make sure the wal_level has been changed.
+    When I shut down postgres1
+    And I start postgres1
+    Then postgres1 is a leader after 10 seconds
+    And "members/postgres1" key in DCS has role=master after 3 seconds
+    # Make sure Patroni has finished telling Postgres it should be accepting writes.
+    And postgres1 role is the primary after 20 seconds
+    # 1. Create our test logical replication slot.
+    # Test that ny subset of attributes in the ignore slots matcher is enough to match a slot
+    # by using 3 different slots.
+    When I create a logical replication slot unmanaged_slot_0 on postgres1 with the test_decoding plugin
+    And I create a logical replication slot unmanaged_slot_1 on postgres1 with the test_decoding plugin
+    And I create a logical replication slot unmanaged_slot_2 on postgres1 with the test_decoding plugin
+    And I create a logical replication slot unmanaged_slot_3 on postgres1 with the test_decoding plugin
+    And I create a logical replication slot dummy_slot on postgres1 with the test_decoding plugin
+    # It seems like it'd be obvious that these slots exist since we just created them,
+    # but Patroni can actually end up dropping them almost immediately, so it's helpful
+    # to verify they exist before we begin testing whether they persist through failover
+    # cycles.
+    Then postgres1 has a logical replication slot named unmanaged_slot_0 with the test_decoding plugin
+    And postgres1 has a logical replication slot named unmanaged_slot_1 with the test_decoding plugin
+    And postgres1 has a logical replication slot named unmanaged_slot_2 with the test_decoding plugin
+    And postgres1 has a logical replication slot named unmanaged_slot_3 with the test_decoding plugin
+
+    When I start postgres0
+    Then "members/postgres0" key in DCS has role=replica after 3 seconds
+    And postgres0 role is the secondary after 20 seconds
+    # Verify that the replica has advanced beyond the point in the WAL
+    # where we created the replication slot so that on the next failover
+    # cycle we don't accidentally rewind to before the slot creation.
+    And replication works from postgres1 to postgres0 after 20 seconds
+    When I shut down postgres1
+    Then "members/postgres0" key in DCS has role=master after 3 seconds
+
+    # 2. After a failover the server (now a replica) still has the slot.
+    When I start postgres1
+    Then postgres1 role is the secondary after 20 seconds
+    And "members/postgres1" key in DCS has role=replica after 3 seconds
+    And postgres1 has a logical replication slot named unmanaged_slot_0 with the test_decoding plugin
+    And postgres1 has a logical replication slot named unmanaged_slot_1 with the test_decoding plugin
+    And postgres1 has a logical replication slot named unmanaged_slot_2 with the test_decoding plugin
+    And postgres1 has a logical replication slot named unmanaged_slot_3 with the test_decoding plugin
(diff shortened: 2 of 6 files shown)


patroni/postgresql/slots.py
```diff
@@ -48,6 +48,20 @@ def sync_replication_slots(self, cluster):
...
+                        if (
+                                (ignore_slots_matcher.get("name") is None or ignore_slots_matcher["name"] == name)
+                                and (ignore_slots_matcher.get("database") is None or ignore_slots_matcher["database"] == slot["database"])
+                                and (ignore_slots_matcher.get("plugin") is None or ignore_slots_matcher["plugin"] == slot["plugin"])
+                                and (ignore_slots_matcher.get("type") is None or ignore_slots_matcher["type"] == slot["type"])
+                            ):
+                            ignore_slot = True
+                            break
+                    if ignore_slot:
+                        break
```
CyberDem0n, line 63: I think we should not break the outer loop, there might be some slots which must be deleted.
I.e. we need something like:
```python
if not ignore_slot and not self.drop_replication_slot(name):
...
```
```

File: patroni/validator.py [hunk 2]
```diff
@@ -365,6 +365,26 @@ def assert_(condition, message="Wrong value"):
...
+    ),
+  },
+  Optional("ignore_slots"): [
+    {
+      Optional("name"): str,
+      Optional("database"): str,
+      Optional("plugin"): str,
+      Optional("type"): str,
+    }
+  ],
```
CyberDem0n, line 387: `slots` and `ignore_slots` are stored in DCS (and also always accessed via cluster.config), therefore validator in its current state will not check them.

patroni/postgresql/slots.py
```diff
@@ -48,6 +48,20 @@ def sync_replication_slots(self, cluster):
...
                 for name in set(self._replication_slots) - set(slots):
+                    ignore_slot = False
+                    for ignore_slots_matcher in cluster.config.ignore_slots_matchers:
+                        slot = self._replication_slots[name]
+                        if (
+                                (ignore_slots_matcher.get("name") is None or ignore_slots_matcher["name"] == name)
+                                and (ignore_slots_matcher.get("database") is None or ignore_slots_matcher["database"] == slot["database"])
+                                and (ignore_slots_matcher.get("plugin") is None or ignore_slots_matcher["plugin"] == slot["plugin"])
+                                and (ignore_slots_matcher.get("type") is None or ignore_slots_matcher["type"] == slot["type"])
+                            ):
```
CyberDem0n, line 59: Flake8 (on travis) complains about these lines:
```
patroni/postgresql/slots.py:56:121: E501 line too long (138 > 120 characters)
patroni/postgresql/slots.py:57:121: E501 line too long (132 > 120 characters)
patroni/postgresql/slots.py:58:121: E501 line too long (126 > 120 characters)
patroni/postgresql/slots.py:59:29: E125 continuation line with same indent as next logical line
```

You can also check it locally as:
```
python3 setup.py flake8
```

patroni/postgresql/slots.py
jcoleman: Fixing.

patroni/validator.py
jcoleman: If they are configured in the yml and you run `--validate-config` would it be checked? Should we instead do validation somewhere else?

CyberDem0n: If they are configured in yaml (`postgresql.slots` and `postgresql.ignore_slots`), this [code](https://github.com/zalando/patroni/pull/1742/files#diff-e77f860f36a4c7c541f165a9ded2335eb58ed020bab147af0a2315d912dde812R52) will not pick it up.

jcoleman: I believe both of those are top-level parameters not on the `postgresql` hash, right?

Does that mean you can't configure `slots` at startup? Or just that the way we're accessing it needs to change? Or that we need to load the config into the DCS when we start up?

patroni/postgresql/slots.py
jcoleman: Fixing.

patroni/validator.py
CyberDem0n: Oh, sorry missed it.
Yes, they are sort of "top" level, but in the global configuration, which is stored in the DCS, in the /config key.
When you run `patronictl edit-config` it works with the /config key.

But, in the Paroni config there is `bootstrap.dcs` section. These parameters could be placed there and Patroni will write the content of bootstrap.dcs to the config key when the new cluster is created (initdb or custom bootstrap).

Having some validation for bootstrap.dcs would be good, but it is a different task.

jcoleman: OK, we'll remove the validation for now.

patroni/postgresql/slots.py
```diff
@@ -33,6 +33,19 @@ def load_replication_slots(self):
...
+        for matcher in cluster.config.ignore_slots_matchers:
+            if (
+                    (matcher.get("name") is None or matcher["name"] == slot["name"])
+                    and (matcher.get("database") is None or matcher["database"] == slot["database"])
+                    and (matcher.get("plugin") is None or matcher["plugin"] == slot["plugin"])
+                    and (matcher.get("type") is None or matcher["type"] == slot["type"])
+               ):
+                ignore_slot = True
+                break
+        ignore_slot
```
CyberDem0n, line 47: The `slot` doesn't always have plugin and database keys, but only for logical slots, and that's the reason why tests sometimes fail.

```suggestion
    def ignore_replication_slot(self, cluster, name):
        slot = self._replication_slots[name]
        for matcher in cluster.config.ignore_slots_matchers:
            if all(not matcher.get(a) or matcher[a] == slot.get(a) for a in ('name', 'database', 'plugin', 'type')):
                return True
        return False
```

patroni/postgresql/slots.py
```diff
@@ -48,7 +61,8 @@ def sync_replication_slots(self, cluster):
 
                 # drop old replication slots which are not presented in desired slots
                 for name in set(self._replication_slots) - set(slots):
-                    if not self.drop_replication_slot(name):
+                    slot = self._replication_slots[name]
+                    if not self.ignore_replication_slot(cluster, slot) and not self.drop_replication_slot(name):
```
CyberDem0n, line 65:
```suggestion
                    if not self.ignore_replication_slot(cluster, name) and not self.drop_replication_slot(name):
```

features/ignored_slots.feature
```diff
@@ -0,0 +1,54 @@
...
+    # Test that ny subset of attributes in the ignore slots matcher is enough to match a slot
+    # by using 3 different slots.
+    When I create a logical replication slot unmanaged_slot_0 on postgres1 with the test_decoding plugin
+    And I create a logical replication slot unmanaged_slot_1 on postgres1 with the test_decoding plugin
+    And I create a logical replication slot unmanaged_slot_2 on postgres1 with the test_decoding plugin
+    And I create a logical replication slot unmanaged_slot_3 on postgres1 with the test_decoding plugin
+    Then postgres1 has a logical replication slot named unmanaged_slot_0 with the test_decoding plugin
+    And postgres1 has a logical replication slot named unmanaged_slot_1 with the test_decoding plugin
+    And postgres1 has a logical replication slot named unmanaged_slot_2 with the test_decoding plugin
+    And postgres1 has a logical replication slot named unmanaged_slot_3 with the test_decoding plugin
```
CyberDem0n, line 26: We just created slots, and creation didn't fail, so I think this block probably could be dropped.

features/ignored_slots.feature
```diff
@@ -0,0 +1,54 @@
...
+    When I issue a PATCH request to http://127.0.0.1:8009/config with {"loop_wait": 2, "ignore_slots": [{"name": "unmanaged_slot_0", "database": "postgres", "plugin": "test_decoding", "type": "logical"}, {"name": "unmanaged_slot_1", "database": "postgres", "plugin": "test_decoding"}, {"name": "unmanaged_slot_2", "database": "postgres"}, {"name": "unmanaged_slot_3"}], "postgresql": {"parameters": {"wal_level": "logical"}}}
+    Then I receive a response code 200
+    # Make sure the wal_level has been changed.
+    When I shut down postgres1
+    And I start postgres1
+    Then postgres1 is a leader after 10 seconds
+    And "members/postgres1" key in DCS has role=master after 3 seconds
+    # Make sure Patroni has finished telling Postgres it should be accepting writes.
+    And postgres1 role is the primary after 20 seconds
+    And Response on GET http://127.0.0.1:8009/config contains ignore_slots after 10 seconds
```
CyberDem0n, line 15: The `/config` endpoint is more or less proxy to the `/config` key in DCS. Since the [PATCH returned 200](https://github.com/zalando/patroni/pull/1742/files#diff-ff1c3e9e4d3e4c711f4d2c6f500b4ea23d93fe454b02ab5ebb86eeddc7771f26R6-R7), values were persisted in the DCS.

docs/SETTINGS.rst
```diff
@@ -38,6 +38,11 @@ Dynamic configuration is stored in the DCS (Distributed Configuration Store) and
         -  **type**: slot type. Could be ``physical`` or ``logical``. If the slot is logical, you have to additionally define ``database`` and ``plugin``.
         -  **database**: the database name where logical slots should be created.
         -  **plugin**: the plugin name for the logical slot.
+-  **ignore_slots**: list of sets of replication slot properties for which Patroni should ignore matching slots. This configuration/feature/etc. is useful when some replication slots are managed outside of Patroni. Any subset of matching properties will cause a slot to be ignored.
+    -  **name**: the name of the replication slot.
+    -  **type**: slot type. Can be ``physical`` or ``logical``. If the slot is logical, you may additionally define ``database`` and/or ``plugin``.
+    -  **database**: the database name (when matching a ``logical`` slot).
+    -  **plugin**: the logical decoding plugin (when matching a ``logical`` slot).
```
CyberDem0n, line 45: `slots` is a dict, while `ignore_slots` is a list.
What do you think about adding the YAML code snippet with the example, two slots and two ignore_slots? IMO it would help to avoid confusion.

patroni/postgresql/slots.py
jcoleman: It must be the case that some DCSs (etcd3 and kubernettes) handle errors differently. We consistently have the raft suite pass locally, but we noticed that actually the key errors that are logged in the Travis builds that failed are also logged when running the tests against raft even though they don't result in tests failing.

Looking into that further we realized that when we refactored out the slot check into a method we added a bug -- the slot name isn't present in the slot dict, so a key error was being thrown on basically every invocation (if `ignore_slots` had any config). Our tests passed though with raft because that error was being swallowed, and all of the tests were about slots being ignored. On the other hand etcd3 and kubernetes appear to stop doing anything when that key error occurs, which results in the failing test since Patroni never actually gets to the code that configured the state of the server.

We're pushing a fix that handles this issue plus updates the tests to guarantee it.

features/ignored_slots.feature
jcoleman: We think these tests should stay because in our testing the slots got almost immediately deleted after this point without our change, so it's helpful to guarantee they properly exist before moving forward with failover cycles.

features/ignored_slots.feature
jcoleman: We actually copied that assertion from the first test in `features/standby_cluster.feature`, but it does seem good to move it up to be part of the assertions that follow the PATCH, so we're changing that.

docs/SETTINGS.rst
jcoleman: Added an example chunk of YAML.

jcoleman:
@CyberDem0n Any news on this?

docs/SETTINGS.rst
Jan-M: For ignored slots, why does one need the extensive config? isn't `name` enough?

jcoleman: It's potentially useful (this is in our use case) to exclude on other combinations of config options without having to specify the specific slot. For example, we want to be able to specify that all logical slots using a specific output plugin (where we know the plugin manages the slots) should be ignored.

Jan-M: Ah did not get that smart "matching" from the example.

Jan-M:
:+1:

Jan-M:
Looks good to me, thanks for your patience and contributing.

CyberDem0n:
:+1:

CyberDem0n:
@jcoleman first we need to finish https://github.com/zalando/patroni/pull/1748, otherwise tests on the master will fail (due to absent zookeeper binaries)

jcoleman:
@CyberDem0n I think all of the comments have been resolved on #1748 now.

Issue #1759, opened 2020-11-14
Multi-instance raft


mbanck:
It seems pure raft based on pysyncobj does not namespace the configuration values like e.g. etcd does (which uses the `/$SCOPE/` as prefix/namespace). So you can't run two patroni instances on the same raft port.

Is this not possible due to restrictions in the way the key-value store works (no directories/nesting), or even due to the fact that it would be intrinsically tied to the patroni instance creating it (maybe like the systemid in postgresql)?

It would of course be possible to just select a different port for each instance, but this makes it more cumbersome to setup multiple instances the way Debian does it, because it currently assumes the DCS configuration is the same for each patroni instance.

As an aside, is there some default port assigned for raft? I see both 1234 and 222[234] in the code.

Maybe it could be made to just use API port + 10000 (so 18008 by default) if no port is specified, (assuming that this class could even get at the API port)?

CyberDem0n:
> It seems pure raft based on pysyncobj does not namespace the configuration values like e.g. etcd does (which uses the /$SCOPE/ as prefix/namespace).

In this regard, the raft implementation isn't different from any other DCS. It uses `/$namespace/$scope/` as prefix for keys.

> So you can't run two patroni instances on the same raft port.

IMO, the main use-case for using raft, is when you just have a single cluster, with two or three PostgreSQL nodes.
In theory, (and I believe it wouldn't be hard to show it in practice) it is possible to run the pysyncobj cluster of three `patroni_raft_controler`. This cluster could be used by Patroni. For that, you just don't need to specify the `raft.self_addr` in the patroni.yaml and it will not listen for incoming connections, but use the aforementioned cluster. But again, IMO, instead of using such a setup, it would be better to rely on well tested DCS, like Etcd, Consul, and Zookeeper.

> It would of course be possible to just select a different port for each instance, but this makes it more cumbersome to setup multiple instances the way Debian does it, because it currently assumes the DCS configuration is the same for each patroni instance.

Yes, that's the major difference of raft with other DCS. You have to specify in the config the list of other nodes of this specific PostgreSQL cluster.

> As an aside, is there some default port assigned for raft? I see both 1234 and 222[234] in the code.

I don't think that pysyncobj recommends or advertises any default port. It just happened that in my experiments I used some "random" ports, so they ended up in the sample configs and code.

> Maybe it could be made to just use API port + 10000 (so 18008 by default) if no port is specified, (assuming that this class could even get at the API port)?

[...]

PR #1774, opened 2020-12-01
Allow to omit the RAFT port (#1759)

mbanck:
If no ports are specified for self_addr, bind_addr and/or partner_addrs, then
Patroni will self-assign the startup API port + 10000 as RAFT port.

### DIFF
--- a/docs/SETTINGS.rst
+++ b/docs/SETTINGS.rst
@@ -237,6 +237,9 @@ Raft
 
     A: Yes, on the third node you can run ``patroni_raft_controller`` (without Patroni and PostgreSQL). In such setup one can temporary loose one node without affecting the primary.
 
+  - Q: Do I need to specify a port?
+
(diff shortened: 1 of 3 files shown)


CyberDem0n (commented):
With raft port being deduced from the REST API port, the change of API port will implicitly change the pysyncobj identifier (host:port), what could create bring the cluster to an unrecoverable state, because on start Patroni checks whether this node is already known as a cluster member and if it is not - adds itself.

Example. We have a cluster from `10.0.0.1:18080,10.0.0.2:18080,10.0.0.3:18080`.
After the change of the REST API port on the first node to 8081, the raft address will become `10.0.0.1:18081`, and it will add itself to the cluster members. The list of members will become `10.0.0.1:18080,10.0.0.2:18080,10.0.0.3:18080,10.0.0.1:18081`. This configuration will stop working as soon as we stop the first node.

For me it looks like a footgun.

PR #1777, opened 2020-12-04
Docs / README - mention currently supported PostgreSQL versions

kmoppel:
Hi,
I think would make sense to hightlight supported PG versions more prominently...lot of questions on that somehow from our customers. Currently one needs to dive into the Issues history to get some clues...

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -12,6 +12,8 @@ Patroni is a template for you to create your own customized, high-availability s
 
 We call Patroni a "template" because it is far from being a one-size-fits-all or plug-and-play replication system. It will have its own caveats. Use wisely.
 
+Currently supported PostgreSQL versions: 9.3 to 14devel.
+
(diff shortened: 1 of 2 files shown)


bradnicholson:
Even though Patroni works with older versions, would it make more sense to take a support stance that Patroni officially supports the currently supported versions of Postgres?  Seems like it would be nice to mirror those and drop off support for older versions.

kmoppel:
Sure, if Patroni project's policy is to follow the official Postgres support window then this would of course be the message here...so the question is I guess if there's a support strategy and are there some automatic tests covering only last version etc. ...@CyberDem0n you probably have some input here? :)

CyberDem0n:
There is no big value in removing support of PostgreSQL versions which are EOL.
The only problem with legacy, that they are not covered by behave tests.
Support 14devel is a bit too ambitious, it is better to stop at 13 :)

kmoppel:
@CyberDem0n ok probably a good idea - although I checked that 14devel works things might still change there too much...

CyberDem0n:
:+1:

FxKu:
:+1:

CyberDem0n:
Merged, thank you @kmoppel.

PR #1779, opened 2020-12-04
Warn the user when the required watchdog is not healthy

orgrim:
When the watchdog device is not writable or missing in required mode,
the member cannot be promoted. It only keeps logging that it is not the
healthiest member. Add a warning to show the user where to search for
this misconfiguration.


FxKu:
:+1:

CyberDem0n:
:+1:

PR #1786, opened 2020-12-10
Choose synchronous node based on replication lag

ksarabu1:
Patroni keeps the same synchrounous_standby node even when it is in an unhealthy & hung state (sync standby or the replay process was hung on sync standby for e.g.). Fix is to select a new healthy node to replace the unhealthy sync node in `synchronous_standby_names` on the leader.


patroni/postgresql/__init__.py
```diff
@@ -984,16 +984,17 @@ def pick_synchronous_standby(self, cluster, sync_node_count=1):
         # receiving changes faster than the sync member (very rare but possible). Such cases would
         # trigger sync standby member swapping frequently and the sort on sync_state desc should
         # help in keeping the query result consistent.
-        for app_name, state, sync_state in self.query(
+        for app_name, state, sync_state, replica_rank in self.query(
                 "SELECT pg_catalog.lower(application_name), state, sync_state"
+                ", rank() over (order by {0}_{1} DESC) as replica_rank"
```
CyberDem0n, line 989: It seems that we are back to roots, i.e. `ORDER BY flush_lsn DESC, sync_state DESC`.

The issue with the `RANK()` and `replica_rank == 1` check that it would require all replicas to be on the same LSN, which IMO on the system with a lot of TPS would rarely be the case.

ksarabu1: Agree. We could probably do the opposite, maybe identifying unhealthy sync (if the replication delay is over the threshold) and taking it out from `sync_nodes` list?

CyberDem0n: There is no guaranty that `replay/flush/write` for sync replicas would have the same values.
It is very likely that soon or later Patroni will catch the moment when different replicas are off from each other by a few bytes/kilobytes, therefore not all of them will have the same rank.

> maybe identifying unhealthy sync (if the delay is over the threshold) and taking it out from sync_nodes list?

It could be an option, there are a few questions though:
1. What would be the threshold? Should it be configurable?
2. If there are many replicas the threshold could be applied from the most up-to-date. What should we do if there is only one replica? Should we use the current wal lsn of the primary?

ksarabu1: Thanks. The safest approach would be going with identifying unhealthy synchronous replica and demoting it to asynchronous:

1. Yes, This will be a configurable parameter, with an acceptable default value.
2. Do you foresee any issues using current lsn of the primary in case of both, i.e. multiple replicas and single replica?

CyberDem0n: Transactions are first always written to the WAL on the primary and only after that they are sent to replicas.
New transactions don't necessarily wait until other transactions are confirmed by sync standbys.
If the transaction rate on the primary is high enough, the current lsn would be most likely is ahead of feedback values reported via pg_stat_replication. If the threshold is too low that would become a problem, because Patroni will switch to async replication.

ksarabu1: Agree & Your suggestion above makes sense in this case. For a single replica, we could use current wal lsn of the primary. Will start making changes. Thanks.

features/steps/basic_replication.py
```diff
@@ -33,6 +33,46 @@ def add_table(context, table_name, pg_name):
...
+
+@step('I resume wal replay on {pg_name:w}')
+def resume_wal_replay(context, pg_name):
+    # resume the wal replay process
+    try:
+       version = context.pctl.query(pg_name, "select pg_catalog.pg_read_file('PG_VERSION', 0, 2)").fetchone()
+       wal = version and version[0] and int(version[0].split('.')[0]) < 10 and "xlog" or "wal"
+       context.pctl.query(pg_name, "SELECT pg_{0}_replay_resume()".format(wal))
+    except pg.Error as e:
+        assert False, "Error resuming wal recovery on {0}: {1}".format(pg_name, e)
```
CyberDem0n, line 55: It could be just a single function:
```python
@step('I {action:w} wal replay on {pg_name:w}')
def toggle_wal_replay(context, action, pg_name):
```

patroni/postgresql/__init__.py
```diff
@@ -984,16 +988,30 @@ def pick_synchronous_standby(self, cluster, sync_node_count=1):
         # receiving changes faster than the sync member (very rare but possible). Such cases would
         # trigger sync standby member swapping frequently and the sort on sync_state desc should
         # help in keeping the query result consistent.
-        for app_name, state, sync_state in self.query(
-                "SELECT pg_catalog.lower(application_name), state, sync_state"
+        for app_name, state, sync_state, replica_lag in self.query(
+                "SELECT app_name, state, sync_state,"
+                "       (case when replica_count > 0 then max_replica_lsn - replica_lsn"
+                "             else current_lsn - replica_lsn"
+                "        end) replica_lag"
```
CyberDem0n, line 995: There is no guaranty that the most up-to-date replica is actually the member of the cluster (it could be pg_receivewal for example).

Taking into account that there are also potential issues with pg_current_wal_lsn() (see below), I think it would make sense to perform post-processing in the python code.

Instead of calling pg_current_wal_lsn() you can use the values returned by the `last_operation()` method:
```python
current_wal_lsn = int(str(self.last_operation()))
```
This value could be a bit not up-to-date, but that shouldn't cause a big issue.

features/steps/basic_replication.py
ksarabu1: Fixed

features/basic_replication.feature
```diff
@@ -28,6 +28,26 @@ Feature: basic replication
...
+    And I sleep for 10 seconds
+    And I load data on postgres0
+    And I sleep for 20 seconds
+    Then "sync" key in DCS has sync_standby=postgres1 after 5 seconds
+    And I issue a GET request to http://127.0.0.1:8009/sync
+    Then I receive a response code 200
+    And I resume wal replay on postgres2
+    And I sleep for 20 seconds
+    When I issue a GET request to http://127.0.0.1:8010/async
+    Then I receive a response code 200
```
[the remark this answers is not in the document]
ksarabu1: We have already tested the scenario on `postgres2`, not sure if it adds any value in repeating the same test.

features/basic_replication.feature
```diff
@@ -28,6 +28,26 @@ Feature: basic replication
     When I issue a GET request to http://127.0.0.1:8009/async
     Then I receive a response code 200
 
+  Scenario: check stuck sync replica
+    Given I issue a PATCH request to http://127.0.0.1:8008/config with {"maximum_lag_on_syncnode": 100000000, "postgresql": {"parameters": {"synchronous_commit": "remote_apply"}}}
```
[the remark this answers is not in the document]
ksarabu1: Sure. Will reduce the txn size.

features/basic_replication.feature
```diff
@@ -28,6 +28,26 @@ Feature: basic replication
     When I issue a GET request to http://127.0.0.1:8009/async
     Then I receive a response code 200
 
+  Scenario: check stuck sync replica
```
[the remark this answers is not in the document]
ksarabu1: Sleep options were added part of debug when smoke tests failed initially. will remove unnecessary sleeps.

features/basic_replication.feature
```diff
@@ -28,6 +28,26 @@ Feature: basic replication
...
+    Given I issue a PATCH request to http://127.0.0.1:8008/config with {"maximum_lag_on_syncnode": 100000000, "postgresql": {"parameters": {"synchronous_commit": "remote_apply"}}}
+    Then I receive a response code 200
+    And I create table on postgres0
+    And I sleep for 5 seconds
+    When I pause wal replay on postgres2
+    And I sleep for 10 seconds
```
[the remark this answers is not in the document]
ksarabu1: Sleep options were added part of debug when smoke tests failed initially. will remove unnecessary sleeps.

features/basic_replication.feature
```diff
@@ -28,6 +28,26 @@ Feature: basic replication
...
+    Then "sync" key in DCS has sync_standby=postgres1 after 5 seconds
+    And I issue a GET request to http://127.0.0.1:8009/sync
+    Then I receive a response code 200
+    And I resume wal replay on postgres2
+    And I sleep for 20 seconds
+    When I issue a GET request to http://127.0.0.1:8010/async
+    Then I receive a response code 200
+    When I issue a PATCH request to http://127.0.0.1:8008/config with {"maximum_lag_on_syncnode": -1, "postgresql": {"parameters": {"synchronous_commit": "on"}}}
+    Then I receive a response code 200
+    And I sleep for 10 seconds
```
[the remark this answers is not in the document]
ksarabu1: Sure, will remove/cleanup.

features/basic_replication.feature
```diff
@@ -28,6 +28,26 @@ Feature: basic replication
...
 
+  Scenario: check stuck sync replica
+    Given I issue a PATCH request to http://127.0.0.1:8008/config with {"maximum_lag_on_syncnode": 100000000, "postgresql": {"parameters": {"synchronous_commit": "remote_apply"}}}
+    Then I receive a response code 200
+    And I create table on postgres0
+    And I sleep for 5 seconds
+    When I pause wal replay on postgres2
+    And I sleep for 10 seconds
+    And I load data on postgres0
+    And I sleep for 20 seconds
```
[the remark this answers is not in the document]
ksarabu1: Sleep options were added part of debug when smoke tests failed initially. will remove unnecessary sleeps.

patroni/postgresql/__init__.py
```diff
@@ -984,16 +988,30 @@ def pick_synchronous_standby(self, cluster, sync_node_count=1):
...
+                " FROM ("
+                " SELECT pg_catalog.lower(application_name) app_name, state, sync_state"
+                "       , pg_{2}_{1}_diff(pg_current_{2}_{1}(), '0/0')::bigint current_lsn"
```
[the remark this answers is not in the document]
ksarabu1: Ok. Updated the code and replaced the logic with `self.last_operation()` as suggested below.

patroni/postgresql/__init__.py
```diff
@@ -984,16 +988,31 @@ def pick_synchronous_standby(self, cluster, sync_node_count=1):
...
+        for app_name, state, sync_state, replica_lag in self.query(
+                "SELECT app_name, state, sync_state,"
+                "       (case when replica_count > 0 then max_replica_lsn - replica_lsn"
+                "             else {3} - replica_lsn"
+                "        end) replica_lag"
+                " FROM ("
+                " SELECT pg_catalog.lower(application_name) app_name, state, sync_state"
+                "       , pg_{2}_{1}_diff({0}_{1}, '0/0')::bigint replica_lsn"
+                "       , max( pg_{2}_{1}_diff({0}_{1}, '0/0')::bigint ) over () max_replica_lsn"
+                "       , count(1) over () replica_count"
```
CyberDem0n, line 1001: The problem is that not all connected replicas are real cluster members.
It could be for example pg_receivewal, or standby cluster and they should not be used in the `MAX()` and `COUNT()`.

We have two options to handle it:
1. Filter by application_name in the query
2. Fetch the whole result set to python and process it.

I would prefer the second one.

ksarabu1: Thanks. Fixed the code.

patroni/postgresql/__init__.py
```diff
@@ -984,17 +989,22 @@ def pick_synchronous_standby(self, cluster, sync_node_count=1):
...
-                " ORDER BY sync_state DESC, {0}_{1} DESC".format(sort_col, self.lsn_name)):
+                " ORDER BY sync_state DESC, {0}_{1} DESC".format(sort_col, self.lsn_name, self.wal_name)):
             member = members.get(app_name)
-            if not member or member.tags.get('nosync', False):
-                continue
-            candidates.append(member.name)
-            if sync_state == 'sync':
-                sync_nodes.append(member.name)
+            if member and not member.tags.get('nosync', False):
+                replica_list.append((member.name, state, sync_state, replica_lsn))
```
CyberDem0n, line 999: Since the `state` is not used anymore outside the query I think it could be removed.

patroni/api.py
```diff
@@ -197,7 +197,7 @@ def _read_json_content(self, body_is_optional=False):
     def do_PATCH_config(self):
         request = self._read_json_content()
         if request:
-            cluster = self.server.patroni.dcs.get_cluster()
+            cluster = self.server.patroni.dcs.get_cluster(True)
```
sdudoladov, line 200: If it is meant to be [used with Zookeeper](https://github.com/zalando/patroni/blob/d7f579ee61822cd0fc181a0783c2b920f902734a/patroni/dcs/__init__.py#L657-L662), it is worth an explicit check the DCS is indeed Zookeper.

sdudoladov:
:+1:

patroni/api.py
CyberDem0n: Caller doesn't know anything about DCS implementation.
It is sort of possible to try to load the very specific DCS and use isinstance for comparison, but loading unnecessary modules increases memory footprint.

CyberDem0n:
:+1:

CyberDem0n:
Merged, thank you @ksarabu1!

PR #1788, opened 2020-12-11
Using hot_stanby = off when custom bootstrap prevents from failing …

yanchenko-igor:
… if we are restoring WAL files which increase values like max_connections.


Jan-M:
:+1:

CyberDem0n:
:+1:

CyberDem0n:
Merged, thank you @yanchenko-igor!

PR #1791, opened 2020-12-14
Submit coverage to codacy only if secret is available

CyberDem0n:
If PR is open from the external GH repo secrets are not set due to security reasons. It makes codacy coverage report to fail.


FxKu:
:+1:

CyberDem0n:
:+1:

PR #1792, opened 2020-12-14
Fix issues with zookeeper

CyberDem0n:
1. The `ttl` was incorrectly returned 1000 times higher then it should
2. The `watch()` method must return True if the parent method returned True. Not doing so resulted in the incorrect calculation of sleep time.
3. Move mock of exhibitor api to the features/environment.py. It simplifies testing with behave.

### DIFF
--- a/.github/workflows/install_deps.py
+++ b/.github/workflows/install_deps.py
@@ -154,13 +154,6 @@ users:
     return 0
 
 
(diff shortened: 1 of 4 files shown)


Jan-M:
:+1:

CyberDem0n:
:+1:

PR #1794, opened 2020-12-14
Fixes around pg_rewind

CyberDem0n:
1. If the superuser name is different from postgres, the pg_rewind in the standby cluster was failing because the connection string didn't contain the database name.
2. Provide output if the single-user mode recovery failed.

Close https://github.com/zalando/patroni/pull/1736

### DIFF
--- a/patroni/postgresql/rewind.py
+++ b/patroni/postgresql/rewind.py
@@ -59,7 +59,8 @@ class Rewind(object):
         if self.can_rewind_or_reinitialize_allowed and self._state != REWIND_STATUS.NEED:
             self._state = REWIND_STATUS.CHECK
 
-    def check_leader_is_not_in_recovery(self, conn_kwargs):
+    @staticmethod
+    def check_leader_is_not_in_recovery(conn_kwargs):
         try:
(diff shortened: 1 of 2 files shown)


patroni/postgresql/rewind.py
```diff
@@ -59,11 +59,10 @@ def trigger_check_diverged_lsn(self):
         if self.can_rewind_or_reinitialize_allowed and self._state != REWIND_STATUS.NEED:
             self._state = REWIND_STATUS.CHECK
 
-    def check_leader_is_not_in_recovery(self, **kwargs):
-        if not kwargs.get('database'):
-            kwargs['database'] = self._postgresql.database
+    @staticmethod
+    def check_leader_is_not_in_recovery(conn_kwargs):
```
FxKu, line 63:
```suggestion
    def check_leader_is_not_in_recovery(conn_kwargs={}):
```
add default? So you don't have to call it like `check_leader_is_not_in_recovery({})`

CyberDem0n: No, it is supposed to always get some connection parameters.
The empty `dict` is only passed from the unit-tests.

FxKu:
:+1:

CyberDem0n:
👍

Issue #1727, opened 2020-10-05
Failure during call_post_bootstrap


android-leha:
Hi,

Bug presents on line: https://github.com/zalando/patroni/blob/master/patroni/postgresql/bootstrap.py#L133
In case there is no superuser password, _env_ varible wouldn't be initializated(None) and next line will fail with exception.

This is really small bug, related only for passwordless patroni config file

Need to define certificats authantication for superuser without password and ad post_init script

env should be initialized with empty dictionary

- Patroni version: 2.0.0
- PostgreSQL version: 12
- DCS (and its version): zookeeper 3.6.1

Unfortunately didn't save stacktrace.
Was exception on next line:
**TypeError: 'NoneType' object does not support item assignment**

I did manual patch to fix the issue:
This line
`env = self._postgresql.config.write_pgpass(r) if 'password' in r else None`
was replaced with this
`env = self._postgresql.config.write_pgpass(r) if 'password' in r else {}`

CyberDem0n:
Hi @android-leha, thank you for a bug-report.
The correct fix would be just:
```python
env = self._postgresql.config.write_pgpass(r)
```

Could you please create a PR?

PR #1795, opened 2020-12-14
Fix bug in the post_bootstrap method()

CyberDem0n:
The config.write_pgpass(r) should be called unconditionally, no matter whether the password defined in the config or not.

Close https://github.com/zalando/patroni/issues/1727


FxKu:
:+1:

CyberDem0n:
:+1:

Issue #1767, opened 2020-11-21
Primary demoted after concurrent update


le-duane:
**Describe the bug**
Primary demoted after only 1 concurrent update

**To Reproduce**
I'm not sure - in my testing, the leader lock update was retried after concurrent updates were encountered. It is possible something else was going on in this case which did not show up in Patroni or API logs; some other Patroni instances reported timeout errors around the same time.

```
2020-11-16 13:35:19,571 ERROR: Request to server https://172.20.0.1:443 failed: ReadTimeoutError("HTTPSConnectionPool(host='172.20.0.1', port=443): Read timed out. (read timeout=12.409012882038951)",)
``` 

We use managed Kubernetes (EKS) so there is not a lot of visibility as to what happened to the API server to cause the timeout.

**Expected behavior**
Patroni to retry after 409 response from API server. Patroni does not demote before retry_timeout expires

**Environment**
- Patroni version: 2.0.0
- PostgreSQL version: 11.9
- DCS (and its version): Kubernetes 14.9

**patronictl show-config**
```
loop_wait: 10
master_start_timeout: 300
master_stop_timeout: 135
maximum_lag_on_failover: 0
postgresql:
  parameters:
    max_connections: 100
    max_locks_per_transaction: 64
    max_prepared_transactions: 0
    max_replication_slots: 10
    max_wal_senders: 10
    max_worker_processes: 8
    track_commit_timestamp: 'off'
    wal_keep_segments: 64
    wal_level: logical
    wal_log_hints: 'off'
  use_pg_rewind: false
  use_slots: false
retry_timeout: 90
synchronous_mode: true
ttl: 200
```

**Have you checked Patroni logs?**
```
2020-11-16 13:34:52,439 INFO: no action.  i am the leader with the lock
2020-11-16 13:35:02,386 INFO: Lock owner: connections-profile-gato-clr-c-0; I am connections-profile-gato-clr-c-0
2020-11-16 13:35:02,440 WARNING: Concurrent update of connections-profile-gato-clr
2020-11-16 13:35:03,441 ERROR: failed to update leader lock
2020-11-16 13:35:03.442 UTC [3481] LOG:  received immediate shutdown request
2020-11-16 13:35:03.444 UTC [20085] DETAIL:  The postmaster has commanded this server process to roll back the current transaction and exit, because another server process exited abnormally and possibly corrupted shared memory.
...
2020-11-16 13:35:03.460 UTC [3481] LOG:  database system is shut down
[...]

PR #1796, opened 2020-12-14
Improve handling of K8s concurrent update error

CyberDem0n:
The old strategy was waiting for 1 second and hoping that we will get an update event from the WATCH connection.
Unfortunately, it didn't work well in practice. Instead, we will get the current value from the API by performing an explicit read request.

Close https://github.com/zalando/patroni/issues/1767


FxKu:
:+1:

CyberDem0n:
:+1:

PR #1797, opened 2020-12-14
 Fix small issues with ignore-slots feature

CyberDem0n:
When there is no config key in DCS Patroni shouldn't try accessing ignore_slots, otherwise an exception is raised.

In addition to that implement missing unit-tests and fix linting issues in behave tests.


FxKu:
:+1:

CyberDem0n:
:+1:

Issue #1785, opened 2020-12-10
patronictl list command exception handling


cobolbaby:
**Describe the bug**

`patronictl list` command sometimes throw exception:

```
postgres@pg01:~$ patronictl list
+ Cluster: pgcluster_itc (6728244408812797982) +-----------+
| Member | Host       | Role    | State   | TL | Lag in MB |
+--------+------------+---------+---------+----+-----------+
| pg01   | 10.13.0.18 | Leader  | running | 92 |           |
| pg02   | 10.13.0.20 | Replica | running | 92 |         0 |
| pg03   | 10.13.0.30 | Replica | running | 92 |         0 |
+--------+------------+---------+---------+----+-----------+
Exception ignored in: <bound method EtcdClient.__del__ of <patroni.dcs.etcd.EtcdClient object at 0x7ff6fd6624e0>>
Traceback (most recent call last):
  File "/usr/local/lib/python3.5/dist-packages/patroni/dcs/etcd.py", line 406, in __del__
  File "/usr/local/lib/python3.5/dist-packages/urllib3/poolmanager.py", line 222, in clear
  File "/usr/local/lib/python3.5/dist-packages/urllib3/_collections.py", line 100, in clear
  File "/usr/local/lib/python3.5/dist-packages/urllib3/poolmanager.py", line 173, in <lambda>
  File "/usr/local/lib/python3.5/dist-packages/urllib3/connectionpool.py", line 494, in close
AttributeError: 'NoneType' object has no attribute 'Empty'
```

**Environment**
- Patroni version: 2.0.1
- PostgreSQL version: 10.15
- DCS (and its version): Etcd

vladloskut:
Got the same error:

```
root@pg_node_2:~# patronictl -c /etc/patroni/patroni.yml list
+ Cluster: patroni_lsk_2 (6916504509894852761) --+----+-----------+-----------------+
| Member    | Host           | Role    | State   | TL | Lag in MB | Pending restart |
+-----------+----------------+---------+---------+----+-----------+-----------------+
| pg_node_1 | 10.105.241.135 | Replica | running | 12 |         0 | *               |
| pg_node_2 | 10.105.241.137 | Leader  | running | 12 |           | *               |
+-----------+----------------+---------+---------+----+-----------+-----------------+
Exception ignored in: <bound method EtcdClient.__del__ of <patroni.dcs.etcd.EtcdClient object at 0x7ff5174b7710>>
Traceback (most recent call last):
[...]

PR #1801, opened 2020-12-15
Handle AttributeError in etcd.py

CyberDem0n:
When the__del__() method is executed the python interpreter already unloaded some of the modules that are still used down the http.clear() method.
The only we could do in this case - silence some exceptions like ReferenceError, TypeError, AttributeError.

Close https://github.com/zalando/patroni/issues/1785


FxKu:
:+1:

CyberDem0n:
:+1:

PR #1802, opened 2020-12-15
Silence unhandled exceptions in Thread.run() during unit-tests

CyberDem0n:
Python 3.8 changed the way how exceptions raised from the Thread.run() method are handled.
It resulted in unit-tests showing a couple of warnings. They are not important and we just silence them.


FxKu:
:+1:

CyberDem0n:
:+1:

Issue #1805, opened 2020-12-31
Error at startup when only 2 out of 3 running etcd3 servers


tintin759:
Hi all,
**The problem**
I'm facing some problem when starting patroni with this configuration:
- 3 etcd3 nodes but only 2 out of 3 running (192.168.1.126 isn't running)

Patroni will sometimes refuse to start and sometimes it starts without any problem.
Thank you for your help.

**To Reproduce**
1. configure 3 etcd nodes but only 2 out of 3 running
2. start the first patroni node multiple times until success

**Expected behavior**
It should always start with 2 etcd alived. Am I wrong ?

**Environment**
- Patroni version: 2.0.1
- PostgreSQL version: 11
- DCS (and its version): etcd 3.4.14

**Patroni configuration file**
```
etcd3:
  hosts:
  - 192.168.1.128:2379 
  - 192.168.1.126:2379
  - 192.168.1.127:2379
  username: root
  password: ChANgEmE
  protocol: http
  cacert: None
tags:
    nofailover: false
    noloadbalance: false
    clonefrom: false 
    nosync: false

```

**patronictl show-config**
```
loop_wait: 10
maximum_lag_on_failover: 1048576
postgresql:
  parameters: null
  use_pg_rewind: true
retry_timeout: 10
ttl: 30
```

**Have you checked Patroni logs?**
Here are the logs when it refuses to start:
```
2020-12-31 10:47:34,486 INFO: Selected new etcd server http://192.168.1.126:2379
2020-12-31 10:47:34,487 INFO: Trying to authenticate on Etcd...
2020-12-31 10:47:35,898 INFO: Selected new etcd server http://192.168.1.127:2379
2020-12-31 10:47:35,899 ERROR: Request to server http://192.168.1.126:2379 failed: MaxRetryError("HTTPConnectionPool(host='192.168.1.126', port=2379): Max retr
ies exceeded with url: /v3/auth/authenticate (Caused by NewConnectionError('<urllib3.connection.HTTPConnection object at 0x7fc9124f0518>: Failed to establish a
 new connection: [Errno 113] No route to host',))",)
2020-12-31 10:47:35,900 INFO: Reconnection allowed, looking for another server.
2020-12-31 10:47:35,901 CRITICAL: Etcd3 authentication failed: EtcdException('POST /v3/auth/authenticate request failed',)
```

**Additional context**
My setup is a cluster of 3 VM:
[...]

PR #1806, opened 2021-01-04
Exit only if authentication explicitly failed

CyberDem0n:
It could happen that one of etcd servers is not accessible on Patroni start.
In this case Patroni was trying to perform authentication and exiting, while it should exit only if Etcd explicitly responded with the
AuthFailed error.

Close https://github.com/zalando/patroni/issues/1805


FxKu:
:+1:

CyberDem0n:
:+1:

Issue #1810, opened 2021-01-08
Usage of python-consul


wilfriedroset:
**Describe the bug**
When it comes to consul dcs, patroni leverage `python-consul`, as per [requirements.txt](https://github.com/zalando/patroni/blob/master/requirements.txt#L8). This library is inactive for [more than 2 years now](https://github.com/cablehead/python-consul/commits/master).

For several other usecases related to consul we have been using happily [python-consul2](https://github.com/poppyred/python-consul2) is actively maintained. Users starts moving away from `python-consul` and there is a PR to warn about it: https://github.com/hashicorp/consul/pull/9079

While working on the setup of patroni on consul with `python-consul2`, we have stumble upon the traceback below.
Everything works has expected with `python-consul`.

Before diving in the code I would like to have your thoughts about that. Do you have plan to move away from `python-consul`?

I can see several solutions:
* leave it as it is and warn about it in the documentation
* completely move away the unsupported version
* support both `python-consul` and `python-consul2` (for how long?)
* work with @poppyred to be able to use `python-consul2` as a drop-in replacement which works with patroni

**To Reproduce**
use patroni with python-consul2 instead of python-consul

**Expected behavior**
A working patroni just like with python-consul

**Environment**
- os: debian 9
- patroni version: 2.0.1 from pgdg
- postgresql version: 9.6.20 from pgdg
- python-consul: 0.7.1
- python-consul2: 0.1.1
- consul: 1.8.5

**Patroni configuration file**
```yml
consul:
  host: http://localhost:8500
  consistency: consistent
```

**Have you checked Patroni logs?**
```
Jan 07 15:58:52 redacted-hostname systemd[1]: Started Runners to orchestrate a high-availability PostgreSQL.
Jan 07 15:58:52 redacted-hostname patroni[19400]: Traceback (most recent call last):
Jan 07 15:58:52 redacted-hostname patroni[19400]:   File "/usr/bin/patroni", line 11, in <module>
Jan 07 15:58:52 redacted-hostname patroni[19400]:     load_entry_point('patroni==2.0.1', 'console_scripts', 'patroni')()
Jan 07 15:58:52 redacted-hostname patroni[19400]:   File "/usr/lib/python3/dist-packages/patroni/__init__.py", line 170, in main
Jan 07 15:58:52 redacted-hostname patroni[19400]:     return patroni_main()
[...]

PR #1812, opened 2021-01-11
Add compatibility with python-consul2

CyberDem0n:
the good old python-consul is not maintained for a few years in a row, therefore someone forked under a different name, but package files are installed into the same location as for the old.

The API of both modules is mostly compatible therefore it wasn't hard to add the support of both modules in Patroni.

Taking into account that python-consul is not a direct requirement for Patroni, but extra, now the end-user has a choice what to install.

Close https://github.com/zalando/patroni/issues/1810

### DIFF
--- a/patroni/dcs/consul.py
+++ b/patroni/dcs/consul.py
@@ -8,6 +8,7 @@ import ssl
 import time
 import urllib3
 
+from collections import namedtuple
 from consul import ConsulException, NotFound, base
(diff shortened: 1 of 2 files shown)


wilfriedroset:
Thank you for this. I've successfully tested your patch. It works has expected.

FxKu:
:+1:

CyberDem0n:
:+1:

Issue #1817, opened 2021-01-18
WAL folder softlinks can be easily wiped on reinitialize


kmoppel:
**Describe the bug**
Currently if Patroni takes over an existing cluster with a linked pg_wal folder by default the link will not be preserved after a manual reinit request. This could lead to quite dangerous situations where this small change is not detected initially as everything will keep working but later volumes could get filled, performance suffer etc.

NB! Patroni will work correctly though if the `--waldir` option is specified in `postgresql.baseback` section, but it's only available from v10.

**Proposed fix**
I was thinking of adding below items - does it sound OK?
1. A check rejecting the reinit request in API handler if currently WAL is linked but the basebackup sections does not provide the `--waldir` option.
https://github.com/zalando/patroni/blob/master/patroni/api.py#L339
2. An according `--waldir` line in the sample postgres1.yml config.

CyberDem0n:
Hi @kmoppel,

this is so rare case when people "taking over" existing clusters, that I don't want to invest my time working on it.

IMO it should be just enough to add a few words to documentation about this potential issue. Adding further possible arguments (with comments) to the basebackup section, also will raise awareness.

kmoppel:
Hey @CyberDem0n, thanks for the quick response! Btw, seen from the consulting world it's a very common use case to "patronify" existing plain instances or moving away from Pacemaker. And I could also do the coding...just wanted to discuss what would be the best approach to tackle this. But if you think changing code is overkill I'll just update the documentation, I won't step on the same rake again for sure.

PR #1818, opened 2021-01-18
Docs: document the need for special configuration for symlinked pg_wal

kmoppel:
If an existing instance was configured with WAL residing outside of
PGDATA then currently a 'reinit' would lose such symlinks. So add some
bits of information on that to draw attention to this cornercase issue
and also add the --waldir option to the sample `postgresql.basebackup`
configuration sections to increase visibility.

Discussion: https://github.com/zalando/patroni/issues/1817


sdudoladov:
:+1:

CyberDem0n:
:+1:

CyberDem0n:
Merged, thank you @kmoppel!

PR #1825, opened 2021-01-29
Add support for encrypted TLS keys for REST API

jkatz:
The Python SSL library allows for the inclusion of a password
in its [`load_cert_chain`](https://docs.python.org/3/library/ssl.html#ssl.SSLContext.load_cert_chain) function when setting up a `SSLContext`.
This allows for loading an encrypted key file in PEM representation
to be loaded into the certificate chain.

This commit adds the optional `keyfile_password` parameter to
the REST API block of configuration so that Patroni can load in
encrypted private keys when establishing its TLS socket.

This also adds the corollary `PATRONI_RESTAPI_KEYFILE_PASSWORD`
environmental variable, which has the same effect.


sdudoladov:
:+1:

CyberDem0n:
:+1:

CyberDem0n:
Merged, thank you @jkatz!

jkatz:
@CyberDem0n Thank you for accepting the PR! Happy to contribute.

I may have a few more in this area in the near future :smile:

PR #1830, opened 2021-02-01
Don't use bypass_api_service when running patronictl

CyberDem0n:
It could happen that the cluster role wither not configured or doesn't provide enough permissions. In this case bypass_api_service is ignored, but the warning is logged, which is rather annoying when patronictl is used.
Since the bypass_api_service is most useful for Patroni, we will simply ignore it when patronictl is used.


sdudoladov:
:+1:

CyberDem0n:
:+1:

PR #1865, opened 2021-03-05
Add support for patroni replication user socket connection

melrifa:
Issue: https://github.com/zalando/patroni/issues/1866


CyberDem0n (commented):
There is a typo in the documentation.
Also, I reshuffled code a bit in the resolve_connection_addresses().

edysli:
I've tried this patch and it allows me to work around #1884, using UNIX sockets instead of a TLS connection to localhost for the replication user. :smiley:

FxKu:
:+1:

CyberDem0n:
:+1:

Issue #1873, opened 2021-03-10
Inconsistent heartbeat status logging


RajKiranS:
Hello, We upgraded from Patroni-1.6.5 to Patroni-2.0.1. And post the upgrade we started noticing inconsistent logs for heartbeat status.

For example, here are the logs from the leader node:
```
2021-03-10 06:08:44,891 INFO: Lock owner: node-01.xyz.com; I am node-01.xyz.com
2021-03-10 06:08:44,892 INFO: no action.  i am the leader with the lock
2021-03-10 06:08:54,891 INFO: Lock owner: node-01.xyz.com; I am node-01.xyz.com
2021-03-10 06:08:54,892 INFO: no action.  i am the leader with the lock
2021-03-10 06:09:04,891 INFO: Lock owner: node-01.xyz.com; I am node-01.xyz.com
2021-03-10 06:09:04,901 INFO: no action.  i am the leader with the lock
2021-03-10 06:09:04,923 INFO: Lock owner: node-01.xyz.com; I am node-01.xyz.com
2021-03-10 06:09:04,924 INFO: no action.  i am the leader with the lock
2021-03-10 06:09:14,902 INFO: Lock owner: node-01.xyz.com; I am node-01.xyz.com
2021-03-10 06:09:14,904 INFO: no action.  i am the leader with the lock
2021-03-10 06:09:24,902 INFO: Lock owner: node-01.xyz.com; I am node-01.xyz.com
2021-03-10 06:09:24,916 INFO: no action.  i am the leader with the lock
2021-03-10 06:09:24,938 INFO: Lock owner: node-01.xyz.com; I am node-01.xyz.com
2021-03-10 06:09:24,940 INFO: no action.  i am the leader with the lock
2021-03-10 06:09:34,918 INFO: Lock owner: node-01.xyz.com; I am node-01.xyz.com
2021-03-10 06:09:34,920 INFO: no action.  i am the leader with the lock
```
Logs from the replica node:
```
2021-03-10 06:08:04,871 INFO: Lock owner: node-01.xyz.com; I am node-02.xyz.com
2021-03-10 06:08:04,871 INFO: does not have lock
2021-03-10 06:08:04,876 INFO: no action.  i am a secondary and i am following a leader
2021-03-10 06:08:04,894 INFO: Lock owner: node-01.xyz.com; I am node-02.xyz.com
2021-03-10 06:08:04,894 INFO: does not have lock
[...]

PR #1875, opened 2021-03-10
Fix excessive HA loop runs with Zookeeper

CyberDem0n:
1. Commit 04b9fb9dd4c082a7a009e50e67641191e0ca9a5d introduced additional conditions for updating cached version of the leader optime. It was required for implementing health-checks based on replication lag in the https://github.com/zalando/patroni/pull/1599.
  What in fact was forgotten, the event should be cleared after the new value of the optime was fetched. Not doing so results in running the HA loop more frequently than is required.
  
2. Don't watch for sync members.
  The watch for sync member(s) was introduced in order to give a signal to the leader that one of the members set the `nosync` tag to true.
  Since that time we have got a few more conditions that should be notified about, therefore instead of watching for all members of the cluster every cluster member checks whether the condition is met, and instead of updating ZNode performs delete+create.
  Since every member is already watching for new ZNodes to be created inside the $scope/members/, they automatically get notified about important changes, and therefore watching for sync members is redundant.

3. In addition to that, slightly increase watch timeout, it will keep HA loops in sync across all nodes in the cluster.
Close https://github.com/zalando/patroni/pull/1873


avandras:
👍

FxKu:
:+1:

CyberDem0n:
:+1:

PR #1887, opened 2021-03-18
fix: reload api if certificate changed on disk

### LINKED ISSUE #1886: Hot certificate change (reload patroni?)
Hello,
I'm securing the rest api with TLS+auth for my lovely patroni 2.0.2 (Debian 10 with python 3.7.3).

Is there a way to reload patroni in order to switch the certificate? I tried:
```
systemctl reload patroni
patronictl -c /etc/patroni/ reload cluster_name node_name
patronictl -c /etc/patroni/ restart cluster_name node_name
``` 
with no effect the certificate being used. I'm checking the cert with
```
openssl s_client -connect node_name.domain.com:8008 </dev/null 2>/dev/null | openssl x509 -noout -text | grep -E 'Subject:|DNS:'
```

The only thing that made it change is a violent
```
systemctl restart patroni
```
which makes the cluster to switch.

Is there another way to reload the certificate? I may try to provide a PR for this if it's not supported yet.

Thanks!

michael-todorovic:
This PR fixes #1886.

We get the certificate serial number on server startup and store it in `api.__ssl_serial_number`
On reload, we get again the serial number from disk and compare it to the one stored in `api.__ssl_serial_number`: if different, then the api will be reloaded (even if the config file didn't change)

### DIFF
--- a/patroni/__init__.py
+++ b/patroni/__init__.py
@@ -71,7 +71,7 @@ class Patroni(AbstractPatroniDaemon):
     def reload_config(self, sighup=False, local=False):
         try:
             super(Patroni, self).reload_config(sighup, local)
-            if local:
+            if local or self.api.reload_local_certificate():
                 self.tags = self.get_tags()
                 self.request.reload_config(self.config)
                 self.api.reload_config(self.config['restapi'])
--- a/patroni/__init__.py
+++ b/patroni/__init__.py
@@ -71,9 +71,10 @@ class Patroni(AbstractPatroniDaemon):
     def reload_config(self, sighup=False, local=False):
         try:
             super(Patroni, self).reload_config(sighup, local)
-            if local or self.api.reload_local_certificate():
+            if local:
                 self.tags = self.get_tags()
                 self.request.reload_config(self.config)
+            if local or self.api.reload_local_certificate():
                 self.api.reload_config(self.config['restapi'])
             self.watchdog.reload_config(self.config)
             self.postgresql.reload_config(self.config['postgresql'], sighup)
--- a/patroni/api.py
+++ b/patroni/api.py
@@ -703,6 +703,7 @@ class RestApiServer(ThreadingMixIn, HTTPServer, Thread):
 
         self.__listen = listen
         self.__ssl_options = ssl_options
+        self._received_new_cert = False # reset to False after reload_config()
 
         self.__httpserver_init(host, port)
         Thread.__init__(self, target=self.serve_forever)
--- a/patroni/api.py
+++ b/patroni/api.py
@@ -616,6 +616,8 @@ class RestApiServer(ThreadingMixIn, HTTPServer, Thread):
         self.http_extra_headers = {}
         self.reload_config(config)
         self.daemon = True
+        self.__ssl_serial_number = None
+        self._received_new_cert = False
 
     def query(self, sql, *params):
         cursor = None
@@ -723,6 +725,7 @@ class RestApiServer(ThreadingMixIn, HTTPServer, Thread):
                     ctx.verify_mode = modes[verify_client]
                 else:
                     logger.error('Bad value in the "restapi.verify_client": %s', verify_client)
+            self.__ssl_serial_number = self.get_certificate_serial_number()
             self.socket = ctx.wrap_socket(self.socket, server_side=True)
         if reloading_config:
             self.start()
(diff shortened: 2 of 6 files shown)


patroni/api.py
```diff
@@ -750,6 +753,27 @@ def shutdown_request(self, request):
...
         return super(RestApiServer, self).shutdown_request(request)
 
+    def get_certificate_serial_number(self):
+        from cryptography import x509, hazmat
+        try:
+            with open(self.__ssl_options['certfile'], "rb") as f:
+                crt = x509.load_pem_x509_certificate(f.read(), hazmat.backends.default_backend())
+                return crt.serial_number
+        except EnvironmentError:
+            logger.exception("Failed to get serial number from certificate %s", self.__ssl_options['certfile'])
```
CyberDem0n, line 763: I don't like pulling unnecessary dependencies, they are heavy and also create stress for packagers.

Below is an alternative that relies only on the `ssl` module. 
```suggestion
        if self.__ssl_options.get('certfile'):
            import ssl
            try:
                crt = ssl._ssl._test_decode_cert(self.__ssl_options['certfile'])
                return crt.get('serialNumber')
            except ssl.SSLError as e:
                logger.error('Failed to get serial number from certificate %s: %r', self.__ssl_options['certfile'], e)
```

Yes, I know, `ssl._ssl._test_decode_cert()` is undocumented, but it works on all versions of python.

patroni/daemon.py
```diff
@@ -54,7 +54,11 @@ def run(self):
         while not self.received_sigterm:
             if self._received_sighup:
                 self._received_sighup = False
-                self.reload_config(True, self.config.reload_local_configuration())
+                if self.config.reload_local_configuration() or self.api.reload_local_certificate():
+                    reload = True
+                else:
+                    reload = False
+                self.reload_config(True, reload)
```
CyberDem0n, line 61: This is not really the right place (abstract class, which is also used by patroni_raft_controller).
The `api.reload_conf()` should be called from the https://github.com/zalando/patroni/blob/master/patroni/__init__.py#L77
Condition would be: `if local or self.api.reload_local_certificate()`.

patroni/api.py
```diff
@@ -750,6 +753,27 @@ def shutdown_request(self, request):
...
+        if self.__protocol == 'https':
+            on_disk_cert_serial_number = self.get_certificate_serial_number()
+            if on_disk_cert_serial_number != self.__ssl_serial_number:
+                self._received_new_cert = True
+                self.__ssl_serial_number = on_disk_cert_serial_number
+                return True
+            else:
+                return False
+        else:
+            return False
```
CyberDem0n, line 775: These else branches are redundant...
By default None is returned, which is treated as not True.

patroni/__init__.py
```diff
@@ -71,7 +71,7 @@ def nosync(self):
     def reload_config(self, sighup=False, local=False):
         try:
             super(Patroni, self).reload_config(sighup, local)
-            if local:
+            if local or self.api.reload_local_certificate():
```
CyberDem0n, line 74: This condition should apply only to `self.api.reload_config(self.config['restapi'])`.
The remaining two lines should still be executed only `if local:`

patroni/api.py
```diff
@@ -750,6 +753,23 @@ def shutdown_request(self, request):
...
+                crt = ssl._ssl._test_decode_cert(self.__ssl_options['certfile'])
+                return crt.get('serialNumber')
+            except ssl.SSLError as e:
+                logger.error('Failed to get serial number from certificate %s: %r', self.__ssl_options['certfile'], e)
+
+    def reload_local_certificate(self):
+        if self.__protocol == 'https':
+            on_disk_cert_serial_number = self.get_certificate_serial_number()
+            if on_disk_cert_serial_number != self.__ssl_serial_number:
+                self._received_new_cert = True
```
CyberDem0n, line 769: The state (`self._received_new_cert`) is never reset after being set to True.
I think the proper place to do it would be after the self.__initialize() call: https://github.com/zalando/patroni/pull/1887/files#diff-13329cd3342d1757c2c9b4dd7a26c0170f4deee5d32eff5c2fd719cc9ff9375bR787

patroni/__init__.py
michael-todorovic: That's right, sorry for this dumb thing

FxKu:
:+1:

CyberDem0n:
:+1:

PR #1888, opened 2021-03-21
Don't create pgpass dir if kerberos auth is used

### LINKED ISSUE #1551: Disable creation of PGPASS if using kerberos
Hello,

It doesn't make sense to create `PGPASS` file if we are using [kerberos](https://github.com/zalando/patroni/pull/1015). Can we disable in that case?

PGPASS is being created in [here](https://github.com/zalando/patroni/blob/881bba9e1cd081f7cd4c340f9a6ad977daa92c58/patroni/postgresql/config.py#L369).

I know that we don't populate or use PGPASS when kerberos is enabled, but creating parent dir of default PGPASS file could be a trouble when the default(~) directory doesn't exist on a node.

kostiantyn-nemchenko:
Fixes #1551

### DIFF
--- a/patroni/postgresql/config.py
+++ b/patroni/postgresql/config.py
@@ -336,8 +336,9 @@ class ConfigHandler(object):
         if "stats_temp_directory" in self._server_parameters:
             self.try_to_create_dir(self._server_parameters["stats_temp_directory"],
                                    "'{}' is defined in stats_temp_directory, {}")
-        self.try_to_create_dir(os.path.dirname(self._pgpass),
-                               "'{}' is defined in `postgresql.pgpass`, {}")
+        if "krbsrvname" not in self._config:
+            self.try_to_create_dir(os.path.dirname(self._pgpass),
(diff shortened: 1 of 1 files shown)


patroni/postgresql/config.py
```diff
@@ -336,8 +336,9 @@ def check_directories(self):
         if "stats_temp_directory" in self._server_parameters:
             self.try_to_create_dir(self._server_parameters["stats_temp_directory"],
                                    "'{}' is defined in stats_temp_directory, {}")
-        self.try_to_create_dir(os.path.dirname(self._pgpass),
-                               "'{}' is defined in `postgresql.pgpass`, {}")
+        if "krbsrvname" not in self._config:
```
CyberDem0n, line 339: I think it would be more appropriate to use `self._krbsrvname`:
```suggestion
        if not self._krbsrvname:
```

FxKu:
:+1:

CyberDem0n:
:+1:

Issue #1860, opened 2021-02-26
question for patroni auto failover issue


louteq:
**Describe the bug**
Hi team, we have thousands of postgres instance managed by patroni, so for such amount instances we always try to handle instance crash/machine shutdown... issue automatically.

But one scenario we found patroni can't auto failover, can't elet new leader though no any lag on standby. I saw one friend submitted same issue recently. the link is : https://github.com/zalando/patroni/issues/1858

I checked it but not find a solution , seems it is already closed.

under some circumstances  our postgres instance server will shutdown for many reasons, we found patroni can't elect new leader even though there's no any lag between primary and standby, so no data loss. under such cases, we hope patroni can elect new leader, if not , all instance will become replica and read only, our business will be impacted. And we have to check and handle manually, it's very time consuming.

but if we kill patroni process or kill -9 patroni process, new leader can select out.

After investigation, I guess it is caused as when machine shutdown , postgres will do a checkpoint and wal senders were teminated , so lead patroni checked wal lag behind one wal segment size.

but same time, when compared with kill patroni process, we found postgres also will do a checkpoint.

below is postgres log for above 2 scenarios:
*****kill patroni process, can elect new cluster leader*****
2021-02-26 16:41:15.519 CST,,,2246,,6038b139.8c6,3,,2021-02-26 16:28:41 CST,,0,LOG,00000,"received fast shutdown request",,,,,,,,,""
2021-02-26 16:41:15.520 CST,,,2246,,6038b139.8c6,4,,2021-02-26 16:28:41 CST,,0,LOG,00000,"aborting any active transactions",,,,,,,,,""
2021-02-26 16:41:15.520 CST,"postgres","postgres",2262,"127.0.0.1:49178",6038b13a.8d6,3,"idle",2021-02-26 16:28:42 CST,4/0,0,FATAL,57P01,"terminating connection due to administrator command",,,,,,,,,"Patroni"
2021-02-26 16:41:15.520 CST,"postgres","postgres",2262,"127.0.0.1:49178",6038b13a.8d6,4,"idle",2021-02-26 16:28:42 CST,,0,LOG,00000,"disconnection: session time: 0:12:32.596 user=postgres database=postgres host=127.0.0.1 port=49178",,,,,,,,,"Patroni"
[...]

PR #1889, opened 2021-03-22
Write prev LSN before the checkpoint to optime if achive_mode=on

CyberDem0n:
The #1527 introduced a feature of updating `/optime/leader` with the location of the last checkpoint after the Postgres was shutdown cleanly.

If wal archiving is enabled, Postgres always switching the WAL file before writing the checkpoint shutdown record. Normally it is not an issue, but for databases without too much write activity it could lead to the situation that the visible replication lag becomes equal to the size of a single WAL file. In fact, the previous WAL file is mostly empty and contains only a few records.

Therefore it should be safe to report the LSN of the SWITCH record before the shutdown checkpoint.
In order to do that, Patroni first gets the output of the pg_controldata and based on it calls pg_waldump two times:
* The first call reads the checkpoint record (and verifies that this is really the shutdown checkpoint).
* The next call reads the previous record and in case if it is the 'xlog switch' (for 9.3 and 9.4) or 'SWITCH' (for 9.5+), the LSN
of the SWITCH record is written to the `/optime/leader`.

In case of any mismatch, failure to call pg_waldump or parse its output, the old behavior is retained, i.e. `Latest checkpoint location` from the pg_controldata is used.

Close https://github.com/zalando/patroni/issues/1860


sdudoladov:
:+1:

CyberDem0n:
:+1:

Issue #1870, opened 2021-03-09
Database is not restored to recovery point (ignored recovery_target_time)


vasilievs:
**Describe the bug**
I tries to created | restore PITR backup for patroni cluster.
Cluster restored **latest state in wal files**, and not to the desired point (recovery_target_time)
base_backup 2021-03-04 16:27:03 
need to recovery 2021-03-04 16:32:20
clueter restored: 2021-03-04 16:32:20

**To Reproduce**
I set up a shared NFS folder for wal files (use archive_command).
```
  parameters:
    archive_mode: on
    archive_command: 'test ! -f /mnt/nfs_pg_backup/%f && cp %p /mnt/nfs_pg_backup/%f'

```
after that I made several transactions, switched the master.
Created base backup
`pg_basebackup -U replication -h localhost -D /mnt/pg_backup/base_backup/"$(date +%Y-%m-%d_%H-%M)" -Ft -Xf -Z7`

made a few more transactions, and run pg_switch_wal() and copied the wal files in difficult folder.
```
  PGPASSWORD=postgres psql -U postgres -h $HOSTNAME -c "select pg_switch_wal();"
  rsync /mnt/nfs_pg_backup/* $d
```

after that i tries to recovery patroni cluster to PITR.

delete existing data and cluster:
```
systemctl stop patroni
rm -rf /var/lib/pgsql/12/data/*
patronictl remove patroni
```
Add recovery conf on postgres section config patroni (not bootstap)
vi /etc/patroni/patroni.yml
```
        restore_command: 'cp /tmp/restore_point/2021-03-04_16/%f %p'
        recovery_target_time: '2021-03-04 16:32:20'
        recovery_target_inclusive: true

```
create recovery signal:
```
touch /var/lib/pgsql/12/data/recovery.signal
chown -R postgres:postgres /var/lib/pgsql/12/data/recovery.signal
```
and start patroni:
`systemctl start patroni`

Cluster restored **latest state in wal files**, and not to the desired point (recovery_target_time)
base_backup 2021-03-04 16:27:03 
need to recovery 2021-03-04 16:32:20
clueter restored: 2021-03-04 16:32:20

 in recovery_target_time i tried to set formats:
 UTC, timezone etc ( not results)

**Environment**
- Patroni version: patronictl version 1.6.5
- PostgreSQL: 12
- DCS (and its version): consul (Consul v1.8.0)

**Have you checked PostgreSQL logs?**
```
[...]

PR #1891, opened 2021-03-25
Fix little issues with custom bootstrap

CyberDem0n:
1. Set hot_standby=off only when we do PITR
2. Restart postgres after PITR is done to avoid warnings
3. Address invalid config issue https://github.com/zalando/patroni/issues/1870#issuecomment-800088643


FxKu:
:+1:

CyberDem0n:
:+1:

df7cb:
Hi,
could we have a new release that includes this fix? A customer is running into that problem when a custom bootstrap method is in use (which the Debian package is using via `pg_createcluster_patroni`).

The symptom is that after the first switchover, the old leader is connecting to the new leader, but is stuck in "starting" because it has `hot_standby=off` in the config. Restarting PostgreSQL does not help; restarting patroni fixes the problem and the old leader can connect successfully to the new leader.

Thanks!

Issue #1899, opened 2021-04-07
allow-group-access clause in initdb seems not woking correctly


dan-aksenov:
**Describe the bug**
When using allow-group-access clause in initdb still have some files stlil created with 600 permisstions
```
find /data/postgresql/11/data/ -type f  -perm 600
/data/postgresql/11/data/patroni.dynamic.json
/data/postgresql/11/data/postgresql.base.conf.backup
/data/postgresql/11/data/postgresql.conf.backup
/data/postgresql/11/data/recovery.done
```
As far as I know first 3 files are definitely created by patroni. recovery.* file also appears after restarting patroni daemon.

**To Reproduce**
1. Create patroni config with allow-group-access option
```
initdb:
    - allow-group-access
```
2. (Re)initialize cluster.
3. Observe file permissions in data directory.

**Expected behavior**
All files mast have group access permissions

**Environment**
- Patroni version: patroni 1.6.5
- PostgreSQL version: PostgreSQL 11.8 on x86_64-pc-linux-gnu, compiled by gcc (Ubuntu 7.5.0-3ubuntu1~18.04) 7.5.0, 64-bit
- DCS (and its version): etcd Version: 3.2.17

I am not allowed to upgrade PostgreSQL and Patroni versions right now.

**Patroni configuration file**
Not sure it's needed. Please let me know if it is.

**patronictl show-config**
```
patronictl -c /etc/patroni/config.yml show-config
loop_wait: 10
maximum_lag_on_failover: 1048576
postgresql:
  parameters:
    max_connections: '200'
    max_locks_per_transaction: '64'
    max_prepared_transactions: '3'
    max_worker_processes: '1'
    track_commit_timestamp: false
    wal_level: replica
    wal_log_hints: true
  use_pg_rewind: true
  use_slots: true
retry_timeout: 10
ttl: 30
```

**Have you checked Patroni logs?**
Not sure it's needed. Please let me know if it is.

**Have you checked PostgreSQL logs?**
Not sure it's needed. Please let me know if it is.

**Have you tried to use GitHub issue search?**
Was unable to find anything usefull for allow-group-access

CyberDem0n:
Good catch.
But, in order to fix it we should not rely on `--allow-group-access` option of initdb, but rather set `umask` based on permissions of PGDATA. This is exactly what Postgres does.

Do you want to work on that?

dan-aksenov:
If "work on that" means pull request, I'm afraid I dont have enough skill for the job.

[...]

PR #1901, opened 2021-04-08
Set permissions for config files that created by Patroni as mode from PGDATA dir

anikin-aa:
Fixes: https://github.com/zalando/patroni/issues/1899

Right now tests are missing, may be it's worth to add them. 

@CyberDem0n, please, advise

### DIFF
--- a/patroni/config.py
+++ b/patroni/config.py
@@ -111,6 +111,9 @@ class Config(object):
         self._load_cache()
         self._cache_needs_saving = False
 
+    def __get_permission_for_data_dir(self):
+        return int(oct(os.stat(self._data_dir).st_mode), base=8)
+
     @property
     def config_file(self):
         return self._config_file
@@ -166,6 +169,7 @@ class Config(object):
                     fd = None
                     json.dump(self.dynamic_configuration, f)
                 tmpfile = shutil.move(tmpfile, self._cache_file)
+                os.chmod(self._cache_file, self.__get_permission_for_data_dir())
                 self._cache_needs_saving = False
             except Exception:
                 logger.exception('Exception when saving file: %s', self._cache_file)
--- a/patroni/postgresql/config.py
+++ b/patroni/postgresql/config.py
@@ -320,6 +320,9 @@ class ConfigHandler(object):
         self._recovery_params = {}
         self.reload_config(config)
 
+    def __get_permission_for_data_dir(self):
+        return int(oct(os.stat(self._postgresql.data_dir).st_mode), base=8)
+
     def setup_server_parameters(self):
         self._server_parameters = self.get_server_parameters(self._config)
         self._adjust_recovery_parameters()
@@ -364,6 +367,7 @@ class ConfigHandler(object):
                     backup_file = os.path.join(self._postgresql.data_dir, f + '.backup')
                     if os.path.isfile(config_file):
                         shutil.copy(config_file, backup_file)
+                        os.chmod(backup_file, self.__get_permission_for_data_dir())
             except IOError:
                 logger.exception('unable to create backup copies of configuration files')
         return True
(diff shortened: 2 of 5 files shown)


patroni/config.py
```diff
@@ -111,6 +111,9 @@ def __init__(self, configfile, validator=default_validator):
         self._load_cache()
         self._cache_needs_saving = False
 
+    def __get_permission_for_data_dir(self):
+        return int(oct(os.stat(self._data_dir).st_mode), base=8)
```
CyberDem0n, line 115: I afraid it will not work on Windows

anikin-aa: yep, forgot about Windows case, will fix it

anikin-aa: actually, it works for me on Windows machine, or may be I am missing something ?

```
import os
import sys

if __name__ == '__main__':
    print(sys.platform)
    f = open("./test", "a")
    f.close()
    os.chmod("./test", 0o777)
    print(oct(os.stat("./test").st_mode))
    rights = int(oct(os.stat("./test").st_mode), base=8)
    print(rights)
    os.chmod("./test", int(oct(os.stat("./test").st_mode), base=8))
```

And the output:
```
win32
0o100666
33206
```

CyberDem0n: Oh, true, I mixed it up with some other issue.
Anyway, we should not copy directory permissions as it is, but replicate https://github.com/postgres/postgres/blob/REL_13_STABLE/src/common/file_perm.c#L33-L50
That is, set group access only when it is necessary and by default only user should be allowed.

CyberDem0n: @anikin-aa please take a look at my last comment https://github.com/zalando/patroni/pull/1901#discussion_r609657320

anikin-aa: @CyberDem0n , yep, i see. Sorry for long response, pretty busy these days. I believe, that i will comeback on the next week and continue to work on this.

CyberDem0n: @anikin-aa ping

CyberDem0n: @anikin-aa ping

anikin-aa: @CyberDem0n , sorry for long response, i will to fix it during next week

anikin-aa: @CyberDem0n , let me verify the requested changes (just to double check, that i understood everything correctly)
i've to check if mode on pg data dir equals to PG_DIR_MODE_GROUP, then return group file mode otherwise return file mode owner ? because right now i am changing permissions only for files

CyberDem0n: It is a bit more complex than that.
Patroni not only creates files, but sometimes also directories (for example when copying logical slots to replicas).
In all cases umask and permissions should be set correctly and according to expectations of Postgres.
I.e., default values (owner only): https://github.com/postgres/postgres/blob/20c22333630111718bc97ce9a19b392f23e3c525/src/common/file_perm.c#L17-L25
and https://github.com/postgres/postgres/blob/20c22333630111718bc97ce9a19b392f23e3c525/src/common/file_perm.c#L37-L42 if group access is allowed.

Correct permissions should be "calculated" when:
1. We start Patroni with existing $PGDATA
2. After [bootstrap](https://github.com/zalando/patroni/blob/1b6e23ab6a5a92892c7948afde415f10cb95acdb/patroni/postgresql/bootstrap.py#L290-L294) (initdb or custom bootstrap) and after [create_replica](https://github.com/zalando/patroni/blob/1b6e23ab6a5a92892c7948afde415f10cb95acdb/patroni/postgresql/bootstrap.py#L278)

anikin-aa: ahh, ok, thank you for detailed explanation 

> but sometimes also directories (for example when copying logical slots to replicas).

didn't know about that, will check it too

CyberDem0n: Hi @anikin-aa any progress on it?

Issue #1907, opened 2021-04-19
Don't Remove unexpected parameters in Patroni config.yml from postgresql.conf file


fourpastmidnight:
## Describe the bug

I am configuring a new PostgreSQL 12.6 database cluster and have specified the setting `recovery_min_apply_delay = 3600` as a property under the `bootstrap.dcs.postgresql.parameters` section of the Patroni configuration file. Upon successfully starting Patroni, the following warning appears in the systemd log for the Patroni service:

```log
Apr 18 23:12:11 pgsql0 patroni[26714]: 2021-04-18 23:12:11,035 WARNING: Removing unexpected parameter=recovery_min_apply_delay value=3600 from the config
```

## To Reproduce

Specify the following property in your Patroni configuration file. (Note, minimal file example shown.)

```yaml
bootstrap:
  dcs:
    postgresql:
      parameters:
        recovery_min_apply_delay_value: 3600
```

Upon starting a cluster containing this PostgreSQL parameter, the following log line will be displayed in `journalctl`:

```log
Apr 18 23:12:11 pgsql0 patroni[26714]: 2021-04-18 23:12:11,035 WARNING: Removing unexpected parameter=recovery_min_apply_delay value=3600 from the config
```

## Expected behavior

I expect that _any_ valid PostgreSQL setting should not be removed by Patroni. If Patroni does not recognize a setting, it should just leave it alone. I understand this _may_ be to prevent people from "shooting themselves in the foot" with bad parameters, but PostgreSQL will do that for you. It'll either log a warning, an error, or outright terminate itself. In short, Patroni is here to manage the cluster, and not the database settings, other than ensuring that all cluster members contain the same settings pursuant to having a functional, highly-available cluster insofar as correct settings have been supplied by those setting up the database.

## Screenshots

Not Applicable.

## Environment

- Patroni version: 2.0.2
- PostgreSQL version: 12.6
- DCS (and its version): Etcd 3.2.26

## Patroni configuration file

```yaml
namespace: /jfrog
scope: artifactory
name: pgsql0

restapi:
  listen: 192.168.10.70:8008
  connect_address: 192.168.10.70:8008
  certfile: /etc/patroni/pki/pgsql0.crt
  keyfile: /etc/patroni/pki/pgsql0.key
  cafile: /etc/patroni/pki/ca.crt
  verify_client: required

ctl:
  insecure: false
  certfile: /etc/patroni/pki/pgsql0.crt
[...]

PR #1908, opened 2021-04-19
Get rid of false warning about invalid parameter

CyberDem0n:
Despite all recovery parameters became GUCs in PostgreSQL 12, there are very good reasons to keep them separated in the Patroni internals.

While implementing PostgreSQL parameters validation in #1674 one little oversight occurred. The parameters validation happens before the recovery parameters are skipped from the list, which produces a false warning.

Close https://github.com/zalando/patroni/issues/1907


FxKu:
:+1:

CyberDem0n:
:+1:

PR #1909, opened 2021-04-19
Add Patronis default postgres configs to the sample yml files

tommyzli:
This adds the default postgres settings enforced by Patroni to the `postgres{n}.yml` files provided in the repo. The documentation does call out the defaults that Patroni will set, but it can be missed if you (like me) download postgres0.yml and use that as a starting point. Hopefully the extra commented out configs serve as a visual cue to save the next person from the same mistake :)

### DIFF
--- a/postgres0.yml
+++ b/postgres0.yml
@@ -57,10 +57,15 @@ bootstrap:
       parameters:
 #        wal_level: hot_standby
 #        hot_standby: "on"
+#        max_connections: 100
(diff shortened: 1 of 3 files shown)


FxKu:
:+1:

CyberDem0n:
:+1:

CyberDem0n:
Merged, thank you @tommyzli!

PR #1912, opened 2021-04-21
Run behave tests with Python 3.9

tnir:
- Run behave tests with Python 3.9, instead of 3.8 on Linux
- Run behave tests with Python 3.9, instead of 3.7 on macOS
- Run behave tests with Python 3.6, instead of 3.5 on Linux since Python 3.5 reached EOL in 2020
- Stop to run uni tests with Python 3.5
- Push less number of code coverage from matrix jobs to coveralls
    - before this PR, 46 jobs https://coveralls.io/builds/38961070
    - with this PR, would be 33 jobs

### DIFF
--- a/.github/workflows/tests.yaml
+++ b/.github/workflows/tests.yaml
@@ -30,15 +30,6 @@ jobs:
       run: python .github/workflows/run_tests.py
       if: matrix.os != 'windows'
 
-    - name: Set up Python 3.5
-      uses: actions/setup-python@v2
-      with:
-        python-version: 3.5
-    - name: Install dependencies
-      run: python .github/workflows/install_deps.py
-    - name: Run tests and flake8
-      run: python .github/workflows/run_tests.py
-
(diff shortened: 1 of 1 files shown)


CyberDem0n (commented):
Sorry, but I don't agree with most of the points.

Database folks a veeery conservative and prefer not to touch the working system without a special need. Maybe 3.5 is EOL (2.7 actually is also EOL), but they still in use in plenty of places.

Running unit tests with 3.5 is not a big deal. Not running behave tests with 3.5 is actually a big deal. There is a slim difference between 3.6 and 3.5 and we already got a few times into a situation when 3.5 support was broken due to not being properly tested.

As far as I know, 3.9 is not yet included into any LTS Linux distros, therefore running behave tests with 3.9 is also a bit too early. Although, I don't mind changing it from 3.7 to 3.9 for MacOS.

I also don't understand the point of submitting coverage results only for the latest python. What benefits do we get from it?

PR #1917, opened 2021-04-23
The /leader endpoint returns 200 if node holds the lock

CyberDem0n:
Promoting the standby cluster requires updating load-balancer health checks, which is not very convenient and easy to forget.
In order to solve it, we change the behavior of the `/leader` health-check endpoint. It will return 200 without taking into account whether PostgreSQL is running as the primary or the standby_leader.


Jan-M:
:+1:

CyberDem0n:
:+1:

Issue #1746, opened 2020-11-04
Cluster init with RAFT traffic encryption


lalbrecht:
**Describe the bug**
Can't start Patroni cluster of 2 nodes with RAFT traffic password encryption. Secondary node doesn't start

**To Reproduce**
1. Configure Patroni with RAFT controller: 2 PG node + 1 Patroni RAFT Controller
2. Set same [password](https://patroni.readthedocs.io/en/latest/SETTINGS.html#raft) in RAFT configuration. 
3. Start Patroni on all nodes

**Expected behavior**
All nodes are up and running. Cluster has a stable leader.

**Screenshots**

**Environment**
- Patroni version: 2.0.1
- PostgreSQL version: 12
- DCS (and its version): no DCS, pure RAFT

**Patroni configuration file**
Patroni RAFT controller configuration:
```
scope: test
name: AA.BB.CC.DD

raft:
  data_dir: /var/data/raft
  self_addr: AA.BB.CC.DD:6666
  partner_addrs:
    - AA.BB.CC.DD:6666
    - AA.BB.CC.DD:6666
  password: '123456'
```

Patroni PG node:
```
scope: test
name: AA.BB.CC.DD

raft:
  data_dir: /var/data/raft
  self_addr: AA.BB.CC.DD:6666
  partner_addrs:
    - AA.BB.CC.DD:6666
    - AA.BB.CC.DD:6666
  password: '123456'

restapi:
  listen: 0.0.0.0:8008
  connect_address: AA.BB.CC.DD:8008
  authentication:
    username: username
    password: '123456'

```

**patronictl show-config**
Unfortunately patronictl not working too:
```
patronictl -c /etc/patroni.yml show-config
Exception in thread Thread-1:
Traceback (most recent call last):
  File "/usr/lib/python3.6/threading.py", line 916, in _bootstrap_inner
    self.run()
  File "/usr/lib/python3.6/threading.py", line 864, in run
    self._target(*self._args, **self._kwargs)
  File "/usr/local/lib/python3.6/dist-packages/pysyncobj/syncobj.py", line 508, in _autoTickThread
    self._onTick(self.__conf.autoTickPeriod)
  File "/usr/local/lib/python3.6/dist-packages/patroni/dcs/raft.py", line 250, in _onTick
    super(KVStoreTTL, self)._onTick(timeToWait)
  File "/usr/local/lib/python3.6/dist-packages/pysyncobj/syncobj.py", line 620, in _onTick
[...]

PR #1919, opened 2021-04-26
Fix issues with raft traffic encryption

CyberDem0n:
and run raft behave tests with encryption enabled.

Using the new `pysyncobj` release allowed us to get rid of a lot of hacks with accessing private properties and methods of the parent class and reduce the size of the `raft.py`.

Close https://github.com/zalando/patroni/issues/1746


sdudoladov:
:+1:

CyberDem0n:
:+1:

blaluc:
@CyberDem0n Will this fix be included in the 2.0.3 version and when, more or less, will it be released?
Many thanks

PR #1933, opened 2021-05-10
Fix bug with unix socket connections

CyberDem0n:
When the unix_socket_directories is not known Patroni was immediately going back to tcp connection via the localhost.

The bug was introduced in https://github.com/zalando/patroni/pull/1865


Jan-M:
.+1:

Jan-M:
:+1:

CyberDem0n:
:+1:

PR #1934, opened 2021-05-10
Correct log message for PG restart on replication config changes

NikolayS:
When some replication-related changes happen (e.g., restore_command
is changed), Patroni restarts Postgres and the following can bee seen in logs:

INFO: changing primary_conninfo and restarting in progress   

This message is misleading because it mentions one particular parameter
but the change could happen with a different one (e.g., restore_command).

This is a quick improvement that makes the messaging better. In the future,
it would make sense to report the diff of changes (hence, particular root cause
of the restart) that is analyzed in config.py:record_missmatch.

### DIFF
--- a/patroni/ha.py
+++ b/patroni/ha.py
@@ -426,7 +426,8 @@ class Ha(object):
             change_required, restart_required = self.state_handler.config.check_recovery_conf(node_to_follow)
             if change_required:
                 if restart_required:
-                    self._async_executor.try_run_async('changing primary_conninfo and restarting',
+                    self._async_executor.try_run_async('replication-related parameters (such as primary_conninfo, restore_command) has recently changed, '
+                                                       'restarting PostgreSQL to apply the changes',
                                                        self.state_handler.follow, args=(node_to_follow, role))
                 else:
                     self.state_handler.follow(node_to_follow, role, do_reload=True)


NikolayS:
@CyberDem0n we've discussed this in person – what do you think to adjust the log message (maybe making it a bit shorter), while planning to implement config diff reported in logs explicitly? (The main idea I have is that we need to see what exactly has changed – but I guess it might be non-trivial and long, so let's have some simple improvement first).

patroni/ha.py
```diff
@@ -426,7 +426,8 @@ def follow(self, demote_reason, follow_reason, refresh=True):
             change_required, restart_required = self.state_handler.config.check_recovery_conf(node_to_follow)
             if change_required:
                 if restart_required:
-                    self._async_executor.try_run_async('changing primary_conninfo and restarting',
+                    self._async_executor.try_run_async('replication-related parameters (such as primary_conninfo, restore_command) has recently changed, '
+                                                       'restarting PostgreSQL to apply the changes',
```
CyberDem0n, line 430:
```suggestion
                    self._async_executor.try_run_async('changing replication-related parameters and restarting',
```

CyberDem0n (changes requested):
@NikolayS if we go this way, I believe it would be also nice to log what parameters have been changed in the [check_recovery_conf()](https://github.com/zalando/patroni/blob/master/patroni/postgresql/config.py#L707-L730) method.
Only parameter names, without values (values could be sensitive!), and only if change results in a restart.

PR #1957, opened 2021-06-11
Track /status key updates

CyberDem0n:
The #1820 introduced a new key in DCS, named /status, which contains the leader optime and maybe flush LSN of logical slots.

In case of etcd3 and raft we should use updates of this key for syncing HA loop across nodes (before we were using /optime/leader).


sdudoladov:
:+1:

CyberDem0n:
:+1:

Issue #1734, opened 2020-10-14
Authenticated RCE via Patroni HTTP REST API


german-namestnikov:
Hello,

Seems like there is a way to compromise Patroni nodes with RCE in HTTP REST API. I placed a short write-up in my blog, you will find it here: https://illegalbytes.com/2020-10-14/patroni-remote-code-execution/

Changing the PostgreSQL configuration file with PATCH requests to Patroni HTTP API allows a remote authenticated* attacker to execute arbitrary OS commands. I placed the asterisk near the word 'authenticated' because most of the Patroni clusters I have met in my everyday work had no authentication enabled at least at one node, which potentially makes the whole cluster vulnerable to RCE.

I don't have enough experience with Patroni, so, I hope you will help me to determine the possible impact and - which is more important - mitigation. 

Thanks!

CyberDem0n:
Hi @german-namestnikov,

I will disappoint you, but you haven't discovered anything new.

First of all it is possible to protect the REST API with basic auth and/or tls, optionally check client key: https://patroni.readthedocs.io/en/latest/SETTINGS.html#rest-api.
Second, local config file always takes precedence over global configuration, therefore it is enough to have these options defined there: https://patroni.readthedocs.io/en/latest/dynamic_configuration.html
And third, there is a section in the documentation regarding security: https://patroni.readthedocs.io/en/latest/security.html.

This "vulnerability" is more or less the same as using COPY TO PROGRAM when you have postgres credentials.

And maybe a few words about why basic auth is not enabled by default:
1. Without tls basic auth doesn't really improves security (anyone could sniff unencrypted traffic)
2. If there'll be a default password in the config people are too lazy to change it, what is the same as not having a password at all.

german-namestnikov:
Thanks for quick and so useful response, @CyberDem0n !

Well, you didn't disappoint me. I found no mentions about this ability before my research - if it is so obvious as COPY TO PROGRAM, why I could not find any publications about this 'feature'? So, I think it is still useful :)

Authentication is Ok (and I agree with your words about default passwors) but seems like people neglect reading documentation and prefer to use ymls you provide by default. This is my own experience and I don't really think everyone has their HTTP API without auth. Just to note. 

Could you please tell me more about precedence of local configs over global configuration? Do you mean that this is enough to set 'archive_command' in yml to mitigate this technique?

CyberDem0n:
It is enough to put the following in the config.yam:
```yaml
postgresql:
  parameters:
[...]

PR #1959, opened 2021-06-14
Implemented allowlist for REST API

CyberDem0n:
If configured, only IPs that matching rules would be allowed to call unsafe endpoints.
In addition to that, it is possible to automatically include IPs of members of the cluster to the list.
If neither of the above is configured the old behavior is retained.

Partially address https://github.com/zalando/patroni/issues/1734


sdudoladov:
:+1:

CyberDem0n:
:+1:

Issue #1930, opened 2021-05-07
Patroni w/ Raft sometimes won't elect a new leader when the leader node goes down in docker


The-TT-Hacker:
**Describe the bug**

Note: I am using the new Raft option for DCS as we intend to use it when it is stable.
I have a 3 node setup in docker swarm and each instance of Patroni tied to its own physical server in the swarm. Sometimes when I disconnect the server with the leader node from the network, the other 2 nodes get into an infinite loop where Postgres fails to resolve the hostname (docker-compose service name) of the leader node and Patroni does not elect a new leader. The docker DNS seems to remove entries when a container stops, however it looks like there's a race condition which means this error does not always happen. If the DNS entry for the master node is removed after Patroni does a leader election, the cluster fails over correctly, however if the DNS entry is removed quickly by docker then this infinite loop occurs.

Postgres failing to resolve the hostname of the leader
FATAL:  could not connect to the primary server: could not translate host name "postgres1" to address: Name or service not known

Patroni not electing a new leader
2021-05-06 23:51:00,613 INFO: Lock owner: postgres1; I am postgres3
2021-05-06 23:51:00,613 INFO: does not have lock
2021-05-06 23:51:02,617 INFO: no action.  i am a secondary and i am following a leader

This occurs on both standby nodes while the leader is down. The network connection with postgres1 has been physically broken yet the 2 standby nodes do not elect a new leader.

**To Reproduce**

My setup is here: https://github.com/SPG-Controls/patroni-raft-docker
I used 3 physical servers so I could unplug network cables however I think this could also be reproduced on a single node swarm by scaling the service acting as the leader to 0 replicas.

**Expected behavior**
One of the two standby nodes should be elected as leader and be promoted as the primary node

**Environment**
- Patroni version: 2.0.2 Commit 99626a07f23151aa1abd63b9fb07819ab10e6900
- PostgreSQL version: 13
- DCS (and its version): Raft

**Patroni configuration file**
```
bootstrap:
  dcs:
    ttl: 30
    loop_wait: 10
    retry_timeout: 10
    maximum_lag_on_failover: 1048576
    synchronous_mode: true
    synchronous_mode_strict: true
    synchronous_node_count: 1

    postgresql:
      parameters:
        max_connections: 256
      use_pg_rewind: true
      remove_data_directory_on_diverged_timelines: false
      remove_data_directory_on_rewind_failure: true

  initdb:
  - encoding: UTF8
  - data-checksums

  pg_hba:
[...]

Issue #1931, opened 2021-05-07
Patroni w/ Raft in Docker fails to start when using service hostnames for PATRONI_RAFT_PARTNER_ADDRS


The-TT-Hacker:
**Describe the bug**

Starting up a 3 node Patroni cluster in docker swarm using Raft for DCS will fail to start when you use hostnames instead of IP addresses because docker dynamically adds and removes DNS entries when containers startup and shut down. During the Raft bootstrap process each node will try to resolve each others hostnames, however the first node docker starts will immediately fail because the hostnames of the other 2 have not been entered into the DNS yet. The services will keep starting up and failing immediately unless they happen to all start at the exact same time which is rare. I solved this by adding a sleep in the entry-point script so that all services start before Patroni starts and tries resolve the hostnames, however this does not solve the scenario where 1 of the 3 servers is down and one of the other nodes restarts as it will never be able to resolve the hostname of the node which is down on startup.

Other projects which implement distributed consensus algorithms such as all the other DCS options Patroni supports do not seem to have this problem. I haven't looked into how they solve it, but maybe Patroni has to perform some DNS resolution retries for the partner nodes and then if it is unsuccessful behave as if that node is down instead of crashing out of the program.

I also tried assigning static IP addresses to the services to remove the DNS resolution but it seems like docker overlay networks do not support static IPs for docker services.

**To Reproduce**

Use this setup https://github.com/SPG-Controls/patroni-raft-docker/blob/main/docker-compose.yml

set the INIT_WAIT_TIME environment variable to 0 to remove the sleep in the entry-point script
start the stack

**Expected behavior**

Patroni w/ Raft should be able to:
- bootstrap a quorum in a dynamic environment like docker swarm
- behave as if a node is down when the DNS resolution for it fails

**Environment**
- Patroni version: 2.0.2 99626a07f23151aa1abd63b9fb07819ab10e6900
- PostgreSQL version: 13
- DCS (and its version): Raft

**Patroni configuration file**
```
bootstrap:
  dcs:
    ttl: 30
    loop_wait: 10
    retry_timeout: 10
    maximum_lag_on_failover: 1048576
    synchronous_mode: true
    synchronous_mode_strict: true
    synchronous_node_count: 1

    postgresql:
      parameters:
        max_connections: 256
      use_pg_rewind: true
      remove_data_directory_on_diverged_timelines: false
      remove_data_directory_on_rewind_failure: true

  initdb:
  - encoding: UTF8
  - data-checksums

[...]

PR #1960, opened 2021-06-14
Handle DNS issues in Raft implementation

CyberDem0n:
- Resolve Node IP for every connection attempt
- Handle exception with connection failures due to failed resolve
- Set PySyncObj DNS Cache timeouts aligned with `loop_wait` and `ttl`

In addition to that,  postpone the leader race for freshly started Raft nodes. It will help with the situation when the leader node was alone and demoted the Postgres and after that, the replica arrives, and quickly takes the leader lock without really performing the leader race.

Close https://github.com/zalando/patroni/issues/1930, https://github.com/zalando/patroni/issues/1931

### DIFF
--- a/patroni/dcs/raft.py
+++ b/patroni/dcs/raft.py
@@ -30,11 +30,11 @@ class _TCPTransport(TCPTransport):
             return False
 
 
-class _TCPNode(TCPNode):
+def resolve_host(self):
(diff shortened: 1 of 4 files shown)


sdudoladov:
:+1:

CyberDem0n:
:+1:

Issue #1958, opened 2021-06-12
Feature Request: Add user defined tags and healthchecks to instances


jooof:
We are running Patroni with 5 nodes, 3 for short queries which are also participating in leader election and 2 for OLAP queries which can lag behind the leader and are not participating.

There are 2 things that I felt could really be helpful if added to patroni. 

1- the ability to add user defined tags to each instance. for example I would be able to tag a particular instance as analyze and also be able to check if an instance is tagged with a health check api like:
 `GET /tag/analyze?lag=50MB`
and also be abled to see the tags for each instance in requests like `GET /patroni `and `GET /cluster` .

2- the ability to attach a health check script to a specific health check api. For example I don't want any read or analyze query to be   redirected to leader, unless there is no other running db to do the job. So by being able to attach other (preferably python) scripts to patroni api, We can define any kind of complex health check and redirect mechanism to the databases. 

If this seems like a promising feature, I am also very happy to be able to contribute to the project this way:)

CyberDem0n:
> the ability to add user defined tags to each instance

The `tags` section is not limited by documented tags. You can add any arbitrary information that you like and it is already visible everywhere:
```json
$ curl -sL localhost:8008/cluster | jq .
{
  "members": [
    {
      "name": "postgresql0",
      "tags": {
        "foo": "bar"
      },
      "timeline": 17,
      "state": "running",
      "host": "127.0.0.1",
      "role": "leader",
      "port": 5432,
      "api_url": "http://127.0.0.1:8008/patroni"
    }
  ]
}

$ curl -sL localhost:8008/patroni | jq .
{
  "database_system_identifier": "6971790382475598570",
  "postmaster_start_time": "2021-06-14 07:19:32.185583+02:00",
  "tags": {
    "foo": "bar"
  },
  "timeline": 17,
  "cluster_unlocked": false,
  "patroni": {
    "scope": "batman",
    "version": "2.0.2"
  },
  "state": "running",
  "role": "master",
  "xlog": {
    "location": 50491592
  },
  "server_version": 130003
}

$ patronictl list
+ Cluster: batman (6971790382475598570) -----+----+-----------+----------+
| Member      | Host      | Role   | State   | TL | Lag in MB | Tags     |
+-------------+-----------+--------+---------+----+-----------+----------+
| postgresql0 | 127.0.0.1 | Leader | running | 17 |           | foo: bar |
+-------------+-----------+--------+---------+----+-----------+----------+
```

[...]

PR #1964, opened 2021-06-15
add health check on user defined tags

jooof:
Add health check api for user defined tags based on the conversation in #1958

### DIFF
--- a/docs/SETTINGS.rst
+++ b/docs/SETTINGS.rst
@@ -374,7 +374,7 @@ Tags
 - **replicatefrom**: The IP address/hostname of another replica. Used to support cascading replication.
 - **nosync**: ``true`` or ``false``. If set to ``true`` the node will never be selected as a synchronous replica.
 
-In addition to these predefined tags, You can also add your own tags to mark instances:
+In addition to these predefined tags, You can also add your own ones:
 
 - **key1**: ``true``
 - **key2**: ``false``
--- a/docs/SETTINGS.rst
+++ b/docs/SETTINGS.rst
@@ -374,11 +374,11 @@ Tags
 - **replicatefrom**: The IP address/hostname of another replica. Used to support cascading replication.
 - **nosync**: ``true`` or ``false``. If set to ``true`` the node will never be selected as a synchronous replica.
 
-In addition to these predefined tags, You can also add your own ones:
+In addition to these predefined tags, you can also add your own ones:
 
 - **key1**: ``true``
 - **key2**: ``false``
 - **key3**: ``1.4``
 - **key4**: ``"RandomString"``
 
-You can get these tags from the rest api. You can also check for an instance health using these tags. If the tag isn't defined for an instance, or if the respective value doesn't match the querying value, it will return HTTP Status Code 503.
+Tags are visible in the :ref:`REST API <rest_api>` and ``patronictl list`` You can also check for an instance health using these tags. If the tag isn't defined for an instance, or if the respective value doesn't match the querying value, it will return HTTP Status Code 503.
--- a/docs/SETTINGS.rst
+++ b/docs/SETTINGS.rst
@@ -373,3 +373,12 @@ Tags
 - **noloadbalance**: ``true`` or ``false``. If set to ``true`` the node will return HTTP Status Code 503 for the ``GET /replica`` REST API health-check and therefore will be excluded from the load-balancing. Defaults to ``false``.
 - **replicatefrom**: The IP address/hostname of another replica. Used to support cascading replication.
 - **nosync**: ``true`` or ``false``. If set to ``true`` the node will never be selected as a synchronous replica.
+
+In addition to these predefined tags, You can also add your own tags to mark instances:
+
+- **key1**: ``true``
+- **key2**: ``false``
+- **key3**: ``1.4``
+- **key4**: ``"RandomString"``
+
+You can get these tags from the rest api. You can also check for an instance health using these tags. If the tag isn't defined for an instance, or if the respective value doesn't match the querying value, it will return HTTP Status Code 503.
--- a/patroni/api.py
+++ b/patroni/api.py
@@ -122,14 +122,13 @@ class RestApiHandler(BaseHTTPRequestHandler):
 
         status_code = 503
 
-        request_for_master = 'master' in path or \
-                             'leader' in path or \
-                             'primary' in path or \
-                             'read-write' in path
-        if request_for_master:
-            status_code = primary_status_code
-        elif 'standby_leader' in path or 'standby-leader' in path:
+        ignore_tags = False
+        if 'standby_leader' in path or 'standby-leader' in path:
             status_code = standby_leader_status_code
+            ignore_tags = True
+        elif 'master' in path or 'leader' in path or 'primary' in path or 'read-write' in path:
+            status_code = primary_status_code
+            ignore_tags = True
         elif 'replica' in path:
             status_code = replica_status_code
         elif 'read-only' in path:
@@ -145,7 +144,7 @@ class RestApiHandler(BaseHTTPRequestHandler):
                 status_code = replica_status_code
 
         # check for user defined tags in query params
-        if not request_for_master and status_code == 200:
+        if not ignore_tags and status_code == 200:
             qs_tag_prefix = "tag_"
             for qs_key, qs_value in self.path_query.items():
                 if not qs_key.startswith(qs_tag_prefix):
--- a/patroni/api.py
+++ b/patroni/api.py
@@ -153,7 +153,7 @@ class RestApiHandler(BaseHTTPRequestHandler):
                 if instance_tag_value is None:
                     status_code = 503
                     break
-                if not isinstance(instance_tag_value, str):
+                if not isinstance(instance_tag_value, six.string_types):
                     instance_tag_value = str(instance_tag_value).lower()
                 if instance_tag_value != qs_value:
                     status_code = 503
--- a/patroni/api.py
+++ b/patroni/api.py
@@ -155,11 +155,11 @@ class RestApiHandler(BaseHTTPRequestHandler):
         # check for user defined tags in query params
         if status_code == 200:
             qs_tag_prefix = "tag_"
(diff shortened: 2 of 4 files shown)


patroni/api.py
```diff
@@ -140,6 +152,22 @@ def do_GET(self, write_status_code_only=False):
             elif path in ('/async', '/asynchronous') and not is_synchronous:
                 status_code = replica_status_code
 
+        # check for user defined tags in query params
+        if status_code == 200:
```
CyberDem0n, line 156: It makes sense to check tags only for non-leader health checks

patroni/api.py
```diff
@@ -140,6 +152,22 @@ def do_GET(self, write_status_code_only=False):
...
+            qs_tag_prefix = "tag_"
+            qs_tags = {tag_name[len(qs_tag_prefix):]: tag_values[0]
+                       for tag_name, tag_values in self.path_query.items()
+                       if tag_name.startswith(qs_tag_prefix)}
+            for qs_key, qs_value in qs_tags.items():
```
CyberDem0n, line 161: Is there a need to do a list comprehension if later we still have a for loop over the just built dict?

patroni/api.py
```diff
@@ -140,6 +152,22 @@ def do_GET(self, write_status_code_only=False):
...
+                qs_value = self.cast_query_param(qs_value)
+                instance_tag_value = patroni.tags.get(qs_key)
```
CyberDem0n, line 163: IMO, since we are interested in comparing precise values it doesn't make much sense to cast str to other objects.
The opposite makes better sense.
I.e. we can cast `instance_tag_value` to a lower string if it is not None and not a string and use this value for comparison with query parameter value.

CyberDem0n (commented):
What about documentation?
And unit-tests?

patroni/api.py
jooof: The only reason I didn't do that was due to the simplicity of comparing floating point and integer values. Do you prefer the system to return 503 if 1.0 is compared to 1 or 200?

patroni/api.py
jooof: I was thinking if that's what the user wants then let them have it. But if this way is better then I'll change it

patroni/api.py
jooof: oops:D

jooof:
> What about documentation?
> And unit-tests?

Will add the unit tests right away. 
But since there was no documentation on the tags themselves I didn't consider doing so. I'll get to it.

patroni/api.py
```diff
@@ -140,6 +140,25 @@ def do_GET(self, write_status_code_only=False):
...
+                if not qs_key.startswith(qs_tag_prefix):
+                    continue
+                qs_key = qs_key[len(qs_tag_prefix):]
+                qs_value = qs_value[0]
+                instance_tag_value = patroni.tags.get(qs_key)
+                # tag not registered for instance
+                if instance_tag_value is None:
+                    status_code = 503
+                    break
+                if not isinstance(instance_tag_value, str):
```
CyberDem0n, line 156:
```suggestion
                if not isinstance(instance_tag_value, six.string_types):
```

patroni/api.py
CyberDem0n: Tags have nothing to do with application-level sharding.
If you run many shards (clusters), they must have a different cluster name (scope).

IMO, the leader should always be reported as leader, no matter what tags are requested in a health check.

CyberDem0n:
> But since there was no documentation on the tags themselves I didn't consider doing so. I'll get to it.

There is: https://github.com/zalando/patroni/blob/master/docs/SETTINGS.rst#tags

patroni/api.py
jooof: done``

> > But since there was no documentation on the tags themselves I didn't consider doing so. I'll get to it.
> 
> There is: https://github.com/zalando/patroni/blob/master/docs/SETTINGS.rst#tags

I meant user defined tags. I'll add it to the doc

jooof:
> > But since there was no documentation on the tags themselves I didn't consider doing so. I'll get to it.
> 
> There is: https://github.com/zalando/patroni/blob/master/docs/SETTINGS.rst#tags

I meant user defined tags. I'll add it to the doc

patroni/api.py
```diff
@@ -140,6 +144,25 @@ def do_GET(self, write_status_code_only=False):
             elif path in ('/async', '/asynchronous') and not is_synchronous:
                 status_code = replica_status_code
 
+        # check for user defined tags in query params
+        if not request_for_master and status_code == 200:
```
CyberDem0n, line 148: Well, the standby-leader should be also excluded, therefore, these [changes](https://github.com/zalando/patroni/pull/1964/files#diff-13329cd3342d1757c2c9b4dd7a26c0170f4deee5d32eff5c2fd719cc9ff9375bR125-R132) more or less unnecessary.

tests/test_api.py
```diff
@@ -143,14 +143,16 @@ def sendall(self, *args, **kwargs):
...
         self.socket = 0
         self.serve_forever = Mock()
+        if patroni is None:
+            patroni = MockPatroni()
         MockRestApiServer._BaseServer__is_shut_down = Mock()
         MockRestApiServer._BaseServer__shutdown_request = True
         config = config or {'listen': '127.0.0.1:8008', 'auth': 'test:test', 'certfile': 'dumb', 'verify_client': 'a',
                             'http_extra_headers': {'foo': 'bar'}, 'https_extra_headers': {'foo': 'sbar'}}
-        super(MockRestApiServer, self).__init__(MockPatroni(), config)
+        super(MockRestApiServer, self).__init__(patroni, config)
```
CyberDem0n, line 155: IMO, this refactoring is not necessary. Please just add the necessary tags to the `MockPatroni.tags`, they are harmless for remaining tests.

tests/test_api.py
```diff
@@ -197,6 +199,38 @@ def test_do_GET(self):
...
+            MockRestApiServer(RestApiHandler, 'GET /replica?lag=1M&tag_key1=true&tag_key2=False&tag_key3=1&tag_key4=1.4&tag_key5=RandomTag', patroni=mock_patroni)
+            MockRestApiServer(RestApiHandler, 'GET /replica?lag=1M&tag_key1=true&tag_key2=false&tag_key3=1.0&tag_key4=1.4&tag_key5=RandomTag', patroni=mock_patroni)
+            MockRestApiServer(RestApiHandler, 'GET /replica?lag=1M&tag_key1=true&tag_key2=false&tag_key3=1&tag_key4=1.4&tag_key5=RandomTag&tag_key6=RandomTag2', patroni=mock_patroni)
+        #
+        MockRestApiServer(RestApiHandler, 'GET /read-write?lag=1M&tag_key1=true&tag_key2=false&tag_key3=1&tag_key4=1.4&tag_key5=RandomTag', patroni=mock_patroni)
+        MockRestApiServer(RestApiHandler, 'GET /read-write?lag=1M&tag_key1=true&tag_key2=False&tag_key3=1&tag_key4=1.4&tag_key5=RandomTag', patroni=mock_patroni)
+        MockRestApiServer(RestApiHandler, 'GET /read-write?lag=1M&tag_key1=true&tag_key2=false&tag_key3=1.0&tag_key4=1.4&tag_key5=RandomTag', patroni=mock_patroni)
+        MockRestApiServer(RestApiHandler, 'GET /read-write?lag=1M&tag_key1=true&tag_key2=false&tag_key3=1&tag_key4=1.4&tag_key5=RandomTag&tag_key6=RandomTag2', patroni=mock_patroni)
```
CyberDem0n, line 231: Lines are clearly bigger than 120 characters, It will not pass flake8 validation (you can call it as `python setup.py flake8`)

docs/SETTINGS.rst
```diff
@@ -373,3 +373,12 @@ Tags
...
 - **nosync**: ``true`` or ``false``. If set to ``true`` the node will never be selected as a synchronous replica.
+
+In addition to these predefined tags, You can also add your own ones:
+
+- **key1**: ``true``
+- **key2**: ``false``
+- **key3**: ``1.4``
+- **key4**: ``"RandomString"``
+
+You can get these tags from the rest api. You can also check for an instance health using these tags. If the tag isn't defined for an instance, or if the respective value doesn't match the querying value, it will return HTTP Status Code 503.
```
CyberDem0n, line 384:
```suggestion
In addition to these predefined tags, you can also add your own ones:

- **key1**: ``true``
- **key2**: ``false``
- **key3**: ``1.4``
- **key4**: ``"RandomString"``

Tags are visible in the :ref:`REST API <rest_api>` and ``patronictl list``. You can also check for an instance health using these tags. If the tag isn't defined for an instance, or if the respective value doesn't match the querying value, it will return HTTP Status Code 503.
```

CyberDem0n (changes requested):

tests/test_api.py
```diff
@@ -197,6 +197,90 @@ def test_do_GET(self):
...
+                                          'tag_key1=true&tag_key2=False&'
+                                          'tag_key3=1&tag_key4=1.4&tag_key5=RandomTag')
+        MockRestApiServer(RestApiHandler, 'GET /read-write?lag=1M&'
+                                          'tag_key1=true&tag_key2=false&'
+                                          'tag_key3=1.0&tag_key4=1.4&tag_key5=RandomTag')
+        MockRestApiServer(RestApiHandler, 'GET /read-write?lag=1M&'
+                                          'tag_key1=true&tag_key2=false&'
+                                          'tag_key3=1&tag_key4=1.4&tag_key5=RandomTag&tag_key6=RandomTag2')
+
+
```
CyberDem0n, line 283: flake8 complains about too many blank lines:
```suggestion

```

CyberDem0n (dismissed):
Please revert changes of postgres0.yml, postgres1.yml, and requirements.txt

jooof:
> Please revert changes of postgres0.yml, postgres1.yml, and requirements.txt

Oh, so sorry about that one

sdudoladov:
:+1:

CyberDem0n:
:+1:

PR #1970, opened 2021-06-17
Compatibility with psycopg2 2.9+

CyberDem0n:
the autocommit = True is ignored in the `with connection` block


sdudoladov:
:+1:

CyberDem0n:
:+1:

PR #1972, opened 2021-06-18
Add scope as label to Prometheus metrics

mbanck:
The Prometheus metrics look nice, but I think it would make sense to add (at least) the scope as a label to them.

That way, if you have lots of Patroni clusters, you (i) get a better overview and (ii) can actually tell which nodes belong to which cluster.

Or did I miss something, and this is already there somehow?

### DIFF
--- a/patroni/api.py
+++ b/patroni/api.py
@@ -187,48 +187,49 @@ class RestApiHandler(BaseHTTPRequestHandler):
 
         metrics = []
 
+        scope_label = '{{scope="{0}"}}'.format(patroni.postgresql.scope)
         metrics.append("# HELP patroni_version Patroni semver without periods.")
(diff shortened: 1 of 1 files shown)


mbanck:
cc @mamercad

CyberDem0n (changes requested):
Hi @mbanck, I think adding labels should not hurt.

Please fix the formatting:
```python
$ python setup.py flake8
running flake8
patroni/api.py:227:25: E128 continuation line under-indented for visual indent
patroni/api.py:233:25: E128 continuation line under-indented for visual indent
patroni/api.py:245:25: E128 continuation line under-indented for visual indent
patroni/api.py:250:25: E128 continuation line under-indented for visual indent
```

PR #1978, opened 2021-06-22
Add scope as label to Prometheus metrics

mbanck:
The Prometheus metrics look nice, but I think it would make sense to add (at least) the scope as a label to them.

That way, if you have lots of Patroni clusters, you (i) get a better overview and (ii) can actually tell which nodes belong to which cluster.

This also fixes indenting issues.

Sorry I had a typo in the original branch name and didn't realize Github would auto-close the PR

### DIFF
--- a/patroni/api.py
+++ b/patroni/api.py
@@ -191,73 +191,74 @@ class RestApiHandler(BaseHTTPRequestHandler):
 
         metrics = []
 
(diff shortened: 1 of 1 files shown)


FxKu:
:+1:

CyberDem0n:
:+1:

CyberDem0n:
Merged, thank you @mbanck!

Issue #1981, opened 2021-06-24
If Zookeeper is unavailable, master will not shutdown or demote


alwyndavis:
**Describe the bug**
When Patroni is setup to use Zookeeper as the DCS, making Zookeeper unavailable (e.g. stopping all the Zookeeper nodes) doesn't cause the master to shutdown (expected behaviour based on documentation and when etcd is the DCS).

**To Reproduce**
1. Setup Zookeeper
2. Setup two Patroni nodes and let them join normally
3. Stop a majority of the Zookeeper nodes
4. Patroni agents will go into a connection retry loop:
```
2021-06-24 06:31:26,806 WARNING: Connection dropped: socket connection error: Connection refused
2021-06-24 06:31:26,806 INFO: Connecting to 3.232.116.122(3.232.116.122):2182, use_ssl: True
2021-06-24 06:31:26,832 WARNING: Connection dropped: socket connection broken
2021-06-24 06:31:27,915 INFO: Connecting to 174.129.244.110(174.129.244.110):2182, use_ssl: True
2021-06-24 06:31:27,916 WARNING: Connection dropped: socket connection error: Connection refused
```
Whilst stuck in this loop, the master will continue to be writeable

* Use Kazoo 2.8.0 (discussed on #patroni Slack channel and regression might be from 2.7)

**Expected behavior**
Current master should shutdown

**Environment**
- Patroni version: 2.0.2
- PostgreSQL version: 13.3
- DCS (and its version): Zookeeper 3.5.8

**Patroni configuration file**
```
scope: zk-side
namespace: /service/
name: pg-1

restapi:
    listen: 10.0.0.10:8008
    connect_address: 10.0.0.10:8008

bootstrap:
    dcs:
        ttl: 30
        loop_wait: 10
        retry_timeout: 10
        maximum_lag_on_failover: 1048576
        synchronous_mode_strict: false
        postgresql:
            use_pg_rewind: true

    postgresql:
      use_pg_rewind: true

    initdb:
    - encoding: UTF8
    - data-checksums

    pg_hba:
    - host replication all all md5
    - host all all all trust

zookeeper:
    hosts:
    - 52.86.244.240:2182
    - 3.232.116.122:2182
    - 174.129.244.110:2182
    use_ssl: true
    verify: false

postgresql:
    listen: 0.0.0.0:5432
    connect_address: 10.0.0.10:5432
    data_dir: /data/patroni
    bin_dir: /var/lib/postgresql/13.3/bin
[...]

PR #1982, opened 2021-06-24
Compatibility with kazoo-2.7+

CyberDem0n:
Old versions of `kazoo` immediately discarded all requests to Zookeeper if the connection is in the `SUSPENDED` state. This is absolutely fine because Patroni is handling retries on its own.
Starting from 2.7, kazoo started queueing requests instead of discarding and as a result, the Patroni HA  loop was getting stuck until the connection to Zookeeper is reestablished, causing no demote of the Postgres.
In order to return to the old behavior we override the `KazooClient._call()` method.

In addition to that, we ensure that the `Postgresql.reset_cluster_info_state()` method is called even if DCS failed (the order of calls was changed in the #1820).

Close https://github.com/zalando/patroni/issues/1981

### DIFF
--- a/patroni/dcs/zookeeper.py
+++ b/patroni/dcs/zookeeper.py
@@ -4,8 +4,9 @@ import select
 import time
 
 from kazoo.client import KazooClient, KazooState, KazooRetry
-from kazoo.exceptions import NoNodeError, NodeExistsError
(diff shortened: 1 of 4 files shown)


sdudoladov:
:+1:

CyberDem0n:
:+1:

PR #1993, opened 2021-06-30
support dynamically registering/deregistering as a consul service

### LINKED ISSUE #1988: sighup doesnt reload consul.register_service
**Describe the bug**
The register_service setting for Consul isn't dynamically adjustable.

**To Reproduce**
1. Start Patroni with Consul as the DCS and without setting consul.register_service in patroni.yml, so it defaults to false
2. set consul.register_service: true in patroni.yml
3. reload patroni (either by sighup or by the reload endpoint)

**Expected behavior**
I'd expect this to enable Consul service registration

**Environment**
- Patroni version: 2.0.2
- PostgreSQL version: 13..3
- DCS (and its version): Consul 1.10.0

I'm happy to put together a PR if this is something that yall are interested in accepting

tommyzli:
closes #1988

### DIFF
--- a/patroni/dcs/consul.py
+++ b/patroni/dcs/consul.py
@@ -227,12 +227,10 @@ class Consul(AbstractDCS):
         self._last_session_refresh = 0
         self.__session_checks = config.get('checks', [])
         self._register_service = config.get('register_service', False)
+        self._previous_loop_register_service = self._register_service
         if self._register_service:
             self._service_tags = config.get('service_tags', [])
-            self._service_name = service_name_from_scope_name(self._scope)
-            if self._scope != self._service_name:
-                logger.warning('Using %s as consul service name instead of scope name %s', self._service_name,
-                               self._scope)
+            self._set_service_name()
         self._service_check_interval = config.get('service_check_interval', '5s')
         if not self._ctl:
             self.create_session()
@@ -250,7 +248,17 @@ class Consul(AbstractDCS):
 
     def reload_config(self, config):
         super(Consul, self).reload_config(config)
-        self._client.reload_config(config.get('consul', {}))
+
+        consul_config = config.get('consul', {})
+        self._client.reload_config(consul_config)
+        self._service_tags = consul_config.get('service_tags', [])
+
+        should_register_service = consul_config.get('register_service', False)
+        if should_register_service and not self._register_service:
+            self._set_service_name()
+
+        self._previous_loop_register_service = self._register_service
+        self._register_service = should_register_service
 
     def set_ttl(self, ttl):
         if self._client.http.set_ttl(ttl/2.0):  # Consul multiplies the TTL by 2x
@@ -408,7 +416,7 @@ class Consul(AbstractDCS):
         try:
             args = {} if permanent else {'acquire': self._session}
             self._client.kv.put(self.member_path, json.dumps(data, separators=(',', ':')), **args)
-            if self._register_service:
+            if self._register_service or self._previous_loop_register_service:
                 self.update_service(not create_member and member and member.data or {}, data)
             return True
         except InvalidSession:
@@ -417,6 +425,11 @@ class Consul(AbstractDCS):
         except Exception:
             logger.exception('touch_member')
         return False
+
+    def _set_service_name(self):
+        self._service_name = service_name_from_scope_name(self._scope)
+        if self._scope != self._service_name:
+            logger.warning('Using %s as consul service name instead of scope name %s', self._service_name, self._scope)
 
     @catch_consul_errors
     def register_service(self, service_name, **kwargs):
@@ -449,7 +462,8 @@ class Consul(AbstractDCS):
             'tags': tags
         }
 
-        if state == 'stopped':
+        if state == 'stopped' or (not self._register_service and self._previous_loop_register_service):
+            self._previous_loop_register_service = self._register_service
             return self.deregister_service(params['service_id'])
 
         if role in ['master', 'replica', 'standby-leader']:
@@ -470,7 +484,7 @@ class Consul(AbstractDCS):
             if old_data.get(key) != new_data[key]:
                 update = True
 
-        if force or update:
+        if force or update or (self._register_service != self._previous_loop_register_service):
             return self._update_service(new_data)
 
     @catch_consul_errors
--- a/patroni/dcs/consul.py
+++ b/patroni/dcs/consul.py
@@ -228,9 +228,9 @@ class Consul(AbstractDCS):
         self.__session_checks = config.get('checks', [])
         self._register_service = config.get('register_service', False)
         self._previous_loop_register_service = self._register_service
+        self._service_tags = sorted(config.get('service_tags', []))
+        self._previous_loop_service_tags = self._service_tags
         if self._register_service:
-            self._service_tags = sorted(config.get('service_tags', []))
-            self._previous_loop_service_tags = self._service_tags
             self._set_service_name()
         self._service_check_interval = config.get('service_check_interval', '5s')
         if not self._ctl:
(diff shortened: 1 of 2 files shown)


CyberDem0n (commented):
Changes are looking good to me, but I have a little question.
Shouldn't we also handle the change of `service_tags` dynamically?

tommyzli:
Ah right, will add

tommyzli:
hey @CyberDem0n, mind taking a second look?

CyberDem0n (commented):
Tags reload doesn't work:
1. I started Patroni with `service_tags: [foo, buzz]`
```
2021-07-14 15:13:28.721 CEST [23459] LOG:  database system is ready to accept connections
2021-07-14 15:13:29,677 INFO: Lock owner: postgresql0; I am postgresql0
2021-07-14 15:13:29,711 INFO: Register service batman, params {'check': {'DeregisterCriticalServiceAfter': '150.0s', 'interval': '5s', 'http': 'http://127.0.0.1:8008/master'}, 'service_id': 'batman/postgresql0', 'tags': ['bar', 'foo', 'master'], 'port': 5432, 'address': '127.0.0.1'}
2021-07-14 15:13:29,739 INFO: no action. I am (postgresql0) the leader with the lock
2021-07-14 15:13:29,767 INFO: no action. I am (postgresql0) the leader with the lock
```
2. changed `service_tags` to `[foo, buzz]`, sent a SIGHUP and nothing happened:
```
2021-07-14 15:14:49,757 INFO: no action. I am (postgresql0) the leader with the lock
2021-07-14 15:14:59,757 INFO: no action. I am (postgresql0) the leader with the lock
2021-07-14 15:15:09,760 INFO: Reloading PostgreSQL configuration.
server signaled
2021-07-14 15:15:09.764 CEST [23459] LOG:  received SIGHUP, reloading configuration files
2021-07-14 15:15:10,769 INFO: no action. I am (postgresql0) the leader with the lock
2021-07-14 15:15:19,756 INFO: no action. I am (postgresql0) the leader with the lock
```

Consul API also reports old tags:
```json
curl -s localhost:8500/v1/agent/services | jq .
{                                                                                                                                                                                                                                                                               
  "batman/postgresql0": {                                                                                                                                                                                                                                                       
    "ID": "batman/postgresql0",                                                                                                                                                                                                                                                 
    "Service": "batman",
    "Tags": [
      "bar",
      "foo",
      "master"
    ],
    "Meta": {},
    "Port": 5432,
    "Address": "127.0.0.1",
    "Weights": {
      "Passing": 1,
      "Warning": 1
    },
    "EnableTagOverride": false
  }
}

```

tommyzli:
Apparently by default, Consul silently ignores changes to service tags. Adding `enable_tag_override: True` to the service definition resolves that

patroni/dcs/consul.py
```diff
@@ -229,7 +229,8 @@ def __init__(self, config):
         self._register_service = config.get('register_service', False)
         self._previous_loop_register_service = self._register_service
         if self._register_service:
-            self._service_tags = config.get('service_tags', [])
+            self._service_tags = sorted(config.get('service_tags', []))
+            self._previous_loop_service_tags = self._service_tags
```
CyberDem0n, line 233: If upon start the `register_service` isn't set in the config, the next call of `reload_config()` will fail due to undefined `self._service_tags`.

CyberDem0n (commented):
> Adding enable_tag_override: True to the service definition resolves that

No, it doesn't. The issue is that the `agent.service.register()` method isn't called. Even the `Consulupdate_service()` isn't called, probably due to the https://github.com/zalando/patroni/pull/1993/files#diff-99372e752231ac1fddf9cda076dd985f0ed8293e8ba1c06ea1d95a0530ccb959R421

tommyzli:
Oh, I see what happened. I've mostly been testing this on short lived patroni processes, and during the first minutes of its life the [Member](https://github.com/zalando/patroni/blob/master/patroni/dcs/consul.py#L394) data changes a few times as the node gets promoted from standby to leader
That causes this [short circuit](https://github.com/zalando/patroni/blob/master/patroni/dcs/consul.py#L405) to do nothing because `deep_compare` indicates a change, and as a side effect allows the config reload to correctly change the service definition since it happened to get wrapped up in the same loop iteration

Question for you @CyberDem0n - how important do you think that short circuiting logic is for Consul? I checked the other DCS implementations and it's not consistent. Raft and Etcd will write the key each time, and the others will exit early if no data has changed.

I can remove the short circuit. Or I can work around the logic, it just might result in some hard to read code. Up to you

CyberDem0n:
> how important do you think that short circuiting logic is for Consul? I checked the other DCS implementations and it's not consistent. Raft and Etcd will write the key each time, and the others will exit early if no data has changed.

It all depends on how the expiration mechanism is implemented. Etcdv3, Consul, Zookeeper - we are using session mechanisms, and the member key is associated with a sessionId. If the session expires it causes also associated keys to expire. But, the sessions mechanism also allows us not to update keys if the value didn't change. In case of Etcdv2 or Raft every key has its own TTL, and therefore they have to be updated all the time.

IMO, it is better to avoid doing unnecessary writes to DCS, therefore short-circuting was implemented.

tommyzli:
A bit of duplicated logic, but now it does the correct thing:
```
# upon startup, with service_tags = ['baz', 'biz']
2021-07-21 14:34:38,648 INFO: Register service batman, params {'service_id': 'batman/postgresql0', 'address': '127.0.0.1', 'port': 5432, 'check': {'http': 'http://127.0.0.1:8008/master', 'interval': '5s', 'DeregisterCriticalServiceAfter': '150.0s'}, 'tags': ['baz', 'biz', 'master'], 'enable_tag_override': True}
...

# removed service_tags
2021-07-21 14:36:08.717 PDT [41668] LOG:  received SIGHUP, reloading configuration files
2021-07-21 14:36:09,726 INFO: Lock owner: postgresql0; I am postgresql0
2021-07-21 14:36:09,728 INFO: Register service batman, params {'service_id': 'batman/postgresql0', 'address': '127.0.0.1', 'port': 5432, 'check': {'http': 'http://127.0.0.1:8008/master', 'interval': '5s', 'DeregisterCriticalServiceAfter': '150.0s'}, 'tags': ['master'], 'enable_tag_override': True}
2021-07-21 14:36:09,770 INFO: no action. I am (postgresql0) the leader with the lock
2021-07-21 14:36:18,699 INFO: Lock owner: postgresql0; I am postgresql0
...

# turned off register_service
2021-07-21 14:46:26.018 PDT [41989] LOG:  received SIGHUP, reloading configuration files
2021-07-21 14:46:27,027 INFO: Lock owner: postgresql0; I am postgresql0
2021-07-21 14:46:27,030 INFO: Deregister service batman/postgresql0
2021-07-21 14:46:27,033 INFO: no action. I am (postgresql0) the leader with the lock
```

Other odd combinations look good as well (starting with service_tags defined but not register_service does not register, turning on register_service afterwards registers with previously set tags, etc) 
I feel like this has become complex enough to warrant it's own suite of behave tests 😅

patroni/dcs/consul.py
```diff
@@ -402,14 +412,23 @@ def touch_member(self, data, permanent=False):
...
+        if self._register_service or self._previous_loop_register_service:
+            try:
+                self.update_service(not create_member and member and member.data or {}, data)
+            except InvalidSession:
+                self._session = None
+                logger.error('Our session disappeared from Consul, can not "touch_member"')
+                return False
+            except Exception:
+                logger.exception('touch_member')
+                return False
```
CyberDem0n, line 424: The `update_service()` isn't supposed to raise exceptions, at least not the `InvalidSession`.

```suggestion
            try:
                self.update_service(not create_member and member and member.data or {}, data)
            except Exception:
                logger.exception('update_service')
```

CyberDem0n (commented):
> I feel like this has become complex enough to warrant it's own suite of behave tests sweat_smile

Well, yeah... But so far behave tests are not really DCS specific.

FxKu:
:+1:

CyberDem0n:
:+1:

CyberDem0n:
Merged, thank you @tommyzli!

PR #1994, opened 2021-07-01
Fix problem with PyInstaller-frozen code and dot in the path

rokm:
If `PyInstaller`-frozen application using `patroni` is placed in a location that contains a dot in the path, the `dcs_modules()` function in `patroni.dcs` breaks because `pkgutil.iter_importers()` treats the given path as a package name.

Ref: pyinstaller/pyinstaller#5944

This can be avoided altogether by not passing a path to `iter_importers()`, because `PyInstaller`'s `FrozenImporter` is a singleton and registered as top-level finder.

### DIFF
--- a/patroni/dcs/__init__.py
+++ b/patroni/dcs/__init__.py
@@ -68,7 +68,11 @@ def dcs_modules():
 
     if getattr(sys, 'frozen', False):
         toc = set()
-        for importer in pkgutil.iter_importers(dcs_dirname):
+        # dcs_dirname may contain a dot, which causes pkgutil.iter_importers()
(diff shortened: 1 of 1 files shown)


rokm:
As a side note, we're planning to add support for `pkgutil.iter_importers()` to `PyInstaller` (pyinstaller/pyinstaller#5959), which will alleviate the need for that PyInstaller-specific codepath in `dcs_modules()`.

sdudoladov:
:+1:

CyberDem0n:
:+1:

CyberDem0n:
Merged, thank you @rokm

Issue #2003, opened 2021-07-09
"Address already in use" error randomly occurs when starting Patroni 2.1.0


EtienneM:
**Describe the bug**

When starting Patroni 2.1.0, I randomly get the following error in my logs:

```
2021-07-09 15:36:18.174856098 +0200 CEST [postgresql-1] 2021-07-09 13:36:18,173 ERROR: Couldn't start a service on '0.0.0.0:8008', please check your `restapi.listen` configuration
2021-07-09 15:36:18.245599227 +0200 CEST [postgresql-1] 2021-07-09 13:36:18,173 ERROR: Failed to reload config_file=/etc/patroni/patroni.yml
2021-07-09 15:36:18.245608423 +0200 CEST [postgresql-1] Traceback (most recent call last):
2021-07-09 15:36:18.245609691 +0200 CEST [postgresql-1] File "/usr/local/lib/python2.7/dist-packages/patroni/__init__.py", line 78, in reload_config
2021-07-09 15:36:18.245610678 +0200 CEST [postgresql-1] self.api.reload_config(self.config['restapi'])
2021-07-09 15:36:18.245612089 +0200 CEST [postgresql-1] File "/usr/local/lib/python2.7/dist-packages/patroni/api.py", line 863, in reload_config
2021-07-09 15:36:18.245613119 +0200 CEST [postgresql-1] self.__initialize(config['listen'], ssl_options)
2021-07-09 15:36:18.245614224 +0200 CEST [postgresql-1] File "/usr/local/lib/python2.7/dist-packages/patroni/api.py", line 768, in __initialize
2021-07-09 15:36:18.245615615 +0200 CEST [postgresql-1] self.__httpserver_init(host, port)
2021-07-09 15:36:18.245665701 +0200 CEST [postgresql-1] File "/usr/local/lib/python2.7/dist-packages/patroni/api.py", line 747, in __httpserver_init
2021-07-09 15:36:18.245669373 +0200 CEST [postgresql-1] HTTPServer.__init__(self, info[0][-1][:2], RestApiHandler)
2021-07-09 15:36:18.245670568 +0200 CEST [postgresql-1] File "/usr/lib/python2.7/SocketServer.py", line 417, in __init__
2021-07-09 15:36:18.245824292 +0200 CEST [postgresql-1] self.server_bind()
2021-07-09 15:36:18.245851871 +0200 CEST [postgresql-1] File "/usr/lib/python2.7/BaseHTTPServer.py", line 108, in server_bind
[...]

PR #2004, opened 2021-07-09
Reload REST API certificate only on SIGHUP

CyberDem0n:
And make sure that cert number is cached when the RestApiServer object is created.

Close https://github.com/zalando/patroni/issues/2003


sdudoladov:
:+1:

CyberDem0n:
👍

Issue #1867, opened 2021-03-07
deploy patroni on Minikube


mgori24:
Gents, 
I'm testing Patroni manifest on Windows 10. Where I have Minikube installed. First, I built the image of Patroni from the Dockerfile from https://github.com/zalando/patroni/blob/master/kubernetes/Dockerfile.
 Then, I execute the kubectl to create from the manifest on this link: https://github.com/zalando/patroni/blob/master/kubernetes/patroni_k8s.yaml.
 I had an error of Missing property "selector". So I added the selector. After that I used kubectl create after the file update. I have error of Image bulling back. So I create a docker image and push it docker hub and then I updated the manifest to use the image that I created. Now I have the below error
/entrypoint.sh: line 2: $'\r': command not found
/entrypoint.sh: line 38: syntax error: unexpected end of file

for the entrypoint.sh I used the following file https://github.com/zalando/patroni/blob/master/kubernetes/entrypoint.sh.
Would please support or advice.

CyberDem0n:
There was a similar complaint about the docker-compose: https://github.com/zalando/patroni/issues/1841
Since I neither use Windows myself nor have it available for testing the stuff, I have absolutely no idea what is wrong with Docker on Windows.

mbanck:
Well, I get the same error about "selector" using minikube on Linux; I think it's probably a patroni <-> k8s issue?
```
patroni/kubernetes$ kubectl apply -f patroni_k8s.yaml 
service/patronidemo-config unchanged
error: error validating "patroni_k8s.yaml": error validating data: ValidationError(StatefulSet.spec): missing required field "selector" in io.k8s.api.apps.v1.StatefulSetSpec; if you choose to ignore these errors, turn validation off with --validate=false
```

CyberDem0n:
@mbanck the issue is in the yaml file in the StatefulSet definition.
You need to add the following to the spec:
```diff
 apiVersion: apps/v1
 kind: StatefulSet
 metadata:
   name: &cluster_name patronidemo
   labels:
     application: patroni
     cluster-name: *cluster_name
 spec:
+  selector:
+    matchLabels:
+      application: patroni
+      cluster-name: *cluster_name
   replicas: 3
   serviceName: *cluster_name
```

[...]

PR #2011, opened 2021-07-20
Add selector to StatefulSet definition

mbanck:
This adds the `selector` to the Patroni Kubernetes StatefulSet spec

Without it, one gets errors like
```
error: error validating "patroni_k8s.yaml": error validating data: ValidationError(StatefulSet.spec): missing required field "selector" in io.k8s.api.apps.v1.StatefulSetSpec; if you choose to ignore these errors, turn validation off with --validate=false
```
(as mentioned in #1867)

### DIFF
--- a/kubernetes/patroni_k8s.yaml
+++ b/kubernetes/patroni_k8s.yaml
@@ -20,6 +20,10 @@ metadata:
 spec:
   replicas: 3
   serviceName: *cluster_name
+  selector:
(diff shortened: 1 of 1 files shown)


sdudoladov:
:+1:

CyberDem0n:
:+1:

CyberDem0n:
Merged, thank you @mbanck

Issue #1758, opened 2020-11-14
Patroni wrong cert error


andrei-matei:
Hello all,

I am opening this here because I am not sure if it's a bug or if I am doing something wrong.

I am using Patroni (in Crunchy postgres-operator) and at the time it start, I get a weird error regarding SSL failing when trying to connect to Kubernetes API. Is there a place where I need to setup a CA so that communication works?

Here is the error:
`2020-11-14 09:59:41,806 WARNING: Retrying (Retry(total=0, connect=None, read=None, redirect=0, status=None)) after connection broken by 'SSLError(SSLError(1, '[SSL: CERTIFICATE_VERIFY_FAILED] certificate verify failed (_ssl.c:877)'),)': /api/v1/namespaces/pgo/pods?labelSelector=vendor%3Dcrunchydata%2Ccrunchy-pgha-scope%3Dhippo1
**2020-11-14 09:59:41,811 ERROR: Request to server https://10.32.0.1:443 failed: MaxRetryError("HTTPSConnectionPool(host='10.32.0.1', port=443): Max retries exceeded with url: /api/v1/namespaces/pgo/configmaps?labelSelector=vendor%3Dcrunchydata%2Ccrunchy-pgha-scope%3Dhippo1 (Caused by SSLError(SSLError(1, '[SSL: CERTIFICATE_VERIFY_FAILED] certificate verify failed (_ssl.c:877)'),))",)
2020-11-14 09:59:41,812 ERROR: Request to server https://10.32.0.1:443 failed: MaxRetryError("HTTPSConnectionPool(host='10.32.0.1', port=443): Max retries exceeded with url: /api/v1/namespaces/pgo/pods?labelSelector=vendor%3Dcrunchydata%2Ccrunchy-pgha-scope%3Dhippo1 (Caused by SSLError(SSLError(1, '[SSL: CERTIFICATE_VERIFY_FAILED] certificate verify failed (_ssl.c:877)'),))",)**
2020-11-14 09:59:42,734 ERROR: get_cluster
**Traceback (most recent call last):
  File "/usr/local/lib/python3.6/site-packages/patroni/dcs/kubernetes.py", line 700, in _load_cluster
    self._wait_caches(stop_time)
  File "/usr/local/lib/python3.6/site-packages/patroni/dcs/kubernetes.py", line 692, in _wait_caches
    raise RetryFailedError('Exceeded retry deadline')
patroni.utils.RetryFailedError: 'Exceeded retry deadline'**
[...]

PR #2025, opened 2021-08-06
make the CA bundle configurable for in-cluster Kubernetes config 

knackaron:
For instances where the full CA chain is not provided by the `ServiceAccount` secret, allow the location to be overridden by an environment variable.

`SSL_CERT_FILE` is the generally accepted environment variable to override the system CA bundle:
* https://pkg.go.dev/crypto/x509#SystemCertPool
* https://www.openssl.org/docs/manmaster/man7/openssl-env.html

An alternative environment variable to use could be the name of the variable used in the code already, `SERVICE_CERT_FILENAME`; this could avoid an potential inadvertent changes in the off-chance that someone already has `SSL_CERT_FILE` set and doesn't want Patroni using it.

The official Python Kubernetes client allows configuring an alternative CA bundle already: https://github.com/kubernetes-client/python/blob/d8f283e7483848647804eab345645106b6fb357d/kubernetes/client/configuration.py#L135

fixes https://github.com/zalando/patroni/issues/1758


CyberDem0n:
Naming is hard...
There is no "official" environment variable for specifying cacert for the K8s API server.
That means the variable name is very Patroni specific and would require a special treatment:
1. Probably it would make sense to support overriding the default value via `kubernetes.cacert` in the Patroni config
2. Therefore, naturally we can derive the name for the environment variable as `PATRONI_KUBERNETES_CACERT` (https://github.com/zalando/patroni/blob/master/docs/ENVIRONMENT.rst#kubernetes)

And of course, the documentation should be updated.

knackaron:
@CyberDem0n I've made the CA bundle configurable via the Patroni configuration file and the environment variable.  Documentation is also updated for both methods of configuration.

CyberDem0n (commented):
Now one unit test needs to be fixed: https://github.com/zalando/patroni/blob/master/tests/test_kubernetes.py#L71
Tests could be executed by running: `python setup.py test`

knackaron:
> Now one unit test needs to be fixed: https://github.com/zalando/patroni/blob/master/tests/test_kubernetes.py#L71
> Tests could be executed by running: `python setup.py test`

Updated and verified all tests pass.

FxKu:
:+1:

CyberDem0n:
:+1:

CyberDem0n:
Merged, thank you @knackaron!

PR #2032, opened 2021-08-12
Enrich history with the new leader

huiyalin525:
image
I added a column of 'master' in the patronictl history command,then the user can use the 'histroy' command to know who is the lmaster in the history

### DIFF
--- a/patroni/ctl.py
+++ b/patroni/ctl.py
@@ -1299,7 +1299,7 @@ def version(obj, cluster_name, member_names):
 def history(obj, cluster_name, fmt):
     cluster = get_dcs(obj, cluster_name).get_cluster()
     history = cluster.history and cluster.history.lines or []
-    table_header_row = ['TL', 'LSN', 'Reason', 'Timestamp', 'Master']
+    table_header_row = ['TL', 'LSN', 'Reason', 'Timestamp', 'New Leader']
     for line in history:
         if len(line) < len(table_header_row):
             add_coloumn_num = len(table_header_row) - len(line)
--- a/patroni/ctl.py
+++ b/patroni/ctl.py
@@ -1299,10 +1299,13 @@ def version(obj, cluster_name, member_names):
 def history(obj, cluster_name, fmt):
     cluster = get_dcs(obj, cluster_name).get_cluster()
     history = cluster.history and cluster.history.lines or []
+    table_header_row = ['TL', 'LSN', 'Reason', 'Timestamp', 'Master']
     for line in history:
-        if len(line) < 4:
-            line.append('')
-    print_output(['TL', 'LSN', 'Reason', 'Timestamp'], history, {'TL': 'r', 'LSN': 'r'}, fmt)
+        if len(line) < len(table_header_row):
+            add_coloumn_num = len(table_header_row) - len(line)
+            for _ in range(add_coloumn_num):
+                line.append('')
+    print_output(table_header_row, history, {'TL': 'r', 'LSN': 'r'}, fmt)
 
 
 def format_pg_version(version):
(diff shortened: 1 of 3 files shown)


patroni/ha.py
```diff
@@ -554,14 +554,16 @@ def update_cluster_history(self):
         elif not cluster_history or cluster_history[-1][0] != master_timeline - 1 or len(cluster_history[-1]) != 4:
```
CyberDem0n, line 554:
```suggestion
        elif not cluster_history or cluster_history[-1][0] != master_timeline - 1 or len(cluster_history[-1]) != 5:
```

patroni/ctl.py
```diff
@@ -1299,10 +1299,13 @@ def version(obj, cluster_name, member_names):
 def history(obj, cluster_name, fmt):
     cluster = get_dcs(obj, cluster_name).get_cluster()
     history = cluster.history and cluster.history.lines or []
+    table_header_row = ['TL', 'LSN', 'Reason', 'Timestamp', 'Master']
```
CyberDem0n, line 1302: Lets better call it Leader, which is in fact the New Leader, because it shows name of the leader on the new timeline.
```suggestion
    table_header_row = ['TL', 'LSN', 'Reason', 'Timestamp', 'New Leader']
```

Also, please keep in mind that information about the who was the leader on timeline=1 is lost forewer.

patroni/ha.py
```diff
@@ -554,14 +554,16 @@ def update_cluster_history(self):
...
             cluster_history = {line[0]: line for line in cluster_history or []}
             history = self.state_handler.get_history(master_timeline)
+            history[-1].append(self.cluster.leader.name)
```
CyberDem0n, line 557: The update_cluster_history() method is executed only by the leader.
```suggestion
            history[-1].append(self.state_handler.name)
```

patroni/ha.py
```diff
@@ -554,14 +554,16 @@ def update_cluster_history(self):
...
             if history and self.cluster.config:
                 history = history[-self.cluster.config.max_timelines_history:]
                 for line in history:
                     # enrich current history with promotion timestamps stored in DCS
                     if len(line) == 3 and line[0] in cluster_history \
-                            and len(cluster_history[line[0]]) == 4 \
+                            and len(cluster_history[line[0]]) == 5 \
                             and cluster_history[line[0]][1] == line[1]:
                         line.append(cluster_history[line[0]][3])
+                        line.append(cluster_history[line[0]][4])
```
CyberDem0n, line 566: What if we are dealing with the "old" format and there are only 4 columns in a row?
We need to support it as well.

Example:
old history in DCS
```json
[
[1,1,"no recovery target specified", "current date"]
]
```

After that we upgrade and restart Patroni with pause enabled and it should be rewritten as:
```json
[
[1,1,"no recovery target specified", "current date", "my-name"]
]
```

CyberDem0n (changes requested):

patroni/ha.py
huiyalin525: I have made changesmade change

patroni/ha.py
huiyalin525: I will get the leader operation and put it in the function patroni/postrgresql/init/get_history

patroni/ha.py
huiyalin525: I don't know this elif what to do

patroni/ctl.py
huiyalin525: fine this is nice

FxKu:
:+1:

CyberDem0n:
:+1:

CyberDem0n:
Merged, thank you @huiyalin525!

PR #2051, opened 2021-09-07
Add dcs_last_seen field to API

mbanck:
This field notes the last time (as unix epoch) a cluster member has
successfully communicated with the DCS. This is useful to identify and/or
analyse network partitions.

Also, expose dcs_last_seen in the MemberStatus class and its from_api_response
method.

### DIFF
--- a/patroni/api.py
+++ b/patroni/api.py
@@ -288,6 +288,10 @@ class RestApiHandler(BaseHTTPRequestHandler):
         metrics.append("# TYPE patroni_postgres_timeline counter")
         metrics.append("patroni_postgres_timeline{0} {1}".format(scope_label, postgres.get('timeline', 0)))
 
+        metrics.append("# HELP patroni_dcs_last_seen Epoch timestamp when DCS was last contacted successfully by Patroni.")
+        metrics.append("# TYPE patroni_dcs_last_seen gauge")
+        metrics.append("patroni_dcs_last_seen{0} {1}".format(scope_label, postgres.get('dcs_last_seen', 0)))
+
         self._write_response(200, '\n'.join(metrics)+'\n', content_type='text/plain')
 
     def _read_json_content(self, body_is_optional=False):
--- a/patroni/api.py
+++ b/patroni/api.py
@@ -583,9 +583,6 @@ class RestApiHandler(BaseHTTPRequestHandler):
         try:
             cluster = self.server.patroni.dcs.cluster
 
-            if cluster:
-                postgresql.dcs_last_seen = cluster.dcs_last_seen
-
             if postgresql.state not in ('running', 'restarting', 'starting'):
                 raise RetryFailedError('')
             stmt = ("SELECT " + postgresql.POSTMASTER_START_TIME + ", " + postgresql.TL_LSN + ","
@@ -603,7 +600,7 @@ class RestApiHandler(BaseHTTPRequestHandler):
                 'role': 'replica' if row[1] == 0 else 'master',
                 'server_version': postgresql.server_version,
                 'cluster_unlocked': bool(not cluster or cluster.is_unlocked()),
-                'dcs_last_seen': postgresql.dcs_last_seen,
+                'dcs_last_seen': self.server.patroni.ha.dcs_last_seen,
                 'xlog': ({
                     'received_location': row[4] or row[3],
                     'replayed_location': row[3],
--- a/patroni/api.py
+++ b/patroni/api.py
@@ -600,7 +600,7 @@ class RestApiHandler(BaseHTTPRequestHandler):
                 'role': 'replica' if row[1] == 0 else 'master',
                 'server_version': postgresql.server_version,
                 'cluster_unlocked': bool(not cluster or cluster.is_unlocked()),
-                'dcs_last_seen': self.server.patroni.ha.dcs_last_seen,
+                'dcs_last_seen': self.server.patroni.dcs.last_seen,
                 'xlog': ({
                     'received_location': row[4] or row[3],
                     'replayed_location': row[3],
--- a/patroni/api.py
+++ b/patroni/api.py
@@ -288,7 +288,8 @@ class RestApiHandler(BaseHTTPRequestHandler):
         metrics.append("# TYPE patroni_postgres_timeline counter")
         metrics.append("patroni_postgres_timeline{0} {1}".format(scope_label, postgres.get('timeline', 0)))
 
-        metrics.append("# HELP patroni_dcs_last_seen Epoch timestamp when DCS was last contacted successfully by Patroni.")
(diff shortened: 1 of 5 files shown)


patroni/ha.py
```diff
@@ -134,6 +137,9 @@ def load_cluster_from_dcs(self):
             self.old_cluster = cluster
         self.cluster = cluster
 
+        # Note time we last updated from DCS
+        self.cluster.dcs_last_seen = int(time.time())
+
```
CyberDem0n, line 142: 1. The `Cluster` object is inherited from a `namedtuple` and therefore immutable: https://github.com/zalando/patroni/blob/93efa91bbdc5a50cad4f99c4e99286340ab94ce8/patroni/dcs/__init__.py#L447-L461
2. Logically, the `Cluster` object contains only info that represents the state of Patroni cluster in DCS. I think the `dcs_last_seen` field doesn't fit there.

The AbstractDCS class is a better place for dcs_last_seen: https://github.com/zalando/patroni/blob/93efa91bbdc5a50cad4f99c4e99286340ab94ce8/patroni/dcs/__init__.py#L627-L656. It will also simplify changes in api.py: https://github.com/zalando/patroni/pull/2051/files#diff-13329cd3342d1757c2c9b4dd7a26c0170f4deee5d32eff5c2fd719cc9ff9375bR586-R588

mbanck: Thanks for the feedback Alexander, I'll look into it

mbanck: So you're saying there should be a new `_DCS_LAST_SEEN` = 'dcs_last_seen'` in there with all the boilerplate to it? AFAICT, that'd be a new REST API call like `https://localhost:8008/dcs_last_seen` am I getting this right?

CyberDem0n: No. It is only about adding a new property to AbstractDCS. In the API if could be exposed like you suggested, but there will be no need to copy it's value (line 587)

mbanck: I've added a commit, can you tell whether this is more like what you had in mind?

patroni/ha.py
```diff
@@ -227,6 +231,10 @@ def touch_member(self):
             if self.is_paused():
                 data['pause'] = True
 
+            # Note time we last updated DCS
+            self.dcs_last_seen = int(time.time())
+            data['dcs_last_seen'] = self.dcs_last_seen
+
```
CyberDem0n, line 237: This doesn't make much sense. If we can update the member information in DCS, it automatically means that we just have seen it.

patroni/dcs/__init__.py
```diff
@@ -652,6 +652,7 @@ def __init__(self, config):
         self._cluster_valid_till = 0
         self._cluster_thread_lock = Lock()
         self._last_lsn = ''
+        self._dcs_last_seen = 0
```
CyberDem0n, line 655: This is exactly what I meant.
Now we just need to set the value every time the `get_cluster()` call succeeds (right after `cluster = self._load_cluster()`)

patroni/api.py
```diff
@@ -600,6 +600,7 @@ def get_postgresql_status(self, retry=False):
                 'role': 'replica' if row[1] == 0 else 'master',
                 'server_version': postgresql.server_version,
                 'cluster_unlocked': bool(not cluster or cluster.is_unlocked()),
+                'dcs_last_seen': self.server.patroni.ha.dcs_last_seen,
```
CyberDem0n, line 603:
```suggestion
                'dcs_last_seen': self.server.patroni.ha.dcs.last_seen,
```

patroni/dcs/__init__.py
CyberDem0n: Maybe we also should rename it to the `_last_seen` and the respective property.

patroni/ha.py
```diff
@@ -75,6 +78,7 @@ def __init__(self, patroni):
         self._start_timeout = None
         self._async_executor = AsyncExecutor(self.state_handler.cancellable, self.wakeup)
         self.watchdog = patroni.watchdog
+        self.dcs_last_seen = 0
```
CyberDem0n, line 81: This one we don't need.

mbanck: Done

patroni/api.py
mbanck: `self.server.patroni..dcs.last_seen` (without `ha`) I guess, but done

patroni/dcs/__init__.py
mbanck: Done

patroni/ha.py
mbanck: Right ok, done

CyberDem0n (commented):
Should we also add `dcs_last_seen` to the `/metrics` (prometeus) endpoint?

patroni/ha.py
```diff
@@ -37,12 +39,13 @@ class _MemberStatus(namedtuple('_MemberStatus', ['member', 'reachable', 'in_reco
     def from_api_response(cls, member, json):
         is_master = json['role'] == 'master'
         timeline = json.get('timeline', 0)
+        dcs_last_seen = json.get('dcs_last_seen', 0)
         wal = not is_master and max(json['xlog'].get('received_location', 0), json['xlog'].get('replayed_location', 0))
-        return cls(member, True, not is_master, timeline, wal, json.get('tags', {}), json.get('watchdog_failed', False))
+        return cls(member, True, not is_master, dcs_last_seen, timeline, wal, json.get('tags', {}), json.get('watchdog_failed', False))
```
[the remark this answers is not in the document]
mbanck: I assume it complains about the overly long line and the s/wal/wall/ is incidental/errornous, will add the line break.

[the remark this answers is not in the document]
mbanck: About `/metrics`, that's a good point, will add.

patroni/api.py
```diff
@@ -288,6 +288,10 @@ def do_GET_metrics(self):
         metrics.append("# TYPE patroni_postgres_timeline counter")
         metrics.append("patroni_postgres_timeline{0} {1}".format(scope_label, postgres.get('timeline', 0)))
 
+        metrics.append("# HELP patroni_dcs_last_seen Epoch timestamp when DCS was last contacted successfully by Patroni.")
```
CyberDem0n, line 291: Flake8 doesn't like lines longer than 120 symbols: https://github.com/zalando/patroni/pull/2051/checks#step:5:8

mbanck: I've fixed that as well now. Sorry, I should've checked that myself.

FxKu:
:+1:

CyberDem0n:
:+1:

CyberDem0n:
Merged, thank you @mbanck!

PR #2054, opened 2021-09-10
Delay the next attempt of recovery till next HA loop

CyberDem0n:
If Postgres crashed due to out of disk space (for example) and fails to start because of that Patroni is too eagerly trying to recover it and producing too many logs


erthalion:
:+1:

CyberDem0n:
:+1:

Issue #2046, opened 2021-08-31
"Failed to drop replication slot" after patroni upgrade


kvaikla:
OS: RH 7.9
Pg: 13.3
Patroni: 2.1.0
Barman: 2.12

After patroni upgrade 1.6.5 -> 2.1.0, patroni logs errors:
```
Aug 31 17:39:22 pgrep1.energia.sise patroni[18285]: 2021-08-31 17:39:22,412 INFO: Lock owner: pgrep1; I am pgrep1
Aug 31 17:39:22 pgrep1.energia.sise patroni[18285]: 2021-08-31 17:39:22,520 ERROR: Failed to drop replication slot 'barman'
Aug 31 17:39:22 pgrep1.energia.sise patroni[18285]: 2021-08-31 17:39:22,626 INFO: no action. I am (pgrep1) the leader with the lock
```
Backup, replica, failover etc works fine. No errors in pg server log.
Solts in master (pgrep1) server:
```
postgres=# select * from pg_replication_slots ;
-[ RECORD 1 ]-------+-----------
slot_name           | pgrep2
plugin              |
slot_type           | physical
datoid              |
database            |
temporary           | f
active              | t
active_pid          | 90559
xmin                |
catalog_xmin        |
restart_lsn         | 4/EE0001E8
confirmed_flush_lsn |
wal_status          | reserved
safe_wal_size       |
-[ RECORD 2 ]-------+-----------
slot_name           | barman
plugin              |
slot_type           | physical
datoid              |
database            |
temporary           | f
active              | t
active_pid          | 94142
xmin                |
catalog_xmin        |
restart_lsn         | 4/EE000000
confirmed_flush_lsn |
wal_status          | reserved
safe_wal_size       |
```
Barman receivewal command
```
barman    94141  94136  0 17:25 ?        00:00:00 /usr/pgsql-13/bin/pg_receivewal --dbname=dbname=replication host=pgrep1.energia.sise options=-cdatestyle=iso port=5432 replication=true user=barman application_name=barman_receive_wal --verbose --no-loop --no-password --directory=/pgdata/backup/pgrep1:5432/streaming --slot=barman
```
pg config
```
$ patronictl show-config
loop_wait: 10
maximum_lag_on_failover: 1048576
postgresql:
  parameters:
    archive_command: null
    archive_mode: 'off'
    hot_standby: 'on'
    log_connections: 'on'
    log_disconnections: 'on'
    log_temp_files: '1'
    max_replication_slots: 8
[...]

PR #2055, opened 2021-09-10
Skip temporary replication slots while doing slot management

CyberDem0n:
Starting from v10 `pg_basebackup` creates a temporary replication slot for WAL streaming and Patroni was trying to drop it because the slot name looks unknown. In order to fix it, we skip all temporary slots when querying `pg_stat_replication_slots` view.

Another option to solve the problem would be running `pg_basebackup` with `--slot=current_node_name` option, but unfortunately at the moment when `pg_basebackup` is executed, we don't yet know the major version (the `--slot` option was added in v9.6).

Ref: https://github.com/zalando/patroni/issues/2046#issuecomment-912521502


RafiaSabih:
:+1:

CyberDem0n:
:+1:

PR #2056, opened 2021-09-11
Improve "I am" status messages

mbanck:
Currently, Patroni prints this every (by default) 10 seconds:
```
INFO: no action. I am (pg3) the leader with the lock
```
and
```
INFO: no action. I am a secondary (pg2) and following a leader (pg3)
```

I find the `I am foo` vs. `I am a secondary foo` galling, and propose to have that either
`I (pg3) am the leader [...] / I (pg2) am a secondary and [...]` or
`I am (pg3), the leader [...] / I am (pg2), a secondary, and [...]`.

I like the first better, but the latter flows somewhat better with other (infrequent) messages like
```
logger.info('Lock owner: %s; I am %s', lock_owner, self.state_handler.name)
[...]
return 'no action. I am ({0})'.format(self.state_handler.name)
```

### DIFF
--- a/patroni/ha.py
+++ b/patroni/ha.py
@@ -997,13 +997,13 @@ class Ha(object):
                     # in case of standby cluster we don't really need to
                     # enforce anything, since the leader is not a master.
                     # So just remind the role.
-                    msg = 'no action. I am ({0}) the standby leader with the lock'.format(self.state_handler.name) \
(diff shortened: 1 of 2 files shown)


CyberDem0n:
Please also update unit-tests and we are good.

RafiaSabih:
:+1:

CyberDem0n:
:+1:

CyberDem0n:
Merged, thank you @mbanck!

PR #2057, opened 2021-09-11
Add log before demoting (which can take some time)

mbanck:
While testing switchovers/failovers, I've noticed that it can take some time for the demote to finish and it might not be obvious from looking at the logs what exactly is going on.

So I propose to add a line at the beginning of the `demote` function.


RafiaSabih:
:+1:

CyberDem0n:
:+1:

CyberDem0n:
Merged, thank you @mbanck!

PR #2060, opened 2021-09-17
Ensure pg_replication_slot_advance() doesn't timeout

CyberDem0n:
The bigger gap between the slot flush LSN and the LSN we want to advance to becomes more time it takes for the call to finish.
Once started failing the "lag" will grow more or less infinitely, that have the following negative side-effects:
1. Size of pg_wal on the replica will grow
2. Size of pg_wal on the primary will grow
3. Since the hot_standby_feedback is forcefully enabled, the primary will stop cleaning up dead tuples
I.e., we are not only in danger of running out of disk space, but also increasing chances of transaction wraparound to happen.

In order to mitigate it, we want to set the `statement_timeout` to 0 before calling `pg_replication_slot_advance()`.

Since the call is happening from the main HA loop and could take more than `loop_wait`, the next heartbeat run could be delayed.
There is also a possibility that the call could take longer than `ttl` and the member key/session in DCS for a given replica expires, but, the slot LSN in DCS is updated by the primary every `loop_wait` seconds. Hence, we don't expect that the slot_advance() call will take significantly longer than the `loop_wait` and therefore chances of the member key/session to expire are very low.

### DIFF
--- a/patroni/postgresql/slots.py
+++ b/patroni/postgresql/slots.py
@@ -142,9 +142,9 @@ class SlotsHandler(object):
                 self._schedule_load_slots = True
 
     @contextmanager
-    def _get_local_connection_cursor(self, database):
+    def _get_local_connection_cursor(self, **kwargs):
         conn_kwargs = self._postgresql.config.local_connect_kwargs
-        conn_kwargs['database'] = database
+        conn_kwargs.update(kwargs)
         with get_connection_cursor(**conn_kwargs) as cur:
             yield cur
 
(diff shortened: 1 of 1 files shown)


erthalion:
:+1:

CyberDem0n:
:+1:

treyhendon:
Thanks for this!  Just curious if or when this will release because we are being affected by this issue with our use of Debezium.

K3ndu:
+1 @treyhendon when will this be released as we are also affected by this on live

treyhendon:
Updates for those following or who find this issue:
- Patroni included this fix in the [Patroni 2.1.2 release](https://github.com/zalando/patroni/blob/master/docs/releases.rst#version-212).
- Spilo updated from Patroni 2.1.1 to Patroni 2.1.3 in the [Spilo 2.1-p4 release](https://github.com/zalando/spilo/releases/tag/2.1-p4).
- If using the Postgres Operator, you should be able to update the Spilo image tag to `2.1-p4`.

PR #2065, opened 2021-09-22
Check only sync nodes when assessing failover capabilities

CyberDem0n:
When synchronous_mode is enabled we should check only synchronous nodes in is_failover_possible().


sdudoladov:
:+1:

CyberDem0n:
:+1:

PR #2067, opened 2021-09-24
Release the leader lock when pg_controldata reports "shut down"

CyberDem0n:
Due to different reasons, it could happen that WAL archiving on the primary stuck or significantly delayed. If we try to do a switchover or shut it down, the shutdown will take forever and will not finish until the whole backlog of WALs is processed.
In the meantime, Patroni keeps updating the leader lock, which prevents other nodes from starting the leader race even if it is known that they received/applied all changes.

The `Database cluster state:` is changed to `"shut down"` after:
- all data is fsynced to disk and the latest checkpoint is written to WAL
- all streaming replicas confirmed that they received all changes (including the latest checkpoint)
- at the same time, the archiver process continues to do its job and the postmaster process is still running.

In order to solve this problem and make the switchover more reliable/fast in a case when `archive_command` is slow/failing, Patroni will remove the leader key immediately after `pg_controldata` started reporting PGDATA as `"shut down"` cleanly and it verified that there is at least one replica that received all changes. If there are no replicas that fulfill the condition the leader key isn't removed and the old behavior is retained, i.e. Patroni will keep updating it.

### DIFF
--- a/patroni/ha.py
+++ b/patroni/ha.py
@@ -688,11 +688,11 @@ class Ha(object):
                         logger.info('Ignoring the former leader being ahead of us')
         return True
 
-    def is_failover_possible(self, members, only_syncronous=False, cluster_lsn=None):
+    def is_failover_possible(self, members, check_synchronous=True, cluster_lsn=None):
         ret = False
         cluster_timeline = self.cluster.timeline
         members = [m for m in members if m.name != self.state_handler.name and not m.nofailover and m.api_url]
-        if only_syncronous and self.is_synchronous_mode():
+        if check_synchronous and self.is_synchronous_mode():
             members = [m for m in members if self.cluster.sync.matches(m.name)]
         if members:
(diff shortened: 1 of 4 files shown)


patroni/ha.py
```diff
@@ -687,18 +688,21 @@ def _is_healthiest_node(self, members, check_replication_lag=True):
                         logger.info('Ignoring the former leader being ahead of us')
         return True
 
-    def is_failover_possible(self, members, check_synchronous=True):
+    def is_failover_possible(self, members, only_syncronous=False, cluster_lsn=None):
```
FxKu, line 691: The default of the flag changed. This function is called from `manual_failover_process_no_leader` without any flag. So it does not matter in that case - and can be false (because no leader)?

FxKu:
:+1:

CyberDem0n:
:+1:

PR #2068, opened 2021-09-25
Add sslcrldir connection parameter support

kostiantyn-nemchenko:
This allows setting the `sslcrldir` connection parameter that will be available since PostgreSQL 14.


FxKu:
:+1:

CyberDem0n:
:+1:

Issue #2074, opened 2021-09-30
switchover on async replica


glushakov:
**Describe the bug**
when doing manual switchover , i did not specify target node (just pressed enter) and patroni sometimes switchs on async replica

**To Reproduce**
make 3 nodes (1 leader, 1 sync, 1 async), and try to switch without specify target node

**Expected behavior**
patroni should not accept empty value OR should switchover only on sync standby (if available)

**Screenshots**
patronictl -c /etc/patroni/postgres1.yml switchover
Master [postgres3]:
Candidate ['postgres1', 'postgres2'] []:
When should the switchover take place (e.g. 2021-09-30T22:48 )  [now]:
Current cluster topology
+---------------------+---------------------+--------------+---------+----+-----------+
| Member              | Host                | Role         | State   | TL | Lag in MB |
+ Cluster: main (7013793607822447442) ------+--------------+---------+----+-----------+
| postgres1 | postgres1 | Sync Standby | running |  2 |         0 |
| postgres2 | postgres2 | Replica      | running |  2 |         0 |
| postgres3 | postgres3 | Leader       | running |  2 |           |
+---------------------+---------------------+--------------+---------+----+-----------+
Are you sure you want to switchover cluster main, demoting current master postgres3? [y/N]: y
2021-09-30 21:48:38.96713 Successfully switched over to "postgres2"
+---------------------+---------------------+---------+---------+----+-----------+
| Member              | Host                | Role    | State   | TL | Lag in MB |
+ Cluster: main (7013793607822447442) ------+---------+---------+----+-----------+
| postgres1 | postgres1 | Replica | running |  2 |         0 |
| postgres2 | postgres2 | Leader  | running |  2 |           |
| postgres3 | postgres3 | Replica | stopped |    |   unknown |
+---------------------+---------------------+---------+---------+----+-----------+
(patroni_venv) [09:48:39 postgres@postgres1:]:~$ patronictl -c /etc/patroni/postgres1.yml  list
+---------------------+---------------------+--------------+---------+----+-----------+
| Member              | Host                | Role         | State   | TL | Lag in MB |
+ Cluster: main (7013793607822447442) ------+--------------+---------+----+-----------+
| postgres1 | postgres1 | Replica      | running |  3 |         0 |
| postgres2 | postgres2 | Leader       | running |  3 |           |
| postgres3 | postgres3 | Sync Standby | running |  3 |         0 |
+---------------------+---------------------+--------------+---------+----+-----------+

**Environment**
[...]

PR #2076, opened 2021-10-04
Allow switchover only to sync nodes when synchronous replication is on

CyberDem0n:
Close https://github.com/zalando/patroni/issues/2074


CyberDem0n:
:+1:

Issue #2073, opened 2021-09-28
pg_rewind run before checkpoint issued on promoted standby


bradnicholson:
**Describe the bug**
We've had a number of replicas that have been lost with `FATAL:  requested timeline 6 does not contain minimum recovery point 0/6700C138 on timeline 5` type of errors, which I've traced back to an issue with Patroni triggering pg_rewind. Our experience mirrors [this](https://www.postgresql.org/message-id/20180713005241.GI1167%40paquier.xyz) issue on pgsql-hackers.

These occurred during problems unrelated to Patroni that was causing severe resource degradation on the host. Clearly a lot is wrong - statement_timeout is set to 2 seconds  and is being tripped here.

Considering the potentially destructive nature of pg_rewind, I wonder Patroni should be more defensive about not triggering in such conditions? As in, if it can't verify that the last completed checkpoint on the promoted leader is newer than the promotion time - don't run pg_rewind.

A high level pattern of behaviour during the issue:

2021-09-23 02:27:17 UTC  - last checkpoint completes on m-1 <- PG Control references this checkpoint after pg_rewind
2021-09-24 00:56:56 UTC - m-1 promotes to leader
2021-09-24 00:57:35 UTC - m-0 promotes to leader
2021-09-24 01:04:13 UTC - m-1 promotes to leader
2021-09-24 01:09:24 UTC - pg_rewind starts on m-0 from m-1 - pg_controldata references the checkpoint from the first line
2021-09-24 01:10:17 UTC - pg_rewind completes successfully.
2021-09-24 01:12:47 UTC next checkpoint completes on m-1

After this, we have a timeline divergence and the replica needs a resync.

These stack traces were present in m-1 logs `2021-09-24 01:04:13`

```
2021-09-24T01:03:59.84416731Z stdout F localhost:5432 - accepting connections
2021-09-24T01:03:59.91322582Z stderr F 2021-09-24 01:03:59,897 INFO: establishing a new patroni connection to the postgres cluster
2021-09-24T01:04:00.305352585Z stderr F 2021-09-24 01:04:00,240 ERROR: Exception when changing replication slots
2021-09-24T01:04:00.423345512Z stderr F Traceback (most recent call last):
2021-09-24T01:04:00.423380547Z stderr F   File "/usr/local/lib/python3.6/site-packages/patroni/ha.py", line 1433, in _run_cycle
2021-09-24T01:04:00.423392825Z stderr F     msg = self.process_healthy_cluster()
[...]

PR #2082, opened 2021-10-07
Use cached role as a fallback when postgres is slow

CyberDem0n:
In some extreme cases Postgres could be so slow that the normal monitoring query doesn't finish in a few seconds. It results in
the exception being raised from the `Postgresql._cluster_info_state_get()` method, which could lead to the situation that postgres isn't demoted on time.
In order to make it reliable we will catch the exception and use the cached state of postgres (`is_running()` and `role`) to determine whether postgres is running as a primary.

Close https://github.com/zalando/patroni/issues/2073


FxKu:
:+1:

PR #2084, opened 2021-10-08
Fix a few issues in kubernetes.py

CyberDem0n:
1. Two `TypeError`-s raised from `ApiClient.request()` method
2. Use the _retry() wrapper function instead of callable object in the `_update_leader_with_retry()` when trying to workaround concurrent updates of the leader object.


FxKu:
:+1:

PR #2088, opened 2021-10-15
Compatibility with psycopg 3.0

CyberDem0n:
By default `psycopg2` is preferred. The `psycopg>=3.0` will be used only if `psycopg2` is not available or its version is too old.


erthalion:
:+1:

CyberDem0n:
:+1:

Issue #2085, opened 2021-10-09
Do not setup replication slots if the status is in Starting state


rsmoorthy:
We have a production server, where there are many write operations going on. We are adding replicas to it. Because postgres replication process is single process (processing WAL files) , it takes a lot of time (>48 hours) until the replica changes the state from "Starting" to "Replica".
However, since the replication slots are created even when the replica is in Starting state (but the replica does not need those WAL files), the pg_xlog folder fills on the Primary server.

Our current option is to disable replication slots by "use_slots: false" and enable it manually when it reaches the "Replica" stage, but it would be good if patroni does not setup replication slots, if a replica is in Starting state.

Please let me know if this is acceptable for all scenarios and if the above suggestion makes sense.

rsmoorthy:
Hi @CyberDem0n 

Probably, I did not explain the problem sufficiently.

Scenario 1:
---------------
A cluster has a primary node, with lots of writes being done by application (200-300 WAL files copied/min). We are adding a replica to the cluster, which will start with "starting" state, bootstraps using base backup and then replaying WAL files via the "restore" command.  In a large database as ours (>10TB), it takes more than 9-10 hours for the replica to move from "starting" state to "running" state.
During this time,  if the replication slots are created on the primary, it will hold up the WAL files on the primary for these many hours, which will create a disk space issue.

Proposed Solution: Patroni does not setup replication slots, if the replica is not in running state.

Scenario 2:
---------------
In continuation with the above scenario or due to other reasons, if the replica has a huge lag (I am not entirely sure what is the threshold value that we should consider, it could be x times maximum_lag_on_failover), the replica is likely restoring the WAL files via the restore_command and not necessarily via streaming replication from the master. In large databases, when the lag is typically could be as high as 500GB (when the replica is coming up for the first time or stopped/started for a short period of time), it would not be helpful to have the replication slots created for this replica, as it could take 24-48 hours for the replica to catchup to 0 lag, which will again force the master to keep all the WAL files and eventually run out of disk space.

Proposed Solution: Patroni does not setup replication slots, if the replica has a huge lag. (I admit, I am not very sure of how much threshold should be considered as huge lag. I think it could even be 5-10 times of maximum_lag_on_failover, though the PR is just taking it as maximum_lag_on_failover)

[...]

PR #2092, opened 2021-10-18
Issue #2085: Do not setup replication slots on certain scenarios

rsmoorthy:
Refer to https://github.com/zalando/patroni/issues/2085 for details.

In short, there are two scenarios where we do not prefer to setup replication slots. 1) When the replica is not yet in running state 2) When there is a huge replication lag on the replica


rsmoorthy:
ok got it. thanks for your time.

Issue #2089, opened 2021-10-17
pick_synchronous_standby() does not seem to take nofailover tag into account


mbanck:
**Describe the bug**

Maybe I broke something while experimenting, but I wanted to file this in case it is a real problem.

If I have a 3-node cluster, and switch on sync mode; one replica is chosen to be the sync standby.

If one of the nodes has the `nofailover: true` tag for whatever reason, in my testing, that did not stop Patroni from assigning this node the sync standby role (and a quick look through `pick_synchronous_standby()` seems to indicate that there is indeed no provision for that).

```
root@pg1:~# patronictl -c /etc/patroni/14-main.yml list
+ Cluster: 14-main (7019633072855448066) -+----+-----------+------------------+
| Member | Host       | Role    | State   | TL | Lag in MB | Tags             |
+--------+------------+---------+---------+----+-----------+------------------+
| pg1    | 10.0.3.46  | Leader  | running | 47 |           |                  |
| pg2    | 10.0.3.165 | Replica | running | 47 |         0 | nofailover: true |
| pg3    | 10.0.3.252 | Replica | running | 47 |         0 |                  |
+--------+------------+---------+---------+----+-----------+------------------+
root@pg1:~# patronictl -c /etc/patroni/14-main.yml edit-config
--- 
+++ 
@@ -18,5 +18,5 @@
   remove_data_directory_on_rewind_failure: true
   use_pg_rewind: true
 retry_timeout: 10
-synchronous_mode: false
+synchronous_mode: true
 ttl: 30

Apply these changes? [y/N]: y
Configuration changed
root@pg1:~# patronictl -c /etc/patroni/14-main.yml list
+ Cluster: 14-main (7019633072855448066) ------+----+-----------+------------------+
| Member | Host       | Role         | State   | TL | Lag in MB | Tags             |
+--------+------------+--------------+---------+----+-----------+------------------+
| pg1    | 10.0.3.46  | Leader       | running | 47 |           |                  |
| pg2    | 10.0.3.165 | Sync Standby | running | 47 |         0 | nofailover: true |
| pg3    | 10.0.3.252 | Replica      | running | 47 |         0 |                  |
+--------+------------+--------------+---------+----+-----------+------------------+
```

This also then leaves cluster without a possibility to switchover, because Patroni only suggests `pg2` as target, but then aborts the switchover:

```
root@pg1:~# patronictl -c /etc/patroni/14-main.yml switchover
Master [pg1]: 
[...]

PR #2108, opened 2021-11-03
Prefer members without nofailover when picking sync nodes

CyberDem0n:
Previously sync nodes were selected only based on replication lag and hence the node with `nofailover` tag had the same chances to become synchrounous and any other node. That behavior was confusing and dangerous at the same time, because in case of failed primary the failover couldn't happen automatically.

Close https://github.com/zalando/patroni/issues/2089


sdudoladov:
:+1:

CyberDem0n:
:+1:

PR #2114, opened 2021-11-10
Optimize checkpoint after promote

CyberDem0n:
1. Avoid doing CHECKPOINT if `pg_control` is already updated.
2. Explicitly call ensure_checkpoint_after_promote() right after the bootstrap finished successfully.


erthalion:
:+1:

CyberDem0n:
:+1:

PR #2115, opened 2021-11-10
Avoid unnecessary updates of the member ZNode in Zookeeper

CyberDem0n:
When deciding whether the ZNode should be updated we rely on the cached version of the cluster, which is updated only when members ZNodes are deleted/created or the `/status`, `/sync`, `/failover`, `/config`, or `/history` ZNodes are updated.

I.e. after the update of the current member ZNode succeeded the cache becomes stale and all further updates are always performed even if the value didn't change. In order to solve it, we introduce the new attribute in the Zookeeper class and will use it for memorizing the actual value and for later comparison.

### DIFF
--- a/patroni/dcs/zookeeper.py
+++ b/patroni/dcs/zookeeper.py
@@ -105,6 +105,7 @@ class ZooKeeper(AbstractDCS):
 
         self._fetch_cluster = True
         self._fetch_status = True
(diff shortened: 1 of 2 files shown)


Jan-M:
:+1:

CyberDem0n:
:+1:

Issue #2113, opened 2021-11-09
Sometimes exception from prometheus monitoring


Thoro:
**Describe the bug**
Error is logged into console, no further issues found

**To Reproduce**
Run patroni on kubernetes with prometheus monitoring active and failover between instances.
I am not sure what causes the state.

**Expected behavior**
No error to be logged.

**Screenshots**

Errors being logged:

```
2021-11-09 16:31:04,268 INFO: PAUSE: postgres is not running
Exception in thread Thread-20:
Traceback (most recent call last):
  File "/usr/lib/python3.6/socketserver.py", line 654, in process_request_thread
    self.finish_request(request, client_address)
  File "/usr/lib/python3.6/socketserver.py", line 364, in finish_request
    self.RequestHandlerClass(request, client_address, self)
  File "/usr/lib/python3.6/socketserver.py", line 724, in __init__
    self.handle()
  File "/usr/lib/python3.6/http/server.py", line 418, in handle
    self.handle_one_request()
  File "/usr/local/lib/python3.6/dist-packages/patroni/api.py", line 634, in handle_one_request
    BaseHTTPRequestHandler.handle_one_request(self)
  File "/usr/lib/python3.6/http/server.py", line 406, in handle_one_request
    method()
  File "/usr/local/lib/python3.6/dist-packages/patroni/api.py", line 285, in do_GET_metrics
    metrics.append("patroni_cluster_unlocked{0} {1}".format(scope_label, int(postgres['cluster_unlocked'])))
KeyError: 'cluster_unlocked'

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/usr/lib/python3.6/threading.py", line 916, in _bootstrap_inner
    self.run()
  File "/usr/lib/python3.6/threading.py", line 864, in run
    self._target(*self._args, **self._kwargs)
  File "/usr/local/lib/python3.6/dist-packages/patroni/api.py", line 802, in process_request_thread
    super(RestApiServer, self).process_request_thread(request, client_address)
  File "/usr/lib/python3.6/socketserver.py", line 656, in process_request_thread
    self.handle_error(request, client_address)
  File "/usr/local/lib/python3.6/dist-packages/patroni/api.py", line 870, in handle_error
    address, port = client_address
ValueError: too many values to unpack (expected 2)
```

**Environment**
- Patroni version: 2.1-p3
- PostgreSQL version: 14
- DCS (and its version): Kubernetes 1.20 / etcd

[...]

PR #2116, opened 2021-11-10
Fix a few issues with Patroni API

CyberDem0n:
1. The `client_address` tuple may have more than two elements in case of IPv6
2. Return `cluster_unlocked` only when the value is true and handle it respectively in the do_GET_metrics()
3. Return `cluster_unlocked` and `dcs_last_seen` even if Postgres isn't running/queries timing out

Close https://github.com/zalando/patroni/issues/2113


Jan-M:
:+1:

CyberDem0n:
:+1:

Issue #2118, opened 2021-11-15
Patroni doesn't rewind standby if timeline is missing on master


fischerman:
**Describe the bug**
Given is a 2 node cluster with member A and B. A is master at TLI 132. Network problems cause A to demote. B is promoted and picks TLI 133. Some time later, B is demoted (apparently without replicating TLI 133 to A). Member A is promoted a switches to TLI 134, branching from TLI 132 (i.e. itself) at a point later than the branch point of TLI 133. Member B now has a diverging (old) timeline and must be rewound. However, Patroni doesn't attempt to rewind and instead starts the node. WAL replications fails because the timelines diverged.

**To Reproduce**

**Environment**
- Patroni version: 2.1.1
- PostgreSQL version: 12
- DCS (and its version): Kubernetes

**Patroni configuration file**: generated by Spilo
```
bootstrap:
  dcs:
    loop_wait: 10
    maximum_lag_on_failover: 33554432
    postgresql:
      parameters:
        archive_mode: 'on'
        archive_timeout: 1800s
        autovacuum_analyze_scale_factor: 0.02
        autovacuum_max_workers: 5
        autovacuum_vacuum_scale_factor: 0.05
        checkpoint_completion_target: 0.9
        hot_standby: 'on'
        log_autovacuum_min_duration: 0
        log_checkpoints: 'on'
        log_connections: 'on'
        log_disconnections: 'on'
        log_line_prefix: '%t [%p]: [%l-1] %c %x %d %u %a %h '
        log_lock_waits: 'on'
        log_min_duration_statement: 500
        log_statement: ddl
        log_temp_files: 0
        max_connections: 260
        max_replication_slots: 10
        max_wal_senders: 10
        tcp_keepalives_idle: 900
        tcp_keepalives_interval: 100
        track_functions: all
        wal_compression: 'on'
        wal_level: hot_standby
        wal_log_hints: 'on'
      use_pg_rewind: true
      use_slots: true
    retry_timeout: 10
    ttl: 30
  initdb:
  - auth-host: md5
  - auth-local: trust
  post_init: /scripts/post_init.sh "zalandos"
  users:
    zalandos:
      options:
      - CREATEDB
      - NOLOGIN
      password: ''
kubernetes:
  bypass_api_service: true
  labels:
    application: spilo
  pod_ip: 10.233.117.147
  ports:
  - name: postgresql
    port: 5432
  role_label: spilo-role
  scope_label: cluster-name
  use_endpoints: true
[...]

PR #2120, opened 2021-11-16
Handle missing timelines in history file when deciding to rewind

CyberDem0n:
When restore_command is configured Postgres is trying to fetch/apply all possible WAL segments and also fetch history files in order to select the correct timeline. It could result in a situation where the new history file will be missing some timelines.

Example:
- node1 demotes/crashes on timeline 1
- node2 promotes to timeline 2 and archives `00000002.history` and crashes
- node1 recovers as a replica, "replays" `00000002.history` and promotes to timeline 3

As a result, the `00000003.history` will not have the line with timeline 2, because it never replayed any WAL segment from it.
The `pg_rewind` tool is supposed to correctly handle such case when rewinding node2 from node1, but Patroni when deciding whether the rewind should happen was searching for the exact timeline in the history file from the new primary.

The solution is to assume that rewind is required if the current replica timeline is missing.

In addition to that this PR makes sure that the primary isn't running in recovery before starting the procedure of rewind check.

Close https://github.com/zalando/patroni/issues/2118 and https://github.com/zalando/patroni/issues/2124

### DIFF
--- a/patroni/dcs/kubernetes.py
+++ b/patroni/dcs/kubernetes.py
@@ -1011,7 +1011,7 @@ class Kubernetes(AbstractDCS):
     def touch_member(self, data, permanent=False):
         cluster = self.cluster
         if cluster and cluster.leader and cluster.leader.name == self._name:
-            role = 'promoted' if data['role'] in ('replica', 'promoted') else 'master'
+            role = 'master'
         elif data['state'] == 'running' and data['role'] != 'master':
             role = data['role']
         else:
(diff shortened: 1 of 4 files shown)


fischerman:
If timeline 2 is missing from the master node1, isn't that reason enough to need a rewind? Why do we still have to compare LSNs?

CyberDem0n:
@fischerman maybe you are right, in this case the fix would be even simpler:
```diff
                 elif parent_timeline > local_timeline:
+                    need_rewind = True
                     break
+            else:  # loop exit without break
+                need_rewind = True
```

FxKu:
:+1:

CyberDem0n:
:+1:

PR #2121, opened 2021-11-18
Optimize case when we don't have permanent logical slots

CyberDem0n:
The unnecessary call of SlotsHandler.process_permanent_slots() results in one additional query to pg_replication_slots view every HA loop.


sdudoladov:
:+1:

CyberDem0n:
:+1:

PR #2125, opened 2021-11-23
Clear event on the leader node when /status was updated

CyberDem0n:
Not doing so causing excessive HA loop runs with Zookeeper.
This moment wasn't fixed correctly in the #1875


FxKu:
:+1:

CyberDem0n:
:+1:

PR #2127, opened 2021-11-23
Remove duplicate hosts from the etcd machine cache

### LINKED ISSUE #2126: etcd machines_cache can have duplicate entries
**Describe the bug**

While researching retries in case of DCS failures, I noticed that `_calculate_timeouts()` gets 7 as `etcd_nodes` when I use etcd, but only 3 if I use etcd3. In both cases, etcd is running locally on a 3-node Patroni cluster, and all three nodes are mentioned in the `etcd:` (or `etcd3`) section.

Not sure where the difference is from, but `machines_cache()` for `etcd` gets three times three nodes, and removes only the duplicate `base_uri` (the etcd leader it seems) hosts, not the others, so 7 remain.

This implies that `retry_timeout` with  a 3-node etcd cluster is divided by 7 for `etcd` and only 3 for `etcd3`, which looks wrong to me? On the other hand, it is not a big deal.

**Expected behavior**

Same timeout for etcd and etcd3.

**Environment**
- Patroni version: 2.1.1
- PostgreSQL version: 14.1
- DCS (and its version): etcd-3.3.25

**Patroni configuration file**
```
etcd:
  hosts:
    - 10.0.3.195:2379
    - 10.0.3.237:2379
    - 10.0.3.178:2379
```

mbanck:
Fixes #2126

### DIFF
--- a/patroni/dcs/etcd.py
+++ b/patroni/dcs/etcd.py
@@ -184,7 +184,7 @@ class AbstractEtcdClientWithFailover(etcd.Client):
 
         for base_uri in machines_cache:
             try:
-                machines = list(self._get_members(base_uri, **kwargs))
+                machines = list(set(self._get_members(base_uri, **kwargs)))
                 logger.debug("Retrieved list of machines: %s", machines)
                 if machines:
                     random.shuffle(machines)


patroni/dcs/etcd.py
```diff
@@ -165,7 +165,7 @@ def _get_members(self, base_uri, **kwargs):
     @property
     def machines_cache(self):
         base_uri, cache = self._base_uri, self._machines_cache
-        return ([base_uri] if base_uri in cache else []) + [machine for machine in cache if machine != base_uri]
+        return ([base_uri] if base_uri in cache else []) + list(set([machine for machine in cache if machine != base_uri]))
```
CyberDem0n, line 168: IMO, https://github.com/zalando/patroni/blob/256a359a1ecd60ae7b81e1ac6524e1c85d2e78bd/patroni/dcs/etcd.py#L187 is a better place for it.

mbanck: Fair enough, I have changed it

FxKu:
:+1:

CyberDem0n:
:+1:

PR #2136, opened 2021-12-02
Release 2.1.2

CyberDem0n:
* Implement missing unit-tests
* Bump version
* Update release notes


docs/releases.rst
```diff
@@ -3,6 +3,110 @@
...
+
+  By default the ``psycopg2`` is preferred. The `psycopg>=3.0` will be used only if ``psycopg2`` is not available or its version is too old.
+
+- Add ``dcs_last_seen`` field to the REST API (Michael Banck)
+
+  This field notes the last time (as unix epoch) a cluster member has successfully communicated with the DCS. This is useful to identify and/or analyze network partitions.
+
+- Release the leader lock when ``pg_controldata`` reports "shut down" (Alexander)
+
+  In order to solve this problem of the slow switchover/shutdown in a case when ``archive_command`` is slow/failing, Patroni will remove the leader key immediately after ``pg_controldata`` started reporting PGDATA as ``shut down`` cleanly and it verified that there is at least one replica that received all changes. If there are no replicas that fulfill the condition the leader key isn't removed and the old behavior is retained, i.e. Patroni will keep updating the lock.
```
FxKu, line 21:
```suggestion
  To solve the problem of slow switchover/shutdown in case ``archive_command`` is slow/failing, Patroni will remove the leader key immediately after ``pg_controldata`` started reporting PGDATA as ``shut down`` cleanly and it verified that there is at least one replica that received all changes. If there are no replicas that fulfill this condition the leader key is not removed and the old behavior is retained, i.e. Patroni will keep updating the lock.
```

docs/releases.rst
```diff
@@ -3,6 +3,110 @@
...
+
+  It can take some time for the demote to finish and it might not be obvious from looking at the logs what exactly is going on.
+
+- Improve "I am" status messages (Michael)
+
+  ``no action. I am a secondary ({0})`` vs ``no action. I am ({0}), a secondary``
+
+- Cast to int ``wal_keep_segments`` when converting to ``wal_keep_size`` (Jorge Solórzano)
+
+  It is possible to specify ``wal_keep_segments`` as a string in the gloabl :ref:`dynamic configuration <dynamic_configuration>` and due to the python being dynamically typed language the string was simply multiplied. Example: ``wal_keep_segments: "100"`` was converted to ``100100100100100100100100100100100100100100100100MB``.
```
FxKu, line 48:
```suggestion
  It is possible to specify ``wal_keep_segments`` as a string in the global :ref:`dynamic configuration <dynamic_configuration>` and due to Python being a dynamically typed language the string was simply multiplied. Example: ``wal_keep_segments: "100"`` was converted to ``100100100100100100100100100100100100100100100100MB``.
```

docs/releases.rst
```diff
@@ -3,6 +3,110 @@
...
 =============
 
+Version 2.1.2
+-------------
+
+**New features**
+
+- Compatibility with ``psycopg>=3.0`` (Alexander Kukushkin)
+
+  By default the ``psycopg2`` is preferred. The `psycopg>=3.0` will be used only if ``psycopg2`` is not available or its version is too old.
```
FxKu, line 13:
```suggestion
  By default ``psycopg2`` is preferred. `psycopg>=3.0` will be used only if ``psycopg2`` is not available or its version is too old.
```

docs/releases.rst
```diff
@@ -3,6 +3,110 @@
...
+
+  It is possible to specify ``wal_keep_segments`` as a string in the gloabl :ref:`dynamic configuration <dynamic_configuration>` and due to the python being dynamically typed language the string was simply multiplied. Example: ``wal_keep_segments: "100"`` was converted to ``100100100100100100100100100100100100100100100100MB``.
+
+- Allow switchover only to sync nodes when synchronous replication is enabled (Alexander)
+
+  In addition to that do the leader race only against known synchronous nodes.
+
+- Use cached role as a fallback when Postgres is slow (Alexander)
+
+  In some extreme cases Postgres could be so slow that the normal monitoring query doesn't finish in a few seconds. The ``statement_timeout`` exception not being properly handled could lead to the situation that the Postgres wasn't demoted on time if the leader key expired or the update failed. In case of such exception Patroni will use the cached ``role`` in order to determine whether the Postgres is running as a primary.
```
FxKu, line 56:
```suggestion
  In some extreme cases Postgres could be so slow that the normal monitoring query does not finish in a few seconds. The ``statement_timeout`` exception not being properly handled could lead to the situation where Postgres was not demoted on time when the leader key expired or the update failed. In case of such exception Patroni will use the cached ``role`` to determine whether Postgres is running as a primary.
```

docs/releases.rst
```diff
@@ -3,6 +3,110 @@
...
+
+- Avoid unnecessary updates of the member ZNode (Alexander)
+
+  If there are no values have changed in the members data, the update shouldn't happen.
```
FxKu, line 60:
```suggestion
  If no values have changed in the members data, the update should not happen.
```

docs/releases.rst
```diff
@@ -3,6 +3,110 @@
...
+
+- Optimize checkpoint after promote (Alexander)
+
+  Avoid doing ``CHECKPOINT`` if the latest timeline is already stored in the ``pg_control``. It helps to avoid unnecessary ``CHECKPOINT`` right after initializing the new cluster with the ``initdb``.
```
FxKu, line 64:
```suggestion
  Avoid doing ``CHECKPOINT`` if the latest timeline is already stored in ``pg_control``. It helps to avoid unnecessary ``CHECKPOINT`` right after initializing the new cluster with ``initdb``.
```

docs/releases.rst
```diff
@@ -3,6 +3,110 @@
...
+
+- Prefer members without ``nofailover`` when picking sync nodes (Alexander)
+
+  Previously sync nodes were selected only based on the replication lag and hence the node with ``nofailover`` tag had the same chances to become synchrounous as any other node. That behavior was confusing and dangerous at the same time, because in case of failed primary the failover couldn't happen automatically.
```
FxKu, line 68:
```suggestion
  Previously sync nodes were selected only based on the replication lag, hence the node with ``nofailover`` tag had the same chances to become synchronous as any other node. That behavior was confusing and dangerous at the same time because in case of a failed primary the failover could not happen automatically.
```

docs/releases.rst
```diff
@@ -3,6 +3,110 @@
...
+
+- Remove duplicate hosts from the etcd machine cache (Michael)
+
+  Advertised client URLs in the etcd cluster could be missconfigured. Removing duplicates in Patroni in this case is a low-hanging fruit.
```
FxKu, line 72:
```suggestion
  Advertised client URLs in the etcd cluster could be misconfigured. Removing duplicates in Patroni in this case is a low-hanging fruit.
```
Does one say such things as "low-hanging fruit" in release notes? :smile: :apple:

docs/releases.rst
```diff
@@ -3,6 +3,110 @@
...
+
+  Starting from v10 ``pg_basebackup`` creates a temporary replication slot for WAL streaming and Patroni was trying to drop it because the slot name looks unknown. In order to fix it, we skip all temporary slots when querying ``pg_stat_replication_slots`` view.
+
+- Ensure ``pg_replication_slot_advance()`` doesn't timeout (Alexander)
+
+  Patroni was using the default ``statement_timeout`` in this case and once the call failed there are very high chances that it will never recover, resulting in increased size of ``pg_wal`` and ``pg_catalog`` bloat.
+
+- The ``/status`` wasn't updated on demote (Alexander)
+
+  After demoting PostgreSQL the old leader leader updating the last LSN in DCS. Starting from ``2.1.0`` the new ``/status`` key was introduced, but the optime was still written to the ``/optime/leader``.
```
FxKu, line 87:
```suggestion
  After demoting PostgreSQL the old leader updates the last LSN in DCS. Starting from ``2.1.0`` the new ``/status`` key was introduced, but the optime was still written to the ``/optime/leader``.
```

docs/releases.rst
```diff
@@ -3,6 +3,110 @@
...
+
+- Handle DCS exceptions when demoting (Alexander)
+
+  While doing demote due to failure to update the leader lock it could happen that DCS goes completely down and the ``get_cluster()`` call raise the exception. Not being properly handled it results in Postgres remaining stopped until DCS recovers.
```
FxKu, line 91:
```suggestion
  While demoting the master due to failure to update the leader lock it could happen that DCS goes completely down and the ``get_cluster()`` call raises an exception. Not being handled properly it results in Postgres remaining stopped until DCS recovers.
```

FxKu:
:+1:

treyhendon:
Looking forward to this release!  Will this require an update to the Postgres Operator as well?  Or is the `2.1-p2` a floating tag?

CyberDem0n:
:+1:

CyberDem0n:
@treyhendon Patroni is independent from Postgres Operator and/or Spilo. Eventually (maybe next week) we will rebuild Spilo with the latest Patroni.

Issue #2137, opened 2021-12-06
Switchover/Failover fails if cascade replication and replication slots are used


velechac:
**Describe the bug**
Patroni is not able to provide switchover or failover if cascade replication and replication slots are used.

**To Reproduce**
* 4 nodes in cascade replication

   ![pg-replication-schema_2-2](https://user-images.githubusercontent.com/1170359/144832821-f8639a33-ace6-41aa-b9f1-67ac4e3f1e98.png)

* `tags.replicatefrom` some node must be set on replica2
* `postgresql.use_slots: True`
* call `patronictl switchover` or `patronictl failover`

**Expected behavior**
Patroni should provide switchover or failover.

**Environment**
- Patroni version: >=2.1.0
- PostgreSQL version: tested with 12, 13, 14
- DCS (and its version): etcd

**Patroni configuration file**
```
bootstrap:
  dcs:
    loop_wait: 10
    maximum_lag_on_failover: 0
    postgresql:
      parameters:
        max_connections: '100'
        max_locks_per_transaction: 64
        max_prepared_transactions: 0
        max_replication_slots: '5'
        max_wal_senders: '10'
        max_worker_processes: 8
        ssl_cert_file: ... 
        ssl_key_file: .. 
        track_commit_timestamp: 'off'
        wal_keep_segments: '100'
        wal_keep_size: '1600'
        wal_level: hot_standby
        wal_log_hints: 'on'
      use_pg_rewind: true
    retry_timeout: 10
    ttl: 30
  initdb:
  - encoding: UTF8
  - data-checksums
  pg_hba:
  - host replication replicator 127.0.0.1/32 md5
  - host replication replicator sandbox1 md5
  - host replication replicator sandbox2 md5
  - host replication replicator sandbox3 md5
  - host replication replicator sandbox4 md5
  - local all postgres  peer map=admins
  - local all all  md5 
  - hostssl all postgres 0.0.0.0/0 cert 
  - hostssl all postgres ::0/0 cert 
  - host all all 0.0.0.0/0 md5 
  - host all all ::0/0 md5 
  pg_ident:
  - admins postgres postgres
  - admins root postgres
  - admins prometheus postgres
  post_bootstrap: /tmp/post_bootstrap.sh
etcd:
  cacert: ...
  cert: ...
  hosts:
   - etcd1:2379
   - etcd2:2379
   - etcd3:2379
  key: ... 
  password: XXX
  protocol: https
  username: sandbox-test
log:
  level: INFO
[...]

PR #2138, opened 2021-12-06
Fix bug with failover to cascading standby

CyberDem0n:
When figuring out which slots should be created on cascading standby we forgot to take into account that the leader might be absent.

Close: https://github.com/zalando/patroni/issues/2137


FxKu:
:+1:

CyberDem0n:
:+1:

petqa:
good day, how soon will be release with this pr?

PR #2144, opened 2021-12-17
Fix bug in the bootstrap standby-leader

CyberDem0n:
When starting postgres after bootstrap of the standby-leader the `follow()` method is used to always return `True`.
This behavior was changed in the #2054 in order to avoid hammering logs if postgres is failing to start.

Since now the method returns `None` if postgres didn't start accepting connections after 60s, the change broke the standby-leader bootstrap code.

As the solution, we will assume that the clone was successful if the `follow()` method returned anything different from `False`.


FxKu:
:+1:

CyberDem0n:
:+1:

PR #2146, opened 2021-12-17
Rearrange tests

CyberDem0n:
- remove codacy steps: they removed legacy organizations and there seems to be no easy way of installing codacy app to the Zalando GH.
- Don't run behave on MacOS: recently worker became way to slow
- Disable behave for combination of kubernetes and python 2.7
- Remove python 3.5 (it will be removed by GH from workers in January) and add 3.10
- Run behave with 3.6 and 3.9 instead of 3.5 and 3.8

### DIFF
--- a/.github/workflows/install_deps.py
+++ b/.github/workflows/install_deps.py
@@ -110,7 +110,7 @@ def install_etcd():
 
 
 def install_postgres():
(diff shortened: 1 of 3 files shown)


FxKu:
:+1:

CyberDem0n:
:+1:

PR #2159, opened 2022-01-05
Annual cleanup

CyberDem0n:
-  Simplify setup.py: remove unneeded features and get rid of deprecation warnings
-  Compatibility with Python 3.10: handle `threading.Event.isSet()` deprecation
-  Make sure setup.py could run without `six`:: move Patroni class and main function to the __main__.py. The __init__.py will have only a few functions used by the Patroni class and from the setup.py


sdudoladov:
:+1:

CyberDem0n:
:+1:

Issue #2162, opened 2022-01-08
Requires superuser connection to primary on standby restart if rewind user is not set


mbanck:
**Describe the bug**
If the rewind user is not set, and Patroni is restarted on a follower node (initial bootstrap or reinit go fine), `check_leader_is_not_in_recovery()` will use the postgres user (I added a debug output line to show `conn_kwargs`), and if no `postgres` password is set, Patroni outputs error messages:
```
Jan 08 13:54:14 pg2 systemd[1]: Started Patroni instance 11-main.
Jan 08 13:54:15 pg2 patroni@11-main[7909]: 2022-01-08 13:54:15,089 INFO: Selected new etcd server http://10.0.3.178:2379
Jan 08 13:54:15 pg2 patroni@11-main[7909]: 2022-01-08 13:54:15,095 INFO: No PostgreSQL configuration items changed, nothing to reload.
Jan 08 13:54:15 pg2 patroni@11-main[7909]: 2022-01-08 13:54:15,106 WARNING: Postgresql is not running.
Jan 08 13:54:15 pg2 patroni@11-main[7909]: 2022-01-08 13:54:15,107 INFO: Lock owner: pg1; I am pg2
Jan 08 13:54:15 pg2 patroni@11-main[7909]: 2022-01-08 13:54:15,111 INFO: pg_controldata:
Jan 08 13:54:15 pg2 patroni@11-main[7909]:   pg_control version number: 1100
[...]

Jan 08 13:54:15 pg2 patroni@11-main[7909]: 2022-01-08 13:54:15,113 INFO: Lock owner: pg1; I am pg2
Jan 08 13:54:15 pg2 patroni@11-main[7909]: 2022-01-08 13:54:15,117 INFO: Local timeline=1 lsn=0/4000108
Jan 08 13:54:15 pg2 patroni@11-main[7909]: 2022-01-08 13:54:15,117 INFO: check_leader_is_not_in_recovery({'host': '10.0.3.178', 'port': 5432, 'dbname': 'postgres', 'user': 'postgres'})
Jan 08 13:54:15 pg2 patroni@11-main[7909]: 2022-01-08 13:54:15,125 ERROR: Exception when working with leader
Jan 08 13:54:15 pg2 patroni@11-main[7909]: Traceback (most recent call last):
[...]

PR #2165, opened 2022-01-10
Use replication credentials when checking leader status

CyberDem0n:
It could be that `remove_data_directory_on_diverged_timelines` is set, but there is no `rewind_credentials` defined and superuser access between nodes is not allowed.

Close https://github.com/zalando/patroni/issues/2162

### DIFF
--- a/patroni/postgresql/rewind.py
+++ b/patroni/postgresql/rewind.py
@@ -165,7 +165,7 @@ class Rewind(object):
             return
 
         if not self.check_leader_is_not_in_recovery(
-                self._conn_kwargs(leader, self._postgresql.config.rewind_credentials)):
+                self._conn_kwargs(leader, self._postgresql.config.replication)):
             return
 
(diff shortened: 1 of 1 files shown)


mbanck:
LGTM, I tested this successfully

FxKu:
:+1:

CyberDem0n:
:+1:

v4nTh3M4n:
When `remove_data_directory_on_diverged_timelines` is not set and there are no `rewind_credentials` the check_leader_is_not_in_recovery function fails. This function gets called when the container has restarted (not at initialization).

According to the postgres logs - the replication user does not have connect permissions on the postgres database to check if the leader is in recovery or not.
Should that grant be created and applied automatically based on the replication user defined in the template? 

When I do a `grant connect on database postgres to replication_user` the rewind script works again and it is able to query the required info from the leader.

Issue #2180, opened 2022-01-21
etcd3 _load_cluster() does not fail if etcd is down


mbanck:
**Describe the bug**

I noticed the `dcs_last_seen` flag is being updated by Patroni internally (I added some debugging to print it out) even if etcd is unavailable. The same seems to be not true if I use etcd (i.e. etcdv2)

**To Reproduce**

I added debug logging at the beginning and end of `_load_cluster()` as well, with etcd3, I get:

```
Jan 21 20:46:52 pg1 patroni@11-main2[47010]: 2022-01-21 20:46:52,283 INFO: _load_cluster start
Jan 21 20:46:52 pg1 patroni@11-main2[47010]: 2022-01-21 20:46:52,283 INFO: _load_cluster end
Jan 21 20:46:52 pg1 patroni@11-main2[47010]: 2022-01-21 20:46:52,284 INFO: Updating _last_seen to 1642798012
Jan 21 20:46:52 pg1 patroni@11-main2[47010]: 2022-01-21 20:46:52,289 INFO: no action. I am (pg1), a secondary, and following a leader (pg3)
[at this point I sent SIGSTOP to etcd]
Jan 21 20:47:02 pg1 patroni@11-main2[47010]: 2022-01-21 20:47:02,283 INFO: _load_cluster start
Jan 21 20:47:02 pg1 patroni@11-main2[47010]: 2022-01-21 20:47:02,283 INFO: _load_cluster end
[_load_cluster returns immediately, and _last_seen is updated:]
Jan 21 20:47:02 pg1 patroni@11-main2[47010]: 2022-01-21 20:47:02,284 INFO: Updating _last_seen to 1642798022
Jan 21 20:47:02 pg1 patroni@11-main2[47010]: 2022-01-21 20:47:02,287 INFO: no action. I am (pg1), a secondary, and following a leader (pg3)
Jan 21 20:47:12 pg1 patroni@11-main2[47010]: 2022-01-21 20:47:12,283 INFO: _load_cluster start
Jan 21 20:47:12 pg1 patroni@11-main2[47010]: 2022-01-21 20:47:12,284 INFO: _load_cluster end
Jan 21 20:47:12 pg1 patroni@11-main2[47010]: 2022-01-21 20:47:12,284 INFO: Updating _last_seen to 1642798032
Jan 21 20:47:15 pg1 patroni@11-main2[47010]: 2022-01-21 20:47:12,284 INFO: Lock owner: pg3; I am pg1
[...]

PR #2186, opened 2022-01-27
Restart etcd3 watcher if all etcd nodes don't respond

CyberDem0n:
Close https://github.com/zalando/patroni/issues/2180


FxKu:
:+1:

Issue #2189, opened 2022-01-27
standby_cluster-side failover


mbanck:
I tried to look through all the (open or closed) issues with "standby_cluster" in it and couldn't find a discussion about this, so here goes:

1. It is well-known that libpq application-side failover is possible by putting all the hosts of a cluster in the connection string (like `host=pg1,pg2,pg3`), along with `target_session_attrs=read-write`. On failover or switchover the application will reconnect to the new primary.
2. In principle, I think it is possible to have this work with a stand-alone replication standby, i.e. put all the hosts in the `primary_conninfo`, along with `target_sesion_attrs`. The latter is not something one can do on the `pg_basebackup` command line (unless maybe one passes the full DSN and not `-h pg1,pg2,pg3`, not sure), but one can edit `postgresql.auto.conf` (for v12+) as written by `pg_basebackup`, add `target_session_attrs=read-write` and restart the standby. When there is a switchover/failover of the cluster, the stand-alone standby will automatically switch to following the new leader.
3. Turning to Patroni, when setting up a standby cluster, is possible to put more than one host in the `standby_cluster/host:` config, like this:
```
dcs:
  standby_cluster:
    host: pg1,pg2,pg3
```
4. This will put the hosts correctly into `primary_conninfo`, however, `primary_conninfo` is (for v12+) managed in `postgresql.conf` and it does not have `target_session_attrs=read-write`; maybe this could just be added as it shouldn't hurt (or are there use-cases for as standby-cluster not replicating from a leader?). One can edit it in of course, but then one has to make sure Patroni doesn't overwrite it all the time.
5. The problem with using `host: pg1,pg2,pg3` is that the pgpass logic just copies that host string into the designated pgpass file, and then PostgreSQL does not recognize the host. Again, one can edit this, but in my experience Patroni overwrites this as well.
6. An alternative would be something like
```
dcs:
  standby_cluster:
    host:
    - pg1
    - pg2
    - pg3
```
However, that fails because Patroni then passes something like `['pg1, 'pg2, 'pg3']` as host to `pg_basebackup` which is not a valid hostname: `pg_basebackup: error: could not translate host name "['pg1'" to address: Name or service not known`
7. So either the pgpass or the `pg_basebackup` logic would need to be changed along with adding `primary_session_attrs=read-write` to `primary_conninfo`.
8. Anything else I'm missing?

[...]

PR #2192, opened 2022-01-28
Write pgpass line per host if more than one is specified in connstr

mbanck:
Partly addresses #2189

### DIFF
--- a/patroni/postgresql/config.py
+++ b/patroni/postgresql/config.py
@@ -745,7 +745,12 @@ class ConfigHandler(object):
                 return re.sub(r'([:\\])', r'\\\1', str(value))
 
             record = {n: escape(record.get(n) or '*') for n in ('host', 'port', 'user', 'password')}
-            return '{host}:{port}:*:{user}:{password}'.format(**record)
+            # 'host' could be several comma-separated hostnames, in this case
+            # we need to write on pgpass line per host
+            line = ''
+            for hostname in record.get('host').split(','):
+                line += hostname + ':{port}:*:{user}:{password}'.format(**record) + '\n'
+            return line.rstrip()
 
     def write_pgpass(self, record):
         line = self._pgpass_line(record)
--- a/patroni/postgresql/config.py
+++ b/patroni/postgresql/config.py
@@ -618,19 +618,19 @@ class ConfigHandler(object):
 
     def _check_passfile(self, passfile, wanted_primary_conninfo):
         # If there is a passfile in the primary_conninfo try to figure out that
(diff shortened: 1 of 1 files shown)


mbanck:
Thanks, will look

mbanck:
I'm having a hard time hitting `_check_passfile()` - it seems at least for my setups `_check_primary_conninfo()` always succesfully edits in the password and then pgpass isn't being checked.

That said, I also tried to force running `_check_passfile()` but did not have success so far in figuring out how to successfully compare those multi-line files/strings reliably. I will look further at this, but wanted to check back whether `_check_passfile()` is maybe obsolete?

CyberDem0n:
> I'm having a hard time hitting _check_passfile() - it seems at least for my setups _check_primary_conninfo() always succesfully edits in the password and then pgpass isn't being checked.

There is a little optimization involved. The content of the file isn't checked if it is known that the file hasn't been modified since the last time (mtime).

> did not have success so far in figuring out how to successfully compare those multi-line files/strings reliably.

I think the simplest approach would be making a `set` from these lines with later comparison of two sets.

mbanck:
Ok, I came up with [8c30ab4](https://github.com/zalando/patroni/pull/2192/commits/8c30ab458ff4547befe4c6da1c65d9f6192f63c1) which works in my testing; not sure about the unit test failure on Windows though, is it related?

patroni/postgresql/config.py
```diff
@@ -618,19 +618,19 @@ def _read_recovery_params_pre_v12(self):
...
                 with open(passfile) as f:
-                    wanted_line = self._pgpass_line(wanted_primary_conninfo).strip()
-                    for raw_line in f:
-                        if raw_line.strip() == wanted_line:
-                            self._passfile = passfile
-                            self._passfile_mtime = passfile_mtime
-                            return True
+                    wanted_lines = self._pgpass_line(wanted_primary_conninfo).splitlines()
+                    file_lines = f.read().splitlines()
+                    if (wanted_lines == file_lines):
```
CyberDem0n, line 630: Since the order of lines doesn't really matter in the pgpassfile we better compare sets:
```suggestion
                    if set(wanted_lines) == set(file_lines):
```

mbanck: Thanks, did it this way in [3c7aac7](https://github.com/zalando/patroni/pull/2192/commits/3c7aac785a04d36dde7f2b673b2bb434b5e8b53c)

FxKu:
:+1:

CyberDem0n:
:+1:

PR #2193, opened 2022-01-28
Add target_session_attrs=read-write to standby_leader primary_conninfo

mbanck:
This allows to have multiple hosts in a standby_cluster and ensures that the
standby leader follows the main cluster's new leader after a switchover.

Addresses #2189

### DIFF
--- a/patroni/postgresql/config.py
+++ b/patroni/postgresql/config.py
@@ -480,13 +480,19 @@ class ConfigHandler(object):
             ret['krbsrvname'] = self._krbsrvname
         if 'dbname' in ret:
             del ret['dbname']
+        # Add target_session_attrs in case more than one hostname is specified
+        # (libpq client-side failover) making sure we hit the primary
+        if not 'target_session_attrs' in ret and self._postgresql.major_version >= 100000:
+            ret.setdefault('target_session_attrs', 'read-write')
+        logger.info('primary_conninfo_params, returning %s', ret)
         return ret
 
     def format_dsn(self, params, include_dbname=False):
         # A list of keywords that can be found in a conninfo string. Follows what is acceptable by libpq
         keywords = ('dbname', 'user', 'passfile' if params.get('passfile') else 'password', 'host', 'port',
                     'sslmode', 'sslcompression', 'sslcert', 'sslkey', 'sslpassword', 'sslrootcert', 'sslcrl',
-                    'sslcrldir', 'application_name', 'krbsrvname', 'gssencmode', 'channel_binding')
+                    'sslcrldir', 'application_name', 'krbsrvname', 'gssencmode', 'channel_binding',
+                    'target_session_attrs')
         if include_dbname:
             params = params.copy()
             if 'dbname' not in params:
--- a/patroni/postgresql/config.py
+++ b/patroni/postgresql/config.py
@@ -484,7 +484,6 @@ class ConfigHandler(object):
         # (libpq client-side failover) making sure we hit the primary
         if 'target_session_attrs' not in ret and self._postgresql.major_version >= 100000:
             ret.setdefault('target_session_attrs', 'read-write')
-        logger.info('primary_conninfo_params, returning %s', ret)
         return ret
 
     def format_dsn(self, params, include_dbname=False):
--- a/patroni/postgresql/config.py
+++ b/patroni/postgresql/config.py
@@ -482,7 +482,7 @@ class ConfigHandler(object):
             del ret['dbname']
         # Add target_session_attrs in case more than one hostname is specified
         # (libpq client-side failover) making sure we hit the primary
-        if not 'target_session_attrs' in ret and self._postgresql.major_version >= 100000:
+        if 'target_session_attrs' not in ret and self._postgresql.major_version >= 100000:
             ret.setdefault('target_session_attrs', 'read-write')
         logger.info('primary_conninfo_params, returning %s', ret)
         return ret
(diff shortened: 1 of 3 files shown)


patroni/postgresql/config.py
```diff
@@ -480,13 +480,19 @@ def primary_conninfo_params(self, member):
             ret['krbsrvname'] = self._krbsrvname
         if 'dbname' in ret:
             del ret['dbname']
+        # Add target_session_attrs in case more than one hostname is specified
+        # (libpq client-side failover) making sure we hit the primary
+        if not 'target_session_attrs' in ret and self._postgresql.major_version >= 100000:
```
CyberDem0n, line 485:
```
E713 test for membership should be 'not in'
```

and the same for https://github.com/zalando/patroni/pull/2193/files#diff-4ff0e58c1acc3485e8ceb3239cb81a7ad629c6a0c2c8d43f291509658372e645R159

CyberDem0n:
And I think this change breaks

https://github.com/zalando/patroni/blob/0980838cb3df9d4810743deba872d1b719360912/patroni/postgresql/config.py#L638-L664

mbanck:
> And I think this change breaks
> 
> https://github.com/zalando/patroni/blob/0980838cb3df9d4810743deba872d1b719360912/patroni/postgresql/config.py#L638-L664

Hrm, why is that? In my testing, `wal_receiver_primary_conninfo` has `target_session_attrs=any` without this PR and `target_session_attrs=read-write` with it, but I don't see off-hand why that should matter here?

patroni/postgresql/config.py
mbanck: Fixed in 08f8d8d

CyberDem0n:
> Hrm, why is that?

I don't know yet, but all behave tests with Postgres 10+ are failing.

Do we actually need to add `target_session_attrs` to `primary_conninfo`?

mbanck:
> > Hrm, why is that?
> 
> I don't know yet, but all behave tests with Postgres 10+ are failing.

Yeah, I noticed; I couldn't figure out the exact issue yet though, will try harder.

> Do we actually need to add `target_session_attrs` to `primary_conninfo`?

Well, we need to have it somewhere in the replication connection string in order to have the standby leader connect to the main leader (especialy if physical replication slots are in play, but I think it is useful in general?). So `primary_conninfo` in the replication section of the standby leader was the obvious choice for me, do you have another proposal, like injecting it via `PGOPTIONS` or so?

CyberDem0n:
> Well, we need to have it somewhere in the replication connection string in order to have the standby leader connect to the main leader (especialy if physical replication slots are in play, but I think it is useful in general?)

Replication slots would probably be the only reason to have it in the case of a standby leader.
Actually, now I think I understand why behave tests are failing. Patoni also supports cascading replication and it is also covered by tests. If we add `target_session_attrs=read-write` to the primary_conninfo one replica can't stream from another replica anymore. Also it breaks replicas in the standby cluster, because they are always supposed to connect to a read-only node.

mbanck:
> > For all other use-cases, we should leave primary_conninfo alone, do you agree?
> 
> For pg_rewind in the standby cluster we also better use `target_session_attrs=read-write`

Yeah absolutely, that's the first thing I noticed when I got back to this branch.

mbanck:
I think the tests are passing now, the two tests currently failing seem to be flukes

FxKu:
:+1:

CyberDem0n:
:+1:

foger:
@CyberDem0n @mbanck Hello. 

If I need to implement this scheme as experiment:

DC1: Leader, 2 Replicas
              |
              V
DC2: Standby Leader, 2 Replicas
              |
              V
DC3: Standby Leader, 2 Replicas

Cascading with 3 DC, not only 2.

With `target_session_attrs=read-write` automatically added by Patroni on DC3 if node is a Standby Leader and `use_slots: true`, this scheme will not work, because a DC2 Standby Leader is read-only
 
If I set use_slots: false - then it works (target_session_attrs is not appended)

What is a reason of "read-write" usage in case of replication between r/o nodes? Is a pg_rewind will correctly work only with `target_session_attrs=read-write` and `use_slots: true`? Is there a way to use slots in replication from Standby Leader DC2 to Standby Leader DC3, or in Patroni multi-DC replication is possible only in scheme and not more:

DC1: Leader, 2 replicas
              |
              V
DC2: Standby Leader, 2 replicas

? Thanks!

mbanck:
I don't know whether Patroni should consider multi-level standby-clusters; I agree that it currently does not work great.
Patroni adds `target_session_attrs=read-write` so that the standby leader in DC2 can have all members of the cluster in DC1 in its hosts list and does not need to care about which one is the leader in DC1.

foger:
Yes, `target_session_attrs=read-write` and multiple hosts in `primary_conninfo` allows not to use external Haproxy to forward connections to r/w Primary.

I see a benefit from SL to SL replication - maybe network bandwidth between Primary and Standby Leaders + Primary load will be less.

Without `target_session_attrs=read-write`  - target_session_attrs will be "any" by default as in documentation: 

```
target_session_attrs
This option determines whether the session must have certain properties to be acceptable. It's typically used in combination with multiple host names to select the first acceptable alternative among several hosts. There are six modes:

any (default)
any successful connection is acceptable

read-write
session must accept read-write transactions by default (that is, the server must not be in hot standby mode and the default_transaction_read_only parameter must be off)

read-only
session must not accept read-write transactions by default as PostgreSQL documentation

primary
server must not be in hot standby mode

standby
server must be in hot standby mode

prefer-standby
first try to find a standby server, but if none of the listed hosts is a standby server, try again in any mode
```

This will accept Sl to SL replication with slots (with haproxy forwarding only to standby leader of upstream)

May be Patroni as a feature will allow to configure this parameter for flexibility?

Issue #2194, opened 2022-01-30
standby_leader tries to run checkpoint on main cluster leader for rewind


mbanck:
**Describe the bug**

I have a standby cluster (well only the single leader for now), that I restarted after it was down during a primary cluster switchover. During that, Patroni (rightfully) decided that it should run `pg_rewind` from the primary cluster's leader, then subsequently also decided to run `CHECKPOINT` on the primary cluster's leader and failed to find the superuser credentials (I did not put in any).

In any case, I don't think the standby cluster should be in the business of having the primary cluster's superuser credentials, or should those be the same everywhere? I guess rewind credentials are inevitable, if one wants to use that.

I debugged this a bit and what I see is that here:
```
INFO: running pg_rewind from remote_master:b642b724-81c0-11ec-85b1-00163ea51852
```
it still tries to use the `rewind_credentials`, but after this line in `rewind.py`, `checkpoint_status` is `None` (I guess because we are not a leader on startup?):

```
        checkpoint_status = leader.checkpoint_after_promote if isinstance(leader, Leader) else None
```

and it then wants to run checkpoint on the leader with superuser credentials:

```
        if checkpoint_status is None:  # master still runs the old Patroni
            leader_status = self._postgresql.checkpoint(self._conn_kwargs(leader, self._postgresql.config.superuser))
```

**To Reproduce**

1. Have a standby cluster that is configured to run `pg_rewind` and is healthy.
2. Shutdown patroni on the standby cluster
3. Introduce a timeline bump on the primary cluster, e.g. by switching away and back to the primary cluster leader
4. Start patroni on the standby cluster

**Expected behavior**

I would like the standby cluster to come up fine, or worst-case, re-clone from the primary cluster.

**Screenshots**

```
Jan 30 11:34:39 pg4 systemd[1]: patroni@11-main-standby.service: Consumed 1.949s CPU time.
Jan 30 11:35:23 pg4 patroni@11-main-standby[1371]: 2022-01-30 11:35:23,287 INFO: Selected new etcd server http://localhost:2379
Jan 30 11:35:23 pg4 patroni@11-main-standby[1371]: 2022-01-30 11:35:23,293 INFO: No PostgreSQL configuration items changed, nothing to reload.
Jan 30 11:35:23 pg4 patroni@11-main-standby[1371]: 2022-01-30 11:35:23,309 WARNING: Postgresql is not running.
[...]

PR #2195, opened 2022-01-30
Avoid running CHECKPOINT on remote master if credentials are missing

mbanck:
This adds pg_start_backup and pg_stop_backup to the list of functions that are
granted to the rewind user on bootstrap, and uses the rewind credentials to run
those functions rather than `CHECKPOINT` itself in the checkpoint() function.

I wonder whether `checkpoint()` should figure out whether it is being run as a
superuser or not, and still run `CHECKPOINT` directly if it could?

See https://github.com/zalando/patroni/issues/2194#issuecomment-1025138990 for context.

### DIFF
--- a/patroni/postgresql/rewind.py
+++ b/patroni/postgresql/rewind.py
@@ -65,6 +65,20 @@ class Rewind(object):
         except Exception:
             return logger.exception('Exception when working with leader')
 
+    @staticmethod
+    def check_leader_has_run_checkpoint(conn_kwargs):
+        try:
+            with get_connection_cursor(connect_timeout=3, options='-c statement_timeout=2000', **conn_kwargs) as cur:
+                cur.execute("SELECT NOT pg_catalog.pg_is_in_recovery()" +
+                            " AND ('x' || pg_catalog.substr(pg_catalog.pg_walfile_name(" +
+                            " pg_catalog.pg_current_wal_lsn()), 1, 8))::bit(32)::int = timeline_id" +
+                            " FROM pg_catalog.pg_control_checkpoint()")
+                if not cur.fetchone()[0]:
+                    return 'leader has not run a checkpoint yet'
+        except Exception:
+            logger.exception('Exception when working with leader')
+            return 'not accessible or not healty'
+
     def _get_checkpoint_end(self, timeline, lsn):
         """The checkpoint record size in WAL depends on postgres major version and platform (memory alignment).
         Hence, the only reliable way to figure out where it ends, read the record from file with the help of pg_waldump
@@ -337,9 +351,14 @@ class Rewind(object):
         #   running a checkpoint or
         #   waiting until Patroni on the master will expose checkpoint_after_promote=True
         checkpoint_status = leader.checkpoint_after_promote if isinstance(leader, Leader) else None
-        if checkpoint_status is None:  # master still runs the old Patroni
-            leader_status = self._postgresql.checkpoint(self._conn_kwargs(leader, self._postgresql.config.superuser))
-            if leader_status:
+        if checkpoint_status is None:  # we are the standby-cluster leader or master still runs the old Patroni
+            # superuser credentials match rewind_credentials if the latter are not provided or we run 10 or older
+            if self._postgresql.config.superuser == self._postgresql.config.rewind_credentials:
+                leader_status = self._postgresql.checkpoint(
+                        self._conn_kwargs(leader, self._postgresql.config.superuser))
+            else:  # we run 11+ and have a dedicated pg_rewind user
+                leader_status = self.check_leader_has_run_checkpoint(r)
+            if leader_status:  # we tried to run/check for a checkpoint on the remote leader, but it failed
                 return logger.warning('Can not use %s for rewind: %s', leader.name, leader_status)
         elif not checkpoint_status:
             return logger.info('Waiting for checkpoint on %s before rewind', leader.name)
--- a/tests/test_ha.py
+++ b/tests/test_ha.py
@@ -431,7 +431,6 @@ class TestHa(PostgresInit):
         self.assertEqual(self.ha.run_cycle(), 'PAUSE: no action. I am (postgresql0)')
 
     @patch.object(Rewind, 'rewind_or_reinitialize_needed_and_possible', Mock(return_value=True))
-    @patch.object(Rewind, 'check_leader_has_run_checkpoint', Mock(return_value=True))
     @patch.object(Rewind, 'can_rewind', PropertyMock(return_value=True))
     def test_follow_triggers_rewind(self):
         self.p.is_leader = false
@@ -793,7 +792,6 @@ class TestHa(PostgresInit):
         self.assertEqual(self.ha.run_cycle(), 'promoted self to a standby leader by acquiring session lock')
 
     @patch.object(Rewind, 'rewind_or_reinitialize_needed_and_possible', Mock(return_value=True))
-    @patch.object(Rewind, 'check_leader_has_run_checkpoint', Mock(return_value=True))
     @patch.object(Rewind, 'can_rewind', PropertyMock(return_value=True))
     def test_process_unhealthy_standby_cluster_as_cascade_replica(self):
         self.p.is_leader = false
--- a/tests/test_ha.py
+++ b/tests/test_ha.py
@@ -431,6 +431,7 @@ class TestHa(PostgresInit):
         self.assertEqual(self.ha.run_cycle(), 'PAUSE: no action. I am (postgresql0)')
 
     @patch.object(Rewind, 'rewind_or_reinitialize_needed_and_possible', Mock(return_value=True))
+    @patch.object(Rewind, 'check_leader_has_run_checkpoint', Mock(return_value=True))
     @patch.object(Rewind, 'can_rewind', PropertyMock(return_value=True))
     def test_follow_triggers_rewind(self):
         self.p.is_leader = false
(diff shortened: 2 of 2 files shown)


mbanck:
```
=================================== FAILURES ===================================
________________________ TestPostgresql.test_checkpoint ________________________

self = <tests.test_postgresql.TestPostgresql testMethod=test_checkpoint>

    def test_checkpoint(self):
        with patch.object(MockCursor, 'fetchone', Mock(return_value=(True, ))):
            self.assertEqual(self.p.checkpoint({'user': 'postgres'}), 'is_in_recovery=true')
        with patch.object(MockCursor, 'execute', Mock(return_value=None)):
            self.assertIsNone(self.p.checkpoint())
>       self.assertEqual(self.p.checkpoint(timeout=10), 'not accessible or not healty')
E       AssertionError: None != 'not accessible or not healty'

tests/test_postgresql.py:239: AssertionError
```

Not quite sure what this test is checking, is it really checking that non-superuser can't run checkpoints? In that case, should this be removed or reversed?

CyberDem0n:
> Not quite sure what this test is checking

It is more or less to improve code coverage. The `CHECKPOINT` in this case is supposed to raise an exception from here:
https://github.com/zalando/patroni/blob/0980838cb3df9d4810743deba872d1b719360912/tests/__init__.py#L90-L91

Originally pg_rewind was always using with superuser credentials, later, when we added support of dedicated user, I thought about what to do with the CHECKPOINT before pg_rewind (it is required in order to make sure that pg_control is updated with the latest timeline id). At the end, the checkpoint problem was solved by shifting it to the primary. I.e., Patroni on the leader is executing CHECKPOINT right after promote and until it finished the member key in DCS has `checkpoint_after_promote: false`. Other replicas are simply waiting until checkpoint_after_promote disappears from the member key of the leader.

The checkpoint before pg_rewind is only executed by nodes of the standby cluster (or in case if Patroni leader is 1.5.6 or older). To be honest, I don't really like the idea of granting more permissions to the rewind user, especially since this user only works for Postgres 11+. But, in fact I was considering a different approach to do checkpoints. We can abuse a [replication protocol](https://www.postgresql.org/docs/current/protocol-replication.html#PROTOCOL-REPLICATION-BASE-BACKUP) to start/stop backup and hence trigger a checkpoint (`BASE_BACKUP FAST NOWAIT`).

mbanck:
I thought about `BASE BACKUP` as well, but wouldn't that actually start to stream all the changes as well and one would then have to (async?) cancel it? I don't think we want to trigger the whole basebackup just for this.

CyberDem0n:
In python we have more control on the process and can close the connection after first bytes arrived:
```python
import psycopg2

class BaseBackupInterceptor(object):

    def __init__(self, cur):
        self.checkpoint = False
        self._cur = cur

    def write(self, buf):
        self.checkpoint = True
        self._cur.connection.close()


conn = psycopg2.connect(user='standby', host='localhost', replication=1)
conn.autocommit = True
with conn.cursor() as cur:
    b = BaseBackupInterceptor(cur)
    try:
        cur.copy_expert('BASE_BACKUP FAST', b)
    except:
        pass
    print(b.checkpoint)
conn.close()
```

CyberDem0n:
Another approach would be waiting until pg_control is updated.
For that we could use pg_catalog.pg_control_checkpoint() function, which doesn't need superuser.
```sql
SELECT NOT pg_catalog.pg_is_in_recovery()
AND ('x' || pg_catalog.substr(pg_catalog.pg_walfile_name(
  pg_catalog.pg_current_wal_lsn()), 1, 8))::bit(32)::int = timeline_id 
FROM pg_catalog.pg_control_checkpoint();
```

If the above query returns true pg_rewind could be run safely.

mbanck:
> Another approach would be waiting until pg_control is updated. For that we could use pg_catalog.pg_control_checkpoint() function, which doesn't need superuser.
> 
> ```sql
> SELECT NOT pg_catalog.pg_is_in_recovery()
> AND ('x' || pg_catalog.substr(pg_catalog.pg_walfile_name(
>   pg_catalog.pg_current_wal_lsn()), 1, 8))::bit(32)::int = timeline_id 
> FROM pg_catalog.pg_control_checkpoint();
> ```
> 
> If the above query returns true pg_rewind could be run safely.

Won't this introduce a possibly significant delay, if `checkpoint_timeout` is more than a couple of minutes?

By the way, v15 will introduce a `pg_checkpointer` predefined role: 
https://git.postgresql.org/gitweb/?p=postgresql.git;a=commitdiff;h=4168a4745492cd54a0ffffc271b452525ef4dc60

CyberDem0n:
> Won't this introduce a possibly significant delay, if checkpoint_timeout is more than a couple of minutes?

Yes, it might, but if on the other side is Patroni, it runs a checkpoint on the leader ASAP.
Also, we could still try to execute it with superuser credentials and wait only after, but of course, it will increase the code complexity.

P.S. `BASE_BACKUP FAST` is still an option, but it seems that there is no nice way to make it work with psycopg3 :(

mbanck:
I've tried to implement that now in 6db49592ed4429d9900c67116880ce9d9c0e7469

patroni/postgresql/rewind.py
```diff
@@ -334,7 +347,13 @@ def execute(self, leader):
...
         checkpoint_status = leader.checkpoint_after_promote if isinstance(leader, Leader) else None
         if checkpoint_status is None:  # master still runs the old Patroni
-            leader_status = self._postgresql.checkpoint(self._conn_kwargs(leader, self._postgresql.config.superuser))
+            if self._postgresql.config.superuser['password'] is not None:
+                # we have superuser credentials, so run checkpoint on the leader
+                leader_status = self._postgresql.checkpoint(
+                        self._conn_kwargs(leader, self._postgresql.config.superuser))
+            else:
+                # query the leader whether a checkpoint happened using the rewind credentials
+                leader_status = self.check_leader_has_run_checkpoint(r)
```
CyberDem0n, line 356: I think it is a bit more complex:
- With Postgres 10 and older, we don't have any other option rather than relying on superuser access.
- The `postgresql.authentication.superuser` could miss `username` and/or `password` or could be even totally absent from the config, but it doesn't mean that superuser access isn't possible, because libpq is using the local user by default. So the only way to figure out that we have superuser access is to try opening a connection with `self._postgresql.config.superuser` and if it doesn't work - switch to a waiting mode.

mbanck: To your first point, a valid (but painful for all but small clusters) option might be to re-clone; I think this is currently not being done no matter what, right? I think I tried both `remove_data_directory_on_rewind_failure` and `remove_data_directory_on_diverged_timelines` and it still failed not being able to do the Checkpoint. Is this a bug, and if so, is it worth fixing?

To your second point, fair enough. I already had a `have_superuser_credentials` method started, but then realized I don't need it for this PR (so far). Now I think it would make sense and might be of use elsewhere, what do you think?

mbanck:
For now, I decided to just switch things around, i.e. trying to check via rewind_credentials first, and then run a Checkpoint with Superuser credentials (unconditionally again)

patroni/postgresql/rewind.py
```diff
@@ -333,10 +346,13 @@ def execute(self, leader):
...
-            leader_status = self._postgresql.checkpoint(self._conn_kwargs(leader, self._postgresql.config.superuser))
-            if leader_status:
-                return logger.warning('Can not use %s for rewind: %s', leader.name, leader_status)
+        if checkpoint_status is None:  # we are the standby-cluster leader or master still runs the old Patroni
+            timeline_status = self.check_leader_has_run_checkpoint(r)
+            if timeline_status:  # we checked the leader's timeline in pg_control, and it is outdated
+                leader_status = self._postgresql.checkpoint(
+                        self._conn_kwargs(leader, self._postgresql.config.superuser))
+                if leader_status:  # we tried to run a checkpoint on the remote leader, but it failed
+                    return logger.warning('Can not use %s for rewind: %s', leader.name, leader_status)
```
CyberDem0n, line 355: The `check_leader_has_run_checkpoint()` doesn't return anything. Also it has the Postgres version-specific query, i.e. it will fail with 9.6 and older (I know that it is EOL, but the reality shows that many people still run it).

With Postgres 10 and older, we don't even need to call the `check_leader_has_run_checkpoint()`, because the superuser access must be allowed anyway for pg_rewind to work and we can immediately do a CHECKPOINT.

So the code should do something like:
```suggestion
            # The superuser credential match with rewind_credentials if the last one are not provided or we run 10 or older
            if self._postgresql.config.superuser == self._postgresql.config.rewind_credentials:
                leader_status = self._postgresql.checkpoint(
                        self._conn_kwargs(leader, self._postgresql.config.superuser))
            else:  # We run 11+ and have a dedicated pg_rewind user
                leader_status = self.check_leader_has_run_checkpoint(r)

            if leader_status:  # we tried to run a checkpoint on the remote leader, but it failed
                return logger.warning('Can not use %s for rewind: %s', leader.name, leader_status)
```

Of course, the `check_leader_has_run_checkpoint()` should return a string explaining what went wrong.

mbanck: OK, I did that now, many thanks for the suggestions

tests/test_ha.py
```diff
@@ -431,6 +431,7 @@ def test_follow_in_pause(self):
         self.assertEqual(self.ha.run_cycle(), 'PAUSE: no action. I am (postgresql0)')
 
     @patch.object(Rewind, 'rewind_or_reinitialize_needed_and_possible', Mock(return_value=True))
+    @patch.object(Rewind, 'check_leader_has_run_checkpoint', Mock(return_value=True))
```
CyberDem0n, line 434: Please revert changes in the test_ha.py.
With the latest changes in rewind.py they are not necessary.

mbanck: done so now

PR #2208, opened 2022-02-09
Make the code honor the connect_address setting

darix:
Setting connection address seems necessary as psycopg does not handle
IP: SAN entries properly.

The whole code is a bit weird. It first checks for the socket and falls
back to the IP and then does the reverse.

I made the fix minimal without doing the rewrite of the larger logic.

one could argue if connect_address should be parsed and the port taken
from it instead of the extra port. we changed our config to only have
the hostname instead.

### DIFF
--- a/patroni/postgresql/config.py
+++ b/patroni/postgresql/config.py
@@ -883,13 +883,13 @@ class ConfigHandler(object):
         tcp_local_address = self._get_tcp_local_address()
         netloc = self._config.get('connect_address') or tcp_local_address + ':' + port
 
+        tcp_local_address = {'host': netloc, 'port': port}
+
         unix_local_address = {'port': port}
         unix_socket_directories = self._server_parameters.get('unix_socket_directories')
         if unix_socket_directories is not None:
             # fallback to tcp if unix_socket_directories is set, but there are no suitable values
             unix_local_address['host'] = self._get_unix_local_address(unix_socket_directories) or tcp_local_address
-
-        tcp_local_address = {'host': tcp_local_address, 'port': port}
 
         self._local_address = unix_local_address if self._config.get('use_unix_socket') else tcp_local_address
         self.local_replication_address = unix_local_address\


CyberDem0n:
The thing is that the  connect_address isn't always guaranteed to be accessible from the current node.

If the connection is rejected because of ssl certificates you can always switch to unix_socket

darix:
well in our setup it never did fallback to the unix domain socket as the psycopg exception stopped that. I will open a bug for psycopg as well to handle IP based SAN entries properly.

CyberDem0n:
Psycopg2 isn't supposed to fallback, you have configure Patoni to use unix sockets for local connections

darix:
the bug is still there that connect_address is overwritten and never being able to use despite being documented.

yes i could work around using unix_socket as well. but TBH making connect_address work is a good idea as well.
the other route via "take the first listen address" and assuming that we can connect to it might also  not always be true and also needs to be paid attention to when configuring patroni.

CyberDem0n:
I don't get your question.  The only place it is used is for writing it to a member key in dcs.

If the connect_address is missing in the config (backward compatibility with governor) the listen_address is used as a fallback

darix:
see the context of my patch ... there is code that was supposed to use `connect_address` ... but due to how the code was structured that code path is broken.

CyberDem0n:
The code isn't broken, it is working exactly as it is supposed to work.
Your "patch" actually breaks the intended behavior

darix:
so we could remove the line before my change?

CyberDem0n:
Which line?

darix:
https://github.com/zalando/patroni/blob/master/patroni/postgresql/config.py#L884

that line gave us the impression "yep netloc should probably have been used at some point but then down here it will be overriden by the first local address no matter what"

CyberDem0n:
The netloc is used later in the method: https://github.com/zalando/patroni/blob/0980838cb3df9d4810743deba872d1b719360912/patroni/postgresql/config.py#L898

darix:
so why not also use it for the other connections to solve tls problems? :)

CyberDem0n:
We are running in circles. Please re-read my messages, I already explained it.

darix:
which use cases get broken?

CyberDem0n:
One should understand, that `listen_address` and `connect_address` in the Patroni configuration are independent of each other.
This is made for a purpose. Sometimes people put a load-balancer/proxy/connection pooler in front of every postgres node and set the `connect_address` to this middleware.

The middleware could down (even temporary) _independently_ of the Postgres. For remote nodes it is not a big issue, they will retry. But for the Patroni that is supposed to start/stop/promote Postgres on the current node it could be a huge problem. In order to make a decision on the current state and/or change the state we need to have a **reliable** connection to the Postgres we manage. Any kind of middleware seriously impacts such reliability and hence the `localhost`/`unix_socket` is the only choice we have.

While solving your little problem by forcing Patroni connection to the local Postgres via the `connect_address` you also potentially break setups where the `connect_address` doesn't point directly to Postgres.

As I already explained, the following addition to patroni.yaml will also solve the issue you have:
```yaml
postgresql:
  use_unix_socket: true
  use_unix_socket_repl: true
```

Just make sure to also properly configure `pg_hba`.

PR #2210, opened 2022-02-10
Make Kubernetes.cancel_initialization() method similar to other DCS

CyberDem0n:
I.e., do delete unconditionally and return the success


FxKu:
:+1:

CyberDem0n:
:+1:

Issue #2220, opened 2022-02-17
Always removing data directory on manual switchover. Is this something new?


victor-sudakov:
**Describe the bug**

After a manual switchover, the PostgreSQL data directory is removed and cloned from scratch from the new leader. I clearly remember it was not the case with previous versions of Patroni. 

**To Reproduce**

Perform `patronictl switchover` and watch pg_clonecluster_patroni (or pg_basebackup or whatever your create_replica_method is) pulling gigabytes of your data from the new leader to the new replica.

**Expected behavior**

I clearly remember that in previous versions of Patroni, running `patronictl switchover` resulted in a quick and graceful switchover, data was never pulled. The `remove_data_directory_on_diverged_timelines: true` setting has always been there since my first installation of Patroni. Why would timelines be diverging when performing a controlled manual switchover?

**Screenshots**
Please see the attached log.

**Environment**
- Patroni version: 2.1.2-2.pgdg100+1
- PostgreSQL version: 13.6
- DCS (and its version): Consul v1.11.3
- OS: Debian GNU/Linux 10 (buster)

**Patroni configuration file**

[13-vas1.yml.txt](https://github.com/zalando/patroni/files/8087567/13-vas1.yml.txt)

**patronictl show-config**
```
loop_wait: 10
master_start_timeout: 300
maximum_lag_on_failover: 1048576
postgresql:
  parameters:
    max_connections: 960
    max_locks_per_transaction: 256
    max_prepared_transactions: 960
    max_replication_slots: 10
    max_wal_senders: 10
    max_worker_processes: 16
    track_commit_timestamp: 'on'
    wal_keep_size: 16GB
  pg_hba:
  - local   all             all                               trust
  - host    all             all       0.0.0.0/0               md5
  - local   replication   repmgr                              trust
  - host    replication   repmgr      0.0.0.0/0               md5
  remove_data_directory_on_diverged_timelines: true
  remove_data_directory_on_rewind_failure: true
  use_pg_rewind: false
retry_timeout: 10
synchronous_mode: true
synchronous_node_count: 3
ttl: 30

```

**Have you checked Patroni logs?**
[log.txt](https://github.com/zalando/patroni/files/8087527/log.txt)

**Have you tried to use GitHub issue search?**
Found nothing relevant in the open issues. OTOH, I'm not even sure it's a bug. Maybe it's a feature, I'm just wondering why it had a  different behavior in previous versions.

victor-sudakov:
[...]

PR #2221, opened 2022-02-17
Fix bug in divergence timeline check

CyberDem0n:
Patroni was falsely assuming that timelines have diverged.
For pg_rewind it didn't create any problem, but if pg_rewind is not allowed and the `remove_data_directory_on_diverged_timelines` is set, it resulted in reinitializing the former leader.

Close https://github.com/zalando/patroni/issues/2220


CyberDem0n:
:+1:

PR #2224, opened 2022-02-18
Feature/raft logging and metric

hmilkovi:
More logging and metrics to easier monitor and debug.

**Example logs**
```
2022-02-18 09:38:19,890 ERROR: failed to update leader lock
2022-02-18 09:38:19,891 WARNING: raft has lost majority
```

**Example metric**
Also notice lalvel dcs_type as when we have multiple clusters with other dcs we can easier filter them by DCS type
```
# HELP patroni_version Patroni semver without periods.
# TYPE patroni_version gauge
patroni_version{scope="batman",dcs_type="raft"} 020102
...
# HELP patroni_dcs_raft_majority Value is 1 if cluster has raft majority, 0 if not
# TYPE patroni_dcs_raft_majority gauge
patroni_dcs_raft_majority{scope="batman",dcs_type="raft"} 0
...
```

### DIFF
--- a/patroni/api.py
+++ b/patroni/api.py
@@ -218,90 +218,98 @@ class RestApiHandler(BaseHTTPRequestHandler):
         epoch = datetime.datetime(1970, 1, 1, tzinfo=tzutc)
 
         metrics = []
+        dc_type_label = 'dcs_type="{0}"'.format(patroni.dcs.__class__.__name__.lower())
+        scope_label = 'scope="{0}"'.format(patroni.postgresql.scope)
+        all_labels = '{{{0},{1}}}'.format(scope_label, dc_type_label)
 
-        scope_label = '{{scope="{0}"}}'.format(patroni.postgresql.scope)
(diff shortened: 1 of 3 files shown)


CyberDem0n (commented):
It is not nice that DCS abstraction is leaking into Ha and RestApi.

hmilkovi:
When you have time can you please elaborate on what do you mean leak?
As:
we already have abstraction “leak” in https://github.com/zalando/patroni/blob/master/patroni/ha.py#L1395 (other way would be to implement get dcs name in DCS)
 I want to help as much as I can as in our company we have

CyberDem0n:
> we already have abstraction “leak” 

Correct, and I really hate it, but there was no real way around it.

In your case the info is exposed in the API could be easily abstracted. And logging that you put into Ha could be actually done from the raft.py

hmilkovi:
I close this as dcs last seen metrics can be used to calculate when the raft members where online.
Also I'm switching to etcd for stability.
Thx @CyberDem0n for checking :)

PR #2231, opened 2022-02-28
consul: Allow setting TLSServerName on consul service checks

grembo:
See also https://www.consul.io/api-docs/agent/check#tlsservername
Useful in case checks are done by IP and the consul `node_name`
is not an FQDN.

### DIFF
--- a/patroni/dcs/consul.py
+++ b/patroni/dcs/consul.py
@@ -233,7 +233,7 @@ class Consul(AbstractDCS):
         if self._register_service:
             self._set_service_name()
         self._service_check_interval = config.get('service_check_interval', '5s')
-        self._service_check_tls_server_name = config.get('service_check_tls_server_name', "")
+        self._service_check_tls_server_name = config.get('service_check_tls_server_name', None)
         if not self._ctl:
             self.create_session()
 
@@ -459,7 +459,8 @@ class Consul(AbstractDCS):
         conn_parts = urlparse(data['conn_url'])
         check = base.Check.http(api_parts.geturl(), self._service_check_interval,
                                 deregister='{0}s'.format(self._client.http.ttl * 10))
-        check['TLSServerName'] = self._service_check_tls_server_name
+        if self._service_check_tls_server_name is not None:
+            check['TLSServerName'] = self._service_check_tls_server_name
         tags = self._service_tags[:]
         tags.append(role)
         self._previous_loop_service_tags = self._service_tags
--- a/patroni/dcs/consul.py
+++ b/patroni/dcs/consul.py
@@ -233,6 +233,7 @@ class Consul(AbstractDCS):
         if self._register_service:
             self._set_service_name()
         self._service_check_interval = config.get('service_check_interval', '5s')
+        self._service_check_tls_server_name = config.get('service_check_tls_server_name', "")
         if not self._ctl:
             self.create_session()
 
@@ -458,6 +459,7 @@ class Consul(AbstractDCS):
         conn_parts = urlparse(data['conn_url'])
         check = base.Check.http(api_parts.geturl(), self._service_check_interval,
                                 deregister='{0}s'.format(self._client.http.ttl * 10))
+        check['TLSServerName'] = self._service_check_tls_server_name
         tags = self._service_tags[:]
         tags.append(role)
         self._previous_loop_service_tags = self._service_tags
--- a/docs/ENVIRONMENT.rst
+++ b/docs/ENVIRONMENT.rst
@@ -49,6 +49,7 @@ Consul
 -  **PATRONI\_CONSUL\_CHECKS**: (optional) list of Consul health checks used for the session. By default an empty list is used.
 -  **PATRONI\_CONSUL\_REGISTER\_SERVICE**: (optional) whether or not to register a service with the name defined by the scope parameter and the tag master, replica or standby-leader depending on the node's role. Defaults to **false**
 -  **PATRONI\_CONSUL\_SERVICE\_CHECK\_INTERVAL**: (optional) how often to perform health check against registered url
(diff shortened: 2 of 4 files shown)


patroni/dcs/consul.py
```diff
@@ -458,6 +459,7 @@ def _update_service(self, data):
         conn_parts = urlparse(data['conn_url'])
         check = base.Check.http(api_parts.geturl(), self._service_check_interval,
                                 deregister='{0}s'.format(self._client.http.ttl * 10))
+        check['TLSServerName'] = self._service_check_tls_server_name
```
CyberDem0n, line 462: We probably don't want to send it if the value is empty and/or missing from the config.

patroni/dcs/consul.py
```diff
@@ -233,6 +233,7 @@ def __init__(self, config):
         if self._register_service:
             self._set_service_name()
         self._service_check_interval = config.get('service_check_interval', '5s')
+        self._service_check_tls_server_name = config.get('service_check_tls_server_name', "")
```
CyberDem0n, line 236: the better place for it is the `reload_config()` method, so one could change it on the fly.

CyberDem0n (commented):
Please also make it work with environment variables. For that you need to add it to the https://github.com/zalando/patroni/blob/333d41d9f039b5a799940c8a6fbc75dcbe0e9a31/patroni/config.py#L351-L355

And in addition to that you will have to update https://github.com/zalando/patroni/blob/master/docs/ENVIRONMENT.rst

patroni/dcs/consul.py
grembo: > We probably don't want to send it if the value is empty and/or missing from the config.

Consul always sets the value itself anyway, it's an empty string by default (see "example" below), so it doesn't really make a difference (we also use an empty string by default).

If you insist, I could of course add an extra check `if self._service_check_tls_server_name:`, the practical result will be the same.

Example of an existing consul check (configured without specifying TLSServerName):
```json
{
  "Check": {
    "Node": "patroni-node1",
    "CheckID": "service:postgresql/patroni-node1",
    "Name": "Service 'postgresql' check",
    "Status": "critical",
    "Notes": "",
    "Output": "",
    "ServiceID": "postgresql/patroni-node1",
    "ServiceName": "postgresql",
    "ServiceTags": [
      "replica"
    ],
    "Type": "http",
    "Interval": "",
    "Timeout": "",
    "ExposedPort": 0,
    "Definition": {},
    "CreateIndex": 0,
    "ModifyIndex": 0
  },
  "ChkType": {
    "CheckID": "",
    "Name": "",
    "Status": "",
    "Notes": "",
    "ScriptArgs": null,
    "HTTP": "https://10.1.1.1:8008/replica",
    "H2PING": "",
    "Header": null,
    "Method": "",
    "Body": "",
    "TCP": "",
    "Interval": 15000000000,
    "AliasNode": "",
    "AliasService": "",
    "DockerContainerID": "",
    "Shell": "",
    "GRPC": "",
    "GRPCUseTLS": false,
    "TLSServerName": "",
    "TLSSkipVerify": false,
    "Timeout": 0,
    "TTL": 0,
    "SuccessBeforePassing": 0,
    "FailuresBeforeCritical": 0,
    "ProxyHTTP": "",
    "ProxyGRPC": "",
    "DeregisterCriticalServiceAfter": 150000000000,
    "OutputMaxSize": 0
  },
  "Token": "xxx",
  "Source": "remote"
}
```

grembo:
@CyberDem0n I checked the [the reload_config method](https://github.com/zalando/patroni/blob/c824ef1e68951af9b97d1b0936c53ea620d8c2a9/patroni/dcs/consul.py#L251), but I'm not 100% clear about its semantics - does it actually apply changes (e.g., new value for "dc"), and what would that mean for changing the new value I added (and potentially other values like the check interval)?

In an ideal world, I would expect those changes to catch on immediately (so imagine, patroni is running, but health checks are failing, since the tls name doesn't match - I would like to change patroni's config to override TLSServerName and then do a reload, which should change the existing health check in consul). Not sure if the current logic would work like that (only in this case it would make sense to have "hot" reload, otherwise I think it would be more confusing then helpful).

Also, if config reload actually applied the changes to an already running health check, setting TLSServerName to an empty string (what this patch does right now) will actually be important, so one could switch back the configuration setting to default (e.g., set it to a server name in patroni.yml, do a reload, it gets changed in the health check, realize that not what you wanted, remove it from patroni.yml, do another reload, it gets set to an empty string (=unset)).

patroni/dcs/consul.py
CyberDem0n: > Consul always sets the value itself anyway


What about compatibility with older Consul versions? For example, v1.9.x doesn't know about `TLSServerName` and I can expect that such requests will fail.

grembo: @CyberDem0n That's a very good point (wasn't aware that 1.9.x is still a supported release).

patroni/dcs/consul.py
grembo: I tested moving these values to reload_config, changing them has no effect on reload - for that, touch_member would have to be called. I don't know patroni well enough to understand where I would add that and what the consequences would be.

patroni/dcs/consul.py
grembo: Changed it, so that it's only set if configured.

patroni/dcs/consul.py
CyberDem0n: Maybe you are right, supporting reloading `service_check_tls_server_name` on the fly isn't really required.

FxKu:
:+1:

CyberDem0n:
:+1:

CyberDem0n:
Merged, thank you @grembo!

grembo:
@CyberDem0n Thanks for your time spent on reviewing, giving feedback, and merging 👍

Issue #2242, opened 2022-03-14
Old primary was stuck in archive recovery state


yy8848:
**Describe the bug**
The cluster was setup by [Crunchy Data PGO](https://github.com/CrunchyData/postgres-operator).  Due to network partition, node running primary was isolated. Primary lost connection to DCS (K8s API server), repo and replicas. After that node was shutdown. Some time later network partition was fixed and old primary was restarted. But it was stuck in archive recovery state.

**Environment**
patronictl version 2.1.1

**Observation**

1. After network partition happened, primary received immediate shutdown request and was restarted. It entered into standby mode. But due to no connection to repo, the database recovered from local wal logs. 

```sql
2022-03-04 05:13:34.720 UTC [100] LOG:  received immediate shutdown request
...
2022-03-04 05:13:39.933 UTC [30234] LOG:  starting PostgreSQL 13.5 on x86_64-pc-linux-gnu, compiled by gcc (GCC) 8.5.0 20210514 (Red Hat 8.5.0-4), 64-bit
...
2022-03-04 05:13:49.984 UTC [30240] LOG:  entering standby mode
2022-03-04 05:13:49.984 UTC [30240] LOG:  database system was not properly shut down; automatic recovery in progress
2022-03-04 05:13:49.987 UTC [30240] LOG:  redo starts at 0/5014228
2022-03-04 05:13:49.988 UTC [30240] LOG:  invalid record length at 0/501E8B8: wanted 24, got 0
2022-03-04 05:13:49.990 UTC [30240] LOG:  consistent recovery state reached at 0/501E8B8
2022-03-04 05:13:49.991 UTC [30234] LOG:  database system is ready to accept read only connections
```
The mode is sync replication. But unfortunately the last wal log entry is a standby snapshot log.  So MinRecoveryPoint was updated in pg_control, which caused divergence later.
```sql
rmgr: Standby     len (rec/tot):     50/    50, tx:          0, lsn: 0/0501E880, prev 0/0501E850, desc: RUNNING_XACTS nextXid 672 latestCompletedXid 671 oldestRunningXid 672
pg_waldump: fatal: error in WAL record at 0/501E880: invalid record length at 0/501E8B8: wanted 24, got 0
```
In async mode, it leads to divergence in higher probability.

2. Node was shutdown. Old primary database received smart shutdown request. 
```sql
2022-03-04 05:17:25.377 UTC [30234] LOG:  received smart shutdown request
```
But the time was very short and database state was still in archive recovery.
[...]

PR #2252, opened 2022-03-24
Handle rewind when demoted node was shut down

CyberDem0n:
In case of DCS unavailability Patroni restarts Postgres in read-only.
It will cause pg_control to be updated with the `Database cluster state: in archive recovery` and also could set the `MinRecoveryPoint`.

When the next time Patroni is started it will assume that Postgres was running as a replica and rewind isn't required and will try to start the Postgres up. In this situation there is the chance that the start will be aborted with the FATAL error message that looks like `requested timeline 2 does not contain minimum recovery point 0/501E8B8 on timeline 1`.
On the next heart-beat Patroni will again notice that Postgres isn't running which would lead to another start-fail attempt.

This loop is endless.

In order to mitigate the problem we do the following:
1. While figuring out whether the rewind is required we consider `in archive recovery` along with `shut down in recovery`.
2. If pg_rewind is required and the cluster state is `in archive recovery` we also perform recovery in a single-user mode.

Close https://github.com/zalando/patroni/issues/2242

### DIFF
--- a/patroni/ha.py
+++ b/patroni/ha.py
@@ -293,12 +293,25 @@ class Ha(object):
 
         return result
 
+    def _handle_crash_recovery(self):
+        if not self._crash_recovery_executed and (self.cluster.is_unlocked() or self._rewind.can_rewind):
+            self._crash_recovery_executed = True
+            self._crash_recovery_started = time.time()
+            msg = 'doing crash recovery in a single user mode'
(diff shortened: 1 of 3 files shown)


FxKu:
:+1:

CyberDem0n:
:+1:

PR #2276, opened 2022-04-14
Handle the TypeError exception

CyberDem0n:
```
TypeError: '<' not supported between instances of 'NoneType' and 'int'
```

It could happen when the replica didn't start to stream using the replication slot on the primary.

Also, it could be the symptom of a different issue: `hot_standby_feedback` is disabled.

The last one we can check and show the annoying error in the log.


CyberDem0n:
superseded by https://github.com/zalando/patroni/pull/2285

PR #2282, opened 2022-04-27
Add plugin installation step for running tests in CONTRIBUTING.rst

rheaton:
- when compiling postgres from source, this plugin is not automatically
  installed and the features will fail

Co-authored-by: Jacob Champion <pchampion@vmware.com>


CyberDem0n (commented):
I would probably better add following to the step 1:
```
including `contrib <https://www.postgresql.org/docs/current/contrib.html>`__ modules
```

CyberDem0n:
@rheaton are you going to address my comment?

PR #2285, opened 2022-05-02
Enhanced checks of replica logical slots safety

CyberDem0n:
The logical slot on a replica is safe to use when the physical replica
slot on the primary:
1. has a nonzero/non-null `catalog_xmin`
2. has a `catalog_xmin` that is not newer (greater) than the  `catalog_xmin` of any slot on the standby
3. the `catalog_xmin` is known to overtake `catalog_xmin` of logical slots on the primary observed during `1`

In case if `1` doesn't take place, Patroni will run an additional check whether the `hot_standby_feedback` is actually in effect and shows the warning in case it is not.

### DIFF
--- a/patroni/postgresql/slots.py
+++ b/patroni/postgresql/slots.py
@@ -254,9 +254,9 @@ class SlotsHandler(object):
                     catalog_xmin = slots.pop(slot_name)
             except Exception as e:
                 return logger.error("Failed to check %s physical slot on the primary: %r", slot_name, e)
-            # Remember catalog_xmin of logical slots on the primary when catalog_xmin on
-            # physical's slot became valid. Logical slots on replica will be safe to use
-            # after promote when the catalog_xmin of the physical slot overtakes these values.
+            # Remember catalog_xmin of logical slots on the primary when catalog_xmin of
+            # the physical slot became valid. Logical slots on replica will be safe to use after
+            # promote when catalog_xmin of the physical slot overtakes these values.
             if catalog_xmin:
                 for name, value in slots.items():
                     self._unready_logical_slots[name] = value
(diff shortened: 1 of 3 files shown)


patroni/postgresql/slots.py
```diff
@@ -245,17 +245,38 @@ def check_logical_slots_readiness(self, cluster, nofailover, replicatefrom):
...
+                                ([n for n, v in self._unready_logical_slots.items() if v is None] + [slot_name],))
+                    slots = {row[0]: row[1] for row in cur}
+                    if slot_name not in slots:
                         return logger.warning('Physical slot %s does not exist on the primary', slot_name)
-                    catalog_xmin = cur.fetchone()[0]
+                    catalog_xmin = slots.pop(slot_name)
             except Exception as e:
                 return logger.error("Failed to check %s physical slot on the primary: %r", slot_name, e)
+            # Remember catalog_xmin of logical slots on the primary when catalog_xmin on
+            # physical's slot became valid. Logical slots on replica will be safe to use
```
FxKu, line 258:
```suggestion
            # the physical slot became valid. Logical slots on replica will be safe to use
```

patroni/postgresql/slots.py
```diff
@@ -245,17 +245,38 @@ def check_logical_slots_readiness(self, cluster, nofailover, replicatefrom):
...
+            # after promote when the catalog_xmin of the physical slot overtakes these values.
```
FxKu, line 259:
```suggestion
            # Remember catalog_xmin of logical slots on the primary when catalog_xmin of
            # the physical slot became valid. Logical slots on replica will be safe to use after
            # promote when catalog_xmin of the physical slot overtakes these values.
```

patroni/postgresql/slots.py
```diff
@@ -245,17 +245,38 @@ def check_logical_slots_readiness(self, cluster, nofailover, replicatefrom):
...
+            if catalog_xmin:
+                for name, value in slots.items():
+                    self._unready_logical_slots[name] = value
+            else:  # Replica isn't streaming or the hot_standby_feedback isn't enabled
+                try:
+                    cur = self._query("SELECT pg_catalog.current_setting('hot_standby_feedback')::boolean")
+                    if not cur.fetchone()[0]:
+                        return logger.error('Logical slots failover requires "hot_standby_feedback".'
```
FxKu, line 267:
```suggestion
                        return logger.error('Logical slot failover requires "hot_standby_feedback".'
```
I know, there can be multiple slots but for this message I believe singular is also fine

FxKu:
:+1:

CyberDem0n:
:+1:

PR #2287, opened 2022-05-04
k8s: Support refreshing service account tokens

hligit:
Since Kubernetes v1.21, with projected service account token feature,
service account tokens expire in 1 hour. Kubernetes clients are expected
to reread the token file to refresh the token. This patch re-reads the
token file very minute for in-cluster config.

Signed-off-by: Haitao Li <hli@atlassian.com>

### DIFF
--- a/patroni/dcs/kubernetes.py
+++ b/patroni/dcs/kubernetes.py
@@ -48,14 +48,29 @@ class K8sConfig(object):
 
     def __init__(self):
         self.pool_config = {'maxsize': 10, 'num_pools': 10}  # configuration for urllib3.PoolManager
+        self._token_expires_at = datetime.datetime.max
         self._make_headers()
+
+    def _set_token(self, token):
+        self._headers['authorization'] = 'Bearer ' + token
 
     def _make_headers(self, token=None, **kwargs):
         self._headers = urllib3.make_headers(user_agent=USER_AGENT, **kwargs)
         if token:
-            self._headers['authorization'] = 'Bearer ' + token
+            self._set_token(token)
 
-    def load_incluster_config(self, ca_certs=SERVICE_CERT_FILENAME):
+    def _read_token_file(self):
+        if not os.path.isfile(SERVICE_TOKEN_FILENAME):
+            raise self.ConfigException('Service token file does not exists.')
+        with open(SERVICE_TOKEN_FILENAME) as f:
+            token = f.read()
+            if not token:
+                raise self.ConfigException('Token file exists but empty.')
+            self._token_expires_at = datetime.datetime.now() + self._token_refresh_interval
+            return token
+
+    def load_incluster_config(self, ca_certs=SERVICE_CERT_FILENAME,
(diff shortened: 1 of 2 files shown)


patroni/dcs/kubernetes.py
```diff
@@ -109,6 +114,8 @@ def server(self):
 
     @property
     def headers(self):
+        if self._token_expires_at and self._token_expires_at <= datetime.datetime.now():
+            self._read_token_file()
```
CyberDem0n, line 118: In case something is wrong with the token file the `_read_token_file()` will raise an exception.
Not being properly handled, it will cause Patroni to demote Postgres to a read-only.
With 1m refresh interval and 1h token lifetime, the best strategy would be to continue using the old token.

CyberDem0n: Also, when after updating the token we don't need to call `_make_headers()` but just set the new value of `self._headers['authorization']`.

CyberDem0n: Last, but not least, you forgot to set the new expiration time and hence the token file will be read for every API request after 1m passed with the start of Patroni.

CyberDem0n (changes requested):

patroni/dcs/kubernetes.py
hligit: Thanks for the quick review! I've pushed a change addressing these points.

patroni/dcs/kubernetes.py
```diff
@@ -48,14 +48,26 @@ class ConfigException(Exception):
 
     def __init__(self):
         self.pool_config = {'maxsize': 10, 'num_pools': 10}  # configuration for urllib3.PoolManager
+        self._token_expires_at = None
```
CyberDem0n, line 51:
```suggestion
        self._token_expires_at = datetime.datetime.max
```

patroni/dcs/kubernetes.py
```diff
@@ -109,6 +116,12 @@ def server(self):
 
     @property
     def headers(self):
+        if self._token_expires_at and self._token_expires_at <= datetime.datetime.now():
```
CyberDem0n, line 119:
```suggestion
        self._token_expires_at <= datetime.datetime.now():
```

patroni/dcs/kubernetes.py
```diff
@@ -109,6 +116,12 @@ def server(self):
...
+            try:
+                token = self._read_token_file()
+                self._headers['authorization'] = 'Bearer ' + token
```
CyberDem0n, line 122: The same line is used twice, here and https://github.com/zalando/patroni/pull/2287/files#diff-0b47c783038695efd3a30f9797eb9699abc996d28c5ec9f040b81e63e995b068R57
Let's extract it to a separate method with the signature `def _set_token(self, token)`?

CyberDem0n (approved):

sdudoladov:
:+1:

CyberDem0n:
:+1:

Issue #2289, opened 2022-05-06
Replica in crash-loop after Patroni upgrade from 2.0.1 to 2.1.3


le-duane:
**Describe the bug**
After patching a replica from Patroni 2.0.1 to 2.1.3 and restarting, Postgres is getting restarted by Patroni before it has a chance to fully stand up.

Patroni starts Postgres

```
2022-05-06 05:17:14.961 UTC [113] LOG:  database system was shut down in recovery at 2022-05-06 05:16:58 UTC
2022-05-06 05:17:14.961 UTC [114] LOG:  connection received: host=[local]
2022-05-06 05:17:14.962 UTC [114] FATAL:  the database system is starting up
```

This repeats for a while

```
2022-05-06 05:17:37,384 INFO: Lock owner: db-c-0; I am db-b-0
2022-05-06 05:17:37,384 INFO: restarting after failure in progress
2022-05-06 05:17:38.158 UTC [236] LOG:  connection received: host=[local]
2022-05-06 05:17:38.158 UTC [236] FATAL:  the database system is starting up
```

Patroni assumes cached role

```
2022-05-06 05:18:17,783 WARNING: Retry got exception: 'connection problems'
2022-05-06 05:18:17,783 WARNING: Failed to determine PostgreSQL state from the connection, falling back to cached role
```

Patroni restarts Postgres after failed checkpoint

```
2022-05-06 05:18:41,557 INFO: no action. I am (db-b-0), a secondary, and following a leader (db-c-0)
2022-05-06 05:18:41.559 UTC [436] LOG:  connection received: host=[local]
2022-05-06 05:18:41.559 UTC [436] FATAL:  the database system is starting up
2022-05-06 05:18:41,558 WARNING: Loop time exceeded, rescheduling immediately.
2022-05-06 05:18:41,559 ERROR: Exception during CHECKPOINT
Traceback (most recent call last):
  File "/usr/local/lib/python3.6/site-packages/patroni/postgresql/__init__.py", line 600, in checkpoint
    with get_connection_cursor(**connect_kwargs) as cur:
  File "/usr/local/lib/python3.6/contextlib.py", line 81, in __enter__
    return next(self.gen)
  File "/usr/local/lib/python3.6/site-packages/patroni/postgresql/connection.py", line 44, in get_connection_cursor
    conn = psycopg.connect(**kwargs)
  File "/usr/local/lib/python3.6/site-packages/psycopg2/__init__.py", line 122, in connect
[...]

PR #2292, opened 2022-05-09
Special handling of check_recovery_conf for v12+

CyberDem0n:
When starting as a replica it may take some time before Postgres starts accepting new connections, but meanwhile, it could happen that the leader transitioned to a different member and the `primary_conninfo` must be updated.

On pre v12 Patroni regularly checks `recovery.conf` in order to check that recovery parameters match the expectation. Starting from v12 recovery parameters were converted to GUC's and Patroni gets current values from the `pg_settings` view. The last one creates a problem when it takes more than a minute for Postgres to start accepting new connections.

Since Patroni attempts to execute at least `pg_is_in_recovery()` every HA loop, and it is raising at exception, the `check_recovery_conf()` effectively wasn't reachable until recovery is finished, but it changed when #2082 was introduced.

As a result of #2082 we got the following behavior:
1. Up to v12 (not including) everything was working as expected
2. v12 and v13 - Patroni restarting Postgres after 1m of recovery
3. v14+ - the `check_recovery_conf()` is not executed because the `replay_paused()` method raising an exception.

In order to properly handle changes of recovery parameters or leader transitioned to a different node on v12+, we will rely on the cached values of recovery parameters until Postgres becomes ready to execute queries.

Close https://github.com/zalando/patroni/issues/2289

### DIFF
--- a/patroni/ha.py
+++ b/patroni/ha.py
@@ -423,9 +423,10 @@ class Ha(object):
                 self.state_handler.get_history(self._leader_timeline + 1):
             self._rewind.trigger_check_diverged_lsn()
 
-        msg = self._handle_rewind_or_reinitialize()
-        if msg:
-            return msg
+        if not self.state_handler.is_starting():
+            msg = self._handle_rewind_or_reinitialize()
+            if msg:
+                return msg
 
         if not self.is_paused():
             self.state_handler.handle_parameter_change()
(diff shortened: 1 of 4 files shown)


FxKu:
:+1:

CyberDem0n:
:+1:

lalbrecht:
Hello! Can you please tell when and in what release this fix will be included?

PR #2299, opened 2022-05-16
Compatibility with upcoming PostgreSQL 15 beta1

CyberDem0n:
- pg_rewind doesn't like if there are unix sockets in PGDATA
- update postgresql/validator.py


FxKu:
:+1:

CyberDem0n:
:+1:

Issue #2305, opened 2022-05-18
Missing API endpoint for synchronous setups?


Dennis4b:
I am using Patroni in a 2-server synchronous setup. This works well.

I am using `haproxy` to route read-only queries to both database instances, and read-write requests to the primary (using two different pools in the application, connecting to different ports).

This looks (partially) like this in the `haproxy` config:

```
frontend db_ro 
        mode tcp 
        bind 127.0.0.1:5433
 
        default_backend all_db 
 
frontend db_rw 
        mode tcp 
        bind 127.0.0.1:5434 
 
        default_backend only_primary 
 
backend all_db 
        option httpchk OPTIONS /read-only 
        http-check expect status 200 
 
        server db1 192.168.1.1:5432 check port 8008 on-marked-down shutdown-sessions 
        server db2 192.168.1.2:5432 check port 8008 on-marked-down shutdown-sessions 
 
backend only_primary 
        option httpchk OPTIONS /master 
        http-check expect status 200 
 
        server db1 192.168.1.1:5432 check port 8008 on-marked-down shutdown-sessions 
        server db2 192.168.1.2:5432 check port 8008 on-marked-down shutdown-sessions
```

The problem with this setup is that *while* a standby has reconnected and is catching up (but is *not yet* a synchronous standby), the `/read-only` endpoint already returns successfully. The standby is set up as `remote_apply` and it is important that it not return stale data, so it should not be queried until it has fully transitioned back to the `synchronous standby` state.

There is the `/synchronous` endpoint, which is great, but this only returns success on the standby and not on the primary. 

As far as I have found so far, i can only set one `httpchk` method for the `haproxy` `backend`, and not per server. So I don't have a single endpoint that returns `true if primary or synchronous standby` (with `/read-only` being `true if primary or any standby`).

I've looked at the code to add this, but it's a rather long nested `if-then-else` where I'm not familiar with all the details to confidently add a `/read-only-sync` endpoint.

There's probably `haproxy` tricks to be done (and I of course welcome any tips - I'd also like to `leastconn` balance across both `backends` etc), but it's perhaps worth adding such an endpoint as this is probably a common use-case?

CyberDem0n:
[...]

PR #2311, opened 2022-05-20
Add /read-only-sync endpoint (#2305)

Dennis4b:
`/read-only-sync` mirrors `/read-only`, but only returns `200` on a replica if this replica is a synchronous standby.

This can be used to ensure read-only queries are only sent to synchronous replicas (or master), and not yet to sync-replicas-to-be which are in the process of catching up and thus would still return stale data.

### DIFF
--- a/patroni/api.py
+++ b/patroni/api.py
@@ -152,6 +152,11 @@ class RestApiHandler(BaseHTTPRequestHandler):
                 status_code = replica_status_code
             elif path in ('/async', '/asynchronous') and not is_synchronous:
                 status_code = replica_status_code
+            elif path in ('/read-only-sync', '/read-only-synchronous'):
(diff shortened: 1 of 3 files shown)


patroni/api.py
```diff
@@ -152,6 +152,11 @@ def do_GET(self, write_status_code_only=False):
                 status_code = replica_status_code
             elif path in ('/async', '/asynchronous') and not is_synchronous:
                 status_code = replica_status_code
+            elif path in ('/read-only-sync', '/read-only-synchronous'):
+                if primary_status_code == 200:
```
CyberDem0n, line 156: I would suggest to include the standby_leader here:
```suggestion
                if 200 in (primary_status_code, standby_leader_status_code):
```

jopadi:
👍

PR #2318, opened 2022-05-26
Add new static_primary config option

thedodd:
In essence, this configuration option will ensure that a
Patroni cluster running with a static primary will not demote
the master unnecessarily.

Transient failures to update the leader lock in the DCS will not cause a
demotion when running with a static primary. When running as leader
under normal circumstances, DCS exceptions will not cause a demotion
when running with `static_primary=thisNode`.

Even if replicas are added to the Patroni cluster, Patroni will
be able to protect itself from entering into unsafe states by checking
the value of static_primary. If the configured static_primary is not the
host node, then the replica will refuse to progress to postmaster boot.
If the configuration was dynamic, then the replica will shutdown when
such a state is detected.

Added behavioral tests for static primary feature.


CyberDem0n (commented):
Hello @thedodd.

I agree, that the `static_primary` feature is useful and there is a certain demand for it from the community - I've been asked many times about the possibility to run clusters with Patroni, but without automatic failover, because people often want to simplify configuration management.

Right now the feature is very limited, and supports only single-node clusters, but I don't see any problem that it could be extended to support clusters of arbitrary size:
- Node that thinks that is it configured as static_primary should continue running as primary (or promote) even if DCS is down/not accessible or updated_lock()/acquire_lock() calls have failed.
- Other nodes should always use the static_primary instead of whatever is stored in the leader key in DCS.
- We can also support a switchover via Patroni REST API/patronictl. It will simply change the static_primary field in the config.
- If the static_primary value is changed in DCS, the new node should not promote until the old one is shut down cleanly. We can promote though if the old one isn't accessible.
- We need to document that in case of a network spilt it is the responsibility of the user to do STONITH of the isolated old primary. Such a situation could happen for example with Etcd, where we don't use quorum reads - the old static_primary will continue to see the stale data.

.github/workflows/tests.yaml
```diff
@@ -15,6 +15,8 @@ jobs:
       fail-fast: false
       matrix:
         os: [ubuntu, windows, macos]
+    env:
+      PYTHONWARNINGS: ignore
```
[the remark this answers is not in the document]
thedodd: Yea, basically it was causing CI failure (instead of just logging and continuing). Happy to update this, but correcting the issue seemed pertinent for a separate PR.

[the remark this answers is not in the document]
thedodd: Now that I've rebased this PR, I'll go ahead and remove this to see if the dependency issues have perhaps been resolved.

docs/SETTINGS.rst
```diff
@@ -18,7 +18,8 @@ Dynamic configuration is stored in the DCS (Distributed Configuration Store) and
...
+-  **master\_stop\_timeout**: the number of seconds Patroni is allowed to wait when stopping Postgres and effective only when synchronous_mode is enabled. When set to > 0 and the synchronous_mode is enabled, Patroni sends SIGKILL to the postmaster if the stop operation is running for more than the value set by master_stop_timeout. Set the value according to your durability/availability tradeoff. If the parameter is not set or set <= 0, master_stop_timeout does not apply.
+-  **static\_primary**: enables a few optimizations to ensure that a cluster configured with a static primary will not unnecessarily demote the cluster primary. This is useful for cases where a cluster is running as a single-node cluster. When this value is set, replicas will refuse to boot until the config value is removed from DCS config.
```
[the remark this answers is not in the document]
thedodd: > IMO, replicas should run, and replicate only from the node that is declared as a static_primary (or from the member that is configured in tags.replicatefrom).

I think that would be a great "next step" feature to build upon this foundation. Given that all members must sync with the DCS before they can start, and that they sync as part of each control iteration, I think we can build a fairly abstraction around this.

The main reason I push for this being a "next step" is because this feature as it is will address the single-node use case quite well. From there, we could open an issue in this repo to discuss the design and safety considerations for having something like this apply to a cluster of >1 node.

As far as the warning, I will definitely update the docs to include the info you've suggested.

Thoughts?

features/static_primary.feature
```diff
@@ -0,0 +1,20 @@
+Feature: static primary
+  We should check that static primary behavior is safe
+
+  Scenario: check static primary config in dcs blocks replica from starting
+    Given I start postgres0 as static primary
+    Then postgres0 is a leader after 10 seconds
+    And there is a non empty initialize key in DCS after 15 seconds
+    When I issue a PATCH request to http://127.0.0.1:8008/config with {"ttl": 20, "loop_wait": 2, "synchronous_mode": true}
```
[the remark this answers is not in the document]
thedodd: Was mostly just a copy/paste of other test code. I will remove that bit.

patroni/ha.py
```diff
@@ -128,6 +128,19 @@ def is_leader(self):
         with self._is_leader_lock:
             return self._is_leader > time.time()
 
+    def is_static_primary(self):
+        """Check if this node is configured as the static primary of the cluster."""
+        static_primary = self.patroni.config.get('static_primary')
```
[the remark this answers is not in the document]
thedodd: Hmm, well the DCS config would be updated on the next loop, and given the nature of the static_primary config (that it dictates cluster leadership) I think it should only be considered once it has been synced to the DCS. Otherwise there is no guarantee that other members will see the config.

Thoughts?

patroni/ha.py
```diff
@@ -689,6 +702,9 @@ def _is_healthiest_node(self, members, check_replication_lag=True):
     def is_failover_possible(self, members, check_synchronous=True, cluster_lsn=None):
         ret = False
         cluster_timeline = self.cluster.timeline
+        if self.is_static_primary():
+            logger.warning('manual failover: not possible when instance is static primary')
+            return ret
```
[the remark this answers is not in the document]
thedodd: I can agree with that. Will remove.

patroni/ha.py
```diff
@@ -966,7 +982,6 @@ def process_manual_failover_from_leader(self):
 
     def process_unhealthy_cluster(self):
         """Cluster has no leader key"""
-
```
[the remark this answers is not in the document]
thedodd: 🪓'ing

thedodd:
@CyberDem0n thanks for taking the time to review these changes! I appreciate it.

> Right now the feature is very limited, and supports only single-node clusters, but I don't see any problem that it could be extended to support clusters of arbitrary size:

The points that you've enumerated are all solid points, and many of them I agree with. One difficulty that I've run into with this PR so far is that there are quite a few differing opinions on what this should look like, and what would and would not be safe to implement. For example, making any assumptions about cluster roles outside of what the DCS state indicates, seems to be unsafe; however, as you've stated above, as long as all nodes are configured with the correct topology, then things will be fine. 

Others have mentioned that this project's intent overall is to help folks avoid footguns, which is definitely a great standard, and thus have argued against doing things outside of the authority of the DCS. 

All in all, my point is just that there are multiple good ways to implement something like this for Patroni clusters with > 1 member, but they all seem to have this minimal implementation pattern in common which can be used as a building block for perhaps what we could call `static topology` clusters.

My hope is that:
- We can get this PR landed as the foundation of a larger `static topology` feature.
- The current PR will continue to only apply to single-node clusters (replicas will just shutdown when a static primary is detected).
- We open an issue to discuss what the best path forward would be for larger clusters with static topology.

Thoughts?

patroni/ha.py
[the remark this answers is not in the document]
CyberDem0n: > the DCS config would be updated on the next loop

That's exactly my point.
If the value of `static_primary` was changed we need to take action ASAP, not in `loop_wait` seconds.

features/static_primary.feature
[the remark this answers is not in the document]
thedodd: It is possible that removing that setting has actually caused these tests to start failing. Added it back, and testing again.

patroni/ha.py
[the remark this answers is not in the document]
thedodd: The question I am left with in my mind is this: is it safe? IMHO, I am tempted to say that we should ONLY consider the DCS config to be authoritative.

My working mental model is that, for a configuration element like this, it would really only be safe to take the DCS as authoritative. In such a model, the local config is not considered authoritative for this element.

Consider the case where we have a replica which has been added to a cluster, but the initially given local config did not include the `static_primary` element which is present in the DCS. This could become problematic if we treat the local config as authoritative for this element. Instead, if we treat only the DCS as authoritative, then divergent local configuration is forced to conform to the DCS for this element, ensuring that safety considerations are upheld.

I do believe that this is probably why Patroni is currently implemented (irrespective of this PR) to abort startup if the DCS can not be contacted.

Thoughts? Also, thanks for the discussion and consideration.

features/static_primary.feature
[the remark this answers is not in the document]
CyberDem0n: Nope, tests shouldn't fail because of that.
My point is that in a single-node cluster it doesn't make any sense to enable `synchronous_mode`.

patroni/ha.py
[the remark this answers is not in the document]
CyberDem0n: > we should ONLY consider the DCS config to be authoritative.

Yes, in the ideal world we should, but it is not always possible.
This specific feature could be built only by relying on the DCS config cached by Patroni.

> Consider the case where we have a replica which has been added to a cluster, but the initially given local config did not include the static_primary element

Ok, now I see what you don't understand. There is no such "local" config.
Here is what we have:
1. The global config that is stored in DCS (typically in the /config key). Every heart-beat cycle Patroni loads (at least tries) the cluster view by calling https://github.com/zalando/patroni/blob/fb06af9adbf1267dbc0da72c2d8c3eaf802561a6/patroni/ha.py#L135-L141. As a result, the cluster view is cached in the HA, and in the DCS implementation. The last one could be accessed via the `cluster` property https://github.com/zalando/patroni/blob/fb06af9adbf1267dbc0da72c2d8c3eaf802561a6/patroni/dcs/__init__.py#L763-L766.
2. There is also a cached config. It is stored in the `$PGDATA/patroni.dynamic.json`. After every heart-beat cycle it could be updated from https://github.com/zalando/patroni/blob/fb06af9adbf1267dbc0da72c2d8c3eaf802561a6/patroni/__main__.py#L110-L112
3. Last, there is `bootstrap.dcs`, which only works if DCS is empty and there is no `patroni.dynamic.json`.

When Patroni starts, it first loads config from the `patroni.dynamic.json`, and updates it by whatever is stored in DCS. If both are missing, it will use the `bootstrap.dcs` as the last resort: https://github.com/zalando/patroni/blob/fb06af9adbf1267dbc0da72c2d8c3eaf802561a6/patroni/__main__.py#L38-L49

> I do believe that this is probably why Patroni is currently implemented (irrespective of this PR) to abort startup if the DCS can not be contacted.

Exactly, on start we always want to see what is stored in DCS.

> Thoughts?

Coming back to my original comment. The `self.patroni.config` is updated only _after_ the heart-beat loop. The most up-to-date value is available in the `self.cluster.config`. Hence, the `is_static_primary()`  and `is_static_primary_configured()` methods should be implemented similar to the `get_standby_cluster_config()`.

features/static_primary.feature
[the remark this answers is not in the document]
thedodd: Agreed, however it only starts as a single-node cluster. We then add two additional replicas immediately after it starts, and their behavior after startup is important in this case.

patroni/ha.py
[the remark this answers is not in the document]
thedodd: > There is no such "local" config.

Thank you for taking the time to clarify all of that for me. Such being the case, I will go ahead and update this code as you have suggested.

docs/SETTINGS.rst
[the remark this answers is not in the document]
thedodd: Done. We can discuss the higher-level bits on the main PR thread.

patroni/ha.py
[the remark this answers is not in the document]
thedodd: Done.

patroni/ha.py
```diff
@@ -128,6 +128,26 @@ def is_leader(self):
         with self._is_leader_lock:
             return self._is_leader > time.time()
 
+    def is_static_primary(self):
+        """Check if this node is configured as the static primary of the cluster."""
+        if self.cluster and self.cluster.config and self.cluster.config.modify_index:
+            config = self.cluster.config.data
+        else:
+            config = self.patroni.config.dynamic_configuration
+        static_primary = config.get('static_primary')
```
CyberDem0n, line 137: I would suggest to extract this block into a method in order to avoid duplication with https://github.com/zalando/patroni/pull/2318/files#diff-b6a3a0ef64783a6d32feb908d74a3a009e9f25008b8ca5313c9c9747e3b74a63R145-R149

thedodd: Will do.

thedodd: Done.

features/environment.py
```diff
@@ -666,6 +666,13 @@ def start(self, name, max_wait_limit=40, custom_config=None):
                                                       self._output_dir, custom_config)
         self._processes[name].start(max_wait_limit)
 
+    def start_with_expected_failure(self, name, max_wait_limit=40, custom_config=None):
+        try:
+            self.start(name, max_wait_limit, custom_config)
+            assert False, 'expected startup to fail'
+        except:
```
CyberDem0n, line 673: _E722 do not use bare 'except'_

```suggestion
        except Exception:
```

features/steps/static_primary.py
```diff
@@ -0,0 +1,25 @@
+import json
+import patroni.psycopg as pg
```
CyberDem0n, line 2: _F401 'patroni.psycopg as pg' imported but unused_

```suggestion
```

features/steps/static_primary.py
```diff
@@ -0,0 +1,25 @@
...
+
+from behave import step, then
```
CyberDem0n, line 4: _F401 'behave.then' imported but unused_

```suggestion
from behave import step
```

features/steps/static_primary.py
```diff
@@ -0,0 +1,25 @@
...
+from time import sleep, time
```
CyberDem0n, line 5: _F401 'time.time' imported but unused_

```suggestion
from time import sleep
```

features/steps/static_primary.py
```diff
@@ -0,0 +1,25 @@
...
+def start_patroni_as_replica_with_static_primary(context, name, time_limit):
+    return context.pctl.start_with_expected_failure(name, max_wait_limit=time_limit)
+
+
+@step('"{name}" key not in DCS after waiting {time_limit:d} seconds')
+def check_member_not_present(context, name, time_limit):
+    sleep(time_limit)
+    try:
+        json.loads(context.dcs_ctl.query(name))
+        assert False, "found value under DCS key {} after {} seconds".format(name, time_limit)
```
CyberDem0n, line 23: Meh, where is the actual check?

features/static_primary.feature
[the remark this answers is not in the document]
CyberDem0n: I still believe enabling synchronous_mode in a single-node cluster doesn't make sense.

patroni/ha.py
```diff
@@ -1357,6 +1386,14 @@ def _run_cycle(self):
                 self.state_handler.reset_cluster_info_state(None, self.patroni.nofailover)
                 raise
 
+            # If the cluster has been configured with a static primary,
+            # and we are not that primary, then do not proceed.
+            if self.is_static_primary_configured() and not self.is_static_primary():
+                self.shutdown()
+                return 'patroni cluster is configured with a static primary, \
+                        and this node is not the primary, shutting down and \
+                        refusing to start'
```
CyberDem0n, line 1395: `self.shutdown()` will not cause Patroni to shut down, but I believe it should be.
That behavior could be achieved the same way as here: https://github.com/zalando/patroni/blob/d8d634125c547122af22646bd9988686e75adc99/patroni/ha.py#L1448-L1451

features/static_primary.feature
```diff
@@ -0,0 +1,20 @@
...
+    And I start postgres2 with a configured static primary will not boot after 20 seconds
+    And "sync" key not in DCS after waiting 20 seconds
+    And "members/postgres1" key not in DCS after waiting 10 seconds
+    And "members/postgres2" key not in DCS after waiting 10 seconds
+
+  Scenario: check removing static primary config from dcs allows replica startup
+    Given I issue a PATCH request to http://127.0.0.1:8008/config with {"static_primary": null}
+    Then "sync" key in DCS has leader=postgres0 after 20 seconds
+    And "members/postgres1" key in DCS has state=running after 10 seconds
+    And "members/postgres2" key in DCS has state=running after 10 seconds
```
CyberDem0n, line 20: Since postgres1 and postgres2 are not supposed to be running this scenario will fail and should be removed.

thedodd: This scenarios does indeed succeed, because the Patroni instances are still running. It is only the Postgres processes which are not starting due to the static_primary config (which is intended behavior).

Once that static_primary config is set to null in the DCS, then the Postgres replicas are able to start.

I think this is a point of confusion based on one of your recent comments in ha.py. The replica Patroni instances are not intended to sys.exit. The idea is to just keep Postgres from starting, and the Patroni process stays alive and will continue to check the DCS values periodically based on its control loop.

This will allow for a more flexible (and still safe) operator experience. Basically, cluster operators do not need to synchronize the steps of removing static_primary config and adding of new replicas. The operations are safe to overlap (fewer footguns).

patroni/ha.py
thedodd: See my [comment here](https://github.com/zalando/patroni/pull/2318#discussion_r952957565). The intention is not to shutdown Patroni, only to ensure that Postgres itself is not running.

features/environment.py
thedodd: Done.

features/steps/static_primary.py
thedodd: Done.

features/steps/static_primary.py
thedodd: Done.

features/steps/static_primary.py
thedodd: The check is the call to `json.loads`. If it succeeds, then this indicates that we found a member (`name`) and it was not supposed to be there.

CyberDem0n: The thing is that the running Patroni continues to call touch_member() even if Postgres isn't running.
Therefore, the member key never expires.

The `assert` raises the exception which is just eaten on the next line. Or in other words, we wasted `time_limit` just for nothing.

features/steps/static_primary.py
```diff
@@ -0,0 +1,29 @@
...
+    found_value = False
+    try:
+        res = json.loads(context.dcs_ctl.query(name))
+        if res is not None:
+            found_value = True
+    except Exception:
+        pass
+
+    if found_value:
+        assert False, "found value under DCS key {} after {} seconds".format(name, time_limit)
```
CyberDem0n, line 29: 1. Member keys will not expire because they are updated by running Patroni
2. In case of K8s it is even worse, because the info is stored in Pod annotations.

features/steps/static_primary.py
```diff
@@ -0,0 +1,51 @@
...
+        res = json.loads(context.dcs_ctl.query(name))
+        if res is not None:
+            found_value = True
+    except Exception:
+        pass
+
+    if found_value:
+        print("found value under DCS key {}: {}".format(name, res))
+        assert False, "found value under DCS key {} after {} seconds".format(name, time_limit)
+    return True
```
CyberDem0n, line 31: The member key will not expire if Patroni is running.

features/static_primary.feature
```diff
@@ -0,0 +1,21 @@
...
+
+  Scenario: check static primary config in dcs blocks replica from starting
+    Given I start postgres0 as static primary
+    Then postgres0 is a leader after 10 seconds
+    And there is a non empty initialize key in DCS after 15 seconds
+    When I issue a PATCH request to http://127.0.0.1:8008/config with {"ttl": 20, "loop_wait": 2}
+    Then I receive a response code 200
+    When I start postgres1 with a configured static primary will not boot after 20 seconds
+    And I start postgres2 with a configured static primary will not boot after 20 seconds
+    And "sync" key not in DCS after waiting 20 seconds
```
CyberDem0n, line 12:
```suggestion
```

features/static_primary.feature
```diff
@@ -0,0 +1,21 @@
...
+    And "members/postgres1" is stopped and uninitialized after waiting 10 seconds
+    # NOTE: no need to wait an additional 10 seconds here.
+    And "members/postgres2" is stopped and uninitialized after waiting 1 seconds
+
+  Scenario: check removing static primary config from dcs allows replica startup
+    Given I issue a PATCH request to http://127.0.0.1:8008/config with {"static_primary": null}
+    Then "sync" key in DCS has leader=postgres0 after 20 seconds
```
CyberDem0n, line 19:
```suggestion
```

features/steps/static_primary.py
```diff
@@ -0,0 +1,53 @@
...
```
CyberDem0n, line 31:
```suggestion
```

features/steps/static_primary.py
```diff
@@ -0,0 +1,53 @@
...
+
+
+@step('"{name}" is stopped and uninitialized after waiting {time_limit:d} seconds')
+def member_is_stopped_and_uninitialized(context, name, time_limit):
+    sleep(time_limit)
```
CyberDem0n, line 36: It is much better to do it in a loop, like here: https://github.com/zalando/patroni/blob/cea1fa869baafc077d17dd5eb1b60350d09d0592/features/steps/cascading_replication.py#L26-L41

CyberDem0n (changes requested):
I suggest removing everything related to sync replication.
This is not what we test.

hughcapet:
closing in favour of DCS failsafe mode feature

PR #2322, opened 2022-05-31
Release 2.1.4

CyberDem0n:
- bump version
- update release notes
- implement missing unit-tests


jopadi:
👍

docs/releases.rst
```diff
@@ -3,6 +3,78 @@
...
 =============
 
+Version 2.1.4
+-------------
+
+**New features**
+
+- Improve ``pg_rewind`` behavior on typical Debian/Ubuntu systems (Gunnar "Nick" Bluth)
+
+  On Debian/Ubuntu systems it is common to keep Postgres config files outside of the data directory. That makes ``pg_rewind --restore-target-wal`` to fail because it can't find ``postgresql.conf`` when trying to figure out``resolve_command``.
```
sdudoladov, line 13: `make ... fail` works without `to`

docs/releases.rst
```diff
@@ -3,6 +3,78 @@
...
+
+**Stability improvements**
+
+- Don't copy the logical slot if there is a mismatch with the config (Alexander)
+
+  Previously Patroni was copying a logical slot to a replica with restart, noticing that ``plugin`` or ``database`` are different from the config and immediately removing it.
+
+- Special handling of recovery configuration parameters for PostgreSQL v12+ (Alexander)
+
+  While starting as replica Patroni is be able to update ``postgresql.conf`` and restart/reload if the leader address has changed by caching current parameters values instead of querying them from ``pg_settings``.
```
sdudoladov, line 44: > is be able to update ``postgresql.conf``

I don't quite get what is meant here. I guess it is either `need to update` or `can update` 

>  by caching current parameters values instead of querying them from ``pg_settings``.

Is there a period (.) missing ? that part does not make sense to me grammatically: neither update of ``postgresql.conf`` nor leader address change require caching.

docs/releases.rst
```diff
@@ -3,6 +3,78 @@
...
+
+  Tests weren't failing only because it was also imported from other modules.
+
+- Ensure that ``optime`` annotation is a string (Sebastian Hasler)
+
+  In certain cases Patroni was trying to pass it as numeric.
+
+- Better handling of failed ``pg_rewind`` attempt (Alexander)
+
+  If the primary becomes unavailable during ``pg_rewind`` the last one may leave ``$PGDATA`` in a broken state, what was causing removal of the data directory even if it wasn't allowed by configuration
```
sdudoladov, line 67: `the last one` == the rewind attempt ?

docs/releases.rst
```diff
@@ -3,6 +3,78 @@
...
+
+  While starting as replica Patroni is be able to update ``postgresql.conf`` and restart/reload if the leader address has changed by caching current parameters values instead of querying them from ``pg_settings``.
+
+- Better handling of IPv6 addresses in the ``postgresql.listen`` parameters (Alexander)
+
+  Since the ``listen`` parameter has a port people try to put IPv6 addresses into square braces which were not correctly stripped when there is more than one IP in the list.
+
+- Use ``replication`` credentials when performing divergence check only on PostgreSQL v10 and older (Alexander)
+
+  If ``rewind`` is allowed Patroni will use either ``superuser`` or ``rewind`` credentials as before.
```
alexeyklyukin, line 52:
```suggestion
  If ``rewind`` is enabled, Patroni will again use either ``superuser`` or ``rewind`` credentials on newer Postgres versions.
```

docs/releases.rst
alexeyklyukin, line 67:
```suggestion
  If the primary becomes unavailable during ``pg_rewind``, ``$PGDATA`` will be left a broken state. Following that,  Patroni will remove the data directory even if this is not allowed by the configuration.
```

docs/releases.rst
```diff
@@ -3,6 +3,78 @@
...
+
+  If the primary becomes unavailable during ``pg_rewind`` the last one may leave ``$PGDATA`` in a broken state, what was causing removal of the data directory even if it wasn't allowed by configuration
+
+- Don't remove ``slots`` annotations from the leader ``ConfigMap``/``Endpoint`` when PostgreSQL isn't ready (Alexander)
+
+  If ``slots`` value isn't passed the annotation will keep the current value.
+
+- Handle concurrency problem with K8s API watchers (Alexander)
+
+  Under certain (unknown) conditions watchers might become stale, which was causing ``attempt_to_acquire_leader()`` method to fail due to the HTTP status code 409. In this case we reset watchers connections and restart from scratch.
```
alexeyklyukin, line 75:
```suggestion
  Under certain (unknown) conditions watchers might become stale; as a result, ``attempt_to_acquire_leader()`` method could fail due to the HTTP status code 409. In that case we reset watchers connections and restart from scratch.
```

alexeyklyukin (commented):
Looks good, some nits in the comments to the release notes.

docs/releases.rst
CyberDem0n, line 44:
```suggestion
  While starting as replica Patroni should be able to update ``postgresql.conf`` and restart/reload if the leader address has changed by caching current parameters values instead of querying them from ``pg_settings``.
```

docs/releases.rst
```diff
@@ -3,6 +3,78 @@
...
+
+  Previously Patroni was copying a logical slot to a replica with restart, noticing that ``plugin`` or ``database`` are different from the config and immediately removing it.
+
+- Special handling of recovery configuration parameters for PostgreSQL v12+ (Alexander)
+
+  While starting as replica Patroni is be able to update ``postgresql.conf`` and restart/reload if the leader address has changed by caching current parameters values instead of querying them from ``pg_settings``.
+
+- Better handling of IPv6 addresses in the ``postgresql.listen`` parameters (Alexander)
+
+  Since the ``listen`` parameter has a port people try to put IPv6 addresses into square braces which were not correctly stripped when there is more than one IP in the list.
```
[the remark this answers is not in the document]
CyberDem0n: > Why does the presence of the port in the listen parameter suggest putting an IPv6 address into square brackets?

Because this is what people typically do when they want to make it clear that the last item isn't a hextet but port: `::1:8080` vs `[::1]:8080`.

docs/releases.rst
```diff
@@ -3,6 +3,78 @@
...
+- Added ``/read-only-sync`` monitoring endpoint (Dennis4b)
+
+  It is similar to the ``/read-only`` but includes only synchronous replicas.
+
+
+**Stability improvements**
+
+- Don't copy the logical slot if there is a mismatch with the config (Alexander)
+
+  Previously Patroni was copying a logical slot to a replica with restart, noticing that ``plugin`` or ``database`` are different from the config and immediately removing it.
```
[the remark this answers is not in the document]
CyberDem0n: If someone does `patronictl edit-config` and changes for example the `database` for an existing logical slot the following happens:
- primary will notice that the existing slot is created in the "wrong" database and try to remove it. If the slot is active `pg_drop_replication_slot()` will fail.
- replica will notice that the existing slot is created in the "wrong" database and removes it. After that it connects to the primary, gets the replication slot file, stops postgres, writes the slot file to PGDATA, and starts postgres. But, since the slot on the primary exists in the "wrong" database, slot on the replica is also wrong, and the action repeates.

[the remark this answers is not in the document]
alexeyklyukin: Thanks @CyberDem0n , here is my attempt to rephrase it:
```suggestion
  A replica won't copy a logical replication slot from the primary anymore if the slot doesn't match the ``plugin`` or ``database`` configuration options. Previously, the check for whether the slot matches those configuration options was not performed until after the replica copied the slot and started with it, resulting in unnecessary and repeated restarts.
```

sdudoladov:
:+1:

docs/releases.rst
```diff
@@ -3,6 +3,78 @@
...
+
+  Tests weren't failing only because it was also imported from other modules.
+
+- Ensure that ``optime`` annotation is a string (Sebastian Hasler)
+
+  In certain cases Patroni was trying to pass it as numeric.
+
+- Better handling of failed ``pg_rewind`` attempt (Alexander)
+
+  If the primary becomes unavailable during ``pg_rewind``, ``$PGDATA`` will be left a broken state. Following that,  Patroni will remove the data directory even if this is not allowed by the configuration.
```
CyberDem0n, line 67:
```suggestion
  If the primary becomes unavailable during ``pg_rewind``, ``$PGDATA`` will be left in a broken state. Following that,  Patroni will remove the data directory even if this is not allowed by the configuration.
```

CyberDem0n:
:+1:

Issue #2331, opened 2022-06-12
Patroni failed to grant lease when system recover from OOM


monsterxx03:
**Describe the bug**
A three node patroni cluster, using etcdv3 (3.5.4) as dcs.

When a follower node OOM (it's not patroni's problem, QA's doing some fault injection testing). After it came back, patroni reporting error in loop:

```
2022-06-09 09:08:07,376 ERROR: refresh_lease
Traceback (most recent call last):
  File "/usr/local/lib/python3.9/dist-packages/patroni/dcs/etcd3.py", line 617, in refresh_lease
    return self.retry(self._do_refresh_lease)
  File "/usr/local/lib/python3.9/dist-packages/patroni/dcs/etcd.py", line 447, in retry
    return retry(*args, **kwargs)
  File "/usr/local/lib/python3.9/dist-packages/patroni/utils.py", line 334, in __call__
    return func(*args, **kwargs)
  File "/usr/local/lib/python3.9/dist-packages/patroni/dcs/etcd3.py", line 610, in _do_refresh_lease
    self._lease = self._client.lease_grant(self._ttl, retry)
  File "/usr/local/lib/python3.9/dist-packages/patroni/dcs/etcd3.py", line 319, in lease_grant
    return self.call_rpc('/lease/grant', {'TTL': ttl}, retry)['ID']
  File "/usr/local/lib/python3.9/dist-packages/patroni/dcs/etcd3.py", line 558, in call_rpc
    ret = super(PatroniEtcd3Client, self).call_rpc(method, fields, retry)
  File "/usr/local/lib/python3.9/dist-packages/patroni/dcs/etcd3.py", line 262, in call_rpc
    return self.api_execute(self.version_prefix + method, self._MPOST, fields)
  File "/usr/local/lib/python3.9/dist-packages/patroni/dcs/etcd.py", line 257, in api_execute
    return self._handle_server_response(response)
  File "/usr/local/lib/python3.9/dist-packages/patroni/dcs/etcd3.py", line 220, in _handle_server_response
    _raise_for_data(data, response.status)
  File "/usr/local/lib/python3.9/dist-packages/patroni/dcs/etcd3.py", line 136, in _raise_for_data
    raise err(code, error, status_code)
patroni.dcs.etcd3.InvalidAuthToken: <InvalidAuthToken error: 'etcdserver: invalid auth token', code: 16>
[...]

PR #2332, opened 2022-06-12
Handle expired token for etcd lease_grant (#2331)

monsterxx03:
If pass an expired token for lease_grant, it will
report `invalid auth token` error.

For lease_keepalive, etcd server don't check token
header at all.


sdudoladov:
:+1:

CyberDem0n:
:+1:

CyberDem0n:
Merged, thank you @monsterxx03!

deefha:
Hi, I would like to ask when this fix will be released. Looking forward to 2.1.5 maybe? :-)

PR #2333, opened 2022-06-12
Bump Postgres version in Dockerfile to 14

lvjonok:
I tried to use your tool and encountered error, when tried to build Dockerfile

Problem seemed to be in missing dependency on setuptools. After adding it, everything built properly


CyberDem0n (changes requested):
Sorry, but no. Just bump `PG_MAJOR` to 14 at the top of the Dockerfile and everything will work.
Version 10 is nearly 5 years old and will become EOL in November 2022.

lvjonok:
Thank you, it worked!

Anyway, I updated default argument of ```PG_MAJOR```, so next people will not have this problem again.

sdudoladov:
:+1:

CyberDem0n:
:+1:

CyberDem0n (approved):

PR #2344, opened 2022-06-17
Improvements to 'patroni --validate-config'

### LINKED ISSUE #2345: 'patroni --validate-config' exists 0 upon invalid configuration
`patroni --validate-config` exists 0 upon invalid configuration:
```
$ patroni --validate-config postgres0.yml
restapi.connect_address 127.0.0.1:8008 didn't pass validation: 'must not contain "127.0.0.1", "0.0.0.0", "*", "::1", "localhost"'
etcd.host 127.0.0.1:2379 didn't pass validation: '127.0.0.1:2379 is not reachable'
postgresql.connect_address 127.0.0.1:5432 didn't pass validation: 'must not contain "127.0.0.1", "0.0.0.0", "*", "::1", "localhost"'
$ echo $?
0
```
This is misleading since one cannot rely on exit code to detect an error; also errors are printed to stdout whereas stderr would be more appropriate.
I'd suggest to:
* exit 1 upon invalid configuration
* print errors to stderr

and I'm willing to prepare a PR for this.

**Environment**
- Patroni version: master at b8a6387236d619345c2f72f2f8e858dc7a2bd791
- PostgreSQL version: n/a
- DCS (and its version): n/a

dlax:
Fix #2345

### DIFF
--- a/patroni/daemon.py
+++ b/patroni/daemon.py
@@ -1,3 +1,5 @@
+from __future__ import print_function
+
 import abc
 import os
 import signal
@@ -92,7 +94,7 @@ def abstract_main(cls, validator=None):
         config = Config(args.configfile)
     except ConfigParseError as e:
         if e.value:
-            print(e.value)
+            print(e.value, file=sys.stderr)
         if not validate_config:
             parser.print_help()
         sys.exit(1)
--- a/patroni/daemon.py
+++ b/patroni/daemon.py
@@ -83,8 +83,9 @@ def abstract_main(cls, validator=None):
                         help='Patroni may also read the configuration from the {0} environment variable'
                         .format(Config.PATRONI_CONFIG_VARIABLE))
     args = parser.parse_args()
+    validate_config = validator and args.validate_config
     try:
-        if validator and args.validate_config:
+        if validate_config:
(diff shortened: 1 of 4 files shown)


patroni/daemon.py
```diff
@@ -83,16 +83,18 @@ def abstract_main(cls, validator=None):
...
+        if validate_config:
             Config(args.configfile, validator=validator)
             sys.exit()
 
         config = Config(args.configfile)
     except ConfigParseError as e:
         if e.value:
-            print(e.value)
-        parser.print_help()
+            print(e.value, file=sys.stderr)
```
CyberDem0n, line 95:
```suggestion
            sys.exit(e.value)
```

I think tests with python 2.7 are failing because of that.

dlax: With the `sys.exit()` here, we'd no longer display help; is this okay?
Instead, I have used `from __future__ import print_function` to work around python2 hopefully.

CyberDem0n: It also works.

CyberDem0n:
:+1:

CyberDem0n:
Merged, thank you @dlax!

PR #2349, opened 2022-06-22
Always return checkpoint location as integer

CyberDem0n:
before it was also returned as str in some cases


sdudoladov:
:+1:

CyberDem0n:
:+1:

PR #2358, opened 2022-07-13
Automatically skip some behave tests on legacy Postgres

CyberDem0n:
previously behave had to be started with `--tags=-skip` argument.

### DIFF
--- a/features/standby_cluster.feature
+++ b/features/standby_cluster.feature
@@ -14,7 +14,7 @@ Feature: standby cluster
     Then "members/postgres0" key in DCS has state=running after 10 seconds
     And replication works from postgres1 to postgres0 after 15 seconds
 
-  @skip
+  @skip-legacy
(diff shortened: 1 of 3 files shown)


features/standby_cluster.feature
```diff
@@ -14,7 +14,7 @@ Feature: standby cluster
     Then "members/postgres0" key in DCS has state=running after 10 seconds
     And replication works from postgres1 to postgres0 after 15 seconds
 
-  @skip
```
hughcapet, line 17: Maybe we better rename it to smth like `@skip-no-slot-advance`? In case we will need to skip another subset of test because of some other newly implemented features in the future

hughcapet: The error msg 'Unsupported PostgreSQL version' is also seems to be misleading for me

hughcapet:
👍

CyberDem0n:
:+1:

PR #2371, opened 2022-07-29
Make backup copies of configuration files in config_dir

### LINKED ISSUE #2370: Brittle configuration file backup/restore for external config_dir
**Describe the bug**

1. Have a primary with `config_dir` outside the data directory
2. Patroni (on startup/restart e.g.) copies `config_dir/postgresql.base.conf` to `PGDATA/postgresql.base.conf.backup`
3. On standby bootstrap, basebackup method streams `postgresql.base.conf.backup` into standby's `PGDATA` along with the rest
4. Patroni restores `PGDATA/postgresql.base.conf.backup` to `config_dir/postgresql.base.conf`, overwriting the original one (if there was one)
5. Postgres potentially does not start, in case the new version of `postgresql.base.conf` has wrong data for the local node (e.g. wrong `data_directory`)

**Expected behavior**

I think Patroni should not overwrite local configuration files with a backup from the primary, but I'm not exactly what happens in the general case. One possibility would be to store the `.backup` files in `config_dir` as well and then restore them from there (thus avoiding the `.backup` files in PGDATA), but not sure whether that would break on update or something?

A work-around is to make sure `PGDATA/postgresql.base.conf.backup` does not exits on the primary before bootstrapping the standby.

Well, or Patroni needs to be more explicit that `postgresql.base.conf` must be identical on all nodes, which I am not sure is a reasonable requirement seeing how people put e.g. local IP addresses or hostnames into things like `syslog_ident` etc.

**Environment**
- Patroni version: 2.1.4
- PostgreSQL version: 13.7
- DCS (and its version): etcd3

mbanck:
Before, the .backup files were always stored in the data directory. However, if
there is a distinct configuration directory, it seems logical to put them
there. If config_dir is not configured, it defaults back to the data directory,
so this will not change things for setups where postgresql.conf is in the data
directory.

On the other hand, it will help with setups where postgresql.conf is outside
the data directory and has local changes. In this case, those no longer get
overwritten on boostrap/clone from the primary's backup configuration file that
is streamed to the boostrapped node.

Close #2370

### DIFF
--- a/patroni/postgresql/config.py
+++ b/patroni/postgresql/config.py
@@ -362,7 +362,7 @@ class ConfigHandler(object):
             try:
                 for f in self._configuration_to_save:
                     config_file = os.path.join(self._config_dir, f)
-                    backup_file = os.path.join(self._postgresql.data_dir, f + '.backup')
+                    backup_file = os.path.join(self._config_dir, f + '.backup')
                     if os.path.isfile(config_file):
                         shutil.copy(config_file, backup_file)
             except IOError:
@@ -374,7 +374,7 @@ class ConfigHandler(object):
         try:
             for f in self._configuration_to_save:
                 config_file = os.path.join(self._config_dir, f)
-                backup_file = os.path.join(self._postgresql.data_dir, f + '.backup')
+                backup_file = os.path.join(self._config_dir, f + '.backup')
                 if not os.path.isfile(config_file):
                     if os.path.isfile(backup_file):
                         shutil.copy(backup_file, config_file)


CyberDem0n:
The whole point of creating backups of configuration files is that configs are not actually backed up by some backup tools (for example wal-e). After restoring such a backup we copy these files back.

If config files are stored in the external place things get complex (as usual). It is absolutely normal that the config directory is empty during the bootstrap or replica creation and we want to copy backup files over (current behavior). And your patch will break it.

mbanck:
> The whole point of creating backups of configuration files is that configs are not actually backed up by some backup tools (for example wal-e). After restoring such a backup we copy these files back.
> 
> If config files are stored in the external place things get complex (as usual). It is absolutely normal that the config directory is empty during the bootstrap or replica creation and we want to copy backup files over (current behavior). 

So what you actually want is to (optionally, or maybe by default, but configurable) copy over the primary's `postgresql.base.conf` during standby clone if it is not in the data directory, is that correct? And in order to do that, we copy it into the data directory, and let pg_basebackup/the clone method handle it.

> And your patch will break it.

Well fair enough, but then Patroni is currently breaking if there's already a local (and different) configuration file. Maybe it should not overwrite existing configuration files? Right now, if there is any change between primary and standby, those get overwritten, resulting in configuration loss.

Would it be ok to make this configurable, i.e. let Patroni know whether we want the `.backup` files in the data directory or not?

PR #2372, opened 2022-07-31
systemd service options: restart patroni service if it crashed

NikolayS:
It makes little sense to have ` for the Patroni service – if it goes down and we don't notice it (I doubt that patroni service uptime is well monitored in most cases), then we lose autofailover for Postgres, implying bigger risks of downtime.

`Restart=on-failure` makes more sense in this case.


idanovinda:
👍

CyberDem0n:
:+1:

PR #2379, opened 2022-08-10
DCS failsafe mode

CyberDem0n:
If enabled it will allow Patroni to cope with DCS outages.
In case of a DCS outage the leader tries to call all remaining members in the cluster via API and if all of them respond with success the leader will not be demoted.

The failsafe_mode could be enabled by running
```sh
patronictl edit-config -s failsafe_mode=true
```

or by calling the `/config` REST API endpoint.


CyberDem0n:
@thedodd yes and no.
This implementation is trying to make a difference between the reasons why the update of the leader lock failed and the failsafe mechanism triggers only when we got a network communication exception or some sort of InternalError exception from DCS.
Right now it is hard to guarantee that all possible corner cases are covered.

The #2318 simply considers any failed attempt to update the leader lock as a failsafe trigger and should not bring many surprises. Also, it is a very common request to have a Patroni just to handle configuration management (yeah, I know, it defeats the purpose of Patroni :) )

thedodd (approved):
+1

Nice threading construct to make the peer communication concurrent. I will be pulling this code to do some testing of this feature within a K8s development cluster quite soon. I'll circle back if I run into any issues.

thedodd:
@CyberDem0n just some quick feedback. I've created a docker image to test these changes, and my setup is as follows:
- All code is present as of this branch's latest commit `772fe19`.
- I've configured a 2 member cluster, with failsafe_mode=true, running in a K8s cluster.
- When I query their /failsafe endpoint, they return:
  ```
    "member0": "http://10.22.27.41:8008/patroni",
    "member1": "http://10.22.10.123:8008/patroni"
  ```
- Cluster is stable and has a master & a replica.

If I then delete the K8s RoleBinding which grants Patroni access to the K8s API (for updating endpoints/services/labels &c), it looks like the primary is still demoting.

Technically, the K8s API will have returned a 403 (which I see in the Patroni logs). Does this circumvent the failsafe logic? Upon reviewing the code, it is not immediately obvious to me.

CyberDem0n:
> If I then delete the K8s RoleBinding which grants Patroni access to the K8s API (for updating endpoints/services/labels &c), it looks like the primary is still demoting.
Technically, the K8s API will have returned a 403 (which I see in the Patroni logs). Does this circumvent the failsafe logic? Upon reviewing the code, it is not immediately obvious to me.

Hi @thedodd. The failsafe logic is triggered when DCS doesn't respond in time or when it is asking to repeat the request after some timeout. But, when retrying there could still be the same issue.
In both cases Patroni is giving up after `retry_timeout` seconds and the failsafe mode is triggered. The 403 status code only indicates that DCS rejects requests from the current node, but in fact, it is healthy. I can't immediately tell whether it would be safe to use 403 in order to trigger the failsafe mechanism.

CyberDem0n:
@alexeyklyukin yeah, nodes must be able to communicate with each other using credentials and/or client certificates.
Sure, we can call the `POST /failsafe` in order to make sure that authentication succeeds, but there are still many questions:
1. Should we completely ignore the failsafe mode in this case or just don't add the node to the list?
2. On the next iteration the check will be executed again. It will not only spam logs, but cause some increased resource usage.
3. What if the check succeeds initially but starts failing later? For example, someone changed credentials on the other node (it is not easy to detect). Or maybe some certificates expired.

thedodd:
@CyberDem0n ok, quick update: by way of manually crippling the K8s API server pods of the testing cluster, I was able to induce the needed downtime scenario. Failsafe mode appears to have held the Patroni cluster steady (1 master, 1 replica). I could connect, reconnect, and continue to use the master as expected. No cluster topology changes during the control plane downtime.

anyafit:
hi, i try use this PR and get error in communicating with DCS, but i can't see that this feature is work

CyberDem0n:
> What is my mistake?

@anyafit for existing clusters one should use `patronictl edit-config` to enable it: https://github.com/zalando/patroni/blob/e54f736651743a2521bdf393f067ea191ae08e19/docs/dcs_failsafe_mode.rst
You can also check whether the feature is enabled using `patronictl show-config`.

CyberDem0n:
> by way of manually crippling the K8s API server pods of the testing cluster, I was able to induce the needed downtime scenario. Failsafe mode appears to have held the Patroni cluster steady

@thedodd about simple cases I am quite confident that it works. We have behave tests for that, that send SIGSTOP to the localkube process in order to simulate outage.

thedodd:
@CyberDem0n thoughts on when you would like to land this PR? Waiting on a few other reviews?

CyberDem0n:
@thedodd it should be more sophisticated than just stopping the control plane.
For example:
- slow network
- overloaded control-plane
- leader not being able to access the control plane, with replicas not having this problem

OlleLarsson:
Hi @CyberDem0n, any updates on this feature 🤗 ?

CyberDem0n:
@OlleLarsson more people will test it - faster it will be merged and released. Feel free to participate :)

OlleLarsson:
> @OlleLarsson more people will test it - faster it will be merged and released. Feel free to participate :)

Perhaps I can find some time.
Is there are specific scenario that's extra interesting to test?

CyberDem0n:
@OlleLarsson ideally it should be something close to real problems with DCS.

anyafit:
hi, i have some problem with connect to etcd. It's mean that problem solved too fast?```
2022-10-27 20:18:44,385 INFO: Lock owner: patroni_hostname; I am patroni_hostname
2022-10-27 20:18:47,724 ERROR: Request to server http://etcd_hostname:2379 failed: ReadTimeoutError("HTTPConnectionPool(host='etcd_hostname', port=2379): Read timed out. (read timeout=3.333218111967047)",)
2022-10-27 20:18:47,724 INFO: Reconnection allowed, looking for another server.
2022-10-27 20:18:47,724 INFO: Retrying on http://etcd_hostname_2:2379
2022-10-27 20:18:47,907 INFO: Selected new etcd server http://etcd_hostname_2:2379
2022-10-27 20:18:47,908 ERROR: watchprefix failed: ProtocolError('Connection broken: IncompleteRead(0 bytes read)', IncompleteRead(0 bytes read))
2022-10-27 20:18:48,401 INFO: no action. I am (patroni_hostname), the leader with the lock
2022-10-27 20:18:49,577 ERROR: Request to server http://etcd_hostname_2:2379 failed: MaxRetryError("HTTPConnectionPool(host='etcd_hostname_2', port=2379): Max retries exceeded with url: /v3/kv/range (Caused by ConnectTimeoutError(<urllib3.connection.HTTPConnection object at 0x7f48d2f91c50>, 'Connection to etcd_hostname_2 timed out. (connect timeout=1.6666666666666667)'))",)
2022-10-27 20:18:49,577 INFO: Reconnection allowed, looking for another server.
2022-10-27 20:18:49,577 INFO: Retrying on http://etcd_hostname:2379
2022-10-27 20:18:49,755 INFO: Selected new etcd server http://etcd_hostname:2379
2022-10-27 20:18:55,924 INFO: no action. I am (patroni_hostname), the leader with the lock

```

File: features/dcs_failsafe_mode.feature [hunk 1]
```diff
@@ -19,8 +19,8 @@ Feature: dcs failsafe mode
   @dcs-failsafe
   Scenario: check one-node cluster is functioning while DCS is down
     Given DCS is down
-    When I sleep for 12 seconds
-    Then postgres0 role is the primary after 10 seconds
+    Then Response on GET http://127.0.0.1:8008/primary contains failsafe_mode_is_active after 12 seconds
```
CyberDem0n, line 22: There are couple of more places where the similar trick could be done, but unfortunately it doesn't work with Raft because 2+ Patroni nodes are enough for quorum when `patroni_raft_controller` is paused.

patroni/ha.py
```diff
@@ -114,6 +122,7 @@ def __init__(self, patroni):
         self._is_leader = False
         self._is_leader_lock = RLock()
         self._failsafe = Failsafe(patroni.dcs)
+        self._failsafe_is_active = False
```
CyberDem0n, line 125: Should be removed.

patroni/api.py
```diff
@@ -302,6 +302,11 @@ def do_GET_metrics(self):
         metrics.append("# TYPE patroni_cluster_unlocked gauge")
         metrics.append("patroni_cluster_unlocked{0} {1}".format(scope_label, int(postgres.get('cluster_unlocked', 0))))
 
+        metrics.append("# HELP patroni_failsafe_mode_is_active Value is 1 if the cluster is unlocked, 0 if locked.")
```
hughcapet, line 305: off-topic: seems `/metrics` endpoint is not documented

patroni/ha.py
```diff
@@ -101,6 +101,14 @@ def update_cluster(self, cluster):
             cluster = Cluster(*cluster)
         return cluster
 
+    def is_active(self):
+        with self._lock:
```
hughcapet, line 105: ~one day this lock will have a more informative name~

patroni/dcs/__init__.py
```diff
@@ -444,7 +444,7 @@ def from_node(index, value):
         return TimelineHistory(index, value, lines)
 
 
-class Cluster(namedtuple('Cluster', 'initialize,config,leader,last_lsn,members,failover,sync,history,slots')):
+class Cluster(namedtuple('Cluster', 'initialize,config,leader,last_lsn,members,failover,sync,history,slots,failsafe')):
```
hughcapet, line 447: failsafe field's description is omitted in the docstring

hughcapet:
👍

CyberDem0n:
:+1:

Issue #2314, opened 2022-05-24
[Question] patroni drops replication slots while in maintenance mode


1exx:
Hello.
Why Patroni drops replication slots of the stopped standby node (PostgreSQL & patroni service) when cluster was in maintenance mode (pause)? 
Is this expected behavior?

I supposed that in maintenance mode patroni will not be change postgres configuration
I'm using Patroni 2.1.2

CyberDem0n:
It was always like that. To be honest, I don't know what would be better, to keep them or remove.

Dsolik:
In what version can this appear?

CyberDem0n:
@Dsolik the next one after it is implemented and committed to the master branch.

PR #2383, opened 2022-08-15
Don't drop replication slots in pause

CyberDem0n:
If replication slots are enabled Patroni automatically creates them for any cluster member that is supposed to stream from a given node and for any permanent slot defined in the global configuration. If the member disappears from the DCS Patroni automatically removes the replication slot for it. The same behavior was in the maintenance mode (pause).

This commit disables removal of any replication slots that don't match Patroni's expectations in pause.

Close https://github.com/zalando/patroni/issues/2314


hughcapet:
👍

CyberDem0n:
:+1:

PR #2384, opened 2022-08-15
Archive possibly missing WALs before rewind

hughcapet:
There is currently a risk to lose some WAL segments entirely in case
archive_mode was set to 'on' before a promote and there are some WALs
with .ready files on the former leader we are trying to rewind. It happens
because of the pg_rewind's modus operandi: it simply syncs the content of
pg_wal directory of the old leader with the new leader's one. Including
deletion of all WALs that are not present on the current leader, regardless
their archive status on the former one. Thus, in case the new leader has
already recycled such files, we just remove them entirely.

In case archive_mode was set to 'always' and the .ready WALs are actually
present in archive, it is for end user who writes the archive_command to
avoid overwriting and to properly test it. 
According to [pg docs](https://www.postgresql.org/docs/current/continuous-archiving.html):   

> The archive command should generally be designed to refuse to  overwrite any pre-existing archive file. This is an important safety feature to preserve the integrity of your archive in case of  administrator error (such as sending the output of two different servers to the same archive directory).    
It is advisable to test your proposed archive command to ensure that it indeed does not overwrite an existing file, and that it returns nonzero status in this case. The example command above for Unix ensures this by including a separate test step.

### DIFF
--- a/patroni/postgresql/rewind.py
+++ b/patroni/postgresql/rewind.py
@@ -1,6 +1,8 @@
 import logging
 import os
+import re
 import shlex
+import shutil
 import six
 import subprocess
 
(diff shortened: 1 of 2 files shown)


CyberDem0n:
:+1:

hughcapet:
👍

PR #2387, opened 2022-08-17
Check if .ready file exists in _archive_ready_wals

hughcapet:
Add missing check that .ready file actually exists before calling archive_command in _archive_ready_wals(). It might happen that archive tool has already done this work for us after we called it once (wal-e uses this optimisation)


CyberDem0n:
:+1:

hughcapet:
👍

Issue #1132, opened 2019-08-02
Patroni requires services create permission in Kubernetes


jorsol:
I'm using patroni in Kubernetes, create a ServiceAccount and a Role without the "create" permision for "Services" since I create the service manually (the headless service to avoid deletion of example-config endpoint) but starting patroni gives me this error:

```
2019-08-02 14:48:15,882 ERROR: create_config_service failed
Traceback (most recent call last):
  File "/usr/lib/python2.7/site-packages/patroni/dcs/kubernetes.py", line 298, in _create_config_service
    if not self._api.create_namespaced_service(self._namespace, body):
  File "/usr/lib/python2.7/site-packages/patroni/dcs/kubernetes.py", line 50, in wrapper
    return getattr(self._api, func)(*args, **kwargs)
  File "/usr/lib/python2.7/site-packages/kubernetes/client/apis/core_v1_api.py", line 6820, in create_namespaced_service
    (data) = self.create_namespaced_service_with_http_info(namespace, body, **kwargs)
  File "/usr/lib/python2.7/site-packages/kubernetes/client/apis/core_v1_api.py", line 6905, in create_namespaced_service_with_http_info
    collection_formats=collection_formats)
  File "/usr/lib/python2.7/site-packages/kubernetes/client/api_client.py", line 321, in call_api
    _return_http_data_only, collection_formats, _preload_content, _request_timeout)
  File "/usr/lib/python2.7/site-packages/kubernetes/client/api_client.py", line 155, in __call_api
    _request_timeout=_request_timeout)
  File "/usr/lib/python2.7/site-packages/kubernetes/client/api_client.py", line 364, in request
    body=body)
  File "/usr/lib/python2.7/site-packages/kubernetes/client/rest.py", line 266, in POST
    body=body)
  File "/usr/lib/python2.7/site-packages/kubernetes/client/rest.py", line 222, in request
    raise ApiException(http_resp=r)
ApiException: (403)
Reason: Forbidden
HTTP response headers: HTTPHeaderDict({'Date': 'Fri, 02 Aug 2019 14:48:15 GMT', 'Content-Length': '314', 'Content-Type': 'application/json', 'X-Content-Type-Options': 'nosniff'})
[...]

PR #2390, opened 2022-08-22
Ignore 403s when access to Kubernetes Service is not needed

nhudson:
When running patroni on Kubernetes the logs are spammed with access errors checking if the Service exists.  According to the [documentation](https://github.com/zalando/patroni/blob/0b1b1e3b542714b47634391f7dbb00016d885bff/kubernetes/patroni_k8s.yaml#L190-L193) access to the Service is not needed.  This is an attempt to fix issue #1132 or at least do not log when the ServiceAccount doesn't have access to Service.


hughcapet:
👍

CyberDem0n:
:+1:

Issue #2388, opened 2022-08-18
No Failover Following NFS Disk Malfunction


andrewlecuyer:
**Describe the bug**
When the `PGDATA` directory is on an NFS disk is no longer accessible/reachable (i.e., because the NFS server is no longer running and/or accessible over the network), the TTL of the leader lock expires, but a failover does not occur.

**To Reproduce**
This was detected while testing failure scenarios in a lab.

Deploy two Patroni instances in a Kubernetes cluster, with Endpoints for the cluster's DCS.  When deploying the cluster, ensure the current primary is running on NFS, with the other using another form of storage (as to not be impacted by a NFS failure).

After the cluster is fully up-and-running, with the replica successfully replicating from the primary, terminate the NFS server (e.g. shut down the host running NFS).

At this point, the leader lock expires, and the primary simply stops logging:

```bash
$ kubectl logs -c database test1-instance1-hjlg-0 | tail -n 5
2022-08-16 00:05:07,009 INFO: no action. I am (test1-instance1-hjlg-0), the leader with the lock
2022-08-16 00:05:16,980 INFO: no action. I am (test1-instance1-hjlg-0), the leader with the lock
2022-08-16 00:05:26,959 INFO: no action. I am (test1-instance1-hjlg-0), the leader with the lock
2022-08-16 00:05:37,021 INFO: no action. I am (test1-instance1-hjlg-0), the leader with the lock
2022-08-16 00:05:46,984 INFO: no action. I am (test1-instance1-hjlg-0), the leader with the lock
# Nothing else is logged after this for the duration of the test.
```

Additionally, the readniness check for the primary (implemented using the Patroni `/readiness` REST endpoint) starts to fail:

```bash
$ kubectl describe pod test-instance1-hjlg-0 | tail -1
Warning  Unhealthy         2m9s (x75 over 17m)  kubelet            Readiness probe failed: HTTP probe failed with statuscode: 503
```

Yet, no failover occurs:

```bash
$ kubectl exec -c database test1-instance2-g5px-0 -- patronictl list
+------------------------+-----------------------------------+---------+---------+----+-----------+
| Member                 | Host                              | Role    | State   | TL | Lag in MB |
+ Cluster: test1-ha (7129659798145802323) -------------------+---------+---------+----+-----------+
| test1-instance1-hjlg-0 | test1-instance1-hjlg-0.test1-pods | Leader  | running |  1 |           |
[...]

PR #2395, opened 2022-08-24
Improve liveness probe

CyberDem0n:
it will start failing if the heartbeat loop isn't running longer than `ttl` on the primary or `2*ttl` on the replica.

Close https://github.com/zalando/patroni/issues/2388

### DIFF
--- a/tests/test_api.py
+++ b/tests/test_api.py
@@ -46,6 +46,10 @@ class MockPostgresql(object):
     def replica_cached_timeline(_):
         return 2
 
+    @staticmethod
+    def is_running():
+        return True
+
 
 class MockWatchdog(object):
     is_healthy = False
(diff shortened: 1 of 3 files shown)


tests/test_api.py
```diff
@@ -46,6 +46,10 @@ def postmaster_start_time():
     def replica_cached_timeline(_):
         return 2
 
+    @staticmethod
+    def is_running():
+        return True
```
hughcapet, line 51: Why don't we mock this call? We are sure we will not need to return False in this suit?

CyberDem0n: We mock the entire Postgresql.

hughcapet: Oh, I see. Sorry  🤦‍♀️

hughcapet:
👍

Issue #2414, opened 2022-09-16
"Primary" vs "Master" usage in docs


jchancojr:
**Describe the bug**
The documentation uses “master” in place of “Primary”. While unintentional, occasionally there is feedback that the usage is insensitive. Can we `s/master/primary` in the docs without needing to change the ~1500 references to master in the code? I'll modify the docs in my fork with the proposed changes to documentation and create a PR in the meantime. Kindly let me know? Thanks!

**To Reproduce**
https://patroni.readthedocs.io/en/latest/search.html?q=master&check_keywords=yes&area=default

**Expected behavior**
Replication relationships are referred to by Primary, Standby, Replica, etc

**Screenshots**
https://patroni.readthedocs.io/en/latest/search.html?q=master&check_keywords=yes&area=default

**Environment**
- Patroni version:
- PostgreSQL version:
- DCS (and its version):

**Patroni configuration file**

**patronictl show-config**

**Have you checked Patroni logs?**
NA

**Have you checked PostgreSQL logs?**
NA

**Have you tried to use GitHub issue search?**
Looked but found nothing.

**Additional context**
NA

CyberDem0n:
PR's are always welcome.
But, I would actually make a difference between the "primary" and the "leader"
The first one is more or less to indicate that Postgres is accepting writes, while the second one is more like a Patroni concept.
For example, there is no primary in a standby cluster, or there could be more than one primary in a pause mode.
That is, the simple search&replace will not work.

jchancojr:
Thanks very much @CyberDem0n. I'll get working on that for review.

PR #2417, opened 2022-09-20
docs: Change term "Master" to "primary" or "leader"

jchancojr:
Change the term "master" to either "primary" or "leader" for PostgreSQL and Patroni respectively. I've left the release notes alone to preserve release history.


jchancojr:
Fixes #2414

docs/ENVIRONMENT.rst
```diff
@@ -104,7 +104,7 @@ Kubernetes
 -  **PATRONI\_KUBERNETES\_NAMESPACE**: (optional) Kubernetes namespace where the Patroni pod is running. Default value is `default`.
 -  **PATRONI\_KUBERNETES\_LABELS**: Labels in format ``{label1: value1, label2: value2}``. These labels will be used to find existing objects (Pods and either Endpoints or ConfigMaps) associated with the current cluster. Also Patroni will set them on every object (Endpoint or ConfigMap) it creates.
 -  **PATRONI\_KUBERNETES\_SCOPE\_LABEL**: (optional) name of the label containing cluster name. Default value is `cluster-name`.
--  **PATRONI\_KUBERNETES\_ROLE\_LABEL**: (optional) name of the label containing Postgres role (`master` or `replica`). Patroni will set this label on the pod it is running in. Default value is `role`.
+-  **PATRONI\_KUBERNETES\_ROLE\_LABEL**: (optional) name of the label containing Postgres role (`primary` or `replica`). Patroni will set this label on the pod it is running in. Default value is `role`.
```
CyberDem0n, line 107: This one will be a master.

docs/SETTINGS.rst
```diff
@@ -17,21 +17,21 @@ Dynamic configuration is stored in the DCS (Distributed Configuration Store) and
...
--  **master\_start\_timeout**: the amount of time a master is allowed to recover from failures before failover is triggered (in seconds). Default is 300 seconds. When set to 0 failover is done immediately after a crash is detected if possible. When using asynchronous replication a failover can cause lost transactions. Worst case failover time for master failure is: loop\_wait + master\_start\_timeout + loop\_wait, unless master\_start\_timeout is zero, in which case it's just loop\_wait. Set the value according to your durability/availability tradeoff.
+-  **master\_start\_timeout**: the amount of time a leader is allowed to recover from failures before failover is triggered (in seconds). Default is 300 seconds. When set to 0 failover is done immediately after a crash is detected if possible. When using asynchronous replication a failover can cause lost transactions. Worst case failover time for leader failure is: loop\_wait + master\_start\_timeout + loop\_wait, unless master\_start\_timeout is zero, in which case it's just loop\_wait. Set the value according to your durability/availability tradeoff.
```
CyberDem0n, line 20: This one is about the primary.
```suggestion
-  **master\_start\_timeout**: the amount of time a primary is allowed to recover from failures before failover is triggered (in seconds). Default is 300 seconds. When set to 0 failover is done immediately after a crash is detected if possible. When using asynchronous replication a failover can cause lost transactions. Worst case failover time for primary failure is: loop\_wait + master\_start\_timeout + loop\_wait, unless master\_start\_timeout is zero, in which case it's just loop\_wait. Set the value according to your durability/availability tradeoff.
```

docs/SETTINGS.rst
```diff
@@ -17,21 +17,21 @@ Dynamic configuration is stored in the DCS (Distributed Configuration Store) and
...
-    -  **create\_replica\_methods**: an ordered list of methods that can be used to bootstrap standby leader from the remote master, can be different from the list defined in :ref:`postgresql_settings`
-    -  **restore\_command**: command to restore WAL records from the remote master to standby leader, can be different from the list defined in :ref:`postgresql_settings`
+    -  **host**: an address of remote primary
+    -  **port**: a port of remote primary
+    -  **primary\_slot\_name**: which slot on the remote primary to use for replication. This parameter is optional, the default value is derived from the instance name (see function `slot_name_from_member_name`).
+    -  **create\_replica\_methods**: an ordered list of methods that can be used to bootstrap standby leader from the remote primary, can be different from the list defined in :ref:`postgresql_settings`
+    -  **restore\_command**: command to restore WAL records from the remote primary to standby leader, can be different from the list defined in :ref:`postgresql_settings`
```
CyberDem0n, line 34: It is not necessarily the primary. Could be any remote node, even if it is running as a standby.

docs/ENVIRONMENT.rst
```diff
@@ -127,7 +127,7 @@ PostgreSQL
 -  **PATRONI\_POSTGRESQL\_CONFIG\_DIR**: The location of the Postgres configuration directory, defaults to the data directory. Must be writable by Patroni.
 -  **PATRONI\_POSTGRESQL\_BIN_DIR**: Path to PostgreSQL binaries. (pg_ctl, pg_rewind, pg_basebackup, postgres) The  default value is an empty string meaning that PATH environment variable will be used to find the executables.
 -  **PATRONI\_POSTGRESQL\_PGPASS**: path to the `.pgpass <https://www.postgresql.org/docs/current/static/libpq-pgpass.html>`__ password file. Patroni creates this file before executing pg\_basebackup and under some other circumstances. The location must be writable by Patroni.
--  **PATRONI\_REPLICATION\_USERNAME**: replication username; the user will be created during initialization. Replicas will use this user to access master via streaming replication
+-  **PATRONI\_REPLICATION\_USERNAME**: replication username; the user will be created during initialization. Replicas will use this user to access the primary via streaming replication
```
CyberDem0n, line 130: It is not necessarily the primary. Just a node.

docs/SETTINGS.rst
```diff
@@ -262,7 +262,7 @@ PostgreSQL
             -  **gssencmode**: (optional) maps to the `gssencmode <https://www.postgresql.org/docs/current/libpq-connect.html#LIBPQ-CONNECT-GSSENCMODE>`__ connection parameter, which determines whether or with what priority a secure GSS TCP/IP connection will be negotiated with the server
             -  **channel_binding**: (optional) maps to the `channel_binding <https://www.postgresql.org/docs/current/libpq-connect.html#LIBPQ-CONNECT-CHANNEL-BINDING>`__ connection parameter, which controls the client's use of channel binding.
         -  **replication**:
-            -  **username**: replication username; the user will be created during initialization. Replicas will use this user to access master via streaming replication
+            -  **username**: replication username; the user will be created during initialization. Replicas will use this user to access the primary via streaming replication
```
CyberDem0n, line 265: It is not necessarily the primary. Just a node.

docs/replica_bootstrap.rst
```diff
@@ -63,7 +63,7 @@ Building replicas
 -----------------
 
 Patroni uses tried and proven ``pg_basebackup`` in order to create new replicas. One downside of it is that it requires
-a running master node. Another one is the lack of 'on-the-fly' compression for the backup data and no built-in cleanup
+a running primary node. Another one is the lack of 'on-the-fly' compression for the backup data and no built-in cleanup
```
CyberDem0n, line 66:
```suggestion
a running leader node. Another one is the lack of 'on-the-fly' compression for the backup data and no built-in cleanup
```

docs/replica_bootstrap.rst
```diff
@@ -176,10 +176,10 @@ Standby cluster
 ---------------
 
 Another available option is to run a "standby cluster", that contains only of
-standby nodes replicating from some remote master. This type of clusters has:
+standby nodes replicating from some remote primary. This type of clusters has:
```
CyberDem0n, line 179:
```suggestion
standby nodes replicating from some remote node. This type of clusters has:
```

docs/replica_bootstrap.rst
```diff
@@ -176,10 +176,10 @@ Standby cluster
...
 
 * "standby leader", that behaves pretty much like a regular cluster leader,
-  except it replicates from a remote master.
+  except it replicates from a remote leader.
```
CyberDem0n, line 182:
```suggestion
  except it replicates from a remote node.
```

docs/replication_modes.rst
```diff
@@ -13,7 +13,7 @@ In asynchronous mode the cluster is allowed to lose some committed transactions
 
 The amount of transactions that can be lost is controlled via ``maximum_lag_on_failover`` parameter. Because the primary transaction log position is not sampled in real time, in reality the amount of lost data on failover is worst case bounded by  ``maximum_lag_on_failover`` bytes of transaction log plus the amount that is written in the last ``ttl`` seconds (``loop_wait``/2 seconds in the average case). However typical steady state replication delay is well under a second.
 
-By default, when running leader elections, Patroni does not take into account the current timeline of replicas, what in some cases could be undesirable behavior. You can prevent the node not having the same timeline as a former master become the new leader by changing the value of ``check_timeline`` parameter to ``true``.
+By default, when running leader elections, Patroni does not take into account the current timeline of replicas, what in some cases could be undesirable behavior. You can prevent the node not having the same timeline as a former leader become the new leader by changing the value of ``check_timeline`` parameter to ``true``.
```
CyberDem0n, line 16: Timelines are created when Postgres is promoted to the primary.

```suggestion
By default, when running leader elections, Patroni does not take into account the current timeline of replicas, what in some cases could be undesirable behavior. You can prevent the node not having the same timeline as a former primary become the new leader by changing the value of ``check_timeline`` parameter to ``true``.
```

docs/watchdog.rst
```diff
@@ -3,7 +3,7 @@
 Watchdog support
 ================
 
-Having multiple PostgreSQL servers running as master can result in transactions lost due to diverging timelines. This situation is also called a split-brain problem. To avoid split-brain Patroni needs to ensure PostgreSQL will not accept any transaction commits after leader key expires in the DCS. Under normal circumstances Patroni will try to achieve this by stopping PostgreSQL when leader lock update fails for any reason. However, this may fail to happen due to various reasons:
+Having multiple PostgreSQL servers running as leader can result in transactions lost due to diverging timelines. This situation is also called a split-brain problem. To avoid split-brain Patroni needs to ensure PostgreSQL will not accept any transaction commits after leader key expires in the DCS. Under normal circumstances Patroni will try to achieve this by stopping PostgreSQL when leader lock update fails for any reason. However, this may fail to happen due to various reasons:
```
CyberDem0n, line 6:
```suggestion
Having multiple PostgreSQL servers running as primary can result in transactions lost due to diverging timelines. This situation is also called a split-brain problem. To avoid split-brain Patroni needs to ensure PostgreSQL will not accept any transaction commits after leader key expires in the DCS. Under normal circumstances Patroni will try to achieve this by stopping PostgreSQL when leader lock update fails for any reason. However, this may fail to happen due to various reasons:
```

docs/watchdog.rst
```diff
@@ -13,7 +13,7 @@ Having multiple PostgreSQL servers running as master can result in transactions
...
 
-Patroni will try to activate the watchdog before promoting PostgreSQL to master. If watchdog activation fails and watchdog mode is ``required`` then the node will refuse to become master. When deciding to participate in leader election Patroni will also check that watchdog configuration will allow it to become leader at all. After demoting PostgreSQL (for example due to a manual failover) Patroni will disable the watchdog again. Watchdog will also be disabled while Patroni is in paused state.
+Patroni will try to activate the watchdog before promoting PostgreSQL to leader. If watchdog activation fails and watchdog mode is ``required`` then the node will refuse to become leader. When deciding to participate in leader election Patroni will also check that watchdog configuration will allow it to become leader at all. After demoting PostgreSQL (for example due to a manual failover) Patroni will disable the watchdog again. Watchdog will also be disabled while Patroni is in paused state.
```
CyberDem0n, line 16:
```suggestion
Patroni will try to activate the watchdog before promoting PostgreSQL to primary. If watchdog activation fails and watchdog mode is ``required`` then the node will refuse to become leader. When deciding to participate in leader election Patroni will also check that watchdog configuration will allow it to become leader at all. After demoting PostgreSQL (for example due to a manual failover) Patroni will disable the watchdog again. Watchdog will also be disabled while Patroni is in paused state.
```

jchancojr:
Thanks very much @CyberDem0n

Let me know if there's anything else I missed.

docs/pause.rst
```diff
@@ -17,13 +17,13 @@ When Patroni runs in a paused mode, it does not change the state of PostgreSQL,
...
+- If there is no primary lock in the cluster, the running primary acquires the lock. If there is more than one primary node, then the first primary to acquire the lock wins. If there are no primary altogether, Patroni does not try to promote any replicas. There is an exception in this rule: if there is no primary lock because the old primary has demoted itself due to the manual promotion, then only the candidate node mentioned in the promotion request may take the primary lock. When the new primary lock is granted (i.e. after promoting a replica manually), Patroni makes sure the replicas that were streaming from the previous primary will switch to the new one.
```
hughcapet, line 26: I guess "leader" was replaced with "primary" by mistake here

hunleyd, line 26:
```suggestion
- If there is no leader lock in the cluster, the running primary acquires the lock. If there is more than one primary node, then the first primary to acquire the lock wins. If there are no primary altogether, Patroni does not try to promote any replicas. There is an exception in this rule: if there is no leader lock because the old primary has demoted itself due to the manual promotion, then only the candidate node mentioned in the promotion request may take the leader lock. When the new leader lock is granted (i.e. after promoting a replica manually), Patroni makes sure the replicas that were streaming from the previous leader will switch to the new one.
```

hunleyd (changes requested):

docs/pause.rst
jchancojr: lgtm

hughcapet:
👍

docs/SETTINGS.rst
```diff
@@ -17,21 +17,21 @@ Dynamic configuration is stored in the DCS (Distributed Configuration Store) and
...
-    -  **create\_replica\_methods**: an ordered list of methods that can be used to bootstrap standby leader from the remote master, can be different from the list defined in :ref:`postgresql_settings`
-    -  **restore\_command**: command to restore WAL records from the remote master to standby leader, can be different from the list defined in :ref:`postgresql_settings`
+    -  **host**: an address of remote node
+    -  **port**: a port of remote node
+    -  **primary\_slot\_name**: which slot on the remote node to use for replication. This parameter is optional, the default value is derived from the instance name (see function `slot_name_from_member_name`).
+    -  **create\_replica\_methods**: an ordered list of methods that can be used to bootstrap standby node from the remote primary, can be different from the list defined in :ref:`postgresql_settings`
+    -  **restore\_command**: command to restore WAL records from the remote primary to standby node, can be different from the list defined in :ref:`postgresql_settings`
```
CyberDem0n, line 34:
```suggestion
    -  **restore\_command**: command to restore WAL records from the remote primary to standby leader, can be different from the list defined in :ref:`postgresql_settings`
```

docs/existing_data.rst
```diff
@@ -29,11 +29,11 @@ Major Upgrade of PostgreSQL Version
 The only possible way to do a major upgrade currently is:
 
 1. Stop Patroni
-2. Upgrade PostgreSQL binaries and perform `pg_upgrade <https://www.postgresql.org/docs/current/pgupgrade.html>`_ on the master node
+2. Upgrade PostgreSQL binaries and perform `pg_upgrade <https://www.postgresql.org/docs/current/pgupgrade.html>`_ on the primary node
 3. Update patroni.yml
 4. Remove the initialize key from DCS or wipe complete cluster state from DCS. The second one could be achieved by running ``patronictl remove <cluster-name>``. It is necessary because pg_upgrade runs initdb which actually creates a new database with a new PostgreSQL system identifier.
 5. If you wiped the cluster state in the previous step, you may wish to copy patroni.dynamic.json from old data dir to the new one.  It will help you to retain some PostgreSQL parameters you had set before.
-6. Start Patroni on the master node.
+6. Start Patroni on the leader node.
```
CyberDem0n, line 36: If all Patroni nodes in the cluster are stopped there will be no leader.

```suggestion
6. Start Patroni on the primary node.
```

docs/replica_bootstrap.rst
```diff
@@ -137,7 +137,7 @@ A special ``no_params`` parameter, if defined, restricts passing parameters to c
 A ``basebackup`` method is a special case: it will be used if
 ``create_replica_methods`` is empty, although it is possible
 to list it explicitly among the ``create_replica_methods`` methods. This method initializes a new replica with the
-``pg_basebackup``, the base backup is taken from the master unless there are replicas with ``clonefrom`` tag, in which case one
+``pg_basebackup``, the base backup is taken from the primary unless there are replicas with ``clonefrom`` tag, in which case one
```
CyberDem0n, line 140:
```suggestion
``pg_basebackup``, the base backup is taken from the leader unless there are replicas with ``clonefrom`` tag, in which case one
```

docs/replica_bootstrap.rst
```diff
@@ -123,11 +123,11 @@ to execute and any custom parameters that should be passed to that command. All
 --role
     Always 'replica'
 --connstring
-    Connection string to connect to the cluster member to clone from (master or other replica). The user in the
+    Connection string to connect to the cluster member to clone from (primary or other replica). The user in the
     connection string can execute SQL and replication protocol commands.
 
 A special ``no_master`` parameter, if defined, allows Patroni to call the replica creation method even if there is no
-running master or replicas. In that case, an empty string will be passed in a connection string. This is useful for
+running primary or replicas. In that case, an empty string will be passed in a connection string. This is useful for
```
CyberDem0n, line 130:
```suggestion
running leader or replicas. In that case, an empty string will be passed in a connection string. This is useful for
```

docs/SETTINGS.rst
cbandy: Last I checked, this setting affects both standby leader and standby replicas: #1341

CyberDem0n: @cbandy, I agree, it wasn't a good idea to change only one word.
Now I rephrased it a little bit (the previous comment is modified)

docs/pause.rst
jchancojr: _should_ be fixed now. kindly let me know? thanks!

CyberDem0n:
:+1:

hughcapet:
👍

CyberDem0n:
Merged, thank you @jchancojr, @cbandy, @hunleyd

PR #2419, opened 2022-09-22
Avoid cloning while bootstrap is running

ants:
If cluster has a create replica method that does not require a leader it can get triggered while bootstrap is running. If that method comes up with an accessible cluster faster than the bootstrap completes it will get promoted as the leader, only to lose leader lock when bootstrap completes. To fix this we only consider leaderless create replica methods if sysid is non-empty, i.e. there is no bootstrap running.


hughcapet:
👍

CyberDem0n:
:+1:

PR #2423, opened 2022-09-29
Fix exception handling in create_config_service

hughcapet:
The recently updated condition used to produce erroneous exception line in logs


CyberDem0n:
:+1:

hughcapet:
👍

Issue #2424, opened 2022-10-11
Unexpected EOF failure with OpenSSL 3.0


reddare:
**Describe the bug**

In an attempt to perform a switchover I get this error:
urllib3.exceptions.MaxRetryError: HTTPSConnectionPool(host='patronitest2.smartly.io', port=8008): Max retries exceeded with url: /switchover (Caused by SSLError(SSLError("read error: Error([('SSL routines', '', 'unexpected eof while reading')])")))

**The unexpected EOF failure was introduced in OpenSSL 3.0 to prevent truncation attack**
https://www.openssl.org/docs/manmaster/man3/SSL_CTX_set_options.html#SSL_OP_IGNORE_UNEXPECTED_EOF

OpenSSL 3.0 installed by default in Ubuntu 22.04. Therefore, now Patroni and Ubuntu 22.04 are incompatible.

It seems that this is still a Patroni problem: to ensure correct operation for the OpenSSL 3.0 according to the new requirements?

**Expected behavior** : switchover

**Environment**
    OS: Ubuntu 22.04.1 LTS
    Patroni version: 2.1.4
    PostgreSQL version: 14
    DCS (and its version): etcd3-0.4.5

**Additional context**
```
During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/usr/local/lib/python3.10/dist-packages/urllib3/connectionpool.py", line 723, in urlopen
    response = self.ResponseCls.from_httplib(
  File "/usr/local/lib/python3.10/dist-packages/urllib3/response.py", line 652, in from_httplib
    resp = ResponseCls(
  File "/usr/local/lib/python3.10/dist-packages/urllib3/response.py", line 265, in __init__
    self._body = self.read(decode_content=decode_content)
  File "/usr/local/lib/python3.10/dist-packages/urllib3/response.py", line 565, in read
    with self._error_catcher():
  File "/usr/lib/python3.10/contextlib.py", line 153, in __exit__
    self.gen.throw(typ, value, traceback)
  File "/usr/local/lib/python3.10/dist-packages/urllib3/response.py", line 454, in _error_catcher
    raise SSLError(e)
urllib3.exceptions.SSLError: ("read error: Error([('SSL routines', '', 'unexpected eof while reading')])",)

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/usr/local/lib/python3.10/dist-packages/patroni/ctl.py", line 719, in _do_failover_or_switchover
[...]

PR #2425, opened 2022-10-12
Explicitly shut down SSL connection before socket

CyberDem0n:
This is handled by calling the unwrap() method on SSLSocket.
In addition to that simplify code that handles deferred handshakes.

Close https://github.com/zalando/patroni/issues/2424


hughcapet:
👍

idanovinda:
👍

PR #2427, opened 2022-10-12
A small fix in unit tests

CyberDem0n:
Not all external resources were properly mocked

### DIFF
--- a/tests/test_consul.py
+++ b/tests/test_consul.py
@@ -161,6 +161,7 @@ class TestConsul(unittest.TestCase):
         self.c.write_leader_optime('1')
 
     @patch.object(consul.Consul.Session, 'renew', Mock())
+    @patch.object(consul.Consul.KV, 'put', Mock(side_effect=ConsulException))
     def test_update_leader(self):
         self.c.update_leader(12345)
 
@@ -217,6 +218,7 @@ class TestConsul(unittest.TestCase):
         d['role'] = 'bla'
         self.assertIsNone(self.c.update_service({}, d))
 
+    @patch.object(consul.Consul.KV, 'put', Mock(side_effect=ConsulException))
     def test_reload_config(self):
         self.assertEqual([], self.c._service_tags)
         self.c.reload_config({'consul': {'token': 'foo', 'register_service': True, 'service_tags': ['foo']},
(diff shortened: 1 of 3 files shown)


tests/test_patroni.py
```diff
@@ -54,6 +54,7 @@ def test_no_config(self):
         self.assertRaises(SystemExit, patroni_main)
 
     @patch('sys.argv', ['patroni.py', '--validate-config', 'postgres0.yml'])
+    @patch('socket.socket.connect_ex', Mock(return_value=1))
```
hughcapet, line 57: [Off-topic] I am wondering, is validate_host_port() with arg listen=False covered anyhow? 

https://github.com/zalando/patroni/blob/641647eeacbcdfeaba91aa48631d2018d8d06176/patroni/validator.py#L48-L52

CyberDem0n: They are (according to coverage).
And the `connect_ex()` is mocked here: https://github.com/zalando/patroni/blob/db9b5962ec5b921c4c98823f623084f6be58b648/tests/test_validator.py#L111-L131

hughcapet: oh, thanks. Missed that part

tests/test_consul.py
```diff
@@ -217,6 +218,7 @@ def test_update_service(self):
         d['role'] = 'bla'
         self.assertIsNone(self.c.update_service({}, d))
 
+    @patch.object(consul.Consul.KV, 'put', Mock(side_effect=ConsulException))
```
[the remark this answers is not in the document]
CyberDem0n: Further down it calls `touch_member()` in order to make sure that new settings are applied (service should be deregistered)

tests/test_patroni.py
hughcapet: although I still don't see any explicit listen=False setting

hughcapet:
👍

tests/test_patroni.py
CyberDem0n: `False` is a default value.
Calls happen from here: https://github.com/zalando/patroni/blob/db9b5962ec5b921c4c98823f623084f6be58b648/patroni/validator.py#L60-L66

hughcapet: 🤦‍♀️ sure, thanks again

idanovinda:
👍

PR #2428, opened 2022-10-12
Compatibility with kazoo-2.9.0

CyberDem0n:
Now the select() method may raise `TypeError` and `IOError` exceptions if the socket is closed

### DIFF
--- a/patroni/dcs/zookeeper.py
+++ b/patroni/dcs/zookeeper.py
@@ -1,6 +1,7 @@
 import json
 import logging
 import select
+import six
 import time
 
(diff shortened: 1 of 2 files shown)


patroni/dcs/zookeeper.py
```diff
@@ -53,8 +54,10 @@ def select(self, *args, **kwargs):
         """Python3 raises `ValueError` if socket is closed, because fd == -1"""
```
hughcapet, line 54: And when other exceptions are raised? :)

hughcapet:
👍

idanovinda:
👍

PR #2432, opened 2022-10-17
Behave tests on Windows

CyberDem0n:
Windows doesn't support `SIGTERM`, but our behave tests in majority of cases relying on Patroni graceful shutdown.
In order to emulate the behaviour we introduced the new REST API endpoint `POST /sigterm`. The endpoint works only on Windows and when `BEHAVE_DEBUG` environment variable is set.
Besides that some minor adjustments in behave tests were done. Mainly related to backslash-slash handling.

In addition to that improve test coverage on Windows by properly mocking access to filesystem and avoiding calling
 `subprocess.call()`. Specifically, symlink creation on Windows requires Admin privileges and there is no `true.exe`.

### DIFF
--- a/features/environment.py
+++ b/features/environment.py
@@ -14,6 +14,7 @@ import yaml
 
 import patroni.psycopg as psycopg
 
+from patroni.request import PatroniRequest
 from six.moves.BaseHTTPServer import BaseHTTPRequestHandler, HTTPServer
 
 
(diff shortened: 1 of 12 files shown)


hughcapet:
Nice to see we finally have such an experienced Patroni hacker, who runs it on Windows :)

hughcapet:
👍

hughcapet:
👍

idanovinda:
👍

Issue #2431, opened 2022-10-16
patronictl timeout with raft


leagueofsoups:
**Describe the bug**
I have question about patroni cluster with three nodes with raft. There are probability to get infinity time of execution patronictl -c <cfg_path> topology, after one node shutdown. How I can fix this timeout?

**To Reproduce**

1) My test cluster topology
```
+----------+-----------+---------+---------+----+-----------+
| Member   | Host      | Role    | State   | TL | Lag in MB |
+ Cluster: demo-cluster-1 (7155048679605235377) +-----------+
| node-2   | 10.2.0.21 | Leader  | running |  1 |           |
| + node-1 | 10.2.0.6  | Replica | running |  1 |         0 |
| + node-3 | 10.2.0.25 | Replica | running |  1 |         0 |
+----------+-----------+---------+---------+----+-----------+
```
2) Then, I exec halt -p on node-3
3) Check partner_addrs on node-1
```
cat /etc/patroni.yml | grep part
  partner_addrs: [ "10.2.0.21:5010","10.2.0.25:5010" ]
```
4) Command for problem reproduce:
```
while true; do strace patronictl -c /etc/patroni.yml topology; sleep 1; done
```
patronictl try communicate with inaccessible node:
```
connect(3, {sa_family=AF_INET, sin_port=htons(5010), sin_addr=inet_addr("10.2.0.25")}, 16) = -1 EINPROGRESS (Operation now in progress)
poll([{fd=3, events=POLLIN|POLLOUT|POLLERR}], 1, 500) = 0 (Timeout)
poll([{fd=3, events=POLLIN|POLLOUT|POLLERR}], 1, 500) = 0 (Timeout)
poll([{fd=3, events=POLLIN|POLLOUT|POLLERR}], 1, 500) = 0 (Timeout)
poll([{fd=3, events=POLLIN|POLLOUT|POLLERR}], 1, 500) = 0 (Timeout)
poll([{fd=3, events=POLLIN|POLLOUT|POLLERR}], 1, 500) = 0 (Timeout)
poll([{fd=3, events=POLLIN|POLLOUT|POLLERR}], 1, 500) = 0 (Timeout)
poll([{fd=3, events=POLLIN|POLLOUT|POLLERR}], 1, 500) = 0 (Timeout)
```
This is because TcpUtility (pysyncobj/utility.py) have timeout = 900 in constructor. 

https://github.com/bakwc/PySyncObj/blob/master/pysyncobj/utility.py#L42
[...]

PR #2435, opened 2022-10-18
Introduce configurable timeout in TcpUtility

CyberDem0n:
the value is calculated based on the number of nodes and retry_timeout.

Close https://github.com/zalando/patroni/issues/2431

### DIFF
--- a/patroni/dcs/raft.py
+++ b/patroni/dcs/raft.py
@@ -39,9 +39,9 @@ setattr(TCPNode, 'ip', property(resolve_host))
 
 class SyncObjUtility(object):
 
-    def __init__(self, otherNodes, conf):
+    def __init__(self, otherNodes, conf, retry_timeout=10):
         self._nodes = otherNodes
-        self._utility = TcpUtility(conf.password)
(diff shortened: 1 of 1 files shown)


patroni/dcs/raft.py
```diff
@@ -121,7 +121,8 @@ def __init__(self, on_ready, on_set, on_delete, **config):
                            journalFile=(file_template + '.journal' if self_addr else None),
                            onReady=on_ready, dynamicMembershipChange=True)
 
-        super(KVStoreTTL, self).__init__(self_addr, partner_addrs, conf)
+        retry_timeout = int(config.get('retry_timeout') or 10)
```
hughcapet, line 124: Can `retry_timeout` be explicitly set to 0? We want it to be 10 then?

hughcapet: Am also wondering: don't we want this code-wide default timeout = 10 be set via some variable?

hughcapet:
👍

idanovinda:
👍

PR #2437, opened 2022-10-19
Inptroduce postgresql.proxy_address

CyberDem0n:
It will be written to member key in DCS as the `proxy_url` and could be used/useful for service discovery.


idanovinda:
👍

Issue #2438, opened 2022-10-20
TypeError: '<=' not supported between instances of 'NoneType' and 'NoneType'


jopadi:
The error appears after adding 2 new replicas at the same time. 

one replica was showing the error  below, but the error is gone after the replica is synced:

 ```
2022-10-20 12:51:02,581 ERROR: Exception when changing replication slots
Traceback (most recent call last):
  File "/usr/local/lib/python3.6/dist-packages/patroni/postgresql/slots.py", line 226, in sync_replication_slots
    self.check_logical_slots_readiness(cluster, nofailover, replicatefrom)
  File "/usr/local/lib/python3.6/dist-packages/patroni/postgresql/slots.py", line 278, in check_logical_slots_readiness
    if not value or self._unready_logical_slots[name] <= catalog_xmin <= value['catalog_xmin']:
TypeError: '<=' not supported between instances of 'NoneType' and 'NoneType'
```

one replica is showing an error and is not streaming correctly for a while after showing the previous error only once the replica was synced:
 `
2022-10-20 13:06:12,577 WARNING: Physical slot i_04eaab40edb01b3ac does not exist on the primary
`

environment:
instances: 1 master and 2 replicas
instance type: i3.2xlarge
PG Version: 13.7
1  replication slot for FES/logical replication

```
 slot_name |  plugin  | slot_type | datoid |       database       | temporary | active | active_pid | xmin | catalog_xmin |  restart_lsn  | confirmed_flush_lsn | wal_status | safe_wal_size
-----------+----------+-----------+--------+----------------------+-----------+--------+------------+------+--------------+---------------+---------------------+------------+---------------
 fes       | wal2json | logical   |  16426 | dbname | f         | f      |            |      |   2444389396 | 105A/C6DB88B0 | 105A/C7039FF8       | reserved   |
```

**patronictl show-config**
```
loop_wait: 10
maximum_lag_on_failover: 33554432
pause: false
postgresql:
  parameters:
    archive_mode: 'on'
    archive_timeout: 1800s
    autovacuum_analyze_scale_factor: 0.02
    autovacuum_max_workers: 5
    autovacuum_vacuum_scale_factor: 0.05
    checkpoint_completion_target: 0.9
    hot_standby: 'on'
    log_autovacuum_min_duration: 0
    log_checkpoints: 'on'
    log_connections: 'on'
    log_disconnections: 'on'
    log_line_prefix: '%t [%p]: [%l-1] %c %x %d %u %a %h '
[...]

PR #2439, opened 2022-10-24
Fix a little bug in check_logical_slots_readiness()

CyberDem0n:
If there is no `catalog_xmin` on physical slot on the primary the only further check that makes sense is whether the `hot_standby_feedback` is enabled.

Close https://github.com/zalando/patroni/issues/2438


hughcapet:
👍

FxKu:
:+1:

CyberDem0n:
👍

PR #2440, opened 2022-10-25
Make sure self.__retry_timeout is set

CyberDem0n:
The default value was None, which is the same as no timeout.


hughcapet:
👍

CyberDem0n:
:+1:

Issue #1800, opened 2020-12-14
Log status on change only


stephan2012:
**Describe the bug**

Patroni currently floods logs with status messages every five seconds:

```
2020-12-14 17:11:58,430 INFO: Lock owner: my-application-pg-0; I am my-applcation-pg-1
2020-12-14 17:11:58,430 INFO: does not have lock
2020-12-14 17:11:58,431 INFO: no action.  i am a secondary and i am following a leader
```

Emitting messages on status changes (or errors) only would not only reduce the number of log events but also make the log more clear.

**To Reproduce**
Setup any replicated database.

**Expected behavior**
Emit a log message only if there is news …

**Screenshots**
N/A

**Environment**
- Patroni version: `1.6.5`
- PostgreSQL version: Postgres 11
- DCS (and its version): N/A

**Patroni configuration file**
Generated by the Postgres Operator, N/A here

**patronictl show-config**
```
loop_wait: 5
maximum_lag_on_failover: 33554432
postgresql:
  parameters:
    archive_mode: 'on'
    archive_timeout: 1800s
    autovacuum_analyze_scale_factor: 0.02
    autovacuum_max_workers: 5
    autovacuum_vacuum_scale_factor: 0.05
    checkpoint_completion_target: 0.9
    hot_standby: 'on'
    log_autovacuum_min_duration: 0
    log_checkpoints: 'on'
    log_connections: 'on'
    log_disconnections: 'on'
    log_line_prefix: '%t [%p]: [%l-1] %c %x %d %u %a %h '
    log_lock_waits: 'on'
    log_min_duration_statement: 500
    log_statement: ddl
    log_temp_files: 0
    max_connections: '130'
    max_replication_slots: 10
    max_wal_senders: 10
    tcp_keepalives_idle: 900
    tcp_keepalives_interval: 100
    track_functions: all
    wal_keep_segments: 8
    wal_level: hot_standby
    wal_log_hints: 'on'
  use_pg_rewind: true
  use_slots: true
retry_timeout: 5
ttl: 10
```

**Have you checked Patroni logs?**
N/A

**Have you checked PostgreSQL logs?**
N/A

**Have you tried to use GitHub issue search?**
Yes.

**Additional context**

CyberDem0n:
Let's use correct words, it doesn't flood logs, but provides update on heartbeat status. In case of problems such heartbeat logs are extremely useful.
[...]

Issue #1203, opened 2019-10-02
Patroni Lag Problem


rootqu:
Hello,

we have 3 servers and was installed as sql1, sql2 and etcd.

etcd1 - 10.10.10.101 - etcdctl version: 3.2.22
sql1 - 10.10.10.102 - patronictl version 1.5.1
sql2 - 10.10.10.103 - patronictl version 1.5.1

# patronictl -d etcd://10.10.10.101:2379 list sql
+---------+--------+--------------+--------+---------+-----------+
| Cluster | Member |     Host     |  Role  |  State  | Lag in MB |
+---------+--------+--------------+--------+---------+-----------+
|   sql|  sql1| 10.10.10.102 |        | running |   13591.0 |
|   sql|  sql2| 10.10.10.103| Leader | running |       0.0 |
+---------+--------+--------------+--------+---------+-----------+

we are having a LAG problem and I didn't seen problems with network access.

Error Logs:

sql1:

2019-10-02 13:24:26.611 +04 [24938] LOG:  started streaming WAL from primary at 96/B7000000 on timeline 26
2019-10-02 13:24:26.611 +04 [24938] FATAL:  could not receive data from WAL stream: ERROR:  requested WAL segment 0000001B00000096000000B7 has already been removed

sql2:

2019-10-02 13:25:38.671 +04 [10876] ERROR:  requested WAL segment 0000001B00000096000000B7 has already been removed

sql2 is Leader to sql cluster. what could be problem of this error? why is LAG happening and giving errors? what would you offer me?

Thanks for all help.

CyberDem0n:
Hi @rootqu,

please use [search](https://github.com/zalando/patroni/issues?q=is%3Aissue+requested+WAL+segment+has+already+been+removed+is%3Aclosed)

Issue #621, opened 2018-02-06
Logging


mrquokka:
Hi there, can someone help to clear unnecessary logs, such us
```
фев 06 14:42:07 user-VirtualBox patroni[14532]: 2018-02-06 14:42:07,473 INFO: Lock owner: postgresql0; I am postgresql0
фев 06 14:42:07 user-VirtualBox patroni[14532]: 2018-02-06 14:42:07,482 INFO: no action.  i am the leader with the lock
```
I tried to change postgresql parameters, such us:

```
    log_min_messages: 'NOTICE'
    client_min_messages: 'NOTICE'
    log_min_error_statement: 'NOTICE'
```
and 
__init__.py on source:

`    logging.basicConfig(format=logformat, level=logging.WARNING)`

But nothing changes

alexeyklyukin:
Hi,

Those are not PostgreSQL, but Patroni logs, so changing PostgreSQL parameters won't affect them. Why do you think they are 'unnecessary'? Patroni reports every iteration of the event loop its role and who is the leader. Knowing that is important in order to understand, for instance, why did the failover happen.

You can mute the INFO messages from the event loop by setting the level in a `ha.py`.

We may also provide some way of controlling log levels in the future. Feel free to contribute a PR if you are interested.

antineutron:
I've submitted #622 to make log level configurable using environment vars, but as these are at INFO level it'd still require hacking ha.py to mute the event loop messages.

mrquokka:
Alexey, thanks for answer. Because system work stable, without reboot and changes, but they write to syslog every second. I really need only first message who i master and messanges when patroni change master, no more

alexeyklyukin:
@mrquokka feel free to submit a PR. I think it makes sense to have a group of N messages (so that Patroni says 'Leader with a lock (repeated N times)'.

@antineutron I think your PR needs to change how the loggers are spawned, so that there is a single place to set the logging level. Also, we might want to set this parameter in the configuration file rather than via the environment variables (and change it with a reload of Patroni without stopping its operations).

combrs:
+ to log repeating messages about leader & lock in batches.
For example, in config set logging option 'log_batches: 100' 
and then buffer repeated messages in memory. Write it to disk like 

[Repeated 100 times] : Lock owner: postgresql0; I am postgresql0; no action; i'm a leader with the lock

In one string
Other messages, of course, write immediately.

PR #2442, opened 2022-10-27
logging: Log no action messages at debug level

### LINKED ISSUE #1154: Excessive log messages
During normal operation the logs are very noisy with low value information.  I was going to see if I could change the log level (we use the Postgres operator), but at a glance the important messages (secondary/leader swapping) are also logged with the INFO log level...

```
2019-08-23 05:50:08,234 INFO: Lock owner: ask-pg-1; I am ask-pg-0
2019-08-23 05:50:08,234 INFO: does not have lock
2019-08-23 05:50:08,253 INFO: no action.  i am a secondary and i am following a leader
2019-08-23 05:50:18,183 INFO: Lock owner: ask-pg-1; I am ask-pg-0
2019-08-23 05:50:18,183 INFO: does not have lock
2019-08-23 05:50:18,211 INFO: no action.  i am a secondary and i am following a leader
```

mcassaniti:
There's a few issues floating around about the verbosity of logged messages that start with 'no action'. PR #2161 attempts to resolve this by reducing how often the messages are printed. This PR instead changes the severity level which is easy enough to reconfigure.

This should fix #1154, #1800 and #621. It would also help with #1203.

### DIFF
--- a/patroni/__main__.py
+++ b/patroni/__main__.py
@@ -105,7 +105,11 @@ class Patroni(AbstractPatroniDaemon):
         super(Patroni, self).run()
 
     def _run_cycle(self):
-        logger.info(self.ha.run_cycle())
+        msg = self.ha.run_cycle()
+        if msg.startswith('no action'):
+            logger.debug(msg)
+        else:
+            logger.info(msg)
 
         if self.dcs.cluster and self.dcs.cluster.config and self.dcs.cluster.config.data \
                 and self.config.set_dynamic_configuration(self.dcs.cluster.config):


CyberDem0n:
Sorry, but just silencing log messages will not be accepted.

mcassaniti:
> Sorry, but just silencing log messages will not be accepted.

Correct me if I am wrong, but the logging level can be changed from info to debug as part of the configuration. When changed the output will be logged. I certainly don't want to drop these log messages always, but instead to configure when they are produced

CyberDem0n:
The DEBUG level is too verbose.
The number of people who don't have any issues with seeing heartbeat messages every 10 seconds is certainly much higher than people who would like to run Patroni with log.level=DEBUG.

mcassaniti:
OK, when should the logs actually show heartbeat messages? From the issues I linked and the number of comments on each it seems logging these by default is not what users wish to see.

How many users wish to see heartbeat messages every 10 seconds under the default configuration? That is a different question than how many users are indifferent.

CyberDem0n:
> From the issues I linked and the number of comments on each it seems logging these by default is not what users wish to see.

Issues don't always reflect the real world. The total amount of Patroni users is much higher than the number of users opening/commenting issues.

mcassaniti:
@CyberDem0n Just nudging this one again. Would a non-default configuration option be accepted that silenced any 'no action' messages? If so I'm happy to write a PR and we can nut it out.

Issue #2449, opened 2022-11-10
consul token rotation causes catalog registry errors


johnalotoski:
**Describe the bug**
In the case that a consul DCS is being used and consul tokens that patroni uses are rotated, when patroni receives a new token via the configuration file and is signaled to reload configuration, the service is not updated with the new token and after the old token expires, consul will begin logging catalog registration errors:
```
[ERROR] agent.client: RPC failed to server: method=Catalog.Register server=$IP:8300 error="rpc error making call: rpc error making call: ACL not found"
```

**To Reproduce**
Implement a consul token rotation and reload mechanism; not necessarily trivial, example code [here](https://github.com/input-output-hk/bitte-cells/blob/c429ee88509702123c7e24c1f1317da5c3936eff/cells/patroni/entrypoints/default.nix).

**Expected behavior**
Consul token rotation with a corresponding service update for the newly rotated and valid token so that consul catalog updates continued to be handled properly.

**Environment**
- Patroni version: `2.1.4`
- PostgreSQL version: `12`
- DCS (and its version): `consul 1.13.2`

**Additional context**
The patroni configuration file is being updated periodically with a new consul token as the old consul token approaches end of life.  Patroni is then SIGHUP signaled to reload the configuration file with fresh new token.
```
...
consul:
  ...
  token: <TOKEN>
...
```

PR with suggested fix below.  With this PR in our deployment the issue is resolved.

PR #2450, opened 2022-11-10
fix: update service on consul token rotation

johnalotoski:
* Triggers a consul service update upon consul token rotation
* Addresses issue #2449

### DIFF
--- a/patroni/dcs/consul.py
+++ b/patroni/dcs/consul.py
@@ -236,6 +236,7 @@ class Consul(AbstractDCS):
         self._service_check_tls_server_name = config.get('service_check_tls_server_name', None)
         if not self._ctl:
             self.create_session()
+        self._previous_loop_token = self._client.token
 
     def retry(self, *args, **kwargs):
         return self._retry.copy()(*args, **kwargs)
@@ -464,6 +465,7 @@ class Consul(AbstractDCS):
         tags = self._service_tags[:]
         tags.append(role)
         self._previous_loop_service_tags = self._service_tags
+        self._previous_loop_token = self._client.token
 
         params = {
             'service_id': '{0}/{1}'.format(self._scope, self._name),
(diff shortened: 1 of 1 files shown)


patroni/dcs/consul.py
```diff
@@ -236,6 +236,8 @@ def __init__(self, config):
         self._service_check_tls_server_name = config.get('service_check_tls_server_name', None)
         if not self._ctl:
             self.create_session()
+        self._token = kwargs.get('token')
```
CyberDem0n, line 239: We already have `self._client.token`

patroni/dcs/consul.py
```diff
@@ -472,6 +476,7 @@ def _update_service(self, data):
             'check': check,
             'tags': tags,
             'enable_tag_override': True,
+            'token': self._token,
```
CyberDem0n, line 479: https://developer.hashicorp.com/consul/api-docs/api-structure#authentication
> Previously this was provided via a `?token=` query parameter. This functionality exists on many endpoints for backwards compatibility, but its use is highly discouraged, since it can show up in access logs as part of the URL.

Patroni is already using `X-Consul-Token` header for all HTTP requests: https://github.com/zalando/patroni/blob/ef2922fe37206ac376abb87d53b5a01d4e3b57c8/patroni/dcs/consul.py#L120-L122
E.g., there is no need to explicity add the token parameter. We only need to call register_service() if the token has changed. Also it will make https://github.com/zalando/patroni/pull/2450/files#diff-99372e752231ac1fddf9cda076dd985f0ed8293e8ba1c06ea1d95a0530ccb959R446 unnecessary.

CyberDem0n (changes requested):

patroni/dcs/consul.py
johnalotoski: Fixed

patroni/dcs/consul.py
johnalotoski: Ah, good point.  Ok, I think I've fixed it up the way you are suggesting.  Let me know if I've missed anything.  Thanks.

hughcapet:
👍

CyberDem0n:
👍

CyberDem0n (approved):

PR #2465, opened 2022-11-28
Switch from localkube to kind and/or k3d

CyberDem0n:
The only advantage of localkube was being a low weight. Anything else started creating only problems:
1. It is not properly maintained for many years.
2. It effectively worked only on Linux, but stopped on modern version due to changes in iptables.

Instead, we will use widely adoped tools like kind or k3s. The "kind-kind" is the default K8s context (see ~/.kube/config), but it could be overriden using `PATRONI_KUBERNETES_CONTEXT` environment variable. When executed from GH actions the context is set to k3d-k3s-default, because K3s is much faster to start.


tests/test_kubernetes.py
```diff
@@ -120,6 +121,19 @@ def test_load_kube_config(self):
         with patch.object(builtins, 'open', mock_open(read_data=json.dumps(config))):
             k8s_config.load_kube_config()
             self.assertEqual(k8s_config.headers.get('authorization'), 'Bearer token')
+        config["users"][0]["user"]["client-key-data"] = base64.b64encode(b'foobar').decode('utf-8')
```
hughcapet, line 124:
```suggestion

        config["users"][0]["user"]["client-key-data"] = base64.b64encode(b'foobar').decode('utf-8')
```

patroni/dcs/kubernetes.py
```diff
@@ -28,12 +31,34 @@
...
+    __temp_files = []
+
+
+def _create_temp_file(content):
+    if len(__temp_files) == 0:
+        atexit.register(_cleanup_temp_files)
+
+    fd, name = tempfile.mkstemp()
+    os.write(fd, content)
+    os.close(fd)
```
hughcapet, line 57: shouldn't we close `fd` if write failed?

hughcapet: possible errors in os calls are maybe good additional test cases:)

CyberDem0n: If one of those fails that would mean something is terribly wrong and we don't want to run.

hughcapet:
👍

CyberDem0n:
👍

Issue #2466, opened 2022-11-29
Exception with the API after 2.1.5 upgrade


aurelien-reeves-scalingo:
**Describe the bug**
Right after upgraded Patroni to 2.1.5, the logs started to show an exception related to the API:

```console
2022-11-29 15:54:44.027356116 +0100 CET [postgresql-1] Traceback (most recent call last):
2022-11-29 15:54:44.028621535 +0100 CET [postgresql-1] File "/usr/local/lib/python3.9/dist-packages/patroni/api.py", line 844, in process_request_thread
2022-11-29 15:54:44.027211556 +0100 CET [postgresql-1] Exception in thread Thread-25:
2022-11-29 15:54:44.027359108 +0100 CET [postgresql-1] File "/usr/lib/python3.9/threading.py", line 954, in _bootstrap_inner
2022-11-29 15:54:44.028881590 +0100 CET [postgresql-1] File "/usr/lib/python3.9/socketserver.py", line 654, in process_request_thread
2022-11-29 15:54:44.028614950 +0100 CET [postgresql-1] self._target(*self._args, **self._kwargs)
2022-11-29 15:54:44.028383168 +0100 CET [postgresql-1] File "/usr/lib/python3.9/threading.py", line 892, in run
2022-11-29 15:54:44.028825824 +0100 CET [postgresql-1] super(RestApiServer, self).process_request_thread(request, client_address)
2022-11-29 15:54:44.029650842 +0100 CET [postgresql-1] self.shutdown_request(request)
2022-11-29 15:54:44.028376841 +0100 CET [postgresql-1] self.run()
2022-11-29 15:54:44.029692733 +0100 CET [postgresql-1] File "/usr/local/lib/python3.9/dist-packages/patroni/api.py", line 848, in shutdown_request
2022-11-29 15:54:44.029946159 +0100 CET [postgresql-1] File "/usr/lib/python3.9/ssl.py", line 1285, in unwrap
2022-11-29 15:54:44.030793778 +0100 CET [postgresql-1] ConnectionResetError: [Errno 104] Connection reset by peer
2022-11-29 15:54:44.029896642 +0100 CET [postgresql-1] request.unwrap()
2022-11-29 15:54:44.030790053 +0100 CET [postgresql-1] s = self._sslobj.shutdown()
2022-11-29 15:54:56.039742434 +0100 CET [postgresql-1] Exception in thread Thread-28:
[...]

PR #2468, opened 2022-11-30
Fix annoying exceptions on ssl socket shutdown

CyberDem0n:
The HAProxy is closing connections as soon as it got the HTTP Status code leaving no time for Patroni to properly shutdown SSL connection.

Close https://github.com/zalando/patroni/issues/2466


hughcapet:
👍

PR #2471, opened 2022-12-01
fix: only keep latest failed data directory

### LINKED ISSUE #1421: Standby failed bootstrap fills up disk with timestamped directories
Hi,

We've got a standby cluster where we're clearly doing something wrong in our bootstrap script (a wal-g hack of your wal-e one), but on failure the postgres/current directory is renamed with a timestamp, and the process starts again.  This soon fills up a disk.

Is there any reason not to just rename once to `current.failed` as if this is a repeated problem then only the last failure is really relevant?

Thanks.

walbertus:
Use constant postfix when moving data directory due to failure so it only keeps data from the latest failure.
I need some input, especially on the naming convention.

Close #1421

### DIFF
--- a/patroni/postgresql/__init__.py
+++ b/patroni/postgresql/__init__.py
@@ -1061,7 +1061,7 @@ class Postgresql(object):
                     logger.info('renaming WAL directory and updating symlink: %s', pg_wal_realpath)
                     new_name = '{0}.{1}'.format(pg_wal_realpath, postfix)
                     if os.path.exists(new_name):
-                        os.remove(new_name)
+                        shutil.rmtree(new_name)
                     os.rename(pg_wal_realpath, new_name)
                     os.unlink(source)
                     os.symlink(new_name, source)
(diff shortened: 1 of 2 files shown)


patroni/postgresql/__init__.py
```diff
@@ -1023,27 +1023,33 @@ def pg_tblspc_realpaths(self):
...
+                postfix = 'failed'
 
                 # let's see if the wal directory is a symlink, in this case we
                 # should move the target
                 for (source, pg_wal_realpath) in self.pg_wal_realpath().items():
                     logger.info('renaming WAL directory and updating symlink: %s', pg_wal_realpath)
-                    new_name = '{0}_{1}'.format(pg_wal_realpath, postfix)
+                    new_name = '{0}.{1}'.format(pg_wal_realpath, postfix)
+                    if os.path.exists(new_name):
+                        os.remove(new_name)
```
CyberDem0n, line 1034: We better use `shutil.rmtree()` instead of `os.remove()`.

walbertus: updated using `shutil.rmtree`

walbertus:
sorry for the late update, added the test to cover the new lines

walbertus:
updated the branch to include codacy fix commit

hughcapet:
👍

CyberDem0n:
👍

hughcapet:
Merged, thank you, @walbertus!

PR #2475, opened 2022-12-06
Remove patronictl configure

hughcapet:
Remove `patronictl configure` command
\+ Change name of the "secret" ENV variable and the corresponding patronictl option (to avoid mixing it up with the one from tests)


CyberDem0n:
👍

hughcapet:
👍

PR #2476, opened 2022-12-06
Improve behaviour of the insecure option

CyberDem0n:
It didn't worked correctly when client certificates are used for REST API requests.


hughcapet:
👍

PR #2484, opened 2022-12-13
Raise DCSError when communication with DCS fails

CyberDem0n:
Previously such an exception was raised only from the `get_cluster()` method, and now we will to do the same from the `update_leader()` and `attempt_to_acquire_leader()` methods.

These methods influence Postgres promotion and demotion and we want to make a difference between different types of failures. Specifically, if calls have failed because DCS isn't accessible or due to a timeout.

This commit is extracted from the #2379


hughcapet:
👍

CyberDem0n:
👍

PR #2489, opened 2022-12-14
Adjust Dockerfile for arm64

hughcapet:
- Remove explicit amd64/x86_64
- Add `/lib/$arch-linux-gnu/libnss_files.so.* ` to excludes


CyberDem0n:
👍

hughcapet:
👍

PR #2491, opened 2022-12-16
Prevent pg_stat_statements from recording secrets

feikesteenbergen:
pg_stat_statements is enabled by many by default, but will by default also track utility commands including an

        ALTER USER john WITH PASSWORD 's3cret'

We can prevent this leaking by ensuring that our session currently does not track utility commands when running a sensitive query.

Local testing shows that this command works fine, even for those that do not have `pg_stat_statements` configured in their
`shared_preload_libraries`.


manute (approved):

hughcapet:
👍

CyberDem0n:
👍

CyberDem0n:
Merged, thank you @feikesteenbergen!

PR #2496, opened 2022-12-19
Enforce search_path=pg_catalog for non-replication connections

CyberDem0n:
There is a known [vector of attact](https://pganalyze.com/blog/5mins-postgres-security-patch-releases-pgspot-pghostile) by creating functions and/or operators in a public scheme with the same name and signature as corresponding objects in `pg_catalog`.

Since Patroni is heavily relying on superuser connections we want to mitigate it by enforcing `search_path=pg_catalog` for all connections created by Patroni (except replication connections). It is achieved by introducing a new function, that wraps psycopg.connect() and appends ` -c search_path=pg_catalog` to `options` parameter.

In addition to that, we set connection.autocommit to True before returning it.

### DIFF
--- a/features/environment.py
+++ b/features/environment.py
@@ -177,7 +177,7 @@ class PatroniController(AbstractController):
         patroni_config_name = self.PATRONI_CONFIG.format(name)
         patroni_config_path = os.path.join(self._output_dir, patroni_config_name)
 
-        with open(patroni_config_name) as f:
+        with open('postgres0.yml') as f:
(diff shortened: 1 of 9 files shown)


hughcapet:
Should it actually also be the case for `patronictl query`? If yes, seems it is worth mentioning in docs/command description

hughcapet:
👍

CyberDem0n:
👍

PR #2498, opened 2022-12-21
Enable SSL in REST API and Postgres if possible when running behave

CyberDem0n:
If openssl binary is available use it to generate a self-signed certificate. Use it to protect Patroni REST API
(`verify_client: required`).

In case if Postgres is compiled with SSL support enable it in the configuration and configure pg_hba.conf to check client certificates (`verify-ca`) in addition to passwords. Also configure superuser/replication/rewind users to use client certificates and verify server certificate (`verify-ca`)


hughcapet:
👍

CyberDem0n:
👍

PR #2501, opened 2022-12-27
The Consul TTL is off by twice from reality

CyberDem0n:
we use `ttl/2.0` when setting the value on the HTTPClient, but forgot to multiply the current value by 2.


hughcapet:
can we have this description as the code comment as well? 🥺

hughcapet:
👍

CyberDem0n:
👍

Issue #2499, opened 2022-12-23
Should active unrecognized slots throw an error each HA cycle?


mbanck:
**Describe the bug**
If you create an (external) slot and use it (e.g. with `pg_recvlogical`), Patroni runs fine for now.

Once you put the cluster in pause mode, restart Patroni on the leader, and unpause the cluster, the error `ERROR: Failed to drop replication slot 'slot_recvlogical` is logged for each HA cycle.

First off, it's a bit unclear why this doesn't happen immediately, is that a bug/consistency issue in itself?

Also, this is an active slot that probably has some purpose; I understand that from Patroni's logic this slot isn't recognized, but logging this at `ERROR` level each HA cycle makes it difficult to maintain a strict "no errors in log" policy.

On the other hand, if you add those slots to `ignore_slots`, Patroni won't drop them once they are inactive (after a switchover or a demote) and they hang around with all the problems that makes.

So I think it might be better to log this once at `ERROR` (or `WARNING`?) level and then only at `DEBUG` at most? Maybe I am missing something which makes it important to have this logged as error each time?

One alternative would be to register those slots properly (and let Patroni do logical slot failover), but if those slots are manually maintained externally that might not be an option.

**To Reproduce**

1. on follower (or elsewhere), run `pg_recvlogical -h pg3  -S slot_recvlogical -f archive_logical -U $REPL_USER -d postgres --create-slot --start`
2. Verify that no errors are logged waiting for one HA cycle
3. put patroni in maintenance mode with `patronictl pause`
4. restart patroni on leader
5. put patroni back in production mode with `patronictl resume`
6. now I see the `ERROR`
7. Add `ignore_slots` via `patronictl edit-config`:
```
ignore_slots:
- database: postgres
  name: pg1_recvlogical
  type: logical
```
8. Error is gone
9. Do a switchover with `patronictl switchover`
10. Verify that slot is still around and inactive via `SELECT * FROM pg_replication_slots;` on old leader

**Environment**
- Patroni version: 2.1.5
- PostgreSQL version: n/a
- DCS (and its version): n/a

CyberDem0n:
If the slot is active there is not danger that it will result in pg_wal growth. The error/warning for active slots isn't reported just not to spam logs and to avoid some other potential problems: https://github.com/zalando/patroni/pull/266

[...]

PR #2502, opened 2022-12-27
Suppress recurring errors when dropping unknown but active replication slots

mbanck:
When a replication slot is not registered with Patroni but is active, Patroni would log an error during each HA cycle in certain conditions (after a restart or role change). To avoid this, first check if the replication slot we are about to drop is still active and if so, only log a warning. Otherwise, log the slot we are dropping for informational purposes.

Close: #2499

The warning is only logged once according to my tests (unlike the error, which was carried over for each HA cycle).

### DIFF
--- a/patroni/postgresql/slots.py
+++ b/patroni/postgresql/slots.py
@@ -200,10 +200,18 @@ class SlotsHandler(object):
     def _drop_incorrect_slots(self, cluster, slots, paused):
         # drop old replication slots which are not presented in desired slots
         for name in set(self._replication_slots) - set(slots):
-            if not paused and not self.ignore_replication_slot(cluster, name) and not self.drop_replication_slot(name):
-                logger.error("Failed to drop replication slot '%s'", name)
-                self._schedule_load_slots = True
-
+            if not paused and not self.ignore_replication_slot(cluster, name):
+                if not self.drop_replication_slot(name):
+                    self._schedule_load_slots = True
+                    # Check if slot is still active, this is not considered an error
+                    cursor = self._query(('SELECT 1 FROM pg_catalog.pg_replication_slots WHERE ' +
+                                          'slot_name = %s AND active'), name)
+                    if (cursor.rowcount == 1):
+                        logger.debug("Unable to drop unknown replication slot '%s', slot is still active", name)
+                    else:
+                        logger.error("Failed to drop replication slot '%s'", name)
+                else:
+                    logger.info("Dropped unknown replication slot '%s'", name)
         for name, value in slots.items():
             if name in self._replication_slots and not compare_slots(value, self._replication_slots[name]):
                 logger.info("Trying to drop replication slot '%s' because value is changing from %s to %s",


patroni/postgresql/slots.py
```diff
@@ -192,10 +192,19 @@ def ignore_replication_slot(self, cluster, name):
...
-                              'FROM pg_catalog.pg_replication_slots WHERE slot_name = %s AND NOT active)'), name, name)
-        # In normal situation rowcount should be 1, otherwise either slot doesn't exists or it is still active
-        return cursor.rowcount == 1
+        # First check if the replication slot is still active; not being able
+        # to drop it in this case is not considered an error
+        cursor_slot_active = self._query(('SELECT 1 FROM pg_catalog.pg_replication_slots WHERE ' +
+                                          'slot_name = %s AND active'), name)
+        if (cursor_slot_active.rowcount == 1):
+            logger.warning("Unable to drop unknown replication slot '%s', slot is still active", name)
+            return True
```
CyberDem0n, line 201: Since the method returns True from here there will be no further attempt to remove the slot.

mbanck: oh right I see - if the slots becomes inactive, we'd never drop it.

So let's get back to logging this at debug each cycle, will revise

mbanck: I rearranged this now, I hope this is better

patroni/postgresql/slots.py
```diff
@@ -201,8 +201,17 @@ def _drop_incorrect_slots(self, cluster, slots, paused):
...
         for name in set(self._replication_slots) - set(slots):
             if not paused and not self.ignore_replication_slot(cluster, name) and not self.drop_replication_slot(name):
-                logger.error("Failed to drop replication slot '%s'", name)
                 self._schedule_load_slots = True
+                # Check if slot is still active, this is not considered an error
+                cursor = self._query(('SELECT 1 FROM pg_catalog.pg_replication_slots WHERE ' +
+                                      'slot_name = %s AND active'), name)
+                if (cursor.rowcount == 1):
+                    logger.debug("Unable to drop unknown replication slot '%s', slot is still active", name)
+                    return True
```
CyberDem0n, line 210: If we return from here it will prevent other slots from removal...

mbanck: Yeah, that was a copy-pasto I saw independently and have now removed

patroni/postgresql/slots.py
```diff
@@ -201,8 +201,16 @@ def _drop_incorrect_slots(self, cluster, slots, paused):
...
                 self._schedule_load_slots = True
+                # Check if slot is still active, this is not considered an error
+                cursor = self._query(('SELECT 1 FROM pg_catalog.pg_replication_slots WHERE ' +
+                                      'slot_name = %s AND active'), name)
+                if (cursor.rowcount == 1):
+                    logger.debug("Unable to drop unknown replication slot '%s', slot is still active", name)
+                else:
+                    logger.error("Failed to drop replication slot '%s'", name)
+            else:
+                logger.info("Dropped unknown replication slot '%s'", name)
```
hughcapet, line 213: I'm afraid, this will be erroneously logged for 'paused' mode or an ignored slot as well

CyberDem0n: @mbanck please take a look

mbanck: You're right, that was a thinko; I've pushed a new version now

hughcapet:
👍

CyberDem0n:
:+1:

CyberDem0n:
Merged, thank you @mbanck!

PR #2503, opened 2022-12-28
Do not rely on 'role' value when checking other nodes via REST API

CyberDem0n:
When doing the leader race we need to check that the former primary isn't alive anymore. For that we relied on non-inclusive terms. In order to simplify future work on getting rid from all non-inclusive words we change the check to rely on a difference in format of wal/xlog field. There is only "location" for the primary and "replayed_location" + "received_location" for standbys.

In addition to that we start supporting "wal" field as well as deprecated "xlog".

### DIFF
--- a/patroni/ha.py
+++ b/patroni/ha.py
@@ -39,6 +39,11 @@ class _MemberStatus(namedtuple('_MemberStatus', ['member', 'reachable', 'in_reco
     """
     @classmethod
     def from_api_response(cls, member, json):
+        """
+        :param member: dcs.Member object
+        :param json: RestApiHandler.get_postgresql_status() result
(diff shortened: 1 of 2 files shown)


patroni/ha.py
```diff
@@ -39,11 +39,15 @@ class _MemberStatus(namedtuple('_MemberStatus', ['member', 'reachable', 'in_reco
     """
     @classmethod
     def from_api_response(cls, member, json):
```
hughcapet, line 41: such a nice occasion to add docsting with input/output params description 😈

hughcapet:
👍

CyberDem0n:
👍

PR #2504, opened 2022-12-29
Citus integration

CyberDem0n:
Citus cluster (coordinator and workers) will be stored in DCS as a fleet of Patroni logically grouped together:
```
/service/batman/
/service/batman/0/
/service/batman/0/initialize
/service/batman/0/leader
/service/batman/0/members/
/service/batman/0/members/m1
/service/batman/0/members/m2
/service/batman/
/service/batman/1/
/service/batman/1/initialize
/service/batman/1/leader
/service/batman/1/members/
/service/batman/1/members/m1
/service/batman/1/members/m2
...
```

Where 0 is a Citus group for coordinator and 1, 2, etc are worker groups.

Such hierarchy allows reading the entire Citus cluster with a single call to DCS (except Zookeeper).

The get_cluster() method will be reading the entire Citus cluster on the coordinator because it needs to discover workers. For the worker cluster it will be reading the subtree of its own group.

Besides that we introduce a new method  get_citus_coordinator(). It will be used only by worker clusters.

Since there is no hierarchical structures on K8s we will use the citus group suffix on all objects that Patroni creates.
E.g.
```
batman-0-leader  # the leader config map for the coordinator
batman-0-config  # the config map holding initialize, config, and history "keys"
...
batman-1-leader  # the leader config map for worker group 1
batman-1-config
...
```

Citus integration is enabled from patroni.yaml:
```yaml
citus:
  database: citus
  group: 0  # 0 is for coordinator, 1, 2, etc are for workers
```

If enabled, Patroni will create the database, citus extension in it, and INSERTs INTO `pg_dist_authinfo` information required for Citus nodes to communicate between each other, i.e. 'password', 'sslcert', 'sslkey' for superuser if they are defined in the Patroni configuration file.

When the new Citus coordinator/worker is bootstrapped, Patroni adds `synchronous_mode: on` to the `bootstrap.dcs` section.

Besides that, Patroni takes over management of some Postgres GUCs:
- `shared_preload_libraries` - Patroni ensures that the "citus" is added to the first place
- `max_prepared_transactions` - if not set or set to 0, Patroni changes the value to `max_connections*2`
- wal_level - automatically set to logical. It is used by Citus to move/split shards. Under the hood Citus is creating/removing replication slots and they are automatically added by Patroni to the `ignore_slots` configuration to avoid accidental removal.

The coordinator primary actively discovers worker primary nodes and registers/updates them in the `pg_dist_node` table using
citus_add_node() and citus_update_node() functions.

Patroni running on the coordinator provides the new REST API endpoint: `POST /citus`. It is used by workers to facilitate controlled switchovers and restarts of worker primaries.
When the worker primary needs to shut down Postgres because of restart or switchover, it calls the `POST /citus` endpoint on the coordinator and the Patroni on the coordinator starts a transaction and calls `citus_update_node(nodeid, 'host-demoted', port)` in order to pause client connections that work with the given worker.
Once the new leader is elected or postgres started back, they perform another call to the `POST/citus` endpoint, that does another `citus_update_node()` call with actual hostname and port and commits a transaction. After transaction is committed, coordinator reestablishes connections to the worker node and client connections are unblocked.
If clients don't run long transaction the operation finishes without client visible errors, but only a short latency spike.

All operations on the `pg_dist_node` are serialized by Patroni on the coordinator. It allows to have more control and ROLLBACK transaction in progress if its lifetime exceeding a certain threshold and there are other worker nodes should be updated.


docs/citus.rst
```diff
@@ -0,0 +1,334 @@
...
+
+1. The ``list`` and the ``topology`` by default output all members of the Citus
+   formation (coordinators and workers). The new column ``Group`` indicates
+   which Citus group they belong to.
+2. For all ``patronictl`` commands the new option is introduced, named
+   ``--group``. For some commands the default value for the group might be
+   taken from the ``patroni.yaml``. For example, ``patronictl pause`` will
+   enable the maintenance mode by default for the ``group`` that is set in the
+   ``citus`` section, but for example for ``patronictl  switchover`` or
+   ``patronictl remove`` the group must be explicitly specified.
```
JelteF, line 66: What's the reasoning for this difference in behavior with regards to patroni.yaml between commands?

CyberDem0n: It is a long story.
Originally every command required explicit cluster name to be specified in the command-line, because it was supposed to work for an arbitrary Patroni clusters registered in DCS.

But, it turned out that people mostly run `patronictl` on the same host where Patroni runs.
Providing the cluster name was very annoying (especially knowing that it is already in the config file), therefore for most commands we made the cluster name optional. If the cluster name isn't supplied, it is picked from the config file.
Known exceptions to this rule are:
1. commands which are followed by a member name, e.g.: `patronictl reinit $CLUSTER_NAME $MEMBER_NAME`.
2. "dangerous" command: `patronictl remove $CLUSTER_NAME`. I call in dangerous because it wipes all keys for a given cluster from DCS.

Since a Citus cluster is effectively just a bunch of different Patroni clusters logically connected using the groupid I had to introduce the `--group` option for all commands. And now we have more or less the same story.
1. For `patronictl list` we probably don't want to use the default group value from the config because typically users want to see all coordinator and worker nodes as a whole.
2. `patronictl remove` command still remains dangerous, and since we require the cluster name to be explicitly specified we have to do the same for the `--group`. Actually, the `remove` command has only an interactive mode and asks to confirm the cluster name. Probably we should do the same for the group.
3. `patronictl switchover/failover` -- the workflow of these commands implies that it works only for a single Patroni cluster, therefore the group must be specified either explicitly from the command line, or it will ask for while working in the interactive mode.

For remaining commands the workflow will stay as it is. Lets say you investigate the problem with one of the workers. In this case most likely you will be on the host where the worker is running and want `patronictl` do some actions (edit-config/show-config/pause/resume) for a given group.

patroni/postgresql/citus.py
```diff
@@ -0,0 +1,386 @@
...
+import time
+
+from six.moves.urllib_parse import urlparse
+from threading import Condition, Event, Thread
+
+from .connection import Connection
+from ..dcs import CITUS_COORDINATOR_GROUP_ID
+from ..psycopg import connect, quote_ident
+
+CITUS_SLOT_NAME_RE = re.compile(r'^citus_shard_(move|split)_slot(_[1-9][0-9]*){3}$')
```
JelteF, line 12: The third number is only added in the current main branch of citus. To be compatible with citus 11.1, it should also match slots with 2 numbers.
```suggestion
CITUS_SLOT_NAME_RE = re.compile(r'^citus_shard_(move|split)_slot(_[1-9][0-9]*){2,3}$')
```

patroni/postgresql/citus.py
```diff
@@ -0,0 +1,386 @@
...
+        This method returns `True` if node was updated (optionally,
+        transaction was committed) as an indicator that
+        the `self._pg_dist_node` cache should be updated.
+
+        The maximum lifetime of the transaction in progress
+        is controlled outside of this method."""
+
+        if task.event == 'after_promote':
+            if not self._in_flight or self._in_flight.host != task.host or self._in_flight.port != task.port:
+                self.update_node(task)
```
JelteF, line 229: When reading this condition I was quite confused when it would return `True`: `self._in_flight.host != task.host or self._in_flight.port != task.port`

After reading the code more I think that I understand that there are two cases this is supposed to handle:
1. it's a convoluted way of checking if `self._in_flight` was a before_demote event.
2. If somehow the promotion failed, then a new election could happen and the target of after_promote could be a different node than the one from before_promote.

I think some code comments are definitely deserving here.

features/patroni_api.feature
```diff
@@ -14,9 +14,9 @@ Scenario: check API requests on a stand-alone server
 	Then I receive a response code 200
 	When I issue a GET request to http://127.0.0.1:8008/replica
 	Then I receive a response code 503
-	When I run patronictl.py reinit batman postgres0 --force
-	Then I receive a response returncode 0
-	And I receive a response output "Failed: reinitialize for member postgres0, status code=503, (I am the leader, can not reinitialize)"
+	When I issue a POST request to http://127.0.0.1:8008/reinitialize with {"force": true}
+	Then I receive a response code 503
+	And I receive a response text I am the leader, can not reinitialize
```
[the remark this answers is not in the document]
CyberDem0n: Yes, it was necessary.
I did some refactoring of patronictl while introducing `--group` option and now it doesn't allow to call `POST /reinitialize` if member is a leader.
Since this behave test is about REST API I decided to call it directly, without patronictl.

features/steps/basic_replication.py
```diff
@@ -29,8 +29,12 @@ def add_table(context, table_name, pg_name):
     # parse the configuration file and get the port
     try:
         context.pctl.query(pg_name, "CREATE TABLE public.{0}()".format(table_name))
-    except pg.Error as e:
-        assert False, "Error creating table {0} on {1}: {2}".format(table_name, pg_name, e)
+    except pg.Error:
+        try:
+            context.pctl.query(pg_name, "CREATE TABLE public.{0}()".format(table_name))
```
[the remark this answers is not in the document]
CyberDem0n: I will probably change it back.
After switchover the connection is closed and exception is raised.
Now there is `{name:w} role is the primary after {timeout:d} seconds` is all required places and it can handle exception and repoen connection.

features/steps/basic_replication.py
```diff
@@ -29,8 +29,12 @@ def add_table(context, table_name, pg_name):
...
+        except pg.Error as e:
+
+            assert False, "Error creating table {0} on {1}: {2}".format(table_name, pg_name, e)
```
CyberDem0n, line 37:
```suggestion
    except pg.Error as e:
        assert False, "Error creating table {0} on {1}: {2}".format(table_name, pg_name, e)
```

patroni/postgresql/citus.py
CyberDem0n: Sure, I will add a comment

hughcapet:
👍

hughcapet:
👍

CyberDem0n:
:+1:

PR #2509, opened 2023-01-02
Start etcd with V2 API enabled for V2 etcd acceptance tests

mbanck:
Otherwise, the etcd (not etcd3) behave tests fail to connect:
```
Jan 02 09:56:18 HOOK-ERROR in before_all: AssertionError: etcd instance is not available for queries after 5 seconds
```


hughcapet:
👍

CyberDem0n:
👍

PR #2511, opened 2023-01-02
Bump postgres version to 11 when running  github action tests

### LINKED ISSUE #2510: We should bump the PG version to 11 in `mapping.py`
When running tests when pulling changes, some tests are failing due to packages no longer being available. This is because the tests try to install 9.6 when the DCS is etcd. We should stick with the oldest supported version of Postgres, in this case PG 11.

martinmarques:
When running the github tests, versions of postgres are taken from the `version` dictionary in `mapping.py`.

The version for DCS `etcd` is to use postgres 9.6. Sadly, this version is out of support and not available in some distribution repositories (we'd need to use archived repositories to get older versions of postgres).

As stated above, 9.6, and also 10, are not supported anymore so it's much better to test on supported software.

The commit pushes the version to be used in tests wity etcd to postgres 11, and closes #2510.

Signed-off-by: Martín Marqués <martin.marques@enterprisedb.com>

### DIFF
--- a/.github/workflows/mapping.py
+++ b/.github/workflows/mapping.py
@@ -1,1 +1,1 @@
-versions = {'etcd': '9.6', 'etcd3': '14', 'consul': '13', 'exhibitor': '12', 'raft': '11', 'kubernetes': '15'}
+versions = {'etcd': '11', 'etcd3': '14', 'consul': '13', 'exhibitor': '12', 'raft': '11', 'kubernetes': '15'}


CyberDem0n:
Everything that is inside `.github/workflows` is used only for GH actions and there is no problem installing 9.6.
When running tests locally or using your own CI you don't have to use run_tests.py. The typical workflow is descibed in the https://github.com/zalando/patroni/blob/master/docs/CONTRIBUTING.rst

Now regarding testing or not nesting with 9.6: in the documentation we write "Currently supported PostgreSQL versions: 9.3 to 15", hence IMO it would be nice to test at least with one EOL version.

martinmarques:
Ignore this. I missunderstood the failure I was seeing, which was from another GH action in my internal repo (@matthbakeredb shed light on it so I could see what was really wrong)

We can ignore this PR all together

PR #2515, opened 2023-01-05
Fix behave tests on GH actions MacOS

CyberDem0n:
- the new MacOS doesn't play well with old go binaries (bump etcd)
- use brew to install Postgres and expect (unbuffer, to make behave output colorful) and use the latest version
- upload failed logs instead of grepping them to stdout


hughcapet:
👍

hughcapet:
👍

CyberDem0n:
👍

PR #2516, opened 2023-01-09
Run only one query per HA loop

CyberDem0n:
If the cluster is stable (no nodes are joining/leaving/lagging) we want to run at most one monitor query per every HA loop. So far it worker perfectly except when synchronous_mode is enabled, where we run two additional queries:
1. SHOW synchronous_mode
2. SELECT ... FROM pg_stat_replication

In order to solve it, we will include these "queries" to the common monitoring query is synchronous_mode is enabled.

In addition to that we make sure that synchronous_standby_names is reset on replicas that used to by a primary.

P.S.: in the monitoring query we also extract the current value of synchronous_standby_names, because it will be useful for the quorum commit feature.

### DIFF
--- a/patroni/postgresql/__init__.py
+++ b/patroni/postgresql/__init__.py
@@ -22,6 +22,7 @@ from .connection import Connection, get_connection_cursor
 from .misc import parse_history, parse_lsn, postgres_major_version_to_int
 from .postmaster import PostmasterProcess
 from .slots import SlotsHandler
+from .validator import CaseInsensitiveDict
 from .. import psycopg
 from ..exceptions import PostgresConnectionException
 from ..utils import Retry, RetryFailedError, polling_loop, data_directory_is_empty, parse_int
@@ -106,6 +107,7 @@ class Postgresql(object):
         self._cluster_info_state = {}
         self._has_permanent_logical_slots = True
         self._enforce_hot_standby_feedback = False
+        self._is_synchronous_mode = True
         self._cached_replica_timeline = None
 
         # Last known running process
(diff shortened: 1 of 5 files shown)


patroni/postgresql/__init__.py
```diff
@@ -158,19 +160,30 @@ def lsn_name(self):
 
     @property
     def cluster_info_query(self):
```
hughcapet, line 162: After making this method even more complex we probably need a description here

patroni/postgresql/__init__.py
```diff
@@ -1093,29 +1106,28 @@ def pick_synchronous_standby(self, cluster, sync_node_count=1, sync_node_maxlag=
...
         # Pick candidates based on who has higher replay/remote_write/flush lsn.
-        sync_commit_par = self._get_synchronous_commit_param()
-        sort_col = {'remote_apply': 'replay', 'remote_write': 'write'}.get(sync_commit_par, 'flush')
+        synchronous_commit = self._cluster_info_state_get('synchronous_commit')
+        sort_col = {'remote_apply': 'replay', 'remote_write': 'write'}.get(synchronous_commit, 'flush') + '_lsn'
+        pg_stat_replication = [(r['application_name'], r['sync_state'], r[sort_col])
+                               for r in self._cluster_info_state_get('pg_stat_replication') or []
+                               if r[sort_col] is not None]
         # pg_stat_replication.sync_state has 4 possible states - async, potential, quorum, sync.
         # Sort clause "ORDER BY sync_state DESC" is to get the result in required order and to keep
```
hughcapet, line 1120: seems it also makes sense to adjust the comment

patroni/postgresql/__init__.py
```diff
@@ -1093,29 +1106,28 @@ def pick_synchronous_standby(self, cluster, sync_node_count=1, sync_node_maxlag=
...
-                " FROM pg_catalog.pg_stat_replication"
-                " WHERE state = 'streaming' AND {0}_{1} IS NOT NULL"
-                " ORDER BY sync_state DESC, {0}_{1} DESC".format(sort_col, self.lsn_name, self.wal_name)):
+        for app_name, sync_state, replica_lsn in sorted(pg_stat_replication, key=lambda r: (r[1], r[2]), reverse=True):
             member = members.get(app_name)
-            if member and not member.tags.get('nosync', False):
+            if member and member.is_running and not member.tags.get('nosync', False):
                 replica_list.append((member.name, sync_state, replica_lsn, bool(member.nofailover)))
 
-        max_lsn = max(replica_list, key=lambda x: x[2])[2] if len(replica_list) > 1 else int(str(self.last_operation()))
```
hughcapet, line 1118: can't get it - what has changed so that we don't need `int(str())` anymore?

tests/test_postgresql.py
```diff
@@ -638,44 +638,46 @@ def time_in_state(*args):
...
-        mock_cursor = Mock()
-        mock_cursor.fetchone.return_value = ('remote_apply',)
-
-        with patch.object(Postgresql, "query", side_effect=[
-                    mock_cursor,
-                    [(self.leadermem.name, 'sync', 1),
-                     (self.me.name, 'async', 2),
-                     (self.other.name, 'async', 2)]
+        mock_cluster_info_state_get = Mock()
+        mock_cluster_info_state_get.return_value = 'remote_apply'
```
hughcapet, line 646: is it actually still needed?

patroni/postgresql/__init__.py
CyberDem0n: The last_operation() method used to return str. It was changed in https://github.com/zalando/patroni/commit/935805087f8805a4c9dad3401311e05972ef4f20#diff-72030eab48ce77175c998a2fd1ee9e84fe950e132b66972aa72496532a3923e5R876-R878

hughcapet:
👍

CyberDem0n:
:+1:

PR #2521, opened 2023-01-11
Improve /(a)sync checks in behave tests

CyberDem0n:
They are frequently failing because sometimes replicas are a bit slow realizing that they are synchronous. Instead of instroducing more sleeps we will poll for required http status code with some timeout.

### DIFF
--- a/features/steps/patroni_api.py
+++ b/features/steps/patroni_api.py
@@ -131,6 +131,20 @@ def add_tag_to_config(context, tag, value, pg_name):
     context.pctl.add_tag_to_config(pg_name, tag, value)
 
 
+@then('Status code on GET {url:url} is {code:d} after {timeout:d} seconds')
+def check_http_code(context, url, code, timeout, negate=False):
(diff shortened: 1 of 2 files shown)


features/steps/patroni_api.py
```diff
@@ -131,6 +131,20 @@ def add_tag_to_config(context, tag, value, pg_name):
     context.pctl.add_tag_to_config(pg_name, tag, value)
 
 
+@then('Status code on GET {url:url} is {code:d} after {timeout:d} seconds')
+def check_http_code(context, url, code, timeout, negate=False):
```
hughcapet, line 135: is `negate` accidentally here?

CyberDem0n: copy&paste :facepalm:

hughcapet:
👍

CyberDem0n:
:+

CyberDem0n:
:+1:

PR #2528, opened 2023-01-16
Submit coverage to codacy only if secret is available

CyberDem0n:
If PR is open from the external GH repo secrets are not set due to security reasons. It makes codacy coverage report to fail.

### DIFF
--- a/.github/workflows/tests.yaml
+++ b/.github/workflows/tests.yaml
@@ -8,6 +8,10 @@ on:
     tags:
     - v.*
 
+env:
+  CODACY_PROJECT_TOKEN: ${{ secrets.CODACY_PROJECT_TOKEN }}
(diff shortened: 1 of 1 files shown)


.github/workflows/tests.yaml
```diff
@@ -149,6 +149,8 @@ jobs:
       run: bash <(curl -Ls https://coverage.codacy.com/get.sh) report -r cobertura.xml -l Python --partial
       env:
         CODACY_PROJECT_TOKEN: ${{ secrets.CODACY_PROJECT_TOKEN }}
+        SECRETS_AVAILABLE: ${{ secrets.CODACY_PROJECT_TOKEN != '' }}
```
hughcapet, line 152: should we move them both to the workflow level 🤔

CyberDem0n: Please try

CyberDem0n:
:+1:

hughcapet:
👍

PR #2532, opened 2023-01-19
Fixes and improvements in failsafe

CyberDem0n:
1. Fix problem with logical slots not advancing when only the primary lost access to DCS
2. Don't let Patroni to join as a raft voting member when running failsafe behave tests. It allows to test exactly the same conditions as for other DCS
3. Speed up dcs_failsafe_mode behave tests by getting rid from long sleeps, slight reshuffling of places when we start/stop outage, and by killing Patroni/Postgres to avoid long shutdown due to the leader key removal attempts.

### DIFF
--- a/features/environment.py
+++ b/features/environment.py
@@ -184,6 +184,8 @@ class PatroniController(AbstractController):
             config.pop('etcd', None)
 
         raft_port = os.environ.get('RAFT_PORT')
+        # If patroni_raft_controller is suspended two Patroni members is enough to get a quorum,
+        # therefore we don't want Patroni to join as a voting member when testing dcs_failsafe_mode.
         if raft_port and not self._output_dir.endswith('dcs_failsafe_mode'):
             os.environ['RAFT_PORT'] = str(int(raft_port) + 1)
             config['raft'] = {'data_dir': self._output_dir, 'self_addr': 'localhost:' + os.environ['RAFT_PORT']}
--- a/features/environment.py
+++ b/features/environment.py
@@ -184,7 +184,7 @@ class PatroniController(AbstractController):
             config.pop('etcd', None)
 
         raft_port = os.environ.get('RAFT_PORT')
-        if raft_port:
+        if raft_port and not self._output_dir.endswith('dcs_failsafe_mode'):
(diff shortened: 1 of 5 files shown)


features/environment.py
```diff
@@ -184,7 +184,7 @@ def _make_patroni_test_config(self, name, custom_config):
             config.pop('etcd', None)
 
         raft_port = os.environ.get('RAFT_PORT')
-        if raft_port:
+        if raft_port and not self._output_dir.endswith('dcs_failsafe_mode'):
```
hughcapet, line 187: Can we have the point from the PR desc as a comment here as well?

patroni/ha.py
```diff
@@ -1695,15 +1692,18 @@ def _handle_dcs_error(self):
...
-                if cluster:
-                    self.state_handler.slots_handler.sync_replication_slots(cluster,
-                                                                            self.patroni.nofailover,
-                                                                            self.patroni.replicatefrom,
-                                                                            self.is_paused())
-
+                self._sync_replication_slots()
         return 'DCS is not accessible'
 
+    def _sync_replication_slots(self):
```
hughcapet, line 1698: and the explanation how syncing of replication slots (=this function) works both for failsafe mode enabled and failing dcs (incl. this corner case of readonly dcs) and in the "normal operational mode"? (now, after an hour of despair, I think I got it)

patroni/ha.py
```diff
@@ -1694,15 +1694,33 @@ def _handle_dcs_error(self):
...
+        cluster = self._failsafe.update_cluster(self.cluster)\
+            if self.is_failsafe_mode() and not self.is_leader() else self.cluster
+        if cluster:
+            slots = self.state_handler.slots_handler.sync_replication_slots(cluster,
                                                                             self.patroni.nofailover,
                                                                             self.patroni.replicatefrom,
                                                                             self.is_paused())
-
-        return 'DCS is not accessible'
+            # Don't copy replication slots if failsafe_mode is is active
```
hughcapet, line 1722:
```suggestion
            # Don't copy replication slots if failsafe_mode is active
```

hughcapet:
👍

CyberDem0n:
:+1:

PR #2534, opened 2023-01-20
Refactor drop_replication_slot() and _drop_incorrect_slots()

CyberDem0n:
Use CTE to avoid running the second query if pg_drop_replication_slot() failed

### DIFF
--- a/patroni/postgresql/slots.py
+++ b/patroni/postgresql/slots.py
@@ -179,31 +179,33 @@ class SlotsHandler(object):
         return False
 
     def drop_replication_slot(self, name):
-        cursor = self._query(('SELECT pg_catalog.pg_drop_replication_slot(%s) WHERE EXISTS (SELECT 1 ' +
-                              'FROM pg_catalog.pg_replication_slots WHERE slot_name = %s AND NOT active)'), name, name)
-        # In normal situation rowcount should be 1, otherwise either slot doesn't exists or it is still active
-        return cursor.rowcount == 1
+        """Returns a tuple(active, dropped)"""
+        cursor = self._query(('WITH slots AS (SELECT slot_name, active' +
+                              ' FROM pg_catalog.pg_replication_slots WHERE slot_name = %s),' +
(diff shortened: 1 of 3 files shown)


patroni/postgresql/slots.py
```diff
@@ -179,31 +179,33 @@ def ignore_replication_slot(self, cluster, name):
...
-                    logger.info("Dropped unknown replication slot '%s'", name)
         for name, value in slots.items():
             if name in self._replication_slots and not compare_slots(value, self._replication_slots[name]):
                 logger.info("Trying to drop replication slot '%s' because value is changing from %s to %s",
                             name, self._replication_slots[name], value)
-                if self.drop_replication_slot(name):
+                if self.drop_replication_slot(name) == (False, True):
                     self._replication_slots.pop(name)
                 else:
                     logger.error("Failed to drop replication slot '%s'", name)
```
hughcapet, line 211: don't we want to differentiate if drop failed due to the slot being active or any other error here as well?

tests/test_slots.py
```diff
@@ -52,7 +53,7 @@ def test_sync_replication_slots(self):
         self.p.set_role('master')
         with mock.patch('patroni.postgresql.Postgresql.role', new_callable=PropertyMock(return_value='replica')):
             self.s.sync_replication_slots(cluster, False)
-        with patch.object(SlotsHandler, 'drop_replication_slot', Mock(return_value=True)),\
+        with patch.object(SlotsHandler, 'drop_replication_slot', Mock(return_value=(False, True))),\
```
hughcapet, line 56: do we need it at all after you've added this as the default result in `MockCursor.execute()`?

hughcapet:
maybe it is the time to add a unit test for `_drop_incorrect_slots()` with logging assertion? 😇

patroni/postgresql/slots.py
CyberDem0n: It is purely theoretical. If slot isn't active and it can't be dropped most likely there is a filesystem corruption, what is already a much bigger problem.

hughcapet:
👍

CyberDem0n:
:+1:

Issue #2535, opened 2023-01-23
pyinstaller relative import


zloyded:
```Traceback (most recent call last):
  File "patroni/__main__.py", line 6, in <module>
ImportError: attempted relative import with no known parent package```
while using `pyintaller --clean patroni.spec` for rpm pkg
Think need update patroni.spec for fix.

zloyded:
And one more, after fix pyinstaller spec and successful build package, start failed with error:
` File "six.py", line 87, in _import_module
ModuleNotFoundError: No module named 'http.server'`

CyberDem0n:
> after fix pyinstaller spec

Could you please elaborate more on that?
What exactly have you changed?

zloyded:
> > after fix pyinstaller spec
> 
> Could you please elaborate more on that? What exactly have you changed?

i`ve fix locally pyinstaller  spec file 
`a = Analysis(['patroni.py'],
....`
and compile binary and run it i have 
`ModuleNotFoundError: No module named 'http.server'`
Build whith python 3.11.1, 3.10 3.9.16

zloyded:
`File "patroni/api.py", line 15, in <module>
File "six.py", line 237, in create_module
File "six.py", line 213, in load_module
File "six.py", line 120, in _resolve
File "six.py", line 87, in _import_module
ModuleNotFoundError: No module named 'http.server'
patroni[5001]: [5002] Failed to execute script 'patroni' due to unhandled exception!`

PR #2537, opened 2023-01-24
Compatibility with pyinstaller

CyberDem0n:
it doesn't like relative imports and not recognise `http.server` imported with `six`.
The last one is explicitly added to the list of `hiddenimports()` and will break compatibility with python 2.7, which support will be dropped in the next Patroni release anyway.

Close https://github.com/zalando/patroni/issues/2535


hughcapet:
👍

CyberDem0n:
:+1:

zloyded:
Super! You so fast!

PR #2538, opened 2023-01-24
 Check unexpected exceptions in Patroni logs after behave

CyberDem0n:
and make behave fail if there are anything unexpected found.

In addition to that fix globing rule when uploading artifacts with logs.

### DIFF
--- a/features/environment.py
+++ b/features/environment.py
@@ -1088,7 +1088,7 @@ def after_all(context):
         with open(log) as f:
             for line in f:
                 if 'please report it as a BUG' in line:
-                    print(':'.join([log, line]), end='')
+                    print(':'.join([log, line.rstrip()]))
                     found = True
     if found:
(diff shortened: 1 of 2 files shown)


features/environment.py
```diff
@@ -1111,8 +1100,20 @@ def after_feature(context, feature):
...
         shutil.rmtree(data)
     context.dcs_ctl.cleanup_service_tree()
-    if feature.status == 'failed':
+
+    found = False
+    logs = glob.glob(context.pctl.output_dir + '/patroni_*.log')
+    for log in logs:
+        with open(log) as f:
+            for line in f:
+                if 'Unexpected errors in Patroni log files' in line:
```
hughcapet, line 1109: why did `if 'please report it as a BUG'` disappear ? :)

hughcapet:
👍

CyberDem0n:
:+1:

Issue #2495, opened 2022-12-19
Feature request: make value of replication role labels configurable


matteoolivi:
**Describe the feature**
Currently, on K8s the value of the labels that identify the replication role of each patroni/postgresql Pod is hardcoded to be "master" for the leader and "replica" for the followers.

I would like to make those values configurable.

That allows us to stop using the non-inclusive "master" term and replace it with something like "leader" or primary.

I would also like to opt out from using "replica". The reason is that I'm running patroni/spilo in a context where "replica" is already widely used to describe something different than postgresql's replication replicas. So currently there's a name overloading that causes confusion.

CyberDem0n:
Getting rif from non-inclusive terms is in our radar.
The thing is that the problem is much deeper than just labels.
There are plenty of internal checks on the "master" and we can't rename it without doing an intermediate release (compatible between old and new terms).

Even with labels it is much harder than you probably anticipate. If Patroni is managing the endpoint on it's own - everything will be fine. If Patroni is using configmaps and there is a Service with label selector, then they could not be migrated to the new scheme without a downtime, because set-based requirements aren't supported for Services: https://kubernetes.io/docs/concepts/overview/working-with-objects/labels/#resources-that-support-set-based-requirements

matteoolivi:
@CyberDem0n thank you for your reply. I now have a better understanding of the problem and its caveats.

Strictly speaking my change request is not for Patroni to replace the non-inclusive terms. It's just to make the values of the replication role label for both master and replica pods configurable. This could be implemented in a backward-compatible way with a change that:

1. adds two env variables/config parameters that determine the value of the replication role label for leader and followers, respecitvely (e.g. `PATRONI_KUBERNETES_LEADER_LABEL_VALUE` and `PATRONI_KUBERNETES_FOLLOWER_LABEL_VALUE`
2. keeps the defaults for the two aforementioned env vars/config parameters to the legacy values (`master` and `replica`); this is what ensures backward compatibility
3. makes patroni's code use the value of the aforementioned config parameters rather than the hardcoded current values.

For my use case this would be already enough.

The change I described could be the intermediate non-breaking release you mentioned. Its release notes could advertise that in a few releases the defaults would be changed to the inclusive terms, so users could get ready.

> If Patroni is using configmaps and there is a Service with label selector, then they could not be migrated to the new scheme without a downtime, because set-based requirements aren't supported for Services

A couple of random thoughts on this (not sure whether they make sense):
[...]

PR #2541, opened 2023-01-25
Change master->primary/leader/member

CyberDem0n:
keep as much backward compatibility as possible.

Following changes were made:
1. All internal checks are performed as `role in ('master', 'primary')`
2. All internal variables/functions/methods are renamed
3. `GET /metrics` endpoint returns `patroni_primary` in addition to `patroni_master`.
4. Logs are changed to use leader/primary/member/remote depending on the context
5. Unit-tests are using only role = 'primary' instead of 'master' to verify that 1 works.
6. patronictl still supports old syntax, but also accepts `--leader` and `--primary`.
7. `master_(start|stop)_timeout` is automatically translated to `primary_(start|stop)_timeout` if the last one is not set.
8. updated the documentation and some examples

Future plan: in the next major release switch role name from `master` to `primary` and maybe drop `master` altogether.
The Kubernetes implementation will require more work and keep two labels in parallel. Label values should probably be configurable as described in https://github.com/zalando/patroni/issues/2495.


CyberDem0n:
@matthbakeredb, I noticed you liked the PR. Do you want to take a look/try/test it? :)

matthbakeredb:
> @matthbakeredb, I noticed you liked the PR. Do you want to take a look/try/test it? :)

hehe yes of course :smile:

tests/test_api.py
```diff
@@ -208,42 +208,42 @@ def test_do_GET(self):
...
         MockPatroni.dcs.cluster = None
-        with patch.object(RestApiHandler, 'get_postgresql_status', Mock(return_value={'role': 'master'})):
-            MockRestApiServer(RestApiHandler, 'GET /master')
+        with patch.object(RestApiHandler, 'get_postgresql_status', Mock(return_value={'role': 'primary'})):
+            MockRestApiServer(RestApiHandler, 'GET /primary')
         with patch.object(MockHa, 'restart_scheduled', Mock(return_value=True)):
-            MockRestApiServer(RestApiHandler, 'GET /master')
-        self.assertIsNotNone(MockRestApiServer(RestApiHandler, 'GET /master'))
+            MockRestApiServer(RestApiHandler, 'GET /primary')
+        self.assertIsNotNone(MockRestApiServer(RestApiHandler, 'GET /primary'))
```
matthbakeredb, line 215: Perhaps should check the content of this query as it's returning `master`

```
postgres@patroni2:~$ curl -s http://127.0.0.1:8008/primary|jq .role
"master"
```

patroni/ctl.py
```diff
@@ -241,24 +242,28 @@ def watching(w, watch, max_count=None, clear=True):
...
 
 
-def get_all_members(obj, cluster, group, role='master'):
+def get_all_members(obj, cluster, group, role='leader'):
     clusters = {0: cluster}
     if obj.get('citus') and group is None:
         clusters.update(cluster.workers)
-    if role == 'master':
+    if role in ('leader', 'master', 'primary', 'standby-leader'):
+        role = {'primary': 'master', 'standby-leader': 'standby_leader'}.get(role, role)
```
matthbakeredb, line 250: Looks like `master` is still the actual value, so still a "first class citizen", rather than replaced with `primary`. Shouldn't this be replaced both here and where it is being retrieved from?

matthbakeredb (commented):
Some thoughts; So the config and ctl cli is backwards compatible afaict but you're removing the /master api endpoint? Is this intended to be transitional, ie master to be removed entirely in a future release? If this is to be released using a new major version (non-backwards compatible) then perhaps `master` could be removed entirely. If it is transitional then I'd expect all interfaces to support the new and old naming, with the new naming as a "first class citizen" in all defaults, internal refs, docs, etc.

Do we want to leave it to the package maintainers to include migration tooling for the required changes to config?

Otherwise, everything looks good. Runs OK locally, behave tests passing, basic interaction seems good.

patroni/ctl.py
CyberDem0n: Yes, in this place we can reverse primary<->master in dict. 
But, REST API will still return role=master and so will be stored in the member key in DCS.
We can't simply change everything in one shot because it will break backward compatibility.
For example, during leader race all versions up to 2.1.5 are checking for role==master: https://github.com/zalando/patroni/blob/v2.1.5/patroni/ha.py#L42

tests/test_api.py
CyberDem0n: Yes, it is expected that new versions will still use "master" for a while.
Right now we change only tests in order to verify that the actual code correctly handles both cases, when role == 'master' or role == 'primary'.
Now we are planning a release 3.0.
My plan is to completely get rid from master only in 3.1, where in release notes we could write - "you should be running at least 3.0.0" in order to safely upgrade to "3.1.0" (mainly because of https://github.com/zalando/patroni/blob/v2.1.5/patroni/ha.py#L42). Upgrading directly from older version is sort of also possible, but there will be very easy to screw things up :(

matthbakeredb: I understand that it is a big change, bumping the major version notifies the user of backwards incompatible changes. Perhaps, if this change is backwards compatible this could go into 2.2, by allowing the user to get used to using the new syntax/interfaces whilst fully supporting all the old, with the main switch to not having `master` in 3.0? That would be more in-line with semver imo.

patroni/ctl.py
matthbakeredb: Perhaps there could be data migration in the dcs code that renames stored values? That could be optional in backwards compatible versions and enforced in non-backwards compat versions.

tests/test_api.py
CyberDem0n: This is not only about breaking the API, which indeed we don't break it yet.
There are two huge features recently merged: DCS failsafe and Citus support.
IMO, it deserves 3.0 :)

patroni/ctl.py
CyberDem0n: There is no data migration required.
Every node maintains its own member key.
The only problem is that the client tooling and new Patroni are required to understand new values.

tests/test_api.py
matthbakeredb: ah ha! Fair enough.

matthbakeredb (approved):
:+1:

CyberDem0n:
> So the config and ctl cli is backwards compatible afaict but you're removing the /master api endpoint?

No, it is not removed. All kind of "monitoring" endpoints are handled in the "do_GET()" method. It supports /primary since a long time.

> Is this intended to be transitional, ie master to be removed entirely in a future release?

Yes, in some future release it will be removed entirely.

> If this is to be released using a new major version (non-backwards compatible) then perhaps master could be removed entirely. If it is transitional then I'd expect all interfaces to support the new and old naming, with the new naming as a "first class citizen" in all defaults, internal refs, docs, etc.

That's more or less what I did. Docs are using only the "primary" or "leader".
There are only three places left where the "master" is used:
- REST API response: `{"role":"master", ...}`
- "/members/$name" in DCS: `{"role":"master", ...}`
- K8s "role" label on the leader pod

When everything is migrated to 3.0+, we can change REST API and DCS to `{"role":"primary", ...}`, entirely remove "master" from the code, and release 3.1.

With K8s it is way more complex. More details is in the #2495 

> Do we want to leave it to the package maintainers to include migration tooling for the required changes to config?

There is nothing package maintainers could do about it. Also, many people install Patroni with pip.

> Otherwise, everything looks good. Runs OK locally, behave tests passing, basic interaction seems good.

Well, yeah with behave tests are out of question, because we run them for every commit :)
Thank you for you

CyberDem0n:
:+1:

PR #2544, opened 2023-01-27
Remove patronictl scaffold

CyberDem0n:
The only reason for having it was a hacky way of running standby clusters.


hughcapet:
👍

CyberDem0n:
:+1:

PR #2548, opened 2023-02-03
Switch to Citus 11.2

CyberDem0n:
- Update Dockerfile.citus files
- Enable behave tests with Citus


Dockerfile.citus
```diff
@@ -27,7 +27,7 @@ RUN set -ex \
                             python3-etcd python3-kazoo python3-pip busybox \
                             net-tools iputils-ping --fix-missing \
     && curl https://install.citusdata.com/community/deb.sh | bash \
-    && apt-get -y install postgresql-$PG_MAJOR-citus-11.1 \
+    && apt-get -y install postgresql-$PG_MAJOR-citus-11.2 \
```
hughcapet, line 30: don't we want to have Citus version configurable as well?

hughcapet:
👍

PR #2554, opened 2023-02-08
Pass 'master' role to a callback script instead of 'promoted'

hughcapet:
Fix the regression introduced in https://github.com/zalando/patroni/pull/2504


CyberDem0n:
:+1:

hughcapet:
👍

PR #2557, opened 2023-02-09
Docker entrypoint uses *CONNECT* vars when set

sdenovan:
The entrypoint.sh for containers does not honor these documented environment variables:
PATRONI_RESTAPI_CONNECT_ADDRESS
PATRONI_POSTGRESQL_CONNECT_ADDRESS

This patch adds support for these variables. 

Why it's needed: When running Patroni instances under Docker / Podman on different hosts (eg. geographically distributed on a non-meshed network) the public addresses need to be used instead of the internal $DOCKER_IP (private/un-routable).


sdenovan:
Added a second commit to also allow PATRONI_ETCD_HOST which has the same issue.

CyberDem0n:
I am sorry, but I don't see a point in doing that.
This docker setup exists only for testing/debugging/demo purposes, and never meant for production usage.

sdenovan:
I agree, but as the only example in the project it's going to be point of reference.

This PR does not change the default behavior but it makes the example entrypoint.sh compliant with the documentation -- which I think is a win for people starting out with the project. 

I don't think there's any downside, but if you don't want it let me know I'll close the PR.

docker/entrypoint.sh
```diff
@@ -36,7 +36,7 @@ EOT
         exec dumb-init "$@"
         ;;
     etcd)
-        exec "$@" -advertise-client-urls "http://$DOCKER_IP:2379"
+        exec "$@" -advertise-client-urls "http://${PATRONI_ETCD_HOST:-$DOCKER_IP:2379}"
```
CyberDem0n, line 39: Meh. The `PATRONI_ETCD_HOST` is a Patroni configuration, not Etcd and should not be used in advertise-client-urls

Issue #2569, opened 2023-02-27
Recovery error when using custom bootstrap and recovery_target_action: pause


Aytuar:
**Describe the bug**
In issue [1891](https://github.com/zalando/patroni/pull/1891/files#diff-723fad026148c1592a96b17e06f7d1921abc8fa832ef864705b4a35770c14ce8R1075) a change was made that breaks patroni behavior when restoring via custom bootstrap.

**Expected behavior**
In postgres-Mon.log
```log
recovery stopping before commit of transaction 1858, time 2023-02-22 12:51:26.046508+03
```

This is the point in time I need to recover.
I go to the database and see if the necessary data has been restored, if not, stop patroni, change recovery_target_time and start patroni.
And I can repeat this several times, moving the restore point forward without having to restore the entire cluster all over again.
After reaching the desired timestamp, I simply execute the psql command `SELECT pg_xlog_replay_resume()` and postgres exits recovery mode and becomes the primary node.

If I rename `data.failed -> data` and start the patroni service, then postgres starts up, but immediately does promote, which I don’t need, because Otherwise, I can not do the recovery procedure described above.

**To Reproduce**
**Environment**
-   Patroni version: 2.0.2+
-   PostgreSQL version: 9.6+
-   DCS (and its version): etcd 3.5.1
-  postgres backup util: wal-g 2.0.1

**Patroni configuration file**
```yaml
bootstrap:
  method: wal-g
  wal-g:
    command: /usr/local/bin/wal_g_bootstrap.sh
    keep_existing_recovery_conf: False
    recovery_conf:
      recovery_target_time: '2023-02-22 12:51:26'
      recovery_target_action: pause
      recovery_target_timeline: 4
      restore_command: '/usr/local/bin/wal-g wal-fetch %f %p'
  dcs:
    ttl: 30
    loop_wait: 10
    retry_timeout: 10
    maximum_lag_on_failover: 1048576
    master_start_timeout: 30
    synchronous_mode: false
    synchronous_mode_strict: false
    postgresql:
      use_pg_rewind: true
      parameters:
        archive_mode: 'on'
        archive_command: '/usr/local/bin/wal-g wal-push %p'
```

**bootstrap script**
wal_g_bootstrap.sh
```bash
#!/usr/bin/env bash

while getopts ":-:" optchar; do
  [[ "${optchar}" == "-" ]] || continue
  case "${OPTARG}" in
    datadir=* )
        DATA_DIR=${OPTARG#*=}
        ;;
[...]

PR #2570, opened 2023-02-27
Set hot_standby=off only if recovery_target_action=promote

CyberDem0n:
During custom bootstrap the `hot_standby` is set to off to protect postgres from panicking and shutting down when some parameters like `max_connections` are increased on the primary.

According to the [documentation](https://www.postgresql.org/docs/current/runtime-config-wal.html#GUC-RECOVERY-TARGET-ACTION), `hot_standby` set to `off` affects behavior of the `recovery_target_action`, and `pause` starts acting as the `shutdown`:
> If [hot_standby](https://www.postgresql.org/docs/current/runtime-config-replication.html#GUC-HOT-STANDBY) is not enabled, a setting of pause will act the same as shutdown

 This is not what users expect/need, because normally they resolve pause state on their own.

To solve the problem we will set `hot_standby` to `off` during custom bootstrap only if `recovery_target_action` is set to 'promote'.

Close https://github.com/zalando/patroni/issues/2569

### DIFF
--- a/patroni/postgresql/config.py
+++ b/patroni/postgresql/config.py
@@ -1094,12 +1094,22 @@ class ConfigHandler(object):
                 effective_configuration[name] = cvalue
                 self._postgresql.set_pending_restart(True)
 
-        # If we are using custom bootstrap with PITR it could fail when values
-        # like max_connections are increased, therefore we disable hot_standby.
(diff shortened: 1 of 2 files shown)


hughcapet:
👍

hughcapet:
👍

PR #2571, opened 2023-02-28
Remove python 2.7 support

CyberDem0n:
- get rid from 2.7 specific modules: `six`, `ipaddress`
- use Python3 unpacking operator
- use `shutil.which()` instead of `find_executable()`


patroni/api.py
```diff
@@ -7,15 +7,14 @@
...
-import six
 import socket
 import sys
 
-from ipaddress import ip_address, ip_network as _ip_network
-from six.moves.BaseHTTPServer import BaseHTTPRequestHandler, HTTPServer
-from six.moves.socketserver import ThreadingMixIn
-from six.moves.urllib_parse import urlparse, parse_qs
+from http.server import BaseHTTPRequestHandler, HTTPServer
+from ipaddress import ip_address, ip_network
```
hughcapet, line 14: does this mean `ipaddress` is not a python2-specific lib...?

hughcapet:
👍

Issue #2445, opened 2022-10-31
[QUESTION] Patroni callbacks getting killed


RajKiranS:
We have noticed that Patroni itself is killing the callback it had just fired. Wanted to understand the reason behind it. And its happening randomly for different set of roles too.
```
Oct 31 10:28:06 foo.com patroni-foo: 2022-10-31 10:28:06,975 WARNING: Killed ['/usr/bin/patroni-callback-handler', '/etc/foo/patroni-callback.yaml', 'on_role_change', 'master', 'foo'] because it was still running
```
Why and when does this happen? How to prevent it?

Thanks in advance.

CyberDem0n:
Well, there was some other callback executed, but the previous was still running.
In order not to get to the race condition the previous is killed.

RajKiranS:
Yes, that is correct! In next 11 seconds, `on_reload` callback was fired because it assigned `synchronous_standby` to a replica.

But, how do we avoid such killings? That particular callback for `on_role_change` is very important as it would update and restart few things. Is there a flag to toggle such behavior? Having some configurable timeout makes sense than killing it..

CyberDem0n:
Well, there are two questons:
1. why it takes so long to execute callback script?
2. what do you need the on_reload callback for?

RajKiranS:
> why it takes so long to execute callback script?

We do restart a couple of services on the restart and that could sometimes take a while. And in this case, the on_reload callback was fired just within 11seconds...

> what do you need the on_reload callback for?

We do not need this... Is there a way to disable it?

CyberDem0n:
> We do not need this... Is there a way to disable it?

Remove it from the patroni.yaml?

RajKiranS:
> Remove it from the patroni.yaml?

Thought that they were mandatory! Will try this out. Thank you!

RajKiranS:
Thanks @CyberDem0n . Removing unwanted callbacks has helped. Closing.

PR #2578, opened 2023-03-01
Don't allow on_reload callback kill other callbacks

CyberDem0n:
Since a long time Patroni enforcing only one callback script running at a time. If the new callback is executed while the old one is still running, the old one is killed (including all child processes).

Such behavior is fine for all callbacks but on_reload, because the last one may accidentally cancel important ones, that for example updating DNS or assigning/removing Virtual IP.

To mitigate the problem we introduce a dedicated executor for on_reload callbacks, so that on_reload may only cancel another on_reload.

Ref: https://github.com/zalando/patroni/issues/2445

### DIFF
--- a/patroni/postgresql/callback_executor.py
+++ b/patroni/postgresql/callback_executor.py
@@ -1,9 +1,28 @@
 import logging
 
-from patroni.postgresql.cancellable import CancellableExecutor
+from patroni.postgresql.cancellable import CancellableExecutor, CancellableSubprocess
 from threading import Condition, Thread
+from typing import List
 
 logger = logging.getLogger(__name__)
+
+ACTION_NOOP = "noop"
+ACTION_ON_START = "on_start"
+ACTION_ON_STOP = "on_stop"
+ACTION_ON_RESTART = "on_restart"
+ACTION_ON_RELOAD = "on_reload"
+ACTION_ON_ROLE_CHANGE = "on_role_change"
+
+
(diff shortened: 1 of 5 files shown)


patroni/postgresql/callback_executor.py
```diff
@@ -1,22 +1,52 @@
...
+
+class OnReloadExecutor(CancellableSubprocess):
+
+    def call_nowait(self, cmd: List[str]) -> None:
+        """Run one `on_reload` callback at most.
+
+        To achive it we always kill already running command including child processes."""
+        self.cancel(kill=True)
+        self._kill_children()
+        self._start_process(cmd, close_fds=True)
```
hughcapet, line 25: why don't we need lock acquisition here?

patroni/postgresql/callback_executor.py
```diff
@@ -1,22 +1,60 @@
...
+    def __repr__(self):
+        return self.value
+
+
+class OnReloadExecutor(CancellableSubprocess):
+
+    def call_nowait(self, cmd: List[str]) -> None:
+        """Run one `on_reload` callback at most.
+
+        To achive it we always kill already running command including child processes."""
```
hughcapet, line 29:
```suggestion
        To achieve it we always kill already running command including child processes."""
```

hughcapet:
👍

CyberDem0n:
:+1:

PR #2579, opened 2023-03-02
Compatibility with flake8>=5.0

CyberDem0n:
The main() function now returns exit code instead of exiting on it's own


hughcapet:
👍

CyberDem0n:
:+1:

Issue #2434, opened 2022-10-17
recovery.conf file created for Patroni 2.1.1 - PG13


gerardo-ongres:
recovery.conf files are being created by Patroni in case of replica re-create fails. Recovery.conf files are incompatible with Postgres v13 or higher

**To Reproduce**
Rebooting a replica node

**Screenshots**
For this setup, they are 2 ways for replica bootstraping: PgBackrest and pgBasebackup. If both fails, a recovery.conf file seems to be created.

Traces from the logs:
Error log from Postgres:
```
[root@hvnplrrbdb02 log]# grep 'recovery.conf' postgresql-2022-10-1*
postgresql-2022-10-13_050403.log:2022-10-13 05:04:03.964 GMT [1888] FATAL:  using recovery command file "recovery.conf" is not supported
postgresql-2022-10-13_050416.log:2022-10-13 05:04:16.285 GMT [2663] FATAL:  using recovery command file "recovery.conf" is not supported
```

pgBackrest failed attempt to bootstrap:
```
2022-10-13 10:07:57,222 DEBUG: Converted retries value: 0 -> Retry(total=0, connect=None, read=None, redirect=0, status=None)
2022-10-13 10:07:57,235 DEBUG: http://10.34.52.167:2379/ "GET /v2/keys/service/tprrbdb01/?recursive=true HTTP/1.1" 200 None
2022-10-13 10:07:57,235 DEBUG: Writing {"conn_url":"postgres://10.40.0.160:5432/postgres","api_url":"http://10.40.0.160:8679/patroni","state":"stopped","role":"uninitialized","version":"2.1.1"} to
 key /service/tprrbdb01/members/host_2 ttl=30 dir=False append=False
2022-10-13 10:07:57,236 DEBUG: Converted retries value: 0 -> Retry(total=0, connect=None, read=None, redirect=0, status=None)
:
2022-10-13 10:07:57,273 DEBUG: http://10.34.52.167:2379/ "PUT /v2/keys/service/tprrbdb01/members/host_2 HTTP/1.1" 201 355
2022-10-13 10:07:57,273 INFO: Lock owner: host_3; I am host_2
2022-10-13 10:07:57,339 DEBUG: Writing {"conn_url":"postgres://10.40.0.160:5432/postgres","api_url":"http://10.40.0.160:8679/patroni","state":"stopped","role":"uninitialized","version":"2.1.1"} to key /service/tprrbdb01/members/host_2 ttl=30 dir=False append=False
[...]

PR #2583, opened 2023-03-03
Don't write to PGDATA if major version is not known

CyberDem0n:
It could happen that Patroni is started up before PGDATA was mounted. In this case Patroni can't determine major Postgres version from PG_VERSION file. Later, when PGDATA is mounted, Patroni was trying to create the recovery.conf even if the actual Postgres major version is newver than 12.

To mitigate the problem we double check that the `Postgresql._major_version` is set before writing recovery configuration or starting postgres up.

Close https://github.com/zalando/patroni/issues/2434

### DIFF
--- a/patroni/postgresql/__init__.py
+++ b/patroni/postgresql/__init__.py
@@ -13,6 +13,7 @@ from datetime import datetime
 from dateutil import tz
 from psutil import TimeoutExpired
 from threading import current_thread, Lock
+from typing import Optional
 
 from .bootstrap import Bootstrap
 from .callback_executor import CallbackExecutor
@@ -25,6 +26,7 @@ from .postmaster import PostmasterProcess
 from .slots import SlotsHandler
 from .sync import SyncHandler
 from .. import psycopg
+from ..dcs import Member
 from ..exceptions import PostgresConnectionException
 from ..utils import Retry, RetryFailedError, polling_loop, data_directory_is_empty, parse_int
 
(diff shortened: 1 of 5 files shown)


patroni/postgresql/__init__.py
```diff
@@ -932,7 +938,28 @@ def get_history(self, timeline):
...
+               do_reload: Optional[bool] = False) -> Optional[bool]:
+        """Reconfigure postgres to follow a new member or use different recovery parameters.
+
+        Method may call `on_role_change` callback if role is changing.
+
+        :param member: The member to follow
+        :param role: The desired role, normally 'replica', but could also be a 'standby_leader'
+        :path The desired role, normally 'replica', but could also be 'standby_leader'
+        :param timeout: start timeout, how long should the `start()` method wait for postgres accepting connections
+        :param do_reload: indicates that after updating postgresql.conf we just need to to a reload instead of restart
```
hughcapet, line 954:
```suggestion
        :param do_reload: indicates that after updating postgresql.conf we just need to do a reload instead of restart
```

patroni/postgresql/__init__.py
```diff
@@ -932,7 +938,28 @@ def get_history(self, timeline):
...
+               role: Optional[str] = 'replica',
+               timeout: Optional[float] = None,
+               do_reload: Optional[bool] = False) -> Optional[bool]:
+        """Reconfigure postgres to follow a new member or use different recovery parameters.
+
+        Method may call `on_role_change` callback if role is changing.
+
+        :param member: The member to follow
+        :param role: The desired role, normally 'replica', but could also be a 'standby_leader'
+        :path The desired role, normally 'replica', but could also be 'standby_leader'
```
hughcapet, line 952: what is `:path`?

CyberDem0n: Nothing, seems to be accidental copy&paste, removed.

hughcapet:
👍

CyberDem0n:
:+1:

PR #2588, opened 2023-03-08
Enable multihost docker compose

pflaeging:
This are some small changes in the docker container for patroni which enables a real production rollout on multiple hosts with docker-compose.

The main changes are:

- make almost all variables in `docker/entrypoint.sh` overwritable by environment
- if there's no database after initialising AND the standard variables like POSTGRES_DB, POSTGRES_USER, POSTGRES_PASSWD are set, create a database

There's a small project helping to set up such a cluster: <https://github.com/pflaeging/patroni-compose>

This setup is in production use by customers (mainly in cooperation with a redundant KeyCloak 19+ installation also on docker compose).

---
:peter


CyberDem0n:
As it is written in the [top of the Dockerfile](https://github.com/zalando/patroni/blob/master/Dockerfile#L1) it serves the only purpose - playing with Patroni on your machine, therefore we have no interest in extending it beyond that.

Also I really dislike changes to sample config files (`postgres[012].yml`). The `post_bootstrap` hook that you added makes it impossible to start Patroni locally because it references non-existing script.

If you want to build something with Patroni just create your own Dockerfile where you will add everything you want and install Patroni with `pip`. But, please always keep in mind that the _real production_ isn't possible without doing all kind of things around backups, continues archiving, and PITR.

PR #2597, opened 2023-03-13
Fix bug with metadata after coordinator failover

CyberDem0n:
We made incorrect assupmtion that `citus_set_coordinator_host()` will trigget `pg_dist_node` sync. Instead we should also use `citus_update_node()` and call `citus_set_coordinator_host()` only during the bootstrap.

Adjust behave tests to verify that coordinator failover is visible on workers.


hughcapet:
👍

CyberDem0n:
:+1:

Issue #2604, opened 2023-03-17
`patronictl edit-config` throws an exception if neither `less` nor `more` is available in the host


barthisrael:
### What happened?

I attempted to run the following command in a host that has neither `less` nor `more` installed:

```
patronictl -c postgres0.yml edit-config --pg work_mem=8MB --force batman
```

That caused Patroni to throw this exception:

```
Traceback (most recent call last):
  File "/usr/bin/patronictl", line 11, in <module>
    load_entry_point('patroni==3.0.1', 'console_scripts', 'patronictl')()
  File "/usr/lib/python3/dist-packages/click/core.py", line 764, in __call__
    return self.main(*args, **kwargs)
  File "/usr/lib/python3/dist-packages/click/core.py", line 717, in main
    rv = self.invoke(ctx)
  File "/usr/lib/python3/dist-packages/click/core.py", line 1137, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
  File "/usr/lib/python3/dist-packages/click/core.py", line 956, in invoke
    return ctx.invoke(self.callback, **ctx.params)
  File "/usr/lib/python3/dist-packages/click/core.py", line 555, in invoke
    return callback(*args, **kwargs)
  File "/usr/lib/python3/dist-packages/click/decorators.py", line 27, in new_func
    return f(get_current_context().obj, *args, **kwargs)
  File "/usr/lib/python3/dist-packages/patroni/ctl.py", line 1250, in edit_config
    show_diff(before_editing, after_editing)
  File "/usr/lib/python3/dist-packages/patroni/ctl.py", line 1102, in show_diff
    markup_to_pager(PatchStream(buf), opts)
  File "/usr/lib/python3/dist-packages/cdiff.py", line 639, in markup_to_pager
    pager = subprocess.Popen(
  File "/usr/lib/python3.8/subprocess.py", line 858, in __init__
    self._execute_child(args, executable, preexec_fn, close_fds,
  File "/usr/lib/python3.8/subprocess.py", line 1704, in _execute_child
    raise child_exception_type(errno_num, err_msg, err_filename)
FileNotFoundError: [Errno 2] No such file or directory: 'less'
```

### How can we reproduce it (as minimally and precisely as possible)?

1. Make sure both `less` and `more` are unavailable
2. Create a Patroni cluster using the `postgres0.yml` defined in the repository
3. Run `patronictl -c postgres0.yml edit-config --pg work_mem=8MB --force batman`

[...]

PR #2605, opened 2023-03-17
Handle `patronictl edit-config` diff pager in a more user friendly way

barthisrael:
`patronictl edit-config` requires a pager to show the diff output back to the user. It used to be hard-coded to use either `less` or `more`.

When these tools were not available in the host that would cause `patronictl` to face an exception in `ydiff` module and to show the stack trace in the console.

This PR changes `patronictl edit-config` command to behave like this:

- If `PAGER` environment variable is set, attempt to find the corresponding executable.
- If `PAGER` is not set or is set with an invalid executable, then attempt to use either `less` or `more` as it used to do.
- If no executable is find at all then throw a `PatroniCtlException` to show an user friendly message

Unit tests in `tests/test_ctl.py` were modified accordingly.

References: PAT-21
References: #2604

### DIFF
--- a/patroni/ctl.py
+++ b/patroni/ctl.py
@@ -1092,11 +1092,27 @@ def show_diff(before_editing, after_editing):
             width = 80
             tab_width = 8
             wrap = True
-            if shutil.which('less'):
-                pager = None
-            else:
-                pager = os.path.basename(shutil.which('more') or 'more')
+            pager = next(
+                (
+                    os.path.basename(p)
+                    for p in (os.environ.get('PAGER', None), "less", "more")
+                    if p is not None and shutil.which(p)
+                ),
+                None,
+            )
             pager_options = None
+
+        if opts.pager is None:
+            raise PatroniCtlException(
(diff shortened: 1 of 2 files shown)


patroni/ctl.py
```diff
@@ -1092,12 +1092,28 @@ class opts:
...
             tab_width = 8
             wrap = True
-            if shutil.which('less'):
-                pager = None
-            else:
-                pager = os.path.basename(shutil.which('more') or 'more')
+            pager = next(
+                (
+                    os.path.basename(p)
+                    for p in (os.environ.get('PAGER', None), "less", "more")
```
CyberDem0n, line 1098: `dict.get` returns `None` by default if value is missing
```suggestion
                    for p in (os.environ.get('PAGER'), "less", "more")
```

barthisrael: I missed changing the unit test, doing that now

barthisrael: done!

CyberDem0n:
:+1:

PR #2612, opened 2023-03-21
Add docstrings and type annotations to `patroni/validator.py`

barthisrael:
References: PAT-42

### DIFF
--- a/patroni/validator.py
+++ b/patroni/validator.py
@@ -244,7 +244,7 @@ def validate_data_dir(data_dir: str) -> bool:
 class Result(object):
     """Represent the result of a given validation that was performed.
 
-    :ivar status: Iif the validation succeeded.
+    :ivar status: If the validation succeeded.
     :ivar path: YAML tree path of the configuration option.
     :ivar data: value of the configuration option.
     :ivar level: error level, in case of error.
--- a/patroni/validator.py
+++ b/patroni/validator.py
@@ -540,12 +540,12 @@ class Schema(object):
         for i in self.iter():
             yield i
 
-    def iter(self) -> Generator[Result, None, None]:
+    def iter(self) -> Iterator[Result]:
         """Iterate over ``validator``, if it is an iterable object, to validate the corresponding entries in ``data``.
 
         Only :class:`dict`, :class:`list`, :class:`Directory` and :class:`Or` objects are considered iterable objects.
 
-        :return: result of a validation that has been performed against the configuration.
+        :rtype: Iterator[:class:`Result`] objects with the error message related to the failure, if any check fails.
         """
         if isinstance(self.validator, dict):
             if not isinstance(self.data, dict):
--- a/patroni/validator.py
+++ b/patroni/validator.py
@@ -22,7 +22,7 @@ def data_directory_empty(data_dir: str) -> bool:
     """Check if PostgreSQL data directory is empty.
 
     :param data_dir: path to the PostgreSQL data directory to be checked.
-    :return: if the data directory is empty.
+    :returns: True if the data directory is empty.
     """
     if os.path.isfile(os.path.join(data_dir, "global", "pg_control")):
         return False
--- a/patroni/validator.py
+++ b/patroni/validator.py
@@ -275,7 +275,7 @@ class Result(object):
         else:
             self.error = None
 
-    def __repr__(self):
+    def __repr__(self) -> str:
         """Show configuration path and value. If the validation failed, also show the error message."""
         return str(self.path) + (" " + str(self.data) + " " + str(self._error) if self.error else "")
 
--- a/patroni/validator.py
+++ b/patroni/validator.py
@@ -277,7 +277,7 @@ class Result(object):
 
     def __repr__(self):
         """Show configuration path and value. If the validation failed, also show the error message."""
-        return self.path + (" " + str(self.data) + " " + self._error if self.error else "")
+        return str(self.path) + (" " + str(self.data) + " " + str(self._error) if self.error else "")
 
 
 class Case(object):
--- a/patroni/validator.py
+++ b/patroni/validator.py
@@ -10,6 +10,7 @@ import re
 import shutil
 import socket
 import subprocess
+
 from typing import Any, Union, Generator, List
 
 from .utils import split_host_port, data_directory_is_empty
--- a/patroni/validator.py
+++ b/patroni/validator.py
@@ -367,11 +367,11 @@ class Directory(object):
         self.contains = contains
         self.contains_executable = contains_executable
 
-    def validate(self, name: str) -> Generator[Result, None, None]:
+    def validate(self, name: str) -> Iterator[Result]:
         """Check if the expected paths and executables can be found under *name* directory.
 
         :param name: path to the base directory against which paths and executables will be validated.
-        :return: a :class:`Result` object with the error message related to the failure, if any check fails.
+        :rtype: Iterator[:class:`Result`] objects with the error message related to the failure, if any check fails.
         """
         if not name:
             yield Result(False, "is an empty string")
--- a/patroni/validator.py
+++ b/patroni/validator.py
@@ -397,7 +397,7 @@ class Schema(object):
 
     It contains all the configuration options that are available in each scope, including the validation(s) that should
     be performed against each one of them. The validations will be performed whenever the :class:`Schema` object is
-    called, or its :funct`validate` method is called.
+    called, or its :func:`validate` method is called.
 
     :ivar validator: validator of the configuration schema. Can be any of these
         * :class:`str`: defines that a string value is required; or
--- a/patroni/validator.py
+++ b/patroni/validator.py
@@ -667,7 +667,7 @@ def _get_type_name(python_type: Any) -> str:
                     python_type, getattr(python_type, __name__, "unknown type"))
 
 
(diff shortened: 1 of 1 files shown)


patroni/validator.py
```diff
@@ -135,30 +274,104 @@ def __init__(self, status, error="didn't pass validation", level=0, path="", dat
...
+        """
         self.contains = contains
         self.contains_executable = contains_executable
 
-    def validate(self, name):
+    def validate(self, name: str) -> Generator[Result, None, None]:
+        """Check if the expected paths and executables can be found under *name* directory.
+
+        :param name: path to the base directory against which paths and executables will be validated.
+        :return: a :class:`Result` object with the error message related to the failure, if any check fails.
```
CyberDem0n, line 373:
```suggestion
    def validate(self, name: str) -> Iterator[Result]:
        """Check if the expected paths and executables can be found under *name* directory.

        :param name: path to the base directory against which paths and executables will be validated.
        :rtype: Iterator[:class:`Result`] objects with the error message related to the failure, if any check fails.
```

patroni/validator.py
```diff
@@ -177,18 +390,126 @@ def validate(self, name):
...
+        :return:list of errors identified while validating the *data*, if any.
+        """
         errors = []
         for i in self.validate(data):
             if not i.status:
                 errors.append(str(i))
         return errors
 
-    def validate(self, data):
+    def validate(self, data: Any) -> Generator[Result, None, None]:
```
CyberDem0n, line 492:
```suggestion
    def validate(self, data: Any) -> Iterator[Result]:
```

patroni/validator.py
```diff
@@ -177,18 +390,126 @@ def validate(self, name):
...
+
+        Additionally:
+        * If ``validator`` attribute is a callable object, calls it to validate *data* argument. Before doing so, if
+            `validator` contains an ``expected_type`` attribute, check if *data* argument is compliant with that
+            expected type.
+        * If ``validator`` attribute is an iterable object (:class:`dict`, :class:`list`, :class:`Directory` or
+            :class:`Or`), then it iterates over it to validate each of the corresponding entries in *data* argument.
+
+        :param data: configuration to be validated against ``validator``.
+        :return: result of a validation that has been performed.
```
CyberDem0n, line 505:
```suggestion
        :rtype: Iterator[:class:`Result`] objects with the error message related to the failure, if any check fails.
```

patroni/validator.py
```diff
@@ -218,7 +539,13 @@ def validate(self, data):
         for i in self.iter():
             yield i
 
-    def iter(self):
+    def iter(self) -> Generator[Result, None, None]:
+        """Iterate over ``validator``, if it is an iterable object, to validate the corresponding entries in ``data``.
+
+        Only :class:`dict`, :class:`list`, :class:`Directory` and :class:`Or` objects are considered iterable objects.
+
+        :return: result of a validation that has been performed against the configuration.
```
CyberDem0n, line 547:
```suggestion
    def iter(self) -> Iterator[Result]:
        """Iterate over ``validator``, if it is an iterable object, to validate the corresponding entries in ``data``.

        Only :class:`dict`, :class:`list`, :class:`Directory` and :class:`Or` objects are considered iterable objects.

        :rtype: Iterator[:class:`Result`] objects with the error message related to the failure, if any check fails.
```

patroni/validator.py
```diff
@@ -240,7 +572,13 @@ def iter(self):
             for i in self.iter_or():
                 yield i
 
-    def iter_dict(self):
+    def iter_dict(self) -> Generator[Result, None, None]:
+        """Iterate over a :class:`dict` based ``validator`` to validate the corresponding entries in ``data``.
+
+        :return: result of a validation that has been performed against the configuration.
```
CyberDem0n, line 578:
```suggestion
    def iter_dict(self) -> Iterator[Result]:
        """Iterate over a :class:`dict` based ``validator`` to validate the corresponding entries in ``data``.

        :rtype: Iterator[:class:`Result`] objects with the error message related to the failure, if any check fails.
```

patroni/validator.py
```diff
@@ -251,20 +589,31 @@ def iter_dict(self):
...
                                      path=(d + ("." + v.path if v.path else "")), level=v.level, data=v.data)
 
-    def iter_or(self):
+    def iter_or(self) -> Generator[Result, None, None]:
+        """Perform all validations defined in an `Or` object for a given configuration option.
+
+        This method can be only called against leaf nodes in the configuration tree. :class:`Or` objects defined in the
+        ``validator`` keys will be handled by :func:`iter_dict` method.
+
+        :return: result of a validation that has been performed against the configuration.
```
CyberDem0n, line 604:
```suggestion
    def iter_or(self) -> Iterator[Result]:
        """Perform all validations defined in an `Or` object for a given configuration option.

        This method can be only called against leaf nodes in the configuration tree. :class:`Or` objects defined in the
        ``validator`` keys will be handled by :func:`iter_dict` method.

        :rtype: Iterator[:class:`Result`] objects with the error message related to the failure, if any check fails.
```

patroni/validator.py
```diff
@@ -273,28 +622,59 @@ def iter_or(self):
                 max_level = v.level
                 yield Result(v.status, v.error, path=v.path, level=v.level, data=v.data)
 
-    def _data_key(self, key):
+    def _data_key(self, key: Union[str, Optional, Or]) -> Generator[str, None, None]:
```
CyberDem0n, line 625:
```suggestion
    def _data_key(self, key: Union[str, Optional, Or]) -> Iterator[str]:
```

patroni/validator.py
```diff
@@ -273,28 +622,59 @@ def iter_or(self):
...
+        """Map a key from the ``validator`` dictionary to the corresponding key(s) in the ``data`` dictionary.
+
+        :param key: key from the ``validator`` attribute.
+
+        Yields:
+            Key that should be used to access corresponding value in the ``data`` attribute.
```
CyberDem0n, line 631:
```suggestion
        :rtype: Iterator[str], keys that should be used to access corresponding value in the ``data`` attribute.
```

patroni/validator.py
```diff
@@ -273,28 +622,59 @@ def iter_or(self):
...
+    Returns:
+        User friendly name of the given Python type.
+    """
+    return {str: 'a string', int: 'an integer', float: 'a number',
             bool: 'a boolean', list: 'an array', dict: 'a dictionary'}.get(
                     python_type, getattr(python_type, __name__, "unknown type"))
 
 
-def assert_(condition, message="Wrong value"):
+def assert_(condition: bool, message: str = "Wrong value") -> None:
```
CyberDem0n, line 670:
```suggestion
def assert_(condition: bool, message: Optional[str] = "Wrong value") -> None:
```

CyberDem0n (commented):
When adding suggestions I just realized that there is a name collision between `typing.Optional` and `Optional` class defined in this file.
I thing it would be solved by directly referencing `typing.Optional` in function/methods arguments.

Also a little remark, `:return:` and `:returns:` have totally the same meaning, but we prefer the second one. Especially when `bool` is returned:
```python
    :return: if the data directory is empty.
vs
    :returns: `True` if the data directory is empty.
```

IMO the last one looks nicer. I didn't make suggestions to change all of them though, it is up to you.

barthisrael:
> When adding suggestions I just realized that there is a name collision between `typing.Optional` and `Optional` class defined in this file. I thing it would be solved by directly referencing `typing.Optional` in function/methods arguments.
> 
> Also a little remark, `:return:` and `:returns:` have totally the same meaning, but we prefer the second one. Especially when `bool` is returned:
> 
> ```python
>     :return: if the data directory is empty.
> vs
>     :returns: `True` if the data directory is empty.
> ```
> 
> IMO the last one looks nicer. I didn't make suggestions to change all of them though, it is up to you.

I see, thanks! Either way would work for me too. I only went with `:return`: instead of `:returns:` because when using `pdoc` it was rendering only the former in the appropriate way. But that's only a detail.

barthisrael:
Applied all changes, thank you both!

patroni/validator.py
```diff
@@ -56,40 +99,98 @@ def validate_host_port(host_port, listen=False, multiple_hosts=False):
...
+    """
     try:
         socket.inet_pton(socket.AF_INET6, ip)
     except Exception:
         raise ConfigParseError("Is not a valid ipv6 address")
     return True
 
 
-def get_major_version(bin_dir=None):
+def get_major_version(bin_dir: str = None) -> str:
```
CyberDem0n, line 180:
```suggestion
def get_major_version(bin_dir: OptionalType[str] = None) -> str:
```

CyberDem0n: Besides that please also change https://github.com/zalando/patroni/blob/a04612b8fe7c40afeb55095b42ae802d39e54a54/patroni/validator.py#L199-L200 to:
```python
    match = re.match(r'^[^\s]+ [^\s]+ (\d+)(\.(\d+))?', version)
    if not match:
        return version
    return '.'.join([match.group(1), match.group(3)]) if int(match.group(1)) < 10 else match.group(1)
```

Because pyright linter complains about them.

CyberDem0n (commented):
A couple of more comments

hughcapet:
👍

CyberDem0n:
:+1:

PR #2217, opened 2022-02-14
fix AttributeError no attribute 'leader'

hmilkovi:
Hi when I was using Raft DC and sync mode I have encountered:
```
2022-02-14 12:58:56,025 WARNING: Loop time exceeded, rescheduling immediately.
2022-02-14 12:59:02,043 ERROR: Unexpected exception
Traceback (most recent call last):
  File "/usr/local/lib/python3.9/dist-packages/patroni/ha.py", line 1502, in run_cycle
    info = self._run_cycle()
  File "/usr/local/lib/python3.9/dist-packages/patroni/ha.py", line 1464, in _run_cycle
    ret = self.process_unhealthy_cluster()
  File "/usr/local/lib/python3.9/dist-packages/patroni/ha.py", line 970, in process_unhealthy_cluster
    if self.is_healthiest_node():
  File "/usr/local/lib/python3.9/dist-packages/patroni/ha.py", line 799, in is_healthiest_node
    if self.is_synchronous_mode() and self.cluster.sync.leader:
AttributeError: 'NoneType' object has no attribute 'leader'
2022-02-14 12:59:02,043 INFO: Unexpected exception raised, please report it as a BUG
2022-02-14 12:59:02,044 WARNING: Loop time exceeded, rescheduling immediately.
...
```

Basically it doesn't proceed past this point.

PR #2614, opened 2023-03-22
Make sure Cluster.sync is never empty

CyberDem0n:
It was possible to have it empty if the all cluster keys are missing in DCS. In this case the `Cluster` object was manually created with all values set to `None` or `[]` (including sync).
It already resulted in #2217, which is in fact wasn't a correct fix.

In order to solve it and reduce code duplication we introduce `Cluster.empty()` and `SyncState.empty()` methods, which will create corresponding empty objects and start using `Cluster.empty()` from all places where the empty `Cluster` object was manually created.

### DIFF
--- a/patroni/dcs/__init__.py
+++ b/patroni/dcs/__init__.py
@@ -14,6 +14,7 @@ from collections import defaultdict, namedtuple
 from copy import deepcopy
 from random import randint
 from threading import Event, Lock
+from typing import Any, Dict, List, Optional, Union
 from urllib.parse import urlparse, urlunparse, parse_qsl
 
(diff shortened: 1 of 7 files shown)


patroni/dcs/__init__.py
```diff
@@ -374,7 +375,7 @@ class SyncState(namedtuple('SyncState', 'index,leader,sync_standby')):
     """
 
     @staticmethod
-    def from_node(index, value):
+    def from_node(index: Union[str, int], value: Union[str, Dict[str, Any]]):
```
hughcapet: definitely a docstring required

hughcapet:
👍

Issue #2613, opened 2023-03-22
Feature: report replica status "Sync Standby" via /metrics


TLINDEN:
### What happened?

Patroni already reports the status of the master (or standby leader) and the replicas using the metrics:
```
patroni_replica
patroni_master
patroni_standby_leader
```
However, it doesn't report if a replica is the current "Sync Standby". But it would be nice to have this information as a metric (especially when no replica is actually the sync standby).

### How can we reproduce it (as minimally and precisely as possible)?

Scrape the mentioned metrics.

### What did you expect to happen?

A new metric would be nice to have, e.g. `patroni_sync_standby`.

### Patroni/PostgreSQL/DCS version

-   Patroni version: 2.1.4
-   PostgreSQL version:  13.8
-   DCS (and its version): etcd

### Patroni configuration file

```yaml
doesn't matter.
```

### patronictl show-config

```yaml
doesn't matter.
```

### Patroni log files

```shell
doesn't matter.
```

### PostgreSQL log files

```shell
doesn't matter.
```

### Have you tried to use GitHub issue search?

- [X] Yes

CyberDem0n:
Fell free to add it, should be straightforward: https://github.com/zalando/patroni/blob/master/patroni/api.py#L228

TLINDEN:
How about this?

PR #2615, opened 2023-03-22
Add metric to report about sync standby replica status, see #2613

TLINDEN:
Close https://github.com/zalando/patroni/issues/2613

### DIFF
--- a/patroni/api.py
+++ b/patroni/api.py
@@ -271,7 +271,7 @@ class RestApiHandler(BaseHTTPRequestHandler):
 
         metrics.append("# HELP patroni_sync_standby Value is 1 if this node is a sync standby replica, 0 otherwise.")
         metrics.append("# TYPE patroni_sync_standby gauge")
-        metrics.append("patroni_sync_standby{0} {1}".format(scope_label, postgres['sync_standby']))
+        metrics.append("patroni_sync_standby{0} {1}".format(scope_label, int(postgres.get('sync_standby', False))))
 
         metrics.append("# HELP patroni_xlog_received_location Current location of the received"
                        " Postgres transaction log, 0 if this node is not a replica.")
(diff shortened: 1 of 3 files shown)


patroni/api.py
```diff
@@ -686,6 +691,9 @@ def get_postgresql_status(self, retry=False):
             if result['role'] == 'replica' and self.server.patroni.ha.is_standby_cluster():
                 result['role'] = postgresql.role
 
+            if result['role'] == 'replica' and self.server.patroni.ha.is_sync_standby():
+                result['sync_standby'] = 1
```
CyberDem0n, line 695: Here we want to use the most up-to-date `cluster` object:
```suggestion
            if result['role'] == 'replica' and cluster.is_synchronous_mode()\
                    and cluster.sync and postgresql.name in cluster.sync.members:
                result['sync_standby'] = True
```

patroni/api.py
```diff
@@ -269,6 +269,10 @@ def do_GET_metrics(self):
         metrics.append("# TYPE patroni_replica gauge")
         metrics.append("patroni_replica{0} {1}".format(scope_label, int(postgres['role'] == 'replica')))
 
+        metrics.append("# HELP patroni_sync_standby Value is 1 if this node is a sync standby replica, 0 otherwise.")
+        metrics.append("# TYPE patroni_sync_standby gauge")
+        metrics.append("patroni_sync_standby{0} {1}".format(scope_label, postgres['sync_standby']))
```
CyberDem0n, line 274:
```suggestion
        metrics.append("patroni_sync_standby{0} {1}".format(scope_label, int(postgres.get('sync_standby', False))))
```

patroni/api.py
```diff
@@ -673,6 +677,7 @@ def get_postgresql_status(self, retry=False):
                 'state': postgresql.state,
                 'postmaster_start_time': row[0],
                 'role': 'replica' if row[1] == 0 else 'master',
+                'sync_standby': 0,
```
CyberDem0n, line 680: This one should probably be removed, in JSON it doesn't have so much value because absence of this field will automatically mean that the node is async.

```suggestion
```

hughcapet:
we will also need to add the new metric to the docs [here](https://github.com/burakergenn/patroni/blob/master/docs/rest_api.rst#monitoring-endpoint)

hughcapet:
👍

PR #2626, opened 2023-03-29
Compatibility with python 3.6

CyberDem0n:
despite being EOL the number of downloads from pypi for 3.6 is on the first place, hence we need to support it.

Fixed a couple of problems in tests:
1. pytest complains about `call` imported from mock, mock.call() is fine
2. one test of Citus intergration was broken


hughcapet:
👍

CyberDem0n:
:+1:

PR #2629, opened 2023-04-03
Introduce patroni.collections

CyberDem0n:
For now it implements:
- CaseInsensitiveDict()
- CaseInsensitiveSet()

Update `patroni.postgresql.sync.parse_sync_standby_names()` to use `CaseInsensitiveSet()` instead of `CaseInsensitiveDict()`


CyberDem0n:
:+1:

PR #2631, opened 2023-04-03
More use of CaseInsensitiveSet

CyberDem0n:
1. make `SyncHandler.current_state()` return `CaseInsensitiveSet` instead of `list` objects.
2. take `sync_node_count` and `sync_node_maxlag` from the `Postgresql._global_config` instead of passing them as arguments.
3. Make `AbstractDCS.write_sync_state()` accept any `Collection`-like objects.

### DIFF
--- a/patroni/ha.py
+++ b/patroni/ha.py
@@ -473,7 +473,7 @@ class Ha(object):
                 node_to_follow = self._get_node_to_follow(self.cluster)
 
             if self.is_synchronous_mode():
-                self.state_handler.sync_handler.set_synchronous_standby_names([])
+                self.state_handler.sync_handler.set_synchronous_standby_names(CaseInsensitiveSet())
         elif self.has_lock():
             msg = "starting as readonly because i had the session lock"
             node_to_follow = None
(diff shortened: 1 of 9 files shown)


patroni/ha.py
```diff
@@ -578,36 +579,34 @@ def process_sync_replication(self):
         promoting standbys that were guaranteed to be replicating synchronously.
         """
         if self.is_synchronous_mode():
-            sync_node_count = self.global_config.synchronous_node_count
-            sync_node_maxlag = self.global_config.maximum_lag_on_syncnode
-            current = [] if self.cluster.sync.is_empty else self.cluster.sync.members
-            picked, allow_promote = self.state_handler.sync_handler.current_state(self.cluster, sync_node_count,
-                                                                                  sync_node_maxlag)
-            if set(picked) != set(current):
+            current = CaseInsensitiveSet([] if self.cluster.sync.is_empty else self.cluster.sync.members)
```
hughcapet, line 582: isn't `is_empty` check already done in the `members` property?

hughcapet:
👍

Issue #2632, opened 2023-04-04
Arguments for initdb containing spaces require double quoting


matthbakeredb:
### What happened?

During testing new TDE setup in pg15 we noticed that configuration for initdb is unintuitive. Values which contain spaces appear to require quoting twice. The result if this is not done is a failure to bootstrap the cluster.

### How can we reproduce it (as minimally and precisely as possible)?

Create a patroni configuration with extra arguments to setup TDE. Start patroni fresh install.

### What did you expect to happen?

Postgres initdb key-value arguments passed quoted when in a list of dicts form, where the value is a string with spaces.

### Patroni/PostgreSQL/DCS version

-   Patroni version: 
-   PostgreSQL version: 
-   DCS (and its version): 

### Patroni configuration file

```yaml
initdb:
  - encoding: UTF8
  - data-checksums
  - auth-local: peer
  - auth-host: scram-sha-256
  - data-encryption
  - key-wrap-command: "openssl enc -e -pbkdf2 -out %p -pass pass:trustno1"
  - key-unwrap-command: "openssl enc -d -pbkdf2 -in %p -pass pass:trustno1"
```

### patronictl show-config

```yaml
n/a
```

### Patroni log files

```shell
Apr 04 07:42:27 pgext-patroni1 patroni[26555]: 2023-04-04 07:42:27,101 INFO: trying to bootstrap a new cluster
Apr 04 07:42:27 pgext-patroni1 patroni[26555]: /usr/edb/pge15/bin/initdb: invalid option -- 'e'
Apr 04 07:42:27 pgext-patroni1 patroni[26555]: initdb: hint: Try "initdb --help" for more information.
Apr 04 07:42:27 pgext-patroni1 patroni[26555]: pg_ctl: database system initialization failed
Apr 04 07:42:27 pgext-patroni1 patroni[26555]: 2023-04-04 07:42:27,111 INFO: removing initialize key after failed attempt to bootstrap the cluster
Apr 04 07:42:27 pgext-patroni1 patroni[26555]: Traceback (most recent call last):
Apr 04 07:42:27 pgext-patroni1 patroni[26555]:   File "/usr/bin/patroni", line 11, in <module>
Apr 04 07:42:27 pgext-patroni1 patroni[26555]:     load_entry_point('patroni==3.0.2', 'console_scripts', 'patroni')()
[...]

PR #2633, opened 2023-04-04
Quote strings as values for initdb configuration

matthbakeredb:
A user may be confused by the fact that a string value of a configuration dictionary needs to be double-quoted.

If a value is not already quoted, supporting backwards compatibility with existing config, then quote it. Applies to POSIX compliant OSes only.

References: PAT-86 #2632

### DIFF
--- a/patroni/postgresql/bootstrap.py
+++ b/patroni/postgresql/bootstrap.py
@@ -55,7 +55,7 @@ class Bootstrap(object):
             >>> Bootstrap.process_user_options('foo', ['yes', {'foo': 'bar'}], (), print)
             ['--yes', '--foo=bar']
 
-            Options that contain spaces will passed as is to ``subprocess.call``
+            Options that contain spaces are passed as is to ``subprocess.call``
             >>> Bootstrap.process_user_options('foo', [{'foo': 'bar baz'}], (), print)
             ['--foo=bar baz']
 
@@ -72,7 +72,7 @@ class Bootstrap(object):
             * If the options list is not in the required structure.
 
         :param tool: The name of the tool used in error reports to *error_handler*
-        :param options: Options to parse as a list of key, values or single values or a dictionary
+        :param options: Options to parse as a list of key, values or single values, or a dictionary
         :param not_allowed_options: List of keys that cannot be used in the list of key, value formatted options
         :param error_handler: A function which will be called when an error condition is encountered
         :returns: List of long form arguments to pass to the named tool
@@ -86,9 +86,9 @@ class Bootstrap(object):
             return ret
 
         if isinstance(options, dict):
-            for k, v in options.items():
-                if k and v:
-                    user_options.append('--{0}={1}'.format(k, v))
+            for key, val in options.items():
+                if key and val:
+                    user_options.append('--{0}={1}'.format(key, unquote(val)))
         elif isinstance(options, list):
             for opt in options:
                 if isinstance(opt, str) and option_is_allowed(opt):
--- a/patroni/postgresql/bootstrap.py
+++ b/patroni/postgresql/bootstrap.py
@@ -33,6 +33,11 @@ class Bootstrap(object):
                              error_handler: Callable[[str], None]) -> List:
         """Format *options* in a list or dictionary format into command line long form arguments.
 
+        .. note::
+            The format of the output of this method is to prepare arguments for use in the ``pg_ctl``
+            method of `self._postgres`, which will join all options together into a single space delimited
+           string passed to the argument ``-o``.
+
         :Example:
 
             The *options* can be defined as a dictionary of key, values to be converted into arguments:
@@ -50,6 +55,14 @@ class Bootstrap(object):
             Or a combination of single and key, values
             >>> Bootstrap.process_user_options('foo', ['yes', {'foo': 'bar'}], (), print)
             ['--yes', '--foo=bar']
+
+            Options that contain spaces will be quoted
+            >>> Bootstrap.process_user_options('foo', [{'foo': 'bar baz'}], (), print)
+            ["--foo='bar baz'"]
+
+            Options that are already quoted will remain so
+            >>> Bootstrap.process_user_options('foo', [{'foo': '"bar baz"'}], (), print)
+            ['--foo="bar baz"']
 
         .. note::
             The *error_handler* is called when any of these conditions are met:
--- a/patroni/postgresql/bootstrap.py
+++ b/patroni/postgresql/bootstrap.py
@@ -6,7 +6,7 @@ import time
 
 from ..dcs import RemoteMember
 from ..psycopg import quote_ident, quote_literal
-from ..utils import deep_compare
+from ..utils import deep_compare, shell_quote
 
 logger = logging.getLogger(__name__)
 
@@ -48,7 +48,7 @@ class Bootstrap(object):
                     if len(keys) != 1 or not isinstance(opt[keys[0]], str) or not option_is_allowed(keys[0]):
                         error_handler('Error when parsing {0} key-value option {1}: only one key-value is allowed'
                                       ' and value should be a string'.format(tool, opt[keys[0]]))
-                    user_options.append('--{0}={1}'.format(keys[0], opt[keys[0]]))
+                    user_options.append('--{0}={1}'.format(keys[0], shell_quote(opt[keys[0]])))
                 else:
                     error_handler('Error when parsing {0} option {1}: value should be string value'
                                   ' or a single key-value pair'.format(tool, opt))
--- a/patroni/postgresql/bootstrap.py
+++ b/patroni/postgresql/bootstrap.py
@@ -3,6 +3,7 @@ import os
 import shlex
 import tempfile
 import time
+from typing import List, Dict, Union, Callable, Tuple
 
 from ..dcs import RemoteMember
 from ..psycopg import quote_ident, quote_literal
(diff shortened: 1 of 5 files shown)


tests/test_utils.py
```diff
@@ -41,6 +42,15 @@ def test_enable_keepalive(self):
                 with patch('sys.platform', platform):
                     self.assertIsNone(enable_keepalive(Mock(), 10, 5))
 
+    def test_shell_quote(self):
+        self.assertEqual(shell_quote('value'), 'value')
+        self.assertEqual(shell_quote('value with spaces'), "'value with spaces'")
```
CyberDem0n, line 47: It should be skipped on windows.

matthbakeredb: Ah, thanks! Will do that now.

CyberDem0n: What about testing values with a single/double quotes in the beginning/middle/end of the string (and different permutations of these combinations)?

It is important, because the regex checking that the provided string is already quoted is wrong.

patroni/postgresql/bootstrap.py
```diff
@@ -48,7 +48,7 @@ def option_is_allowed(name):
                     if len(keys) != 1 or not isinstance(opt[keys[0]], str) or not option_is_allowed(keys[0]):
                         error_handler('Error when parsing {0} key-value option {1}: only one key-value is allowed'
                                       ' and value should be a string'.format(tool, opt[keys[0]]))
-                    user_options.append('--{0}={1}'.format(keys[0], opt[keys[0]]))
+                    user_options.append('--{0}={1}'.format(keys[0], shell_quote(opt[keys[0]])))
```
CyberDem0n, line 51: Last time I touched this method was at least 4 years ago and already forgot some details and @hughcapet probably never seen it yet.
It would be nice to add a docstring and type hints to process_user_options method, it will help the review.

patroni/utils.py
```diff
@@ -911,3 +910,19 @@ def enable_keepalive(sock: socket.socket, timeout: int, idle: int, cnt: Optional
...
+    """Quote a string intended to be used as part of a shell command.
+
+    Applies quoting on POSIX compliant systems only, on `win32` platforms
+    will just return the original string unquoted.
+
+    :param opt: String to be quoted
+    :return: Quoted string
+
+    """
+    if sys.platform != 'win32' and not re.match(r'^[\'"]', opt):
```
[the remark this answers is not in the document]
matthbakeredb: Yes, very good point. I did think about whether this was good enough test or not and did play with trying to take an already quoted string and convert it back to unquoted to requote it again. So regardless of whether it was quoted or not it got the same treatment. There didn't seem to be an obvious way to do that so added a simple test to just ignore modifying it.

I'll look at adding another func to act as a more rigid test of "quotedness".

patroni/utils.py
```diff
@@ -911,3 +912,31 @@ def enable_keepalive(sock: socket.socket, timeout: int, idle: int, cnt: Optional
...
+
+    :param string: The input string to check.
+    :returns: ``True`` if the input string is a single fully quoted string, ``False`` otherwise.
+    """
+    lexer = shlex(string)
+    # Next 2 assignments ensure shlex does not split the string on whitespace
+    lexer.whitespace_split = False
+    lexer.whitespace = ''
+    tokens = list(lexer)
+    return len(tokens) == 1 and tokens[0][0] == tokens[0][-1] and tokens[0][0] in ('"', "'")
```
CyberDem0n, line 942: Unfortunately it doesn't work with one of my examples:
```python
In [1]: from shlex import quote, shlex

In [2]: a = 'asd \' ads'

In [3]: c = quote(a)

In [4]: lexer = shlex(c)

In [5]: lexer.whitespace_split = False

In [6]: lexer.whitespace = ''

In [7]: tokens = list(lexer)

In [8]: len(tokens)
Out[8]: 3  # <-- HERE
```

patroni/postgresql/bootstrap.py
```diff
@@ -48,7 +86,10 @@ def option_is_allowed(name):
                     if len(keys) != 1 or not isinstance(opt[keys[0]], str) or not option_is_allowed(keys[0]):
                         error_handler('Error when parsing {0} key-value option {1}: only one key-value is allowed'
                                       ' and value should be a string'.format(tool, opt[keys[0]]))
-                    user_options.append('--{0}={1}'.format(keys[0], opt[keys[0]]))
+                    try:
+                        user_options.append('--{0}={1}'.format(keys[0], shell_quote(opt[keys[0]])))
+                    except ValueError as err:
```
CyberDem0n, line 91: IMO this exception should be handled in the `is_single_quoted_string()` function by simply returning a `False`.

tests/test_utils.py
```diff
@@ -55,6 +86,7 @@ def inner():
             else:
                 scope['times'] += 1
                 raise PatroniException('Failed!')
+
```
CyberDem0n, line 89: Unrelated change.
```suggestion
```

patroni/utils.py
matthbakeredb: > Unfortunately it doesn't work with one of my examples:
...
> In [8]: len(tokens)
> Out[8]: 3  # <-- HERE
> ```

Hmnn, drat.

CyberDem0n: > Maybe we need a behave test for this.

I am not so sure. As far as I know TDE isn't available as an extension and installing it on GH is an additional effort.
IMO it would be enough just to check that both cases work fine (simplified, you need to add missing initdb arguments):
```python
subprocess.call(['initdb', '--key-wrap-command="openssl enc -e -pbkdf2 -out %p -pass pass:trustno1"'])
subprocess.call(["initdb", "--key-wrap-command='openssl enc -e -pbkdf2 -out %p -pass pass:trustno1'"])
```

CyberDem0n: `process_user_options()` already checks that the value is `str`, hence we always pass the `str` object to the `shell_quote()`.

patroni/utils.py
```diff
@@ -911,3 +912,35 @@ def enable_keepalive(sock: socket.socket, timeout: int, idle: int, cnt: Optional
...
+    """Quote a string intended to be used as part of a shell command.
+
+    Applies quoting on POSIX compliant systems only, on ``win32`` platforms
+    will just return the original string unquoted.
+
+    :param opt: String to be quoted
+    :returns: Quoted string
+    """
+    if sys.platform != 'win32' and requires_quoting(str(opt)):
+        return quote(str(opt))
```
CyberDem0n, line 927: `opt` is already a string.
```suggestion
    if sys.platform != 'win32' and requires_quoting(opt):
        return quote(opt)
```

patroni/utils.py
matthbakeredb: > > Maybe we need a behave test for this.
> 
> I am not so sure. As far as I know TDE isn't available as an extension and installing it on GH is an additional effort. IMO it would be enough just to check that both cases work fine (simplified, you need to add missing initdb arguments):
> 
> ```python
> subprocess.call(['initdb', '--key-wrap-command="openssl enc -e -pbkdf2 -out %p -pass pass:trustno1"'])
> subprocess.call(["initdb", "--key-wrap-command='openssl enc -e -pbkdf2 -out %p -pass pass:trustno1'"])
> ```

So, I ran this test to mimic what the existing bootstrap code is doing
```
#!/usr/bin/env python3

import sys
import subprocess

args = ['--encoding=UTF8',
        '--data-checksums',
        '--auth-local=peer',
        '--auth-host=scram-sha-256',
        '--data-encryption',
        '--key-wrap-command="openssl enc -e -pbkdf2 -out %p -pass pass:trustno1"',
        '--key-unwrap-command="openssl enc -d -pbkdf2 -in %p -pass pass:trustno1"']

options = ['-o', ' '.join(args)]

res = subprocess.call(['/usr/lib/edb-as/15/bin/pg_ctl', 'initdb', '-D', sys.argv[1]] + options)
```

I also inverted the quoting, double enclosing single. Both worked and resulted in a single quoted value in the postgresql.conf.
```
data_encryption_key_unwrap_command = 'openssl enc -d -pbkdf2 -in %p -pass pass:trustno1'
```

Note that a direct call to initdb does not work, as you gave above in your example to test, the quotes are interpreted literally. It results in an error
```
setting up data encryption ... sh: 1: openssl enc -e -pbkdf2 -out d2/pg_encryption/key.bin -pass pass:trustno1: not found
initdb: error: could not run command "'openssl enc -e -pbkdf2 -out d2/pg_encryption/key.bin -pass pass:trustno1'": command not found
```

So, this only works because the arguments are passed as one via the options var which combines all args as a single argument passed to pg_ctl. If this was handled directly as a call to initdb the args should be sent without quotes but separate items in the list of args.

```
#!/usr/bin/env python3

import sys
import subprocess

args = ['--encoding=UTF8', 
        '--data-checksums', 
        '--auth-local=peer', 
        '--auth-host=scram-sha-256', 
        '--data-encryption', 
        '--key-wrap-command=openssl enc -e -pbkdf2 -out %p -pass pass:trustno1', 
        '--key-unwrap-command=openssl enc -d -pbkdf2 -in %p -pass pass:trustno1']

res = subprocess.call(['/usr/lib/edb-as/15/bin/initdb'] + args + [sys.argv[1]])
```

CyberDem0n: Yeah, it seems that just calling initdb directly would simplify a lot of things.
I don't already recall why it was done like this. Probably because we just can...

patroni/utils.py
matthbakeredb: Sorry, I should have removed that before pushing again.

patroni/utils.py
matthbakeredb: I don't propose to change that now for this PR, although I would be happy to add that as something to look into if you think it's worthwhile?

matthbakeredb:
I've added some more docstring comments, I think this resolves all outstanding review points?

patroni/utils.py
CyberDem0n: Calling initdb directly  would eliminate the need to have shell_quote() and requires_quoting() all together.

CyberDem0n: I want to say - lets do it the right way.
Quoting-unquoting-checking seems to created many strange problems that better solved by avoiding them.

P.S., process_user_options() docstring is beautiful, we should keep it!

matthbakeredb: OK, I'll take a look at how to do that. It would reduce clutter.

matthbakeredb: Of course we have the possible case now where a user is relying on the current way this works, i.e. double quoting.

```
initdb:
  - key-wrap-command: "'openssl enc -e -pbkdf2 -out %p -pass pass:trustno1'"
```

If we want to be backwards compatible we'd need to `unquote` that string. Then we're back where we were before!

CyberDem0n: Oh, you already have it somewhere...

> we'd need to unquote that string

It could be as simple as:
```python
def unquote(value: str) -> str:
    try:
        ret = split(value)
        ret = ret[0] if len(ret) == 1 else value
    except Exception:
        ret = value
    return ret
```

matthbakeredb: OK pushed the code for this. I noticed that the options parser method used in bootstrap is also used by pg_basebackup. I think it should be OK though as it doesn't go through pg_ctl as initdb did.

CyberDem0n: Hmm, I don't see any new commits...

matthbakeredb: Gak, sorry pushed to our internal repo. Should be there now.

tests/test_utils.py
```diff
@@ -1,8 +1,10 @@
+import os
```
CyberDem0n, line 1: Unused import
```suggestion
```

patroni/postgresql/bootstrap.py
```diff
@@ -48,7 +98,7 @@ def option_is_allowed(name):
                     if len(keys) != 1 or not isinstance(opt[keys[0]], str) or not option_is_allowed(keys[0]):
                         error_handler('Error when parsing {0} key-value option {1}: only one key-value is allowed'
                                       ' and value should be a string'.format(tool, opt[keys[0]]))
-                    user_options.append('--{0}={1}'.format(keys[0], opt[keys[0]]))
+                    user_options.append('--{0}={1}'.format(keys[0], unquote(opt[keys[0]])))
```
hughcapet, line 101: shouldn't we do the same for the `options` passed as `dict`?

matthbakeredb: Yes, I think so. It was left off before as we thought perhaps it was used only in custom bootstrap methods but on second look I think it definitely applies.

Pushed a commit for this.

matthbakeredb:
btw I'm not sure what to do about the codeacy check failure. It seems to just be warning about the use of `subprocess.call` and doesn't lead to any solution to silence the error. If we are happy that "appropriate care is taken to sanitize any user input" we can add `# nosec B603` to the line.

It says that the input for a subprocess.call() should be verified but doesn't give any examples of how. It states that it checks for non-shell use, and that it's not vulnerable to shell injection techniques. The "see also" all refer to shell=True use cases, so don't seem relevant.

https://app.codacy.com/gh/zalando/patroni/pullRequest?prid=11497352

hughcapet:
I disabled this pattern (`
subprocess_without_shell_equals_true`) for the whole project

hughcapet:
👍

CyberDem0n:
:+1:

PR #2634, opened 2023-04-06
Add docstrings and type annotations to `patroni/psycopg.py`

barthisrael:
References: PAT-72

### DIFF
--- a/patroni/psycopg.py
+++ b/patroni/psycopg.py
@@ -3,7 +3,7 @@
 This module is able to handle both ``pyscopg2`` and ``psycopg3``, and it exposes a common interface for both.
 ``psycopg2`` takes precedence. ``psycopg3`` will only be used if ``psycopg2`` is either absent or older than ``2.5.4``.
 """
-from typing import Any, Optional
+from typing import Any, Optional, TYPE_CHECKING
 
 
 __all__ = ['connect', 'quote_ident', 'quote_literal', 'DatabaseError', 'Error', 'OperationalError', 'ProgrammingError']
@@ -15,12 +15,15 @@ try:
     if parse_version(__version__) < MIN_PSYCOPG2:
         raise ImportError
     from psycopg2 import connect as _connect, Error, DatabaseError, OperationalError, ProgrammingError
-    from psycopg2.extensions import adapt, connection
+    from psycopg2.extensions import adapt
 
     try:
         from psycopg2.extensions import quote_ident as _quote_ident
     except ImportError:
         _legacy = True
+
+    if TYPE_CHECKING:
+        from psycopg2.extensions import connection
 
     def quote_literal(value: Any, conn: Optional[connection] = None) -> str:
         """Quote *value* as a SQL literal.
--- a/patroni/psycopg.py
+++ b/patroni/psycopg.py
@@ -45,7 +45,7 @@ except ImportError:
         """Call ``psycopg.connect`` with ``*args`` and ``**kwargs``.
 
         .. note::
-            Will create ``server_version`` attribute in the returning connection so it keeps compatibility with the
+            Will create ``server_version`` attribute in the returning connection, so it keeps compatibility with the
             object that would be returned by ``psycopg2.connect``.
 
         :param args: positional arguments to call ``psycopg.connect`` with.
--- a/patroni/psycopg.py
+++ b/patroni/psycopg.py
@@ -3,7 +3,7 @@
 This module is able to handle both ``pyscopg2`` and ``psycopg3``, and it exposes a common interface for both.
 ``psycopg2`` takes precedence. ``psycopg3`` will only be used if ``psycopg2`` is either absent or older than ``2.5.4``.
 """
-from typing import Any, Optional, TYPE_CHECKING
+from typing import Any, Optional
 
 
 __all__ = ['connect', 'quote_ident', 'quote_literal', 'DatabaseError', 'Error', 'OperationalError', 'ProgrammingError']
@@ -22,10 +22,7 @@ try:
     except ImportError:
         _legacy = True
 
-    if TYPE_CHECKING:
-        from psycopg2.extensions import connection
-
-    def quote_literal(value: Any, conn: Optional["connection"] = None) -> str:
+    def quote_literal(value: Any, conn: Optional[Any] = None) -> str:
         """Quote *value* as a SQL literal.
 
         .. note::
@@ -44,8 +41,8 @@ try:
 except ImportError:
     from psycopg import connect as __connect, sql, Error, DatabaseError, OperationalError, ProgrammingError, Connection
 
-    def _connect(*args: Any, **kwargs: Any) -> Connection[Any]:
-        """Call ``psycopg.connect`` with ``*args`` and ``**kwargs``.
+    def _connect(dsn: str, **kwargs: Any) -> Connection[Any]:
+        """Call ``psycopg.connect`` with ``dsn`` and ``**kwargs``.
 
         .. note::
             Will create ``server_version`` attribute in the returning connection, so it keeps compatibility with the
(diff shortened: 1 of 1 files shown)


patroni/psycopg.py
```diff
@@ -14,27 +22,79 @@
...
+    .. note::
+        The connection will have ``autocommit`` enabled.
+
+    :param args: positional arguments to call ``connect`` function from ``psycopg`` module.
+    :param kwargs: keyword arguments to call ``connect`` function from ``psycopg`` module.
+
+    :returns: a connection to the database. Can be either a :class:`psycopg.Connection` if using ``psycopg3``, or a
+        :class:`psycopg2.extensions.connection` if using ``psycopg2``.
+    """
+    # Enforce search_path=pg_catalog for non-replication connections to mitigate security issues as Patroni relies on
```
CyberDem0n, line 96: Should this comment be a part of docstring?

patroni/psycopg.py
```diff
@@ -44,7 +104,15 @@ def connect(*args, **kwargs):
     return ret
 
 
-def quote_ident(value, conn=None):
+def quote_ident(value: Any, conn: Any = None):
```
CyberDem0n, line 107:
```suggestion
def quote_ident(value: Any, conn: Optional[Any] = None):
```

patroni/psycopg.py
```diff
@@ -14,27 +22,79 @@
     except ImportError:
         _legacy = True
 
-    def quote_literal(value, conn=None):
+    if TYPE_CHECKING:
+        from psycopg2.extensions import connection
+
+    def quote_literal(value: Any, conn: Optional["connection"] = None) -> str:
```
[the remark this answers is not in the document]
barthisrael: We were on doubt about how to deal with that, I mean, if we should use `TYPE_CHECKING` when there is no issue with circular importing, but we are only importing `connection` because of type hinting. What are your thoughts on that?

patroni/psycopg.py
barthisrael: I think that makes sense!

patroni/psycopg.py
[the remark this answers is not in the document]
CyberDem0n: But it this case there are no circular dependencies, we already imported from `psycopg2.extensions` on the line 18.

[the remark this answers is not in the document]
matthbakeredb: Yes, that we understood. What we were thinking was whether to make it clear what is imported _just_ for type checking and doesn't form part of the functional code. It seems that `TYPE_CHECKING` is mainly used to avoid circular dependencies and in some cases to avoid large imports for performance reasons, but it appears that, as a dev looking at the import list, that it might be advantageous to separate these use cases, as a documentation aid.

If that doesn't make sense, then yes this should be on line 18.

patroni/psycopg.py
```diff
@@ -14,27 +22,80 @@
...
```
CyberDem0n, line 28: Actually... the difference between psycopg2 connection and psycopg3 Connection makes pyright crazy...
Hence I suggest the following:
```suggestion
    def quote_literal(value: Any, conn: Optional[Any] = None) -> str:
```

patroni/psycopg.py
```diff
@@ -14,27 +22,80 @@
...
         value = adapt(value)
         if conn:
             value.prepare(conn)
         return value.getquoted().decode('utf-8')
 except ImportError:
-    from psycopg import connect as __connect, sql, Error, DatabaseError, OperationalError, ProgrammingError
+    from psycopg import connect as __connect, sql, Error, DatabaseError, OperationalError, ProgrammingError, Connection
+
+    def _connect(*args: Any, **kwargs: Any) -> Connection[Any]:
+        """Call ``psycopg.connect`` with ``*args`` and ``**kwargs``.
```
CyberDem0n, line 48:
```suggestion
    def _connect(dsn: str, **kwargs: Any) -> Connection[Any]:
        """Call ``psycopg.connect`` with ``dsn`` and ``**kwargs``.
```

patroni/psycopg.py
```diff
@@ -14,27 +22,80 @@
...
-    def _connect(*args, **kwargs):
+        :returns: a connection to the database.
+        """
         ret = __connect(*args, **kwargs)
-        ret.server_version = ret.pgconn.server_version  # compatibility with psycopg2
+        setattr(ret, 'server_version', ret.pgconn.server_version)  # compatibility with psycopg2
         return ret
 
-    def _quote_ident(value, conn):
+    def _quote_ident(value: Any, conn: Connection[Any]) -> str:
```
CyberDem0n, line 63:
```suggestion
    def _quote_ident(value: Any, conn: Connection) -> str:
```

patroni/psycopg.py
```diff
@@ -14,27 +22,80 @@
...
+
+        :param value: value to be quoted.
+        :param conn: connection to evaluate the returning string into.
+
+        :returns: *value* quoted as a SQL identifier.
+        """
         return sql.Identifier(value).as_string(conn)
 
-    def quote_literal(value, conn=None):
+    def quote_literal(value: Any, conn: Optional[Connection[Any]] = None) -> str:
```
CyberDem0n, line 73:
```suggestion
    def quote_literal(value: Any, conn: Optional[Any] = None) -> str:
```

patroni/psycopg.py
```diff
@@ -1,3 +1,11 @@
+"""Abstraction layer for ``psycopg`` module.
+
+This module is able to handle both ``pyscopg2`` and ``psycopg3``, and it exposes a common interface for both.
+``psycopg2`` takes precedence. ``psycopg3`` will only be used if ``psycopg2`` is either absent or older than ``2.5.4``.
+"""
+from typing import Any, Optional, TYPE_CHECKING
```
CyberDem0n, line 6:
```suggestion
from typing import Any, Optional
```

CyberDem0n (commented):
Remove most of pyright complaints

patroni/psycopg.py
barthisrael: I applied the suggestion, but by removing `[Any]` I started getting this again in `pyright`:

```
  /Users/israel.barth/Repositories/edb/edb-patroni/patroni/psycopg.py:60:34 - error: Type of parameter "conn" is partially unknown
    Parameter type is "Connection[Unknown]" (reportUnknownParameterType)
```

CyberDem0n: Strange. I don't have it:
```bash
$ pyright patroni/psycopg.py
No configuration file found.
No pyproject.toml file found.
Assuming Python platform Linux
Searching for source files
Found 1 source file
pyright 1.1.302
/home/akukushkin/git/patroni/patroni/psycopg.py
  /home/akukushkin/git/patroni/patroni/psycopg.py:13:26 - error: "__version__" is unknown import symbol (reportGeneralTypeIssues)
1 error, 0 warnings, 0 informations 
Completed in 0.764sec

```

patroni/psycopg.py
barthisrael: By using `dsn` instead of `*args`, behave test start to fail because they don't use `dsn` parameter:

```
        File "features/environment.py", line 283, in _connection
          self._conn = psycopg.connect(**self._connkwargs)
        File "/home/runner/work/patroni/patroni/patroni/psycopg.py", line 100, in connect
          ret = _connect(*args, **kwargs)
      TypeError: _connect() missing 1 required positional argument: 'dsn'
  ```
  
  I attempted a few different options to overcome that issue, but was not able yet.
  
  I wonder if we should just acknowledge that pyright issue?

Comment (CyberDem0n) [reply patroni/psycopg.py]: Sorry, my bad. I didn't try it with behave tests :(
`psycopg.connect()` expects `conninfo` doesn't like None. Setting dsn to en empty string fixes it:
```python
    def _connect(dsn: str = "", **kwargs: Any) -> Any:
```

barthisrael: I can't believe I have not tried with an empty string 😩 . I attempted with `None`, but not with an empty string. Thanks for that!

patroni/psycopg.py
barthisrael: Strange indeed! I'm running the very same `pyright` version.

matthbakeredb: fwiw I ran pyright on that file with all supported python versions and don't see that "partially unknown" message

CyberDem0n:
👍

PR #2636, opened 2023-04-06
Add docstrings and type annotations to `patroni/log.py`

barthisrael:
References: PAT-74

### DIFF
--- a/patroni/log.py
+++ b/patroni/log.py
@@ -20,7 +20,7 @@ _LOGGER = logging.getLogger(__name__)
 
 def debug_exception(logger_obj: logging.Logger, msg: str, *args: Any, **kwargs: Any) -> None:
     """Add full stack trace info to debug log messages and partial to others.
-    
+
     Handle :func:`exception` calls for *logger_obj*.
 
     .. note::
@@ -43,11 +43,12 @@ def debug_exception(logger_obj: logging.Logger, msg: str, *args: Any, **kwargs: 
 
 def error_exception(logger_obj: logging.Logger, msg: str, *args: Any, **kwargs: Any) -> None:
     """Add full stack trace info to error messages.
-    
+
     Handle :func:`exception` calls for *logger_obj*.
 
     .. note::
-        * Issue an ``ERROR`` message with the complete stack trace.
+        * By default issue an ``ERROR`` message with the complete stack trace. If you do not want to show the complete
+        stack trace, call with ``exc_info=False``.
 
     :param logger_obj: logger for which :func:`exception` will be processed.
     :param msg: the message related to the exception to be logged.
@@ -119,7 +120,7 @@ class QueueHandler(logging.Handler):
 
 
 class ProxyHandler(logging.Handler):
-    """Handle log records in place of pending log handlers. 
+    """Handle log records in place of pending log handlers.
 
     .. note::
         This is used to handle log messages while the logger thread has not started yet, in which case the queue-based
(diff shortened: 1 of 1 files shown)


patroni/log.py
```diff
@@ -82,41 +176,71 @@ class PatroniLogger(Thread):
...
+
+        :param config: ``log`` section from Patroni configuration.
+        """
         if self._config is None or not deep_compare(self._config, config):
             with self._queue_handler.queue.mutex:
                 self._queue_handler.queue.maxsize = config.get('max_queue_size', self.DEFAULT_MAX_QUEUE_SIZE)
 
             self._root_logger.setLevel(config.get('level', PatroniLogger.DEFAULT_LEVEL))
             if config.get('traceback_level', PatroniLogger.DEFAULT_TRACEBACK_LEVEL).lower() == 'debug':
+                # show stack traces only if ``log.level`` is ``DEBUG``
```
CyberDem0n, line 240:
```suggestion
                # show stack traces only if ``log.traceback_level`` is ``DEBUG``
```

CyberDem0n (commented):
There are couple of more places where pyright issues should be fixed, but GH doesn't allow to make suggestions :(

```diff
@@ -248,6 +249,7 @@ class PatroniLogger(Thread):
                 if not isinstance(self.log_handler, RotatingFileHandler):
                     new_handler = RotatingFileHandler(os.path.join(config['dir'], __name__))
                 handler = new_handler or self.log_handler
+                assert isinstance(handler, RotatingFileHandler)
                 handler.maxBytes = int(config.get('file_size', 25000000))
                 handler.backupCount = int(config.get('file_num', 4))
             else:
@@ -261,7 +263,7 @@ class PatroniLogger(Thread):
             olddateformat = (self._config or {}).get('dateformat') or None
             dateformat = config.get('dateformat') or None  # Convert empty string to `None`

-            if oldlogformat != logformat or olddateformat != dateformat or new_handler:
+            if (oldlogformat != logformat or olddateformat != dateformat or new_handler) and handler:
                 handler.setFormatter(logging.Formatter(logformat, dateformat))

             if new_handler:
```

Unfortunately even after that there is one problem is left:
```python
patroni/log.py:253:36 - error: Cannot assign member "maxBytes" for type "RotatingFileHandler"
    "int" is incompatible with "str" (reportGeneralTypeIssues)
```

But, it more looks like a bug in pyright, for some reason it assumes that type is str, while in fact it is assigned from int: https://github.com/microsoft/pyright/blob/eb98cdda4ecfb4d2ce2fb1d4b9ce7848ab439c32/packages/pyright-internal/typeshed-fallback/stdlib/logging/handlers.pyi#L47-L54

barthisrael:
Thanks for all suggestions, and sorry about pyright. I'll make sure to run it for other PRs.

Should we leave this one aside for now?

```python
/Users/israel.barth/Repositories/edb/edb-patroni/patroni/log.py
  /Users/israel.barth/Repositories/edb/edb-patroni/patroni/log.py:253:36 - error: Cannot assign member "maxBytes" for type "RotatingFileHandler"
    "int" is incompatible with "str" (reportGeneralTypeIssues)
```

CyberDem0n:
> Cannot assign member "maxBytes" for type "RotatingFileHandler"

I think for now we can leave it as it is.

patroni/log.py
```diff
@@ -50,29 +98,76 @@ def _try_to_report_lost_records(self):
...
 
-    def emit(self, record):
+    def emit(self, record: logging.LogRecord) -> None:
+        """Handle each log record that is emitted.
+
+        Call :func:`_put_record` to enqueue the emitted log record.
+
+        Also check if we have previously lost any log record, and if so, log a ``WARNING`` message.
+
+        :param record: the recorded that was emitted.
```
hughcapet, line 108: the **record** that was emitted?

CyberDem0n:
:+1:

Issue #2638, opened 2023-04-10
`--dcs-url` command-line option of `patronictl` does not cover `etcd3` DCS


barthisrael:
### What happened?

I attempted to use `--dcs-url` commad-line option of `patronictl` command to specify the DCS URI but it did not work -- it throws an error about unknown scheme:

```bash
$ patronictl -c postgres0.yml --dcs-url "etcd://localhost:2379" list batman
+ Cluster: batman -----+-------+----+-----------+
| Member | Host | Role | State | TL | Lag in MB |
+--------+------+------+-------+----+-----------+
+--------+------+------+-------+----+-----------+

$ patronictl -c postgres0.yml --dcs-url "etcd3://localhost:2379" list batman
Error: Unknown dcs scheme: etcd3
```

### How can we reproduce it (as minimally and precisely as possible)?

Execute the following command:

```bash
patronictl -c postgres0.yml --dcs-url "etcd://localhost:2379" list batman
```

### What did you expect to happen?

Output to be:

```bash
$ patronictl -c postgres0.yml --dcs-url "etcd3://localhost:2379" list batman
+ Cluster: batman --------+--------+---------+----+-----------+
| Member      | Host      | Role   | State   | TL | Lag in MB |
+-------------+-----------+--------+---------+----+-----------+
| postgresql0 | 127.0.0.1 | Leader | running |  2 |           |
+-------------+-----------+--------+---------+----+-----------+
```

### Patroni/PostgreSQL/DCS version

-   Patroni version: 3.0.2
-   PostgreSQL version: 15.2
-   DCS (and its version): 3.2.26

### Patroni configuration file

```yaml
scope: batman
name: postgresql0

restapi:
  listen: 127.0.0.1:8008
  connect_address: 127.0.0.1:8008
  authentication:
    username: username
    password: password


etcd3:
  host: 127.0.0.1:2379


bootstrap:
  dcs:
    ttl: 30
    loop_wait: 10
    retry_timeout: 10
    maximum_lag_on_failover: 1048576
    postgresql:
      use_pg_rewind: true
      parameters:

  initdb:  # Note: It needs to be a list (some options need values, others are switches)
  - encoding: UTF8
  - data-checksums

  pg_hba:  # Add following lines to pg_hba.conf after running 'initdb'
  - host replication replicator 127.0.0.1/32 md5
  - host all all 0.0.0.0/0 md5


  users:
    admin:
      password: admin%
      options:
        - createrole
        - createdb

postgresql:
[...]

PR #2639, opened 2023-04-10
Cover `etcd3` in `parse_dcs` function

barthisrael:
Previous to this commit the `parse_dcs` function would fail with a `PatroniCtlException` if the user ever passed an `etcd3` URL through `--dcs-url` command-line option.

As a consequence, the only way of using `etcd3` in `patronictl` was by using a configuration file passed through `-c` command-line option.

This commit fixes that issue and allows `etcd3` to be used in `--dcs-url`.

References: PAT-91 #2638

### DIFF
--- a/patroni/ctl.py
+++ b/patroni/ctl.py
@@ -21,6 +21,8 @@ import tempfile
 import time
 import yaml
 
+from typing import Any, Dict, Union
+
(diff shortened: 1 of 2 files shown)


hughcapet:
👍

Issue #2596, opened 2023-03-10
inquiry: pre-stop callback


le-duane:
Hi,

Sorry, this isn't a bug report, and I couldn't figure out how to get an invite to the Postgres Slack Workspace.

My team is attempting to move to ECTD as our DCS from Kubernetes. We would like to leverage callbacks to manage the role labels for our pods. We also from time to time have logical replication connections which hang shutdown, and would like to kill these before shutdown as well. However, the `on_stop` callback is too late for both of these things.

My thought was to do this via a pre-stop callback before the actual shutdown of the DB; perhaps [here](https://github.com/zalando/patroni/blob/60a7e5a514b1a9982c1ff1f9ee78f644a4e820bb/patroni/postgresql/__init__.py#L699)?

I would like to start a discussion of whether this is feasible or perhaps whether this is a better way to achieve our goals here.

Thanks!

hughcapet:
> I couldn't figure out how to get an invite to the Postgres Slack Workspace

You can join using the link provided here: https://www.postgresql.org/community/

le-duane:
Thanks @hughcapet - that got me going!

PR #2642, opened 2023-04-11
Add before stop hook

le-duane:
We'd like the ability to run scripts prior to shutting down our Postgres instances.

The two cases we have in mind are:
* In spite of following all best practices client-side, logical replication connections can sometimes hang the Postgres shutdown sequence. We'd like to sigterm any misbehaving logical replication connections which remain after x seconds. These will inevitably get killed anyway on master stop timeout.
* remove "role=master" label on current primary when not using k8s as DCS. Waiting until after Postgres fully stops can sometimes be too long for this.

#2596

### DIFF
--- a/patroni/postgresql/__init__.py
+++ b/patroni/postgresql/__init__.py
@@ -12,7 +12,7 @@ from datetime import datetime
 from dateutil import tz
 from psutil import TimeoutExpired
 from threading import current_thread, Lock
-from typing import Optional, Union, TYPE_CHECKING
+from typing import Any, Dict, List, Optional, Union, TYPE_CHECKING
 
 from .bootstrap import Bootstrap
 from .callback_executor import CallbackAction, CallbackExecutor
@@ -30,7 +30,7 @@ from ..exceptions import PostgresConnectionException
 from ..utils import Retry, RetryFailedError, polling_loop, data_directory_is_empty, parse_int
 
 if TYPE_CHECKING:  # pragma: no cover
-    from .config import GlobalConfig
+    from ..config import GlobalConfig
 
 logger = logging.getLogger(__name__)
 
@@ -156,6 +156,11 @@ class Postgresql(object):
     @property
     def lsn_name(self):
         return 'lsn' if self._major_version >= 100000 else 'location'
+
+    @property
+    def supports_multiple_sync(self) -> bool:
+        """:returns: `True` if Postgres version supports more than one synchronous node."""
+        return self._major_version >= 90600
 
     @property
     def cluster_info_query(self):
@@ -419,13 +424,16 @@ class Postgresql(object):
     def received_timeline(self):
         return self._cluster_info_state_get('received_tli')
 
-    def synchronous_commit(self):
+    def synchronous_commit(self) -> str:
+        """:returns: "synchronous_commit" GUC value."""
         return self._cluster_info_state_get('synchronous_commit')
 
-    def synchronous_standby_names(self):
+    def synchronous_standby_names(self) -> str:
+        """:returns: "synchronous_standby_names" GUC value."""
         return self._cluster_info_state_get('synchronous_standby_names')
 
-    def pg_stat_replication(self):
+    def pg_stat_replication(self) -> List[Dict[str, Any]]:
+        """:returns: a result set of 'SELECT * FROM pg_stat_replication'."""
         return self._cluster_info_state_get('pg_stat_replication') or []
 
     def is_leader(self):
@@ -924,7 +932,8 @@ class Postgresql(object):
             self._cached_replica_timeline = self.get_replica_timeline()
         return self._cached_replica_timeline
 
-    def get_primary_timeline(self):
+    def get_primary_timeline(self) -> int:
+        """:returns: current timeline if postgres is running as a primary or 0."""
         return self._cluster_info_state_get('timeline')
 
     def get_history(self, timeline):
--- a/patroni/postgresql/__init__.py
+++ b/patroni/postgresql/__init__.py
@@ -233,6 +233,17 @@ class Postgresql(object):
         pg_ctl = [self.pgcommand('pg_ctl'), cmd]
         return subprocess.call(pg_ctl + ['-D', self._data_dir] + list(args), **kwargs) == 0
 
+    def initdb(self, *args: str, **kwargs: Any) -> bool:
+        """Builds and executes the initdb command.
+
+        :param args: List of arguments to be joined into the initdb command.
+        :param kwargs: Keyword arguments to pass to ``subprocess.call``.
+
+        :returns: ``True`` if the result of ``subprocess.call`, the exit code, is ``0``.
+        """
+        initdb = [self.pgcommand('initdb')] + list(args) + [self.data_dir]
+        return subprocess.call(initdb, **kwargs) == 0
+
     def pg_isready(self):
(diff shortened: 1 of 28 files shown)


CyberDem0n:
Thanks for working on it @le-duane.

Here are my thoughts. Currently callbacks are supposed to be fired and forgot and they will be running in parallel with whatever Patroni does with Postgres.
In some cases they are indeed executed only after reaching certain stage:
- on_start is executed when Postgres starts accepting connections
- on_stop is executed when Postgres is already stopped

But, right after `on_stop` there could be `on_start` and therefore the newly executed callback usually cancels the one that already running (Patroni kills the process and its all children).

The `on_pre_stop` feels like a bit special. There is one very useful use-case when we would like to run it not as fire-and-forget, but actually wait for the result. More specifically I am talking about `pgbouncer PAUSE`.

Of course, after promoting the new node (or when Postgres is back running as a primary after a restart) we want to do `pgbouncer RESUME`, but this one we don't need to wait for.

Also, we already have `postgresql.pre_promote` hook, which exit code is actually being checked and `pg_ctl promote` isn't called if the exit code is different from 0.

Having said that I'm thinking that `on_pre_stop` should not be a callback, but something similar to `pre_promote`, because we may want to wait until it finished before executing `pg_ctl stop -mf`. And probably `before_stop` would be a bit better name.

le-duane:
Thanks for the quick reply @CyberDem0n - I'll look into `pre_promote`

patroni/postgresql/__init__.py
```diff
@@ -1024,6 +1027,20 @@ def _pre_promote(self):
...
+        """
+
+        cmd = self.config.get('before_stop')
+        if not cmd:
+            return True
+
+        ret = self.cancellable.call(shlex.split(cmd))
+        if ret is not None:
+            logger.info('before_stop script `%s` exited with %s', cmd, ret)
+
```
CyberDem0n, line 1042:
```suggestion
        try:
            ret = self.cancellable.call(shlex.split(cmd))
            if ret is not None:
                logger.info('before_stop script `%s` exited with %s', cmd, ret)
        except Exception as e:
            logger.error('Exception when calling `%s`: %r', cmd, e)
```

le-duane:
Thanks for the tips @CyberDem0n. I'm a lot happier that this is running synchronously now. 

I changed things a little so I could add a test.

patroni/postgresql/__init__.py
```diff
@@ -1024,6 +1027,25 @@ def _pre_promote(self):
...
 
+    def _before_stop(self):
+        """
+        Synchronously run a script prior to stopping postgres
+        """
+
+        cmd = self.config.get('before_stop')
+        if not cmd:
+            return True
+        else:
```
CyberDem0n, line 1038:
```suggestion
        if cmd:
```

tests/test_postgresql.py
```diff
@@ -195,6 +196,17 @@ def test_stop(self, mock_is_running):
...
+        # Ensure before_stop script is called only when configured to
+        mock_postmaster.wait.side_effect = [psutil.TimeoutExpired(30), Mock()]
+        self.p.config._config['before_stop'] = ':'
+        self.p.stop(on_safepoint=mock_callback, stop_timeout=30)
+        self.assertTrue(self.p._do_before_stop.call_count == 1)
+        self.p._do_before_stop.reset_mock()
+        self.p.config._config.pop('before_stop')
+        mock_postmaster.wait.side_effect = [psutil.TimeoutExpired(30), Mock()]
+        self.p.stop(on_safepoint=mock_callback, stop_timeout=30)
+        self.assertTrue(self.p._do_before_stop.call_count == 0)
```
CyberDem0n, line 208:
```suggestion
        # Ensure before_stop script is called when configured to                                                                       
        self.p.config._config['before_stop'] = ':'                                                                                     
        mock_postmaster.wait.side_effect = [psutil.TimeoutExpired(30), Mock()]                                                         
        mock_cancellable_call.return_value = 0                                                                                         
        with patch('patroni.postgresql.logger.info') as mock_logger:                                                                   
            self.p.stop(on_safepoint=mock_callback, stop_timeout=30)                                                                   
            self.assertEqual(mock_logger.call_args[0], ('before_stop script `%s` exited with %s', ':', 0))                             
        mock_postmaster.wait.side_effect = [psutil.TimeoutExpired(30), Mock()]                                                         
        mock_cancellable_call.side_effect = Exception                                                                                  
        with patch('patroni.postgresql.logger.error') as mock_logger:                                                                  
            self.p.stop(on_safepoint=mock_callback, stop_timeout=30)                                                                   
            self.assertEqual(mock_logger.call_args_list[1][0][0], 'Exception when calling `%s`: %r')                                   
```

patroni/postgresql/__init__.py
```diff
@@ -1024,6 +1027,25 @@ def _pre_promote(self):
...
+            self._do_before_stop(cmd)
+
+    def _do_before_stop(self, cmd):
```
CyberDem0n, line 1041:
```suggestion
    def _do_before_stop(self, cmd: str) -> None:
```

CyberDem0n (commented):
Please also update https://github.com/zalando/patroni/blob/ff6d728f07e9a378c595115dabe11b0ca8dba84b/patroni/validator.py#L757 as `Optional("before_strop"): str`

tests/test_postgresql.py
```diff
@@ -195,6 +196,19 @@ def test_stop(self, mock_is_running):
...
+        with patch('patroni.postgresql.logger.info') as mock_logger:                                                                   
+            self.p.stop(on_safepoint=mock_callback, stop_timeout=30)                                                                   
+            self.assertEqual(mock_logger.call_args[0], ('before_stop script `%s` exited with %s', ':', 0))                             
+        mock_postmaster.wait.side_effect = [psutil.TimeoutExpired(30), Mock()]                                                         
+        mock_cancellable_call.side_effect = Exception                                                                                  
+        with patch('patroni.postgresql.logger.error') as mock_logger:                                                                  
+            self.p.stop(on_safepoint=mock_callback, stop_timeout=30)                                                                   
+            self.assertEqual(mock_logger.call_args_list[1][0][0], 'Exception when calling `%s`: %r')                                   
+
```
CyberDem0n, line 211:
```suggestion
        # Ensure before_stop script is called when configured to
        self.p.config._config['before_stop'] = ':'
        mock_postmaster.wait.side_effect = [psutil.TimeoutExpired(30), Mock()]
        mock_cancellable_call.return_value = 0
        with patch('patroni.postgresql.logger.info') as mock_logger:
            self.p.stop(on_safepoint=mock_callback, stop_timeout=30)
            self.assertEqual(mock_logger.call_args[0], ('before_stop script `%s` exited with %s', ':', 0))
        mock_postmaster.wait.side_effect = [psutil.TimeoutExpired(30), Mock()]
        mock_cancellable_call.side_effect = Exception
        with patch('patroni.postgresql.logger.error') as mock_logger:
            self.p.stop(on_safepoint=mock_callback, stop_timeout=30)
            self.assertEqual(mock_logger.call_args_list[1][0][0], 'Exception when calling `%s`: %r')

```

hughcapet:
👍

CyberDem0n:
:+1:

CyberDem0n:
Merged, thank you @le-duane!

PR #2648, opened 2023-04-20
Add docstrings and type hints to `patroni/api.py`

barthisrael:
References: PAT-77


barthisrael:
I still need to address `pyright` issues on this file, just noticed it is throwing quite a few of them.

barthisrael:
`878af423e01e79039008091828a26c9af698ccd9` introduced issues to unit tests because of `MockRestApiServer`. Any better idea?

barthisrael:
I've been fixing several things pointed by `pyright`.
There are a few left that I'm not yet sure about how to address properly.
Also, there is the issue with unit tests that started after addressing one of the issues reported by `pyright`.
Any thoughts on the remaining issue? Thanks in advance!

CyberDem0n:
Instead of adding assert to every method we can just redefine constructor:
```diff
diff --git a/patroni/api.py b/patroni/api.py
index afd25b4..66acab6 100644
--- a/patroni/api.py
+++ b/patroni/api.py
@@ -41,6 +41,12 @@ logger = logging.getLogger(__name__)
 class RestApiHandler(BaseHTTPRequestHandler):
     """Define how to handle each of the requests that are made against the REST API server."""
 
+    def __init__(self, request: Union[socket.socket, Tuple[bytes, socket.socket]],
+                 client_address: Tuple[str, int], server: Union['RestApiServer', HTTPServer]) -> None:
+        assert isinstance(server, RestApiServer)
+        super(RestApiHandler, self).__init__(request, client_address, server)
+        self.server: 'RestApiServer' = server
+
     def _write_status_code_only(self, status_code: int) -> None:
         """Write a response that is composed only of the HTTP status.
```

CyberDem0n:
And by the way, last week I spent adding type hints to `patroni.dcs` modules: https://github.com/CyberDem0n/patroni/pull/11
For now only `raft.py` is not yet covered, and some minor things are left to address or silence.

barthisrael:
Thanks for your suggestion, that's way cleaner, and avoid issues with the unit tests :)
Great news about `patroni.dcs` too :)

barthisrael:
I noticed unit tests started to fail after using `staticmethod` decorator in `check_access` function -- `3de6373eda5fac89c03ebeb044a4b2e78239a934`.

From what I could see, that change was introduced to silence these errors from `pyright`:

```
  patroni/api.py:121:22 - warning: Instance methods should take a "self" parameter (reportSelfClsParameterName)
  patroni/api.py:121:28 - error: Type of parameter "func" must be a supertype of its class "RestApiHandler" (reportGeneralTypeIssues)
```

Maybe we should just move that decorator function (`check_access`) out of the `RestApiHandler` class, with no `staticmethod` decorator? That would silent `pyright` and make unit tests to work.

matthbakeredb:
There are still a couple of errors on codacy and I locally have warnings about those same method's signature not matching their subclass

patroni/api.py
```diff
@@ -908,21 +1493,48 @@ def __initialize(self, listen, ssl_options):
...
+        """Get serial number of the certificate used by the REST API.
+
+        :returns: serial number of the certificate configured through ``restapi.certfile`` setting.
+        """
+        # pyright -- ``__ssl_options`` is initially created as ``None``, but right after that it is replaced with a
+        # dictionary through :func:`reload_config`.
+        assert isinstance(self.__ssl_options, dict)
         if self.__ssl_options.get('certfile'):
             import ssl
             try:
```
matthbakeredb, line 1540: imo this should be replaced with something that doesn't use the underlying `_ssl` shared library module (a protected member). e.g.

```
>>> from cryptography import x509
>>> c = x509.load_pem_x509_certificate(open(cert, 'rb').read())
>>> c.serial_number
4835703278459707669005204
```
I guess that depends on whether that cert is in pem format and it's a server cert. If this is used for client cert or der format then detection might be needed.

CyberDem0n: There is a reason why it is done this way: https://github.com/zalando/patroni/pull/1887#discussion_r597204638

barthisrael:
Fixed the 3 Codacy issues, but one more came up as a consequence:

```
Redefining built-in 'format'
def log_message(self, format: str, *args) -> None:
```

But that's because of how `log_message` is defined in the standard `http` module, as we are overloading the method.

CyberDem0n (commented):
I just started reviewing, please expect more comments.
Also, currently pyright complains about many unknown types because hints are missing for Postgresql and HA methods.

patroni/api.py
matthbakeredb: Perhaps that reason could be reviewed?

```
$ grep crypto requirements.txt 
cryptography>=1.4
```

patroni/api.py
```diff
@@ -430,11 +732,41 @@ def parse_schedule(schedule, action):
...
+
+        The request body should be a JSON dictionary, and it can contain the following keys:
+            * ``schedule``: timestamp at which the restart should occur;
+            * ``role``: restart only nodes which role is ``role``. Can be either:
+                * ``primary`` (or ``master``); or
+                * ``replica``.
+            * ``postgres_version``: restart only nodes which PostgreSQL version is less than ``postgres_version``, e.g.
+                ``15.2``;
+            * ``timeout``: if restart takes longer than ``timeout`` return an error and fail over to a replica;
+            * ``restart_pending``: if we should restart only nodes that have ``pending restart`` flag;
```
CyberDem0n, line 751:
```suggestion
            * ``restart_pending``: if we should restart only when have ``pending restart`` flag;
```

patroni/api.py
```diff
@@ -581,8 +960,29 @@ def is_failover_possible(self, cluster: Cluster, leader: str, candidate: str, ac
...
-    def do_POST_failover(self, action: Optional[str] = 'failover') -> None:
-        """Handles manual failovers/switchovers, mainly from "patronictl"."""
+    def do_POST_failover(self, action: str = 'failover') -> None:
+        """Handle a ``POST`` request to ``/failover`` path.
+
+        Handles manual failovers/switchovers, mainly from ``patronictl``.
+
+        The request body should be a JSON dictionary, and it can contain the following keys:
+            * ``leader``: name of the current leader in the cluster;
+            * ``candidate`` or ``member``: name of the Patroni node to be promoted;
```
CyberDem0n, line 970: The `member` is kept only for backward compatibility.
```suggestion
            * ``candidate``: name of the Patroni node to be promoted;
```

patroni/api.py
```diff
@@ -677,9 +1106,39 @@ def query(self, sql, *params, **kwargs):
...
+            * ``sync_standby``: ``True`` if replication mode is synchronous and this is a sync standby;
+            * ``timeline``: PostgreSQL primary node timeline;
+            * ``replication``: :class:`list` of :class:`dict` entries, one for each replication connection. Each entry
+                contains the following keys:
+                * ``application_name``: ``pg_stat_get_activity().application_name``;
+                * ``client_addr``: ``pg_stat_get_activity().client_addr``;
+                * ``state``: ``pg_stat_get_wal_senders().state``;
+                * ``sync_priority``: ``pg_stat_get_wal_senders().sync_priority``;
+                * ``sync_state``: ``pg_stat_get_wal_senders().sync_state``;
+                * ``usename``: name of the role ``pg_stat_get_activity().usesysid``.
```
CyberDem0n, line 1137: I think here we could safely refer to columns in pg_stat_replication view here.
The view isn't used directly only for "performance" reasons to avoid unnecessary joins.
Under the hood the view is calling exactly these functions.

patroni/api.py
```diff
@@ -778,31 +1275,69 @@ def query(self, sql, *params):
...
+        """
         if self.__auth_key:
             if auth_header is None:
                 return 'no auth header received'
             if not auth_header.startswith('Basic ') or not self.check_basic_auth_key(auth_header[6:]):
                 return 'not authenticated'
 
     @staticmethod
-    def __resolve_ips(host, port):
+    def __resolve_ips(host: Optional[str], port: int) -> Iterator[Union[IPv4Network, IPv6Network]]:
```
CyberDem0n, line 1318:
```suggestion
    def __resolve_ips(host: str, port: int) -> Iterator[Union[IPv4Network, IPv6Network]]:
```

patroni/api.py
```diff
@@ -817,13 +1352,29 @@ def __members_ips(self):
...
+        """
         if self.__allowlist or self.__allowlist_include_members:
             incoming_ip = ip_address(rh.client_address[0])
             if not any(incoming_ip in net for net in self.__allowlist + tuple(self.__members_ips())):
                 return rh._write_response(403, 'Access is denied')
 
         if not hasattr(rh.request, 'getpeercert') or not rh.request.getpeercert():  # valid client cert isn't present
+            # pyright -- ``__ssl_options`` is initially created as ``None``, but right after that it is replaced with a
+            # dictionary through :func:`reload_config`.
+            assert isinstance(self.__ssl_options, dict)
```
CyberDem0n, line 1377: If we declare self.__ssl_options as `self.__ssl_options: Dict[str, Any] = {}` there will be no need to do asserts.

patroni/api.py
```diff
@@ -848,7 +1403,15 @@ def __has_dual_stack():
                     sock.close()
         return False
 
-    def __httpserver_init(self, host, port):
+    def __httpserver_init(self, host: Optional[str], port: int) -> None:
```
CyberDem0n, line 1406: The host can't be None here, but pyright will probably complain because later it could be assigned to None. It could be solved by introducing a new variable.
```suggestion
    def __httpserver_init(self, host: str, port: int) -> None:
```

patroni/api.py
```diff
@@ -908,38 +1493,82 @@ def __initialize(self, listen, ssl_options):
...
+                # There's already been thoughts about using ``cryptography`` module to get the serial number. However,
+                # given the overhead we decided to stick to ``_ssl``. Refer to the following discussion for more
+                # details: https://github.com/zalando/patroni/pull/1887#discussion_r597204638
+                crt = ssl._ssl._test_decode_cert(self.__ssl_options['certfile'])  # pyright: ignore
                 return crt.get('serialNumber')
             except ssl.SSLError as e:
                 logger.error('Failed to get serial number from certificate %s: %r', self.__ssl_options['certfile'], e)
 
-    def reload_local_certificate(self):
+    def reload_local_certificate(self) -> Union[bool, None]:
```
CyberDem0n, line 1551:
```suggestion
    def reload_local_certificate(self) -> Optional[bool]:
```

patroni/api.py
```diff
@@ -908,38 +1493,82 @@ def __initialize(self, listen, ssl_options):
...
-                request.unwrap()
+                # pyright
+                if isinstance(request, SSLSocket):
+                    request.unwrap()
             except Exception as e:
                 logger.debug('Failed to shutdown SSL connection: %r', e)
         super(RestApiServer, self).shutdown_request(request)
 
-    def get_certificate_serial_number(self):
+    def get_certificate_serial_number(self) -> Union[str, None]:
```
CyberDem0n, line 1530:
```suggestion
    def get_certificate_serial_number(self) -> Optional[str]:
```

patroni/api.py
```diff
@@ -908,38 +1493,82 @@ def __initialize(self, listen, ssl_options):
         if reloading_config:
             self.start()
 
-    def process_request_thread(self, request, client_address):
+    def process_request_thread(self, request: Union[socket.socket, SSLSocket], client_address: Tuple[str, int]) -> None:
+        """Process a request to the REST API.
+
+        Wrapper for :func:`ThreadingMixIn.process_request_thread` that additionally:
+            * Enable TCP keepalivel
```
CyberDem0n, line 1500:
```suggestion
            * Enable TCP keepalive
```

patroni/api.py
```diff
@@ -817,13 +1352,29 @@ def __members_ips(self):
                         except Exception as e:
                             logger.debug('Failed to parse url %s: %r', member.api_url, e)
 
-    def check_access(self, rh):
+    def check_access(self, rh: RestApiHandler) -> Union[bool, None]:
```
CyberDem0n, line 1355:
```suggestion
    def check_access(self, rh: RestApiHandler) -> Optional[bool]:
```

patroni/api.py
```diff
@@ -778,31 +1275,69 @@ def query(self, sql, *params):
...
+        :param key: the password received through the Basic authorization header of an HTTP request.
+
+        :returns: ``True`` if *key* matches the password configured for the REST API.
+        """
+        # pyright -- ``__auth_key`` was already checked through the caller method (:func:`check_auth_header`).
+        assert self.__auth_key is not None
         return hmac.compare_digest(self.__auth_key, key.encode('utf-8'))
 
-    def check_auth_header(self, auth_header):
+    def check_auth_header(self, auth_header: Union[str, None]) -> Union[str, None]:
```
CyberDem0n, line 1304:
```suggestion
    def check_auth_header(self, auth_header: Optional[str]) -> Optional[str]:
```

patroni/api.py
```diff
@@ -677,9 +1106,39 @@ def query(self, sql, *params, **kwargs):
     def get_postgresql_status(self, retry: Optional[bool] = False) -> Dict[str, Any]:
```
CyberDem0n, line 1106:
```suggestion
    def get_postgresql_status(self, retry: bool = False) -> Dict[str, Any]:
```

patroni/api.py
```diff
@@ -553,10 +926,16 @@ def poll_failover_result(self, leader, candidate, action):
                 logger.debug('Exception occurred during polling %s result: %s', action, e)
         return 503, action.title() + ' status unknown'
 
-    def is_failover_possible(self, cluster: Cluster, leader: str, candidate: str, action: str) -> Union[str, None]:
-        """Checks whether there are nodes that could take it over after demoting the primary.
+    def is_failover_possible(self, cluster: Cluster, leader: Optional[str], candidate: Optional[str],
+                             action: str) -> Union[str, None]:
```
CyberDem0n, line 930:
```suggestion
                             action: str) -> Optional[str]:
```

patroni/api.py
matthbakeredb: I wonder if this is true of all these. Having the `assert` statements like this is strange to me, afaik `assert` should only be used in tests.

patroni/api.py
barthisrael: Makes sense!

patroni/api.py
```diff
@@ -185,14 +321,37 @@ def do_GET(self, write_status_code_only: Optional[bool] = False) -> None:
...
         self.do_GET(write_status_code_only=True)
 
-    def do_HEAD(self):
+    def do_HEAD(self) -> None:
+        """Handle a ``HEAD`` request.
+
+        Write a simple HTTP response that represents the current PostgreSQL status.
+
+        .. note::
+            Similar to :func:`do_OPTIONS`.
```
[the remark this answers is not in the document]
CyberDem0n: Yeah :)
I would also suggest to mention that these methods will send only `200 OK`/`503 Service Unavailable` as a response and nothing more, particularly no headers.

CyberDem0n:
> But unit tests started to fail.

That's because we actually didn't pass a socket object there and tests should be adjusted. See below:

```diff
diff --git a/patroni/api.py b/patroni/api.py
index 5c76169..33ac834 100644
--- a/patroni/api.py
+++ b/patroni/api.py
@@ -1492,7 +1492,8 @@ class RestApiServer(ThreadingMixIn, HTTPServer, Thread):
         if reloading_config:
             self.start()
 
-    def process_request_thread(self, request: socket.socket, client_address: Tuple[str, int]) -> None:
+    def process_request_thread(self, request: Union[socket.socket, Tuple[bytes, socket.socket]],
+                               client_address: Tuple[str, int]) -> None:
         """Process a request to the REST API.
 
         Wrapper for :func:`ThreadingMixIn.process_request_thread` that additionally:
@@ -1502,14 +1503,15 @@ class RestApiServer(ThreadingMixIn, HTTPServer, Thread):
         :param request: socket to handle the client request.
         :param client_address: tuple containing the client IP and port.
         """
-        enable_keepalive(request, 10, 3)
+        if isinstance(request, socket.socket):
+            enable_keepalive(request, 10, 3)
         if hasattr(request, 'context'):  # SSLSocket
             from ssl import SSLSocket
             if isinstance(request, SSLSocket):  # pyright
                 request.do_handshake()
         super(RestApiServer, self).process_request_thread(request, client_address)
 
-    def shutdown_request(self, request: socket.socket) -> None:
+    def shutdown_request(self, request: Union[socket.socket, Tuple[bytes, socket.socket]]) -> None:
         """Shut down a request to the REST API.
 
         Wrapper for :func:`HTTPServer.shutdown_request` that additionally:
diff --git a/tests/test_api.py b/tests/test_api.py
index 85c01b4..eca139e 100644
--- a/tests/test_api.py
+++ b/tests/test_api.py
@@ -621,22 +621,33 @@ class TestRestApiServer(unittest.TestCase):
         try:
             raise Exception()
         except Exception:
-            self.assertIsNone(MockRestApiServer.handle_error(None, ('127.0.0.1', 55555)))
+            self.assertIsNone(self.srv.handle_error(None, ('127.0.0.1', 55555)))
 
     @patch.object(HTTPServer, '__init__', Mock(side_effect=socket.error))
     def test_socket_error(self):
         self.assertRaises(socket.error, MockRestApiServer, Mock(), '', {'listen': '*:8008'})
 
+    def __create_socket(self):
+        sock = socket.socket()
+        try:
+            import ssl
+            ctx = ssl.create_default_context(ssl.Purpose.SERVER_AUTH)
+            ctx.check_hostname = False
+            sock = ctx.wrap_socket(sock=sock)
+            sock.do_handshake = Mock()
+            sock.unwrap = Mock(side_effect=Exception)
+        except Exception:
+            pass
+        return sock
+
     @patch.object(ThreadingMixIn, 'process_request_thread', Mock())
     def test_process_request_thread(self):
-        self.srv.process_request_thread(Mock(), '2')
+        self.srv.process_request_thread(self.__create_socket(), ('2', 54321))
 
     @patch.object(MockRestApiServer, 'process_request', Mock(side_effect=RuntimeError))
     @patch.object(MockRestApiServer, 'get_request')
     def test_process_request_error(self, mock_get_request):
-        mock_request = Mock()
-        mock_request.unwrap.side_effect = Exception
-        mock_get_request.return_value = (mock_request, ('127.0.0.1', 55555))
+        mock_get_request.return_value = (self.__create_socket(), ('127.0.0.1', 55555))
         self.srv._handle_request_noblock()
 
     @patch('ssl._ssl._test_decode_cert', Mock())
```

patroni/api.py
```diff
@@ -743,30 +1203,68 @@ def get_postgresql_status(self, retry: Optional[bool] = False) -> Dict[str, Any]
...
+        """Parse and dispatch a request to the appropriate ``do_*`` method.
+
+        .. note::
+            This is only used to keep track of latency when logging messages through :func:`log_message`.
+        """
         self.__start_time = time.time()
         BaseHTTPRequestHandler.handle_one_request(self)
 
-    def log_message(self, fmt, *args):
+    def log_message(self, format: str, *args) -> None:
```
CyberDem0n, line 1215:
```suggestion
    def log_message(self, format: str, *args: Any) -> None:
```

patroni/api.py
CyberDem0n: Actually, it could be solved differently:
```python
ctx = ssl.SSLContext(ssl.PROTOCOL_TLS_CLIENT)
crts = ctx.load_verify_locations(self.__ssl_options['certfile'])
if crts:
    return crts[0].get('serialNumber')

```

CyberDem0n (commented):
Following patch resolves pyright issues (including usage of private ssl._ssl._test_decode_cert() function). Unfortunately I can't make post is as comments/suggestions.

```patch
diff --git a/patroni/api.py b/patroni/api.py
index 33ac834..5ba3327 100644
--- a/patroni/api.py
+++ b/patroni/api.py
@@ -183,7 +183,7 @@ class RestApiHandler(BaseHTTPRequestHandler):
         if patroni.postgresql.pending_restart:
             response['pending_restart'] = True
         response['patroni'] = {'version': patroni.version, 'scope': patroni.postgresql.scope}
-        if patroni.scheduled_restart and isinstance(patroni.scheduled_restart, dict):
+        if patroni.scheduled_restart:
             response['scheduled_restart'] = patroni.scheduled_restart.copy()
             del response['scheduled_restart']['postmaster_start_time']
             response['scheduled_restart']['schedule'] = (response['scheduled_restart']['schedule']).isoformat()
@@ -467,7 +467,7 @@ class RestApiHandler(BaseHTTPRequestHandler):
         patroni = self.server.patroni
         epoch = datetime.datetime(1970, 1, 1, tzinfo=tzutc)
 
-        metrics = []
+        metrics: List[str] = []
 
         scope_label = '{{scope="{0}"}}'.format(patroni.postgresql.scope)
         metrics.append("# HELP patroni_version Patroni semver without periods.")
@@ -638,7 +638,7 @@ class RestApiHandler(BaseHTTPRequestHandler):
         if request:
             cluster = self.server.patroni.dcs.get_cluster()
-            if not deep_compare(request, cluster.config.data):
+            if not (cluster.config and deep_compare(request, cluster.config.data)):
                 value = json.dumps(request, separators=(',', ':'))
                 if not self.server.patroni.dcs.set_config_value(value):
                     return self.send_error(502)
@@ -915,7 +915,7 @@ class RestApiHandler(BaseHTTPRequestHandler):
             time.sleep(1)
             try:
                 cluster = self.server.patroni.dcs.get_cluster()
-                if not cluster.is_unlocked() and cluster.leader.name != leader:
+                if not cluster.is_unlocked() and cluster.leader and cluster.leader.name != leader:
                     if not candidate or candidate == cluster.leader.name:
                         return 200, 'Successfully {0}ed over to "{1}"'.format(action[:-4], cluster.leader.name)
                     else:
@@ -952,7 +952,7 @@ class RestApiHandler(BaseHTTPRequestHandler):
             if not members:
                 return action + ' is not possible: can not find sync_standby'
         else:
-            members = [m for m in cluster.members if m.name != cluster.leader.name and m.api_url]
+            members = [m for m in cluster.members if not cluster.leader or m.name != cluster.leader.name and m.api_url]
             if not members:
                 return action + ' is not possible: cluster does not have members except leader'
         for st in self.server.patroni.ha.fetch_nodes_statuses(members):
@@ -1181,7 +1181,8 @@ class RestApiHandler(BaseHTTPRequestHandler):
             if row[1] > 0:
                 result['timeline'] = row[1]
             else:
-                leader_timeline = None if not cluster or cluster.is_unlocked() else cluster.leader.timeline
+                leader_timeline = None\
+                    if not cluster or cluster.is_unlocked() or not cluster.leader else cluster.leader.timeline
                 result['timeline'] = postgresql.replica_cached_timeline(leader_timeline)
 
             if row[7]:
@@ -1192,7 +1193,7 @@ class RestApiHandler(BaseHTTPRequestHandler):
             if state == 'running':
                 logger.exception('get_postgresql_status')
                 state = 'unknown'
-            result = {'state': state, 'role': postgresql.role}
+            result: Dict[str, Any] = {'state': state, 'role': postgresql.role}
 
         if global_config.is_paused:
             result['pause'] = True
@@ -1268,7 +1269,7 @@ class RestApiServer(ThreadingMixIn, HTTPServer, Thread):
         cursor = None
         try:
             with self.patroni.postgresql.connection().cursor() as cursor:
-                cursor.execute(sql, params)
+                cursor.execute(sql.encode('utf-8'), params)
                 return [r for r in cursor]
         except psycopg.Error as e:
             if cursor and cursor.connection.closed == 0:
@@ -1346,10 +1347,10 @@ class RestApiServer(ThreadingMixIn, HTTPServer, Thread):
                     if member.api_url:
                         try:
                             r = urlparse(member.api_url)
-                            host = r.hostname
-                            port = r.port or (443 if r.scheme == 'https' else 80)
-                            for ip in self.__resolve_ips(host, port):
-                                yield ip
+                            if r.hostname:
+                                port = r.port or (443 if r.scheme == 'https' else 80)
+                                for ip in self.__resolve_ips(r.hostname, port):
+                                    yield ip
                         except Exception as e:
                             logger.debug('Failed to parse url %s: %r', member.api_url, e)
 
@@ -1536,14 +1537,11 @@ class RestApiServer(ThreadingMixIn, HTTPServer, Thread):
             import ssl
             try:
-                # We silent pyright for this line otherwise it would complain about accessing a protected member from
-                # ``ssl``.
-                # There's already been thoughts about using ``cryptography`` module to get the serial number. However,
-                # given the overhead we decided to stick to ``_ssl``. Refer to the following discussion for more
-                # details: https://github.com/zalando/patroni/pull/1887#discussion_r597204638
-                crt = ssl._ssl._test_decode_cert(self.__ssl_options['certfile'])  # pyright: ignore
-                return crt.get('serialNumber')
-            except ssl.SSLError as e:
+                ctx = ssl.SSLContext(ssl.PROTOCOL_TLS_CLIENT)
+                crts = ctx.load_verify_locations(self.__ssl_options['certfile'])
+                if crts:
+                    return crts[0].get('serialNumber')
+            except Exception as e:
                 logger.error('Failed to get serial number from certificate %s: %r', self.__ssl_options['certfile'], e)
 
     def reload_local_certificate(self) -> Optional[bool]:
diff --git a/tests/test_api.py b/tests/test_api.py
index eca139e..166d3eb 100644
--- a/tests/test_api.py
+++ b/tests/test_api.py
@@ -180,6 +180,7 @@ class MockRestApiServer(RestApiServer):
 
 @patch('ssl.SSLContext.load_cert_chain', Mock())
 @patch('ssl.SSLContext.wrap_socket', Mock(return_value=0))
+@patch('ssl.SSLContext.load_verify_locations', Mock(return_value=[Mock()]))
 @patch.object(HTTPServer, '__init__', Mock())
 class TestRestApiHandler(unittest.TestCase):
 
@@ -588,6 +589,7 @@ class TestRestApiServer(unittest.TestCase):
     @patch('ssl.SSLContext.load_cert_chain', Mock())
     @patch('ssl.SSLContext.set_ciphers', Mock())
     @patch('ssl.SSLContext.wrap_socket', Mock(return_value=0))
+    @patch('ssl.SSLContext.load_verify_locations', Mock(return_value=[Mock()]))
     @patch.object(HTTPServer, '__init__', Mock())
     def setUp(self):
         self.srv = MockRestApiServer(Mock(), '', {'listen': '*:8008', 'certfile': 'a', 'verify_client': 'required',
@@ -650,6 +652,10 @@ class TestRestApiServer(unittest.TestCase):
         mock_get_request.return_value = (self.__create_socket(), ('127.0.0.1', 55555))
         self.srv._handle_request_noblock()
 
-    @patch('ssl._ssl._test_decode_cert', Mock())
+    @patch('ssl.SSLContext.load_verify_locations', Mock(return_value=[Mock()]))
     def test_reload_local_certificate(self):
         self.assertTrue(self.srv.reload_local_certificate())
+
+    @patch('ssl.SSLContext.load_verify_locations', Mock(side_effect=Exception))
+    def test_get_certificate_serial_number(self):
+        self.assertIsNone(self.srv.get_certificate_serial_number())
```

barthisrael:
> ```diff
> @@ -1268,7 +1269,7 @@ class RestApiServer(ThreadingMixIn, HTTPServer, Thread):
>          cursor = None
>          try:
>              with self.patroni.postgresql.connection().cursor() as cursor:
> -                cursor.execute(sql, params)
> +                cursor.execute(sql.encode('utf-8'), params)
>                  return [r for r in cursor]
>          except psycopg.Error as e:
>              if cursor and cursor.connection.closed == 0:
> ```

Should that really be `bytes` instead of `str`? I'd expect that to be a plain `str` instance.

That change introduces issues to `test_api.py`. If we modify the tests to use bytes it will make `test_api.py` happy, but everything else unhappy 😆 .

barthisrael:
`pyright` keeps silent after performing these changes here:

```diff
diff --git a/patroni/api.py b/patroni/api.py
index 0fb3915..5315eb1 100644
--- a/patroni/api.py
+++ b/patroni/api.py
@@ -1269,7 +1269,7 @@ class RestApiServer(ThreadingMixIn, HTTPServer, Thread):
         cursor = None
         try:
             with self.patroni.postgresql.connection().cursor() as cursor:
-                cursor.execute(sql.encode('utf-8'), params)
+                cursor.execute(sql, params)
                 return [r for r in cursor]
         except psycopg.Error as e:
             if cursor and cursor.connection.closed == 0:
```

CyberDem0n:
@barthisrael there are conflicts after my PR with type hits was merged.

barthisrael:
> @barthisrael there are conflicts after my PR with type hits was merged.

Thanks for the heads up! I just resolved the conflicts.

We noticed it started popping up a `pyright` issue:

```python
patroni/api.py:1272:32 - error: Argument of type "str" cannot be assigned to parameter "query" of type "Query" in function "execute"
    Type "str" cannot be assigned to type "Query"
      "str" is incompatible with "LiteralString"
      "str" is incompatible with "bytes"
      "str" is incompatible with "SQL"
      "str" is incompatible with "Composed" (reportGeneralTypeIssues)
```

Interestingly enough it refers to the same thing that you seem to have tried addressing through your `pyright` PR -- related to [this comment](https://github.com/zalando/patroni/pull/2648#issuecomment-1536165184).

From what we could see the `str` goes fine with `psycopg2`, but `psycopg` doesn't like it, apparently because it wants to avoid dynamic SQL that could be malicious:

```python
Query: TypeAlias = Union[LiteralString, bytes, "sql.SQL", "sql.Composed"]
```

We were wondering how to deal with that -- should we add a `# pyright: ignore` for now? Some attempts that we have made so far seem to start a chain of changes to make everything compatible with let's say a `LiteralString`.

matthbakeredb:
> > @barthisrael there are conflicts after my PR with type hits was merged.
> 
> Thanks for the heads up! I just resolved the conflicts.
> 
> We noticed it started popping up a `pyright` issue:
> 
> ```python
> patroni/api.py:1272:32 - error: Argument of type "str" cannot be assigned to parameter "query" of type "Query" in function "execute"
>     Type "str" cannot be assigned to type "Query"
>       "str" is incompatible with "LiteralString"
>       "str" is incompatible with "bytes"
>       "str" is incompatible with "SQL"
>       "str" is incompatible with "Composed" (reportGeneralTypeIssues)
> ```
> 
> Interestingly enough it refers to the same thing that you seem to have tried addressing through your `pyright` PR -- related to [this comment](https://github.com/zalando/patroni/pull/2648#issuecomment-1536165184).
> 
> From what we could see the `str` goes fine with `psycopg2`, but `psycopg` doesn't like it, apparently because it wants to avoid dynamic SQL that could be malicious:
> 
> ```python
> Query: TypeAlias = Union[LiteralString, bytes, "sql.SQL", "sql.Composed"]
> ```
> 
> We were wondering how to deal with that -- should we add a `# pyright: ignore` for now? Some attempts that we have made so far seem to start a chain of changes to make everything compatible with let's say a `LiteralString`.

I've been looking at this and `cusor.execute` seems happier if the queries are passed like this

```

POSTMASTER_START_TIME = "pg_catalog.pg_postmaster_start_time()"
TL_LSN = (
    "CASE WHEN pg_catalog.pg_is_in_recovery() THEN 0 ELSE ('x' || pg_catalog.substr(pg_catalog.pg_{wal_name}file_name("
    "pg_catalog.pg_current_{wal_name}_{lsn_name}()), 1, 8))::bit(32)::int END, "  # primary timeline
    "CASE WHEN pg_catalog.pg_is_in_recovery() THEN 0 "
    "ELSE pg_catalog.pg_{wal_name}_{lsn_name}_diff(pg_catalog.pg_current_{wal_name}_{lsn_name}(), '0/0')::bigint END, "  # write_lsn
    "pg_catalog.pg_{wal_name}_{lsn_name}_diff(pg_catalog.pg_last_{wal_name}_replay_{lsn_name}(), '0/0')::bigint, "
    "pg_catalog.pg_{wal_name}_{lsn_name}_diff(COALESCE(pg_catalog.pg_last_{wal_name}_receive_{lsn_name}(), '0/0'), '0/0')::bigint, "
    "pg_catalog.pg_is_in_recovery() AND pg_catalog.pg_is_{wal_name}_replay_paused()")

stmt = (
    "SELECT "
    + POSTMASTER_START_TIME
    + ", "
    + TL_LSN
    + ","
      " pg_catalog.pg_last_xact_replay_timestamp(),"
      " pg_catalog.array_to_json(pg_catalog.array_agg(pg_catalog.row_to_json(ri))) "
      "FROM (SELECT (SELECT rolname FROM pg_catalog.pg_authid WHERE oid = usesysid) AS usename,"
      " FROM pg_catalog.pg_stat_get_wal_senders() w, pg_catalog.pg_stat_get_activity(pid)) AS ri")

wal_name = psycopg.sql.Identifier('wal')
lsn_name = psycopg.sql.Identifier('lsn')

query = psycopg.sql.SQL(stmt).format(wal_name=wal_name, lsn_name=lsn_name)
```

I made some changes to api.py in this direction but the query method doesn't like the fact the types of the query arg differ between psycopg2 and 3.
```
Union[str, bytes, 'psycopg.sql.Composable']
```
vs
```
Union[LiteralString, bytes, "sql.SQL", "sql.Composed"]
```

So I thought I'd define those types depending on which is loaded in patroni.psycopg but pyright returns a bunch of `partially unknown` errors. I am wondering whether I might need to create some type stubs for this?

CyberDem0n:
> Should that really be bytes instead of str? I'd expect that to be a plain str instance.

Surprisingly (also for me :)), the interface accepts bytes, both psycopg2 and psycopg3.
Unit tests for you are were failing because https://github.com/zalando/patroni/pull/2652/files#diff-001e61d97d9dee27d0c5aa1f23ba6c1972cd8a2e8320bac839656b0ab935b84dR85-R86 wasn't merged yet.
Now everything works with like a charm with `str.encode()`.

patroni/api.py
```diff
@@ -213,34 +384,90 @@ def do_GET_readiness(self):
...
+        * ``patroni_standby_leader``: ``1`` if standby leader node, else ``0``;
+        * ``patroni_replica``: ``1`` if a replica, else ``0``;
+        * ``patroni_sync_standby``: ``1`` if a sync replica, else ``0``;
+        * ``patroni_xlog_received_location``: ``pg_wal_lsn_diff(pg_last_wal_receive_lsn(), '0/0')``;
+        * ``patroni_xlog_replayed_location``: ``pg_wal_lsn_diff(pg_last_wal_replay_lsn(), '0/0)``;
+        * ``patroni_xlog_replayed_timestamp``: ``pg_last_xact_replay_timestamp``;
+        * ``patroni_xlog_paused``: ``pg_is_wal_replay_paused()``;
+        * ``patroni_postgres_server_version``: Postgres version without periods, e.g. ``150002`` for Postgres ``15.2``;
+        * ``patroni_cluster_unlocked``: ``1`` if no one holds the leader lock, else ``0``;
+        * ``patroni_failsafe_mode_is_active``: ``1`` if ``failmode`` is currently active, else ``0``;
```
hughcapet, line 460:
```suggestion
        * ``patroni_failsafe_mode_is_active``: ``1`` if ``failsafe_mode`` is currently active, else ``0``;
```

CyberDem0n:
:+1:

PR #2652, opened 2023-04-24
Enable pyright strict mode

CyberDem0n:
- added pyrightconfig.json with typeCheckingMode=strict
- added type hints to all files except api.py
- fixed most of pyright issues in files with added docstrings
- added type stubs for dns, etcd, consul, kazoo, pysyncobj and so on python modules
- added type stubs for psycopg2 and urllib3 with some little fixes
- fixes most of the issues reported by pyright


CyberDem0n:
@barthisrael, @matthbakeredb, please review.
It is necessary because you started changing the same files in the #2648.

matthbakeredb:
I don't see a `pyrightconfig.json`, did you forget to add it?

CyberDem0n:
> I don't see a `pyrightconfig.json`, did you forget to add it?

Pushed.

patroni/config.py
```diff
@@ -309,6 +313,7 @@ def set_dynamic_configuration(self, configuration):
                 return True
             except Exception:
                 logger.exception('Exception when setting dynamic_configuration')
+        return False
 
     def reload_local_configuration(self):
```
barthisrael, line 318:
```suggestion
    def reload_local_configuration(self) -> Union[bool, None]:
```

patroni/dcs/__init__.py
```diff
@@ -161,13 +171,13 @@ def conn_url(self):
             self.data['conn_url'] = conn_url
             return conn_url
 
-    def conn_kwargs(self, auth=None):
+    def conn_kwargs(self, auth: Union[Any, Dict[str, Any], None] = None) -> Dict[str, Any]:
```
barthisrael, line 174:
```suggestion
    def conn_kwargs(self, auth: Union[Dict[str, Any], None] = None) -> Dict[str, Any]:
```

It apparently handles only `dict` and `None`.

patroni/dcs/__init__.py
```diff
@@ -230,8 +240,8 @@ def version(self):
 
 class RemoteMember(Member):
     """Represents a remote member (typically a primary) for a standby cluster"""
-    def __new__(cls, name, data):
-        return super(RemoteMember, cls).__new__(cls, None, name, None, data)
+    def __new__(cls, name: str, data: Dict[str, Any]) -> 'RemoteMember':
+        return super(RemoteMember, cls).__new__(cls, -1, name, None, data)
 
     @staticmethod
     def allowed_keys():
```
barthisrael, line 247:
```suggestion
    def allowed_keys() -> Tuple[str, ...]:
```

patroni/dcs/__init__.py
```diff
@@ -447,37 +470,40 @@ def matches(self, name: Union[str, None], check_leader: Optional[bool] = False)
...
-class TimelineHistory(namedtuple('TimelineHistory', 'index,value,lines')):
+class TimelineHistory(NamedTuple):
     """Object representing timeline history file"""
+    index: _Version
+    value: Any
+    lines: List[_HistoryTuple]
 
     @staticmethod
-    def from_node(index, value):
+    def from_node(index: _Version, value: str) -> 'TimelineHistory':
```
barthisrael, line 492:
```suggestion
    def from_node(index: _Version, value: Optional[str]) -> 'TimelineHistory':
```

As the `except` blocks handles `TypeError`, I'd say `value` can be `None`.

patroni/dcs/__init__.py
```diff
@@ -491,55 +517,62 @@ class Cluster(namedtuple('Cluster', 'initialize,config,leader,last_lsn,members,'
...
         return any(m for m in self.members if m.name == member_name)
 
-    def get_member(self, member_name, fallback_to_leader=True):
+    def get_member(self, member_name: str, fallback_to_leader: bool = True) -> Union[Member, Leader, None]:
         return ([m for m in self.members if m.name == member_name] or [self.leader if fallback_to_leader else None])[0]
 
-    def get_clone_member(self, exclude):
-        exclude = [exclude] + [self.leader.name] if self.leader else []
+    def get_clone_member(self, exclude_name: str) -> Union[Member, Leader, None]:
+        exclude = [exclude_name] + [self.leader.name] if self.leader else []
```
barthisrael, line 552:
```suggestion
        exclude = [exclude_name] + ([self.leader.name] if self.leader else [])
```

It has nothing to do with this actual PR, but I was wondering if this check is missing parenthesis?

patroni/dcs/__init__.py
```diff
@@ -954,10 +998,11 @@ def attempt_to_acquire_leader(self):
         process requests (hopefuly temporary), the ~DCSError exception should be raised"""
 
     @abc.abstractmethod
-    def set_failover_value(self, value, index=None):
+    def set_failover_value(self, value: str, index: Optional[Any] = None) -> bool:
         """Create or update `/failover` key"""
 
-    def manual_failover(self, leader, candidate, scheduled_at=None, index=None):
+    def manual_failover(self, leader: str, candidate: str, scheduled_at: Optional[datetime.datetime] = None,
```
barthisrael, line 1004:
```suggestion
    def manual_failover(self, leader: Optional[str], candidate: Optional[str], scheduled_at: Optional[datetime.datetime] = None,
```

The `if leader` and `if candidate` that are performed in the function body give me the impression the parameters can assume `None` value.

patroni/dcs/__init__.py
```diff
@@ -1033,8 +1078,8 @@ def sync_state(leader: Union[str, None], sync_standby: Union[Collection[str], No
         """
         return {'leader': leader, 'sync_standby': ','.join(sorted(sync_standby)) if sync_standby else None}
 
-    def write_sync_state(self, leader: Union[str, None], sync_standby: Union[Collection[str], None],
-                         index: Optional[Union[int, str]] = None) -> bool:
+    def write_sync_state(self, leader: Optional[str], sync_standby: Optional[Collection[str]],
+                         index: Optional[_Version] = None) -> bool:
```
barthisrael, line 1082:
```suggestion
                         index: Optional[Any] = None) -> bool:
```

All other occurrences of `index` parameter in methods of `AbstractDcs` class use `Optional[Any]`.
That leads me to believe this one should also follow the same convention.

Besides that, while taking a look at the `index` usage, it seems to expect an `int` which is incremented upon modifications. What is the usage of `str` for `index`?

patroni/dcs/consul.py
```diff
@@ -38,12 +39,17 @@ class InvalidSession(ConsulException):
     """invalid session"""
 
 
-Response = namedtuple('Response', 'code,headers,body,content')
+class Response(NamedTuple):
+    code: int
+    headers: Any
```
barthisrael, line 44:
```suggestion
    headers: Union[Mapping[str, str], Mapping[bytes, bytes], None]
```

Based on [urllib3 source code](https://github.com/urllib3/urllib3/blob/main/src/urllib3/response.py#L533).

patroni/dcs/consul.py
```diff
@@ -88,14 +94,18 @@ def response(response):
...
-    def __getattr__(self, method):
+    def __getattr__(self, method: str) -> Callable[[Callable[[Response], Union[bool, Any, Tuple[str, Any]]],
+                                                    str, Union[None, Dict[str, Any], List[Tuple[str, Any]]],
+                                                    str, Optional[Dict[str, str]]], Union[bool, Any, Tuple[str, Any]]]:
         if method not in ('get', 'post', 'put', 'delete'):
             raise AttributeError("HTTPClient instance has no attribute '{0}'".format(method))
 
-        def wrapper(callback, path, params=None, data='', headers=None):
+        def wrapper(callback: Callable[[Response], Union[bool, Any, Tuple[str, Any]]], path: str,
+                    params: Union[None, Dict[str, Any], List[Tuple[str, Any]]] = None, data: str = '',
```
barthisrael, line 107:
```suggestion
                    params: Union[None, Dict[str, Any], List[Tuple[str, Any]], Tuple[Tuple[str, Any], ...]] = None, data: str = '',
```

Analogous to previous one.

patroni/dcs/consul.py
```diff
@@ -88,14 +94,18 @@ def response(response):
                 raise ConsulInternalError(msg)
         return Response(response.status, response.headers, body, content)
 
-    def uri(self, path, params=None):
+    def uri(self, path: str, params: Optional[Dict[str, Any]] = None) -> str:
         return '{0}{1}{2}'.format(self.base_uri, path, params and '?' + urlencode(params) or '')
 
-    def __getattr__(self, method):
+    def __getattr__(self, method: str) -> Callable[[Callable[[Response], Union[bool, Any, Tuple[str, Any]]],
+                                                    str, Union[None, Dict[str, Any], List[Tuple[str, Any]]],
```
barthisrael, line 101:
```suggestion
                                                    str, Union[None, Dict[str, Any], List[Tuple[str, Any]], Tuple[Tuple[str, Any], ...]],
```

Analogous to previous one.

patroni/dcs/consul.py
```diff
@@ -127,13 +137,13 @@ def wrapper(callback, path, params=None, data='', headers=None):
 
 class ConsulClient(base.Consul):
 
-    def __init__(self, *args, **kwargs):
+    def __init__(self, *args: Any, **kwargs: Any):
```
barthisrael, line 140:
```suggestion
    def __init__(self, *args: Any, **kwargs: Any) -> None:
```

patroni/dcs/consul.py
```diff
@@ -266,20 +278,20 @@ def reload_config(self, config):
         self._previous_loop_register_service = self._register_service
         self._register_service = should_register_service
 
-    def set_ttl(self, ttl):
+    def set_ttl(self, ttl: int) -> Optional[bool]:
```
barthisrael, line 281: This one has `Optional[bool]` to make it compatible with the `set_ttl` function from the abstract class, right? I'm asking because this implementation itself only returns `None`.

patroni/dcs/__init__.py
CyberDem0n: It does, but `postgresql.authentication.{superuser,replication,rewind}` could contain something different from dict and there are no external checks for that.
The one that we have is located down this method: `if auth and isinstance(auth, dict):`.
If we remove Any from Union pyright starts complaining.

patroni/dcs/consul.py
```diff
@@ -8,11 +8,12 @@
 import time
 import urllib3
 
-from collections import defaultdict, namedtuple
+from collections import defaultdict
 from consul import ConsulException, NotFound, base
 from http.client import HTTPException
 from urllib3.exceptions import HTTPError
 from urllib.parse import urlencode, urlparse, quote
+from typing import Any, Callable, Dict, List, NamedTuple, Optional, Union, Tuple
```
CyberDem0n, line 16:
```suggestion
from typing import Any, Callable, Dict, List, Mapping, NamedTuple, Optional, Union, Tuple
```

patroni/dcs/kubernetes.py
```diff
@@ -232,36 +240,37 @@ def __init__(self, bypass_api_service: Optional[bool] = False) -> None:
...
-        def set_read_timeout(self, timeout):
+        def set_read_timeout(self, timeout: Union[int, float]) -> None:
             self._read_timeout = timeout
 
-        def set_api_servers_cache_ttl(self, ttl):
+        def set_api_servers_cache_ttl(self, ttl: int) -> None:
             self._api_servers_cache_ttl = ttl - 0.5
 
-        def set_base_uri(self, value):
+        def set_base_uri(self, value: str):
```
CyberDem0n, line 249:
```suggestion
        def set_base_uri(self, value: str) -> None:
```

patroni/dcs/zookeeper.py
```diff
@@ -136,30 +139,28 @@ def _kazoo_connect(self, *args):
         ret = self._orig_kazoo_connect(*args)
         return max(self.loop_wait - 2, 2) * 1000, ret[1]
 
-    def session_listener(self, state):
+    def session_listener(self, state: str) -> None:
         if state in [KazooState.SUSPENDED, KazooState.LOST]:
             self.cluster_watcher(None)
 
-    def status_watcher(self, event):
+    def status_watcher(self, event: Optional[WatchedEvent]):
```
CyberDem0n, line 146:
```suggestion
    def status_watcher(self, event: Optional[WatchedEvent]) -> None:
```

patroni/utils.py
```diff
@@ -586,7 +590,7 @@ def polling_loop(timeout: Union[int, float], interval: Optional[Union[int, float
     while time.time() < end_time:
         yield iteration
         iteration += 1
-        time.sleep(interval)
+        time.sleep(float(interval))
 
 
 def split_host_port(value: str, default_port: int) -> Tuple[str, int]:
```
barthisrael, line 596:
```suggestion
def split_host_port(value: str, default_port: Optional[int]) -> Tuple[str, int]:
```

Some methods, like `__initialize` from `patroni.api.RestApiServer`, call `split_host_port` with `port=None`.

barthisrael:
I noticed these `flake8` issues in `patroni.config`:

```
patroni/config.py:162:1: D400 First line should end with a period
patroni/config.py:241:1: D205 1 blank line required between summary line and description
patroni/config.py:262:1: D400 First line should end with a period
```

One idea to handle them would be:

```
$ git diff patroni/config.py
diff --git a/patroni/config.py b/patroni/config.py
index 8189748..54b5584 100644
--- a/patroni/config.py
+++ b/patroni/config.py
@@ -159,7 +159,8 @@ def get_global_config(cluster: Union[Cluster, None], default: Optional[Dict[str,


 class Config(object):
-    """
+    """Handle Patroni configuration.
+
     This class is responsible for:

       1) Building and giving access to `effective_configuration` from:
@@ -238,7 +239,8 @@ class Config(object):
         return deepcopy(self._dynamic_configuration)

     def _load_config_path(self, path: str) -> Dict[str, Any]:
-        """
+        """Load configuration found in *path*.
+
         If path is a file, loads the yml file pointed to by path.
         If path is a directory, loads all yml files in that directory in alphabetical order
         """
@@ -259,7 +261,7 @@ class Config(object):
         return overall_config

     def _load_config_file(self) -> Dict[str, Any]:
-        """Loads config.yaml from filesystem and applies some values which were set via ENV"""
+        """Loads config.yaml from filesystem and applies some values which were set via ENV."""
         assert self._config_file is not None
         config = self._load_config_path(self._config_file)
         patch_config(config, self.__environment_configuration)
```

patroni/dcs/etcd.py
```diff
@@ -71,14 +74,14 @@ def resolve(self, host, port):
                 response = new_response
         return response
 
-    def resolve_async(self, host, port, attempt=0):
+    def resolve_async(self, host: str, port: int, attempt: int = 0):
```
barthisrael, line 77:
```suggestion
    def resolve_async(self, host: str, port: int, attempt: int = 0) -> None:
```

patroni/dcs/etcd.py
```diff
@@ -71,14 +74,14 @@ def resolve(self, host, port):
...
         self._resolve_queue.put(((host, port), attempt))
 
-    def remove(self, host, port):
+    def remove(self, host: str, port: int):
```
barthisrael, line 80:
```suggestion
    def remove(self, host: str, port: int) -> None:
```

barthisrael:
Are we willing to address issues like these, or should we instruct `pyright` to ignore them?

```
patroni/dcs/etcd.py
  patroni/dcs/etcd.py:627:18 - error: "_has_failed" is protected and used outside of the class in which it is declared (reportPrivateUsage)
  patroni/dcs/etcd.py:630:18 - error: "_handle_exception" is protected and used outside of the class in which it is declared (reportPrivateUsage)
  patroni/dcs/etcd.py:633:18 - error: "_handle_exception" is protected and used outside of the class in which it is declared (reportPrivateUsage)
  patroni/dcs/etcd.py:633:53 - error: "_client" is protected and used outside of the class in which it is declared (reportPrivateUsage)
4 errors, 0 warnings, 0 informations
```

barthisrael:
As `_handle_auth_errors` is intended to be used a decorator, I'd say to move it out of the class `Etcd3Client`, so we could clear these pyright issues:

```
  patroni/dcs/etcd3.py:298:29 - error: Instance methods should take a "self" parameter (reportSelfClsParameterName)
  patroni/dcs/etcd3.py:298:35 - error: Type of parameter "func" must be a supertype of its class "Etcd3Client" (reportGeneralTypeIssues)
```

patroni/dcs/etcd3.py
```diff
@@ -349,13 +381,15 @@ def watchrange(self, key, range_end=None, start_revision=None, filters=None, rea
...
+                    filters: Optional[List[Dict[str, Any]]] = None,
+                    read_timeout: Optional[float] = None) -> urllib3.response.HTTPResponse:
         return self.watchrange(key, prefix_range_end(key), start_revision, filters, read_timeout)
 
 
 class KVCache(Thread):
 
-    def __init__(self, dcs, client):
+    def __init__(self, dcs: 'Etcd3', client: 'PatroniEtcd3Client'):
         Thread.__init__(self)
```
barthisrael, line 393:
```suggestion
        super(KVCache, self).__init__()
```

barthisrael:
Are we willing to address issues like these, or should we instruct pyright to ignore them?

```
  patroni/dcs/etcd3.py:402:40 - error: "_name" is protected and used outside of the class in which it is declared (reportPrivateUsage)
  patroni/dcs/etcd3.py:583:43 - error: "_retry" is protected and used outside of the class in which it is declared (reportPrivateUsage)
  patroni/dcs/etcd3.py:585:46 - error: "_retry" is protected and used outside of the class in which it is declared (reportPrivateUsage)
```

barthisrael:
Are we willing to address issues like these, or should we instruct pyright to ignore them?

```
  patroni/dcs/kubernetes.py:563:25 - error: "_run_and_handle_exceptions" is protected and used outside of the class in which it is declared (reportPrivateUsage)
  patroni/dcs/kubernetes.py:585:67 - error: "_LEADER" is protected and used outside of the class in which it is declared (reportPrivateUsage)
  patroni/dcs/kubernetes.py:585:109 - error: "_CONFIG" is protected and used outside of the class in which it is declared (reportPrivateUsage)
```

patroni/dcs/kubernetes.py
```diff
@@ -161,30 +166,32 @@ def headers(self):
 
 class K8sObject(object):
 
-    def __init__(self, kwargs):
+    def __init__(self, kwargs: Dict[str, Any]):
```
barthisrael, line 169:
```suggestion
    def __init__(self, kwargs: Dict[str, Any]) -> None:
```

patroni/dcs/kubernetes.py
```diff
@@ -161,30 +166,32 @@ def headers(self):
...
         self._dict = {k: self._wrap(k, v) for k, v in kwargs.items()}
 
-    def get(self, name, default=None):
+    def get(self, name: str, default: Optional[Any] = None) -> Any:
```
barthisrael, line 172:
```suggestion
    def get(self, name: str, default: Optional[Any] = None) -> Optional[Any]:
```

Should we make it explicit that `None` is a valid return value?

patroni/dcs/kubernetes.py
```diff
@@ -161,30 +166,32 @@ def headers(self):
...
         return self._dict.get(name, default)
 
-    def __getattr__(self, name):
+    def __getattr__(self, name: str) -> Any:
```
barthisrael, line 175:
```suggestion
    def __getattr__(self, name: str) -> Optional[Any]:
```

Same thoughts as the previous.

patroni/dcs/kubernetes.py
```diff
@@ -454,14 +471,14 @@ def wrapper(*args, **kwargs):
 
     class _K8sObjectTemplate(K8sObject):
         """The template for objects which we create locally, e.g. k8s_client.V1ObjectMeta & co"""
-        def __init__(self, **kwargs):
+        def __init__(self, **kwargs: Any):
```
barthisrael, line 474:
```suggestion
        def __init__(self, **kwargs: Any) -> None:
```

patroni/dcs/kubernetes.py
```diff
@@ -474,15 +491,15 @@ def __getattr__(self, name):
 
 class KubernetesRetriableException(k8s_client.rest.ApiException):
 
-    def __init__(self, orig):
+    def __init__(self, orig: K8sClient.rest.ApiException):
```
barthisrael, line 494:
```suggestion
    def __init__(self, orig: K8sClient.rest.ApiException) -> None:
```

patroni/dcs/kubernetes.py
```diff
@@ -536,12 +553,12 @@ def wrapper(*args, **kwargs):
...
 
     @property
-    def use_endpoints(self):
+    def use_endpoints(self) -> bool:
         return self._use_endpoints
 
 
-def catch_kubernetes_errors(func):
-    def wrapper(self, *args, **kwargs):
+def catch_kubernetes_errors(func: Callable[..., Any]):
```
barthisrael, line 560:
```suggestion
def catch_kubernetes_errors(func: Callable[..., Any]) -> Callable[..., Any]:
```

patroni/dcs/kubernetes.py
```diff
@@ -536,12 +553,12 @@ def wrapper(*args, **kwargs):
...
+    def wrapper(self: 'Kubernetes', *args: Any, **kwargs: Any):
```
barthisrael, line 561:
```suggestion
    def wrapper(self: 'Kubernetes', *args: Any, **kwargs: Any) -> Any:
```

patroni/dcs/kubernetes.py
```diff
@@ -551,7 +568,8 @@ def wrapper(self, *args, **kwargs):
 
 class ObjectCache(Thread):
 
-    def __init__(self, dcs, func, retry, condition, name=None):
+    def __init__(self, dcs: 'Kubernetes', func: Callable[..., Any], retry: Retry,
+                 condition: Condition, name: Optional[str] = None) -> None:
         Thread.__init__(self)
```
barthisrael, line 573:
```suggestion
        super(ObjectCache, self).__init__()
```

patroni/dcs/kubernetes.py
```diff
@@ -1235,40 +1275,45 @@ def touch_member(self, data):
...
-    def set_history_value(self, value):
+    def set_history_value(self, value: str) -> bool:
         return self.patch_or_create_config({self._HISTORY: value}, None, bool(self._config_resource_version), False)
 
-    def set_sync_state_value(self, value, index=None):
+    def set_sync_state_value(self, value: str, index: Optional[str] = None) -> bool:
         """Unused"""
+        raise NotImplementedError  # pragma: no cover
 
     def write_sync_state(self, leader: Union[str, None], sync_standby: Union[Collection[str], None],
```
barthisrael, line 1315:
```suggestion
    def write_sync_state(self, leader: Optional[str], sync_standby: Optional[Collection[str]],
```

patroni/dcs/raft.py
```diff
@@ -223,50 +228,50 @@ def delete(self, key, recursive=False, **kwargs):
             return False
 
     @staticmethod
-    def __values_match(old, new):
+    def __values_match(old: Dict[str, Any], new: Dict[str, Any]) -> bool:
         return all(old.get(n) == new.get(n) for n in ('created', 'updated', 'expire', 'value'))
 
     @replicated
-    def _expire(self, key, value, callback=None):
+    def _expire(self, key: str, value: Dict[str, Any], callback: Optional[Callable[..., Any]] = None) -> None:
```
barthisrael, line 235:
```suggestion
    def _expire(self, key: str, value: Dict[str, Any]) -> None:
```

It seems we can remove `callback` parameter.

patroni/dcs/raft.py
```diff
@@ -223,50 +228,50 @@ def delete(self, key, recursive=False, **kwargs):
...
         for key, value in self.__data.items():
             if value and 'expire' in value and value['expire'] <= time.time() and \
                     not (key in self.__limb and self.__values_match(self.__limb[key], value)):
                 self.__limb[key] = value
 
-                def callback(*args):
+                def callback(*args: Any) -> None:
                     if key in self.__limb and self.__values_match(self.__limb[key], value):
                         self.__limb.pop(key)
                 self._expire(key, value, callback=callback)
```
barthisrael, line 249:
```suggestion
                self._expire(key, value)
```

`callback` doesn't seem to be used for anything.

patroni/postgresql/bootstrap.py
```diff
@@ -106,11 +111,11 @@ def option_is_allowed(name):
             error_handler('{0} options must be list or dict'.format(tool))
         return user_options
 
-    def _initdb(self, config):
+    def _initdb(self, config: Any) -> bool:
```
barthisrael, line 114:
```suggestion
    def _initdb(self, config: Union[Any, Dict[str, str], List[Union[str, Dict[str, Any]]]]) -> bool:
```

Just so we use the same annotation of `options` argument in `process_user_options`.

patroni/postgresql/bootstrap.py
```diff
@@ -145,7 +150,7 @@ def _post_restore(self):
         if os.path.exists(trigger_file):
             os.unlink(trigger_file)
 
-    def _custom_bootstrap(self, config):
+    def _custom_bootstrap(self, config: Any) -> bool:
```
barthisrael, line 153:
```suggestion
    def _custom_bootstrap(self, config: Union[Any, Dict[str, str], List[Union[str, Dict[str, Any]]]]) -> bool:
```

Same as previous.

barthisrael, line 153:
```suggestion
    def _custom_bootstrap(self, config: Union[Any, Dict[str, Any]]) -> bool:
```

From what I can see `config` is mainly a `dict`, but `pyright` thinks it can receive `None` given how the method is called.

patroni/postgresql/rewind.py
```diff
@@ -5,36 +5,46 @@
...
+    NEED = 3
+    NOT_NEED = 4
+    SUCCESS = 5
+    FAILED = 6
 
 
 class Rewind(object):
 
-    def __init__(self, postgresql):
+    def __init__(self, postgresql: Postgresql):
```
barthisrael, line 33:
```suggestion
    def __init__(self, postgresql: Postgresql) -> None:
```

patroni/postgresql/slots.py
```diff
@@ -55,15 +62,15 @@ def sync_slot(self, cur, database, slot, lsn):
                 if not self._scheduled[database]:
                     self._scheduled.pop(database)
 
-    def sync_slots_in_database(self, database, slots):
+    def sync_slots_in_database(self, database: str, slots: List[str]):
```
barthisrael, line 65:
```suggestion
    def sync_slots_in_database(self, database: str, slots: List[str]) -> None:
```

barthisrael:
Not able to suggest this either through the interface, so sharing the diff:

```diff
diff --git a/patroni/postgresql/sync.py b/patroni/postgresql/sync.py
index 3a96d4c..882e6fd 100644
--- a/patroni/postgresql/sync.py
+++ b/patroni/postgresql/sync.py
@@ -3,7 +3,7 @@ import re
 import time

 from copy import deepcopy
-from typing import Any, Collection, Dict, List, Tuple, TYPE_CHECKING, Union
+from typing import Any, Collection, Dict, List, Optional, Tuple, TYPE_CHECKING, Union

 from ..collections import CaseInsensitiveDict, CaseInsensitiveSet
 from ..dcs import Cluster
@@ -149,7 +149,7 @@ class SyncHandler(object):
         # "sync" replication connections, that were verified to reach self._primary_flush_lsn at some point
         self._ready_replicas = CaseInsensitiveDict({})  # keys: member names, values: connection pids

-    def _handle_synchronous_standby_names_change(self):
+    def _handle_synchronous_standby_names_change(self) -> Optional[bool]:
         """If synchronous_standby_names has changed we need to check that newly added replicas
         have reached self._primary_flush_lsn. Only after that they could be counted as sync."""
         synchronous_standby_names = self._postgresql.synchronous_standby_names()
```

patroni/psycopg.py
```diff
@@ -53,19 +55,19 @@ def _connect(dsn: str = "", **kwargs: Any) -> Any:
...
+        ret = __connect(dsn or "", **kwargs)
         setattr(ret, 'server_version', ret.pgconn.server_version)  # compatibility with psycopg2
         return ret
 
-    def _quote_ident(value: Any, conn: Connection) -> str:
+    def _quote_ident(value: Any, scope: Any) -> str:
         """Quote *value* as a SQL identifier.
 
         :param value: value to be quoted.
         :param conn: connection to evaluate the returning string into.
```
barthisrael, line 66:
```suggestion
        :param scope: connection to evaluate the returning string into.
```

barthisrael:
The following change:

```diff
diff --git a/patroni/request.py b/patroni/request.py
index bb68fe6..24178ac 100644
--- a/patroni/request.py
+++ b/patroni/request.py
@@ -140,7 +140,7 @@ class PatroniRequest(object):

         :returns: the response returned upon request.
         """
-        url = member.api_url
+        url: str = member.api_url or ''
         if endpoint:
             scheme, netloc, _, _, _, _ = urlparse(url)
             url = urlunparse((scheme, netloc, endpoint, '', '', ''))
```

Would address these `pyright` issues:

```
patroni/request.py
  patroni/request.py:146:30 - error: Argument of type "tuple[str | bytes, str | bytes, str, Literal[''], Literal[''], Literal['']]" cannot be assigned to parameter "components" of type "Iterable[AnyStr@urlunparse | None]" in function "urlunparse"
    Type "str | bytes" cannot be assigned to type "bytes | None"
      Type "str" cannot be assigned to type "bytes | None"
        "str" is incompatible with "bytes"
        Type cannot be assigned to type "None"
    Type "str | bytes" cannot be assigned to type "bytes | None"
      Type "str" cannot be assigned to type "bytes | None"
        "str" is incompatible with "bytes"
        Type cannot be assigned to type "None" (reportGeneralTypeIssues)
  patroni/request.py:146:13 - error: Type of "url" is unknown (reportUnknownVariableType)
  patroni/request.py:147:37 - error: Argument of type "Unknown | str | None" cannot be assigned to parameter "url" of type "str" in function "request"
    Type "Unknown | str | None" cannot be assigned to type "str"
      Type "None" cannot be assigned to type "str" (reportGeneralTypeIssues)
3 errors, 0 warnings, 0 informations
```

patroni/utils.py
```diff
@@ -7,9 +7,9 @@
 :var DEC_RE: regular expression to match decimal numbers, signed or unsigned.
 :var HEX_RE: regular expression to match hex strings, signed or unsigned.
 :var DBL_RE: regular expression to match double precision numbers, signed or unsigned. Matches scientific notation too.
+:var WHITESPACE_re: regular expression to match whitespace characters
```
barthisrael, line 10:
```suggestion
:var WHITESPACE_RE: regular expression to match whitespace characters
```

patroni/dcs/kubernetes.py
CyberDem0n: This one doesn't create problems.

patroni/dcs/kubernetes.py
CyberDem0n: This one will indeed create a lot of issues because not all fields in K8s objects are present in JSON.
Example:
```json
{
    "metadata": {
        "resourceVersion": "123"
        "annotations": {}
    }
}
```

In this specific case K8s API sends it as:
```json

{
    "metadata": {
        "resourceVersion": "123"
    }
}
```

We could do some workarounds like memorizing the name of the parent (in this case "metadata") and make some decisions based on it. E.g., raise AttributeError if key isn't present in the `self._dict` with exception if `metadata.annotations`, `subsets` of the endpoint are accessed.

The list of exceptions shouldn't be big, but knowing the speed of K8s development I would not rely on it.
Hence, probably we should live `__getattr__` as it is, returning Any.

patroni/dcs/raft.py
CyberDem0n: No, we can't there is some magic in the `@replicated` decorator.

patroni/dcs/raft.py
CyberDem0n: It is intercepted and called from the function implementing `@replicated` decorator.

patroni/postgresql/bootstrap.py
CyberDem0n: I tried it, but as soon as it becomes Union we have to explicitly assert types in respective methods :(

patroni/validator.py
```diff
@@ -680,14 +676,41 @@ def assert_(condition: bool, message: OptionalType[str] = "Wrong value") -> None
     assert condition, message
 
 
+class IntValidator(object):
```
barthisrael, line 679:
```suggestion
class IntValidator(object):
    """Validate an integer setting.

    :cvar expected_type: the expect Python type for an integer setting (:class:`int`).
    :ivar min: minimum allowed value for the setting, if any.
    :ivar max: maximum allowed value for the setting, if any.
    :ivar base_unit: the base unit to convert the value to before checking if it's within `min` and `max` range.
    :ivar raise_assert: if an ``assert`` call should be performed regarding expected type and valid range.
    """

```

patroni/validator.py
```diff
@@ -680,14 +676,41 @@ def assert_(condition: bool, message: OptionalType[str] = "Wrong value") -> None
...
+    expected_type = int
+
+    def __init__(self, min: OptionalType[int] = None, max: OptionalType[int] = None,
+                 base_unit: OptionalType[str] = None, raise_assert: bool = False) -> None:
```
barthisrael, line 683:
```suggestion
                 base_unit: OptionalType[str] = None, raise_assert: bool = False) -> None:
        """Create an :class:`IntValidator` object with the given rules.

        :param min: minimum allowed value for the setting, if any.
        :param max: maximum allowed value for the setting, if any.
        :param base_unit: the base unit to convert the value to before checking if it's within *min* and *max* range.
        :param raise_assert: if an ``assert`` call should be performed regarding expected type and valid range.
        """
```

patroni/validator.py
```diff
@@ -680,14 +676,41 @@ def assert_(condition: bool, message: OptionalType[str] = "Wrong value") -> None
...
+        ret = isinstance(value, int)\
+            and (self.min is None or value >= self.min)\
+            and (self.max is None or value <= self.max)
+
+        if self.raise_assert:
+            assert_(ret)
+        return ret
+
+
+def validate_watchdog_mode(value: Any):
```
barthisrael, line 701:
```suggestion
def validate_watchdog_mode(value: Any) -> None:
    """Ensure ``watchdog.mode`` setting is valid.

    .. note::
        If *value* is not valid, then an ``AssertionError`` will be triggered.

    :param value: setting value to be validated.
    """
```

pyrightconfig.json
```diff
@@ -0,0 +1,28 @@
...
+  "reportMissingImports": true,
+  "reportMissingTypeStubs": false,
+
+  "pythonVersion": "3.6",
+  "pythonPlatform": "All",
+
+  "typeCheckingMode": "strict"
+
+}
+//  "stubPath": "src/stubs",
```
barthisrael, line 28:
```suggestion
```

barthisrael:
I guess I went over all files, overall they look great IMHO :)

I only have a couple questions:

1. How are you generating the `pyi` files -- are you writing them all on your own or using some kind of aux tool?
2. It seems some modules like `urllib3` already have type hints in their source code, so I was wondering why exactly we have added `pyi` files for these?

patroni/dcs/etcd.py
```diff
@@ -480,7 +502,7 @@ def _handle_exception(self, e, name='', do_sleep=False, raise_ex=None):
         if isinstance(raise_ex, Exception):
             raise raise_ex
 
-    def _run_and_handle_exceptions(self, method, *args, **kwargs):
+    def _run_and_handle_exceptions(self, method: Callable[..., Any], *args: Any, **kwargs: Any) -> Any:
         retry = kwargs.pop('retry', self.retry)
         try:
             return retry(method, *args, **kwargs) if retry else method(*args, **kwargs)
```
matthbakeredb, line 508: The signature of the retry method doesn't match self.retry, which doesn't match the `__call__` of Retry which as `func` as the first argument.

patroni/dcs/etcd.py
```diff
@@ -451,25 +466,32 @@ def _prepare_request(self, kwargs, params=None, method=None):
...
+    @abc.abstractmethod
+    def _client(self) -> AbstractEtcdClientWithFailover:
+        """return correct type of etcd client"""
+
+    def reload_config(self, config: Union['Config', Dict[str, Any]]) -> None:
         super(AbstractEtcd, self).reload_config(config)
         self._client.reload_config(config.get(self.__class__.__name__.lower(), {}))
 
-    def retry(self, *args, **kwargs):
+    def retry(self, *args: Any, **kwargs: Any) -> Any:
```
matthbakeredb, line 488: Should this method have a `func` arg as the first arg to match the Retry class `__call__`?

patroni/dcs/etcd.py
```diff
@@ -613,18 +639,23 @@ def wrapper(self, *args, **kwargs):
...
         self.__do_not_watch = False
 
-    def set_ttl(self, ttl):
+    @property
+    def _client(self) -> EtcdClient:
+        assert isinstance(self._abstract_client, EtcdClient)
+        return self._abstract_client
+
+    def set_ttl(self, ttl: int) -> Optional[bool]:
         self.__do_not_watch = super(Etcd, self).set_ttl(ttl)
```
matthbakeredb, line 652: Suggest adding explicit `return None` to appease my IDE!

patroni/dcs/etcd3.py
```diff
@@ -725,10 +771,11 @@ def _load_cluster(self, path, loader):
         except Exception as e:
             self._handle_exception(e, 'get_cluster', raise_ex=Etcd3Error('Etcd is not responding properly'))
         self._has_failed = False
+        assert cluster is not None
         return cluster
 
     @catch_etcd_errors
-    def touch_member(self, data):
+    def touch_member(self, data: Dict[str, Any]) -> bool:
```
matthbakeredb, line 778:
```suggestion
    def touch_member(self, data: Dict[str, Any]) -> Union[Dict[str, Any], bool]:
```
Due to the return type of put method on line 792

patroni/dcs/etcd3.py
```diff
@@ -593,15 +630,22 @@ def __init__(self, config):
...
+    def set_socket_options(self, sock: socket.socket,
+                           socket_options: Optional[Collection[Tuple[int, int, int]]]) -> None:
+        assert self._retry.deadline is not None
         enable_keepalive(sock, self.ttl, int(self.loop_wait + self._retry.deadline))
 
-    def set_ttl(self, ttl):
+    def set_ttl(self, ttl: int) -> Optional[bool]:
         self.__do_not_watch = super(Etcd3, self).set_ttl(ttl)
         if self.__do_not_watch:
             self._lease = None
```
matthbakeredb, line 646:
```suggestion
            self._lease = None
        return None
```

patroni/dcs/etcd3.py
```diff
@@ -791,30 +839,30 @@ def _retry(*args, **kwargs):
         return ret
 
     @catch_etcd_errors
-    def set_failover_value(self, value, index=None):
+    def set_failover_value(self, value: str, index: Optional[str] = None) -> bool:
```
matthbakeredb, line 842:
```suggestion
    def set_failover_value(self, value: str, index: Optional[str] = None) -> Union[Dict[str, Any], bool]:
```

matthbakeredb: Same for all the next few that use put(). Not sure why you see bool?

patroni/dcs/kubernetes.py
```diff
@@ -771,31 +792,32 @@ def _run_and_handle_exceptions(method, *args, **kwargs):
...
     @property
-    def leader_path(self):
+    def leader_path(self) -> str:
         return super(Kubernetes, self).leader_path[:-7 if self._api.use_endpoints else None]
 
-    def set_ttl(self, ttl):
+    def set_ttl(self, ttl: int) -> Optional[bool]:
         ttl = int(ttl)
         self.__do_not_watch = self._ttl != ttl
         self._ttl = ttl
```
matthbakeredb, line 805:
```suggestion
        self._ttl = ttl
    return None
```

patroni/dcs/kubernetes.py
```diff
@@ -910,55 +932,62 @@ def _cluster_from_nodes(self, group, nodes, pods):
...
             logger.error('Failed to load Citus coordinator cluster from Kubernetes: %r', e)
 
     @staticmethod
-    def compare_ports(p1, p2):
+    def compare_ports(p1: K8sObject, p2: K8sObject) -> bool:
         return p1.name == p2.name and p1.port == p2.port and (p1.protocol or 'TCP') == (p2.protocol or 'TCP')
 
     @staticmethod
-    def subsets_changed(last_observed_subsets, ip, ports):
+    def subsets_changed(last_observed_subsets: List[K8sObject], ip: str, ports: List[K8sObject]) -> bool:
```
matthbakeredb, line 990: To match docstring example
```suggestion
    def subsets_changed(last_observed_subsets: List[K8sObject], ip: Optional[str], ports: List[K8sObject]) -> bool:
```

patroni/dcs/raft.py
```diff
@@ -45,17 +49,17 @@ def resolve_host(self):
...
-    def executeCommand(self, command):
+    def executeCommand(self, command: List[Any]) -> Any:
         try:
             return self._utility.executeCommand(self.__node, command)
         except Exception:
             return None
 
-    def getMembers(self):
+    def getMembers(self) -> Optional[List[str]]:
         for self.__node in self._nodes:
```
matthbakeredb, line 63: My IDE doesn't like this. Setting an attr in a method outside of init and overwriting it in a for loop. Perhaps it should have a new object type for node/command. I know raft isn't important/maintained/supported, so will ignore.

patroni/dcs/kubernetes.py
matthbakeredb, line 1: There's 3 uses of `max` that have foats as args where they're expected to be ints (as per previous)

patroni/dcs/consul.py
```diff
@@ -266,20 +281,20 @@ def reload_config(self, config):
         self._previous_loop_register_service = self._register_service
         self._register_service = should_register_service
 
-    def set_ttl(self, ttl):
+    def set_ttl(self, ttl: int) -> Optional[bool]:
         if self._client.http.set_ttl(ttl / 2.0):  # Consul multiplies the TTL by 2x
             self._session = None
             self.__do_not_watch = True
```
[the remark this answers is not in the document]
CyberDem0n: Returning None and not returning anything is the same, I don't understand why do we need it.

patroni/dcs/etcd.py
```diff
@@ -133,16 +140,17 @@ def _calculate_timeouts(self, etcd_nodes, timeout=None):
...
-    def _get_headers(self):
+    def _get_headers(self) -> Dict[str, str]:
         basic_auth = ':'.join((self.username, self.password)) if self.username and self.password else None
         return urllib3.make_headers(basic_auth=basic_auth, user_agent=USER_AGENT)
 
-    def _prepare_common_parameters(self, etcd_nodes, timeout=None):
-        kwargs = {'headers': self._get_headers(), 'redirect': self.allow_redirect, 'preload_content': False}
+    def _prepare_common_parameters(self, etcd_nodes: int, timeout: Optional[float] = None) -> Dict[str, Any]:
+        kwargs: Dict[str, Any] = {'headers': self._get_headers(),
+                                  'redirect': self.allow_redirect, 'preload_content': False}
```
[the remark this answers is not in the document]
CyberDem0n: will change it to `max(1.0, per_node_timeout / 2.0)`

patroni/dcs/etcd.py
```diff
@@ -291,18 +305,18 @@ def api_execute(self, path, method, params=None, timeout=None):
                 machines_cache = machines_cache[:nodes]
```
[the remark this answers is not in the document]
CyberDem0n: will change it to `max(1.0, per_node_timeout / 2.0)`

patroni/dcs/etcd.py
CyberDem0n: Weird. Which IDE do you use?

patroni/dcs/etcd3.py
CyberDem0n: No, it has to be boolean. I'll just add an explicit cast.

patroni/dcs/kubernetes.py
CyberDem0n: It could never happen on practice...

patroni/dcs/kubernetes.py
CyberDem0n: Will fix the same way as for etcd.

patroni/dcs/consul.py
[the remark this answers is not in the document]
matthbakeredb: I think the logic here is that if the function sometimes returns something it should always `return` even if `None`. So being "explicit is better than implicit" https://peps.python.org/pep-0020/.

Some discussion here https://youtrack.jetbrains.com/issue/PY-20603

CyberDem0n:
> How are you generating the pyi files -- are you writing them all on your own or using some kind of aux tool?

`pyright --createstub $module` does it, but some manual work is required after. At the end I just copied only required pieces manually to match features that are used by Patroni.

> It seems some modules like urllib3 already have type hints in their source code, so I was wondering why exactly we have added pyi files for these?

urllib3 has hints only in the master branch, which was never released. Stubs that come with pyright unfortunately have some issues and types are not always complete :(

patroni/ctl.py
```diff
@@ -834,11 +858,13 @@ def failover(obj, cluster_name, group, leader, candidate, force):
               default=None)
 @option_force
 @click.pass_obj
-def switchover(obj, cluster_name, group, leader, candidate, force, scheduled):
+def switchover(obj: Dict[str, Any], cluster_name: str, group: Optional[int],
+               leader: Optional[str], candidate: Optional[str], force: bool, scheduled: Optional[str]):
```
barthisrael, line 862:
```suggestion
               leader: Optional[str], candidate: Optional[str], force: bool, scheduled: Optional[str]) -> None:
```

barthisrael:
> > How are you generating the pyi files -- are you writing them all on your own or using some kind of aux tool?
> 
> `pyright --createstub $module` does it, but some manual work is required after. At the end I just copied only required pieces manually to match features that are used by Patroni.
> 
> > It seems some modules like urllib3 already have type hints in their source code, so I was wondering why exactly we have added pyi files for these?
> 
> urllib3 has hints only in the master branch, which was never released. Stubs that come with pyright unfortunately have some issues and types are not always complete :(

Thanks for clarifying!

> @barthisrael, @matthbakeredb any other comments or we can already merge it?

I just went through the commits that I was pending to take a look at, and I'd say it's good to go :)

CyberDem0n:
:+1:

matthbakeredb:
> @barthisrael, @matthbakeredb any other comments or we can already merge it?

Hi, sorry, didn't have time to come back to this. I'll take a look in retrospect and raise anything I see that looks problematic.

PR #2656, opened 2023-04-27
Fix behave tests

CyberDem0n:
1. specify `subjectAltName=IP:127.0.0.1` when generating certificate
2. run more behave tests with psycopg2


hughcapet:
👍

CyberDem0n:
:+1:

PR #2659, opened 2023-04-28
Make pod role label configurable

### LINKED ISSUE #2495: Feature request: make value of replication role labels configurable

waynerv:
resolve #2495


waynerv:
> @waynerv any thoughts?

compatibility make sense, will try some solutin later

waynerv:
Have been quite busy recently, will try to make some progress this weekend.

waynerv:
added `tmp_role_label`, but currently can't archive remove label after migration is done because `patch_namespaced_pod` can not remove a label key.

do we need that?

CyberDem0n:
>added tmp_role_label, but currently can't archive remove label after migration is done because patch_namespaced_pod can not remove a label key.

I didn't check the code yet, but...

1. PATCH with `{labels: {foo: null}}` removes label `foo`
2. when pod is restarted labels that were not specified on pod creation will also be gone.

waynerv:
> PATCH with {labels: {foo: null}} removes label foo

Didn't know there was such a convenient way before. Thought we had to use jsonpath method to explicitly remove it.

But we still have a problem in front of us. If user feels that `tmp_role` is no longer needed, how should they remove it? If `tmp_role` is changed to another value or removed, after reloading the configuration, we will not be able to know which key needs to be removed in Patroni.

One feasible approach is to remove the label if the `tmp_role` ends with a hyphen like `tmp_role_label="tmp_role-"`. However, this method obviously adds some complexity.

CyberDem0n:
> If tmp_role is changed to another value or removed, after reloading the configuration

First of all current implementation doesn't support reloading, and second, it is not even necessary to support it, because the typical workflow is updating StatefulSet or Deployment with the new configuration and restarting Pods.

waynerv:
so every time the tmp_role configuration is modified, pod need be recreated to effect, then there is no need to concern about removing the label.

Current code implementation should be able to meet the entire workflow.

CyberDem0n:
K8s is very special environment.
The normal workflow:
1. update StatefulSet or Deployment objects that are responsible of creating Pods
2. do a rolling restart of Pods.

Nobody should be messing up with Patroni configuration by exec-ing into the Pod. These changes are not persistent and will be lost if pod is restarted.

docs/yaml_configuration.rst
```diff
@@ -155,7 +155,10 @@ Kubernetes
...
 -  **labels**: Labels in format ``{label1: value1, label2: value2}``. These labels will be used to find existing objects (Pods and either Endpoints or ConfigMaps) associated with the current cluster. Also Patroni will set them on every object (Endpoint or ConfigMap) it creates.
 -  **scope\_label**: (optional) name of the label containing cluster name. Default value is `cluster-name`.
--  **role\_label**: (optional) name of the label containing role (master or replica). Patroni will set this label on the pod it runs in. Default value is ``role``.
+-  **role\_label**: (optional) name of the label containing role (master or replica or other custom value). Patroni will set this label on the pod it runs in. Default value is ``role``.
+-  **leader\_label\_value**: (optional) value of the pod label when Postgres role is ``master``. Default value is ``master``.
+-  **follower\_label\_value**: (optional) value of the pod label when Postgres role is ``replica``. Default value is ``replica``.
+-  **tmp_\role\_label**: (optional) name of the label containing role (master or replica). Patroni will set this label on the pod it runs in. Default value is role_label thus no extra label will set.
```
CyberDem0n, line 161: The default should be nothing, and when no value is set it shouldn't be even used here: https://github.com/zalando/patroni/pull/2659/files#diff-0b47c783038695efd3a30f9797eb9699abc996d28c5ec9f040b81e63e995b068R1290

patroni/dcs/kubernetes.py
```diff
@@ -1263,19 +1266,28 @@ def set_config_value(self, value: str, version: Optional[str] = None) -> bool:
     def touch_member(self, data: Dict[str, Any]) -> bool:
         cluster = self.cluster
         if cluster and cluster.leader and cluster.leader.name == self._name:
-            role = 'master'
+            tmp_role = 'master'
+            role = self._leader_label_value
         elif data['state'] == 'running' and data['role'] not in ('master', 'primary'):
-            role = data['role']
+            tmp_role = data['role']
+            role = self._follower_label_value if data['role'] == 'replica' else data['role']
```
CyberDem0n, line 1273: That is, we are fine with hardcoded role=standby_leader?

patroni/config.py
```diff
@@ -469,12 +469,14 @@ def _get_auth(name: str, params: Optional[Collection[str]] = None) -> Dict[str,
...
-                              'REGISTER_SERVICE', 'SERVICE_CHECK_INTERVAL', 'SERVICE_CHECK_TLS_SERVER_NAME',
-                              'SERVICE_TAGS', 'NAMESPACE', 'CONTEXT', 'USE_ENDPOINTS', 'SCOPE_LABEL', 'ROLE_LABEL',
-                              'POD_IP', 'PORTS', 'LABELS', 'BYPASS_API_SERVICE', 'RETRIABLE_HTTP_CODES', 'KEY_PASSWORD',
-                              'USE_SSL', 'SET_ACLS', 'GROUP', 'DATABASE') and name:
+                if suffix in (
+                    'HOST', 'HOSTS', 'PORT', 'USE_PROXIES', 'PROTOCOL', 'SRV', 'SRV_SUFFIX', 'URL', 'PROXY',
+                    'CACERT', 'CERT', 'KEY', 'VERIFY', 'TOKEN', 'CHECKS', 'DC', 'CONSISTENCY',
+                    'REGISTER_SERVICE', 'SERVICE_CHECK_INTERVAL', 'SERVICE_CHECK_TLS_SERVER_NAME',
+                    'SERVICE_TAGS', 'NAMESPACE', 'CONTEXT', 'USE_ENDPOINTS', 'SCOPE_LABEL', 'ROLE_LABEL',
+                    'POD_IP', 'PORTS', 'LABELS', 'BYPASS_API_SERVICE', 'RETRIABLE_HTTP_CODES', 'KEY_PASSWORD',
```
CyberDem0n, line 477: Unrelated changes!

patroni/config.py
```diff
@@ -469,12 +469,14 @@ def _get_auth(name: str, params: Optional[Collection[str]] = None) -> Dict[str,
...
+                    'USE_SSL', 'SET_ACLS', 'GROUP', 'DATABASE', 'LEADER_LABEL_VALUE', 'FOLLOWER_LABEL_VALUE', 'TMP_ROLE_LABEL'
```
CyberDem0n, line 478: Flake8 complains about this line

CyberDem0n (commented):
What about unit/behave tests?

patroni/dcs/kubernetes.py
waynerv: added a `standby_leader_label_value`

waynerv:
added some unit tests, but have no idea why did behave test failed 😢

CyberDem0n:
The code looks good, @waynerv.
Now we also need a some documentation that describes the migration from old labels to new ones without downtime.

waynerv:
docs committed, do I need squash all the commits?

docs/kubernetes.rst
```diff
@@ -32,6 +32,55 @@ Configuration
...
 
+Customize role label
+^^^^^^^^^^^^^^^^^
+By default, Patroni will set corresponding labels on the pod it runs in based on node's role, such as ``role=master``.
+The key and value of label can be customized by `kubernetes.role_label`, `kubernetes.leader_label_value`, `kubernetes.follower_label_value` and `kubernetes.standby_leader_label_value`.
+
+Note that if you migrate from default role labels to custom ones, you can reduce downtime by following migration steps:
+
+1. Add a temporary label using original role value for the pod with `kubernetes.tmp_role_label`(like ``tmp_role``). Updated pod looks like:
+
```
CyberDem0n, line 43: I would better prefer using `master`/`primary` as an example. This is something most people will have to deal with, while `replica` probably stay as it is.
Or, showing two different cases (primary and replica) would be even better.

CyberDem0n: > Updated pod looks like:

I would probably rephrase it as: `Once pods are restarted they will get following labels set by Patroni:`

docs/kubernetes.rst
```diff
@@ -32,6 +32,55 @@ Configuration
...
+2. After all pods have been updated, modify the service selector to select the temporary label.
+
+  .. code:: YAML
+
+    selector:
+      cluster-name: foo
+      tmp_role: replica
+
+3. Add your custom role label (e.g., set `kubernetes.follower_label_value=standby`). Updated pod looks like:
```
CyberDem0n, line 59: I would again emphasize here that pods should be restarted and after that Patroni will set new labels.

docs/kubernetes.rst
```diff
@@ -32,6 +32,55 @@ Configuration
 
 Patroni Kubernetes :ref:`settings <kubernetes_settings>` and :ref:`environment variables <kubernetes_environment>` are described in the general chapters of the documentation.
 
+Customize role label
+^^^^^^^^^^^^^^^^^
```
CyberDem0n, line 36:
```suggestion
^^^^^^^^^^^^^^^^^^^^
```

CyberDem0n (commented):
I left a few comments.
Besides that rst formatting is not always correctly rendered.

docs/kubernetes.rst
waynerv: my bad, haven't write rst for 2 years

docs/kubernetes.rst
waynerv: I think a master/primary would be good enough

matthbakeredb (approved):
lgtm

barthisrael (approved):
Looks good!

hughcapet:
👍

CyberDem0n:
:+1:

PR #2660, opened 2023-04-28
Remove watchdog from __DEFAULT_CONFIG

hughcapet:
Watchdog is assumed to be configurable via the dynamic config, however when one tries to add this section using `patronictl edit-config`, it fails to apply changes on the `_safe_copy_dynamic_configuration` stage:

```
2023-04-28 12:37:07,516 ERROR: Exception when setting dynamic_configuration                                                                    
  File "/patroni/config.py", line 305, in set_dynamic_configuration                                                                            
    self.__effective_configuration = self._build_effective_configuration(configuration,                                                        
  File "/patroni/config.py", line 510, in _build_effective_configuration                                                                       
    config = self._safe_copy_dynamic_configuration(dynamic_configuration)                                                                      
  File "/patroni/config.py", line 349, in _safe_copy_dynamic_configuration
    config[name] = int(value)
TypeError: int() argument must be a string, a bytes-like object or a number, not 'dict'
```

Upd: it is not actually assumed to be configurable, so remove it from __DEFAULT_CONFIG to avoid any further confusion

### DIFF
--- a/patroni/config.py
+++ b/patroni/config.py
@@ -196,9 +196,6 @@ class Config(object):
             'use_slots': True,
             'parameters': CaseInsensitiveDict({p: v[0] for p, v in ConfigHandler.CMDLINE_OPTIONS.items()
                                                if p not in ('wal_keep_segments', 'wal_keep_size')})
-        },
-        'watchdog': {
-            'mode': 'automatic',
         }
     }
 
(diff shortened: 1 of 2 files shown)


hughcapet:
you mean if we invent a better workaround we don't need it in the global config?

CyberDem0n:
Yes, something like this. Basically it only covers a case when there is no `watchdog` section in the patroni.yaml at all.
We could do something like `watchdog_config = config.get('watchdog') or {'mode': 'automatic'}` in the `WatchdogConfig` constructor.

CyberDem0n:
:+1:

PR #2666, opened 2023-05-09
More typing in sync.py

CyberDem0n:
- make parse_sync_standby_names() return NamedTuple instead of dict
- fix little issue in Postgresql.reset_cluster_info_state(), it should check global_config independently from cluster and cluster.config


hughcapet:
👍

CyberDem0n:
:+1:

PR #2668, opened 2023-05-09
Quorum based failover

CyberDem0n:
Based on https://github.com/zalando/patroni/pull/672

To enable quorum commit:
```diff
$ patronictl.py edit-config
--- 
+++ 
@@ -5,3 +5,4 @@
   use_pg_rewind: true
 retry_timeout: 10
 ttl: 30
+synchronous_mode: quorum

Apply these changes? [y/N]: y
Configuration changed
```

By default Patroni will use `ANY 1(list,of,stanbys)` in `synchronous_standby_names`. That is, only one node out of listed replicas will be used for quorum.
If you want to increase the number of quorum nodes it is possible to do it with:
```diff
$ patronictl edit-config
--- 
+++ 
@@ -6,3 +6,4 @@
 retry_timeout: 10
 synchronous_mode: quorum
 ttl: 30
+synchronous_node_count: 2

Apply these changes? [y/N]: y
Configuration changed
```


patroni/ha.py
```diff
@@ -846,17 +987,24 @@ def check_failsafe_topology(self) -> bool:
...
-        """This method tries to determine whether I am healthy enough to became a new leader candidate or not."""
-
+        """This method tries to determine whether the current node is healthy enough to became a new leader candidate.
+
+        :param members: the list of nodes to check against
+        :param check_replication_lag: whether to take the replication lag into account.
+                                      If the lag exceeds configured threshold the node disqualifies itself.
+        :returns: `True` in case if the node is eligible to become the new leader. Since this method is executed
+                  on multiple nodes independently it could happen that many nodes will count themselves as
+                  healthiest because they received/replayed up to the same LSN, but it is totally fine.
```
matthbakeredb, line 1006:
```suggestion
        :returns: `True` if the node is eligible to become the new leader. Since this method is executed
                  on multiple nodes independently it is possible that multiple nodes could count 
                  themselves as the healthiest because they received/replayed up to the same LSN,
                  but this is totally fine.
```

patroni/postgresql/sync.py
```diff
@@ -239,52 +232,119 @@ def current_state(self, cluster: Cluster) -> Tuple[CaseInsensitiveSet, CaseInsen
...
+                    # "really" synchronous when sync_state = 'sync' and we known that it managed to catch up
+                    self._ready_replicas[app_name] = pid
+
+    def current_state(self, cluster: Cluster) -> _SyncState:
+        """Finds best candidates to be the synchronous standbys.
+
+        Current synchronous standby is always preferred, unless it has disconnected or does not want to be a
+        synchronous standby any longer.
+
+        Standbys are selected based on values from the global configuration:
```
matthbakeredb, line 261: The bullet points need to be indented and have line breaks after each one to be properly formatted in sphinx

patroni/postgresql/sync.py
```diff
@@ -239,52 +232,119 @@ def current_state(self, cluster: Cluster) -> Tuple[CaseInsensitiveSet, CaseInsen
...
+        - `maximum_lag_on_syncnode`: would help swapping unhealthy sync replica in case if it stops
+          responding (or hung). Please set the value high enough so it won't unncessarily swap sync
+          standbys during high loads. Any value less or equal of 0 keeps the behavior backward compatible.
+          Please note that it will not also swap sync standbys in case where all replicas are hung.
+        - `synchronous_node_count`: controlls how many nodes should be set as synchronous.
+
+        :param cluster: current cluster topology from DCS
+        :returns: current synchronous replication state as a :class:`_SyncState` object
```
matthbakeredb, line 269:
```suggestion
        :param cluster: current cluster topology from DCS
        
        :returns: current synchronous replication state as a :class:`_SyncState` object
```

matthbakeredb (changes requested):
partial review...

There's a lot to understand here that I don't feel equipped to review.

My general feel for some of the methods is that they are very large and are steps in a flow that could be better expressed as separate methods or possibly separate classes. (I don't know if Patroni needs some way to express complex workflows in a way that is easier to understand.)

Some of the concepts are new to me, I can't say I understand what this is doing from reading the documentation. Do you have any material where you've talked about this, perhaps at a conference?

patroni/api.py
```diff
@@ -295,10 +295,13 @@ def do_GET(self, write_status_code_only: bool = False) -> None:
         elif 'health' in path:
             status_code = 200 if response.get('state') == 'running' else 503
         elif cluster:  # dcs is available
+            is_quorum = response.get('quorum_standby')
             is_synchronous = response.get('sync_standby')
             if path in ('/sync', '/synchronous') and is_synchronous:
                 status_code = replica_status_code
-            elif path in ('/async', '/asynchronous') and not is_synchronous:
+            elif path == '/quorum' and is_quorum:
```
barthisrael, line 302: Missing to add `/quorum` to method docstring.

barthisrael (commented):
I still need to continue going through the PR later, but I'm sharing some comments that I had so far.

CyberDem0n:
> There's a lot to understand here that I don't feel equipped to review.

Yes, there is quite a complex logic in `QuorumStateResolver` :(

My general feel for some of the methods is that they are very large and are steps in a flow that could be better expressed as separate methods or possibly separate classes. (I don't know if Patroni needs some way to express complex workflows in a way that is easier to understand.)

In this specific case codacy doesn't complain about added or introduced complexity. Well, it was, but two days ago I split a couple of methods to smaller ones.

> Some of the concepts are new to me, I can't say I understand what this is doing from reading the documentation. Do you have any material where you've talked about this, perhaps at a conference?

I would suggest starting from https://github.com/zalando/patroni/issues/664 and https://github.com/zalando/patroni/pull/672. They will help to understand the most complex part. Also the QuorumStateResolver [docstring](https://github.com/zalando/patroni/pull/2668/files#diff-83dcf5d703da93054bcd0ac903b574fe8cbbda165de60c6f56c8bff62970cbf1R14-R49).

patroni/dcs/etcd3.py
```diff
@@ -939,6 +939,6 @@ def watch(self, leader_version: Optional[str], timeout: float) -> bool:
             timeout += 0.5
 
         try:
-            return super(Etcd3, self).watch(None, timeout)
+            return super(Etcd3, self).watch(None, timeout + 0.5)
```
barthisrael, line 942: What's the motivation of this change?

barthisrael (commented):
Some more thoughts.

docs/replication_modes.rst
```diff
@@ -79,6 +93,24 @@ Patroni will only assign one or more synchronous standby nodes based on ``synchr
...
+Quorum commit mode
+==================
+
+Starting from PostgreSQL v10 Patroni supports quorum-based synchronous replication.
+
+In this mode, Patroni maintains synchronization state in the DCS, containing the latest known primary, the number of nodes required for quorum and the nodes currently eligible to vote on quorum. In steady state, the nodes voting on quorum are the leader and all synchronous standbys. This state is updated with strict ordering constraints, with regards to node promotion and ``synchronous_standby_names``, to ensure that at all times any subset of voters that can achieve quorum includes at least one node with the latest successful commit.
+
+On each iteration of HA loop, Patroni re-evaluates synchronous standby choices and quorum, based on node availability and requested cluster configuration. In PostgreSQL versions above 9.6 all eligible nodes are added as synchronous standbys as soon as their replication catches up to leader.
+
+Quorum commit helps to reduce worst case latencies, even during normal operation, as a higher latency of replicating to one standby can be compensated by other standbys.
```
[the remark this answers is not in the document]
CyberDem0n: > My understanding so far is that roughly:
`synchronous_mode: on`: sets `synchronous_standby_names`: `<synchronous_node_count> (<list_of_standbys>)`

Almost. It will be not a `<synchronous_node_count>`, but `len(<list_of_standbys>)`, where `<list_of_standbys>` are nodes that are eligible (state == 'running', don't have `nosync` tag, and don't lag more than `maximum_lag_on_syncnode`) to be synchronous and the total number of nodes in the list will not exceed `synchronous_node_count`.

> `synchronous_mode: quorum`: sets `synchronous_standby_names:` `ANY <synchronous_node_count> (<list_of_ALL_standbys>)`

Also almost.
- `<list_of_ALL_standbys>` -- not all nodes, but nodes with state == 'running', without `nosync` tag, and not lagging more than `maximum_lag_on_syncnode`.
- `synchronous_node_count` -- there will be a value `N`: `1 <= N <= min(synchronous_node_count, len(<list_of_standbys>))`

The main difference between `synchronous_mode` `on` and `quorum` that in the first case Patroni picks a fixed number of synchronous nodes and they are all synchronous. During failover with `synchronous_mode: on` we can **safely promote any node** from this  list and there will be no client visible data loss.

With `quorum` there will be something like `ANY 2 (n1, n2, n3)` in `synchronous_standby_names` and in order to guaranty that there is no data loss during failover Patroni has to check some minimum subset of nodes from this list to prove that the given node received not less WAL stream than others.
Example:
primary: n0
quorum: ANY 2 (n1, n2, n3)

That is, two nodes from this least could be promoted without client visible data loss, and every node need to figure out whether it has all data.
Lets say n0 and n1 are failed. But, we know that at least two nodes from n1, n2, n3 have the latest data, therefore at least one from n2, n3 also has it. Therefore n2 **must** see n3 and compare LSNs to make decision about promotion. If n2 is not behind of n3, than n2 could be promoted and there will be no client visible data loss.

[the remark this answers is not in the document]
barthisrael: Cool, thanks for clarifying!

I guess it would be valuable to describe this in the docs, even if with a simpler approach, just so users can getter a better understanding of the difference between each of the synchronous modes.

BTW I intend to get back to reviewing this PR today.

docs/replication_modes.rst
```diff
@@ -55,28 +58,122 @@ up.
...
 
+
 Synchronous Replication Factor
-------------------------------
-The parameter ``synchronous_node_count`` is used by Patroni to manage number of synchronous standby databases. It is set to 1 by default. It has no effect when ``synchronous_mode`` is set to off. When enabled, Patroni manages precise number of synchronous standby databases based on parameter ``synchronous_node_count`` and adjusts the state in DCS & synchronous_standby_names as members join and leave.
+==============================
+
+The parameter ``synchronous_node_count`` is used by Patroni to manage number of synchronous standby databases. It is set to 1 by default. It has no effect when ``synchronous_mode`` is set to off. When enabled, Patroni manages precise number of synchronous standby databases based on parameter ``synchronous_node_count`` and adjusts the state in DCS & ``synchronous_standby_names`` in PostgreSQL as members join and leave. If the parameter is set to a value higher than the number of eligible nodes it will be automatically reduced by Patroni down.
```
barthisrael, line 69:
```suggestion
The parameter ``synchronous_node_count`` is used by Patroni to manage the number of synchronous standby databases. It is set to ``1`` by default. It has no effect when ``synchronous_mode`` is set to ``off``. When enabled, Patroni manages the precise number of synchronous standby databases based on parameter ``synchronous_node_count`` and adjusts the state in DCS & ``synchronous_standby_names`` in PostgreSQL as members join and leave. If the parameter is set to a value higher than the number of eligible nodes it will be automatically reduced by Patroni.
```

docs/replication_modes.rst
```diff
@@ -55,28 +58,122 @@ up.
...
+
+postgresql.conf
+^^^^^^^^^^^^^^^
+
+.. code-block:: INI
+
+    synchronous_standby_names = 'ANY 2 (node1,node2,node3)'
+
+
+If the primary (``node0``) failed, in the above example two of the ``node1``, ``node2``, ``node3`` will have the latest transaction received, but we don't know which ones. To figure out whether the node ``node1`` has received the latest transaction, we needs to compare its LSN with the LSN on **at least** one node (``quorum=1`` in the ``/sync`` key) of ``node2`` and ``node3``. If ``node1`` isn't behind of at least one of them, we can guaranty that there will be no user visible data loss if ``node1`` is promoted.
```
barthisrael, line 176:
```suggestion
If the primary (``node0``) failed, in the above example two of the ``node1``, ``node2``, ``node3`` will have the latest transaction received, but we don't know which ones. To figure out whether the node ``node1`` has received the latest transaction, we need to compare its LSN with the LSN on **at least** one node (``quorum=1`` in the ``/sync`` key) among ``node2`` and ``node3``. If ``node1`` isn't behind of at least one of them, we can guarantee that there will be no user visible data loss if ``node1`` is promoted.
```

barthisrael (commented):
More doc-related suggestions.

patroni/utils.py
```diff
@@ -765,11 +765,12 @@ def cluster_as_json(cluster: 'Cluster', global_config: Optional['GlobalConfig']
     cluster_lsn = cluster.last_lsn or 0
 
     ret: Dict[str, Any] = {'members': []}
+    sync_role = 'quorum_standby' if global_config.is_quorum_commit_mode else 'sync_standby'
     for m in cluster.members:
         if m.name == leader_name:
             role = 'standby_leader' if global_config.is_standby_cluster else 'leader'
-        elif cluster.sync.matches(m.name):
-            role = 'sync_standby'
+        elif cluster.sync.matches(m.name, global_config.is_quorum_commit_mode):
```
barthisrael, line 772: Why do we need to check for the leader when `synchronous_mode: quorum`?

patroni/postgresql/sync.py
```diff
@@ -153,6 +153,26 @@ def parse_sync_standby_names(value: str) -> _SSN:
     return _SSN(sync_type, has_star, num, members)
 
 
+class _SyncState(NamedTuple):
+    """Class representing the current synchronous state.
+
+    :ivar sync_type: possible values: 'off', 'priority', 'quorum'
+    :ivar numsync: how many nodes are required to be synchronous (according to ``synchronous_standby_names``).
+                   Is ``0`` in case if synchronous_standby_names value is invalid or has ``*``.
```
barthisrael, line 161:
```suggestion
                   Is ``0`` in case if ``synchronous_standby_names`` value is invalid or has ``*``.
```

patroni/postgresql/sync.py
```diff
@@ -201,35 +221,35 @@ def _handle_synchronous_standby_names_change(self) -> None:
 END;$$""")
         self._postgresql.reset_cluster_info_state(None)  # Reset internal cache to query fresh values
 
-    def current_state(self, cluster: Cluster) -> Tuple[CaseInsensitiveSet, CaseInsensitiveSet]:
-        """Finds best candidates to be the synchronous standbys.
+    def _get_replica_list(self, cluster: Cluster) -> Iterator[Tuple[int, str, str, int, bool]]:
+        """Yields candidates based on higher replay/remote_write/flush lsn.
```
barthisrael, line 225:
```suggestion
        """Yields candidates based on higher replay/write/flush LSN.
```

I guess you mean the fields from ``pg_stat_replication``.

patroni/postgresql/sync.py
```diff
@@ -201,35 +221,35 @@ def _handle_synchronous_standby_names_change(self) -> None:
...
 
-        Current synchronous standby is always preferred, unless it has disconnected or does not want to be a
-        synchronous standby any longer.
+        .. note::
+            Tuples are reverse ordered by sync_state and LSN fields so nodes that already synchronous or having
```
barthisrael, line 228:
```suggestion
            Tuples are reverse ordered by ``sync_state`` and LSN fields so nodes that already synchronous or having
```

patroni/postgresql/sync.py
```diff
@@ -201,35 +221,35 @@ def _handle_synchronous_standby_names_change(self) -> None:
...
+        :param cluster: current cluster topology from DCS.
+
+        :yields: tuples composed of:
 
-        :returns: tuple of candidates :class:`CaseInsensitiveSet` and synchronous standbys :class:`CaseInsensitiveSet`.
+            * pid - a PID of walsender process
+            * member name - matches with the application_name
+            * sync_state - one of ("async", "potential", "quorum", "sync")
+            * LSN - write_lsn, flush_lsn, or replica_lsn, depending on the value of ``synchronous_commit`` GUC
+            * nofailover - whether the member has ``nofailover`` tag set
```
barthisrael, line 239:
```suggestion
            * ``pid`` - PID of the walsender process
            * ``member name`` - matches with the ``application_name```
            * ``sync_state`` - one of (``async``, ``potential``, ``quorum``, ``sync``)
            * ``LSN`` - write_lsn, flush_lsn, or replica_lsn, depending on the value of ``synchronous_commit`` GUC
            * ``nofailover`` - whether the member has ``nofailover`` tag set
```

Just highlighting some literals :)

patroni/postgresql/sync.py
```diff
@@ -239,52 +259,122 @@ def current_state(self, cluster: Cluster) -> Tuple[CaseInsensitiveSet, CaseInsen
         for pid, app_name, sync_state, replica_lsn in sorted(pg_stat_replication, key=lambda r: r[2:4], reverse=True):
             member = members.get(app_name)
             if member and member.is_running and not member.tags.get('nosync', False):
-                replica_list.append((pid, member.name, sync_state, replica_lsn, bool(member.nofailover)))
+                yield (pid, member.name, sync_state, replica_lsn, bool(member.nofailover))
 
-        max_lsn = max(replica_list, key=lambda x: x[3])[3]\
-            if len(replica_list) > 1 else self._postgresql.last_operation()
+    def _process_replica_readiness(self, cluster: Cluster, replica_list: List[Tuple[int, str, str, int, bool]]) -> None:
+        """Flags replicas as truely "synchronous" when they caught up with "_primary_flush_lsn"."""
```
barthisrael, line 265:
```suggestion
        """Flags replicas as truly "synchronous" when they caught up with "_primary_flush_lsn"."""
```

barthisrael: We are missing to describe the parameters of this method in the docstring.

patroni/postgresql/sync.py
barthisrael: Do you think it is worth it mentioning that nodes with `nosync` tag are not considered as the idea of the method is to get a list of replicas that can be synchronous?

patroni/utils.py
CyberDem0n:
```suggestion
        elif cluster.sync.matches(m.name):
```

matthbakeredb (changes requested):
Another batch of comments. Still not reviewed entirely although I think I understand a little better what this is doing. Thanks for the pointers!

patroni/quorum.py
```diff
@@ -0,0 +1,316 @@
...
+    """Calculates a list of state transition tuples of the form `('sync'/'quorum'/'restart',leader,number,set_of_names)`
+
+    Synchronous replication state is set in two places. PostgreSQL configuration sets how many and which nodes are
+    needed for a commit to succeed, abbreviated as `numsync` and `sync` set here. DCS contains information about how
+    many and which nodes need to be interrogated to be sure to see an xlog position containing latest confirmed commit,
+    abbreviated as `quorum` and `voters` set. Both pairs have the meaning "ANY n OF set".
+
+    The number of nodes needed for commit to succeed, `numsync`, is also called the replication factor.
+
+    To guarantee zero lost transactions on failover we need to keep the invariant that at all times any subset of
```
barthisrael, line 24:
```suggestion
    To guarantee zero transaction loss on failover we need to keep the invariant that at all times any subset of
```

patroni/quorum.py
```diff
@@ -0,0 +1,316 @@
...
+    nodes that can acknowledge a commit overlaps with any subset of nodes that can achieve quorum to promote a new
+    leader. Given a desired replication factor and a set of nodes able to participate in sync replication there
+    is one optimal state satisfying this condition. Given the node set `active`, the optimal state is:
```
barthisrael, line 27:
```suggestion
    is one optimal state satisfying this condition. Given the node set ``active``, the optimal state is:
```

patroni/quorum.py
```diff
@@ -0,0 +1,316 @@
...
+
+class QuorumError(Exception):
+    pass
+
+
+class QuorumStateResolver(object):
+    """Calculates a list of state transition tuples of the form `('sync'/'quorum'/'restart',leader,number,set_of_names)`
+
+    Synchronous replication state is set in two places. PostgreSQL configuration sets how many and which nodes are
+    needed for a commit to succeed, abbreviated as `numsync` and `sync` set here. DCS contains information about how
```
barthisrael, line 18:
```suggestion
    needed for a commit to succeed, abbreviated as ``numsync`` and ``sync`` set here. DCS contains information about how
```

patroni/quorum.py
```diff
@@ -0,0 +1,316 @@
...
+    The number of nodes needed for commit to succeed, `numsync`, is also called the replication factor.
+
+    To guarantee zero lost transactions on failover we need to keep the invariant that at all times any subset of
+    nodes that can acknowledge a commit overlaps with any subset of nodes that can achieve quorum to promote a new
+    leader. Given a desired replication factor and a set of nodes able to participate in sync replication there
+    is one optimal state satisfying this condition. Given the node set `active`, the optimal state is:
+
+        sync = voters = active
+        numsync = min(sync_wanted, len(active))
+        quorum = len(active) - numsync
```
barthisrael, line 31: Interesting! So `quorum` in Patroni means something like "how many peers among the possible synchronous standby nodes do I need to consult to be sure I'm the best candidate for promotion?"

patroni/quorum.py
```diff
@@ -0,0 +1,316 @@
...
+
+    We need to be able to produce a series of state changes that take the system to this desired state from any
+    other state arbitrary given arbitrary changes is node availability, configuration and interrupted transitions.
```
barthisrael, line 34:
```suggestion
    other arbitrary state given arbitrary changes is node availability, configuration and interrupted transitions.
```

patroni/quorum.py
```diff
@@ -0,0 +1,316 @@
...
+
+    To keep the invariant the rule to follow is that when increasing `numsync` or `quorum`, we need to perform the
```
barthisrael, line 36:
```suggestion
    To keep the invariant the rule to follow is that when increasing ``numsync`` or ``quorum``, we need to perform the
```

patroni/quorum.py
```diff
@@ -0,0 +1,316 @@
...
+    increasing operation first. When decreasing either, the decreasing operation needs to be performed later.
+
+    Order of adding or removing nodes from sync and voters depends on the state of synchronous_standby_names:
```
barthisrael, line 39:
```suggestion
    Order of adding or removing nodes from ``sync`` and ``voters`` depends on the state of ``synchronous_standby_names``:
```

patroni/quorum.py
```diff
@@ -0,0 +1,316 @@
...
+    When adding new nodes:
+        if sync (synchronous_standby_names) is empty:
```
barthisrael, line 41:
```suggestion
        if ``sync`` (``synchronous_standby_names``) is empty:
```

patroni/quorum.py
```diff
@@ -0,0 +1,316 @@
...
+            add new nodes first to sync and then to voters when numsync_confirmed > 0
```
barthisrael, line 42:
```suggestion
            add new nodes first to ``sync`` and then to ``voters`` when ``numsync_confirmed`` > ``0``
```

patroni/quorum.py
```diff
@@ -0,0 +1,316 @@
...
+        else:
+            add new nodes first to voters and than to sync
```
barthisrael, line 44:
```suggestion
            add new nodes first to ``voters`` and then to ``sync``
```

patroni/quorum.py
```diff
@@ -0,0 +1,316 @@
...
+    When removing nodes:
+        if sync (synchronous_standby_names) will become empty after removal:
```
barthisrael, line 46:
```suggestion
        if ``sync`` (``synchronous_standby_names``) will become empty after removal:
```

patroni/quorum.py
```diff
@@ -0,0 +1,316 @@
...
+            first remove nodes from voters and than from sync
```
barthisrael, line 47:
```suggestion
            first remove nodes from ``voters`` and then from ``sync``
```

patroni/quorum.py
```diff
@@ -0,0 +1,316 @@
...
+        else:
+            first remove nodes from sync and than from voters. make voters empty if numsync_confirmed == 0"""
```
barthisrael, line 49:
```suggestion
            first remove nodes from ``sync`` and then from ``voters``. Make ``voters`` empty if ``numsync_confirmed`` == ``0``"""
```

patroni/quorum.py
```diff
@@ -0,0 +1,316 @@
...
+
+        sync = voters = active
+        numsync = min(sync_wanted, len(active))
+        quorum = len(active) - numsync
+
+    We need to be able to produce a series of state changes that take the system to this desired state from any
+    other state arbitrary given arbitrary changes is node availability, configuration and interrupted transitions.
+
+    To keep the invariant the rule to follow is that when increasing `numsync` or `quorum`, we need to perform the
+    increasing operation first. When decreasing either, the decreasing operation needs to be performed later.
```
barthisrael, line 37: "first" before of what?
"later" after what?

patroni/quorum.py
barthisrael, line 39: Why does it depend on `synchronous_standby_names`?
(Sorry if I'm putting too many questions, I'm still going over files 😅 Maybe some of them will be answered when I take a look somewhere else).

barthisrael (commented):
Posting some comments.

patroni/quorum.py
```diff
@@ -0,0 +1,316 @@
...
+            raise QuorumError("Mismatched sets: quorum only=%s sync only=%s" %
+                              (voters - sync, sync - voters))
+
+    def quorum_update(self, quorum: int, voters: CaseInsensitiveSet, leader: Optional[str] = None,
+                      adjust_quorum: Optional[bool] = True) -> Iterator[Tuple[str, str, int, CaseInsensitiveSet]]:
+        """Updates quorum, voters and optionally leader fields.
+
+        :param quorum: the new value for `self.quorum`, could be adjusted depending
+                       on values of `self.numsync_confirmed` and `adjust_quorum`
+        :param voters: the new value for `self.voters`, could be adjusted if numsync_confirmed == 0
```
barthisrael, line 90:
```suggestion
        :param voters: the new value for :attr:`voters`, could be adjusted if :attr:`numsync_confirmed` == ``0``
```

patroni/quorum.py
```diff
@@ -0,0 +1,316 @@
...
+        :param leader: the new value for `self.leader`, optional
```
barthisrael, line 91:
```suggestion
        :param leader: the new value for :attr:`leader`, optional
```

patroni/quorum.py
```diff
@@ -0,0 +1,316 @@
...
+        :param adjust_quorum: if set to `True` the quorum requirement will be increased by the
```
barthisrael, line 92:
```suggestion
        :param adjust_quorum: if set to ``True`` the quorum requirement will be increased by the
```

patroni/quorum.py
```diff
@@ -0,0 +1,316 @@
...
+                              difference between `self.numsync` and ``self.numsync_confirmed`
```
barthisrael, line 93:
```suggestion
                              difference between :attr:`numsync` and :attr:`numsync_confirmed`
```

patroni/quorum.py
```diff
@@ -0,0 +1,316 @@
...
+        :rtype: Iterator[tuple(type, leader, quorum, voters)] with the new quorum state,
```
barthisrael, line 94:
```suggestion
        :yields: the new quorum state,
```

patroni/quorum.py
```diff
@@ -0,0 +1,316 @@
...
+                where type could be 'quorum' or 'restart'. The latter means that
```
barthisrael, line 95:
```suggestion
                where type could be ``quorum`` or ``restart``. The latter means that
```

patroni/quorum.py
```diff
@@ -0,0 +1,316 @@
...
+                quorum could not be updated with the current input data
+                and the :class:`QuorumStateResolver` should be restarted.
+        :raises `QuorumError`: in case of invalid data or if invariant after transition could not be satisfied
```
barthisrael, line 98:
```suggestion
        :raises:
            :exc:`QuorumError`: in case of invalid data or if invariant after transition could not be satisfied
```

patroni/quorum.py
```diff
@@ -0,0 +1,316 @@
...
+    def check_invariants(self) -> None:
+        """Checks invatiant of synchronous_standby_names and /sync key in DCS.
+
+        :raises `QuorumError`: in case of broken state"""
+        voters = CaseInsensitiveSet(self.voters | CaseInsensitiveSet([self.leader]))
+        sync = CaseInsensitiveSet(self.sync | CaseInsensitiveSet([self.leader_wanted]))
+
+        # We need to verify that subset of nodes that can acknowledge a commit overlaps
+        # with any subset of nodes that can achieve quorum to promote a new leader.
+        if self.voters and not (len(voters | sync) <= self.quorum + self.numsync + 1):
```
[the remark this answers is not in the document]
barthisrael: (I'm trying to understand this code and how it complies with what is described in the docstring in regards to "keeping the invariants". Maybe it's just late for me at this point, and I'm struggling to consume it 😆 . Thoughts are welcome IAC!)

barthisrael (commented):
Some more comments.

patroni/postgresql/sync.py
```diff
@@ -239,52 +260,127 @@ def current_state(self, cluster: Cluster) -> Tuple[CaseInsensitiveSet, CaseInsen
...
             sync_param = next(iter(sync), None)
 
+        if TYPE_CHECKING:  # pragma: no cover
+            assert self._postgresql.global_config is not None
+
+        if self._postgresql.global_config.is_quorum_commit_mode and sync or\
+                self._postgresql.supports_multiple_sync and len(sync) > 1:
+            prefix = 'ANY ' if self._postgresql.global_config.is_quorum_commit_mode\
+                and self._postgresql.supports_quorum_commit else ''
+            sync_param = '{0}{1} ({2})'.format(prefix, num, sync_param)
```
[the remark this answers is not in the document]
matthbakeredb: Yes, agreed. Perhaps it reads better when function call results are stored in variables or variables are used more than once?

I agree, it's not a hard rule, there would be situations where it would be better to use another form. For instance, it is generally seen as better to use `logging`'s own string formatting: `logging.info("This is my %s", message)`.

patroni/postgresql/sync.py
```diff
@@ -273,70 +295,118 @@ def _process_replica_readiness(self, cluster: Cluster, replica_list: _ReplicaLis
...
         if TYPE_CHECKING:  # pragma: no cover
             assert self._postgresql.global_config is not None
         sync_node_count = self._postgresql.global_config.synchronous_node_count\
             if self._postgresql.supports_multiple_sync else 1
         sync_node_maxlag = self._postgresql.global_config.maximum_lag_on_syncnode
 
-        candidates = CaseInsensitiveSet()
-        sync_nodes = CaseInsensitiveSet()
         # Prefer members without nofailover tag. We are relying on the fact that sorts are guaranteed to be stable.
         for replica in sorted(replica_list, key=lambda x: x.nofailover):
```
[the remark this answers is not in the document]
barthisrael: Right, that would be cool indeed! My main concern is to miss anything because of the tags are handled, and that we will have 2 tags that somehow change the behavior of each other.

patroni/quorum.py
barthisrael: Thanks for confirming!

patroni/quorum.py
barthisrael: Hmm, makes sense, so we keep the cluster consistent during the transitional state. Thanks for explaining!
Maybe that description would be good to have in the docstring? It did make it much easier to get the meaning :)

patroni/quorum.py
```diff
@@ -0,0 +1,316 @@
...
+            first remove nodes from voters and than from sync
+        else:
+            first remove nodes from sync and than from voters. make voters empty if numsync_confirmed == 0"""
+
+    def __init__(self, leader: str, quorum: int, voters: Collection[str],
+                 numsync: int, sync: Collection[str], numsync_confirmed: int,
+                 active: Collection[str], sync_wanted: int, leader_wanted: str) -> None:
+        self.leader = leader                        # The leader according to the `/sync` key
+        self.quorum = quorum                        # The number of nodes we need to check when doing leader race
+        self.voters = CaseInsensitiveSet(voters)    # Set of nodes we need to check (both stored in the /sync key)
```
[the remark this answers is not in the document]
barthisrael: Maybe we could put `(stored in the /sync key)` in both lines just in case?

patroni/quorum.py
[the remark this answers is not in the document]
barthisrael: I get that this method is supposed to determine if the nodes that are stored in the DCS sync key, hence the ones that will dictate if a failover can occur, should have an overlap with the nodes that are currently set in `synchronous_standby_names`.

I'm just having a hard time to understand how this overlap operation works and guarantees what we are looking for. I mean, I'm struggling to understand how the implemented math operation ensures that.

I was also looking to understand why we need to check if either of the sets is a subset of the other?

Under a normal situation we would have `sync` = `voters`, and the thing described in the class docstring makes sense.

patroni/quorum.py
```diff
@@ -0,0 +1,326 @@
...
+        self.check_invariants()
+        logger.debug('quorum %s %s %s', self.leader, self.quorum, self.voters)
+        yield 'quorum', self.leader, self.quorum, self.voters
+
+    def sync_update(self, numsync: int, sync: CaseInsensitiveSet) -> Iterator[Tuple[str, str, int, CaseInsensitiveSet]]:
+        """Updates numsync and sync fields.
+
+        :param numsync: the new value for `self.numsync`
+        :param sync: the new value for `self.sync`
+        :rtype: Iterator[tuple('sync', leader, numsync, sync)] with the new state of synchronous_standby_names
```
barthisrael, line 145:
```suggestion
        :yields: new state of synchronous_standby_names: ...description of tuple items...
```

patroni/quorum.py
```diff
@@ -0,0 +1,326 @@
...
+        :raises `QuorumError`: in case of invalid data or if invariant after transition could not be satisfied
+        """
+        if numsync < 0:
+            raise QuorumError("Sync %d < 0 of (%s)" % (numsync, sync))
```
barthisrael, line 149:
```suggestion
            raise QuorumError(f"Sync {numsync} < 0 of ({sync})")
```

patroni/quorum.py
```diff
@@ -0,0 +1,326 @@
...
+        if numsync > len(sync):
+            raise QuorumError("Sync %s > N of (%s)" % (numsync, sync))
+
+        self.numsync = numsync
+        self.sync = sync
+        self.check_invariants()
+        logger.debug('sync %s %s %s', self.leader, self.numsync, self.sync)
+        yield 'sync', self.leader, self.numsync, self.sync
+
+    def __iter__(self) -> Iterator[Tuple[str, str, int, CaseInsensitiveSet]]:
```
barthisrael, line 159: Missing docstring

patroni/quorum.py
```diff
@@ -0,0 +1,326 @@
...
+                yield from self.sync_update(max(self.sync_wanted, len(self.voters) - self.quorum), self.sync)
+            else:
+                logger.debug('Case 4: quorum is bigger than needed')
+                yield from self.quorum_update(len(self.sync) - self.numsync, self.voters)
+        else:
+            safety_margin = self.quorum + self.numsync - len(self.voters | self.sync)
+            if self.numsync == self.sync_wanted and safety_margin > 0 and self.numsync > self.numsync_confirmed:
+                yield from self.quorum_update(len(self.sync) - self.numsync, self.voters)
+
+    def __remove_gone_nodes(self) -> Iterator[Tuple[str, str, int, CaseInsensitiveSet]]:
```
barthisrael, line 216: Missing docstring

patroni/quorum.py
```diff
@@ -0,0 +1,326 @@
...
+                assert self.quorum == 0
+                numsync = self.numsync - len(to_remove)
+                sync = CaseInsensitiveSet(self.sync - to_remove)
+                voters = CaseInsensitiveSet(self.voters - to_remove)
+                sync_decrease = numsync - min(self.sync_wanted, len(sync))
+                quorum = min(sync_decrease, len(voters) - 1) if sync_decrease else 0
+                yield from self.quorum_update(quorum, voters, adjust_quorum=False)
+                yield from self.sync_update(numsync, sync)
+
+    def __add_new_nodes(self) -> Iterator[Tuple[str, str, int, CaseInsensitiveSet]]:
```
barthisrael, line 248: Missing docstring.

patroni/quorum.py
```diff
@@ -0,0 +1,326 @@
...
+                voters = CaseInsensitiveSet(self.voters | add)
+                yield from self.quorum_update(len(voters) - sync_wanted, voters)
+                to_add -= self.sync
+            if to_add:
+                voters = CaseInsensitiveSet(self.voters | to_add)
+                yield from self.quorum_update(len(voters) - sync_wanted, voters,
+                                              adjust_quorum=sync_wanted > self.numsync_confirmed)
+                yield from self.sync_update(sync_wanted, CaseInsensitiveSet(self.sync | to_add))
+
+    def __handle_replication_factor_change(self) -> Iterator[Tuple[str, str, int, CaseInsensitiveSet]]:
```
barthisrael, line 272: Missing docstring

patroni/quorum.py
```diff
@@ -0,0 +1,326 @@
...
+            yield from self.quorum_update(len(self.voters) - self.numsync, self.voters)
+        elif sync_increase < 0:
+            # Reduce replication factor
+            logger.debug("Reducing replication factor to %s", self.numsync + sync_increase)
+            if self.quorum - sync_increase < len(self.voters):
+                yield from self.quorum_update(len(self.voters) - self.numsync - sync_increase, self.voters,
+                                              adjust_quorum=self.sync_wanted > self.numsync_confirmed)
+            yield from self.sync_update(self.numsync + sync_increase, self.sync)
+
+    def _generate_transitions(self) -> Iterator[Tuple[str, str, int, CaseInsensitiveSet]]:
```
barthisrael, line 288: Missing docstring

barthisrael (commented):
More comments.

patroni/quorum.py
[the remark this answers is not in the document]
CyberDem0n: > Do we need that "+ 1"? I'm struggling to understand if that's needed and why, maybe the "<=" is enough?

Yes, we do need it:
- `numsync` is a numeric value from `synchronous_standby_names` and doesn't include the leader.
- `quorum` is a numeric value from the `/sync` key and it's meaning is - how many other nodes we need to see in order to make a decision. It's value is calculated as `len(sync) - numsync`.

Lets say out `a` is the leader name and we have `ANY 1 (b,c)` in `synchronous_standby_names`.
In the /sync key we will have: `{leader: 'a', sync: 'b,c' , quorum: 1}`

If we start performing this check we will get:
- `voters = {a,b,c}`
- `sync = {a,b,c}`
- `len(sync | voters) = 3`
- `quorum + numsync = 2`

That is, even for a normal (stable) configuration we need both, `+1` and `<=`.

When we change the topology (adding/removing/replacing nodes or changing `numsync`/`quorum` values) the rule also must be followed.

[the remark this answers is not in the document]
barthisrael: I guess I missed the leader stuff when I was checking this operation.
I think that makes more sense now, thanks for clarifying!

patroni/quorum.py
```diff
@@ -0,0 +1,407 @@
...
+    def check_invariants(self) -> None:
+        """Checks invatiant of ``synchronous_standby_names`` and ``/sync`` key in DCS.
+
+        :raises:
+            :exc:`QuorumError`: in case of broken state"""
+        voters = CaseInsensitiveSet(self.voters | CaseInsensitiveSet([self.leader]))
+        sync = CaseInsensitiveSet(self.sync | CaseInsensitiveSet([self.leader_wanted]))
+
+        # We need to verify that subset of nodes that can acknowledge a commit overlaps
+        # with any subset of nodes that can achieve quorum to promote a new leader.
```
barthisrael, line 137:
```suggestion
        # with any subset of nodes that can achieve quorum to promote a new leader.
        # ``+ 1`` is required because the leader is included in the set.
```

Maybe this helps explaining the `+ 1` when someone takes a look at this in the future?

patroni/quorum.py
```diff
@@ -0,0 +1,407 @@
...
+        if self.voters and not (len(voters | sync) <= self.quorum + self.numsync + 1):
+            raise QuorumError("Quorum and sync not guaranteed to overlap: nodes %d >= quorum %d + sync %d" %
+                              (len(voters | sync), self.quorum, self.numsync))
```
barthisrael, line 140:
```suggestion
            raise QuorumError(f"Quorum and sync not guaranteed to overlap: nodes {len(voters | sync)} > quorum "
                              f"{self.quorum} + sync {self.numsync} + 1.")
```

Use f-string and adjust the message to reflect the calculation.

patroni/quorum.py
```diff
@@ -0,0 +1,407 @@
...
+        # unstable cases, we are changing synchronous_standby_names and /sync key
+        # one after another, hence one set is allowed to be a subset of another
+        if not (voters.issubset(sync) or sync.issubset(voters)):
+            raise QuorumError("Mismatched sets: quorum only=%s sync only=%s" %
+                              (voters - sync, sync - voters))
```
barthisrael, line 145:
```suggestion
            raise QuorumError(f"Mismatched sets: quorum only={voters - sync} sync only={sync - voters}")
```

Use f-string.

patroni/quorum.py
```diff
@@ -0,0 +1,407 @@
...
+
+    def quorum_update(self, quorum: int, voters: CaseInsensitiveSet, leader: Optional[str] = None,
+                      adjust_quorum: Optional[bool] = True) -> Iterator[Transition]:
+        """Updates quorum, voters and optionally leader fields.
```
barthisrael, line 149:
```suggestion
        """Updates :attr:`quorum`, :attr:`voters` and optionally :attr:`leader` fields.
```

barthisrael (commented):
More comments 😆

patroni/quorum.py
matthbakeredb: Perhaps, it's better to pre-define a var for readability.
```suggestion
            nodes = len(voters | sync)
            raise QuorumError(f"Quorum and sync not guaranteed to overlap: nodes {nodes} > quorum "
                              f"{self.quorum} + sync {self.numsync} + 1."))
```

patroni/quorum.py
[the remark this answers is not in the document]
CyberDem0n: > not (nodes <= quorum)

The reason it is written like this because IMO it is easier to understand the meaning of ot.
We want to check the the value stored in the `quorum` field of the `/sync` key is high enough that we can surely identify whether the current node is one of the most up-to-date (could be count as synchronous).

If the rule is violated (`not`), that would mean the topology is in a bad shape.

patroni/postgresql/sync.py
```diff
@@ -273,70 +295,118 @@ def _process_replica_readiness(self, cluster: Cluster, replica_list: _ReplicaLis
...
-
-        return candidates, sync_nodes
-
-    def set_synchronous_standby_names(self, sync: Collection[str]) -> None:
+                if self._postgresql.global_config.is_quorum_commit_mode:
+                    # add nodes with nofailover tag only to get enough "active" nodes
+                    if not replica.nofailover or len(active) < sync_node_count:
+                        if replica.application_name in self._ready_replicas:
+                            numsync_confirmed += 1
+                        active.add(replica.application_name)
```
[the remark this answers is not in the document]
barthisrael: I'm not sure if I get it.

I mean, I kind of understand that you want to avoid having to consult more nodes during failover.
But wouldn't something like this `if not replica.nofailover -> add to active` increase quorum instead of reduce it?

Also, with your description it seems to deviate a bit from the behavior described in `QuorumStateResolver`:

```
        sync = voters = active

        numsync = min(sync_wanted, len(active))

        quorum = len(active) - numsync
```

Sorry if I'm missing something, just trying to connect the pieces.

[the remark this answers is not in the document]
CyberDem0n: Let's step back and think about it differently:
* n0 -- primary
* n1, n2(nofailover), n3(nofailover), n4(nofailover) -- replicas.

If we have `synchronous_standby_names = "ANY 1 (n1, n2, n3, n4)"` there is 75% chance that `n1` will not qualify itself as eligible to be the new primary because some other node from the list is ahead of it.

That is, by adding nodes tagged with `nofailover` we reduce availability guarantees.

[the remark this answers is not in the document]
barthisrael: Oh man, sorry, I'm just being silly...

I'm seeing one thing but interpreting it in a different way. I don't know why, but the `not nofailover` sounded to me like "this node is not allowed to take the primary role", when it's the opposite, "filter out nodes that are not allowed to take the primary role over".

Never mind! I got it now.

Maybe it's worth it a comment about your explanation near this piece of code. Besides that, an update to `_SyncState`'s docstring:

```
:ivar active: collection of node names that are streaming and have no restrictions to become synchronous.
```

->

```
:ivar active: collection of node names that are streaming and have no restrictions neither to become synchronous nor to be promoted.
```

What do you think?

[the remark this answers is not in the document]
barthisrael: Similarly, I think we can update the docstring of `QuorumStateResolver`:

```
    :ivar active: set of node names that are replicating from the primary (according to ``pg_stat_replication``)
                  and are eligible to be listed in ``synchronous_standby_names``.
```

->

```
    :ivar active: set of node names that are replicating from the primary (according to ``pg_stat_replication``)
                  and are eligible to be listed in ``synchronous_standby_names``, and allowed to be promoted.
```

patroni/quorum.py
```diff
@@ -0,0 +1,427 @@
...
+        if quorum < 0:
+            raise QuorumError(f'Quorum {quorum} < 0 of ({voters})')
+        if quorum > 0 and quorum >= len(voters):
+            raise QuorumError(f'Quorum {quorum} >= N of ({voters})')
+
+        old_leader = self.leader
+        if leader is not None:  # Change of leader was requested
+            self.leader = leader
+        elif self.numsync_confirmed == 0:
+            # If there are no nodes that known to caught up with the primary we want to reset quorum/votes in /sync key
```
barthisrael, line 192:
```suggestion
            # If there are no nodes that known to caught up with the primary we want to reset quorum/voters in /sync key
```

patroni/quorum.py
```diff
@@ -0,0 +1,427 @@
...
+            quorum = 0
+            voters = CaseInsensitiveSet()
+        elif adjust_quorum:
+            # It could be that the number of nodes that are known to catch up with the primary is below desired numsync.
+            # We want to increase quorum to guaranty that the sync node will be found during the leader race.
```
barthisrael, line 197:
```suggestion
            # We want to increase quorum to guarantee that the sync node will be found during the leader race.
```

patroni/quorum.py
```diff
@@ -0,0 +1,427 @@
...
+        self.check_invariants()
+        logger.debug('sync %s %s %s', self.leader, self.numsync, self.sync)
+        yield Transition('sync', self.leader, self.numsync, self.sync)
+
+    def __iter__(self) -> Iterator[Transition]:
+        """Merge two transitions of the same type to a single one.
+
+        .. note::
+            This is always safe because skipping the first transition is equivalent
+            to no one observing the intermediate state.
```
barthisrael, line 240:
```suggestion
        """Iterate over the transitions produced by :meth:`_generate_transitions`.

        .. note::
            Merge two transitions of the same type to a single one.

            This is always safe because skipping the first transition is equivalent
            to no one observing the intermediate state.
```

This line seems more like a note about the method than a description of what it is intended to do.

What about this?

patroni/postgresql/sync.py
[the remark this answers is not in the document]
CyberDem0n: The thing is that sometimes we add `nofailover` nodes to satisfy `synchronous_node_count`...

patroni/quorum.py
```diff
@@ -0,0 +1,429 @@
...
+            if cur_transition.transition_type == 'restart':
+                break
+
+    def __handle_non_steady_cases(self) -> Iterator[Transition]:
+        """Handle cases when set of transitions produces on previous run was interrupted.
+
+        :yields: transitions as :class:`Transition` objects.
+        """
+        if self.sync < self.voters:
+            logger.debug("Case 1: synchronous_standby_names subset of DCS state")
```
barthisrael, line 261: I guess we could include the variables in the debug message.

patroni/quorum.py
```diff
@@ -0,0 +1,429 @@
...
+                yield from self.quorum_update(
+                    quorum=len(self.voters) - len(remove_from_voters) - self.numsync,
+                    voters=CaseInsensitiveSet(self.voters - remove_from_voters),
+                    adjust_quorum=not (self.sync - self.active))
+            # Start syncing to nodes that are in voters and alive
+            add_to_sync = (self.voters & self.active) - self.sync
+            if add_to_sync:
+                yield from self.sync_update(self.numsync, CaseInsensitiveSet(self.sync | add_to_sync))
+        elif self.sync > self.voters:
+            logger.debug("Case 2: synchronous_standby_names superset of DCS state")
```
barthisrael, line 275: I guess we could include the variables in the debug message.

patroni/quorum.py
```diff
@@ -0,0 +1,429 @@
...
+                    numsync=min(self.numsync, len(self.sync) - len(remove_from_sync)),
+                    sync=CaseInsensitiveSet(self.sync - remove_from_sync))
+
+        # After handling these two cases voters and sync must match.
+        assert self.voters == self.sync
+
+        safety_margin = self.quorum + min(self.numsync, self.numsync_confirmed) - len(self.voters | self.sync)
+        if safety_margin > 0:  # In the middle of changing replication factor.
+            if self.numsync > self.sync_wanted:
+                logger.debug('Case 3: replication factor is bigger than needed')
```
barthisrael, line 295: I guess we could include the variables in the debug message.

patroni/quorum.py
```diff
@@ -0,0 +1,429 @@
...
+                yield from self.sync_update(max(self.sync_wanted, len(self.voters) - self.quorum), self.sync)
+            else:
+                logger.debug('Case 4: quorum is bigger than needed')
```
barthisrael, line 298: I guess we could include the variables in the debug message.

patroni/quorum.py
```diff
@@ -0,0 +1,429 @@
...
+                yield from self.quorum_update(len(voters) - self.numsync, voters)
+            # Remove from sync nodes that are dead
+            remove_from_sync = self.sync - self.voters
+            if remove_from_sync:
+                yield from self.sync_update(
+                    numsync=min(self.numsync, len(self.sync) - len(remove_from_sync)),
+                    sync=CaseInsensitiveSet(self.sync - remove_from_sync))
+
+        # After handling these two cases voters and sync must match.
+        assert self.voters == self.sync
```
barthisrael, line 290: I wonder if throwing a QuorumError instead of AssertionError would look more elegant

patroni/quorum.py
```diff
@@ -0,0 +1,429 @@
...
+        if safety_margin > 0:  # In the middle of changing replication factor.
+            if self.numsync > self.sync_wanted:
+                logger.debug('Case 3: replication factor is bigger than needed')
+                yield from self.sync_update(max(self.sync_wanted, len(self.voters) - self.quorum), self.sync)
+            else:
+                logger.debug('Case 4: quorum is bigger than needed')
+                yield from self.quorum_update(len(self.sync) - self.numsync, self.voters)
+        else:
+            safety_margin = self.quorum + self.numsync - len(self.voters | self.sync)
+            if self.numsync == self.sync_wanted and safety_margin > 0 and self.numsync > self.numsync_confirmed:
```
barthisrael, line 302: Will we ever reach this case?

I mean, I would expect the `else` where this `if` is in to be triggered only if `numsync` != `sync_wanted`.

Is there anything that I'm missing?

patroni/quorum.py
```diff
@@ -0,0 +1,429 @@
...
+                yield from self.quorum_update(len(voters) - self.numsync, voters,
+                                              adjust_quorum=not to_remove)
+            if to_remove:
+                assert self.quorum == 0
+                numsync = self.numsync - len(to_remove)
+                sync = CaseInsensitiveSet(self.sync - to_remove)
+                voters = CaseInsensitiveSet(self.voters - to_remove)
+                sync_decrease = numsync - min(self.sync_wanted, len(sync))
+                quorum = min(sync_decrease, len(voters) - 1) if sync_decrease else 0
+                yield from self.quorum_update(quorum, voters, adjust_quorum=False)
```
barthisrael, line 337: I must say I'm having some issues to consume the maths shown in this module.

I wonder if is there any way in which we can improve readability/consumption of the logic that has been implemented?

Maybe we could:

* Expand docstrings
* Expand comments
* Create auxiliary variables with a detailed description
* Describe the workflow

I don't know. I wonder if it's just me who is having bit of a hard time to digest it all?

barthisrael: (I put a localized comment here, but I guess we could look for improvements in other pieces of this module)

CyberDem0n: Yes, I know, it is extremely complex. The original PR was opened more than 5 years ago and for me it took a few attempts to get the idea.
All the complexity exists because we can't atomically update both, `/sync`  and `synchronous_standby_names`.

I would recommend you to start understanding the workflow from the `__add_new_nodes()` method:
- 'a' is a leader
- `/sync` key is empty
- `synchronous_standby_names` is empty
- we have 'b' and 'c' nodes streaming

The `QuorumStateResolver` will produce following transitions:
1. 'sync', 'a', 1, {'b', 'c'}
2. 'restart', because after updating `synchronous_standby_names` we need to wait until 'b' and 'c' will catch up with the LSN when `postgresql.conf` was updated.

That is, we got to one of **non-steady cases,** when  `/sync`  and `synchronous_standby_names` don't match.
Such state will remain until at least one of the nodes 'b' or 'c' will catch up with the LSN when `postgresql.conf` was updated (`numsync_confirmed>0`).


Interrupt might happen not only because of `restart` yielded by  `QuorumStateResolver`, but also because of other external factors like for example not being able to update the `/sync` key.
Lets consider the state:
- `a` is a leader
- `/sync` key: {leader: a, sync_standby: b,c, quorum: 1}
- `synchronous_standby_names`: `{ANY 1 (b, c)}`
- only `b` is streaming

In this case `QuorumStateResolver` will produce following transitions:
1. 'sync', 'a', 1, {'b'}
2. 'quorum', 'a', 0, {'b'}

That is, first 'c' is removed from the `synchronous_standby_names` and only after from the '/sync' key.
If update of the `/sync` key fails, we also get to a **non-steady** state.

Non-steady state might also happen when replication factor (`sync_wanted`) is increased/reduced by the administrator. The change also must be reflected by performing updates of `/sync`  and `synchronous_standby_names` in a very specific order.

Last, but not least, there could also be a combination of different changes. Like for example new nodes started streaming, but admin decided to increase/decrease replication factor.

If you look at unit-tests, you'll see that `QuorumStateResolver` can produce up to 4 transitions until the steady state is reached: https://github.com/zalando/patroni/blob/feature/quorum-commit/tests/test_quorum.py#L410-L422 
And, `QuorumStateResolver` must be able to "resume" the chain from the arbitrary state when it was interrupted.
Most of this magic happens in `__handle_non_steady_cases()` method and this is why it is com complex.

Unit-tests actually check restart `QuorumStateResolver` from all intermediate states and check that yielded transitions are matching the expectation: https://github.com/zalando/patroni/blob/feature/quorum-commit/tests/test_quorum.py#L10-L28

patroni/postgresql/sync.py
```diff
@@ -273,71 +295,118 @@ def _process_replica_readiness(self, cluster: Cluster, replica_list: _ReplicaLis
...
-                    sync_nodes.add(replica.application_name)
-            if len(candidates) >= sync_node_count:
-                break
-
-        return candidates, sync_nodes
-
-    def set_synchronous_standby_names(self, sync: Collection[str]) -> None:
-        """Constructs and sets "synchronous_standby_names" GUC value.
+                if self._postgresql.global_config.is_quorum_commit_mode:
+                    # add nodes with nofailover tag only to get enough "active" nodes
```
barthisrael, line 357:
```suggestion
                    # We do not add nodes with `nofailover` enabled because that reduces availability.
                    # We need to check LSN quorum only among nodes that are promotable because
                    # there is a chance that a non-promotable node is ahead of a promotable one.
```

Would something like this make more sense based on the thread thoughts?

patroni/postgresql/sync.py
```diff
@@ -273,71 +295,118 @@ def _process_replica_readiness(self, cluster: Cluster, replica_list: _ReplicaLis
...
+
+        return _SyncState(
+            self._ssn_data.sync_type,
+            0 if self._ssn_data.has_star else self._ssn_data.num,
+            numsync_confirmed,
+            sync_nodes,
+            active)
+
+    def set_synchronous_standby_names(self, sync: Collection[str], num: Optional[int] = None) -> None:
+        """Constructs and sets ``synchronous_standby_names`` GUC value.
```
barthisrael, line 381:
```suggestion
        """Constructs and sets ``synchronous_standby_names`` GUC value.
        
        .. note::
            standbys in ``synchronous_standby_names`` will be sorted by name.
```

tests/test_sync.py
```diff
@@ -40,34 +40,55 @@ def test_pick_sync_standby(self):
...
+    def test_current_state_quorum(self):
+        self.p._global_config = GlobalConfig({'synchronous_mode': 'quorum'})
+        cluster = Cluster(True, None, self.leader, 0, [self.me, self.other, self.leadermem], None,
+                          SyncState(0, self.me.name, self.leadermem.name, 0), None, None, None)
+
+        pg_stat_replication = [
+            {'pid': 100, 'application_name': self.leadermem.name, 'sync_state': 'quorum', 'flush_lsn': 1},
+            {'pid': 101, 'application_name': self.other.name, 'sync_state': 'quorum', 'flush_lsn': 2}]
+
+        # sync node is a bit behind of async, but we prefer it anyway
```
barthisrael, line 85: I'm not sure if I understood this comment 🤔

tests/test_quorum.py
```diff
@@ -0,0 +1,473 @@
...
+        # also check interrupted transitions
+        if len(result) > 0 and result[0][0] != 'restart' and kwargs['leader'] == result[0][1]:
+            if result[0][0] == 'sync':
+                kwargs.update(numsync=result[0][2], sync=result[0][3])
+            else:
+                kwargs.update(leader=result[0][1], quorum=result[0][2], voters=result[0][3])
+            kwargs['expected'] = expected[1:]
+            self.check_state_transitions(**kwargs)
+
+    def test_1111(self):
```
barthisrael, line 30: What is the convention on these test names that have trailing numbers?

patroni/quorum.py
barthisrael, line 1: There are 3 `assert` calls in this module. Maybe we should replace that with some kind of exception?
I mean, `assert` seems more like something we are testing than something we are running on production.

patroni/quorum.py
```diff
@@ -0,0 +1,431 @@
...
+        to_add = self.active - self.sync
+        if to_add:
+            # First get to requested replication factor
+            logger.debug("Adding nodes: %s", to_add)
+            sync_wanted = min(self.sync_wanted, len(self.sync | to_add))
+            increase_numsync_by = sync_wanted - self.numsync
+            if increase_numsync_by > 0:
+                if self.sync:
+                    add = CaseInsensitiveSet(sorted(to_add)[:increase_numsync_by])
+                    increase_numsync_by = len(add)
```
barthisrael, line 356: We need this because `to_add` maybe have less members that `increase_numsync_by` initially wants, right?

barthisrael:
I think we could give it a try as an experimental feature and see what users say about this. It is not something that would come enabled by default IAC -- only on Citus clusters from what I could see; maybe we should put some kind of warning here?

The feature is very complex, but the unit tests provide a good coverage.
The behave tests make sense to me, and from manual attempts of using the feature it seems to be working fine.

I tried understanding in detail and memorizing the quorum state machine code paths, but I'm having a really hard time on doing that. However, as I mentioned above, the unit tests are a good way to take a look at. While reviewing the PR I tried "guessing" the outputs that were written to the unit tests of `QuorumStateResolver`, and I was successful more or less half of the time. I guess that's a good indication at least.

From a developer perspective what I can say is: it's complex to understand everything that is being included, but the tests make me more confident that it's working mostly fine. I'd like to get users fingers on that, as users are brilliant on breaking things 😆 

Anyway, I would like to hear other people's thoughts.

tests/test_sync.py
CyberDem0n: I think it was a copy&paste from https://github.com/zalando/patroni/pull/2668/files#diff-7cdc82c5ea3109d2377741f06aecdcce8d49257482335f49ee25be99853621b3R40

tests/test_quorum.py
CyberDem0n: Very good question :)
These names are from the original PR: https://github.com/zalando/patroni/pull/672/files#diff-77b9ff9cf878f34d15fd5bada8be62ca3475ab49e9d6f15db4261728837be972R6

I think I'll rename them one day.

patroni/quorum.py
CyberDem0n: Its a corner case / optimization.
Adding nodes and changing the replication factor are two different operations and supposed to be done one after another, which is not always optimal.

Example:
- `/sync`: `{leader: a, sync_standby: b, quorum: 0}`
- `synchronous_standby_names`: `ANY 1 {b}`

We want to add node `c` and increase the replication factor from 1 to 2.
Non optimal way of doing that would be:
1. `/sync`: `{leader: a, sync_standby: b,c, quorum: 1}` (add node /sync key with quorum increase)
2. `synchronous_standby_names`: `ANY 1 {b,c}` (add node no ssn)
3. `synchronous_standby_names`: `ANY 2 {b,c}` (increase replication factor to requested value)
4. `/sync`: `{leader: a, sync_standby: b,c, quorum: 0}` (reduce quorum)
`2` and `3` will be merged by `__iter__` method.

With optimization (because we know that the replication factor should be increased anyway) it produces the following:
1. `synchronous_standby_names`: `ANY 2 {b,c}` (add node no ssn and increase the replication factor)
2. `/sync`: `{leader: a, sync_standby: b,c, quorum: 0}` (add node to /sync key).

Besides that there could be other corner cases. When for example we have only 2 replicas (one is already in ssn and we are adding the second one), but the requested replication factor is 3 or higher.

patroni/quorum.py
CyberDem0n: Yeah... These asserts are purely for tests.
Maybe we can also raise `QuorumError` and "handle" it in the `__iter__()` method when calling `transitions = list(self._generate_transitions())`.

barthisrael: Ah cool, yes, I think so.
This is more of a cosmetic change because we would also get exceptions on assertion errors.
I like the suggestion.

martinmarques:
@CyberDem0n what's the latest state on this PR? Need more testing/reviewing?
I know that it's very large and complex, but at the same time adds a lot of usefulness that some end users would like to have with Patroni.

CyberDem0n:
@martinmarques my current plan is to merge it for v4.0.0. When the release will happen I don't know yet.

hughcapet (approved):

PR #2671, opened 2023-05-11
Extend Postgres GUCs validator

barthisrael:
In order to make the Postgres GUCs validator in Patroni more flexible we changed it from a hard-coded static dictionary of validators to an extendable approach using YAML files. From now on Patroni will parse all validators spec found as YAML files under the `patroni/postgresql/available_parameters` directory. That will allow, for example, that Patroni accepts custom GUCs from custom Postgres builds if a vendor adds an YAML spec file to that directory.

Besides that, we also changed Patroni to use `postgres --describe-config` as an auxiliary tool. It basically serves 2 purposes:

* Do not allow settings that do not exist in a Postgres version to be passed down to Postgres;
* Avoid that a setting would not be passed down to Postgres if it is a GUC that exists, but which comes to have no validator implemented in Patroni due to an eventual oversight.

As a side change, we now check for available validators for GUCs that contain `.` in the name. It is very likely that they are from an extension, but it may be the case that a custom Postgres build contains a namespaced GUC, so we check validators before just accepting the GUC value just in case.

References: PAT-102.

### DIFF
--- a/patroni.spec
+++ b/patroni.spec
@@ -16,7 +16,10 @@ def hiddenimports():
 a = Analysis(['patroni/__main__.py'],
              pathex=[],
              binaries=None,
-             datas=[('patroni/postgresql/available_parameters', 'patroni/postgresql/available_parameters')],
+             datas=[
+                ('patroni/postgresql/available_parameters/*.yml', 'patroni/postgresql/available_parameters'),
+                ('patroni/postgresql/available_parameters/*.yaml', 'patroni/postgresql/available_parameters'),
+            ],
              hiddenimports=hiddenimports(),
              hookspath=[],
              runtime_hooks=[],
--- a/patroni.spec
+++ b/patroni.spec
@@ -16,7 +16,7 @@ def hiddenimports():
 a = Analysis(['patroni/__main__.py'],
              pathex=[],
              binaries=None,
-             datas=None,
+             datas=[('patroni/postgresql/available_parameters', 'patroni/postgresql/available_parameters')],
              hiddenimports=hiddenimports(),
              hookspath=[],
              runtime_hooks=[],
--- a/patroni/postgresql/available_parameters/0_postgres.yml
+++ b/patroni/postgresql/available_parameters/0_postgres.yml
@@ -1,0 +1,2132 @@
+parameters:
+  allow_in_place_tablespaces:
+  - type: Bool
+    version_from: 150000
+    version_till: null
+  allow_system_table_mods:
+  - type: Bool
+    version_from: 90300
+    version_till: null
+  application_name:
+  - type: String
+    version_from: 90300
+    version_till: null
+  archive_command:
+  - type: String
+    version_from: 90300
+    version_till: null
+  archive_library:
+  - type: String
+    version_from: 150000
+    version_till: null
+  archive_mode:
+  - type: Bool
+    version_from: 90300
+    version_till: 90500
+  - type: EnumBool
+    version_from: 90500
+    version_till: null
+    possible_values:
+    - always
+  archive_timeout:
+  - type: Integer
+    version_from: 90300
+    version_till: null
+    min_val: 0
+    max_val: 1073741823
+    unit: s
+  array_nulls:
+  - type: Bool
+    version_from: 90300
+    version_till: null
+  authentication_timeout:
+  - type: Integer
+    version_from: 90300
+    version_till: null
+    min_val: 1
+    max_val: 600
+    unit: s
+  autovacuum:
+  - type: Bool
+    version_from: 90300
+    version_till: null
+  autovacuum_analyze_scale_factor:
+  - type: Real
+    version_from: 90300
+    version_till: null
+    min_val: 0
+    max_val: 100
+    unit: null
+  autovacuum_analyze_threshold:
+  - type: Integer
+    version_from: 90300
+    version_till: null
+    min_val: 0
+    max_val: 2147483647
+    unit: null
+  autovacuum_freeze_max_age:
+  - type: Integer
+    version_from: 90300
+    version_till: null
+    min_val: 100000
+    max_val: 2000000000
+    unit: null
+  autovacuum_max_workers:
+  - type: Integer
+    version_from: 90300
+    version_till: 90600
+    min_val: 1
+    max_val: 8388607
+    unit: null
+  - type: Integer
+    version_from: 90600
+    version_till: null
+    min_val: 1
+    max_val: 262143
+    unit: null
+  autovacuum_multixact_freeze_max_age:
+  - type: Integer
+    version_from: 90300
+    version_till: null
+    min_val: 10000
+    max_val: 2000000000
+    unit: null
+  autovacuum_naptime:
+  - type: Integer
+    version_from: 90300
+    version_till: null
+    min_val: 1
+    max_val: 2147483
+    unit: s
+  autovacuum_vacuum_cost_delay:
+  - type: Integer
+    version_from: 90300
+    version_till: 120000
+    min_val: -1
+    max_val: 100
+    unit: ms
+  - type: Real
+    version_from: 120000
+    version_till: null
+    min_val: -1
+    max_val: 100
+    unit: ms
+  autovacuum_vacuum_cost_limit:
+  - type: Integer
+    version_from: 90300
+    version_till: null
+    min_val: -1
+    max_val: 10000
+    unit: null
+  autovacuum_vacuum_insert_scale_factor:
+  - type: Real
+    version_from: 130000
+    version_till: null
+    min_val: 0
+    max_val: 100
+    unit: null
+  autovacuum_vacuum_insert_threshold:
+  - type: Integer
+    version_from: 130000
+    version_till: null
+    min_val: -1
+    max_val: 2147483647
+    unit: null
+  autovacuum_vacuum_scale_factor:
+  - type: Real
+    version_from: 90300
+    version_till: null
+    min_val: 0
+    max_val: 100
+    unit: null
+  autovacuum_vacuum_threshold:
+  - type: Integer
+    version_from: 90300
+    version_till: null
+    min_val: 0
+    max_val: 2147483647
+    unit: null
+  autovacuum_work_mem:
+  - type: Integer
+    version_from: 90400
+    version_till: null
+    min_val: -1
+    max_val: 2147483647
+    unit: kB
+  backend_flush_after:
+  - type: Integer
+    version_from: 90600
+    version_till: null
+    min_val: 0
+    max_val: 256
+    unit: 8kB
+  backslash_quote:
+  - type: EnumBool
+    version_from: 90300
+    version_till: null
+    possible_values:
+    - safe_encoding
+  backtrace_functions:
+  - type: String
+    version_from: 130000
+    version_till: null
+  bgwriter_delay:
+  - type: Integer
+    version_from: 90300
+    version_till: null
+    min_val: 10
+    max_val: 10000
+    unit: ms
+  bgwriter_flush_after:
+  - type: Integer
+    version_from: 90600
+    version_till: null
+    min_val: 0
+    max_val: 256
+    unit: 8kB
+  bgwriter_lru_maxpages:
+  - type: Integer
+    version_from: 90300
+    version_till: 100000
+    min_val: 0
+    max_val: 1000
+    unit: null
+  - type: Integer
+    version_from: 100000
+    version_till: null
+    min_val: 0
+    max_val: 1073741823
+    unit: null
+  bgwriter_lru_multiplier:
+  - type: Real
+    version_from: 90300
+    version_till: null
+    min_val: 0
+    max_val: 10
+    unit: null
+  bonjour:
+  - type: Bool
+    version_from: 90300
+    version_till: null
+  bonjour_name:
+  - type: String
+    version_from: 90300
+    version_till: null
+  bytea_output:
+  - type: Enum
+    version_from: 90300
+    version_till: null
+    possible_values:
+    - escape
+    - hex
+  check_function_bodies:
+  - type: Bool
+    version_from: 90300
+    version_till: null
+  checkpoint_completion_target:
+  - type: Real
+    version_from: 90300
+    version_till: null
+    min_val: 0
+    max_val: 1
+    unit: null
+  checkpoint_flush_after:
+  - type: Integer
+    version_from: 90600
+    version_till: null
+    min_val: 0
+    max_val: 256
+    unit: 8kB
+  checkpoint_segments:
+  - type: Integer
+    version_from: 90300
+    version_till: 90500
+    min_val: 1
+    max_val: 2147483647
+    unit: null
+  checkpoint_timeout:
+  - type: Integer
+    version_from: 90300
+    version_till: 90600
+    min_val: 30
+    max_val: 3600
+    unit: s
+  - type: Integer
+    version_from: 90600
+    version_till: null
+    min_val: 30
+    max_val: 86400
+    unit: s
+  checkpoint_warning:
+  - type: Integer
+    version_from: 90300
+    version_till: null
+    min_val: 0
+    max_val: 2147483647
+    unit: s
+  client_connection_check_interval:
+  - type: Integer
+    version_from: 140000
+    version_till: null
+    min_val: 0
+    max_val: 2147483647
+    unit: ms
+  client_encoding:
+  - type: String
+    version_from: 90300
+    version_till: null
+  client_min_messages:
+  - type: Enum
+    version_from: 90300
+    version_till: null
+    possible_values:
+    - debug5
+    - debug4
+    - debug3
+    - debug2
+    - debug1
+    - log
+    - notice
+    - warning
+    - error
+  cluster_name:
+  - type: String
+    version_from: 90500
+    version_till: null
+  commit_delay:
+  - type: Integer
+    version_from: 90300
+    version_till: null
+    min_val: 0
+    max_val: 100000
+    unit: null
+  commit_siblings:
+  - type: Integer
+    version_from: 90300
+    version_till: null
+    min_val: 0
+    max_val: 1000
+    unit: null
+  compute_query_id:
+  - type: EnumBool
+    version_from: 140000
+    version_till: 150000
+    possible_values:
+    - auto
+  - type: EnumBool
+    version_from: 150000
+    version_till: null
+    possible_values:
+    - auto
+    - regress
+  config_file:
+  - type: String
+    version_from: 90300
+    version_till: null
+  constraint_exclusion:
+  - type: EnumBool
+    version_from: 90300
+    version_till: null
+    possible_values:
+    - partition
+  cpu_index_tuple_cost:
+  - type: Real
+    version_from: 90300
+    version_till: null
+    min_val: 0
+    max_val: 1.79769e+308
+    unit: null
+  cpu_operator_cost:
+  - type: Real
+    version_from: 90300
+    version_till: null
+    min_val: 0
+    max_val: 1.79769e+308
+    unit: null
+  cpu_tuple_cost:
+  - type: Real
+    version_from: 90300
+    version_till: null
+    min_val: 0
+    max_val: 1.79769e+308
+    unit: null
+  cursor_tuple_fraction:
+  - type: Real
+    version_from: 90300
+    version_till: null
+    min_val: 0
+    max_val: 1
+    unit: null
+  data_directory:
+  - type: String
+    version_from: 90300
+    version_till: null
+  data_sync_retry:
+  - type: Bool
+    version_from: 90400
+    version_till: null
+  DateStyle:
+  - type: String
+    version_from: 90300
+    version_till: null
+  db_user_namespace:
+  - type: Bool
+    version_from: 90300
+    version_till: null
+  deadlock_timeout:
+  - type: Integer
+    version_from: 90300
+    version_till: null
+    min_val: 1
+    max_val: 2147483647
+    unit: ms
+  debug_discard_caches:
+  - type: Integer
+    version_from: 150000
+    version_till: null
+    min_val: 0
+    max_val: 0
+    unit: null
+  debug_pretty_print:
+  - type: Bool
+    version_from: 90300
+    version_till: null
+  debug_print_parse:
+  - type: Bool
+    version_from: 90300
+    version_till: null
+  debug_print_plan:
+  - type: Bool
+    version_from: 90300
+    version_till: null
+  debug_print_rewritten:
+  - type: Bool
+    version_from: 90300
+    version_till: null
+  default_statistics_target:
+  - type: Integer
+    version_from: 90300
+    version_till: null
+    min_val: 1
+    max_val: 10000
+    unit: null
+  default_table_access_method:
+  - type: String
+    version_from: 120000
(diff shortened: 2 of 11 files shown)


barthisrael:
I noticed unit tests are failing -- they didn't fail in my environment before because I have `postgres` in my `PATH` 🤦 .

patroni/postgresql/validator.py
```diff
@@ -128,456 +135,370 @@ def transform(self, name: str, value: Optional[Any]) -> Optional[Any]:
...
+            continue
+
+        for section in ['parameters', 'recovery_parameters']:
+            section_var = getattr(module, section)
+
+            config_section = config.get(section, {})
+            for parameter in config_section.keys():
+                for exc in _load_postgres_guc_validators(section_var, config_section, parameter):
+                    logger.warning('Faced an issue while parsing a validator for parameter `%s`, from file `%s`: `%s`',
+                                   parameter, file, str(exc))
```
CyberDem0n, line 416:
```suggestion
                    logger.warning('Faced an issue while parsing a validator for parameter `%s`, from file `%s`: `%r`',
                                   parameter, file, exc)
```

patroni/postgresql/validator.py
```diff
@@ -128,456 +135,370 @@ def transform(self, name: str, value: Optional[Any]) -> Optional[Any]:
...
+    :returns: the return value may be one among:
+        * *value* transformed to the expected format for GUC *name* in Postgres *version*, if *name* is present in
+            *available_gucs* and has a validator in *validators* for the corresponding Postgres *version*; or
+        * The own *value* if *name* is present in *available_gucs* but not in *validators*; or
+        * ``None`` if *name* is not present in *available_gucs*.
+    """
+    if name in available_gucs:
+        name_validators = validators.get(name)
+        if name_validators:
+            for validator in (name_validators if isinstance(name_validators, tuple) else (name_validators,)):
```
CyberDem0n, line 448: Isn't it now always a tuple (due to https://github.com/zalando/patroni/pull/2671/files#diff-c1f5f15afbae52f3272b8520d30aea67a3820c6855b92a90ed5494f27fdd1ec3R258)?

patroni/postgresql/available_parameters/0_postgres.yml
```diff
@@ -0,0 +1,2132 @@
+parameters:
+  allow_in_place_tablespaces:
+  - type: Bool
+    version_from: 150000
+    version_till: null
```
CyberDem0n, line 5: These classes used to be namedetuples, where position of the argument in constructor plays a big role.
Now we can always use named parameters and make some of them (`version_till`, `unit`) default to `None`.
This way we can reduce a bit the size of the YAML file.

patroni/postgresql/validator.py
```diff
@@ -128,456 +135,370 @@ def transform(self, name: str, value: Optional[Any]) -> Optional[Any]:
...
+parameters = CaseInsensitiveDict()
+recovery_parameters = CaseInsensitiveDict()
+
+
+def _get_module() -> 'ModuleType':
+    """Get a reference to the current module.
+
+    :returns: the module object.
+    """
+    return sys.modules[__name__]
```
CyberDem0n, line 153: It is used only to find references to:
1. child classes of `_Transformable`
2. `parameters` + `recovery_parameters`

Both could be just a static mapping.

Or, actually https://github.com/zalando/patroni/pull/2671/files#diff-c1f5f15afbae52f3272b8520d30aea67a3820c6855b92a90ed5494f27fdd1ec3R171 could be easily built  dynamically:
```python
class _Transformable(abc.ABC):
    @classmethod
    def get_subclasses(cls):
        for subclass in cls.__subclasses__():
            for subsubclass in subclass.get_subclasses():
                yield subsubclass
            yield subclass


class ValidatorFactory:
    TYPES = {cls.__name__: cls for cls in _Transformable.get_subclasses()}
```

patroni/postgresql/validator.py
```diff
@@ -128,456 +135,370 @@ def transform(self, name: str, value: Optional[Any]) -> Optional[Any]:
...
+    )
+    yaml_files: List[str] = []
+
+    for root, _, files in os.walk(conf_dir):
+        for file in sorted(files):
+            full_path = os.path.join(root, file)
+            if file.lower().endswith(('.yml', '.yaml')):
+                yaml_files.append(full_path)
+            else:
+                logger.warning('Ignored a non-YAML file found under `available_parameters` directory: `%s`.', full_path)
```
CyberDem0n, line 400:
```suggestion
                logger.info('Ignored a non-YAML file found under `available_parameters` directory: `%s`.', full_path)
```

CyberDem0n (commented):
Besides comments:
1. flake8 complains about formatting of unit test files
2. code coverage is somehow low.

patroni/postgresql/available_parameters/0_postgres.yml
barthisrael: Oh, good point! I'll work on that in a few to reduce the size of that file.
When I initially generated this YAML I had not yet pulled the merged changes from your pyright PR, so I still had the namedtuple definitions. With that in mind at that point I just generated the YAML file with a Python script using the `_fields` attribute and the class names.

patroni/postgresql/validator.py
barthisrael: Good idea, thanks! I just submitted a commit covering that suggestion.

patroni/postgresql/validator.py
```diff
@@ -128,456 +135,370 @@ def transform(self, name: str, value: Optional[Any]) -> Optional[Any]:
...
+          - type: String
+            version_from: 90300
+            version_till: null
+        ```
+    """
+    module = _get_module()
+    conf_dir = os.path.join(
+        os.path.dirname(os.path.abspath(__file__)),
+        'available_parameters',
+    )
```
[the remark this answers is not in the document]
barthisrael: Thanks for the pointer! Let me take a look at that -- I didn't notice the `patroni.spec` in the repo 😅 .

barthisrael:
FYI I started working on increasing coverage for `patroni.postgresql.validator`. I still need to get some time to understand why all behave tests are failing with something like this -- any inputs are welcome :) :

```
Failing scenarios:
  features/basic_replication.feature:80  check rejoin of the former primary with pg_rewind

7 features passed, 1 failed, 1 skipped
36 scenarios passed, 1 failed, 5 skipped
372 steps passed, 1 failed, 62 skipped, 0 undefined
Took 8m8.589s
Error: Process completed with exit code 1.
```

CyberDem0n:
> FYI I started working on increasing coverage for patroni.postgresql.validator

It should be fairly easy, `Postgresql._get_gucs()' mocks should return some meaningful data, and the list of returned GUC's should cover everything what is referenced in Patroni.

> to understand why all behave tests are failing

Well, this is actually straightforward, just check `features/output/basic_replication/patroni_postgres0.log` and find:
``` python
  File "/home/akukushkin/git/patroni/patroni.py", line 6, in <module>
    main()
  File "/home/akukushkin/git/patroni/patroni/__main__.py", line 150, in main
    return patroni_main()
  File "/home/akukushkin/git/patroni/patroni/__main__.py", line 141, in patroni_main
    abstract_main(Patroni, schema)
  File "/home/akukushkin/git/patroni/patroni/daemon.py", line 180, in abstract_main
    controller = cls(config)
  File "/home/akukushkin/git/patroni/patroni/__main__.py", line 34, in __init__
    self.postgresql = Postgresql(self.config['postgresql'])
  File "/home/akukushkin/git/patroni/patroni/postgresql/__init__.py", line 126, in __init__
    self.config.write_postgresql_conf()
  File "/home/akukushkin/git/patroni/patroni/postgresql/config.py", line 416, in write_postgresql_conf
    self._postgresql.available_gucs)
  File "/home/akukushkin/git/patroni/patroni/postgresql/__init__.py", line 220, in available_gucs
    return self._available_gucs
AttributeError: 'Postgresql' object has no attribute '_available_gucs'. Did you mean: 'available_gucs'?
```

Basically, we need to either move declaration of `self._available_gucs` a few lines up, or maybe not cache these values at all and execute `postgres --describe-config` every time when `available_gucs` property is accessed. The second option could be very useful if something is changing between minor versions, but honestly I don't know if it ever happened that new GUC's were added in the minor release.

patroni/postgresql/validator.py
```diff
@@ -128,456 +142,365 @@ def transform(self, name: str, value: Optional[Any]) -> Optional[Any]:
...
+#          different validation rule.
+parameters = CaseInsensitiveDict()
+recovery_parameters = CaseInsensitiveDict()
+
+
+class ValidatorFactoryNoType(Exception):
+    """Raised when a validator spec misses a type."""
+
+
+class ValidatorFactorInvalidType(Exception):
```
CyberDem0n, line 159: Is it a typo?
```suggestion
class ValidatorFactoryInvalidType(Exception):
```

patroni/postgresql/validator.py
```diff
@@ -1,20 +1,34 @@
...
 
+    @classmethod
+    def get_subclasses(cls) -> Iterator[Type['_Transformable']]:
+        """Recursvely get all subclasses of :class:`_Transformable`.
+
+        :rtype: yields each subclass of :class:`_Transformable`.
+        """
+        for subclass in cls.__subclasses__():
+            for subsubclass in subclass.get_subclasses():
+                yield subsubclass
```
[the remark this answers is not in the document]
barthisrael: I did miss that too, thanks!

patroni/postgresql/validator.py
barthisrael: Yeah, good catch!

patroni.spec
```diff
@@ -16,7 +16,7 @@ def hiddenimports():
 a = Analysis(['patroni/__main__.py'],
              pathex=[],
              binaries=None,
-             datas=None,
+             datas=[('patroni/postgresql/available_parameters', 'patroni/postgresql/available_parameters')],
```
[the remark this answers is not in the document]
barthisrael: Works for me too! At a first moment I just added the full directory because the logic in `validator.py` would filter anything other than YAML files out.
But I also think it's better if we make it more explicit about the expected content of the directory in the spec/setup files.
Thanks for the pointer on the pip packaging btw.

CyberDem0n:
> Some details about that are explained in [Versioning Policy](https://www.postgresql.org/support/versioning/).

I know that. And I also know that there are cases with minor upgrade when the standby must be restarted first due to new types of WAL records being introduced. Who knows, maybe one day a new GUC will be introduced in a minor release :)
Also, my point is that `transform_postgresql_parameter_value` and `transform_recovery_parameter_value` are called very infrequently, only when `postgresql.conf` is written, and cashing seems to be unnecessary.

barthisrael:
> > Some details about that are explained in [Versioning Policy](https://www.postgresql.org/support/versioning/).
> 
> I know that. And I also know that there are cases with minor upgrade when the standby must be restarted first due to new types of WAL records being introduced. Who knows, maybe one day a new GUC will be introduced in a minor release :) Also, my point is that `transform_postgresql_parameter_value` and `transform_recovery_parameter_value` are called very infrequently, only when `postgresql.conf` is written, and cashing seems to be unnecessary.

Ah, cool, got it! Right, makes sense, I'll modify the code to behave like that :)

barthisrael:
It seems some behave tests are failing consistently as you re-run the tests but they kept failing. The problem seems related with `--describe-config` on Postgres <= 11:

- [behave (ubuntu, 3.7, etcd)](https://github.com/zalando/patroni/actions/runs/4995385324/jobs/8959963279?pr=2671): Postgres 9.6
- [behave (ubuntu, 3.7, raft)](https://github.com/zalando/patroni/actions/runs/4995385324/jobs/8959964117?pr=2671): Postgres 11
- [behave (ubuntu, 3.10, etcd)](https://github.com/zalando/patroni/actions/runs/4995385324/jobs/8959964322?pr=2671): Postgres 9.6
- [behave (ubuntu, 3.10, raft)](https://github.com/zalando/patroni/actions/runs/4995385324/jobs/8959965215?pr=2671): Postgres 11

While inspecting the `--describe-config` I found the apparent problem:

- Postgres 11:

```bash
postgres@7854224c7ba3:~$ /usr/lib/postgresql/11/bin/postgres --describe-config | awk -F'\t' '{print($1)}' | egrep "standby_mode|primary_slot_name|recovery_target_timeline"
postgres@7854224c7ba3:~$
```

- Postgres 12:

```bash
postgres@f883201129c6:~$ /usr/lib/postgresql/12/bin/postgres --describe-config | awk -F'\t' '{print($1)}' | egrep "standby_mode|primary_slot_name|recovery_target_timeline"
primary_slot_name
recovery_target_timeline
```

We will need some modification so this code works on Postgres <= 11:

```python
def _transform_parameter_value(validators: MutableMapping[str, Tuple[_Transformable, ...]],
                               version: int, name: str, value: Any,
                               available_gucs: CaseInsensitiveSet) -> Optional[Any]:
    """Validate *value* of GUC *name* for Postgres *version* using defined *validators* and *available_gucs*.

    :param validators: a dictionary of all GUCs across all Postgres versions. Each key is the name of a Postgres GUC,
        and the corresponding value is a variable length tuple of :class:`_Transformable`. Each item is a validation
        rule for the GUC for a given range of Postgres versions. Should either contain recovery GUCs or general GUCs,
        not both.
    :param version: Postgres version to validate the GUC against.
    :param name: name of the Postgres GUC.
    :param value: value of the Postgres GUC.
    :param available_gucs: a set of all GUCs available in Postgres *version*. Each item is the name of a Postgres
        GUC. Used for a couple purposes:
        * Disallow writing GUCs to ``postgresql.conf`` (or ``recovery.conf``) that does not exist in Postgres *version*;
        * Avoid ignoring GUC *name* if it does not have a validator in *validators*, but is a valid GUC in Postgres
            *version*.

    :returns: the return value may be one among:
        * *value* transformed to the expected format for GUC *name* in Postgres *version*, if *name* is present in
            *available_gucs* and has a validator in *validators* for the corresponding Postgres *version*; or
        * The own *value* if *name* is present in *available_gucs* but not in *validators*; or
        * ``None`` if *name* is not present in *available_gucs*.
    """
    if name in available_gucs:
        name_validators: Tuple[_Transformable, ...] = validators.get(name, ())
        if name_validators:
            for validator in name_validators:
                if version >= validator.version_from and\
                        (validator.version_till is None or version < validator.version_till):
                    return validator.transform(name, value)
        # Ideally we should have a validator in *validators*. However, if none is available, we will not discard a
        # setting that exists in Postgres *version*, but rather allow the value with no validation.
        return value
    logger.warning('Removing unexpected parameter=%s value=%s from the config', name, value)
```

patroni/postgresql/config.py
```diff
@@ -534,7 +535,8 @@ def _write_recovery_params(self, fd: ConfigWriter, recovery_params: CaseInsensit
                     self._passfile_mtime = mtime(self._pgpass)
                 value = self.format_dsn(value)
             else:
-                value = transform_recovery_parameter_value(self._postgresql.major_version, name, value)
+                value = transform_recovery_parameter_value(self._postgresql.major_version, name, value,
+                                                           self._postgresql.available_gucs)
```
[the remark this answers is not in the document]
barthisrael: > It fails on Postgres v11 and older, because recovery.conf wasn't converted to GUC's yet.
Cool, your review comment came just about the same time I posted the above comment 😆 . Thanks for looking too!

> Good that we run behave tests with many different Postgres version :)
Indeed, what a life saver!

barthisrael:
I can see `pyright` job is failing in `tests.yaml` workflow. This change seems to fix the issue:

```diff
diff --git a/patroni/dcs/raft.py b/patroni/dcs/raft.py
index 1011393..1a4c1ca 100644
--- a/patroni/dcs/raft.py
+++ b/patroni/dcs/raft.py
@@ -10,7 +10,7 @@ from pysyncobj.dns_resolver import globalDnsResolver
 from pysyncobj.node import TCPNode
 from pysyncobj.transport import TCPTransport, CONNECTION_STATE
 from pysyncobj.utility import TcpUtility
-from typing import Any, Callable, Collection, Dict, List, Optional, Union, TYPE_CHECKING
+from typing import Any, Callable, Collection, Dict, List, Optional, Set, Union, TYPE_CHECKING

 from . import AbstractDCS, ClusterConfig, Cluster, Failover, Leader, Member, SyncState, TimelineHistory, citus_group_re
 from ..exceptions import DCSError
@@ -113,8 +113,8 @@ class KVStoreTTL(DynMemberSyncObj):
         self.__limb: Dict[str, Dict[str, Any]] = {}
         self.set_retry_timeout(int(config.get('retry_timeout') or 10))

-        self_addr = config.get('self_addr')
-        partner_addrs = set(config.get('partner_addrs', []))
+        self_addr: Optional[str] = config.get('self_addr')
+        partner_addrs: Set[str] = set(config.get('partner_addrs', []))
         if config.get('patronictl'):
             if self_addr:
                 partner_addrs.add(self_addr)
```

However I wonder if we should push that change through this PR? To be honest I don't know why that issue with `pyright` eventually started between 2 commits in this PR.

IAC, while taking a look at that issue I also noticed that my local `pyright` runs, either through `act` or through direct invoking of `pyright`, result in a very different set of errors -- local runs throw about 830 errors, and none of them are equal to the ones observed in the GitHub Actions runs. Any idea?

patroni/postgresql/validator.py
```diff
@@ -128,456 +142,371 @@ def transform(self, name: str, value: Optional[Any]) -> Optional[Any]:
...
+    :rtype: yields any exception that is faced while parsing a validator spec into a Patroni validator object.
+    """
+    for validator_spec in config.get(parameter, []):
+        try:
+            validator = ValidatorFactory(validator_spec)
+        except (ValidatorFactoryNoType, ValidatorFactoryInvalidType, ValidatorFactoryInvalidSpec) as exc:
+            yield exc
+            continue
+
+        section[parameter] = section.get(parameter, ()) + (validator,)
```
CyberDem0n, line 254: Hmm.
Lets say we have two yaml files.
The default one, 0_postgres.yml and 1_vendor.yml.
The first one defines:
```yaml
  wal_compression:
  - type: Bool
    version_from: 90500
    version_till: 150000
  - type: EnumBool
    version_from: 150000
    possible_values:
    - pglz
    - lz4
    - zstd
```

And the second one:
```yaml
  wal_compression:
  - type: Bool
    version_from: 90500
    version_till: 150000
  - type: EnumBool
    version_from: 150000
    possible_values:
    - brotli
    - pglz
    - lz4
    - zstd
```

With this approach we will have a tuple with 4 objects: `(Bool(), EnumBool(), Bool(), EnumBool())`, what is IMO wrong.
If the same parameter is defined in a different file, it should be completely overwritten.

IMO, this function should simply receive `config[parameter]` as an argument and return a tuple.
All exceptions that are currently yielded from here should be just converted to warnings.

What do you think @barthisrael?

CyberDem0n:
1. Flake8 complains about unused import:
```
tests/test_postgresql.py:13:1: F401 'patroni.collections.CaseInsensitiveDict' imported but unused
```
2. There are still some uncovered lines in `validator.py` (exception handlers) and `__init__.py` (available_gucs()/_get_gucs())

barthisrael:
> 1. Flake8 complains about unused import:
> 
> ```
> tests/test_postgresql.py:13:1: F401 'patroni.collections.CaseInsensitiveDict' imported but unused
> ```
> 
> 2. There are still some uncovered lines in `validator.py` (exception handlers) and `__init__.py` (available_gucs()/_get_gucs())

Thanks for pointing that out!

I could swear I had run `tox` against this branch earlier, then I noticed I even missed rebasing on top of `master` to get `tox` 😆 

Just submitted a couple commits to address both issues.

barthisrael:
> Oh, my bad, posted to a wrong PR :( I got the link from here: [#2671 (comment)](https://github.com/zalando/patroni/pull/2671#issuecomment-1547997892)

Ah, cool! I earlier saw `Build 5080939381` in the title and thought it was not up-to-date because the URL contained something starting with a `6` in the path, in what seemed an ID 🤦

patroni/postgresql/validator.py
```diff
@@ -128,456 +142,369 @@ def transform(self, name: str, value: Optional[Any]) -> Optional[Any]:
...
+#          validation of the GUC across postgres versions. If a GUC validation has never changed over time, then it will
+#          have a single object in the tuple. For example, `password_encryption` used to be a boolean GUC up to Postgres
+#          10, at which point it started being an enum. In that case the value of `password_encryption` would be a tuple
+#          of 2 `_Transformable` objects (`Bool` and `Enum`, respectively), each one reprensenting a different
+#          validation rule.
+parameters = CaseInsensitiveDict()
+recovery_parameters = CaseInsensitiveDict()
+
+
+class ValidatorFactoryNoType(Exception):
```
hughcapet, line 156: Do we want to have all these exceptions as `PatroniException` subclass @CyberDem0n ?

barthisrael: Thanks for the suggestion, applying!

patroni/postgresql/validator.py
```diff
@@ -128,456 +142,369 @@ def transform(self, name: str, value: Optional[Any]) -> Optional[Any]:
...
+        * The own *value* if *name* is present in *available_gucs* but not in *validators*; or
+        * ``None`` if *name* is not present in *available_gucs*.
+    """
+    if name in available_gucs:
+        name_validators: Tuple[_Transformable, ...] = validators.get(name, ())
+        if name_validators:
+            for validator in name_validators:
+                if version >= validator.version_from and\
+                        (validator.version_till is None or version < validator.version_till):
+                    return validator.transform(name, value)
```
hughcapet, line 450:
```suggestion
        for validator in validators.get(name, ()) or ():
            if version >= validator.version_from and\
                    (validator.version_till is None or version < validator.version_till):
                return validator.transform(name, value)
```

patroni/postgresql/validator.py
```diff
@@ -128,456 +142,369 @@ def transform(self, name: str, value: Optional[Any]) -> Optional[Any]:
...
+
+    :returns: The return value may be one among
+        * The original *value* if *name* seems to be an extension GUC (contains a period '.'); or
+        * ``None`` if **name** is a recovery GUC; or
+        * *value* transformed to the expected format for GUC *name* in Postgres *version* using validators defined in
+            ``parameters``. Can also return ``None``. See :func:`_transform_parameter_value`.
+    """
     if '.' in name:
-        return value
+        if name not in parameters:
```
hughcapet, line 477: `if '.' in name and name not in parameters:`

barthisrael: Applied, thanks :)

hughcapet:
👍

patroni/postgresql/validator.py
```diff
@@ -1,20 +1,34 @@
...
 class _Transformable(abc.ABC):
 
-    def __init__(self, version_from: int, version_till: Optional[int]) -> None:
+    def __init__(self, version_from: int, version_till: Optional[int] = None) -> None:
         self.__version_from = version_from
         self.__version_till = version_till
 
+    @classmethod
+    def get_subclasses(cls) -> Iterator[Type['_Transformable']]:
+        """Recursvely get all subclasses of :class:`_Transformable`.
```
matthbakeredb, line 24:
```suggestion
        """Recursively get all subclasses of :class:`_Transformable`.
```

CyberDem0n:
:+1:

hughcapet:
👍

PR #2672, opened 2023-05-11
Fix openssl certificate generation in behave tests

hughcapet:
--addext -> -addext (doesn't work on macOS)
set keyfile permissions to 600 (to avoid "private key file has group or world access")


hughcapet:
👍

PR #2677, opened 2023-05-15
Dev docker images improvements

hughcapet:
- configurable image
- ETCD_UNSUPPORTED_ARCH env in docker-compose
- Build confd and citus for arm64 images


CyberDem0n:
:+1:

PR #2678, opened 2023-05-15
Validate-config fixes

hughcapet:
- fix --validate-config not to error out if bin_dir is an empty string in the yaml config
- mention bin_dir optionality in the docs
- validate bin_dir even if it is not in the yaml config (add optional default value for Optional config params in validator)
- make rewind user optional


CyberDem0n:
:+1:

hughcapet:
👍

PR #2682, opened 2023-05-19
Fix pyright complaints about partner_addrs

hughcapet:
Fix pyright errors raised for `raft.py` that appeared in the new version (v1.1.309)


CyberDem0n:
:+1:

hughcapet:
👍

PR #2690, opened 2023-05-24
Refactor update_leader interface

CyberDem0n:
pass reference to a last known leader object in order to avoid obtaining it from the `AbstractDCS.cluster` cache.

This change is useful for Consul, Etcd3 and Zookeeper implementations.


hughcapet:
👍

PR #2692, opened 2023-05-24
Add support for custom Postgres binary names

barthisrael:
When using a custom Postgres distribution it may be the case that the Postgres binaries are compiled with different names other than the ones used by the community Postgres distribution.

With that in mind we implemented a new set of settings for Patroni, so the user is able to override the default binary names with custom binary names through the new section postgresql.bin_name in the local configuration.

References: PAT-17.

### DIFF
--- a/patroni/validator.py
+++ b/patroni/validator.py
@@ -176,6 +176,16 @@ def is_ipv6_address(ip: str) -> bool:
     return True
 
 
+def get_bin_name(bin_name: str) -> str:
+    """Get the value of ``postgresql.bin_name[*bin_name*]`` configuration option.
+
+    :param bin_name: a key to be retrieved from ``postgresql.bin_name`` configuration.
+
+    :returns: value of ``postgresql.bin_name[*bin_name*]``, if present, otherwise *bin_name*.
+    """
+    return schema.data.get('postgresql', {}).get('bin_name', {}).get(bin_name, bin_name)
+
+
 def get_major_version(bin_dir: OptionalType[str] = None) -> str:
     """Get the major version of PostgreSQL.
 
@@ -191,9 +201,9 @@ def get_major_version(bin_dir: OptionalType[str] = None) -> str:
         * Returns `15` for PostgreSQL 15.2
     """
     if not bin_dir:
-        binary = 'postgres'
+        binary = get_bin_name('postgres')
     else:
-        binary = os.path.join(bin_dir, 'postgres')
+        binary = os.path.join(bin_dir, get_bin_name('postgres'))
     version = subprocess.check_output([binary, '--version']).decode()
     version = re.match(r'^[^\s]+ [^\s]+ (\d+)(\.(\d+))?', version)
     if TYPE_CHECKING:  # pragma: no cover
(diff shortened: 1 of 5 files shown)


patroni/validator.py
```diff
@@ -406,6 +448,29 @@ def validate(self, name: str) -> Iterator[Result]:
...
+class BinDirectory(Directory):
+    """Check if a Postgres binary directory contains the expected files.
+
+    It is a subclass of :class:`Directory` with an extended capability: translating ``BINARIES`` according to configured
+    ``postgresql.bin_name``, if any.
+
+    :cvar BINARIES: list of executable files that should exist directly under a given Postgres binary directory.
+    """
+
+    BINARIES = ["pg_ctl", "initdb", "pg_controldata", "pg_basebackup", "postgres", "pg_isready", "pg_rewind"]
```
CyberDem0n, line 460: The reason we didn't check for `pg_rewind` is that it is nice to have, but it is not required.
Also, it is not available on 9.3 and 9.4 unless someone built/installed it separately.

barthisrael: Oh, right! I forgot about this mention in the docs BTW:

> Currently supported PostgreSQL versions: 9.3 to 15.

The first thing that usually comes to my mind is the 5 years cycle of PGDG 😆 

I'll adjust that.

patroni/validator.py
```diff
@@ -176,6 +176,16 @@ def is_ipv6_address(ip: str) -> bool:
...
 
 
+def get_bin_name(bin_name: str) -> str:
+    """Get the value of ``postgresql.bin_name[*bin_name*]`` configuration option.
+
+    :param bin_name: a key to be retrieved from ``postgresql.bin_name`` configuration.
+
+    :returns: value of ``postgresql.bin_name[*bin_name*]``, if present, otherwise *bin_name*.
+    """
+    return schema.data.get('postgresql', {}).get('bin_name', {}).get(bin_name, bin_name)
```
hughcapet, line 186: why is it different to `pgcommand()`? (not `.get('bin_name', {}) or {}`)

barthisrael: I found it a bit weird too, but I did that to silence `pyright`. In `pgcommand`, if there were no `or {}`, it would raise this issue:

```
$ pyright patroni/postgresql/__init__.py
WARNING: there is a new pyright version available (v1.1.308 -> v1.1.310).
Please install the new version or set PYRIGHT_PYTHON_FORCE_VERSION to `latest`

patroni/postgresql/__init__.py
  patroni/postgresql/__init__.py:240:76 - error: "get" is not a known member of "None" (reportOptionalMemberAccess)
1 error, 0 warnings, 0 informations
```

It seems `pyright` gets a bit lost on the [get function](https://github.com/zalando/patroni/blob/master/patroni/postgresql/config.py#L1175-L1176) call of `ConfigHandler` class.

Here in `get_bin_name` it's an actual dictionary `get` call, and `pyright` runs smoothly with that.

barthisrael: We do something similar in other places of `patroni.postgresql` too, e.g.: https://github.com/zalando/patroni/blob/master/patroni/postgresql/__init__.py#L149-L151.

But I'm open to suggestions if you have a better solution 🙌

hughcapet: I asked because we actually were thinking you are trying to solve the problem of having empty `poatgresql.bin_name` value. 🙈 
Like:
```
postgresql:
  bin_name:
  bin_dir: /bin/dir
```
When after parsing it becomes `None`

Wdyt? Can it be an issue?

barthisrael: Oh, I see! I thought you were concerned about one call having `or {}` and the other not, as `get` method was supposed to fall back to a default (`{}`) already.

Yes, I did that to deal with both existing and non-existing `postgresql.bin_name` in a single line. The idea is something like "Get the value of `postgresql.bin_name.BINARY` if set, otherwise just assume the default name `BINARY` -- the latter will be the case for at least 99% of the Patroni installations 😆 ".

hughcapet: > I thought you were concerned about one call having or {} and the other not

yes, that is what I am concerned about:)

>  get method was supposed to fall back to a default ({}) already

```
% cat test.yaml 
postgresql:
  bin_name:
  bin_dir: /bin/dir
```
```
>>> with open('test.yaml') as f:
...     test = yaml.safe_load(f)
>>> test
{'postgresql': {'bin_name': None, 'bin_dir': '/bin/dir'}}
>>> test.get('postgresql', {}).get('bin_name', {}).get('bin_name', 'bin_name')
AttributeError: 'NoneType' object has no attribute 'get'
```

I mean it is covered in `pgcommand()` but here - not

barthisrael: Ohh 🤦 now I got what you mean, I think you were concerned the other way around, meaning why `pgcommand` had that, and not why `get_bin_name` was missing that.

Thanks for pointing that out, makes sense! I just submitted a commit with that fix.

hughcapet:
👍

CyberDem0n:
:+1:

PR #2695, opened 2023-05-25
Add base image build arg for alt postgres

matthbakeredb:
Allows for running behave tests with an alternative base image than the official postgres image.

Also provides a PG_USER/PG_GROUP should that be different to the default `postgres`.

References: PAT-114


CyberDem0n:
:+1:

hughcapet:
👍

PR #2697, opened 2023-05-26
Refactor daemon entrypoints

hughcapet:
- abstract_main only creates Config object using the passed configfile and instantiates the passed daemon class
- common args parser is extracted into a separate func that is called from daemons' main funcs (specific args can be added afterwards)

### DIFF
--- a/patroni/__main__.py
+++ b/patroni/__main__.py
@@ -1,11 +1,13 @@
 import logging
 import os
 import signal
+import sys
 import time
 
+from argparse import Namespace
 from typing import Any, Dict, Optional, TYPE_CHECKING
 
-from patroni.daemon import AbstractPatroniDaemon, abstract_main
+from patroni.daemon import AbstractPatroniDaemon, abstract_main, get_base_arg_parser
 
 if TYPE_CHECKING:  # pragma: no cover
     from .config import Config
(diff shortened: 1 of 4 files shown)


patroni/__main__.py
```diff
@@ -135,10 +136,25 @@ def _shutdown(self) -> None:
...
+
+    from .config import Config
+    from .daemon import get_base_arg_parser
+    parser = get_base_arg_parser()
+    parser.add_argument('--validate-config', action='store_true', help='Run config validator and exit')
+    args = parser.parse_args()
+
+    if args.validate_config:
+        from patroni.validator import schema
+        from patroni.config import ConfigParseError
```
CyberDem0n, line 149:
```suggestion
    from .daemon import get_base_arg_parser
    parser = get_base_arg_parser()
    parser.add_argument('--validate-config', action='store_true', help='Run config validator and exit')
    args = parser.parse_args()

    if args.validate_config:
        from patroni.validator import schema
        from patroni.config import Config, ConfigParseError
```

patroni/daemon.py
```diff
@@ -15,7 +16,18 @@
 
 if TYPE_CHECKING:  # pragma: no cover
     from .config import Config
-    from .validator import Schema
+
+
+def get_base_arg_parser() -> argparse.ArgumentParser:
```
CyberDem0n, line 21: Docstring?

patroni/__main__.py
```diff
@@ -133,21 +135,40 @@ def _shutdown(self) -> None:
...
+def patroni_main(configfile: str) -> None:
     from multiprocessing import freeze_support
-    from patroni.validator import schema
 
     freeze_support()
-    abstract_main(Patroni, schema)
+    abstract_main(Patroni, configfile)
+
+
+def process_arguments() -> 'Namespace':
```
CyberDem0n, line 145:
```suggestion
def process_arguments() -> Namespace:
```

PR #2698, opened 2023-05-26
Add docstrings to `patroni.__init__`

barthisrael:
References: PAT-111.

### DIFF
--- a/patroni/__init__.py
+++ b/patroni/__init__.py
@@ -19,8 +19,7 @@ def fatal(string: str, *args: Any) -> None:
 
     :param string: message to be written before exiting.
     """
-    sys.stderr.write('FATAL: ' + string.format(*args) + '\n')
-    sys.exit(1)
+    sys.exit('FATAL: ' + string.format(*args))
 
 
 def parse_version(version: str) -> Tuple[int, ...]:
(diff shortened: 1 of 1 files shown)


patroni/__init__.py
```diff
@@ -8,12 +15,41 @@
 
 
 def fatal(string: str, *args: Any) -> None:
+    """Write a fatal message to stderr and exit with code ``1``.
+
+    :param string: message to be written before exiting.
+    """
     sys.stderr.write('FATAL: ' + string.format(*args) + '\n')
     sys.exit(1)
```
CyberDem0n, line 23: Since we deprecated python 2.7 it could be changed to:
```suggestion
    sys.exit('FATAL: ' + string.format(*args))
```

patroni/__init__.py
```diff
@@ -25,6 +61,16 @@ def _parse_version(version: str) -> Iterator[int]:
 # We pass MIN_PSYCOPG2 and parse_version as arguments to simplify usage of check_psycopg from the setup.py
```
CyberDem0n, line 61: Should we move this comment to a note section in the docstring?

hughcapet:
👍

CyberDem0n:
:+1:

PR #2699, opened 2023-05-30
Generate API docs from code with sphinx autodoc

matthbakeredb:
Expanding on the addition of docstrings in code, this adds python module API docs to sphinx documentation.

A developer can preview what this might look like by running this locally:

```
tox -m docs
```

The option `-W` is added to the tox env so that warning messages are considered errors.

Adds doc generation using the above method to the test GitHub workflow to catch documentation problems on PRs.

Some docstrings have been reformatted and fixed to satisfy errors generated with the above setup.


docs/conf.py
```diff
@@ -20,10 +20,34 @@
...
+
+# -- Project information -----------------------------------------------------
+
+project = metadata['project']
+author = metadata['author'][0]
+try:
+    copyright_year = datetime.strptime(metadata['date'], '%d/%m/%Y').year
+except TypeError:
+    copyright_year = metadata['date'].year
+copyright = '{}, {}'.format(copyright_year, author)
```
CyberDem0n, line 46: It all just to parse the `docs/metadata.yaml` file, which is not used anywhere else?
Doesn't look like a good improvement comparing to having just [3 lines](https://github.com/zalando/patroni/pull/2699/files#diff-85933aa74a2d66c3e4dcdf7a9ad8397f5a7971080d34ef1108296a7c6b69e7e3L55-L57) in the `conf.py`.

docs/existing_data.rst
```diff
@@ -17,8 +17,10 @@ A Patroni cluster can be started with a data directory from a single-node Postgr
 
 .. code-block:: sql
 
-   CREATE USER $PATRONI_SUPERUSER_USERNAME WITH SUPERUSER ENCRYPTED PASSWORD '$PATRONI_SUPERUSER_PASSWORD';
-   CREATE USER $PATRONI_REPLICATION_USERNAME WITH REPLICATION ENCRYPTED PASSWORD '$PATRONI_REPLICATION_PASSWORD';
+   CREATE USER PATRONI_SUPERUSER_USERNAME WITH SUPERUSER ENCRYPTED PASSWORD 'PATRONI_SUPERUSER_PASSWORD';
+   CREATE USER PATRONI_REPLICATION_USERNAME WITH REPLICATION ENCRYPTED PASSWORD 'PATRONI_REPLICATION_PASSWORD';
```
CyberDem0n, line 21: Hmm. I would prefer to leave `$` char as it is.

CyberDem0n (commented):
I am not entirely sure that it is a good idea to publish module docs on readthedocs because they are not particularly useful for waste majority of people.
Bu, I don't mind if module docs are correctly generated.

docs/existing_data.rst
matthbakeredb: The reason I changed it is because the sphinx autodoc complains that this isn't a valid SQL command. I don't think those `$` symbols are valid and thought perhaps would confuse the user.

I can revert but I also added something to the tox.ini so that a doc build would fail on any warnings. If I revert I'll have to turn that off. That's not a massive problem, it would mean we'd lose the early warning on doc syntax problems, but we can lint docs with other things, e.g. pydocstyle. I was wondering whether to suggest adding that so we have doc linting. There is a tool called prospector which I've found to be very comprehensive with lint checks. That would at least allow us to add exceptions for specific errors we want to ignore (like this). However, I don't think it has this language inspection like the original method in sphinx.

matthbakeredb:
> I am not entirely sure that it is a good idea to publish module docs on readthedocs because they are not particularly useful for waste majority of people. Bu, I don't mind if module docs are correctly generated.

I would say that they are useful to anyone who has an interest in understanding more of the internals of Patroni. Sometimes high level documentation doesn't go quite far enough. Also, it's a good resource for developers, adding searching and structured code driven documentation. It might not be useful to most, but I also don't think it hurts to be there.

At the moment the module docs link on readthedocs is a 404. (We could try to fix the missing menu item error I get if I take the module docs out of the main menu list.) If the autodoc extension is enabled then the docs are generated in readthedocs. So, we either have module docs generated and linked via the bottom of the index page, or the extension is off and we leave it as a 404, (or perhaps even take that footer menu out of the index entirely - which would be a shame imho).

CyberDem0n:
> I also don't think it hurts to be there

Will it also be included to PDF? That may actually hurt...

docs/conf.py
matthbakeredb: No problem. I'll remove the external metadata load.

matthbakeredb:
> > I also don't think it hurts to be there
> 
> Will it also be included to PDF? That may actually hurt...

I'll take a look at that.

matthbakeredb:
> > > I also don't think it hurts to be there
> > 
> > 
> > Will it also be included to PDF? That may actually hurt...
> 
> I'll take a look at that.

How do you build the pdf at the moment? With `latexpdf` builder?

CyberDem0n:
> How do you build the pdf at the moment? With latexpdf builder?

Good question. TBH, I don't know readthedocs.org does it out of the box.

matthbakeredb:
> > How do you build the pdf at the moment? With latexpdf builder?
> 
> Good question. TBH, I don't know readthedocs.org does it out of the box.

Ah right it's from that end. I'll look into it. Had a brief look and I'm not sure if that can be done. If we can't do it in readthedocs I'll add a tag so that the module docs are only generated if a tag is passed on the command line to sphinx.

matthbakeredb:
> > > How do you build the pdf at the moment? With latexpdf builder?
> > 
> > 
> > Good question. TBH, I don't know readthedocs.org does it out of the box.
> 
> Ah right it's from that end. I'll look into it. Had a brief look and I'm not sure if that can be done. If we can't do it in readthedocs I'll add a tag so that the module docs are only generated if a tag is passed on the command line to sphinx.

I've got a way that's working locally, I want to test it with readthedocs. I've setup a separate account that will read from our fork. I think it will work based on the commands I see that are run there.

I also need to resolve the conflicts...

matthbakeredb:
@CyberDem0n Take a look at https://enterprisedbpatroni.readthedocs.io/en/latest/ which points to this branch at the moment. You can see how the module appears in the main table of contents, the module index and if you select pdf download from the bottom left, that there's no module section or index.

hughcapet:
> Take a look at https://enterprisedbpatroni.readthedocs.io/en/latest/ which points to this branch 

the new section name would benefit from being more descriptive, I think:)

also these "source" links for each method/class that forward you to the source code inside the same doc - what is the practical meaning? Seems a bit weird to me. Forwarding to the upstream gh repo (if it is possible) imho is more logical.

matthbakeredb:
> > Take a look at https://enterprisedbpatroni.readthedocs.io/en/latest/ which points to this branch
> 
> the new section name would benefit from being more descriptive, I think:)

Yes, agreed. I'll look into it.

> also these "source" links for each method/class that forward you to the source code inside the same doc - what is the practical meaning? Seems a bit weird to me. Forwarding to the upstream gh repo (if it is possible) imho is more logical.

The rtd theme supports changing those source links to "Edit in GitHub" links. I'll have a look at getting that working.

matthbakeredb:
> > Take a look at https://enterprisedbpatroni.readthedocs.io/en/latest/ which points to this branch
> 
> the new section name would benefit from being more descriptive, I think:)
> 
> also these "source" links for each method/class that forward you to the source code inside the same doc - what is the practical meaning? Seems a bit weird to me. Forwarding to the upstream gh repo (if it is possible) imho is more logical.

@hughcapet take a look now. What do you think?
https://enterprisedbpatroni.readthedocs.io/en/latest/

hughcapet:
> take a look now. What do you think?

Nice! Maybe one more suggestion: will it make sense if we have Patroni API an "contributing guidelines" in a common contributing section?

matthbakeredb:
> > take a look now. What do you think?
> 
> Nice! Maybe one more suggestion: will it make sense if we have Patroni API and "contributing guidelines" in a common contributing section?

Do you mean like a guide to writing good docstrings/type annotations? That might be good.

hughcapet:
> Do you mean like a guide to writing good docstrings/type annotations? That might be good.

No, I mean [the already existing section ](https://patroni.readthedocs.io/en/latest/CONTRIBUTING.html)which is used by contributors (like the new one being set up in this PR)?

matthbakeredb:
> > Do you mean like a guide to writing good docstrings/type annotations? That might be good.
> 
> No, I mean [the already existing section ](https://patroni.readthedocs.io/en/latest/CONTRIBUTING.html)which is used by contributors (like the new one being set up in this PR)?

Do you mean moving "Contributing Guidelines" and "Patroni API" under a new index heading called "Contributing"?

hughcapet:
> Do you mean moving "Contributing Guidelines" and "Patroni API" under a new index heading called "Contributing"?

Exactly. If you find it reasonable. Sorry for the confusion:)

matthbakeredb:
> > Do you mean moving "Contributing Guidelines" and "Patroni API" under a new index heading called "Contributing"?
> 
> Exactly. If you find it reasonable. Sorry for the confusion:)

Ah right ok. No sorry that makes sense now I re-read your first comment.

In the PDF format the API docs aren't included, so either we'd need to have 2 different sections, one with Contributing and Guidelines+API docs beneath for HTML format, and another with  just Guidelines for non-HTML format OR we always have a Contributing section with just one child link in non-HTML format. tbh I think I'd have to check whether the code I wrote to exclude the API docs from non-HTML will still work with this first.

Without considering that, I am not sure if this reference is only for people who are contributing to development, it would also be useful for people who wish to understand the internals of patroni. In my experience it's more common to have the API docs at the top level, but then that is normally associated with a library that is intended to be used in other software projects, whilst Patroni typically is not. So perhaps yes that would work better.

Let me look into the first problem...

matthbakeredb:
> > > Do you mean moving "Contributing Guidelines" and "Patroni API" under a new index heading called "Contributing"?
> > 
> > Exactly. If you find it reasonable. Sorry for the confusion:)
> Let me look into the first problem...

@hughcapet take a look https://enterprisedbpatroni.readthedocs.io/en/latest/

--
Copying commit notes here for quick reference

Create a new contributing level in docs toctree

The existing "Contributing guidelines" are moved into their own
file and a new "CONTRIBUTING" page has been created with a 
`toctree` which adds this and places API docs beneath.

As the `maxdepth` of the index `toctree` is 2 the subsections on
the Contributing guidelines page are no longer visible, they are
shown on the new CONTRIBUTING `toctree` instead.
--

I wondered whether the top level page should be DEVELOPMENT instead so that the CONTRIBUTING file doesn't need to be renamed. That might make more sense than having 2 levels of "contributing". What do you think?

hughcapet:
👍

CyberDem0n:
:+1:

matthbakeredb:
Whoop! Thanks!

PR #2702, opened 2023-05-31
Add docstrings to `patroni.collections`

barthisrael:
Note: we used `# doctest: +SKIP` in `CaseInsensitiveSet.__str__` doctest because it uses a Python `set` internally, meaning order of values is not guaranteed. The `:Example:` section is only used for demonstrating a sample behavior, it is not deterministic.

References: PAT-118.

### DIFF
--- a/patroni/collections.py
+++ b/patroni/collections.py
@@ -104,10 +104,10 @@ class CaseInsensitiveSet(MutableSet[str]):
 class CaseInsensitiveDict(MutableMapping[str, Any]):
     """A case-insensitive ``dict``-like object.
 
-    Implements all methods and operations of :class:``MutableMapping`` as well as dict's :func:``copy``.
+    Implements all methods and operations of :class:``MutableMapping`` as well as ``dict``'s :func:``copy``.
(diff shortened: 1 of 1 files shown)


hughcapet:
>  meaning order of keys is not guaranteed when querying it

as far as I understand, it is actually guaranteed for 3.7+, while we do not test on 3.6 and older 🤔

patroni/collections.py
```diff
@@ -1,73 +1,203 @@
...
+        """Assign *value* to *key* in this dict.
+
+        *key* is searched/stored case-insensitively in the dict. The corresponding value in the dict is a tuple of:
+            * original *key*;
+            * *value*.
+
+        :param key: key to be created or updated in the dict.
+        :param value: value for *key*.
+        """
         # Use the lowercase key for lookups, but store the actual key alongside the value.
```
hughcapet, line 136: do we still need this comment?

barthisrael: Good point, I don't think so.

CyberDem0n:
:+1:

hughcapet:
👍

PR #2703, opened 2023-05-31
Add docstrings to `patroni.exceptions`

barthisrael:
References: PAT-119.

### DIFF
--- a/patroni/exceptions.py
+++ b/patroni/exceptions.py
@@ -1,33 +1,55 @@
+"""Implement high-level Patroni exceptions.
+
+More specific exceptions can be found in other modules, as subclasses of any exception defined in this module.
+"""
 from typing import Any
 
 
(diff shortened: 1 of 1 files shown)


patroni/exceptions.py
```diff
@@ -1,33 +1,55 @@
...
 
 
 class PostgresException(PatroniException):
+    """Any exception related with Postgres management."""
+
     pass
 
 
 class DCSError(PatroniException):
+    """Parent class for all kind of DCS related exceptions."""
```
hughcapet, line 35: Is "*all* kind" actually correct here? @CyberDem0n

CyberDem0n: I think yes. And they are handled in the Ha.

CyberDem0n:
:+1:

hughcapet:
👍

PR #2707, opened 2023-06-01
Fix kubernetes behave tests

CyberDem0n:
Starting from 1.27 there is containerd process, which also uses k3s binary and being detected by pidof. Therefore we will search for "k3s server" string in the process list instead of just "k3s".


hughcapet:
👍

CyberDem0n:
:+1:

PR #2708, opened 2023-06-02
Add docstrings to `patroni.config`

barthisrael:
Besides adding docstrings to `patroni.config`, a few side changes
have been applied:

* Reference `config_file` property instead of internal attribute
`_config_file` in method `_load_config_file`;
* Have `_AUTH_ALLOWED_PARAMETERS[:2]` as default value of `params`
argument in method `_get_auth` instead of using
`params or _AUTH_ALLOWED_PARAMETERS[:2]` in the body;
* Use `len(PATRONI_ENV_PREFIX)` instead of a hard-coded `8` when
removing the prefix from environment variable names;
* Fix documentation of `wal_log_hints` setting. The previous docs
mentioned it was a dynamic setting that could be changed. However
it is managed by Patroni, which forces `on` value.

References: PAT-123.

### DIFF
--- a/patroni/config.py
+++ b/patroni/config.py
@@ -458,7 +458,8 @@ class Config(object):
 
             When setting parameters from :attr:`~patroni.postgresql.config.ConfigHandler.CMDLINE_OPTIONS` through
             dynamic configuration their value will be validated as per the validator defined in that very same
-            attribute entry.
+            attribute entry. If the given value cannot be validated, a warning will be logged and the default value of
+            the GUC will be used instead.
 
             Some parameters from :attr:`~patroni.postgresql.config.ConfigHandler.CMDLINE_OPTIONS` cannot be set even if
             not *is_local* configuration:
--- a/patroni/config.py
+++ b/patroni/config.py
@@ -56,8 +56,8 @@ def default_validator(conf: Dict[str, Any]) -> List[str]:
 class GlobalConfig(object):
     """A class that wraps global configuration and provides convenient methods to access/check values.
 
-    It is instantiated by calling :func:`get_global_config` method which picks either a
-    configuration from provided :class:`Cluster` object (the most up-to-date) or from the
+    It is instantiated either by calling :func:`get_global_config` or :meth:`Config.get_global_config`, which picks
+    either a configuration from provided :class:`Cluster` object (the most up-to-date) or from the
     local cache if :class:`ClusterConfig` is not initialized or doesn't have a valid config.
     """
 
--- a/patroni/config.py
+++ b/patroni/config.py
@@ -1,3 +1,4 @@
+"""Facilities related to Patroni configuration."""
 import json
 import logging
 import os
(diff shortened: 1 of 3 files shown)


patroni/config.py
```diff
@@ -272,6 +349,11 @@ def _load_cache(self) -> None:
                 logger.exception('Exception when loading file: %s', self._cache_file)
 
     def save_cache(self) -> None:
+        """Save dynamic configuration to ``patroni.dynamic.json``.
```
CyberDem0n, line 352:
```suggestion
        """Save dynamic configuration to ``patroni.dynamic.json`` under Postgres data directory.
```

patroni/config.py
```diff
@@ -272,6 +349,11 @@ def _load_cache(self) -> None:
...
+
+        .. note::
+            ``patroni.dynamic.json`` is created as a temporary file under Postgres data directory.
```
CyberDem0n, line 355:
```suggestion
            ``patroni.dynamic.jsonXXXXXX`` is created as a temporary file and than renamed to ``patroni.dynamic.json``, where ``XXXXXX`` is a random suffix.
```

patroni/config.py
```diff
@@ -34,121 +35,162 @@
...
     return []
 
 
 class GlobalConfig(object):
+    """A class that wraps global configuration and provides convenient methods to access/check values.
 
-    """A class that wrapps global configuration and provides convinient methods to access/check values.
-
-    It is instantiated by calling :func:`Config.global_config` method which picks either a
+    It is instantiated by calling :func:`get_global_config` method which picks either a
```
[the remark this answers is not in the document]
barthisrael: Hmm, I see. What about this?

```diff
diff --git a/patroni/config.py b/patroni/config.py
index 2115504..6dffc13 100644
--- a/patroni/config.py
+++ b/patroni/config.py
@@ -55,8 +55,8 @@ def default_validator(conf: Dict[str, Any]) -> List[str]:
 class GlobalConfig(object):
     """A class that wraps global configuration and provides convenient methods to access/check values.

-    It is instantiated by calling :func:`get_global_config` method which picks either a
-    configuration from provided :class:`Cluster` object (the most up-to-date) or from the
+    It is instantiated either by calling :func:`get_global_config` or :meth:`Config.get_global_config`, which picks
+    either a configuration from provided :class:`Cluster` object (the most up-to-date) or from the
     local cache if :class:`ClusterConfig` is not initialized or doesn't have a valid config.
     """
```

I ask because I can see there is a function `get_global_config`, which can be used directly or through method `Config.get_global_config`.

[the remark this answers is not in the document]
CyberDem0n: Oh... Forgot about a function :(
Your suggestion makes sense.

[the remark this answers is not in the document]
barthisrael: Just pushed a commit with the above diff.

CyberDem0n:
:+1:

matthbakeredb:
Rebased. @barthisrael you might want to check the docstring for _process_postgresql_parameters as a new log message is emitted. I reverted the suggested dict comprehension in that method which was conflicting. I hope that's OK.

patroni/config.py
```diff
@@ -202,12 +250,27 @@ class Config(object):
 
     def __init__(self, configfile: str,
                  validator: Optional[Callable[[Dict[str, Any]], List[str]]] = default_validator) -> None:
+        """Create a new instance of :class:`Config` and validate the loaded configuration using *validator*.
+
+        .. note::
+            Patroni reads the configuration from the command-line argument if it exists (given by *configfile*),
+            otherwise from the environment variable (see :cvar:`PATRONI_CONFIG_VARIABLE`).
+
+            Besides that Patroni also reads configuration values from all supported environment variables -- see
```
CyberDem0n, line 259: We need somehow emphasize that environment variables are used only when there is no `PATRONI_CONFIGURATION` environment variable or when it doesn't contain a yaml document.

patroni/config.py
```diff
@@ -456,7 +660,13 @@ def _parse_list(value: str) -> Union[List[str], None]:
                 if value:
                     ret[first][second] = value
 
-        def _parse_dict(value: str) -> Union[Dict[str, Any], None]:
+        def _parse_dict(value: str) -> Optional[Dict[str, Any]]:
+            """Parse an YAML dictioary *value* as a :class:`dict`.
```
CyberDem0n, line 664:
```suggestion
            """Parse an YAML dictionary *value* as a :class:`dict`.
```

patroni/config.py
matthbakeredb: Made a commit for this

hughcapet:
👍

CyberDem0n:
:+1:

PR #2710, opened 2023-06-09
Fix a few concurrency bugs in Citus support

CyberDem0n:
- the `_in_fligh` attribute is accessed from multiple threads and must be protected with mutex when it is changed
- allow adding tasks for `_in_fligh.group` from the `sync_pg_dist_node()` method when timeout is reached. Not doing so might result is indefinite transaction if REST API request from worker node failed.

### DIFF
--- a/patroni/postgresql/citus.py
+++ b/patroni/postgresql/citus.py
@@ -216,17 +216,18 @@ class CitusHandler(Thread):
     def process_task(self, task: PgDistNode, transaction: Optional[PgDistNode]) -> bool:
         """Updates a single row in `pg_dist_node` table, optionally in a transaction.
 
-        The transaction is started if we do a demote of the worker node
-        or before promoting the other worker if there is not transaction
-        in progress. And, the transaction it is committed when the
-        switchover/failover completed.
+        The transaction is started if we do a demote of the worker node or before promoting the other worker if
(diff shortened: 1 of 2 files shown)


patroni/postgresql/citus.py
```diff
@@ -293,10 +302,16 @@ def _add_task(self, task: PgDistNode) -> bool:
...
+            # The task.timeout is None is an indicator that it was scheduled from the sync_pg_dist_node().
+            # We don't want to override already existing task created from REST API.
+            # In case if there is already _in_flight we don't want to add a task until timeout is reached.
+            # The last one is important to protect from the case when REST API request failed.
+            # But, effectively we only need protection for the first few seconds until the member
+            # key was updated with the new state/role values after demotion has started.
+            if task.timeout is None\
+                and (i is not None and self._tasks[i].timeout is not None
+                     or self._in_flight and self._in_flight.group == task.group
+                     and self._in_flight.deadline > time.time()):
```
JelteF, line 314: This condition is pretty long and thus hard to parse for most humans. For these kinds of cases I prefer to split them up. Either in separate if statements, that simply do the same early return. Or I group them in names, e.g. 
```python
waiting_for_timeout = i is not None and self._tasks[i].timeout is not None
```

hughcapet:
👍

CyberDem0n:
:+1:

PR #2714, opened 2023-06-15
Remove brackets from parse_dsn example docstring

matthbakeredb:
From py3.11 this throws an exception

  File "/usr/lib/python3.11/ipaddress.py", line 54, in ip_address
    raise ValueError(f'{address!r} does not appear to be an IPv4 or IPv6 address')
ValueError: '%2Fhost2' does not appear to be an IPv4 or IPv6 address

### DIFF
--- a/patroni/postgresql/config.py
+++ b/patroni/postgresql/config.py
@@ -113,7 +113,7 @@ def parse_dsn(value: str) -> Optional[Dict[str, str]]:
     and sets the `sslmode`, 'gssencmode', and `channel_binding` to `prefer` if it is not present in
     the connection string. This is necessary to simplify comparison of the old and the new values.
 
-    >>> r = parse_dsn('postgresql://u%2Fse:pass@:%2f123,%2Fhost2/db%2Fsdf?application_name=mya%2Fpp&ssl=true')
+    >>> r = parse_dsn('postgresql://u%2Fse:pass@:%2f123,[::1]/db%2Fsdf?application_name=mya%2Fpp&ssl=true')
     >>> r == {'application_name': 'mya/pp', 'host': ',/host2', 'sslmode': 'require',\
               'password': 'pass', 'port': '/123', 'user': 'u/se', 'gssencmode': 'prefer', 'channel_binding': 'prefer'}
     True
--- a/patroni/postgresql/config.py
+++ b/patroni/postgresql/config.py
@@ -113,7 +113,7 @@ def parse_dsn(value: str) -> Optional[Dict[str, str]]:
     and sets the `sslmode`, 'gssencmode', and `channel_binding` to `prefer` if it is not present in
     the connection string. This is necessary to simplify comparison of the old and the new values.
 
-    >>> r = parse_dsn('postgresql://u%2Fse:pass@:%2f123,[%2Fhost2]/db%2Fsdf?application_name=mya%2Fpp&ssl=true')
+    >>> r = parse_dsn('postgresql://u%2Fse:pass@:%2f123,%2Fhost2/db%2Fsdf?application_name=mya%2Fpp&ssl=true')
     >>> r == {'application_name': 'mya/pp', 'host': ',/host2', 'sslmode': 'require',\
               'password': 'pass', 'port': '/123', 'user': 'u/se', 'gssencmode': 'prefer', 'channel_binding': 'prefer'}
     True
--- a/patroni/postgresql/config.py
+++ b/patroni/postgresql/config.py
@@ -37,15 +37,18 @@ def conninfo_uri_parse(dsn: str) -> Dict[str, str]:
     hosts: List[str] = []
     ports: List[str] = []
     for netloc in r.netloc.split('@')[-1].split(','):
-        host = None
+        # Handle IPv6
         if '[' in netloc and ']' in netloc:
-            host = netloc.split(']')[0][1:]
-        tmp = netloc.split(':', 1)
-        if host is None:
-            host = tmp[0]
-        hosts.append(host)
-        if len(tmp) == 2:
-            ports.append(tmp[1])
+            tmp = netloc.split(']')
+            hosts.append(tmp[0][1:])  # discard leading ``[``
+            if ':' in tmp[1]:
+                ports.append(tmp[1].split(':', 1)[1])
+        # Handle IPv4 or DNS name
+        else:
+            tmp = netloc.split(':', 1)
+            hosts.append(tmp[0])
+            if len(tmp) == 2:
+                ports.append(tmp[1])
     if hosts:
         ret['host'] = ','.join(hosts)
     if ports:
@@ -114,7 +117,7 @@ def parse_dsn(value: str) -> Optional[Dict[str, str]]:
     the connection string. This is necessary to simplify comparison of the old and the new values.
 
     >>> r = parse_dsn('postgresql://u%2Fse:pass@:%2f123,[::1]/db%2Fsdf?application_name=mya%2Fpp&ssl=true')
-    >>> r == {'application_name': 'mya/pp', 'host': ',/host2', 'sslmode': 'require',\
+    >>> r == {'application_name': 'mya/pp', 'host': ',::1', 'sslmode': 'require',\
               'password': 'pass', 'port': '/123', 'user': 'u/se', 'gssencmode': 'prefer', 'channel_binding': 'prefer'}
     True
     >>> r = parse_dsn(" host = 'host' dbname = db\\\\ name requiressl=1 ")


matthbakeredb:
I was wondering whether there is any code that relies on the bracketed host format it's using or if it's just a bogus docstring example? I thought the best way to demonstrate it was to create a commit. Feel free to drop if this is incorrect.

patroni/postgresql/config.py
```diff
@@ -113,7 +113,7 @@ def parse_dsn(value: str) -> Optional[Dict[str, str]]:
     and sets the `sslmode`, 'gssencmode', and `channel_binding` to `prefer` if it is not present in
     the connection string. This is necessary to simplify comparison of the old and the new values.
 
-    >>> r = parse_dsn('postgresql://u%2Fse:pass@:%2f123,[%2Fhost2]/db%2Fsdf?application_name=mya%2Fpp&ssl=true')
+    >>> r = parse_dsn('postgresql://u%2Fse:pass@:%2f123,%2Fhost2/db%2Fsdf?application_name=mya%2Fpp&ssl=true')
```
CyberDem0n, line 116: Square brackets are required in the URL for IPv6.
Up until recently python didn't check whats inside the, but now it wants to check that this is really IPv6 and nothing else (even IPv4 isn't allowed). 

```suggestion
    >>> r = parse_dsn('postgresql://u%2Fse:pass@:%2f123,[::1]/db%2Fsdf?application_name=mya%2Fpp&ssl=true')
```

CyberDem0n: @matthbakeredb JFYI

barthisrael:
The unit tests are now failing because `host` was not adjusted to be `::1`:

```python
>>> parse_dsn('postgresql://u%2Fse:pass@:%2f123,[::1]/db%2Fsdf?application_name=mya%2Fpp&ssl=true')
{'user': 'u/se', 'password': 'pass', 'host': ',::1', 'port': '/123,:1]', 'application_name': 'mya/pp', 'sslmode': 'require', 'gssencmode': 'prefer', 'channel_binding': 'prefer'}
```

But besides that `port` seems bogus. The split by `:` in `conninfo_uri_parse` seems to break an IPv6 address.

CyberDem0n:
Such a weird URL format ($PORT,$HOST) is supported by parse_dsn() function from psycopg2, which is a wrapper on `PQconninfoParse` function from libpq.
And what is interesting, it also produces strange results (psycopg2 version 2.9.6):
```python
In [1]: from psycopg2.extensions import parse_dsn

In [2]: parse_dsn('postgresql://u%2Fse:pass@:%2f123,[::1]/db%2Fsdf?application_name=mya%2Fpp&ssl=true')
Out[2]: 
{'user': 'u/se',
 'password': 'pass',
 'dbname': 'db/sdf',
 'host': ',::1',
 'port': '/123,',
 'application_name': 'mya/pp',
 'sslmode': 'require'}
```

When creating our own `parse_dsn()` I basically replicated the C code in python so that our version produces exactly the same results as the version from psycopg2.
For now we can simply adjust tests: `'host': ',::1',`

barthisrael:
> Such a weird URL format ($PORT,$HOST) is supported by parse_dsn() function from psycopg2, which is a wrapper on `PQconninfoParse` function from libpq. And what is interesting, it also produces strange results (psycopg2 version 2.9.6):
> 
> ```python
> In [1]: from psycopg2.extensions import parse_dsn
> 
> In [2]: parse_dsn('postgresql://u%2Fse:pass@:%2f123,[::1]/db%2Fsdf?application_name=mya%2Fpp&ssl=true')
> Out[2]: 
> {'user': 'u/se',
>  'password': 'pass',
>  'dbname': 'db/sdf',
>  'host': ',::1',
>  'port': '/123,',
>  'application_name': 'mya/pp',
>  'sslmode': 'require'}
> ```
> 
> When creating our own `parse_dsn()` I basically replicated the C code in python so that our version produces exactly the same results as the version from psycopg2. For now we can simply adjust tests: `'host': ',::1',`

I see, thanks for expanding! It seems you mean [this piece of C code](https://github.com/postgres/postgres/blob/efeb12ef0bfef0b5aa966a56bb4dbb1f936bda0c/src/interfaces/libpq/fe-connect.c#L6441-L6693) from Postgres.

That resolves the weird `/123` part, so we are only missing to address the `,:1]'` which is going as part of the port too in our function -- `'port': '/123,:1]'`.

barthisrael:
> > Such a weird URL format ($PORT,$HOST) is supported by parse_dsn() function from psycopg2, which is a wrapper on `PQconninfoParse` function from libpq. And what is interesting, it also produces strange results (psycopg2 version 2.9.6):
> > ```python
> > In [1]: from psycopg2.extensions import parse_dsn
> > 
> > In [2]: parse_dsn('postgresql://u%2Fse:pass@:%2f123,[::1]/db%2Fsdf?application_name=mya%2Fpp&ssl=true')
> > Out[2]: 
> > {'user': 'u/se',
> >  'password': 'pass',
> >  'dbname': 'db/sdf',
> >  'host': ',::1',
> >  'port': '/123,',
> >  'application_name': 'mya/pp',
> >  'sslmode': 'require'}
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > When creating our own `parse_dsn()` I basically replicated the C code in python so that our version produces exactly the same results as the version from psycopg2. For now we can simply adjust tests: `'host': ',::1',`
> 
> I see, thanks for expanding! It seems you mean [this piece of C code](https://github.com/postgres/postgres/blob/efeb12ef0bfef0b5aa966a56bb4dbb1f936bda0c/src/interfaces/libpq/fe-connect.c#L6441-L6693) from Postgres.
> 
> That resolves the weird `/123` part, so we are only missing to address the `,:1]'` which is going as part of the port too in our function -- `'port': '/123,:1]'`.

What about something like this?

```diff
diff --git a/patroni/postgresql/config.py b/patroni/postgresql/config.py
index 857fe6c..b40d373 100644
--- a/patroni/postgresql/config.py
+++ b/patroni/postgresql/config.py
@@ -37,15 +37,18 @@ def conninfo_uri_parse(dsn: str) -> Dict[str, str]:
     hosts: List[str] = []
     ports: List[str] = []
     for netloc in r.netloc.split('@')[-1].split(','):
-        host = None
+        # Handle IPv6
         if '[' in netloc and ']' in netloc:
-            host = netloc.split(']')[0][1:]
-        tmp = netloc.split(':', 1)
-        if host is None:
-            host = tmp[0]
-        hosts.append(host)
-        if len(tmp) == 2:
-            ports.append(tmp[1])
+            tmp = netloc.split(']')
+            hosts.append(tmp[0][1:])  # discard leading ``[``
+            if ':' in tmp[1]:
+                ports.append(tmp[1].split(':', 1)[1])
+        # Handle IPv4 or DNS name
+        else:
+            tmp = netloc.split(':', 1)
+            hosts.append(tmp[0])
+            if len(tmp) == 2:
+                ports.append(tmp[1])
     if hosts:
         ret['host'] = ','.join(hosts)
     if ports:
@@ -114,7 +117,7 @@ def parse_dsn(value: str) -> Optional[Dict[str, str]]:
     the connection string. This is necessary to simplify comparison of the old and the new values.

     >>> r = parse_dsn('postgresql://u%2Fse:pass@:%2f123,[::1]/db%2Fsdf?application_name=mya%2Fpp&ssl=true')
-    >>> r == {'application_name': 'mya/pp', 'host': ',/host2', 'sslmode': 'require',\
+    >>> r == {'application_name': 'mya/pp', 'host': ',::1', 'sslmode': 'require',\
               'password': 'pass', 'port': '/123', 'user': 'u/se', 'gssencmode': 'prefer', 'channel_binding': 'prefer'}
     True
     >>> r = parse_dsn(" host = 'host' dbname = db\\\\ name requiressl=1 ")
```

patroni/postgresql/config.py
```diff
@@ -37,15 +37,18 @@ def conninfo_uri_parse(dsn: str) -> Dict[str, str]:
...
+            tmp = netloc.split(']')
+            hosts.append(tmp[0][1:])  # discard leading ``[``
+            if ':' in tmp[1]:
+                ports.append(tmp[1].split(':', 1)[1])
+        # Handle IPv4 or DNS name
+        else:
+            tmp = netloc.split(':', 1)
+            hosts.append(tmp[0])
+            if len(tmp) == 2:
+                ports.append(tmp[1])
```
CyberDem0n, line 51:
```suggestion
        host = None
        if '[' in netloc and ']' in netloc:                                                                                            
            tmp = netloc.split(']') + ['']                                                                                             
            host = tmp[0][1:]
            netloc = ':'.join(tmp[:2])                                                                                                 
        tmp = netloc.rsplit(':', 1)                                                                                                    
        if host is None:                                                                                                               
            host = tmp[0]
        hosts.append(host)
        ports.append(tmp[1] if len(tmp) == 2 else '')                                                                                  
```

patroni/postgresql/config.py
```diff
@@ -113,8 +116,8 @@ def parse_dsn(value: str) -> Optional[Dict[str, str]]:
     and sets the `sslmode`, 'gssencmode', and `channel_binding` to `prefer` if it is not present in
     the connection string. This is necessary to simplify comparison of the old and the new values.
 
-    >>> r = parse_dsn('postgresql://u%2Fse:pass@:%2f123,[%2Fhost2]/db%2Fsdf?application_name=mya%2Fpp&ssl=true')
-    >>> r == {'application_name': 'mya/pp', 'host': ',/host2', 'sslmode': 'require',\
+    >>> r = parse_dsn('postgresql://u%2Fse:pass@:%2f123,[::1]/db%2Fsdf?application_name=mya%2Fpp&ssl=true')
+    >>> r == {'application_name': 'mya/pp', 'host': ',::1', 'sslmode': 'require',\
               'password': 'pass', 'port': '/123', 'user': 'u/se', 'gssencmode': 'prefer', 'channel_binding': 'prefer'}
```
CyberDem0n, line 121:
```suggestion
    >>> r = parse_dsn('postgresql://u%2Fse:pass@:%2f123,[::1]/db%2Fsdf?application_name=mya%2Fpp&ssl=true')
    >>> r == {'application_name': 'mya/pp', 'host': ',::1', 'sslmode': 'require',\
              'password': 'pass', 'port': '/123,', 'user': 'u/se', 'gssencmode': 'prefer', 'channel_binding': 'prefer'}
```

CyberDem0n (commented):
With my changes it is much closer to the psycopg2.extensions.parse_dsn().

Also, original version had a problem, it discarded a second port (an empty string). Line 51 fixes it.

PR #2716, opened 2023-06-20
Update GUC's validator for PG16 beta1

CyberDem0n:
Plenty of GUC's were added, and some removed.
`force_parallel_mode` was renamed to `debug_parallel_query`, but it is not worth special handling.


hughcapet:
👍

CyberDem0n:
:+1:

PR #2718, opened 2023-06-21
Compatibility with python 3.11

CyberDem0n:
now it checks that inside square brackets there is indeed IPv6.

In addition to that fix a little issue in the function itself so it returns exactly the same result as psycopg2.extensions.parse_dsn().

Close https://github.com/zalando/patroni/pull/2714


hughcapet:
👍

CyberDem0n:
:+1:

PR #2719, opened 2023-06-21
Release v3.0.3

CyberDem0n:
- update release notes
- bump Patroni version
- bump pyright version and resolve issues
- add readthedocs config file

### DIFF
--- a/docs/releases.rst
+++ b/docs/releases.rst
@@ -39,7 +39,7 @@ Version 3.0.3
 
   -  Make ``bootstrap.initdb`` optional. It is only required for new clusters, but ``patroni --validate-config`` was complaining if it was missing in the config.
   -  Don't error out when ``postgresql.bin_dir`` is empty or not set. Try to first find Postgres binaries in the default PATH instead.
-  -  Make ``postgresql.authentication.rewind`` secion optional. If it is missing Patroni is using the superuser.
+  -  Make ``postgresql.authentication.rewind`` section optional. If it is missing, Patroni is using the superuser.
 
 - Improved error reporting in ``patronictl`` (Israel)
 
--- a/docs/releases.rst
+++ b/docs/releases.rst
@@ -52,6 +52,10 @@ Version 3.0.3
 
   If the REST API call from the promoted worker to the coordinator failed during switchover it was leaving the given Citus group blocked during indefinite time.
 
+- Allow `etcd3` URL in `--dcs-url` option of `patronictl` (Israel)
+
+  If users attempted to pass a `etcd3` URL through `--dcs-url` option of `patronictl` they would face an exception.
+
 
 Version 3.0.2
 -------------
--- a/docs/releases.rst
+++ b/docs/releases.rst
@@ -3,6 +3,56 @@
 Release notes
 =============
 
+Version 3.0.3
+-------------
+
+**New features**
+
+- Compatibility with PostgreSQL 16 beta1 (Alexander Kukushkin)
+
+  Extended GUC's validator rules.
+
+- Make PostgreSQL GUC's validator extensible (Israel Barth Rubio)
+
+  Validator rules are loaded from YAML files located in ``patroni/postgresql/available_parameters/`` directory. Files are ordered in alphabetical order and applied one after another. It makes possible to have custom validators for non-standard Postgres distributions.
+
+- Added ``restapi.request_queue_size`` option (Andrey Zhidenkov)
+
+  Sets request queue size for TCP socket used by Patroni REST API. Once the queue is full, further requests get a "Connection denied" error. The default value is 5.
+
+- Call ``initdb`` directly when initializing a new cluster (Matt Baker)
+
+  Previously it was called via ``pg_ctl``, what required a special quoting of parameters passed to ``initdb``.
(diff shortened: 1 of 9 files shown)


CyberDem0n:
It was affecting only unit-tests, the actual code didn't had any problems.

barthisrael:
Should we reference [this PR](https://github.com/zalando/patroni/pull/2639) as a bugfix in the release notes?

CyberDem0n:
> Should we reference https://github.com/zalando/patroni/pull/2639 as a bugfix in the release notes?

I don't even know if someone even use this feature, it is very basic comparing to using normal config files.
But, if you like please make a suggestion in the code :)

barthisrael:
> > Should we reference #2639 as a bugfix in the release notes?
> 
> I don't even know if someone even use this feature, it is very basic comparing to using normal config files. But, if you like please make a suggestion in the code :)

Thanks for your thoughts. Yeah, I don't know either. I personally just hit that issue by coincidence while testing/writing docstrings. I don't think I would have hit that in a day-to-day basis. But I think it wouldn't hurt to add a reference just in case. Let me suggest something then

docs/releases.rst
```diff
@@ -3,6 +3,56 @@
...
+  The ``\n`` symbol was rendered as it is, instead of the actual newline symbol.
+
+
+**Bugfixes**
+
+- Fixed issue in Citus support (Alexander)
+
+  If the REST API call from the promoted worker to the coordinator failed during switchover it was leaving the given Citus group blocked during indefinite time.
+
+
```
barthisrael, line 55: - Allow `etcd3` URL in `--dcs-url` option of `patronictl`

  If users attempted to pass a `etcd3` URL through `--dcs-url` option of `patronictl` they would face an exception.

docs/releases.rst
```diff
@@ -3,6 +3,56 @@
...
+- Call ``initdb`` directly when initializing a new cluster (Matt Baker)
+
+  Previously it was called via ``pg_ctl``, what required a special quoting of parameters passed to ``initdb``.
+
+- Added before stop hook (Le Duane)
+
+  The hook could be configured via ``postgresql.before_stop`` and is executed right before ``pg_ctl stop``. The exit code doesn't impact shutdown process.
+
+- Added support for custom Postgres binary names (Israel, Polina Bungina)
```
martinmarques, line 31: We should put full names

docs/releases.rst
```diff
@@ -3,6 +3,56 @@
...
+
+  When using a custom Postgres distribution it may be the case that the Postgres binaries are compiled with different names other than the ones used by the community Postgres distribution. Custom binary names could be configured using ``postgresql.bin_name.*`` and ``PATRONI_POSTGRESQL_BIN_*`` environment variables.
+
+
+**Improvements**
+
+- Various improvements of ``patroni --validate-config`` (Polina)
```
martinmarques, line 38: Idem before

docs/releases.rst
```diff
@@ -3,6 +3,56 @@
...
+
+  -  Make ``bootstrap.initdb`` optional. It is only required for new clusters, but ``patroni --validate-config`` was complaining if it was missing in the config.
+  -  Don't error out when ``postgresql.bin_dir`` is empty or not set. Try to first find Postgres binaries in the default PATH instead.
+  -  Make ``postgresql.authentication.rewind`` section optional. If it is missing, Patroni is using the superuser.
+
+- Improved error reporting in ``patronictl`` (Israel)
```
martinmarques, line 44: Idem before

docs/releases.rst
```diff
@@ -3,6 +3,56 @@
...
+
+  The ``\n`` symbol was rendered as it is, instead of the actual newline symbol.
+
+
+**Bugfixes**
+
+- Fixed issue in Citus support (Alexander)
```
martinmarques, line 51: Idem before

docs/releases.rst
martinmarques: Use full names for the contributors

martinmarques (changes requested):
Uniform mention of committers in the Release Notes

docs/releases.rst
hughcapet: We actually only put full names at the first mention. Is that a problem?

martinmarques: The problem I see is consistency overall. I read "Polina Bungina" in one place and "Polina" later in the Release Notes and that would at least raise an eye-brow.
Is there a reason not to have the full name on every feature/bug fix? The Postgres release notes follow the same convention, which I've always found pretty nice overall (everything in the Postgres docs is very well written and organized).

hughcapet: I see and ultimately agree. I will change it in a bit, hope @CyberDem0n doesn't mind.

CyberDem0n:
Well...
If we start doing that than the whole file should be updated :)
But IMO, it shouldn't be done in this PR.

hughcapet:
> But IMO, it shouldn't be done in this PR.

Exactly. This is what I wanted to do next

CyberDem0n:
:+1:

hughcapet:
👍

PR #2722, opened 2023-06-23
REST API would not reload SSL certificate upon receiving an SIGHUP

barthisrael:
Patroni used to have basically 3 ways of loading a new certificate file in the REST API server:

* Restarting the Patroni agent
* Changing the value of `restapi.certfile` and issuing an SIGHUP
* Changing the actual certificate pointed by `restapi.certfile` and issuing an SIGHUP

The 3rd method relies on checking differences between serial numbers of the configured certificate before and after an SIGHUP.

Through [#2648](https://github.com/zalando/patroni/pull/2648) we applied a change to method `get_certificate_serial_number` that introduced a bug: REST API's certificate will not be reloaded upon an SIGHUP if the value of `restapi.certfile` remains unchanged, but with a different content in the actual certificate file.

With the code of 3.0.3 we are attempting to use the output of `load_verify_locations` to get the serial number of the certificate. However, that method only loads the information, and we would need to fetch the serial number from entries returned by `get_ca_certs` instead.

However, even with `load_verify_locations` + `get_ca_certs`, there would still be cases where the serial number would not be returned as `get_ca_certs` would return an empty list: certificates created with `CA:FALSE` -- refer to [CPython _ssl](https://github.com/python/cpython/blob/c283a0cff5603540f06d9017e484b3602cc62e7c/Modules/_ssl.c#L4618C14-L4619).

With that in mind, through this PR we are reverting to using `ssl._ssl._test_decode_cert` to get serial number from certificates. In the future we might come up with a more elegant solution, if any, but for now we will stick with the working approach that uses private functions.

References: PAT-130.

### DIFF
--- a/patroni/api.py
+++ b/patroni/api.py
@@ -1541,15 +1541,11 @@ class RestApiServer(ThreadingMixIn, HTTPServer, Thread):
         if self.__ssl_options.get('certfile'):
             import ssl
             try:
-                ctx = ssl.SSLContext(ssl.PROTOCOL_TLS_CLIENT)
-                ctx.load_verify_locations(self.__ssl_options['certfile'])
-                crts = ctx.get_ca_certs()
-                if crts:
-                    serial_number = crts[0].get('serialNumber')
-                    if TYPE_CHECKING:  # pragma: no cover
-                        assert isinstance(serial_number, str)
-                    return serial_number
-            except Exception as e:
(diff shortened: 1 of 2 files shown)


CyberDem0n:
:+1:

hughcapet:
👍

hughcapet:
👍

PR #2724, opened 2023-06-26
Prevent splitbrain from duplicate names in configuration

### LINKED ISSUE #1804: Split-brain possible on simple configuration mistake
Currently Patroni will happily launch a 2nd primary if the "member" name is by accident a duplicate of the current leader.

Sure it's an operator error...but I think would be relatively easy to avoid it by first checking the current leader's (if exists) role directly over the Patroni API.

This situation won't happen on new setups as there's seems to be a check against that (although the message "released leader key voluntarily as data dir empty and currently leader" hints that the new duplicate node still grabbed the key), but rather with a "takeover" of some old Postgres clusters where the PGDATA exists.

**To Reproduce**

1. Start the 1st node as leader.
2. Start a 2nd node that will be bootstrapped as replica.
3. Change the replica config so that the member name (`name`) matches the leader's name.
4. Restart Patroni on 2nd node

mfpekala-nutanix:
Fixes #1804 by ensuring that there is not already another node with the same name that is running in the cluster when joining. Will return an error during initialization if there is already a node with the same name. Added a behave test to ensure functionality.


features/steps/basic_replication.py
```diff
@@ -90,3 +97,8 @@ def replication_works(context, primary, replica, time_limit):
         When I add the table test_{0} to {1}
         Then table test_{0} is present on {2} after {3} seconds
     """.format(int(time()), primary, replica, time_limit))
+
+@then('there are {num_nodes:d} nodes after {time_limit:d} seconds')
+def check_node_count(context, num_nodes, time_limit):
+    assert context.pctl.check_node_count(num_nodes, time_limit),\
+        "There are not {0} nodes in the system after {1} seconds".format(num_nodes, time_limit)
```
CyberDem0n, line 104: IMO, what we really want to check that the node refuse to start with a very specific error message.
It will simplify a lot of things and make unnecessary many changes introduced in the `features/environment.py`.

patroni/__main__.py
```diff
@@ -44,9 +45,16 @@ def __init__(self, config: 'Config') -> None:
 
     def load_dynamic_configuration(self) -> None:
         from patroni.exceptions import DCSError
+
         while True:
             try:
                 cluster = self.dcs.get_cluster()
+                if cluster:
+                    same_name = [member for member in cluster.members if member.name == self.config['name']]
+                    if len(same_name) > 0 and any(member.state == 'running' for member in same_name):
```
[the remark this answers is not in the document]
CyberDem0n: Actually, the check should be more sophisticated than just `state == 'running'`.

Let's imagine that Patroni was killed with "-9" signal and init system immediately restarts it.
With this change Patroni will refuse to start until the member key expires.

We should at least also check that start for a found member `api_url` and `conn_url` are not the same.
Or, in case if they are the same try to call `api_url` and don't error out if it can't be accessed.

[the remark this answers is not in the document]
mfpekala-nutanix: This makes sense, I missed the `get_member` function, thanks for pointing out. Implemented this and moved the logic to a separate function that would be called after `requests` is initialized but before the first run

features/steps/basic_replication.py
mfpekala-nutanix: Yes this is definitely better, removed what was bloating `features/environment.py` and replaced with a check that looks for a specific error from the duplicate node

features/environment.py
```diff
@@ -45,7 +45,7 @@ def _is_accessible(self):
     def _start(self):
         """start process"""
 
-    def start(self, max_wait_limit=5):
+    def start(self, max_wait_limit=5, should_wait=True):
```
CyberDem0n, line 48: Most of the changes to the `environment.py` are not needed:
1. `should_wait` argument isn't needed and in fact even bad. Starting Patroni process takes some time and it also doesn't fail instantly due to `ensure_unique_name()`. That is, we still have to wait in the below loop. But, in addition to `self._is_accessible()` also move `assert self._has_started(), "Process {0} is not running after being started".format(self._name)` to the loop. The exception raised from the loop should be intercepted/handled in `start_duplicate_patroni()` as expected one.
2. Refactoring around `__PG_PORT`,  `__BASE_API_PORT`, and `port_offset` is also not required. Instead you can simply pass `config = {'name': name, 'restapi': {'connect_address': 'localhost:8008'}}` from `start_duplicate_patroni()`. It will work because `self.recursive_update(config, custom_config)` is called after the main config handling in __make_patroni_test_config()` is performed.
3. the new `is_running()` method isn't used.

features/environment.py
```diff
@@ -819,18 +836,20 @@ def patroni_path(self):
...
+        self._processes[name].start(max_wait_limit, should_wait=should_wait)
 
     def __getattr__(self, func):
         if func not in ['stop', 'query', 'write_label', 'read_label', 'check_role_has_changed_to',
                         'add_tag_to_config', 'get_watchdog', 'patroni_hang', 'backup']:
             raise AttributeError("PatroniPoolController instance has no attribute '{0}'".format(func))
 
         def wrapper(name, *args, **kwargs):
+            if name not in self._processes:
+                return None
```
CyberDem0n, line 852: This change isn't necessary.
```suggestion
```

features/environment.py
```diff
@@ -846,6 +865,13 @@ def create_and_set_output_directory(self, feature_name):
             shutil.rmtree(feature_dir)
         os.makedirs(feature_dir)
         self._output_dir = feature_dir
+    
+    def read_patroni_log(self, type: str, node: str) -> 'list[str]':
+        try:
+            with open(str(os.path.join(self._output_dir or '', "patroni_" + node + ".log"))) as f:
+                return [line for line in f.readlines() if line[24:24+len(type)] == type]
+        except IOError:
+            return []
```
CyberDem0n, line 874: Please move this method to the `PatroniController` class.
It could be accessed using https://github.com/zalando/patroni/pull/2724/files (`read_patroni_log` should be added to the list).

CyberDem0n: Of course, arguments should be reordered (`name` comes first).

patroni/__main__.py
```diff
@@ -6,8 +6,10 @@
 
 from argparse import Namespace
 from typing import Any, Dict, Optional, TYPE_CHECKING
+from .exceptions import ConfigParseError
 
 from patroni.daemon import AbstractPatroniDaemon, abstract_main, get_base_arg_parser
+from patroni.dcs import Member
```
CyberDem0n, line 12: Please move this import to the `ensure_unique_name()` method.

features/basic_replication.feature
```diff
@@ -83,3 +83,16 @@ Feature: basic replication
...
     When I add the table buz to postgres1
     Then table buz is present on postgres0 after 20 seconds
+
+  Scenario: check graceful rejection when two nodes have the same name
+    Given I start postgres0
+    And I shut down postgres1
+    And I shut down postgres2
+    Then postgres0 is a leader after 10 seconds
+    When I start postgres1
+    Then postgres1 role is the secondary after 10 seconds
```
CyberDem0n, line 93: It is all not required.
We just need to start a "duplicated" node with the name that is one of `postgres0`, `postgres1`, `postgres2`.

patroni/__main__.py
```diff
@@ -60,6 +65,22 @@ def load_dynamic_configuration(self) -> None:
...
 
+    def ensure_unique_name(self) -> None:
+        """ A helper function to prevent splitbrain from operator error """
+        cluster = self.dcs.get_cluster()
+        if not cluster:
+            return
+        member = cluster.get_member(self.config['name'], False)
+        if not member or type(member) != Member:
+            return
+        if member.state != 'running' or member.api_url == member.conn_url:
```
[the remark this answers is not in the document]
mfpekala-nutanix: Ah, silly mistake, my bad! I removed this, but am stuck on how to get access to the node's `state_handler` from the information on the DCS returned by `get_cluster`. Apologies if I'm missing something simple, but from the `member` returned from the DCS, I see how to inspect `api.connection_string` but not `state_handler.connection_string`. 

Unless I'm mistaken and this is supposed to be comparing the member `api.connection_string` with something from the `state_handler` on the local node? Or something entirely different?

For now I've just removed (so it always goes and tries to ping the `/liveness` endpoint if the state is `running`) but happy to add in the right fix once I figure it out.

mfpekala-nutanix:
Just pushed another round of fixes, thanks for taking the time! I was stumped on one thing (mentioned above), and there seem to be some behave tests failing on OS/DCS configs I haven't tested locally. I'll take a look at this tomorrow.

features/environment.py
```diff
@@ -100,7 +99,7 @@ def __init__(self, context, name, work_directory, output_dir, custom_config=None
             custom_config['watchdog'] = {'driver': 'testing', 'device': self.watchdog.fifo_path, 'mode': 'required'}
         else:
             self.watchdog = None
-
+
```
CyberDem0n, line 102:
```suggestion

```

patroni/__main__.py
```diff
@@ -38,12 +38,15 @@ def __init__(self, config: 'Config') -> None:
...
 
+        self.ensure_unique_name()
+
         self.tags = self.get_tags()
         self.next_run = time.time()
         self.scheduled_restart: Dict[str, Any] = {}
 
     def load_dynamic_configuration(self) -> None:
         from patroni.exceptions import DCSError
+
```
CyberDem0n, line 49: Unrelated change.

```suggestion
```

patroni/__main__.py
```diff
@@ -60,6 +63,23 @@ def load_dynamic_configuration(self) -> None:
...
+        cluster = self.dcs.get_cluster()
+        if not cluster:
+            return
+        member = cluster.get_member(self.config['name'], False)
+        if not isinstance(member, Member):
+            return
+        if member.state != 'running':
+            return
+        resp = self.request(member, endpoint="/liveness")
+        if resp.status == 200:
```
CyberDem0n, line 79:
```suggestion
        try:
            _ = self.request(member, endpoint="/liveness")
```

patroni/__main__.py
```diff
@@ -60,6 +63,23 @@ def load_dynamic_configuration(self) -> None:
...
+            logger.fatal("Can't start; there is already a node named '%s' running", self.config['name'])
+            sys.exit(1)
+
```
CyberDem0n, line 82:
```suggestion
        except Exception:
            return

```

features/environment.py
```diff
@@ -192,12 +191,21 @@ def _make_patroni_test_config(self, name, custom_config):
...
 
         host = config['postgresql']['listen'].rsplit(':', 1)[0]
         config['postgresql']['listen'] = config['postgresql']['connect_address'] = '{0}:{1}'.format(host, self.__PORT)
 
-        config['name'] = name
+        if custom_config and 'name' in custom_config:
+            config['name'] = custom_config['name']
+        else:
+            config['name'] = name
+
```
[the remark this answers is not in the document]
mfpekala-nutanix: I believe if I make this change, in order to keep the integration tests working, I'd need to make a similar change elsewhere. Here's what I was thinking:

- What we want is for two nodes to have the same `name` in their configuration. However, in `environment.py` and `PatroniPoolController` we also use `name` as the unique key in the dictionary keeping track of processes.
- So the way I got around this was by having the `name` be `dup_postgres0` in the PoolController, but using this if statement to ensure that in the configuration for the node, `name` is `postgres0`

Making this change as is would make it so that the `name` parameter of the duplicate node would be `dup_postgres0` which wouldn't cause the issue. If I changed the way the duplicate node is spawned, then it would overwrite the original `postgres0` process in `pctl._processes`. My thought was that this if statement allowed me to track the different processes with different names, but ensure that in their configurations they did indeed have duplicate names.

What are your thoughts on this? Is there something I'm missing that would allow me to cut these lines or solve the same problem in a better way?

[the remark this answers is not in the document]
CyberDem0n: https://github.com/zalando/patroni/blob/e72d3ba79e92708c378f48eec84979349dd44197/features/environment.py#L244-L245 already does everything we need.

mfpekala-nutanix:
Still seems to be running into issues with k8. Haven't been able to track down issue yet, more to come

features/environment.py
```diff
@@ -344,6 +345,12 @@ def backup(self, dest=os.path.join('data', 'basebackup')):
                         '--datadir=' + os.path.join(self._work_directory, dest),
                         '--dbname=' + self.backup_source])
 
+    def read_patroni_log(self, type):
+        try:
+            with open(str(os.path.join(self._output_dir or '', self._name + ".log"))) as f:
+                return [line for line in f.readlines() if line[24:24 + len(type)] == type]
+        except IOError:
+            return []
```
CyberDem0n, line 354: flake8 expects 2 blank lines
```suggestion


```

features/environment.py
```diff
@@ -192,12 +190,15 @@ def _make_patroni_test_config(self, name, custom_config):
...
         host = config['restapi']['listen'].rsplit(':', 1)[0]
+
         config['restapi']['listen'] = config['restapi']['connect_address'] = '{}:{}'.format(host, 8008 + int(name[-1]))
 
+
         host = config['postgresql']['listen'].rsplit(':', 1)[0]
         config['postgresql']['listen'] = config['postgresql']['connect_address'] = '{0}:{1}'.format(host, self.__PORT)
 
         config['name'] = name
+
```
CyberDem0n, line 201:
```suggestion
```

features/environment.py
```diff
@@ -192,12 +190,15 @@ def _make_patroni_test_config(self, name, custom_config):
             config['raft'] = {'data_dir': self._output_dir, 'self_addr': 'localhost:' + os.environ['RAFT_PORT']}
 
         host = config['restapi']['listen'].rsplit(':', 1)[0]
+
         config['restapi']['listen'] = config['restapi']['connect_address'] = '{}:{}'.format(host, 8008 + int(name[-1]))
 
+
```
CyberDem0n, line 196:
```suggestion
```

features/environment.py
```diff
@@ -80,7 +79,7 @@ def stop(self, kill=False, timeout=15, _=False):
 
         if self._log:
             self._log.close()
-
+
```
CyberDem0n, line 82:
```suggestion

```

features/environment.py
```diff
@@ -100,7 +99,6 @@ def __init__(self, context, name, work_directory, output_dir, custom_config=None
             custom_config['watchdog'] = {'driver': 'testing', 'device': self.watchdog.fifo_path, 'mode': 'required'}
         else:
             self.watchdog = None
```
CyberDem0n, line 101:
```suggestion
            self.watchdog = None

```

features/environment.py
```diff
@@ -192,12 +190,15 @@ def _make_patroni_test_config(self, name, custom_config):
             config['raft'] = {'data_dir': self._output_dir, 'self_addr': 'localhost:' + os.environ['RAFT_PORT']}
 
         host = config['restapi']['listen'].rsplit(':', 1)[0]
+
```
CyberDem0n, line 193:
```suggestion
```

features/steps/basic_replication.py
```diff
@@ -9,6 +9,22 @@ def start_patroni(context, name):
     return context.pctl.start(name)
 
 
+@step('I start duplicate {name:w} on port {port:d}')
+def start_duplicate_patroni(context, name, port):
+    config = {
+        "name": name,
+        "restapi": {
+            "listen": "127.0.0.1:{0}".format(port)
+        }
```
CyberDem0n, line 18: trailing whitespaces
```suggestion
        }
```

CyberDem0n (commented):
There are quite a few unrelated changes that should be reverted and some violations of flake8 formatting  rules (see suggestions).

features/environment.py
```diff
@@ -344,6 +345,12 @@ def backup(self, dest=os.path.join('data', 'basebackup')):
                         '--datadir=' + os.path.join(self._work_directory, dest),
                         '--dbname=' + self.backup_source])
 
+    def read_patroni_log(self, type):
+        try:
+            with open(str(os.path.join(self._output_dir or '', self._name + ".log"))) as f:
+                return [line for line in f.readlines() if line[24:24 + len(type)] == type]
```
CyberDem0n, line 351: It is actually log `level`, not `type`. Also, codacy [complains](https://app.codacy.com/gh/zalando/patroni/pullRequest?prid=12071813) about it because `type` is a built-in and it is better to avoid using it as a variable name.
```suggestion
    def read_patroni_log(self, level):
        try:
            with open(str(os.path.join(self._output_dir or '', self._name + ".log"))) as f:
                return [line for line in f.readlines() if line[24:24 + len(level)] == level]
```

features/environment.py
mfpekala-nutanix: Got it, thanks! One thing I've noticed is that I seem to be passing `flake8` locally (although I agree the above is clearly off). When I run both `python setup.py flake8` and `tox -e lint` they pass with no errors. I just reread the contributing docs, and am not sure what might be going wrong here. Any ideas? (Don't want to bother with these kind of fixes in the future.)

I do see where the linting errors showed up in Codacy, so at the very least I'll be sure to look there going forward.

tests/test_patroni.py
```diff
@@ -202,3 +203,38 @@ def test_check_psycopg(self):
             self.assertRaises(SystemExit, check_psycopg)
         with patch('builtins.__import__', mock_import):
             self.assertRaises(SystemExit, check_psycopg)
+
+    def test_ensure_unique_name(self):
+        # None/empty cluster implies unique name
+        with patch('patroni.dcs.AbstractDCS.get_cluster', Mock(return_value=None)):
+            self.p.ensure_unique_name()
```
hughcapet, line 210: I would prefer checking that sys.exit was not called (because when I remove line 240 here, tests still pass, this means we are likely to miss the fact non-erroneous test cases suddenly start to exit with code 1)

Or, even better, per @CyberDem0n suggestion, switch to checking wether an [exception was raised ](https://github.com/zalando/patroni/blob/master/tests/test_patroni.py#L159)/`None` was returned by `ensure_unique_name()`

tests/test_patroni.py
```diff
@@ -202,3 +203,41 @@ def test_check_psycopg(self):
...
+            without_members = empty_cluster._asdict()
+            del without_members['members']
+            # Cluster with members with different names implies unique name
+            okay_cluster = Cluster(
+                members=[Member(version=1, name="distinct", session=1, data={})],
+                **without_members
+            )
+            with patch('patroni.dcs.AbstractDCS.get_cluster', Mock(return_value=okay_cluster)):
+                self.assertIsNone(self.p.ensure_unique_name())
+
```
hughcapet, line 226: please, check flake8

hughcapet:
👍

CyberDem0n:
👍

PR #2725, opened 2023-07-04
Use full names for contributors in the release notes

martinmarques:
Until the last release, contributors' names were fully written on the first occurence during that release. This meant that if Alexander had four contributions in the release, we would use Alexander Kukushkin on the first item in the release, and on all the others just Alexander.

This could, in some cases, create some confusion. For example, if there are more than one contributor with the same first name that has more than one contribution each.

For this reason, in release 3.0.3, we used the full names of contributors on all the items from the release.

This patch is to amend the old release notes and have each entry with the full name of the contributor.

Also fix typo with 2 spaces between first name and last name in one bug fix

### DIFF
--- a/docs/releases.rst
+++ b/docs/releases.rst
@@ -74,18 +74,18 @@ Version 3.0.2
 
   It makes pager configurable via ``PAGER`` environment variable, which overrides default ``less`` and ``more``.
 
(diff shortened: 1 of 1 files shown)


hughcapet:
👍

CyberDem0n:
:+1:

Issue #2720, opened 2023-06-23
[Postger error panic] Loop recover mode db when postgres could not flush dirty data: Structure needs cleaning


cuong-max:
### What happened?

This issue was found given the following scenario:

Node3 is the leader, Node1, 2 is the replica

```
+-----------------------+--------------------+---------+---------+----+-----------+
| Member                | Host               | Role    | State   | TL | Lag in MB |
+ Cluster: postgres-cluster (7145069559047319590) --+---------+----+-----------+
| patroni-IP_1              | IP_1:8432 | Replica | running | 92 |         0 |
| patroni-IP_2              | IP_2:8432 | Replica | running | 92 |         0 |
| patroni-IP_3              | IP_3:8432 | Leader  | running | 94 |           |
+-----------------------+--------------------+---------+---------+----+-----------+
```

Node3 crash  error panic maybe relate to some querry db error (status node Leader from running => crash => running => stuck crash).

```
+-----------------------+--------------------+---------+---------+----+-----------+
| Member                | Host               | Role    | State   | TL | Lag in MB |
+ Cluster: postgres-cluster (7145069559047319590) --+---------+----+-----------+
| patroni-IP_1              | IP_1:8432 | Replica | running | 92 |         0 |
| patroni-IP_2              | IP_2:8432 | Replica | running | 92 |         0 |
| patroni-IP_3              | IP_3:8432 | Leader  | crash | |           |
+-----------------------+--------------------+---------+---------+----+-----------+
```

**Log error:**
```
2023-05-17 06:41:52.760 UTC [2514] PANIC:  could not flush dirty data: Structure needs cleaning
2023-05-17 06:41:52.986 UTC [31] LOG:  checkpointer process (PID 2514) was terminated by signal 6: Aborted
2023-05-17 06:41:52.986 UTC [31] LOG:  terminating any other active server processes
```

Patroni check node leader unhealthy => Patroni automatic recover postgres

```
    def _run_cycle(self):
        dcs_failed = False
        # ...
            if not self.state_handler.is_healthy():
                if self.is_paused():
                    self.state_handler.set_state('stopped')
                    if self.has_lock():
                        self._delete_leader()
                        return 'removed leader lock because postgres is not running'
                    # Normally we don't start Postgres in a paused state. We make an exception for the demoted primary
                    # that needs to be started after it had been stopped by demote. When there is no need to call rewind
                    # the demote code follows through to starting Postgres right away, however, in the rewind case
[...]

PR #2726, opened 2023-07-04
Start postgres not in recovery in some cases

CyberDem0n:
If we know for sure that a few moments ago postgres was still running as a primary and we still have the leader lock and can successfully update it, in this case we can safely start postgres back not in recovery. That will allow to avoid bumping timeline without a reason and hopefully improve reliability because it will address issues similar to #2720.

In addition to that remove `if self.state_handler.is_starting()` check from the `recover()` method. This branch could never be reached because the `starting` state is handled earlier in the `_run_cycle()`. Besides that remove redundant `self._crash_recovery_executed`.

P.S. now we do not cover cases when Patroni was killed along with Postgres.
Lets consider that we just started Patroni, there is no leader, and `pg_controldata` reports `Database cluster state` as `shut down`. It feels logical to use `Latest checkpoint location` and `Latest checkpoint's TimeLineID` to do a usual leader race and start directly as a primary, but it could be totally wrong. The thing is that we run `postgres --single` if standby wasn't shut down cleanly before executing `pg_rewind`. As a result `Database cluster state` transition from `in archive recovery` to `shut down`, but if such a node becomes a leader the timeline must be increased.

### DIFF
--- a/patroni/ha.py
+++ b/patroni/ha.py
@@ -442,7 +442,7 @@ class Ha(object):
         - if ``primary_start_timeout`` is 0 and this node owns the leader lock, the lock
           will be voluntarily released if there are healthy replicas to take it over.
         - if postgres was running as a ``primary`` and this node owns the leader lock, postgres is started as primary.
-        - crash recover in a single-user mode is execute in the following cases:
+        - crash recover in a single-user mode is executed in the following cases:
           - postgres was running as ``primary`` wasn't ``shut down`` cleanly and there is no leader in DCS
           - postgres was running as ``replica`` wasn't ``shut down in recovery`` (cleanly)
             and we need to run ``pg_rewind`` to join back to the cluster.
@@ -474,7 +474,7 @@ class Ha(object):
                 and self.state_handler.state == 'crashed'\
                 and self.state_handler.role in ('primary', 'master')\
                 and not self.state_handler.config.recovery_conf_exists():
-            # We know 100% that we were running as a primary a few  moments ago, therefore could just start postgres
+            # We know 100% that we were running as a primary a few moments ago, therefore could just start postgres
             msg = 'starting primary after failure'
             if self._async_executor.try_run_async(msg, self.state_handler.start,
                                                   args=(timeout, self._async_executor.critical_task)) is None:
--- a/patroni/ha.py
+++ b/patroni/ha.py
@@ -436,6 +436,22 @@ class Ha(object):
             return self._async_executor.try_run_async(msg, self._do_reinitialize, args=(self.cluster,)) or msg
 
     def recover(self) -> str:
+        """Handle the case when postgres ist't running.
+
+        Depending on the state of Patroni, DCS cluster view, and pg_controldata the following could happen:
+        - if ``primary_start_timeout`` is 0 and this node owns the leader lock, the lock
+          will be voluntarily released if there are healthy replicas to take it over.
+        - if postgres was running as a ``primary`` and this node owns the leader lock, postgres is started as primary.
+        - crash recover in a single-user mode is execute in the following cases:
+          - postgres was running as ``primary`` wasn't ``shut down`` cleanly and there is no leader in DCS
+          - postgres was running as ``replica`` wasn't ``shut down in recovery`` (cleanly)
+            and we need to run ``pg_rewind`` to join back to the cluster.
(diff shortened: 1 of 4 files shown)


barthisrael (approved):
I still need to get more used to the HA module, but I took a first look over it.

The proposed changes look good to the extent of my knowledge, and the behave tests outcome seem to confirm they are working as intended.

I'd say we are good to merge this PR.

patroni/ha.py
```diff
@@ -455,6 +451,22 @@ def recover(self) -> str:
 
         data = self.state_handler.controldata()
         logger.info('pg_controldata:\n%s\n', '\n'.join('  {0}: {1}'.format(k, v) for k, v in data.items()))
+
+        # timeout > 0 indicates that we still have the leader lock and it was just updated
+        if timeout and data.get('Database cluster state') in ('in production', 'shutting down', 'shut down')\
+                and self.state_handler.state == 'crashed'\
+                and self.state_handler.role in ('primary', 'master')\
+                and not self.state_handler.config.recovery_conf_exists():
```
matthbakeredb, line 459:
```suggestion
        if (
                timeout 
                and data.get('Database cluster state') in ('in production', 'shutting down', 'shut down')
                and self.state_handler.state == 'crashed'
                and self.state_handler.role in ('primary', 'master')
                and not self.state_handler.config.recovery_conf_exists()
        ):
```

patroni/ha.py
```diff
@@ -455,6 +451,22 @@ def recover(self) -> str:
 
         data = self.state_handler.controldata()
         logger.info('pg_controldata:\n%s\n', '\n'.join('  {0}: {1}'.format(k, v) for k, v in data.items()))
+
+        # timeout > 0 indicates that we still have the leader lock and it was just updated
```
matthbakeredb, line 455:
```suggestion
        # timeout > 0 indicates that we still have the leader lock, and it was just updated
```

patroni/ha.py
```diff
@@ -455,6 +451,22 @@ def recover(self) -> str:
...
+                and self.state_handler.role in ('primary', 'master')\
+                and not self.state_handler.config.recovery_conf_exists():
+            # We know 100% that we were running as a primary a few moments ago, therefore could just start postgres
+            msg = 'starting primary after failure'
+            if self._async_executor.try_run_async(msg, self.state_handler.start,
+                                                  args=(timeout, self._async_executor.critical_task)) is None:
+                self.recovering = True
+                return msg
+
+        # Postgres is not running and we will restart in standby mode. Watchdog is not needed until we promote.
```
matthbakeredb, line 467:
```suggestion
        # Postgres is not running, and we will restart in standby mode. Watchdog is not needed until we promote.
```

patroni/ha.py
matthbakeredb, line 455: This method is quite complex, it could do with being broken down into smaller methods.

pylint
```
patroni/ha.py:438:4: R0912: Too many branches (17/12) (too-many-branches)
```

Also, it doesn't have a docstring.

matthbakeredb (commented):
Code quality comments only. Principal seems sound.

patroni/ha.py
hughcapet: hehe, been there. It is feature, not a bug  🥲 

https://twitter.com/hugh_capet/status/1592487213274918912?s=46&t=EUhTKJup0nNtK1f7R-fEgA

patroni/ha.py
```diff
@@ -438,15 +436,29 @@ def _handle_rewind_or_reinitialize(self) -> Optional[str]:
             return self._async_executor.try_run_async(msg, self._do_reinitialize, args=(self.cluster,)) or msg
 
     def recover(self) -> str:
-        # Postgres is not running and we will restart in standby mode. Watchdog is not needed until we promote.
-        self.watchdog.disable()
-
+        """Handle the case when postgres ist't running.
```
matthbakeredb, line 439:
```suggestion
        """Handle the case when postgres isn't running.
```

matthbakeredb (approved):
:+1:

hughcapet:
> In addition to that remove if self.state_handler.is_starting() check from the recover() method. 

wasn't it done in `is_healthiest_node()`? 🤔

hughcapet:
👍

CyberDem0n:
:+1:

PR #2727, opened 2023-07-04
Set encoding on open call in setup.py

matthbakeredb:
If a host OS does not have a UTF-8 locale set the read() call is unable to read the utf-8 encoded README.rst file.

e.g. on centos:7

```
[root@d9de6028b39f src]# python3 setup.py build
  File "setup.py", line 182, in <module>
    setup_package(__version__)
  File "setup.py", line 157, in setup_package
    long_description=read('README.rst'),
  File "setup.py", line 120, in read
    return fd.read()
  File "/usr/lib64/python3.6/encodings/ascii.py", line 26, in decode
    return codecs.ascii_decode(input, self.errors)[0]
UnicodeDecodeError: 'ascii' codec can't decode byte 0xe2 in position 1004: ordinal not in range(128)
```

I think perhaps the encoding of the README.rst changed in the last release?

note: can work around this by setting LC_ALL, e.g.
```
export LC_ALL=en_US.UTF-8
```

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -8,7 +8,7 @@ You can find a version of this documentation that is searchable and also easier 
 
 There are many ways to run high availability with PostgreSQL; for a list, see the `PostgreSQL Documentation <https://wiki.postgresql.org/wiki/Replication,_Clustering,_and_Connection_Pooling>`__.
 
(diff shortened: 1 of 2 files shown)


CyberDem0n:
There are indeed some strange characters in README. Would be nice to clean them

matthbakeredb:
> There are indeed some strange characters in README. Would be nice to clean them

Yes, that would be good. Done! :-)

CyberDem0n:
:+1:

hughcapet:
👍

PR #2728, opened 2023-07-05
Reduce flakiness of citus behave tests

CyberDem0n:
- make a few attempts with timeout  when checking registered nodes
- get rid from artificial sleep
- allow check_registration() function to check secondaries

These changes are useful for Quorum based failover (#2668) and future PR that enhances Citus support by registering secondaries in `pg_dist_node`.


hughcapet:
👍

CyberDem0n:
:+1:

Issue #2142, opened 2021-12-15
Misleading error message mentioning etcdv2 when etcdv3 is unavailable


mbanck:
**Describe the bug**
The install instructions seem to indicate one should install the `python-etcd` package for `etcd3` support as well; `pip install patroni[etcd3]` also installs that:

```
Installing collected packages: python-etcd
    Running setup.py install for python-etcd ... done
Successfully installed python-etcd-0.4.5
```

However, it seems to me, that python-etcd 0.4.5 does not have support for v3, see e.g. https://github.com/jplana/python-etcd/issues/252

Also, if I try to start patroni, I get a reference to v2 even if `etcd3` is in DCS:

```
2021-12-15 16:52:11,264 DEBUG: New etcd client created for https://XXXX:443
2021-12-15 16:52:11,266 DEBUG: Converted retries value: 2 -> Retry(total=2, connect=None, read=None, redirect=0, status=None)
2021-12-15 16:52:11,266 DEBUG: Starting new HTTPS connection (1): XXX:443
2021-12-15 16:52:11,278 DEBUG: https://XXX:443 "GET /version HTTP/1.1" 502 150
2021-12-15 16:53:16,535 ERROR: Failed to get list of machines from https://XXX:443/v2: <Unknown error: '<html>
<head><title>502 Bad Gateway</title></head>
[...]
```
I checked that manually running `ETCDCTL_API=3 etcdctl --endpoints [...] get members` works.

Is this a problem on my side, is there some code needed, or is this just a documentation issue?

**Environment**
- Patroni version: 2.1.2
- PostgreSQL version: n/a
- DCS (and its version): etcd 3.4.3

**Patroni configuration file**
```
etcd3:
  protocol: "https"
  url: "https://XXX:443"
```

CyberDem0n:
> However, it seems to me, that python-etcd 0.4.5 does not have support for v3,

Patroni implements v3 on its own (via JSON gRPC gateways). But, it still relies on some useful methods from python-etcd.

> I checked that manually running ETCDCTL_API=3 etcdctl --endpoints [...] get members works.

Do you have JSON gRPC gateway enabled? Ref: https://github.com/zalando/patroni/issues/1939

mbanck:
I can ask (the etcd cluster is not operated by me), but I assumed that if `etcdctl` can get the members fine, it should work?

CyberDem0n:
> but I assumed that if etcdctl can get the members fine, it should work?

[...]

PR #2729, opened 2023-07-05
Make sure the version_prefix for etcd3 is set to /v3beta

CyberDem0n:
Setting it before calling the parent constructor didn't really work because it is being overwritten in the `etcd.Client.__init__()`.

The only viable way of doing it is passing a custom value to the parent class.

Close https://github.com/zalando/patroni/issues/2142

### DIFF
--- a/patroni/dcs/etcd.py
+++ b/patroni/dcs/etcd.py
@@ -444,7 +444,7 @@ class EtcdClient(AbstractEtcdClientWithFailover):
     ERROR_CLS = EtcdError
 
     def __init__(self, config: Dict[str, Any], dns_resolver: DnsCachingResolver, cache_ttl: int = 300) -> None:
-        super(EtcdClient, self).__init__({'version_prefix': None, **config}, dns_resolver, cache_ttl)
+        super(EtcdClient, self).__init__({**config, 'version_prefix': None}, dns_resolver, cache_ttl)
 
     def __del__(self) -> None:
         try:
--- a/patroni/dcs/etcd.py
+++ b/patroni/dcs/etcd.py
@@ -443,6 +443,9 @@ class EtcdClient(AbstractEtcdClientWithFailover):
 
     ERROR_CLS = EtcdError
 
+    def __init__(self, config: Dict[str, Any], dns_resolver: DnsCachingResolver, cache_ttl: int = 300) -> None:
+        super(EtcdClient, self).__init__({'version_prefix': None, **config}, dns_resolver, cache_ttl)
+
     def __del__(self) -> None:
         try:
             self.http.clear()
--- a/patroni/dcs/etcd.py
+++ b/patroni/dcs/etcd.py
@@ -99,7 +99,7 @@ class AbstractEtcdClientWithFailover(abc.ABC, etcd.Client):
         self._dns_resolver = dns_resolver
         self.set_machines_cache_ttl(cache_ttl)
         self._machines_cache_updated = 0
-        kwargs = {p: config.get(p) for p in ('host', 'port', 'protocol', 'use_proxies',
+        kwargs = {p: config.get(p) for p in ('host', 'port', 'protocol', 'use_proxies', 'version_prefix',
                                              'username', 'password', 'cert', 'ca_cert') if config.get(p)}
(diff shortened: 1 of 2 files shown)


matthbakeredb:
This seems like a bugfix as an aside from the linked issue which talks about a misleading error message. I presume this alleviates the need for an error message? Should there still be message handling where the error was exhibited?

Should the exposure of the configuration of `version_prefix` be protected so a user can't set it? I guess we have that via validation...

CyberDem0n:
As a result of this change the error message will look like:
```
Failed to get list of machines from https://XXX:443/v3beta: <Unknown error: '<html>
<head><title>502 Bad Gateway</title></head>
...
```

I.e., `/v3beta` instead of `/v2`.

Unfortunately we can't do better than that, because this prefix is [specific to an Etcd version](https://etcd.io/docs/v3.3/dev-guide/api_grpc_gateway/#notes) and if gRPC isn't enabled we can't even figure out the version because `GET /version` call fails.

Regarding `version_prefix` in the configuration:
- for `etcd3` it is always discarded because of the nature of this PR
- for `etcd` we actually could protect from user setting it by overriding `EtcdClient.__init__()`

matthbakeredb:
> As a result of this change the error message will look like:
> 
> ```
> Failed to get list of machines from https://XXX:443/v3beta: <Unknown error: '<html>
> <head><title>502 Bad Gateway</title></head>
> ...
> ```
> 
> I.e., `/v3beta` instead of `/v2`.
> 
> Unfortunately we can't do better than that, because this prefix is [specific to an Etcd version](https://etcd.io/docs/v3.3/dev-guide/api_grpc_gateway/#notes) and if gRPC isn't enabled we can't even figure out the version because `GET /version` call fails.

Ah gotcha. Sounds good.

> Regarding `version_prefix` in the configuration:
> 
> * for `etcd3` it is always discarded because of the nature of this PR
> * for `etcd` we actually could protect from user setting it by overriding `EtcdClient.__init__()`

Cool.

matthbakeredb (approved):
lgtm

barthisrael (approved):
At a first glance I'd say the user would be able to mess with `version_prefix` if they set that key in the configuration file, under `etcd` or `etcd3` sections. See:

```python
>>> config = {'prefix': 'random', 'a': 'b', 'c': 'd'}
>>> {'prefix': 'v3', **config}
{'prefix': 'random', 'a': 'b', 'c': 'd'}

>>> del config['prefix']
>>> {'prefix': 'v3', **config}
{'prefix': 'v3', 'a': 'b', 'c': 'd'}
```

OTOH I would not expect a user to set `version_prefix` in the configuration file as that's not mentioned anywhere in the documentation 😅. With that in mind I'd say we are good to go.

CyberDem0n:
Nice catch @barthisrael, the order should be different and it will work as expected:
```python
In [1]: config = {'prefix': 'random', 'a': 'b', 'c': 'd'}

In [2]:  {'prefix': 'v3', **config}
Out[2]: {'prefix': 'random', 'a': 'b', 'c': 'd'}

In [3]:  {**config, 'prefix': 'v3'}
Out[3]: {'prefix': 'v3', 'a': 'b', 'c': 'd'}
```

Pushed the fix in 19456c21b18ed044b14fff08110464c485f01626.

barthisrael:
> Nice catch @barthisrael, the order should be different and it will work as expected:
> 
> ```python
> In [1]: config = {'prefix': 'random', 'a': 'b', 'c': 'd'}
> 
> In [2]:  {'prefix': 'v3', **config}
> Out[2]: {'prefix': 'random', 'a': 'b', 'c': 'd'}
> 
> In [3]:  {**config, 'prefix': 'v3'}
> Out[3]: {'prefix': 'v3', 'a': 'b', 'c': 'd'}
> ```
> 
> Pushed the fix in [19456c2](https://github.com/zalando/patroni/commit/19456c21b18ed044b14fff08110464c485f01626).

Ah, that's even better :) for some reason I didn't think of switching order yesterday 😆

hughcapet:
👍

CyberDem0n:
:+1:

PR #2734, opened 2023-07-07
Allow integer gucs without units in validation

feikesteenbergen:
Previously, integer gucs, for example `max_connections` would not pass the validation, as these settings have no unit, if and only if they were specified as a string.

This causes problems if the `max_connections` is configured in `patroni.yaml` as a string, for example, the following configuration would not result in the right `max_connections` settings, as `max_connections` is configured as a string:

    bootstrap:
      dcs:
        postgresql:
          parameters:
            log_checkpoints: "on"
            log_connections: "off"
            max_connections: "57"

Allowing a user to specify *all* parameters as a string was accepted before in Patroni and also seems very useful, as many of us will be using Ansible/Helm/Golang to build a Patroni configuration, in which creating a `map[string]string` is easier than having to deal with data types.

Attemps to address issue #2735

### DIFF
--- a/features/patroni_api.feature
+++ b/features/patroni_api.feature
@@ -35,21 +35,21 @@ Scenario: check local configuration reload
 	Then I receive a response code 202
 
 Scenario: check dynamic configuration change via DCS
-	Given I run patronictl.py edit-config -s 'ttl=10' -p 'max_connections=101' --force batman
-	Then I receive a response returncode 0
-	And I receive a response output "+ttl: 10"
+	Given I issue a PATCH request to http://127.0.0.1:8008/config with {"ttl": 20, "postgresql": {"parameters": {"max_connections": "101"}}}
+	Then I receive a response code 200
 	And Response on GET http://127.0.0.1:8008/patroni contains pending_restart after 11 seconds
 	When I issue a GET request to http://127.0.0.1:8008/config
 	Then I receive a response code 200
-	And I receive a response ttl 10
+	And I receive a response ttl 20
 	When I issue a GET request to http://127.0.0.1:8008/patroni
 	Then I receive a response code 200
 	And I receive a response tags {'new_tag': 'new_value'}
 	And I sleep for 4 seconds
 
 Scenario: check the scheduled restart
-	Given I issue a PATCH request to http://127.0.0.1:8008/config with {"postgresql": {"parameters": {"superuser_reserved_connections": "6"}}}
-	Then I receive a response code 200
+	Given I run patronictl.py edit-config -p 'superuser_reserved_connections=6' --force batman
+	Then I receive a response returncode 0
+	And I receive a response output "+    superuser_reserved_connections: 6"
 	And Response on GET http://127.0.0.1:8008/patroni contains pending_restart after 5 seconds
 	Given I issue a scheduled restart at http://127.0.0.1:8008 in 5 seconds with {"role": "replica"}
 	Then I receive a response code 202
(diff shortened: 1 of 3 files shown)


feikesteenbergen:
I haven't fully setup a Python dev environment yet, so this is a shot in the dark, will need to test this solution. However, reading the code, this seems to be the part that causes the issue we have identified.

feikesteenbergen:
To confirm, this does solve the issue I raised, Patroni + this path allows `max_connections: "50"` to be set. Result:

```console
$ patronictl show-config | grep max_conn
    max_connections: '50'
```

sebastianwebber:
hey @feikesteenbergen it will be nice to [add a scenario like this](https://github.com/zalando/patroni/blob/1c36112b44b62bb668f0f23f8ad1ff9b1cff5561/features/patroni_api.feature#L37C1-L48) to ensure that when changing the any of the gucs with problem it will will point that the cluster needs a restart.

CyberDem0n:
@feikesteenbergen indeed, lets change behave tests to something like:
```diff
diff --git a/features/patroni_api.feature b/features/patroni_api.feature
index 0c47fe6..ec04afd 100644
--- a/features/patroni_api.feature
+++ b/features/patroni_api.feature
@@ -35,21 +35,21 @@ Scenario: check local configuration reload
        Then I receive a response code 202
 
 Scenario: check dynamic configuration change via DCS
-       Given I run patronictl.py edit-config -s 'ttl=10' -p 'max_connections=101' --force batman
-       Then I receive a response returncode 0
-       And I receive a response output "+ttl: 10"
+    Given I issue a PATCH request to http://127.0.0.1:8008/config with {"ttl": 20, "postgresql": {"parameters": {"max_connections": "101"}}}
+       Then I receive a response code 200
        And Response on GET http://127.0.0.1:8008/patroni contains pending_restart after 11 seconds
        When I issue a GET request to http://127.0.0.1:8008/config
        Then I receive a response code 200
-       And I receive a response ttl 10
+       And I receive a response ttl 20
        When I issue a GET request to http://127.0.0.1:8008/patroni
        Then I receive a response code 200
        And I receive a response tags {'new_tag': 'new_value'}
        And I sleep for 4 seconds
 
 Scenario: check the scheduled restart
-       Given I issue a PATCH request to http://127.0.0.1:8008/config with {"postgresql": {"parameters": {"superuser_reserved_connections": "6"}}}
-       Then I receive a response code 200
+       Given I run patronictl.py edit-config -p 'superuser_reserved_connections=6' --force batman
+       Then I receive a response returncode 0
+       And I receive a response output "+    superuser_reserved_connections: 6"
        And Response on GET http://127.0.0.1:8008/patroni contains pending_restart after 5 seconds
        Given I issue a scheduled restart at http://127.0.0.1:8008 in 5 seconds with {"role": "replica"}
        Then I receive a response code 202
```

features/patroni_api.feature
```diff
@@ -35,21 +35,23 @@ Scenario: check local configuration reload
 	Then I receive a response code 202
 
 Scenario: check dynamic configuration change via DCS
-	Given I run patronictl.py edit-config -s 'ttl=10' -p 'max_connections=101' --force batman
-	Then I receive a response returncode 0
-	And I receive a response output "+ttl: 10"
+	Given I issue a PATCH request to http://127.0.0.1:8008/config with {"ttl": 20, "postgresql": {"parameters": {"max_connections": "101"}}}
+	Then I receive a response code 200
+	And I receive a response output "ttl: 20"
+	And I receive a response output "max_connections: 101"
```
CyberDem0n, line 41: @feikesteenbergen behave tests are failing because of this.
In my suggestion there was no these lines

hughcapet:
👍

CyberDem0n:
:+1:

PR #2736, opened 2023-07-07
Compatibility with click==8.1.4

CyberDem0n:
They somehow messed up with type hints what made pyright unhappy.
To solve it we explicitly pass the Group class to the group() decorator.

In addition to that bump pyright version.


hughcapet:
👍

PR #2737, opened 2023-07-07
Check py files in features with flake8

CyberDem0n:
They are correctly formatted and there is no reason not to enforce it.


hughcapet:
👍

CyberDem0n:
:+1:

Issue #2735, opened 2023-07-07
Patroni ignores GUC's specified as string if they are validated


feikesteenbergen:
### What happened?

We configure our Patroni using a golang operator. One of the things it does is to configure the `patroni.yaml`, for example:

```yaml
        bootstrap:
          dcs:
            postgresql:
              parameters:
                log_checkpoints: "on"
                log_connections: "off"
                max_connections: "57"
```

This used to work fine, but we just switched to the most recent Patroni, and this fails. `max_connections` is not honored, it is set to the default of 100.

Cmdlines as set by Patroni:
```console
$ cat /proc/154/cmdline | tr '\0' '\n'
/usr/lib/postgresql/14/bin/postgres
-D
/home/postgres/pgdata/data
--config-file=/home/postgres/pgdata/data/postgresql.conf
--listen_addresses=0.0.0.0
--port=5432
--cluster_name=a59r0uuuj3-an
--wal_level=replica
--hot_standby=on
--max_connections=100
--max_wal_senders=10
--max_prepared_transactions=100
--max_locks_per_transaction=69
--track_commit_timestamp=off
--max_replication_slots=10
--max_worker_processes=8
--wal_log_hints=on
```

### How can we reproduce it (as minimally and precisely as possible)?

Set `max_connections` to `"57"` in `patroni.yaml` and see the results.

### What did you expect to happen?

That the parameter was interpreted and used by Patroni.

Notice that `max_connections` is *missing* from patronictl show-config, and therefore defaults to 100.

Other guc's that are not involved in the parameter checking, for example, `autovacuum_max_workers` are handled correctly. (This makes sense to me, Patroni does not inspect these, it only passes them along).

p

### Patroni/PostgreSQL/DCS version

-   Patroni version: 3.0.3
-   PostgreSQL version: 15.3
-   DCS (and its version): Kubernetes

### Patroni configuration file

Slightly redacted

```yaml
bootstrap:
  dcs:
    failsafe_mode: true
    loop_wait: 10
    master_start_timeout: 15
    maximum_lag_on_failover: 16777217
    postgresql:
      authentication: {}

     parameters:
        autovacuum_max_workers: "10"
        max_connections: "79"
```

### patronictl show-config

Slightly redacted

```yaml
postgresql:
  parameters:
    archive_command: envdir /etc/postgresql/environment "/etc/postgresql/scripts/timescaledb/wal_archive_command_pgbackrest.sh" "%p"
    archive_mode: 'on'
[...]

PR #2740, opened 2023-07-10
Warn about failed validation configuration parameters

feikesteenbergen:
In issue #2735 it was discussed that there should be some warning around PostgreSQL parameters that do not pass validation.

This commit ensures something is logged for parameters that fail validation and therefore fall back to default values.

### DIFF
--- a/patroni/config.py
+++ b/patroni/config.py
@@ -329,9 +329,19 @@ class Config(object):
 
     @staticmethod
     def _process_postgresql_parameters(parameters: Dict[str, Any], is_local: bool = False) -> Dict[str, Any]:
-        return {name: value for name, value in (parameters or {}).items()
-                if name not in ConfigHandler.CMDLINE_OPTIONS
-                or not is_local and ConfigHandler.CMDLINE_OPTIONS[name][1](value)}
+        pg_params: Dict[str, Any] = {}
+
+        for name, value in (parameters or {}).items():
+            if name not in ConfigHandler.CMDLINE_OPTIONS:
+                pg_params[name] = value
+            elif not is_local:
+                if ConfigHandler.CMDLINE_OPTIONS[name][1](value):
+                    pg_params[name] = value
+                else:
+                    logging.warning("postgresql parameter %s=%s failed validation, defaulting to %s",
+                                    name, value, ConfigHandler.CMDLINE_OPTIONS[name][0])
+
+        return pg_params
 
     def _safe_copy_dynamic_configuration(self, dynamic_configuration: Dict[str, Any]) -> Dict[str, Any]:
         config = deepcopy(self.__DEFAULT_CONFIG)
--- a/patroni/postgresql/config.py
+++ b/patroni/postgresql/config.py
@@ -15,7 +15,7 @@ from ..collections import CaseInsensitiveDict, CaseInsensitiveSet
 from ..dcs import Leader, Member, RemoteMember, slot_name_from_member_name
 from ..exceptions import PatroniFatalException
 from ..utils import compare_values, parse_bool, parse_int, split_host_port, uri, validate_directory, is_subpath
-from ..validator import IntValidator
+from ..validator import IntValidator, EnumValidator
 
 if TYPE_CHECKING:  # pragma: no cover
     from . import Postgresql
@@ -258,14 +258,6 @@ def _false_validator(value: Any) -> bool:
     return False
 
 
-def _wal_level_validator(value: Any) -> bool:
-    return str(value).lower() in ('hot_standby', 'replica', 'logical')
-
-
-def _bool_validator(value: Any) -> bool:
-    return parse_bool(value) is not None
-
-
 class ConfigHandler(object):
 
     # List of parameters which must be always passed to postmaster as command line options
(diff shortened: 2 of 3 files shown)


patroni/config.py
```diff
@@ -329,9 +329,21 @@ def reload_local_configuration(self) -> Optional[bool]:
...
+
+        for name, value in (parameters or {}).items():
+            if name not in ConfigHandler.CMDLINE_OPTIONS:
+                pg_params[name] = value
+            elif not is_local:
+                if ConfigHandler.CMDLINE_OPTIONS[name][1](value):
+                    pg_params[name] = value
+                else:
+                    default = ConfigHandler.CMDLINE_OPTIONS[name][0]
+                    if value != default:
```
CyberDem0n, line 342: I guess `value == default` is impossible situation if validation failed.

feikesteenbergen: The `_false_validator` always returns `False` and would therefore always log.

I hit this situation when explicitly specifying `hot_standby=on` in our configuration, that *fails* validation currently.

hughcapet:
👍

feikesteenbergen:
I was thinking, that next to `_false_validator`, we could also create something like a `_equals_validator`, which would return `True` if and only if the provided value equals the default. That would make this code better as well I think.

CyberDem0n:
@feikesteenbergen agree, please go ahead

patroni/validator.py
```diff
@@ -758,6 +758,25 @@ def assert_(condition: bool, message: str = "Wrong value") -> None:
     assert condition, message
 
 
+class ListValidator(object):
```
CyberDem0n, line 761: I would actually call in `EnumValidator`.
But, it will not correctly work for `wal_log_hints` and `hot_standby`.
`'on'`, `'true'`, `'1'`, `1`, `True` -- these are all valid values for them.

CyberDem0n: Yes, something like `boolean_is_true_validator()`. it could be defined right in patroni/postgresql/config.py

patroni/validator.py
```diff
@@ -758,6 +758,46 @@ def assert_(condition: bool, message: str = "Wrong value") -> None:
...
+        True
+        >>> ListValidator(allowed=['off','on','blind'], case_insensitive=True)('ON')
+        False
+        """
+        self.case_insensitive = case_insensitive
+
+        if self.case_insensitive:
+            self.allowed = [a.lower() if isinstance(a, str) else a for a in allowed]
+        else:
+            self.allowed = allowed
```
CyberDem0n, line 790: A couple of comments:
1. Enum values in Postgres configuration are always strings.
2. instead of casting it to lowercase please just use `patroni.collections.CaseInsensitiveSet()`
3. if `case_insensitive is False` wrap values to `set()`

2 and 3 will simplify `__call__()` method and make `self.case_insensitive` redundant.

patroni/postgresql/config.py
```diff
@@ -286,18 +278,18 @@ class ConfigHandler(object):
         'listen_addresses': (None, _false_validator, 90100),
         'port': (None, _false_validator, 90100),
         'cluster_name': (None, _false_validator, 90500),
-        'wal_level': ('hot_standby', _wal_level_validator, 90100),
-        'hot_standby': ('on', _false_validator, 90100),
+        'wal_level': ('hot_standby', EnumValidator(allowed=['hot_standby', 'replica', 'logical']), 90100),
+        'hot_standby': ('on', EnumValidator(allowed=[True]), 90100),
```
CyberDem0n, line 282: The input value could be anything, `str`, `bool`, `int`, or even `None`.

patroni/postgresql/config.py
```diff
@@ -286,18 +278,18 @@ class ConfigHandler(object):
...
         'max_connections': (100, IntValidator(min=25), 90100),
         'max_wal_senders': (10, IntValidator(min=3), 90100),
         'wal_keep_segments': (8, IntValidator(min=1), 90100),
         'wal_keep_size': ('128MB', IntValidator(min=16, base_unit='MB'), 130000),
         'max_prepared_transactions': (0, IntValidator(min=0), 90100),
         'max_locks_per_transaction': (64, IntValidator(min=32), 90100),
-        'track_commit_timestamp': ('off', _bool_validator, 90500),
+        'track_commit_timestamp': ('off', EnumValidator(allowed=[True, False]), 90500),
```
CyberDem0n, line 289: I would leave this one as it was.

patroni/validator.py
```diff
@@ -758,6 +758,52 @@ def assert_(condition: bool, message: str = "Wrong value") -> None:
...
+        else:
+            self.allowed = [a.lower() if isinstance(a, str) else a for a in allowed]
+
+    def __call__(self, value: Any) -> bool:
+        """Check if *value* is part of the allowed values."""
+
+        if not self.case_sensitive and isinstance(value, str):
+            value = value.lower()
+
+        return value in self.allowed or parse_bool(value) in self.allowed
```
CyberDem0n, line 804: Sorry, but I really dislike that the `EnumValidator` for some reason also handling boolean values.
Enums in Postgres are generally strings, with some minor exceptions like `synchronous_commit`, where PG treats `1` the same way as `on` or `tRuE`. 

- For `wal_level` we don't expect anything but only string values. Neither we do expect to push more GUCS via command line in a foreseeable future. A very simple `EnumValidator` with `patroni.collections.CaseInsensitiveSet()` will perfectly do the job of validating `wal_level` values.
- For `track_commit_timestamp` and `wal_log_hints` we expect only booleans (even if they arrive as `str` or `int` objects).

Mixing two different things together in a single class is not a good practice.
For the reference, there is already [patroni/postgresql/validator.py](https://github.com/zalando/patroni/blob/d46ca88e6bb4b4788ec86689026c2d2e4da68c79/patroni/postgresql/validator.py#L114-L136), that makes a difference between GUC's that are just enums and GUC's that are a mixture of enum with bool, but unfortunately it can't be reused.

Issue #2738, opened 2023-07-07
Patroni stuck on newly promoted node, does not reply to health endpoint


boudey:
### What happened?

After failover request, newly promoted node patroni appears to be stuck, logs not increasing, most REST endpoints not responding.

```
$ patronictl version
patronictl version 3.0.2
```

failover request issued on leader node_0 and successfully executed

```
2023-07-07 06:59:43,163 INFO: no action. I am (node_0), the leader with the lock
2023-07-07 06:59:53,163 INFO: no action. I am (node_0), the leader with the lock
2023-07-07 07:00:02,579 INFO: received failover request with leader=None candidate=node_1 scheduled_at=None
2023-07-07 07:00:02,587 INFO: Got response from node_1 http://100.64.74.200:8008/patroni: {"state": "running", "postmaster_start_time": "2023-07-07 06:51:36.636384+00:00", "role": "replica", "server_version": 120015, "xlog": {"received_location": 153306304, "replayed_location": 153306304, "replayed_timestamp": "2023-07-07 06:59:58.482268+00:00", "paused": false}, "sync_standby": true, "timeline": 1, "dcs_last_seen": 1688713193, "tags": {"clonefrom": true}, "database_system_identifier": "7252718675178405984", "patroni": {"version": "3.0.2", "scope": "postgres"}}
2023-07-07 07:00:02,623 INFO: Got response from node_1 http://100.64.74.200:8008/patroni: {"state": "running", "postmaster_start_time": "2023-07-07 06:51:36.636384+00:00", "role": "replica", "server_version": 120015, "xlog": {"received_location": 153306304, "replayed_location": 153306304, "replayed_timestamp": "2023-07-07 06:59:58.482268+00:00", "paused": false}, "sync_standby": true, "timeline": 1, "dcs_last_seen": 1688713193, "tags": {"clonefrom": true}, "database_system_identifier": "7252718675178405984", "patroni": {"version": "3.0.2", "scope": "postgres"}}
[...]

PR #2741, opened 2023-07-10
A few fixes in synchronous_mode

CyberDem0n:
- make sure that physical replication slots are created even before the promote happened (when async executor is busy with promote).
- execute `txid_current()` with `synchronous_commit=off` so it doesn't accidentally wait for absent synchronous standbys when `synchronous_mode_strict` is enable and `synchronous_standby_names=*`. These standbys can't connect because replication slots weren't there.
- `synchronous_standby_names` wasn't set to `*` after bootstrap with `synchronous_mode` and `synchronous_mode_strict`.
- add `-c statement_timeout=0` to `PGOPTIONS` when executing `post_bootstrap` script.

Close https://github.com/zalando/patroni/issues/2738

### DIFF
--- a/patroni/ha.py
+++ b/patroni/ha.py
@@ -1711,17 +1711,21 @@ class Ha(object):
                 msg = self.process_healthy_cluster()
                 ret = self.evaluate_scheduled_restart() or msg
 
-            # we might not have a valid PostgreSQL connection here if another thread
-            # stops PostgreSQL, therefore, we only reload replication slots if no
-            # asynchronous processes are running (should be always the case for the primary)
-            if (not self._async_executor.busy or self._async_executor.scheduled_action == 'promote')\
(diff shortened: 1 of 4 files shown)


patroni/ha.py
```diff
@@ -1711,7 +1714,8 @@ def _run_cycle(self) -> str:
             # we might not have a valid PostgreSQL connection here if another thread
             # stops PostgreSQL, therefore, we only reload replication slots if no
             # asynchronous processes are running (should be always the case for the primary)
-            if not self._async_executor.busy and not self.state_handler.is_starting():
+            if (not self._async_executor.busy or self._async_executor.scheduled_action == 'promote')\
```
hughcapet, line 1717: should we adjust the comment above?

hughcapet:
👍

CyberDem0n:
:+1:

PR #2742, opened 2023-07-10
Reduce flakiness of citus behave tests, take 2

CyberDem0n:
Reorder some checks and verify that the old primary is already in the `running` state before checking replication. This check elliminates the race condition when replication started to work but node name is removed from the `synchronous_standby_names` because state isn't `running`.


hughcapet:
👍

CyberDem0n:
:+1:

PR #2743, opened 2023-07-10
Document and refactor dcs/__init__.py and postgres/slots.py

matthbakeredb:
This PR addresses 2 streams of work.

* During analysis of replication slot process for better understanding, and possible future work related to the integration of the extension of `pg_failover_slots`, we have refactored the methods used to make decisions in slot replication. Enhances readability of the code and descriptive naming of methods, variables, arguments, etc.
* Write docstrings for `dcs/__init__.py` and `postgresql/slots.py`.

There are some minor code tweaks to things like rewriting loops to generator statements and some linter fixes.

PR #2745, opened 2023-07-11
Refactor allowed_keys

matthbakeredb:
Method does not perform any computation or modify data as it is a static
tuple, therefore it is better expressed as a class variable.

Split from #2743


CyberDem0n:
:+1:

hughcapet:
👍

PR #2746, opened 2023-07-11
Refactor get replication slots

matthbakeredb:
Reduce complexity of single method and allow for documentation of distinct parts.

No functional changes have been introduced.

Split from #2743

### DIFF
--- a/patroni/dcs/__init__.py
+++ b/patroni/dcs/__init__.py
@@ -594,7 +594,7 @@ class Cluster(NamedTuple):
         :param role: role of this node.
         :param nofailover: ``True`` if this node is tagged to not be a failover candidate.
         :param major_version: postgresql major version.
-        :param show_error: if ``True`` report error if any disabled logical slots are found.
+        :param show_error: if ``True`` report error if any disabled logical slots or conflicting slot names are found.
 
         :returns: final dictionary of slot names, after merging with permanent slots and performing sanity checks.
         """
@@ -628,7 +628,7 @@ class Cluster(NamedTuple):
 
         Perform validation of configured permanent slot name, skipping invalid names.
 
-        Will update *slots* in-line based on ``type`` of slot, ``physical`` or ``logical``, and *role* of node.
+        Will update *slots* in-line based on ``type`` of slot, ``physical`` or ``logical``, and name of node.
         Type is assumed to be ``physical`` if there are no attributes stored as the slot value.
 
         :param slots: Slot names with existing attributes if known.
@@ -709,7 +709,7 @@ class Cluster(NamedTuple):
         :param my_name: name of this node.
         :param role: role of this node, if this is a ``primary`` or ``standby_leader`` return list of members
                      replicating from this node. If not then return a list of members replicating as cascaded
-                     replicas.
+                     replicas from this node.
 
         :returns: list of member names.
         """
--- a/patroni/dcs/__init__.py
+++ b/patroni/dcs/__init__.py
@@ -612,8 +612,9 @@ class Cluster(NamedTuple):
                          "; ".join(f"{', '.join(v)} map to {k}"
                                    for k, v in slot_conflicts.items() if len(v) > 1))
 
+        permanent_slots: dict[str, Any] = self._get_permanent_slots(role, nofailover) if self.use_slots else {}
         disabled_permanent_logical_slots: list[str] = self._merge_permanent_slots(
-            slots, self._get_permanent_slots(role, nofailover), my_name, major_version)
+            slots, permanent_slots, my_name, major_version)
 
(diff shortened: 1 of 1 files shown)


patroni/dcs/__init__.py
```diff
@@ -583,36 +583,63 @@ def use_slots(self) -> bool:
...
+
+        :param my_name: name of this node.
+        :param role: role of this node.
+        :param nofailover: ``True`` if this node is tagged to not be a failover candidate.
+        :param major_version: postgresql major version.
+        :param show_error: if ``True`` report error if any disabled logical slots or conflicting slot names are found.
+
+        :returns: final dictionary of slot names, after merging with permanent slots and performing sanity checks.
+        """
+        slot_members: list[str] = self._get_slot_members(my_name, role)
```
CyberDem0n, line 601: Should we better do something like 
```suggestion
        slot_members: list[str] = self._get_slot_members(my_name, role) if self.use_slots else []
```

and get rid of `use_slots` checks in the `_get_slot_members()`?

patroni/dcs/__init__.py
```diff
@@ -583,36 +583,63 @@ def use_slots(self) -> bool:
...
             for name in slot_members:
                 slot_conflicts[slot_name_from_member_name(name)].append(name)
             logger.error("Following cluster members share a replication slot name: %s",
-                         "; ".join("{} map to {}".format(", ".join(v), k)
+                         "; ".join(f"{', '.join(v)} map to {k}"
                                    for k, v in slot_conflicts.items() if len(v) > 1))
 
-        # "merge" replication slots for members with permanent_replication_slots
+        disabled_permanent_logical_slots: list[str] = self._merge_permanent_slots(
+            slots, self._get_permanent_slots(role, nofailover), my_name, major_version)
```
CyberDem0n, line 616: Similar to https://github.com/zalando/patroni/pull/2746/files#diff-247edd60473c0933e528a1b4621cc612da54b6dc964720719d7623135d312474R601:
```suggestion
        permanent_slots = self._get_permanent_slots(role, nofailover) if self.use_slots else {}
        disabled_permanent_logical_slots: list[str] = self._merge_permanent_slots(
            slots, permanent_slots, my_name, major_version)
```

and get rid of `use_slots` checks in the `_get_permanent_slots()`.

patroni/dcs/__init__.py
matthbakeredb: Nice! Yes that reduces some duplication. Committed.

patroni/dcs/__init__.py
matthbakeredb: Applied also in commit.

CyberDem0n: You missed this part:
```python
        permanent_slots = self._get_permanent_slots(role, nofailover) if self.use_slots else {}
```

matthbakeredb: Um, where? I don't see it?

CyberDem0n: Currently it looks like:
```python
        disabled_permanent_logical_slots: list[str] = self._merge_permanent_slots(
            slots, self._get_permanent_slots(role, nofailover), my_name, major_version)
```

While it should be:
```python
        permanent_slots = self._get_permanent_slots(role, nofailover) if self.use_slots else {}
        disabled_permanent_logical_slots: list[str] = self._merge_permanent_slots(
            slots, permanent_slots, my_name, major_version)
```

matthbakeredb: Oh duh! I was looking for something to remove when I should've been looking for something to add. :roll_eyes: 

Added that now. Sorry!

CyberDem0n (commented):
A couple of small issues.
We better stick with List instead of list, because the latter will probably not with with python 3.6, which is still widely used according to pypi download stats.

patroni/dcs/__init__.py
```diff
@@ -583,36 +583,64 @@ def use_slots(self) -> bool:
...
-        slots = {slot_name_from_member_name(name): {'type': 'physical'} for name in slot_members}
+        :param my_name: name of this node.
+        :param role: role of this node.
+        :param nofailover: ``True`` if this node is tagged to not be a failover candidate.
+        :param major_version: postgresql major version.
+        :param show_error: if ``True`` report error if any disabled logical slots or conflicting slot names are found.
+
+        :returns: final dictionary of slot names, after merging with permanent slots and performing sanity checks.
+        """
+        slot_members: list[str] = self._get_slot_members(my_name, role) if self.use_slots else []
```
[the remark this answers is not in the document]
matthbakeredb: Ah yes, I do normally use `List` so don't know why I didn't this time!

hughcapet:
👍

CyberDem0n:
:+1:

PR #2747, opened 2023-07-11
Refactor get_dcs

matthbakeredb:
Now uses generators instead of for loops and implements importing modules once.

Split from #2743

### DIFF
--- a/patroni/dcs/__init__.py
+++ b/patroni/dcs/__init__.py
@@ -111,7 +111,7 @@ def iter_dcs_classes(
                     yield name, dcs_module
 
             except ImportError:
-                logger.log(logging.INFO if config is not None else logging.DEBUG,
+                logger.log(logging.DEBUG if config is not None else logging.INFO,
                            'Failed to import %s', mod_name)
 
 
--- a/patroni/dcs/__init__.py
+++ b/patroni/dcs/__init__.py
@@ -89,7 +89,7 @@ def dcs_modules() -> List[str]:
 
 def iter_dcs_classes(
         config: Optional[Union['Config', Dict[str, Any]]] = None
-) -> Iterator[Tuple[str, Optional[Type['AbstractDCS']]]]:
+) -> Iterator[Tuple[str, Type['AbstractDCS']]]:
     """Attempt to import DCS modules that are present in the given configuration.
 
     .. note::
@@ -103,11 +103,16 @@ def iter_dcs_classes(
     for mod_name in dcs_modules():
         name = mod_name.rpartition('.')[2]
         if config is None or name in config:
+
             try:
                 module = importlib.import_module(mod_name)
-                yield name, find_dcs_class_in_module(module)
+                dcs_module = find_dcs_class_in_module(module)
+                if dcs_module:
+                    yield name, dcs_module
+
             except ImportError:
-                logger.info('Failed to import %s', mod_name)
+                logger.log(logging.INFO if config is not None else logging.DEBUG,
+                           'Failed to import %s', mod_name)
 
 
 def find_dcs_class_in_module(module: ModuleType) -> Optional[Type['AbstractDCS']]:
@@ -146,20 +151,19 @@ def get_dcs(config: Union['Config', Dict[str, Any]]) -> 'AbstractDCS':
     :returns: The first successfully loaded DCS module which is an implementation of :class:`AbstractDCS`.
     """
     for name, dcs_class in iter_dcs_classes(config):
-        if dcs_class:
-            # Propagate some parameters from top level of config if defined to the DCS specific config section.
-            config[name].update({
-                p: config[p] for p in ('namespace', 'name', 'scope', 'loop_wait',
-                                       'patronictl', 'ttl', 'retry_timeout')
-                if p in config})
-            # From citus section we only need "group" parameter, but will propagate everything just in case.
-            if isinstance(config.get('citus'), dict):
-                config[name].update(config['citus'])
-            return dcs_class(config[name])
+        # Propagate some parameters from top level of config if defined to the DCS specific config section.
+        config[name].update({
+            p: config[p] for p in ('namespace', 'name', 'scope', 'loop_wait',
+                                   'patronictl', 'ttl', 'retry_timeout')
+            if p in config})
+        # From citus section we only need "group" parameter, but will propagate everything just in case.
+        if isinstance(config.get('citus'), dict):
+            config[name].update(config['citus'])
+        return dcs_class(config[name])
 
     raise PatroniFatalException(
         f"Can not find suitable configuration of distributed configuration store\n"
-        f"Available implementations: {', '.join(sorted(set(name for name, _ in iter_dcs_classes())))}")
+        f"Available implementations: {', '.join(sorted([n for n, _ in iter_dcs_classes()]))}")
 
 
 _Version = Union[int, str]
--- a/patroni/dcs/__init__.py
+++ b/patroni/dcs/__init__.py
@@ -95,7 +95,7 @@ def iter_dcs_modules(
     .. note::
             If a module successfully imports we can assume that all its requirements are installed.
 
-    :param config: configuration information with possible DCS names as keys.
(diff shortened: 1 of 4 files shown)


patroni/dcs/__init__.py
```diff
@@ -85,38 +87,78 @@ def dcs_modules() -> List[str]:
...
+                                       'patronictl', 'ttl', 'retry_timeout')
+                if p in config})
+            # From citus section we only need "group" parameter, but will propagate everything just in case.
+            if isinstance(config.get('citus'), dict):
+                config[name].update(config['citus'])
+            return dcs_class(config[name])
+
+    raise PatroniFatalException(
+        f"Can not find suitable configuration of distributed configuration store\n"
+        f"Available implementations: {', '.join(sorted(set(name for name, _ in iter_dcs_modules())))}")
```
CyberDem0n, line 161: The original implementation was also checking that the module has a class implementing `AbstractDCS`: https://github.com/zalando/patroni/pull/2747/files#diff-247edd60473c0933e528a1b4621cc612da54b6dc964720719d7623135d312474L114-L115

I would suggest to move this filter (`find_dcs_class_in_module`) inside `iter_dcs_modules()`, which actually could just return DCS classes.

patroni/dcs/__init__.py
```diff
@@ -85,38 +87,78 @@ def dcs_modules() -> List[str]:
...
-                        config[name].update({p: config[p] for p in ('namespace', 'name', 'scope', 'loop_wait',
-                                             'patronictl', 'ttl', 'retry_timeout') if p in config})
-                        # From citus section we only need "group" parameter, but will propagate everything just in case.
-                        if isinstance(config.get('citus'), dict):
-                            config[name].update(config['citus'])
-                        return item(config[name])
+                yield name, importlib.import_module(mod_name)
             except ImportError:
-                logger.debug('Failed to import %s', module_name)
+                logger.debug('Failed to import %s', mod_name)
```
CyberDem0n, line 108: In case if no modules matching configuration were found we used to have `logger.info()` for failed imports.

matthbakeredb: Oh yes, I'll correct that.

patroni/dcs/__init__.py
matthbakeredb: Pushed a commit, let me know if this is what you were thinking?

patroni/dcs/__init__.py
```diff
@@ -85,38 +87,79 @@ def dcs_modules() -> List[str]:
...
-                                             'patronictl', 'ttl', 'retry_timeout') if p in config})
-                        # From citus section we only need "group" parameter, but will propagate everything just in case.
-                        if isinstance(config.get('citus'), dict):
-                            config[name].update(config['citus'])
-                        return item(config[name])
+                module = importlib.import_module(mod_name)
+                yield name, find_dcs_class_in_module(module)
             except ImportError:
-                logger.debug('Failed to import %s', module_name)
+                logger.info('Failed to import %s', mod_name)
```
CyberDem0n, line 110: Still a problem.
Now it will show unwanted messages when for example we have `kubernetes` in the configuration and no `python-etcd`, `python-consul`, `kazoo`, or `pysyncobj` modules installed.

patroni/dcs/__init__.py
```diff
@@ -85,38 +87,79 @@ def dcs_modules() -> List[str]:
...
+                                       'patronictl', 'ttl', 'retry_timeout')
+                if p in config})
+            # From citus section we only need "group" parameter, but will propagate everything just in case.
+            if isinstance(config.get('citus'), dict):
+                config[name].update(config['citus'])
+            return dcs_class(config[name])
+
+    raise PatroniFatalException(
+        f"Can not find suitable configuration of distributed configuration store\n"
+        f"Available implementations: {', '.join(sorted(set(name for name, _ in iter_dcs_classes())))}")
```
CyberDem0n, line 162: If there is a module without proper class that inherits AbstractDCS we will get unexpected results:
```
Available implementations: etcd, etcd3, foo, kubernetes
```

patroni/dcs/__init__.py
```diff
@@ -85,38 +87,79 @@ def dcs_modules() -> List[str]:
...
-                    if key.lower() == name and inspect.isclass(item) and issubclass(item, AbstractDCS):
-                        # propagate some parameters
-                        config[name].update({p: config[p] for p in ('namespace', 'name', 'scope', 'loop_wait',
-                                             'patronictl', 'ttl', 'retry_timeout') if p in config})
-                        # From citus section we only need "group" parameter, but will propagate everything just in case.
-                        if isinstance(config.get('citus'), dict):
-                            config[name].update(config['citus'])
-                        return item(config[name])
+                module = importlib.import_module(mod_name)
+                yield name, find_dcs_class_in_module(module)
```
CyberDem0n, line 108: Why not yield when class is actually found?

CyberDem0n: In fact, we only need to yield a class.
The name could be calculated using `dcs_class.__name__.lower()`

patroni/dcs/__init__.py
matthbakeredb: I'll add a condition for the log level based on whether the `name` is in `config`. INFO for configured, DEBUG otherwise.

patroni/dcs/__init__.py
matthbakeredb: So the iter_dcs_classes is in 2 places. One to load the module, the other to just get the available mods to generate the exception for no "suitable configuration". The second element of the tuple would need to be `None` to preserve the name of the module.

I could mix the logic up in the test for `config is None or name in config` so that either we get the name or the class object but it feels messy.

```
            module = importlib.import_module(mod_name)
            if config is None:
                yield name
            elif name in config:
                dcs_module = find_dcs_class_in_module(module)
                if dcs_module:
                    yield dcs_module
```

It feels cleaner and more consistent to always return the same object type and tuple structure.

patroni/dcs/__init__.py
matthbakeredb: I created an empty `foo.py` under `patroni.dcs`, I get this:

```
list(name for name, _ in iter_dcs_classes())
['consul', 'etcd', 'etcd3', 'exhibitor', 'foo', 'kubernetes', 'raft', 'zookeeper']
```

What did you do that's different?

patroni/dcs/__init__.py
CyberDem0n: This is what the original code did:
```python
    available_implementations: List[str] = []
    for module_name in modules:
        name = module_name.split('.')[-1]
        try:
            module = importlib.import_module(module_name)
            available_implementations.extend(name for key, item in module.__dict__.items() if key.lower() == name
                                             and inspect.isclass(item) and issubclass(item, AbstractDCS))
        except ImportError:
            logger.info('Failed to import %s', module_name)
```

It is importing modules and trying to find a class derived from `AbstractDCS` and with the `__class__.name.lower()` matching the module_name`.
That is, the empty file `patroni/dcs/foo.py` will be skipped.

matthbakeredb: OK, I've pushed a couple of commits which I hope address this. `foo` is skipped now.

I reverted to passing back `name, dcs_class` instead of using `__class__.__name__`. There was a test which conflicted with this which I have added a suggested workaround. It might need a `patch()` instead? I also thought there is less clutter to using the dunder attrs.

tests/test_ha.py
```diff
@@ -213,6 +213,10 @@ def setUp(self):
         self.ha.old_cluster = self.e.get_cluster()
         self.ha.cluster = get_cluster_initialized_without_leader()
         self.ha.load_cluster_from_dcs = Mock()
+        self.ha_dcs_orig_name = self.ha.dcs.__class__.__name__
+
+    def tearDown(self) -> None:
+        self.ha.dcs.__class__.__name__ = self.ha_dcs_orig_name
```
CyberDem0n, line 219: Hmm.
I don't think that there is something that relies on it.
In any case we don't need to do it in setUp/tearDown, restoring right in the `test_recover_raft()` should be enough.

hughcapet:
👍

CyberDem0n:
:+1:

PR #2763, opened 2023-07-19
Start primary back when it crashed with  "in crash recovery" state

CyberDem0n:
If one of backeds crashed postmaster stops all backend and does crash recovery because shared memory could be corrupted. If something happens during this phase postgres may completely crash leaving pg_control with "in crash recovery" state.

Followup on #2726


hughcapet:
👍

CyberDem0n:
:+1:

PR #2768, opened 2023-07-21
Fixes in `patroni.request`

CyberDem0n:
1.  Take client certificates only from the `ctl` section. Motivation: sometimes there are server-only certificates that can't be used as client certificates. As a result neither Patroni not patronictl work correctly even if `--insecure` option is used.
2. Document that if `restapi.verify_client` is set to `required` then client certificates **must** be provided in the `ctl` section.
3.  Add support for `ctl.authentication` and prefer to use it over `restapi.authentication`.
4. Silence annoying InsecureRequestWarning when `patronictl -k` is used, so that behavior becomes is similar to `curl -k`.


docs/ENVIRONMENT.rst
```diff
@@ -196,10 +196,19 @@ REST API
 -  **PATRONI\_RESTAPI\_HTTPS\_EXTRA\_HEADERS**: (optional) HTTPS headers let the REST API server pass additional information with an HTTP response when TLS is enabled. This will also pass additional information set in ``http_extra_headers``.
 -  **PATRONI\_RESTAPI\_REQUEST\_QUEUE\_SIZE**: (optional): Sets request queue size for TCP socket used by Patroni REST API.  Once the queue is full, further requests get a "Connection denied" error. The default value is 5.
 
+.. warning::
+
+    - The ``PATRONI_RESTAPI_CONNECT_ADDRESS`` must be accessible from all nodes of a given Patroni cluster. Internally Patroni is using it during the leader race to find nodes with minimal replication lag.
+    - If you enabled client certificates validation (``PATRONI_RESTAPI_VERIFY_CLIENT`` is set to ``required``), you also **must** to provide **valid client certificates** in the ``PATRONI_CTL_CERTFILE``, ``PATRONI_CTL_KEYFILE``, ``PATRONI_CTL_KEYFILE_PASSWORD``. If not privided Patroni will not work correctly.
```
hughcapet, line 202:
```suggestion
    - If you enabled client certificates validation (``PATRONI_RESTAPI_VERIFY_CLIENT`` is set to ``required``), you also **must** provide **valid client certificates** in the ``PATRONI_CTL_CERTFILE``, ``PATRONI_CTL_KEYFILE``, ``PATRONI_CTL_KEYFILE_PASSWORD``. If not provided, Patroni will not work correctly.
```

patroni/request.py
```diff
@@ -79,37 +99,39 @@ def reload_config(self, config: Union[Config, Dict[str, Any]]) -> None:
...
-        * ``authorization``: based on Patroni' REST API authentication config;
+        * ``authorization``: based on Patroni' CTL or REST API authentication config;
         * ``user-agent``: based on `patroni.utils.USER_AGENT`.
 
         Also configure SSL related settings for requests:
 
         * ``ca_certs`` is configured if ``ctl -> cacert`` or ``restapi -> cafile`` is available;
-        * ``cert``, ``key`` and ``key_password`` are configured if ``ctl -> certile`` or ``restapi -> certfile`` is
-            available.
+        * ``cert``, ``key`` and ``key_password`` are configured if ``ctl -> certile`` is available.
```
hughcapet, line 108:
```suggestion
        * ``cert``, ``key`` and ``key_password`` are configured if ``ctl -> certfile`` is available.
```

typings/urllib3/poolmanager.pyi
```diff
@@ -1,4 +1,5 @@
-from typing import Any, Dict, Optional
+from typing import Any, Dict, Optional, Type
```
barthisrael, line 1:
```suggestion
from typing import Any, Dict, Optional
```

I couldn't find any reference to `Type` in that file.

typings/urllib3/poolmanager.pyi
```diff
@@ -1,4 +1,5 @@
...
+from .connectionpool import HTTPConnectionPool
```
barthisrael, line 2:
```suggestion
```

I think this one is not needed here either.

barthisrael (changes requested):
Overall looks good. Just some small comments.

hughcapet:
👍

CyberDem0n:
:+1:

PR #2770, opened 2023-07-21
Fix Dockerfile_s

hughcapet:
- Install dumb-init using apt to avoid 'externally managed environment' error
- Remove python 2.7 packages purge (current base image debian-bookworm doesn't have python 2.7)

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -26,7 +26,7 @@ RUN set -ex \
             | xargs apt-get install -y vim curl less jq locales haproxy sudo \
                             python3-etcd python3-kazoo python3-pip busybox \
                             net-tools iputils-ping --fix-missing \
-    && pip3 install dumb-init \
+    && apt-get install -y dumb-init \
 \
     # Cleanup all locales but en_US.UTF-8
     && find /usr/share/i18n/charmaps/ -type f ! -name UTF-8.gz -delete \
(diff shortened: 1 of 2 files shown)


Dockerfile
```diff
@@ -26,7 +26,7 @@ RUN set -ex \
             | xargs apt-get install -y vim curl less jq locales haproxy sudo \
                             python3-etcd python3-kazoo python3-pip busybox \
                             net-tools iputils-ping --fix-missing \
-    && pip3 install dumb-init \
+    && apt-get install -y dumb-init \
```
CyberDem0n, line 29: Lets merge it with previous apt-get install.

Dockerfile.citus
```diff
@@ -42,7 +42,7 @@ RUN set -ex \
         && apt-get update -y \
         && apt-get -y install postgresql-$PG_MAJOR-citus-11.3; \
     fi \
-    && pip3 install dumb-init \
+    && apt-get install -y dumb-init \
```
CyberDem0n, line 45: The same as https://github.com/zalando/patroni/pull/2770/files#diff-dd2c0eb6ea5cfc6c4bd4eac30934e2d5746747af48fef6da689e85b752f39557R29

CyberDem0n:
:+1:

hughcapet:
👍

PR #2773, opened 2023-07-25
Fix `in_recovery` check

CyberDem0n:
The primary that is still alive isn't properly recognized.
Regression was introduced in #2652.


hughcapet:
👍

CyberDem0n:
:+1:

PR #2775, opened 2023-07-26
Fix patronictl list

CyberDem0n:
the `Cluster` name field was missing in tsv, json, and yaml formats

The bug was introduced in #2652


hughcapet:
👍

CyberDem0n:
:+1:

PR #2777, opened 2023-07-26
Add docs to `patroni.dcs.__init__.py`

matthbakeredb:
Also, made some small code changes to satisfy formatting and pylint.

### DIFF
--- a/patroni/dcs/__init__.py
+++ b/patroni/dcs/__init__.py
@@ -396,7 +396,6 @@ class RemoteMember(Member):
         """
         if name in RemoteMember.ALLOWED_KEYS:
             return self.data.get(name)
-        return None
 
 
 class Leader(NamedTuple):
--- a/patroni/dcs/__init__.py
+++ b/patroni/dcs/__init__.py
@@ -379,7 +379,8 @@ class RemoteMember(Member):
         """Factory method to construct instance from given *name* and *data*.
 
         :param name: name of the remote member.
-        :param data: dictionary of member information.
+        :param data: dictionary of member information, which can contain keys from :const:`¬RemoteMember.ALLOWED_KEYS`
+                     but also member connection information ``api_url`` and ``conn_kwargs`, and slot information.
 
         :returns: constructed instance using supplied parameters.
         """
--- a/patroni/dcs/__init__.py
+++ b/patroni/dcs/__init__.py
@@ -767,7 +767,8 @@ class TimelineHistory(NamedTuple):
         """Parse the given JSON serialized string as a list of timeline history lines.
 
         :param version: version number
-        :param value: JSON serialized string.
+        :param value: JSON serialized string, consisting of parsed lines of PostgreSQL timeline history files,
+                      see :class:`TimelineHistory`.
 
         :returns: composed timeline history object using parsed lines.
 
--- a/patroni/dcs/__init__.py
+++ b/patroni/dcs/__init__.py
@@ -846,11 +846,13 @@ class Cluster(NamedTuple('Cluster',
            >>> cluster = Cluster.empty()
            >>> len(cluster)
            0
+
            >>> assert bool(cluster) is False
 
-           >>> cluster = Cluster(None, None, None, 0, [1, 2, 3], None, SyncState.empty(), None, None, None)
+           >>> cluster = Cluster(None, None, None, 0, [1, 2, 3], None, SyncState.empty(), None, None, None, {})
            >>> len(cluster)
            1
+
            >>> assert bool(cluster) is True
 
            This makes it easier to write ``if cluster`` rather than the longer statement.
@@ -1137,19 +1139,19 @@ class Cluster(NamedTuple('Cluster',
         :Example:
 
             No history provided:
-            >>> Cluster(0, 0, 0, 0, 0, 0, 0, 0, 0, None).timeline
+            >>> Cluster(0, 0, 0, 0, 0, 0, 0, 0, 0, None, {}).timeline
             0
 
             Empty history assume timeline is ``1``:
-            >>> Cluster(0, 0, 0, 0, 0, 0, 0, TimelineHistory.from_node(1, '[]'), 0, None).timeline
+            >>> Cluster(0, 0, 0, 0, 0, 0, 0, TimelineHistory.from_node(1, '[]'), 0, None, {}).timeline
             1
 
             Invalid history format, a string of ``a``, returns ``0``:
-            >>> Cluster(0, 0, 0, 0, 0, 0, 0, TimelineHistory.from_node(1, '[["a"]]'), 0, None).timeline
+            >>> Cluster(0, 0, 0, 0, 0, 0, 0, TimelineHistory.from_node(1, '[["a"]]'), 0, None, {}).timeline
             0
 
             History as a list of strings:
-            >>> Cluster(0, 0, 0, 0, 0, 0, 0, TimelineHistory.from_node(1, '[["3", "2", "1"]]'), 0, None).timeline
+            >>> Cluster(0, 0, 0, 0, 0, 0, 0, TimelineHistory.from_node(1, '[["3", "2", "1"]]'), 0, None, {}).timeline
             4
         """
         if self.history:
--- a/patroni/dcs/__init__.py
+++ b/patroni/dcs/__init__.py
@@ -1199,43 +1199,50 @@ class AbstractDCS(abc.ABC):
 
     Functional methods that perform construction of complex data objects:
 
-        * :meth:`~Cluster._cluster_loader`: method which processes the structure of data stored in the DCS used to build
-                                            the :class:`Cluster` object with all relevant associated data.
-        * :meth:`~Cluster._citus_cluster_loader`: Similar to above but specifically representing Citus group and workers
-                                                  information.
-        * :meth:`~Cluster._load_cluster`: main method for calling specific ``loader`` method to build the
-                                          :class:`Cluster` object representing the state and topology of the cluster.
+        * :meth:`~AbstractDCS._cluster_loader`:
+            method which processes the structure of data stored in the DCS used to build the :class:`Cluster` object
+            with all relevant associated data.
+        * :meth:`~AbstractDCS._citus_cluster_loader`:
+            Similar to above but specifically representing Citus group and workers information.
+        * :meth:`~AbstractDCS._load_cluster`:
+            main method for calling specific ``loader`` method to build the :class:`Cluster` object representing the
+            state and topology of the cluster.
 
     Functional methods that are critical in their timing and must be written with ACID transaction properties in mind:
 
-        * :meth:`~Cluster.attempt_to_acquire_leader`: method used in the leader race to attempt to acquire the leader
-                                                      lock by creating the leader key in the DCS, if it does not exist.
-        * :meth:`~Cluster.take_leader`: method to create a new leader key in the DCS.
-        * :meth:`~Cluster.initialize`: method used in the race for cluster initialization which creates the
-                                       ``initialize`` key in the DCS.
+        * :meth:`~AbstractDCS.attempt_to_acquire_leader`:
+            method used in the leader race to attempt to acquire the leader lock by creating the leader key in the DCS,
+            if it does not exist.
+        * :meth:`~AbstractDCS.take_leader`:
+            method to create a new leader key in the DCS.
+        * :meth:`~AbstractDCS.initialize`:
+            method used in the race for cluster initialization which creates the ``initialize`` key in the DCS.
 
     DCS backend getter and setter methods and properties:
 
-        * :meth:`~Cluster.set_ttl`: method for setting TTL value in DCS.
-        * :meth:`~Cluster.ttl`: property which returns the current TTL.
-        * :meth:`~Cluster.set_retry_timeout`: method for setting ``retry_timeout`` in DCS backend.
-        * :meth:`~Cluster._write_leader_optime`: compatibility method to write WAL LSN to DCS.
-        * :meth:`~Cluster._write_status`: method to write WAL LSN for slots to the DCS.
-        * :meth:`~Cluster._write_failsafe`: method to write cluster topology to the DCS, used by failsafe mechanism.
-        * :meth:`~Cluster._update_leader`: method to update ``leader`` key in DCS.
-        * :meth:`~Cluster.set_failover_value`: method to create and/or update the ``failover`` key in the DCS.
-        * :meth:`~Cluster.set_config_value`: method to create and/or update the ``failover`` key in the DCS.
-        * :meth:`~Cluster.touch_member`: method to update individual member key in the DCS.
-        * :meth:`~Cluster.set_history_value`: method to set the ``history`` key in the DCS.
-        * :meth:`~Cluster.set_sync_state_value`: method to set the synchronous state ``sync`` key in the DCS.
+        * :meth:`~AbstractDCS.set_ttl`: method for setting TTL value in DCS.
+        * :meth:`~AbstractDCS.ttl`: property which returns the current TTL.
+        * :meth:`~AbstractDCS.set_retry_timeout`: method for setting ``retry_timeout`` in DCS backend.
+        * :meth:`~AbstractDCS._write_leader_optime`: compatibility method to write WAL LSN to DCS.
+        * :meth:`~AbstractDCS._write_status`: method to write WAL LSN for slots to the DCS.
+        * :meth:`~AbstractDCS._write_failsafe`: method to write cluster topology to the DCS, used by failsafe mechanism.
+        * :meth:`~AbstractDCS._update_leader`: method to update ``leader`` key in DCS.
+        * :meth:`~AbstractDCS.set_failover_value`: method to create and/or update the ``failover`` key in the DCS.
+        * :meth:`~AbstractDCS.set_config_value`: method to create and/or update the ``failover`` key in the DCS.
+        * :meth:`~AbstractDCS.touch_member`: method to update individual member key in the DCS.
+        * :meth:`~AbstractDCS.set_history_value`: method to set the ``history`` key in the DCS.
+        * :meth:`~AbstractDCS.set_sync_state_value`: method to set the synchronous state ``sync`` key in the DCS.
 
     DCS data and key removal methods:
 
-        * :meth:`~Cluster.delete_sync_state`: likewise, a method to remove synchronous state ``sync`` key from the DCS.
-        * :meth:`~Cluster.delete_cluster`: method which will remove cluster information from the DCS.
-        * :meth:`~Cluster._delete_leader`: method used by a member that is the current leader, to remove the ``leader``
-                                           key in the DCS.
-        * :meth:`~Cluster.cancel_initialization`: method to remove the ``initialize`` key for the cluster from the DCS.
+        * :meth:`~AbstractDCS.delete_sync_state`:
+            likewise, a method to remove synchronous state ``sync`` key from the DCS.
+        * :meth:`~AbstractDCS.delete_cluster`:
+            method which will remove cluster information from the DCS.
+        * :meth:`~AbstractDCS._delete_leader`:
+            method used by a member that is the current leader, to remove the ``leader`` key in the DCS.
+        * :meth:`~AbstractDCS.cancel_initialization`:
+            method to remove the ``initialize`` key for the cluster from the DCS.
 
     Care should be taken to consult each abstract method for any additional information and requirements such as
     expected exceptions that should be raised in certain conditions and the object types for arguments and return from
--- a/patroni/dcs/__init__.py
+++ b/patroni/dcs/__init__.py
@@ -394,8 +394,7 @@ class RemoteMember(Member):
         :returns: value of *name* key in :attr:`~RemoteMember.data` if key *name* is in
                   :cvar:`~RemoteMember.ALLOWED_KEYS`, else ``None``.
         """
-        if name in RemoteMember.ALLOWED_KEYS:
-            return self.data.get(name)
+        return self.data.get(name) if name in RemoteMember.ALLOWED_KEYS else None
 
 
 class Leader(NamedTuple):
--- a/patroni/dcs/__init__.py
+++ b/patroni/dcs/__init__.py
@@ -225,7 +225,7 @@ class Member(NamedTuple('Member',
         :param version: modification version of a given member key in a Configuration Store.
         :param name: name of PostgreSQL cluster member.
         :param session: either session id or just ttl in seconds.
-        :param value: JSON encoded string containing arbitrary data i.e. ``conn_url``, ``api_url``, 
+        :param value: JSON encoded string containing arbitrary data i.e. ``conn_url``, ``api_url``,
                       ``xlog_location``, ``state``, ``role``, ``tags``, etc. OR a connection URL
                       starting with ``postgres://``.
 
@@ -538,22 +538,22 @@ class Failover(NamedTuple):
 
     def __len__(self) -> int:
         """Implement ``len`` function capability.
-        
+
         .. note::
            This magic method aids in the evaluation of "emptiness" of a ``Failover`` instance. For example:
-           
+
            >>> failover = Failover.from_node(1, None)
            >>> len(failover)
            0
            >>> assert bool(failover) is False
-           
+
            >>> failover = Failover.from_node(1, {"leader": "cluster_leader"})
            >>> len(failover)
            1
            >>> assert bool(failover) is True
-        
+
            This makes it easier to write ``if cluster.failover`` rather than the longer statement.
-        
+
         """
         return int(bool(self.leader)) + int(bool(self.candidate))
 
@@ -1351,7 +1351,8 @@ class AbstractDCS(abc.ABC):
 
         :param path: the path in DCS where to load Cluster(s) from.
 
-        :returns: all Citus groups as :class:`dict`, with group IDs as keys and :class:`Cluster` objects as values or a :class:`Cluster` object representing the coordinator with filled `Cluster.workers` attribute.
+        :returns: all Citus groups as :class:`dict`, with group IDs as keys and :class:`Cluster` objects as values or a
+                  :class:`Cluster` object representing the coordinator with filled `Cluster.workers` attribute.
         """
 
     @abc.abstractmethod
@@ -1481,7 +1482,7 @@ class AbstractDCS(abc.ABC):
 
         .. note::
             This method abstracts away the required data structure of :meth:`~Cluster.write_status`, so it
-            is not needed in the caller. However, the ``optime/leader`` is only written in 
+            is not needed in the caller. However, the ``optime/leader`` is only written in
             :meth:`~Cluster.write_status` when the cluster has members with a Patroni version that
             is old enough to require it (i.e. the old Patroni version doesn't understand the new format).
 
@@ -1718,7 +1719,6 @@ class AbstractDCS(abc.ABC):
     @staticmethod
     def sync_state(leader: Optional[str], sync_standby: Optional[Collection[str]]) -> Dict[str, Any]:
         """Build ``sync_state`` dictionary.
-
 
         :param leader: name of the leader node that manages ``/sync`` key.
         :param sync_standby: collection of currently known synchronous standby node names.
--- a/patroni/dcs/__init__.py
+++ b/patroni/dcs/__init__.py
@@ -540,7 +540,7 @@ class Failover(NamedTuple):
         """Implement ``len`` function capability.
 
         .. note::
-           This magic method aids in the evaluation of "emptiness" of a ``Failover`` instance. For example:
+           This magic method aids in the evaluation of "emptiness" of a :class:`Failover` instance. For example:
 
            >>> failover = Failover.from_node(1, None)
            >>> len(failover)
@@ -1361,7 +1361,7 @@ class AbstractDCS(abc.ABC):
 
     @abc.abstractmethod
     def set_retry_timeout(self, retry_timeout: int) -> None:
-        """Set the new value for retry_timeout."""
+        """Set the new value for *retry_timeout*."""
 
     def _set_loop_wait(self, loop_wait: int) -> None:
         """Set new *loop_wait* value.
@@ -1548,7 +1548,7 @@ class AbstractDCS(abc.ABC):
 
     @abc.abstractmethod
     def _write_status(self, value: str) -> bool:
-        """Write current WAL LSN and confirmed_flush_lsn of permanent slots into the ``/status`` key in DCS.
+        """Write current WAL LSN and ``confirmed_flush_lsn`` of permanent slots into the ``/status`` key in DCS.
 
         :param value: status serialized in JSON format.
 
--- a/patroni/dcs/__init__.py
+++ b/patroni/dcs/__init__.py
@@ -835,7 +835,7 @@ class Cluster(NamedTuple('Cluster',
         """
         return all((self.initialize is None, self.config is None, self.leader is None, self.last_lsn == 0,
                     self.members == [], self.failover is None, self.sync.version is None,
-                    self.history is None, self.slots is None, self.failsafe is None, not self.workers))
+                    self.history is None, self.slots is None, self.failsafe is None, self.workers == {}))
 
     def __len__(self) -> int:
         """Implement ``len`` function capability.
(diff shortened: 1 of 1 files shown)


patroni/dcs/__init__.py
```diff
@@ -1156,106 +1570,158 @@ def manual_failover(self, leader: Optional[str], candidate: Optional[str],
...
     @staticmethod
     def sync_state(leader: Optional[str], sync_standby: Optional[Collection[str]]) -> Dict[str, Any]:
-        """Build sync_state dict.
-        The sync_standby key being kept for backward compatibility.
-        :param leader: name of the leader node that manages /sync key
-        :param sync_standby: collection of currently known synchronous standby node names
-        :returns: dictionary that later could be serialized to JSON or saved directly to DCS
+        """Build ``sync_state`` dictionary.
+
+        The *sync_standby* key is kept for backward compatibility.
```
CyberDem0n, line 1661: I think we wanted to say that we don't use plural (`sync_standbys`) for backward compatibility.
```suggestion
```

patroni/dcs/__init__.py
```diff
@@ -172,28 +193,46 @@ def get_dcs(config: Union['Config', Dict[str, Any]]) -> 'AbstractDCS':
 
 class Member(NamedTuple):
     """Immutable object (namedtuple) which represents single member of PostgreSQL cluster.
+
+    .. note::
+        There are two mandatory keys in a data:
+
+        ``conn_url``: connection string containing host, user and password which could be used to access this member.
+        ``api_url``: REST API url of patroni instance
```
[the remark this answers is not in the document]
matthbakeredb: "Noteworthy"?

patroni/dcs/__init__.py
```diff
@@ -172,28 +193,46 @@ def get_dcs(config: Union['Config', Dict[str, Any]]) -> 'AbstractDCS':
...
-        >>> Member.from_node(-1, '', '', '{"conn_url": "postgres://foo@bar/postgres"}') is not None
-        True
-        >>> Member.from_node(-1, '', '', '{')
-        Member(version=-1, name='', session='', data={})
+        """Factory method for instantiating :class:`Member` from a JSON serialised string or object.
+
+        :param version: modification version of a given member key in a Configuration Store.
+        :param name: name of PostgreSQL cluster member.
+        :param session: either session id or just ttl in seconds.
+        :param value: dictionary containing arbitrary data i.e. ``conn_url``, ``api_url``, ``xlog_location``, ``state``,
```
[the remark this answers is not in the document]
matthbakeredb: Hmmn right, I remember looking at that code too. I'll change it.

patroni/dcs/__init__.py
```diff
@@ -172,28 +193,46 @@ def get_dcs(config: Union['Config', Dict[str, Any]]) -> 'AbstractDCS':
...
+            ``role``, ``tags``, etc...
```
matthbakeredb, line 225:
```suggestion
        :param value: JSON encoded string containing arbitrary data i.e. ``conn_url``, ``api_url``, 
                      ``xlog_location``, ``state``, ``role``, ``tags``, etc. OR a connection URL
                      starting with ``postgres://``.
```

patroni/dcs/__init__.py
```diff
@@ -302,84 +372,145 @@ class RemoteMember(Member):
...
         return self.member.conn_url
 
     @property
     def data(self) -> Dict[str, Any]:
+        """Data value of the :class:`Member` instance."""
         return self.member.data
 
     @property
     def timeline(self) -> Optional[int]:
+        """Timeline value of :attr:`~Leader.data`."""
```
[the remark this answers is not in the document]
matthbakeredb: Yes that might be copypasta
```suggestion
        """Timeline value of :attr:`~Member.data`."""
```

patroni/dcs/__init__.py
```diff
@@ -302,84 +372,145 @@ class RemoteMember(Member):
...
 
     @property
     def checkpoint_after_promote(self) -> Optional[bool]:
-        """
-        >>> Leader(1, '', Member.from_node(1, '', '', '{"version":"z"}')).checkpoint_after_promote
+        """Determine whether a checkpoint should be made for this leader after promotion.
+
+        :returns: ``True`` if the role is ``master`` or ``primary`` and ``checkpoint_after_promote`` is not set,
+                 ``False`` if not a ``master`` or ``primary`` or if the checkpoint has already been made.
+                 If the version of Patroni is older than 1.5.6, return ``None``.
```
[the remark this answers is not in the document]
matthbakeredb: So it's a truth check? checkpoint **has** occurred?
```suggestion
        """Determine whether a checkpoint has been made for this leader after promotion.

        :returns: ``True`` if the role is ``master`` or ``primary`` and ``checkpoint_after_promote`` is not set,
                 ``False`` if not a ``master`` or ``primary`` or if the checkpoint hasn't been made.
                 If the version of Patroni is older than 1.5.6, return ``None``.
```

patroni/dcs/__init__.py
```diff
@@ -585,50 +795,87 @@ class Cluster(NamedTuple):
...
+        """Produce an empty :class:`Cluster` instance."""
         return Cluster(None, None, None, 0, [], None, SyncState.empty(), None, None, None)
 
     def is_empty(self):
-        return self.initialize is None and self.config is None and self.leader is None and self.last_lsn == 0\
-            and self.members == [] and self.failover is None and self.sync.version is None\
-            and self.history is None and self.slots is None and self.failsafe is None and self.workers == {}
+        """Validate definition of all attributes of this :class:`Cluster` instance.
+
+        :returns: ``True`` if the current :class:`Cluster` has unpopulated attributes.
```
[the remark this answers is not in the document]
matthbakeredb: See my suggestion https://github.com/zalando/patroni/pull/2777/files#r1276421207

patroni/dcs/__init__.py
```diff
@@ -585,50 +795,87 @@ class Cluster(NamedTuple):
...
+        """
+        return all((self.initialize is None, self.config is None, self.leader is None, self.last_lsn == 0,
+                    self.members == [], self.failover is None, self.sync.version is None,
+                    self.history is None, self.slots is None, self.failsafe is None, not self.workers))
 
     def __len__(self) -> int:
+        """Implement ``len`` function capability."""
```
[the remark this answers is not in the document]
matthbakeredb: Ah that makes sense, so this is more of a "is defined" test? `len` doesn't sound right for that, in which case yes I agree.

Would you want to make that a separate changeset/PR?

patroni/dcs/__init__.py
```diff
@@ -800,25 +1062,48 @@ def should_enforce_hot_standby_feedback(self, my_name: str, nofailover: bool, ma
...
-        0
-        >>> Cluster(0, 0, 0, 0, 0, 0, 0, TimelineHistory.from_node(1, '[]'), 0, None).timeline
-        1
-        >>> Cluster(0, 0, 0, 0, 0, 0, 0, TimelineHistory.from_node(1, '[["a"]]'), 0, None).timeline
-        0
+        """Get the cluster history index from the :attr:`~Cluster.history`.
+
+        :returns: If the recorded history is empty assume timeline is ``1``, if it is not defined or the stored history
+                  is not formatted as expected ``0`` is returned and an error will be logged.
+                  Otherwise, the first number stored (indexed starting at 1) is returned.
```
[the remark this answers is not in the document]
matthbakeredb: Ok, thanks for explaining.

patroni/dcs/__init__.py
```diff
@@ -850,6 +1143,7 @@ def wrapper(*args: Any, **kwargs: Any):
 
 
 class AbstractDCS(abc.ABC):
+    """Abstract representation of DCS for preparing and loading :class:`Cluster` objects."""
```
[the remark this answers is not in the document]
matthbakeredb: OK, I'll have a look at this and flesh it out a bit.

patroni/dcs/__init__.py
```diff
@@ -1006,15 +1341,28 @@ def __get_patroni_cluster(self, path: Optional[str] = None) -> Cluster:
         return cluster
 
     def is_citus_coordinator(self) -> bool:
+        """:class:`Cluster` instance has a Citus Coordinator group ID.
+
+        :returns: ``True`` if this object's Citus group ID matches the default Citus Coordinator group ID.
```
matthbakeredb, line 1346:
```suggestion
        :returns: ``True`` if the given node is running as Citus Coordinator (``group=0``).
```

patroni/dcs/__init__.py
```diff
@@ -1006,15 +1341,28 @@ def __get_patroni_cluster(self, path: Optional[str] = None) -> Cluster:
...
+        """
         return self._citus_group == str(CITUS_COORDINATOR_GROUP_ID)
 
     def get_citus_coordinator(self) -> Optional[Cluster]:
+        """Load the cluster for the Citus Coordinator node.
+
```
matthbakeredb, line 1352:
```suggestion

          .. note::
              This method is only executed on the worker nodes (``group!=0``) to find the coordinator.

```

patroni/dcs/__init__.py
```diff
@@ -1006,15 +1341,28 @@ def __get_patroni_cluster(self, path: Optional[str] = None) -> Cluster:
         return cluster
 
     def is_citus_coordinator(self) -> bool:
+        """:class:`Cluster` instance has a Citus Coordinator group ID.
+
+        :returns: ``True`` if this object's Citus group ID matches the default Citus Coordinator group ID.
+        """
```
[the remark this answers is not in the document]
matthbakeredb: See https://github.com/zalando/patroni/pull/2777/files#r1276434611

patroni/dcs/__init__.py
```diff
@@ -1024,6 +1372,17 @@ def _get_citus_cluster(self) -> Cluster:
         return cluster
 
     def get_cluster(self, force: bool = False) -> Cluster:
+        """Retrieve an appropriate cached or fresh view of DCS.
```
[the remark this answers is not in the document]
matthbakeredb: Ah ok, I added an extra note comment https://github.com/zalando/patroni/pull/2777/files#r1276442461

patroni/dcs/__init__.py
```diff
@@ -1043,32 +1402,51 @@ def get_cluster(self, force: bool = False) -> Cluster:
...
 
-        :param last_lsn: absolute WAL LSN in bytes
-        :returns: `!True` on success."""
+        :param last_lsn: absolute WAL LSN in bytes.
+
+        :returns: ``True`` if successfully committed to DCS.
+        """
 
     def write_leader_optime(self, last_lsn: int) -> None:
+        """Write value for WAL LSN to ``optime/leader`` key in DCS.
```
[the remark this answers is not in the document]
matthbakeredb: So perhaps a note to say this is only used on older Patroni versions?

(When I saw that before I did wonder why you need to keep that around. Is it just for clusters upgrading from old versions?)

patroni/dcs/__init__.py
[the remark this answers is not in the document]
CyberDem0n: Yes, this method returns `True` when checkpoint after promote has occurred on the given primary node.

patroni/dcs/__init__.py
```diff
@@ -402,21 +533,37 @@ def from_node(version: _Version, value: Union[str, Dict[str, str]]) -> 'Failover
         return Failover(version, data.get('leader'), data.get('member'), data.get('scheduled_at'))
 
     def __len__(self) -> int:
+        """Implement ``len`` function capability."""
```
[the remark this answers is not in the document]
CyberDem0n: I think it would be worth to mention that this method is called when something like `bool(cluster.failover)` or `if cluster.failover` is executed.

The reason it was done this way because Patroni need to create/update/delete failover key atomically, but ZooKeeper implements atomic operations it very weirdly. In ZooKeeper these operations are based on version of a given ZNode, which is reset to 0 every time when ZNode is dropper and created. That is, we have a /failover key with version=0 and want to remove it, but meanwhile some other thread removed and added it back. As a result we will be removing not a key we wanted to remove.

To mitigate this behavior Patroni just empties the /failover key instead of removing it. When we read the cluster state from DCS, the Failover object will be created, because we need to know its version for the next update.

Without `__len__` we would have to write something like `if cluster.failover and (cluster.failover.leader or cluster.failover.candidate)` most of the time, and now we can just do `if cluster.failover`, which will be evaluated to `False` if object is "empty".

patroni/dcs/__init__.py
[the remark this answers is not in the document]
CyberDem0n: No. ZooKeeper implements caching inside zookeeper.py.

`self._cluster = cluster` -- this cache exists mainly for REST API, in order to not put additional stress on DCS while processing health-check requests.

patroni/dcs/__init__.py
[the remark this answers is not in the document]
CyberDem0n: Actually, this method is still called from ha.py on demote or on Patroni shutdown (which also demotes postgres).
It was kept in order to avoid calling write_status() directly, because write_status() requires dict object to be passed and I wanted to abstract it away.

But In any case it first calls write_status() method and only the last one decides whether it should also update /optime/leader

patroni/dcs/__init__.py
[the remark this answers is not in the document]
CyberDem0n: When Patroni updates the member key it always writes conn_url and api_url, but what if the key was updated by someone else? We want to be resilient for such unexpected scenarios.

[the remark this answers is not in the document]
matthbakeredb: I don't understand this. So `Member` is read from DCS during `dcs.touch_member`, and if it's different to `data` passed into it, then the data constructed in `ha.touch_member` is written (looking at the etcd3 implementation). The values for each of these 2 parameters are taken from `state_handler.connection_string` and `api.connection_string` in the `ha.touch_member`. I don't see where the existing data held in DCS is kept/merged.

Do you mean someone who wrote to the DCS `/member/member` key outside of Patroni? Who would do that? And why would that be considered something to honour?

I am thinking that for this comment it's worth mentioning that those keys are assembled from active connection information when data is written to DCS.

Also, why doesn't touch_member create a `Member` instance instead of constructing a `Dict`?

Thanks for any pointers!

patroni/dcs/__init__.py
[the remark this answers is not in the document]
matthbakeredb: Thanks that makes sense!

patroni/dcs/__init__.py
```diff
@@ -402,21 +534,37 @@ def from_node(version: _Version, value: Union[str, Dict[str, str]]) -> 'Failover
...
```
matthbakeredb, line 537:
```suggestion
        """Implement ``len`` function capability.
        
        .. note::
           This magic method aids in the evaluation of "emptiness" of a ``Failover`` instance. For example:
           
           >>> failover = Failover.from_node(1, None)
           >>> len(failover)
           0
           >>> assert bool(failover) is False
           
           >>> failover = Failover.from_node(1, {"leader": "cluster_leader"})
           >>> len(failover)
           1
           >>> assert bool(failover) is True
        
           This makes it easier to write ``if cluster.failover`` rather than the longer statement.
        
        """
```

patroni/dcs/__init__.py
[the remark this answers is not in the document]
matthbakeredb: See suggestion https://github.com/zalando/patroni/pull/2777/files#r1279176340

patroni/dcs/__init__.py
```diff
@@ -585,50 +796,87 @@ class Cluster(NamedTuple):
...
+        """Validate definition of all attributes of this :class:`Cluster` instance.
+
+        :returns: ``True`` if the current :class:`Cluster` has unpopulated attributes.
+        """
+        return all((self.initialize is None, self.config is None, self.leader is None, self.last_lsn == 0,
+                    self.members == [], self.failover is None, self.sync.version is None,
+                    self.history is None, self.slots is None, self.failsafe is None, not self.workers))
 
     def __len__(self) -> int:
+        """Implement ``len`` function capability."""
```
matthbakeredb, line 812:
```suggestion
        """Implement ``len`` function capability.

        .. note::
           This magic method aids in the evaluation of "emptiness" of a ``Cluster`` instance. For example:

           >>> cluster = Cluster.empty()
           >>> len(cluster)
           0
           >>> assert bool(cluster) is False

           >>> cluster = Cluster(None, None, None, 0, [1, 2, 3], None, SyncState.empty(), None, None, None)
           >>> len(cluster)
           1
           >>> assert bool(cluster) is True

           This makes it easier to write ``if cluster`` rather than the longer statement.

        """
```

patroni/dcs/__init__.py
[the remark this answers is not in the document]
matthbakeredb: As per the other suggestion, see this one for `Cluster` https://github.com/zalando/patroni/pull/2777/files#r1279186080

patroni/dcs/__init__.py
CyberDem0n: Looks good

patroni/dcs/__init__.py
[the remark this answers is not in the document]
CyberDem0n: > So Member is read from DCS during dcs.touch_member

No, touch_member() write member key. Reading happens somewhere inside in get_cluster().

> and if it's different to data passed into it, then the data constructed in ha.touch_member is written (looking at the etcd3 implementation)

This is unrelated and just implementation detail of very specific DCS, because the member key is connected with the session and will not expire while the session exists. Therefore we can skip the update if value didn't changed.
But, with for example etcd v2 the member key is updated every time.

> Do you mean someone who wrote to the DCS /member/member key outside of Patroni? Who would do that? And why would that be considered something to honour?

Exactly, something different from Patroni could change arbitrary keys in DCS and we better be resilient to it.

> Also, why doesn't touch_member create a Member instance instead of constructing a Dict?

Because touch_member() only updates one key that represents current node in DCS. When we read the cluster state we are reading **all** member keys, which happens on all nodes.

patroni/dcs/__init__.py
```diff
@@ -1024,6 +1376,18 @@ def _get_citus_cluster(self) -> Cluster:
         return cluster
 
     def get_cluster(self, force: bool = False) -> Cluster:
+        """Retrieve an appropriate cached or fresh view of DCS.
+
+        .. note::
+            Stores copy of time, status and failsafe values for comparison in DCS update decisions.
+            This works around the unique operational requirement of Zookeeper to make extra queries for keys.
```
matthbakeredb, line 1383:
```suggestion
            Caching is mostly required to avoid overhead placed on the REST API, due to the nature of how
            Zookeeper makes multiple requests are made in comparison to other DCS implementations.
```

patroni/dcs/__init__.py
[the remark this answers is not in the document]
matthbakeredb: Um OK, so that suggestion is incorrect? Confused... how about this? https://github.com/zalando/patroni/pull/2777/files#r1279198612

patroni/dcs/__init__.py
```diff
@@ -1043,32 +1407,51 @@ def get_cluster(self, force: bool = False) -> Cluster:
...
+        :param last_lsn: absolute WAL LSN in bytes.
+
+        :returns: ``True`` if successfully committed to DCS.
+        """
 
     def write_leader_optime(self, last_lsn: int) -> None:
+        """Write value for WAL LSN to ``optime/leader`` key in DCS.
+
+        :param last_lsn: absolute WAL LSN in bytes.
+        """
```
matthbakeredb, line 1433:
```suggestion
        """Write value for WAL LSN to ``optime/leader`` key in DCS.

        .. note::
            This method abstracts away the required data structure of :meth:`~Cluster.write_status`, so it
            is not needed in the caller. However, the ``optime`` are only written in :meth:`~Cluster.write_status`
            when the member Patroni version is old enough to require it.

        :param last_lsn: absolute WAL LSN in bytes.
        """
```

patroni/dcs/__init__.py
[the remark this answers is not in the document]
matthbakeredb: Does this addition make sense? https://github.com/zalando/patroni/pull/2777/files#r1279211995

patroni/dcs/__init__.py
CyberDem0n: I wouldn't mention Zookeeper or any other DCS at all here.

TL;DR - there are two levels of caches:
1. `self._cluster` which is accessed via `cluster` property from REST API and cached only for `ttl` seconds.
2. sometimes DCS are have their own caches:
  - zookeeper does it for the entire Cluster
  - K8s and Etcd3 have caches for all keys associated with the cluster

The second type of caches exists to avoid reading something from DCS on every heart beat (when it is possible), and we shouldn't reveal here any implementation details.

patroni/dcs/__init__.py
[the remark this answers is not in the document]
CyberDem0n: > How are we resilient to it? 

Until Patroni on respective node overwrites it.
That is, we don't want to suddenly crash because can't deserialize something.
But, we still create the Member object because the node that performs update of the member key sometimes should know that the key exists and should be removed before creating a new one.
For example in Zookeeper and Consul keys could be bound to a session and it is not possible (or there is no reliable way of doing) to change the session for the key.

patroni/dcs/__init__.py
```diff
@@ -585,50 +813,106 @@ class Cluster(NamedTuple):
 
     @staticmethod
     def empty() -> 'Cluster':
-        return Cluster(None, None, None, 0, [], None, SyncState.empty(), None, None, None)
+        """Produce an empty :class:`Cluster` instance."""
+        return Cluster(None, None, None, 0, [], None, SyncState.empty(), None, None, None, {})
```
matthbakeredb, line 817: @CyberDem0n looks like there was a bug here? It was preventing the doctest from passing because some Citus tests write to cluster.workers. Should I create a separate PR for this?

matthbakeredb: Bug is there are in fact 11 args to `Cluster` not 10...

matthbakeredb: Hmmn seems to have broken tests, which passed locally. :thinking: 

https://github.com/zalando/patroni/actions/runs/5717044205/job/15489954467

CyberDem0n: Yeah, with workers it is an interesting corner case.
It has a default value `{}`, which is mutated :(
The problem is introduced in #2652, when I changed `namedtuple` with overridden `__new__` method to `NamedTuple`.
I'll open a separate PR with the fix.

patroni/dcs/__init__.py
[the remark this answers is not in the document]
matthbakeredb: Thank you!

patroni/dcs/__init__.py
matthbakeredb: Updated it to a simpler form. Thanks for the explanation.

patroni/dcs/__init__.py
matthbakeredb: Ah, thanks for taking that. I'll rebase when it's merged.

patroni/dcs/__init__.py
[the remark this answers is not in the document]
matthbakeredb: Ah right, got it. So you're referring to the read of DCS being resilient to changes UNTIL we overwrite it with sane values.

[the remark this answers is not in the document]
CyberDem0n: Anything is possible. Maybe the broken member key will never be updated, we just need to make sure that it will not affect normal cluster operations. For example we don't want to create a replication slot for it.

patroni/dcs/__init__.py
CyberDem0n: :+1:

patroni/dcs/__init__.py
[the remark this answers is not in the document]
matthbakeredb: How about this:

```suggestion
    .. note::
        These two keys in data are always written to the DCS, but care is taken to maintain consistency and resilience
        from data that is read:
        
        ``conn_url``: connection string containing host, user and password which could be used to access this member.
        ``api_url``: REST API url of patroni instance
```

patroni/dcs/__init__.py
```diff
@@ -898,6 +1281,12 @@ def __init__(self, config: Dict[str, Any]) -> None:
         self.event = Event()
 
     def client_path(self, path: str) -> str:
```
CyberDem0n, line 1283: Yeah, now the *path* word is very confusing in this context. Originally we used `path` only because of actual hierarchical structures in Etcd v2 and Zookeeper.

It is all about building key names in DCS from the following information:
- namespace (key prefix)
- scope (cluster name)
- key name within the scope

At the end we get something like the following.
1. for etcd, etcd3, and raft:
```
/{namespace}/{scope}/config
/{namespace}/{scope}/leader
/{namespace}/{scope}/members/{my_name}
...
```

2. for consul and zookeeper:
```
{namespace}/{scope}/config
{namespace}/{scope}/leader
{namespace}/{scope}/members/{my_name}
...
```

3. something totally different for K8s (all K8s objects should be created in the same namespace that matches `kuberneted.namespace` parameter):
```
{scope}-leader  # leader configmap
{scope}  # leader endpoint

{scope}-{config}  # config key configmap or endpoint
...
```
Pods (that map to members) may have arbitrary names and identified by their labels.


Now the word `key` would be more appropriate. But we can do it separately.

patroni/dcs/__init__.py
```diff
@@ -864,6 +1190,62 @@ def wrapper(*args: Any, **kwargs: Any):
...
+    Functional methods that perform construction of complex data objects:
+
+        * :meth:`~AbstractDCS._cluster_loader`:
+            method which processes the structure of data stored in the DCS used to build the :class:`Cluster` object
+            with all relevant associated data.
+        * :meth:`~AbstractDCS._citus_cluster_loader`:
+            Similar to above but specifically representing Citus group and workers information.
+        * :meth:`~AbstractDCS._load_cluster`:
+            main method for calling specific ``loader`` method to build the :class:`Cluster` object representing the
+            state and topology of the cluster.
```
CyberDem0n, line 1207: They all time critical. If didn't succeed during `retry_timeout` the primary is demoted because of lack of access to DCS.

matthbakeredb: Thanks for these comments on this section. I'll bundle them together in one shuffle commit.

patroni/dcs/__init__.py
```diff
@@ -302,84 +374,145 @@ class RemoteMember(Member):
 
     @classmethod
     def from_name_and_data(cls, name: str, data: Dict[str, Any]) -> 'RemoteMember':
+        """Factory method to construct instance from given *name* and *data*.
+
+        :param name: name of the remote member.
+        :param data: dictionary of member information.
```
barthisrael, line 380: What do we expecte `data` to contain?

patroni/dcs/__init__.py
```diff
@@ -402,21 +535,54 @@ def from_node(version: _Version, value: Union[str, Dict[str, str]]) -> 'Failover
         return Failover(version, data.get('leader'), data.get('member'), data.get('scheduled_at'))
 
     def __len__(self) -> int:
+        """Implement ``len`` function capability.
+
+        .. note::
+           This magic method aids in the evaluation of "emptiness" of a ``Failover`` instance. For example:
```
barthisrael, line 541:
```suggestion
           This magic method aids in the evaluation of "emptiness" of a :class:`Failover` instance. For example:
```

patroni/dcs/__init__.py
```diff
@@ -527,25 +732,45 @@ def matches(self, name: Optional[str], check_leader: bool = False) -> bool:
...
     @staticmethod
     def from_node(version: _Version, value: str) -> 'TimelineHistory':
-        """
-        >>> h = TimelineHistory.from_node(1, 2)
-        >>> h.lines
-        []
+        """Parse the given JSON serialized string as a list of timeline history lines.
+
+        :param version: version number
+        :param value: JSON serialized string.
```
barthisrael, line 762: Any expected content?

patroni/dcs/__init__.py
```diff
@@ -906,112 +1295,147 @@ def client_path(self, path: str) -> str:
...
     @property
     @abc.abstractmethod
     def ttl(self) -> int:
-        """Get new ttl value"""
+        """Get current ``ttl`` value."""
 
     @abc.abstractmethod
     def set_retry_timeout(self, retry_timeout: int) -> None:
-        """Set the new value for retry_timeout"""
+        """Set the new value for retry_timeout."""
```
barthisrael, line 1362:
```suggestion
        """Set the new value for *retry_timeout*."""
```

patroni/dcs/__init__.py
```diff
@@ -1059,32 +1512,57 @@ def get_cluster(self, force: bool = False) -> Cluster:
...
+            is old enough to require it (i.e. the old Patroni version doesn't understand the new format).
+
+        :param last_lsn: absolute WAL LSN in bytes.
+        """
         self.write_status({self._OPTIME: last_lsn})
 
     @abc.abstractmethod
     def _write_status(self, value: str) -> bool:
-        """write current WAL LSN and confirmed_flush_lsn of permanent slots into the `/status` key in DCS
+        """Write current WAL LSN and confirmed_flush_lsn of permanent slots into the ``/status`` key in DCS.
```
barthisrael, line 1549:
```suggestion
        """Write current WAL LSN and ``confirmed_flush_lsn`` of permanent slots into the ``/status`` key in DCS.
```

patroni/dcs/__init__.py
```diff
@@ -1097,38 +1575,55 @@ def write_status(self, value: Dict[str, Any]) -> None:
...
+        :param value: failsafe topology serialized in JSON format.
+
+        :returns: ``True`` if successfully committed to DCS.
+        """
 
     def write_failsafe(self, value: Dict[str, str]) -> None:
-        if not (isinstance(self._last_failsafe, dict) and deep_compare(self._last_failsafe, value))\
+        """Write the ``/failsafe`` key in DCS.
+
+        :param value: dictionary value to set.
```
barthisrael, line 1586: Any expected content?

matthbakeredb: Updated

patroni/dcs/__init__.py
matthbakeredb: updated

patroni/dcs/__init__.py
matthbakeredb: Updated

CyberDem0n:
After release there are now conflicts.

matthbakeredb:
> After release there are now conflicts.

Rebased

hughcapet:
Decreased coverage here as well:(
https://coveralls.io/builds/61766756/source?filename=patroni%2Fdcs%2F__init__.py#L399

tests/test_postgresql.py
```diff
@@ -347,6 +347,7 @@ def test_write_postgresql_and_sanitize_auto_conf(self):
     def test_follow(self):
         self.p.call_nowait(CallbackAction.ON_START)
         m = RemoteMember('1', {'restore_command': '2', 'primary_slot_name': 'foo', 'conn_kwargs': {'host': 'bar'}})
+        self.assertEqual(m.disallowed_attr, None)
```
CyberDem0n, line 350: I wonder if declaring `RemoteMember.__getattr__()` as returning Optional[Any] would work better?

matthbakeredb: Do you mean in the test or in the code?

The extra `return None` is not part of typing tests. It comes from https://pylint.readthedocs.io/en/latest/user_guide/messages/refactor/inconsistent-return-statements.html

CyberDem0n: I mean that `return None` is redundant, it is returned implicitly.

matthbakeredb: Or

```
    def __getattr__(self, name: str) -> Any:
        """Dictionary style key lookup.

        :param name: key to lookup.

        :returns: value of *name* key in :attr:`~RemoteMember.data` if key *name* is in
                  :cvar:`~RemoteMember.ALLOWED_KEYS`, else ``None``.
        """
        return self.data.get(name) if name in RemoteMember.ALLOWED_KEYS else None
```

matthbakeredb: Removed both `return None` and test.

hughcapet:
👍

CyberDem0n:
:+1:

PR #2778, opened 2023-07-26
Add docstrings to `patroni.postgresql.slots.py`

matthbakeredb:
Also, made some small code changes to satisfy formatting and pylint.

### DIFF
--- a/patroni/postgresql/slots.py
+++ b/patroni/postgresql/slots.py
@@ -252,12 +252,11 @@ class SlotsHandler:
 
         .. note::
             Only supported from PostgreSQL version 9.4 onwards.
-            Runs when scheduled by HA loop.
 
         Store replication slot ``name``, ``type``, ``plugin``, ``database`` and ``datoid``.
         If PostgreSQL version is 10 or newer also store ``catalog_xmin`` and ``confirmed_flush_lsn``.
 
-        When using logical slots, store information separately for slot synchronisation (copying) on replica nodes.
+        When using logical slots, store information separately for slot synchronisation  on replica nodes.
         """
         if self._postgresql.major_version >= 90400 and self._schedule_load_slots:
             replication_slots: Dict[str, Dict[str, Any]] = {}
@@ -395,7 +394,7 @@ class SlotsHandler:
             yield cur
 
     def _ensure_logical_slots_primary(self, slots: Dict[str, Any]) -> None:
-        """Create any missing logical replication *slots*.
+        """Create any missing logical replication *slots* on the primary.
 
         If the logical slot already exists, copy state information into the replication slots structure stored in the
         class instance.
@@ -460,8 +459,9 @@ class SlotsHandler:
         advance_slots: Dict[str, Dict[str, int]] = defaultdict(dict)
         create_slots: List[str] = []  # Collect logical slots to be created on the replica
 
-        logical_slots = {name: value for name, value in slots.items() if value['type'] == 'logical'}
-        for name, value in logical_slots.items():
+        for name, value in slots.items():
+            if value['type'] != 'logical':
+                continue
 
             # If the logical already exists, copy some information about it into the original structure
             if self._replication_slots.get(name, {}).get('datoid'):
--- a/patroni/postgresql/slots.py
+++ b/patroni/postgresql/slots.py
@@ -1,11 +1,15 @@
+"""Replication slot handling.
+
+Provides classes for the creation, monitoring, management and synchronisation of PostgreSQL replication slots.
+"""
+
 import logging
 import os
 import shutil
-
 from collections import defaultdict
 from contextlib import contextmanager
 from threading import Condition, Thread
-from typing import Any, Dict, Iterator, List, Optional, Union, Tuple, TYPE_CHECKING
+from typing import Any, Dict, Iterator, List, Optional, Union, Tuple, TYPE_CHECKING, Collection
 
 from .connection import get_connection_cursor
 from .misc import format_lsn, fsync_dir
@@ -43,9 +47,19 @@ def compare_slots(s1: Dict[str, Any], s2: Dict[str, Any], dbid: str = 'database'
 
 
 class SlotsAdvanceThread(Thread):
+    """Daemon process :class:``Thread`` object for advancing logical replication slots on replicas.
+
+    This ensures that slot advancing queries sent to postgres do not block the main loop and cause leader lock expiry.
+
+    :ivar daemon: Enable background long-lived process thread.
+    """
 
     def __init__(self, slots_handler: 'SlotsHandler') -> None:
-        super(SlotsAdvanceThread, self).__init__()
+        """Create a new thread for handling slot advance queries.
+
+        :param slots_handler: The calling class instance for reference to slot information attributes.
+        """
+        super().__init__()
         self.daemon = True
         self._slots_handler = slots_handler
 
@@ -59,6 +73,13 @@ class SlotsAdvanceThread(Thread):
         self.start()
 
     def sync_slot(self, cur: Union['cursor', 'Cursor[Any]'], database: str, slot: str, lsn: int) -> None:
+        """Execute a ``pg_replication_slot_advance`` query and store success for scheduled synchronisation task.
+
+        :param cur: database connection cursor.
+        :param database: name of the database associated with the slot.
+        :param slot: name of the slot to be synchronised.
+        :param lsn: last known LSN position
+        """
         failed = copy = False
         try:
             cur.execute("SELECT pg_catalog.pg_replication_slot_advance(%s, %s)", (slot, format_lsn(lsn)))
@@ -80,6 +101,11 @@ class SlotsAdvanceThread(Thread):
                     self._scheduled.pop(database)
 
     def sync_slots_in_database(self, database: str, slots: List[str]) -> None:
+        """Synchronise slots for a single database.
+
+        :param database: name of the database.
+        :param slots: list of slot names to synchronise.
+        """
         with self._slots_handler.get_local_connection_cursor(dbname=database, options='-c statement_timeout=0') as cur:
             for slot in slots:
                 with self._condition:
@@ -88,6 +114,7 @@ class SlotsAdvanceThread(Thread):
                     self.sync_slot(cur, database, slot, lsn)
 
     def sync_slots(self) -> None:
+        """Synchronise slots for all scheduled databases."""
         with self._condition:
             databases = list(self._scheduled.keys())
         for database in databases:
@@ -100,6 +127,12 @@ class SlotsAdvanceThread(Thread):
                     logger.error('Failed to advance replication slots in database %s: %r', database, e)
 
     def run(self) -> None:
+        """Thread main loop entrypoint.
+
+        .. note::
+            Thread will wait until a sync is scheduled from outside, normally triggered during the HA loop or a wakeup
+            call.
+        """
         while True:
             with self._condition:
                 if not self._scheduled:
@@ -108,6 +141,14 @@ class SlotsAdvanceThread(Thread):
             self.sync_slots()
 
     def schedule(self, advance_slots: Dict[str, Dict[str, int]]) -> Tuple[bool, List[str]]:
+        """Trigger a synchronisation of slots.
+
+        This is the main entrypoint for Patroni HA loop wakeup call.
+
+        :param advance_slots: dictionary containing slots that need to be advanced
+
+        :return: tuple of failure status and a list of slots to be copied
+        """
         with self._condition:
             for database, values in advance_slots.items():
                 self._scheduled[database].update(values)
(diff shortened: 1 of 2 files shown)


patroni/postgresql/slots.py
```diff
@@ -42,9 +46,19 @@ def compare_slots(s1: Dict[str, Any], s2: Dict[str, Any], dbid: str = 'database'
 
 
 class SlotsAdvanceThread(Thread):
+    """Daemon process :class:``Thread`` object for advancing logical replication slots on replicas.
+
+    This ensures that slot advancing queries sent to postgres do not block the main loop and cause leader lock expiry.
+
+    :ivar daemon: Enable background long-lived process thread.
```
CyberDem0n, line 53: `daemon` is a attribute of the `Thread` class, I don't think we should add it to the docstring.
```suggestion
```

patroni/postgresql/slots.py
```diff
@@ -42,9 +46,19 @@ def compare_slots(s1: Dict[str, Any], s2: Dict[str, Any], dbid: str = 'database'
...
+    """
 
     def __init__(self, slots_handler: 'SlotsHandler') -> None:
-        super(SlotsAdvanceThread, self).__init__()
+        """Create a new thread for handling slot advance queries.
```
CyberDem0n, line 57:
```suggestion
        """Create and start a new thread for handling slot advance queries.
```

patroni/postgresql/slots.py
```diff
@@ -135,23 +188,46 @@ def __init__(self, postgresql: 'Postgresql') -> None:
         self.schedule()
 
     def _query(self, sql: str, *params: Any) -> Union['cursor', 'Cursor[Any]']:
+        """Helper method for :meth:`PostgreSQL.query`.
```
CyberDem0n, line 191:
```suggestion
        """Helper method for :meth:`Postgresql.query`.
```

patroni/postgresql/slots.py
```diff
@@ -135,23 +188,46 @@ def __init__(self, postgresql: 'Postgresql') -> None:
...
     @staticmethod
-    def _copy_items(src: Dict[str, Any], dst: Dict[str, Any], keys: Optional[List[str]] = None) -> None:
+    def _copy_items(src: Dict[str, Any], dst: Dict[str, Any], keys: Optional[Collection[str]] = None) -> None:
+        """Select values from *src* dictionary to update in *dst* dictionary for optional supplied *keys*.
+
+        :param src: source dictionary that *keys* will be looked up from.
+        :param dst: destination dictionary to be updated.
+        :param keys: optional list of keys to be looked up in the source dictionary.
+        """
         dst.update({key: src[key] for key in keys or ('datoid', 'catalog_xmin', 'confirmed_flush_lsn')})
```
CyberDem0n, line 208: IIRC, in the previous PR you moved default values to default value of the `keys` parameter.

patroni/postgresql/slots.py
```diff
@@ -343,6 +525,16 @@ def sync_replication_slots(self, cluster: Cluster, nofailover: bool,
 
     @contextmanager
     def _get_leader_connection_cursor(self, leader: Leader) -> Generator[Union['cursor', 'Cursor[Any]'], None, None]:
+        """Create a new database connection to the leader.
+
+        .. note::
+            Uses rewind user credentials.
```
CyberDem0n, line 531:
```suggestion
            Uses rewind user credentials because it has enough permissions to read files from PGDATA.
```

patroni/postgresql/slots.py
matthbakeredb: Yes, I thought about it a bit and realised that if the arg `keys` was an empty list you'd have no `keys` and therefore no copy. That's different to the current behaviour as there's no test for `is None`. If that's not as intended and you should be able to pass an empty list I'll restore the change.

patroni/postgresql/slots.py
```diff
@@ -173,13 +247,24 @@ def process_permanent_slots(self, slots: List[Dict[str, Any]]) -> Dict[str, int]
         return ret
 
     def load_replication_slots(self) -> None:
+        """Query replication slot information from the database and store it for processing by other tasks.
+
+        .. note::
+            Only supported from PostgreSQL version 9.4 onwards.
+            Runs when scheduled by HA loop.
```
CyberDem0n, line 254:
```suggestion
```

Slots are also queried after any database related error.

patroni/postgresql/slots.py
```diff
@@ -173,13 +247,24 @@ def process_permanent_slots(self, slots: List[Dict[str, Any]]) -> Dict[str, int]
...
+
+        Store replication slot ``name``, ``type``, ``plugin``, ``database`` and ``datoid``.
+        If PostgreSQL version is 10 or newer also store ``catalog_xmin`` and ``confirmed_flush_lsn``.
+
+        When using logical slots, store information separately for slot synchronisation (copying) on replica nodes.
```
CyberDem0n, line 259: It is not used for copying.
Logical slots could already exist, but we have to check that logical slot will be safe to use after promotion.
The actual check is executed in the `_ready_logical_slots()` method, which removes keys from `self._logical_slots_processing_queue` when the slot is fine.

patroni/postgresql/slots.py
```diff
@@ -240,31 +357,57 @@ def _drop_incorrect_slots(self, cluster: Cluster, slots: Dict[str, Any], paused:
...
+
+        :yields: connection cursor object, note implementation varies depending on version of :mod:`psycopg`.
+        """
         conn_kwargs = self._postgresql.config.local_connect_kwargs
         conn_kwargs.update(kwargs)
         with get_connection_cursor(**conn_kwargs) as cur:
             yield cur
 
     def _ensure_logical_slots_primary(self, slots: Dict[str, Any]) -> None:
+        """Create any missing logical replication *slots*.
```
CyberDem0n, line 397:
```suggestion
        """Create any missing logical replication *slots* on the primary.
```

patroni/postgresql/slots.py
```diff
@@ -240,31 +357,57 @@ def _drop_incorrect_slots(self, cluster: Cluster, slots: Dict[str, Any], paused:
...
+                                f" WHERE NOT EXISTS (SELECT 1 FROM pg_catalog.pg_replication_slots"
+                                f" WHERE slot_type = 'physical' AND slot_name = %s)",
+                                name, name)
                 except Exception:
                     logger.exception("Failed to create physical replication slot '%s'", name)
                 self._schedule_load_slots = True
 
     @contextmanager
-    def get_local_connection_cursor(self, **kwargs: Any) -> Generator[Union['cursor', 'Cursor[Any]'], None, None]:
+    def get_local_connection_cursor(self, **kwargs: Any) -> Iterator[Union['cursor', 'Cursor[Any]']]:
```
hughcapet, line 380: why did you change the return type here but not here?
https://github.com/zalando/patroni/blob/be8ad58afbfab90b3cc190d541ae3564c6c49554/patroni/postgresql/slots.py#L525

and maybe also here then
https://github.com/zalando/patroni/blob/b7caf3b7f23924711766ef24250d09c4c198b4a1/patroni/postgresql/connection.py#L47

patroni/postgresql/slots.py
```diff
@@ -286,27 +429,50 @@ def _ensure_logical_slots_primary(self, slots: Dict[str, Any]) -> None:
...
-                            if value['confirmed_flush_lsn'] < int(cluster.slots[name]):
-                                advance_slots[value['database']][name] = int(cluster.slots[name])
-                        except Exception as e:
-                            logger.error('Failed to parse "%s": %r', cluster.slots[name], e)
-                elif cluster.slots and name in cluster.slots:  # We want to copy only slots with feedback in a DCS
-                    create_slots.append(name)
+        create_slots: List[str] = []  # Collect logical slots to be created on the replica
+
+        logical_slots = {name: value for name, value in slots.items() if value['type'] == 'logical'}
+        for name, value in logical_slots.items():
```
hughcapet, line 463: maybe instead of adding the new var and an additional iteration over the dict, we can decrease the num of the nested blocks by doing

```
if value['type'] != 'logical':
    continue
```

(if I got your intention correctly)

patroni/postgresql/slots.py
matthbakeredb:
```suggestion
        When using logical slots, store information separately for slot synchronisation  on replica nodes.
```

patroni/postgresql/slots.py
matthbakeredb: > why did you change the return type here but not here?

That was Israel's suggestion, so I missed add it there. I'll fix that.

> and maybe also here then
> 
> https://github.com/zalando/patroni/blob/b7caf3b7f23924711766ef24250d09c4c198b4a1/patroni/postgresql/connection.py#L47

Perhaps in another PR?

patroni/postgresql/slots.py
matthbakeredb:
```suggestion
        for name, value in slots.items():
            if value['type'] != 'logical':
                continue
```

matthbakeredb: Yes, that's much simpler!

patroni/postgresql/slots.py
```diff
@@ -42,9 +46,17 @@ def compare_slots(s1: Dict[str, Any], s2: Dict[str, Any], dbid: str = 'database'
 
 
 class SlotsAdvanceThread(Thread):
+    """Daemon process :class:``Thread`` object for advancing logical replication slots on replicas.
+
+    This ensures that slot advancing queries sent to postgres do not block the main loop and cause leader lock expiry.
```
CyberDem0n, line 51: It runs only in replicas and therefore consequences are different:

0. Replica will not be able to participate in leader race
1. Member key for this node could expire
2. As a result the leader will start making attempts to remove the replication slot.
3. While physical replication is active the physical slot could not be removed (good)
4. but in case if replica for some reason disconnected the physical slot will be dropped (txid_current on the physical slot will stop preventing autovacuum from cleaning up dead tuples in pg_catalog).

```suggestion
    This ensures that slot advancing queries sent to postgres do not block the main loop.
```

patroni/postgresql/slots.py
```diff
@@ -135,23 +186,46 @@ def __init__(self, postgresql: 'Postgresql') -> None:
...
+            about all replication slots that exists on the current host.
+
+            Based on this information perform the following actions:
+
+            1. For the primary we want to expose to DCS permanent logical slots, therefore build (and return) a dict
+               that maps permanent logical slot names to ``confirmed_flush_lsn``.
+            2. detect if one of the previously known permanent slots is missing and schedule resync.
+            3. Update the local cache with the fresh ``catalog_xmin`` and ``confirmed_flush_lsn`` for every known slot.
+
+               This info is used when performing the check of logical slot readiness on standbys.
```
CyberDem0n, line 224: Wrong indentation?
```suggestion
           This info is used when performing the check of logical slot readiness on standbys.
```

patroni/postgresql/slots.py
```diff
@@ -195,16 +279,34 @@ def load_replication_slots(self) -> None:
...
                     return True
         return self._postgresql.citus_handler.ignore_replication_slot(slot)
 
     def drop_replication_slot(self, name: str) -> Tuple[bool, bool]:
-        """Returns a tuple(active, dropped)"""
+        """Drop a named slot from Postgres.
+
+        :param name: name of the slot to be dropped.
+
+        :returns: a tuple of ``active`` and ``dropped``. ``active`` is ``True`` if the slot was active when dropped,
```
CyberDem0n, line 306: Active replication slot could not be dropped.
```suggestion
        :returns: a tuple of ``active`` and ``dropped``. ``active`` is ``True`` if the slot is active,
```

patroni/postgresql/slots.py
```diff
@@ -286,27 +428,51 @@ def _ensure_logical_slots_primary(self, slots: Dict[str, Any]) -> None:
...
+        :param slots: dictionary containing slot information.
+
+        :return: tuple with the result of the scheduling of slot advancement: ``failed`` and list of slots to copy.
+        """
         if not self._advance:
             self._advance = SlotsAdvanceThread(self)
         return self._advance.schedule(slots)
 
     def _ensure_logical_slots_replica(self, cluster: Cluster, slots: Dict[str, Any]) -> List[str]:
+        """Update missing logical *slots* on replicas.
```
CyberDem0n, line 442:
```suggestion
        """Update logical *slots* on replicas.
```

patroni/postgresql/slots.py
hughcapet: > Perhaps in another PR?

Yes, sure. Just hope we will not forget:)

matthbakeredb: Yeah I had already ;-)

https://github.com/zalando/patroni/pull/2799

patroni/postgresql/slots.py
hughcapet: Now we need to cover this continue in the unit tests

CyberDem0n:
There are conflicts, probably dues to Generator->Iterator refactoring.

matthbakeredb:
> There are conflicts, probably dues to Generator->Iterator refactoring.

Updated

matthbakeredb:
> You've probably missed my comment. We need to cover the [added line](https://coveralls.io/builds/61767032/source?filename=patroni%2Fpostgresql%2Fslots.py#L464) in the unit tests

OK thanks, will take a look at that.

patroni/postgresql/slots.py
matthbakeredb: @hughcapet I added a test that should cover the individual case of the filtering out of non-logical slots in that method.

CyberDem0n:
Meanwhile the line remains uncovered: https://coveralls.io/builds/61823939/source?filename=patroni%2Fpostgresql%2Fslots.py#L464

matthbakeredb:
> Meanwhile the line remains uncovered: https://coveralls.io/builds/61823939/source?filename=patroni%2Fpostgresql%2Fslots.py#L464

Gak, I removed too much from my local test! :-1:

hughcapet:
👍

CyberDem0n:
:+1:

PR #2780, opened 2023-07-26
Feature: failover priority

mfpekala-nutanix:
Added support for configuring a `failover_priority` which will prefer certain nodes during failover. Preserves functionality of `failover/nofailover` tag, but also adds the `failover_priority` tag such that:

- During `is_healthiest_node`, a node that sees another node with a higher priority within the maximum lag window will mark itself as unhealthy
- For nodes with equal priority, functionality remains the same (prefer whoever has highest WAL status)
- Only one of `nofailover` or `failover_priority` can be set (`failover_priority: 0` behaves same as `nofailover: True`)

Will address potential edge case where we care about failing over to a specific node (perhaps because of things like better hardware, less lag) but the WAL status of inferior nodes is equal.


patroni/__main__.py
```diff
@@ -81,13 +81,34 @@ def ensure_unique_name(self) -> None:
...
 
     def get_tags(self) -> Dict[str, Any]:
-        return {tag: value for tag, value in self.config.get('tags', {}).items()
-                if tag not in ('clonefrom', 'nofailover', 'noloadbalance', 'nosync') or value}
+        return {
+            tag: value for tag, value in self.config.get('tags', {}).items()
+            if tag not in (
+                'clonefrom',
+                'nofailover',
+                'failover_priority',
```
CyberDem0n, line 89: should it really be a tag?

patroni/config.py
```diff
@@ -601,3 +602,12 @@ def get_global_config(self, cluster: Union[Cluster, None]) -> GlobalConfig:
...
         return get_global_config(cluster, self._dynamic_configuration)
+
+    def _validate_failover_tags(self):
+        """Ensures that only one of `no_failover` and `failover_priority` is set"""
+        has_nofailover_tag = self.get('tags', {}).get('nofailover', None) is not None
+        has_failover_priority_tag = self.get('tags', {}).get('failover_priority', None) is not None
+        if has_nofailover_tag and has_failover_priority_tag:
+            raise ConfigParseError(
+                'Cannot provide values for both `nofailover` and `failover_priority`. Please set only one.'
+            )
```
CyberDem0n, line 613: Why do we need to have this restriction?

patroni/validator.py
```diff
@@ -1045,6 +1045,7 @@ def validate_watchdog_mode(value: Any) -> None:
     },
     Optional("tags"): {
         Optional("nofailover"): bool,
+        Optional("failover_priority"): int,
```
CyberDem0n, line 1048: what are minimum and maximum values?

patroni/ha.py
```diff
@@ -804,6 +804,34 @@ def fetch_nodes_statuses(self, members: List[Member]) -> List[_MemberStatus]:
...
+        """Parallelizes fetching the status and priority from other nodes"""
+        pool = ThreadPool(len(members))
+        results = pool.map(
+            self.fetch_node_status_and_priority,
+            members
+        )  # Run API calls on members in parallel
+        pool.close()
+        pool.join()
+        return results
+
```
CyberDem0n, line 834: What is the whole point in introducing new methods is existing already do everything we need and don't even have to be changed?

patroni/ha.py
```diff
@@ -804,6 +804,34 @@ def fetch_nodes_statuses(self, members: List[Member]) -> List[_MemberStatus]:
...
+        and priority
+        :returns tuple of `_MemberStatus, int`"""
+
+        try:
+            response = self.patroni.request(member, timeout=2, retries=0)
+            data = response.data.decode('utf-8')
+            logger.info('Got response from %s %s: %s', member.name, member.api_url, data)
+            return (
+                _MemberStatus.from_api_response(member, json.loads(data)),
+                member.failover_priority
```
CyberDem0n, line 818: It could be just another field in the `_MemberStatus`.
Also, the priority should be taken from REST API rather than from the member key (what if it was just updated in the YAML file and Patroni didn't update the member key yet?).

patroni/ha.py
```diff
@@ -877,11 +905,21 @@ def _is_healthiest_node(self, members: Collection[Member], check_replication_lag
...
-            for st in self.fetch_nodes_statuses(members):
+            for st, priority in self.fetch_nodes_statuses_and_priorities(members):
                 if st.failover_limitation() is None:
                     if st.in_recovery is False:
                         logger.warning('Primary (%s) is still alive', st.member.name)
                         return False
+                    if self.is_lagging(st.wal_position):
+                        # The other node appears to have intolerable lag, move on (regardless of priority)
+                        continue
+                    if self.get_effective_tags().get('failover_priority', 1) < priority:
```
CyberDem0n, line 916: The `get_effective_tags()` is only useful for `touch_member()`, because sometimes we need to dynamically set `nosync` tag. In all other cases it is better to directly access property in the `Patroni` class.

patroni/ha.py
```diff
@@ -877,11 +905,21 @@ def _is_healthiest_node(self, members: Collection[Member], check_replication_lag
...
         if members:
-            for st in self.fetch_nodes_statuses(members):
+            for st, priority in self.fetch_nodes_statuses_and_priorities(members):
                 if st.failover_limitation() is None:
                     if st.in_recovery is False:
                         logger.warning('Primary (%s) is still alive', st.member.name)
                         return False
+                    if self.is_lagging(st.wal_position):
+                        # The other node appears to have intolerable lag, move on (regardless of priority)
+                        continue
```
CyberDem0n, line 915: There is no point in having it here. This method is checking if there are members that are ahead of our node in replication and we don't care if these nodes are "lagging" in term of `maximum_lag_on_failover`.

patroni/ha.py
CyberDem0n: Also, this check must be performed after the next condition: https://github.com/zalando/patroni/pull/2780/files#diff-b6a3a0ef64783a6d32feb908d74a3a009e9f25008b8ca5313c9c9747e3b74a63R923

patroni/config.py
mfpekala-nutanix: The way I have it set up right now, `failover_priority: 0` means "Never pursue leadership". This means that having `failover_priority: 0` and `nofailover: false` would be an ambiguous and invalid configuration. I thought it might be simpler to just force only one of these to be provided, since anytime you specify a `failover_priority` you are requiring a specific `nofailover` value.

I could change it so that it validates if the configuration is consistent between the two? Or other thoughts on how to rectify this? (I imagine we still want `nofailover` to have the exact same impact for backwards compatibility)

patroni/validator.py
mfpekala-nutanix: Minimum is 0, no maximum. I see that there is a more specific validator and have updated code accordingly

patroni/ha.py
mfpekala-nutanix: Good catch, removed all this and found a much simpler implementation

patroni/ha.py
mfpekala-nutanix: This makes sense, updated

patroni/ha.py
mfpekala-nutanix: This makes sense. The reason I had this here was because of the following concern:

Consider a situation where we have a three node system. Two of the nodes, call them A and B, have `failover_priority: 2` and the third, call it C, has `failover_priority: 1` (I imagine this will be a pretty typical configuration for people who chose to use the priority setting).

Say that leadership is originally on A but it goes down. Assume as well that something is wrong with B and it is outside of the `maximum_lag_on_failover`, but is still up. This means that B sees itself as lagging and refuses to race for leadership.

If we don't have this check, then C will see that B is up and has higher priority and so it will also not race for leadership. Thus, we'll end up in a scenario where no one goes for leadership even though both B and C have positive `failover_priority` (which we want to mean that 

Is there something I'm missing about why this either isn't a concern, or why this check won't help prevent it?

patroni/__main__.py
mfpekala-nutanix: I chose tag because I wanted to keep things as similar to existing `nofailover` implementation as possible. Happy to try and move elsewhere in the configuration if it makes more sense

patroni/ha.py
CyberDem0n: We need to check priority only when the other node has the same wal_position as ours.
I.e., add an additional check (`if my_wal_position == st.wal_position`) to https://github.com/zalando/patroni/pull/2780/files#diff-b6a3a0ef64783a6d32feb908d74a3a009e9f25008b8ca5313c9c9747e3b74a63R897

patroni/__main__.py
CyberDem0n: @hughcapet @matthbakeredb @barthisrael what do you think?

matthbakeredb: I haven't looked at the code changes but from the initial proposal it sounds reasonable.

patroni/dcs/__init__.py
```diff
@@ -257,7 +257,27 @@ def tags(self) -> Dict[str, Any]:
...
+            # Value of `nofailover` takes precedence over `failover_priority`
+            return bool(from_tags)
+        failover_priority = self.tags.get('failover_priority', 1)
+        return int(failover_priority) <= 0
+
+    @property
+    def failover_priority(self) -> int:
+        from_tags = self.tags.get('nofailover', None)
+        fp = self.tags.get('failover_priority', 1)
+        fp = int(fp) if fp is not None else 1
```
CyberDem0n, line 271: The `int()` could raise an exception.

```suggestion
        fp = parse_int(self.tags.get('failover_priority', 1)) or 1
```

patroni/dcs/__init__.py
```diff
@@ -257,7 +257,27 @@ def tags(self) -> Dict[str, Any]:
...
+        if from_tags is not None:
+            from_tags = bool(from_tags)
+            if from_tags is True:
+                # Value of `nofailover` takes precedence over `failover_priority`
+                return 0
+            elif fp <= 0:
+                # Value of `nofailover` takes precedence over `failover_priority`
+                return 1
+        return fp
```
CyberDem0n, line 280: Something is wrong here...
```suggestion
        return 0 if from_tags else fp
```

patroni/dcs/__init__.py
```diff
@@ -257,7 +257,27 @@ def tags(self) -> Dict[str, Any]:
 
     @property
     def nofailover(self) -> bool:
-        return self.tags.get('nofailover', False)
+        from_tags = self.tags.get('nofailover', None)
+        if from_tags is not None:
+            # Value of `nofailover` takes precedence over `failover_priority`
+            return bool(from_tags)
+        failover_priority = self.tags.get('failover_priority', 1)
```
CyberDem0n, line 264:
```suggestion
        failover_priority = parse_int(self.tags.get('failover_priority', 1)) or 1
```

patroni/dcs/__init__.py
```diff
@@ -257,7 +257,27 @@ def tags(self) -> Dict[str, Any]:
...
+        return int(failover_priority) <= 0
```
CyberDem0n, line 265:
```suggestion
        return failover_priority <= 0
```

patroni/ha.py
```diff
@@ -879,6 +879,7 @@ def _is_healthiest_node(self, members: Collection[Member], check_replication_lag
         if members:
             for st in self.fetch_nodes_statuses(members):
                 if st.failover_limitation() is None:
+                    priority = st.tags.get('failover_priority', 1)
```
CyberDem0n, line 882: I would actually make it another property of _MemberStatus() class.

patroni/ha.py
```diff
@@ -890,6 +891,13 @@ def _is_healthiest_node(self, members: Collection[Member], check_replication_lag
                                 or not self.cluster.sync.leader_matches(st.member.name):
                             return False
                         logger.info('Ignoring the former leader being ahead of us')
+                    if my_wal_position == st.wal_position and self.patroni.failover_priority < priority:
+                        # There's a higher priority non-lagging replica
+                        logger.info(
+                            '%s is not lagging and has a higher priority',
+                            st.member.name
+                        )
```
CyberDem0n, line 899: It perfectly fits to 120 characters.
```suggestion
                        logger.info('%s is not lagging and has a higher priority', st.member.name)
```

Also, maybe add actual priority values to the log.

tests/test_config.py
```diff
@@ -4,6 +4,7 @@
 import io
 
 from mock import MagicMock, Mock, patch
+import patroni.config
```
CyberDem0n, line 7:
```suggestion
```

tests/test_config.py
```diff
@@ -147,3 +148,51 @@ def open_mock(fname, *args, **kwargs):
     @patch('os.path.isdir', Mock(return_value=False))
     def test_invalid_path(self):
         self.assertRaises(ConfigParseError, Config, 'postgres0')
+
+    @patch.object(patroni.config, 'logger')
```
CyberDem0n, line 152:
```suggestion
    @patch('patroni.config.logger')
```

patroni/dcs/__init__.py
mfpekala-nutanix: what if `failover_priority` is 0? Then won't `fp` evaluate to `1` because `0 or 1` would evaluate to 1? What about:
```
fp = parse_int(self.tags.get('failover_priority', 1))
fp = 1 if fp is None else fp
```

patroni/config.py
```diff
@@ -601,3 +602,20 @@ def get_global_config(self, cluster: Union[Cluster, None]) -> GlobalConfig:
...
+                nofailover_tag is True and failover_priority_tag > 0
+                or nofailover_tag is False and failover_priority_tag <= 0
+            ):
+                logger.warning(
+                    'Conflicting configuration between nofailover: %s and failover_priority: %s.'
+                    + ' Defaulting to nofailover: %s',
+                    nofailover_tag,
+                    failover_priority_tag,
+                    nofailover_tag
+                )
```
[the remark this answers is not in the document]
mfpekala-nutanix: yes exactly. I updated the `Defaulting to nofailover: %s` to accurately reflect case where `nofailover: False` and `failover_priority: 0`, but I think we should still log in this case since it's an undesirable configuration to begin with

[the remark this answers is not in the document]
mfpekala-nutanix: just to clarify, what do we want to happen when provided:
`
nofailover: False,
failover_priority: 0
`?

Should this be the same as `nofailover: False, failover_priority: 1` or `nofailover: False, failover_priority: 0`?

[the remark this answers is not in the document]
CyberDem0n: > Should this be the same as nofailover: False, failover_priority: 1 or nofailover: False, failover_priority: 0?

IMO, the second seems to be more logical. Why should failover_priority suddenly change its value?

patroni/validator.py
```diff
@@ -354,6 +354,27 @@ def __init__(self, schema: Dict[str, Any]) -> None:
...
+        :Example:
+
+            OneOf("nofailover", "failover_priority"): CaseOptions({
+                Optional("nofailover"): bool,
+                Optional("failover_priority"): IntValidator(min=0, raise_assert=True),
+            })
+
+        That will make sure that at most one of ``nofailover`` or ``failover_priority`` is provided.
+        """
+        self._schema = schema
```
CyberDem0n, line 375:
```suggestion
        self._schema = {k.name: v for k, v in schema.items()}
```

CyberDem0n (commented):
Changes in validator could be simplified, please see suggestions.
Also, there are couple of conflicts that should be resolved.

patroni/validator.py
```diff
@@ -354,6 +354,27 @@ def __init__(self, schema: Dict[str, Any]) -> None:
...
+
+        :param schema: the optional arguments which may appear in the configuration.
+            Each option may or may not be provided, and will work with `:class:OneOf` or similar validations.
+
+        :Example:
+
+            OneOf("nofailover", "failover_priority"): CaseOptions({
+                Optional("nofailover"): bool,
+                Optional("failover_priority"): IntValidator(min=0, raise_assert=True),
+            })
```
matthbakeredb, line 371: Although there are places in this file that don't use this (yet), it might be good to show a working example that can be validated by `doctest`.

```suggestion
            >>> s = Schema({OneOf("nofailover", "failover_priority"): CaseOptions({
            ...   Optional("nofailover"): bool,
            ...   Optional("failover_priority"): IntValidator(min=0, raise_assert=True),
            ... })})
            ...
            >>> s({'nofailover': True})
            []
            >>> s({'nofailover': True, 'failover_priority': 0})
            ["  Multiple of ('nofailover', 'failover_priority') provided"]
```

patroni/validator.py
```diff
@@ -397,6 +418,22 @@ def __init__(self, name: str, default: OptionalType[Any] = None) -> None:
...
+class OneOf(object):
+    """Mark that at most one option from a `CaseOptions` can be suplied
+
+    Represents a list of possible configuration options in a given scope, where at most one can actually
+    be provided.
+
+    .. note::
+
+        It should be used together with a `:class:CaseOptions` object, where every field in the `:class:CaseOptions`
+        is something `:class:Optional`.
```
matthbakeredb, line 430:
```suggestion
        It should be used together with a :class:`CaseOptions` object, where every field in the :class:`CaseOptions`
        is something that is :class:`Optional`.
```

patroni/validator.py
```diff
@@ -397,6 +418,22 @@ def __init__(self, name: str, default: OptionalType[Any] = None) -> None:
...
+    """
+
+    def __init__(self, *args: Any) -> None:
```
matthbakeredb, line 433: No docstring.

patroni/validator.py
```diff
@@ -660,6 +697,10 @@ def iter_dict(self) -> Iterator[Result]:
         # One key in `validator` attribute (`key` variable) can be mapped to one or more keys in `data` attribute (`d`
         # variable), depending on the `key` type.
         for key in self.validator.keys():
+            if isinstance(key, OneOf):
+                if len(list(self._data_key(key))) > 1:
+                    yield Result(False, "Multiple of " + str(key.args) + " provided")
+                    continue
```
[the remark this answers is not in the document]
matthbakeredb: `f` not `a` ;-)

```suggestion
            if isinstance(key, OneOf) and len(list(self._data_key(key))) > 1:
                yield Result(False, f"Multiple of {str(key.args)} provided")
                continue
```

matthbakeredb (changes requested):
Mostly documentation comments

tests/test_validator.py
```diff
@@ -307,3 +307,35 @@ def test_validate_binary_name_empty_string(self, mock_out, mock_err):
         output = "\n".join(errors)
         self.assertEqual(['postgresql.bin_dir', 'postgresql.bin_name.postgres', 'raft.bind_addr', 'raft.self_addr'],
                          parse_output(output))
+
+    def test_one_of(self, _, __):
```
CyberDem0n, line 311:
```suggestion
    def test_one_of(self, *args):
```

tests/test_validator.py
```diff
@@ -307,3 +307,35 @@ def test_validate_binary_name_empty_string(self, mock_out, mock_err):
...
+        del c["tags"]["nofailover"]
+        c["tags"]["failover_priority"] = 1
+        errors = schema(c)
+        self.assertNotIn("tags  Multiple of ('nofailover', 'failover_priority') provided", errors)
+        # Providing both is not fine
+        c["tags"]["nofailover"] = False
+        errors = schema(c)
+        self.assertIn("tags  Multiple of ('nofailover', 'failover_priority') provided", errors)
+
+    def test_failover_priority_int(self, _, __):
```
CyberDem0n, line 331:
```suggestion
    def test_failover_priority_int(self, *args):
```

patroni/validator.py
```diff
@@ -397,6 +424,38 @@ def __init__(self, name: str, default: OptionalType[Any] = None) -> None:
...
+    Represents a list of possible configuration options in a given scope, where at most one can actually
+    be provided.
+
+    .. note::
+
+        It should be used together with a :class:`CaseOptions` object, where every field in the :class:`CaseOptions`
+        is something that is :class:`Optional`
+    """
+
+    def __init__(self, *args: Any) -> None:
```
CyberDem0n, line 439: Unlike in `Or` here we don't expect anything but strings.
```suggestion
    def __init__(self, *args: str) -> None:
```

tests/test_config.py
```diff
@@ -149,3 +149,51 @@ def open_mock(fname, *args, **kwargs):
...
     def test_invalid_path(self):
         self.assertRaises(ConfigParseError, Config, 'postgres0')
+
+    @patch('patroni.config.logger')
+    def test__validate_failover_tags(self, mock_logger):
+        """Ensures that only one of `nofailover` or `failover_priority` can be provided"""
+        mock_logger.warning = Mock()
+        config = Config("postgres0.yml")
+        mock_get = Mock()
+        with patch.object(config, 'get', mock_get):
```
CyberDem0n, line 159: IMO, it is better to use decorator:
```python
@patch.object(Config, 'get')
@patch('patroni.config.logger')
def test__validate_failover_tags(self, mock_logger, mock_get):
```

docs/yaml_configuration.rst
```diff
@@ -371,7 +371,8 @@ Watchdog
 
 Tags
 ----
--  **nofailover**: ``true`` or ``false``, controls whether this node is allowed to participate in the leader race and become a leader. Defaults to ``false``
+-  **nofailover**: ``true`` or ``false``, controls whether this node is allowed to participate in the leader race and become a leader. Defaults to ``false``. NOTE: provide only one of `nofailover` or `failover_priority`. Providing `nofailover: False` is the same as `failover_priority: 1` and `nofailover: True` is the same as `failover_priority: 0`.
+-  **failover_priority**: ``non-negative int``, controls the priority that this node should have during failover. Nodes with higher priority will be preferred over lower priority nodes. However, it's important to be aware that during failover, nodes are first filtered about using WAL status, and that this setting is only to break ties between replicas with equal WAL statuses. NOTE: Provide only one of `nofailover` or `failover_priority`. Setting `failover_priority: 0` will prevent this node from seeking leadership, and is semantically the same as `nofailover: True`.
```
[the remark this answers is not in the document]
mfpekala-nutanix: I agree that that's better. A bit strange to have the info pretty much duplicated so close together, and will probably do a better job catching the user's attention.

docs/yaml_configuration.rst
```diff
@@ -371,7 +371,8 @@ Watchdog
 
 Tags
 ----
--  **nofailover**: ``true`` or ``false``, controls whether this node is allowed to participate in the leader race and become a leader. Defaults to ``false``
+-  **nofailover**: ``true`` or ``false``, controls whether this node is allowed to participate in the leader race and become a leader. Defaults to ``false``. NOTE: provide only one of `nofailover` or `failover_priority`. Providing `nofailover: False` is the same as `failover_priority: 1` and `nofailover: True` is the same as `failover_priority: 0`.
```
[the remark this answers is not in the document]
mfpekala-nutanix: I also added a short sentence indicating that ``false`` means this node _can_ participate in leader races, which may not be immediately obvious to a new user

patroni/config.py
```diff
@@ -614,3 +615,25 @@ def get_global_config(self, cluster: Union[Cluster, None]) -> GlobalConfig:
         :returns: :class:`GlobalConfig` object
         """
         return get_global_config(cluster, self._dynamic_configuration)
+
+    def _validate_failover_tags(self):
+        """Checks nofailover/failover_priority config and warns user if it's contradictory"""
+        nofailover_tag = self.get('tags', {}).get('nofailover', None)
```
barthisrael, line 621: Maybe you could create a temporary variable to reduce copy'n'paste (`self.get('tags', {})`).

patroni/dcs/__init__.py
```diff
@@ -318,8 +318,14 @@ def tags(self) -> Dict[str, Any]:
 
     @property
     def nofailover(self) -> bool:
-        """The value for ``nofailover`` in :attr:`Member`.tags`` if defined, otherwise ``False``."""
-        return self.tags.get('nofailover', False)
+        """The value for ``nofailover`` in :attr:`Member`.tags`` if defined. Otherwise, returns True
```
barthisrael, line 321:
```suggestion
        """The value for ``nofailover`` in :attr:`Member`.tags`` if defined. Otherwise, returns ``True``
```

patroni/dcs/__init__.py
```diff
@@ -318,8 +318,14 @@ def tags(self) -> Dict[str, Any]:
...
+        if ``failover_priority`` in :attr:`Member`.tags`` is non-positive, else defaults to ``False``."""
+        from_tags = self.tags.get('nofailover', None)
+        if from_tags is not None:
+            # Value of `nofailover` takes precedence over `failover_priority`
+            return bool(from_tags)
+        fp = parse_int(self.tags.get('failover_priority', 1))
+        return fp <= 0 if fp is not None else False
```
barthisrael, line 328: I have the feeling we are duplicating this logic, and I wonder if is there a better way of handling that?

patroni/ha.py
```diff
@@ -34,7 +34,7 @@ class _MemberStatus(NamedTuple):
         dcs_last_seen - timestamp from JSON of last succesful communication with DCS
         timeline - timeline value from JSON
         wal_position - maximum value of `replayed_location` or `received_location` from JSON
-        tags - dictionary with values of different tags (i.e. nofailover)
+        tags - dictionary with values of different tags (i.e. nofailover, failover_priority)
```
barthisrael, line 37:
```suggestion
        tags - dictionary with values of different tags (i.e. ``nofailover``, ``failover_priority``)
```

patroni/ha.py
```diff
@@ -77,6 +77,15 @@ def failover_limitation(self) -> Optional[str]:
...
         return None
 
+    @property
+    def failover_priority(self) -> int:
+        """The failover priority of this node based on its tags.
+
+        :returns: the value of :attr:``~_MemberStatus.tags`` as an integer if possible, otherwise `1`.
+        """
+        fp = parse_int(self.tags.get('failover_priority', 1))
+        return fp if fp is not None else 1
```
barthisrael, line 87: With this one I also have the feeling that we are copying and pasting code. Again I wonder if we really need that, and if there is a better way of handling.

patroni/ha.py
```diff
@@ -909,6 +918,15 @@ def _is_healthiest_node(self, members: Collection[Member], check_replication_lag
                     if not self.sync_mode_is_active() or not self.cluster.sync.leader_matches(st.member.name):
                         return False
                     logger.info('Ignoring the former leader being ahead of us')
+                if my_wal_position == st.wal_position and self.patroni.failover_priority < st.failover_priority:
+                    # There's a higher priority non-lagging replica
+                    logger.info(
+                        '%s is not lagging and has priority %s, while this node has priority %s',
```
barthisrael, line 924: I guess "not lagging" is not an accurate message here.
The node can be considered the best candidate for a promotion even if it's lagging, as long as `maximum_lag_on_failover` is respected.
I'd say that although it's unlikely that two or more candidate nodes are lagging with the very same amount of lag, it is still possible.
Thoughts?

barthisrael (commented):
I still need to resume the review and finish going over other files, but I'm putting some comments that came up so far.

patroni/validator.py
```diff
@@ -354,6 +354,33 @@ def __init__(self, schema: Dict[str, Any]) -> None:
...
+            >>> s({'nofailover': True})
+            []
+            >>> s({'nofailover': True, 'failover_priority': 0})
+            ["  Multiple of ('nofailover', 'failover_priority') provided"]
+
+        That will make sure that at most one of ``nofailover`` or ``failover_priority`` is provided.
+        """
+        self._schema = {k.name: v for k, v in schema.items()}
+
+
```
barthisrael, line 383: Do we really need this new class?

I do see value in `OneOf`, but I think we could just the values like this -- using `Case` instead of `CaseOptions` + `Optional`:

```python
        OneOf("nofailover", "failover_priority"): Case({
            "nofailover": bool,
            "failover_priority": IntValidator(min=0, raise_assert=True),
        }),
```

The result would be apparently the same when running `--validate-config`, and the `OneOf` name already gives me the idea that only one of the options should be set.

Just to be clear, something like this is what I propose:

```diff
diff --git a/patroni/validator.py b/patroni/validator.py
index 3d0c1d1..fca2657 100644
--- a/patroni/validator.py
+++ b/patroni/validator.py
@@ -354,33 +354,6 @@ class Case(object):
         self._schema = schema


-class CaseOptions(object):
-    """Map how a :class:`dict` of optional configurations should be validated."""
-
-    def __init__(self, schema: Dict['Optional', Any]) -> None:
-        """Create a :class:`CaseOptions` object.
-
-        :param schema: the optional arguments which may appear in the configuration.
-                       Each option may or may not be provided, and will work with :class:`OneOf`
-                       or similar validations.
-
-        :Example:
-
-            >>> s = Schema({OneOf("nofailover", "failover_priority"): CaseOptions({
-            ...   Optional("nofailover"): bool,
-            ...   Optional("failover_priority"): IntValidator(min=0, raise_assert=True),
-            ... })})
-            ...
-            >>> s({'nofailover': True})
-            []
-            >>> s({'nofailover': True, 'failover_priority': 0})
-            ["  Multiple of ('nofailover', 'failover_priority') provided"]
-
-        That will make sure that at most one of ``nofailover`` or ``failover_priority`` is provided.
-        """
-        self._schema = {k.name: v for k, v in schema.items()}
-
-
 class Or(object):
     """Represent the list of options that are available.

@@ -425,15 +398,14 @@ class Optional(object):


 class OneOf(object):
-    """Mark that at most one option from a `CaseOptions` can be suplied
+    """Mark that at most one option from a `Case` can be suplied

     Represents a list of possible configuration options in a given scope, where at most one can actually
     be provided.

     .. note::

-        It should be used together with a :class:`CaseOptions` object, where every field in the :class:`CaseOptions`
-        is something that is :class:`Optional`
+        It should be used together with a :class:`Case` object.
     """

     def __init__(self, *args: str) -> None:
@@ -443,9 +415,9 @@ class OneOf(object):

         :Example:

-            OneOf("nofailover", "failover_priority"): CaseOptions({
-                Optional("nofailover"): bool,
-                Optional("failover_priority"): IntValidator(min=0, raise_assert=True),
+            OneOf("nofailover", "failover_priority"): Case({
+                "nofailover": bool,
+                "failover_priority": IntValidator(min=0, raise_assert=True),
             })

         The :class`OneOf` object is used to define that at most one of ``nofailover`` and
@@ -732,7 +704,7 @@ class Schema(object):
                         self.data[d] = key.default
                     validator = self.validator[key]
                     if isinstance(key, Or) and isinstance(self.validator[key], Case) \
-                            or isinstance(key, OneOf) and isinstance(self.validator[key], CaseOptions):
+                            or isinstance(key, OneOf) and isinstance(self.validator[key], Case):
                         validator = self.validator[key]._schema[d]
                     # In this loop we may be calling a new `Schema` either over an intermediate node in the tree, or
                     # over a leaf node. In the latter case the recursive calls in the given path will finish.
@@ -1115,9 +1087,9 @@ schema = Schema({
         Optional("safety_margin"): int
     },
     Optional("tags"): {
-        OneOf("nofailover", "failover_priority"): CaseOptions({
-            Optional("nofailover"): bool,
-            Optional("failover_priority"): IntValidator(min=0, raise_assert=True),
+        OneOf("nofailover", "failover_priority"): Case({
+            "nofailover": bool,
+            "failover_priority": IntValidator(min=0, raise_assert=True),
         }),
         Optional("clonefrom"): bool,
         Optional("noloadbalance"): bool,
```

tests/test_config.py
```diff
@@ -149,3 +149,50 @@ def open_mock(fname, *args, **kwargs):
...
+        mock_get.side_effect = [inconsistent_false] * 2
+        self.assertIsNone(config._validate_failover_tags())
+        mock_logger.warning.assert_called_once_with(
+            'Conflicting configuration between nofailover: %s and failover_priority: %s.'
+            + ' Defaulting to nofailover: %s',
+            False,
+            0,
+            False
+        )
+        mock_logger.warning = Mock()
```
barthisrael, line 188:
```suggestion
        mock_logger.warning.reset_mock()
```

barthisrael (changes requested):
Overall looks nice!
Posted a few comments along the way.

patroni/__main__.py
```diff
@@ -81,12 +82,32 @@ def ensure_unique_name(self) -> None:
...
 
     @property
     def nofailover(self) -> bool:
-        return bool(self.tags.get('nofailover', False))
+        from_tags = self.tags.get('nofailover', None)
+        if from_tags is not None:
+            # Value of `nofailover` takes precedence over `failover_priority`
+            return bool(from_tags)
+        fp = parse_int(self.tags.get('failover_priority', 1))
+        return fp <= 0 if fp is not None else False
```
matthbakeredb, line 103: Avoid using abbreviated variables.
```suggestion
        failover_priority = parse_int(self.tags.get('failover_priority', 1))
        return failover_priority <= 0 if failover_priority is not None else False
```

patroni/config.py
```diff
@@ -614,3 +615,25 @@ def get_global_config(self, cluster: Union[Cluster, None]) -> GlobalConfig:
         :returns: :class:`GlobalConfig` object
         """
         return get_global_config(cluster, self._dynamic_configuration)
+
+    def _validate_failover_tags(self):
+        """Checks nofailover/failover_priority config and warns user if it's contradictory"""
```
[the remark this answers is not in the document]
matthbakeredb:
```suggestion
        """Check ``nofailover``/``failover_priority`` config and warn user if it's contradictory."""
```

patroni/config.py
```diff
@@ -614,3 +615,25 @@ def get_global_config(self, cluster: Union[Cluster, None]) -> GlobalConfig:
...
+            failover_priority_tag = self.get('tags', {}).get('failover_priority', None)
+        # Next validate semantics between `failover_priority` and `nofailover`
+        if nofailover_tag is not None and failover_priority_tag is not None:
+            if (
+                nofailover_tag is True and failover_priority_tag > 0
+                or nofailover_tag is False and failover_priority_tag <= 0
+            ):
+                logger.warning(
+                    'Conflicting configuration between nofailover: %s and failover_priority: %s.'
+                    + ' Defaulting to nofailover: %s',
```
matthbakeredb, line 635: You can use implicit string concatenation inside brackets
```suggestion
                    'Conflicting configuration between nofailover: %s and failover_priority: %s.'
                    ' Defaulting to nofailover: %s',
```

patroni/dcs/__init__.py
matthbakeredb: First line of a docstring must be a one line summary. A more detailed explanation can be given after a line break.

matthbakeredb (changes requested):
I think I could take another look after a master branch rebase as there's conflicts.

Added a few suggestions.

patroni/config.py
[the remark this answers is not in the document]
mfpekala-nutanix: How: It just looks at the tags and gets their value.

> What is contradictory?

When implementing this feature, I adopted the following convention: `failover_priority` <= 0 means that it should _never_ seek leadership. On the flip side, `failover_priority > 0` means leadership is theoretically possible for this node. This made sense to me and was similar to other such priority systems I had seen for this kind of thing (notable MongoDB replication sets).

To preserve sanity (and backwards compatibility) the `nofailover` tag will still exist. A contradictory configuration is one where `nofailover` is `true` but `failover_priority > 0`, or where `nofailover` is `false` but `failover_priority <= 0`. Essentially, `nofailover` and `failover_priority` are communicating different things.

This checks for this edge case (which is a misconfiguration on the part of the user) and warns them. The behaviour is as if `failover_priority` were not provided (i.e `nofailover` is the bedrock source of truth)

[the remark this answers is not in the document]
matthbakeredb: Cool, thanks for the explanation. The question was more to prompt for what a reader of the docstring might think, it would be worth adding some of what you describe above to the docstring.

So, something like:

```suggestion
        """Compare ``nofailover``/``failover_priority`` tags and warns user of conflicting configuration.
        
        .. note::
          To preserve sanity (and backwards compatibility) the ``nofailover`` tag will still exist. A contradictory
          configuration is one where ``nofailover`` is ``True`` but ``failover_priority > 0``, or where 
          ``nofailover`` is ``False``, but ``failover_priority <= 0``. Essentially, ``nofailover`` and 
          ``failover_priority`` are communicating different things.

          This checks for this edge case (which is a misconfiguration on the part of the user) and warns them. 
          The behaviour is as if ``failover_priority`` were not provided (i.e ``nofailover`` is the
          bedrock source of truth)
        
        """
```

(You might want to check line length before applying)

patroni/ha.py
mfpekala-nutanix: This makes sense! This I believe is a holdover from an earlier version that would select for priority among all nodes that were within `maximum_lag_on_failover`. What about:

`'%s has equally tolerable WAL position and has priority %s, while this node has priority %s',`

patroni/validator.py
mfpekala-nutanix: I like this better! Updated.

tests/test_patroni.py
```diff
@@ -173,11 +173,47 @@ def test_noloadbalance(self):
         self.assertTrue(self.p.noloadbalance)
 
     def test_nofailover(self):
```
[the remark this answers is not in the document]
mfpekala-nutanix: Ended up using `unittest.TestCase.subTest` (https://docs.python.org/3/library/unittest.html) as parametrize seemed to have problems working with class methods. Same idea.

mfpekala-nutanix:
I've noticed that my unit tests are failing manually due to error:
```sh
socket.gaierror: [Errno 8] nodename nor servname provided, or not known
```
which I believe is coming from
```python
class TestGenerateConfig(unittest.TestCase):

    no_value_msg = '#FIXME'
    _HOSTNAME = socket.gethostname()
    _IP = sorted(socket.getaddrinfo(_HOSTNAME, 0, socket.AF_UNSPEC, socket.SOCK_STREAM, 0), key=lambda x: x[0])[0][4][0]
```
in `test_config_generator.py`. Any ideas for how to resolve this? I've checked that `HOSTNAME` is set and that `127.0.0.1 localhost` is in `/etc/hosts`.

patroni/ha.py
barthisrael: I think it looks better, yes!

patroni/validator.py
barthisrael: Cool!

features/priority_failover.feature
```diff
@@ -0,0 +1,20 @@
...
+    Given I start postgres0
+    Then postgres0 role is the primary after 10 seconds
+
+  Scenario: check higher failover priority is respected
+    Given I configure and start postgres2 with a tag failover_priority 1
+    And I configure and start postgres3 with a tag failover_priority 2
+    And I sleep for 5 seconds
+    When I shut down postgres0
+    Then postgres3 role is the primary after 10 seconds
+    And there is a "postgres3 is not lagging and has priority 2, while this node has priority 1" INFO in the postgres2 patroni log
```
barthisrael: I think we need to fix this check too :) I mean, I can see you already changed the message through this commit.

patroni/utils.py
```diff
@@ -1061,3 +1061,29 @@ def get_major_version(bin_dir: Optional[str] = None, bin_name: str = 'postgres')
...
+def failover_priority_from_tags(tags: Dict[str, Any]) -> int:
+    """Common logic for obtaining the value of ``failover_priority`` from ``tags`` if defined.
+
+    If ``nofailover`` is defined as True, this will return 0. Otherwise, it will return the value of
+    ``failover_priority``, defaulting to 1 if it is not defined.
+    """
+    from_tags = tags.get('nofailover', None)
+    failover_priority = parse_int(tags.get('failover_priority', 1))
+    failover_priority = 1 if failover_priority is None else failover_priority
+    return 0 if from_tags else failover_priority
```
barthisrael, line 1089: I guess we might want to benefit from the PR opened by Alexander (https://github.com/zalando/patroni/pull/2823).
But that's just a guess, I still need to take a proper look at that PR.

patroni/validator.py
```diff
@@ -392,6 +392,37 @@ def __init__(self, name: str, default: OptionalType[Any] = None) -> None:
         self.default = default
 
 
+class OneOf(object):
+    """Mark that at most one option from a `Case` can be suplied
```
barthisrael, line 396:
```suggestion
"""Mark that at most one option from a :class:`Case` can be suplied
```

(If you mean the class)

features/priority_failover.feature
barthisrael: Never mind, I posted a comment in the commit, not in the "compare all"

barthisrael:
> Also for sphinx (for clarification) it looks like it's only outputting warnings. I'm assuming that it's marked to fail if there are any warnings, so if I fix those I should be good? Or is there something else I should look into to resolve that build

Yes, that's it. We have this in `tox.ini`:

```ini
commands =
    sphinx-build \
    -d "{envtmpdir}{/}doctree" docs "{toxworkdir}{/}docs_out" \
    --color -b html \
    -T -E -W --keep-going \
    {posargs}
```

You're having issues with `-W` (turn warnings into errors) because of this:

```
/home/runner/work/patroni/patroni/patroni/dcs/__init__.py:docstring of patroni.dcs.Member.nofailover:4: WARNING: Field list ends without a blank line; unexpected unindent.
```

barthisrael:
> I've noticed that my unit tests are failing manually due to error:
> 
> ```shell
> socket.gaierror: [Errno 8] nodename nor servname provided, or not known
> ```
> 
> which I believe is coming from
> 
> ```python
> class TestGenerateConfig(unittest.TestCase):
> 
>     no_value_msg = '#FIXME'
>     _HOSTNAME = socket.gethostname()
>     _IP = sorted(socket.getaddrinfo(_HOSTNAME, 0, socket.AF_UNSPEC, socket.SOCK_STREAM, 0), key=lambda x: x[0])[0][4][0]
> ```
> 
> in `test_config_generator.py`. Any ideas for how to resolve this? I've checked that `HOSTNAME` is set and that `127.0.0.1 localhost` is in `/etc/hosts`.

You need to make sure you are able to resolve the hostname returned by `socket.gethostname()`.
The exception that you mentioned is likely being triggered by `socket.getaddrinfo` because it's receiving something from `socket.gethostname` which it's not able to resolve (https://github.com/python/cpython/blob/main/Modules/socketmodule.c#L6761).

patroni/ha.py
```diff
@@ -922,18 +930,29 @@ def _is_healthiest_node(self, members: Collection[Member], check_replication_lag
...
+                            return False
+                        logger.info('Ignoring the former leader being ahead of us')
+                    if my_wal_position == st.wal_position and self.patroni.failover_priority < st.failover_priority:
+                        # There's a higher priority non-lagging replica
+                        logger.info(
+                            '%s has equally tolerable WAL position and priority %s, while this node has priority %s',
+                            st.member.name,
+                            st.failover_priority,
+                            self.patroni.failover_priority,
+                        )
```
CyberDem0n, line 954: I think you need to start from scratch for this block. You reverted some of the changes done in  the https://github.com/zalando/patroni/pull/2804

patroni/validator.py
```diff
@@ -674,7 +708,8 @@ def iter_dict(self) -> Iterator[Result]:
                     if d not in self.data and isinstance(key, Optional):
                         self.data[d] = key.default
                     validator = self.validator[key]
-                    if isinstance(key, Or) and isinstance(self.validator[key], Case):
+                    if isinstance(key, Or) and isinstance(self.validator[key], Case) \
+                            or isinstance(key, OneOf) and isinstance(self.validator[key], Case):
```
CyberDem0n, line 712:
```suggestion
                    if isinstance(key, (Or, OneOf)) and isinstance(self.validator[key], Case):
```

patroni/validator.py
```diff
@@ -1050,7 +1093,10 @@ def validate_watchdog_mode(value: Any) -> None:
         Optional("safety_margin"): int
     },
     Optional("tags"): {
-        Optional("nofailover"): bool,
+        OneOf("nofailover", "failover_priority"): Case({
+            "nofailover": bool,
+            "failover_priority": IntValidator(min=0, raise_assert=True),
+        }),
```
CyberDem0n, line 1099: I read this construction as: exactly one of `nofailover` or `failover_priority` must be specified, while having none of them is also a valid configuration.
Previously inner keys were `Optional()` and therefore if was fine and now it is confusing.
Should maybe `OneOf` be renamed to `Xor`? It will better describe the intention.

patroni/utils.py
mfpekala-nutanix: yes I was taking a look at that one too! Don't think it will be a huge change either way, just remove some of the code dup here I think

patroni/ha.py
mfpekala-nutanix: Good catch! Sorry I missed that

patroni/validator.py
mfpekala-nutanix: I agree, I had changed it from before in response to [this comment](https://github.com/zalando/patroni/pull/2780#discussion_r1295118315), which made a good point about not repeating ourselves with `class Case` vs `class CaseOptions`.

I brought the `Optional` back but made a slight change to `Case` to allow it. I'm hoping this ends up being the best of both worlds.

(Was a bit hesitant to the `Xor` idea because ideally we want to allow the user to specify neither `nofailover` no`failover_priority`, and `Xor` doesn't necessarily communicate that)

matthbakeredb (changes requested):
You have added all the docs build files under

docs/_build
docs/modules
docs/pdf

Can you remove them from the commit?

Also please write more explanatory commit messages?

patroni/tags.py
```diff
@@ -45,8 +47,29 @@ def clonefrom(self) -> bool:
 
     @property
     def nofailover(self) -> bool:
-        """``True`` if ``nofailover`` is ``True``, else ``False``."""
-        return bool(self.tags.get('nofailover', False))
+        """Common logic for obtaining the value of ``nofailover`` from ``tags`` if defined.
+
+        If ``no_failover is not defined, this methods returns True if ``failover_priority`` in
+        :attr:`Member`.tags`` is non-positive, else defaults to ``False``.
```
CyberDem0n, line 53:
```suggestion
        If ``nofailover`` is not defined, this methods returns ``True`` if ``failover_priority`` is non-positive,
        ``False`` otherwise.
```

patroni/tags.py
```diff
@@ -45,8 +47,29 @@ def clonefrom(self) -> bool:
...
+        """
+        from_tags = self.tags.get('nofailover', None)
+        if from_tags is not None:
+            # Value of `nofailover` takes precedence over `failover_priority`
+            return bool(from_tags)
+        failover_priority = parse_int(self.tags.get('failover_priority', 1))
+        return failover_priority <= 0 if failover_priority is not None else False
```
CyberDem0n, line 60:
```suggestion
        failover_priority = parse_int(self.tags.get('failover_priority'))
        return failover_priority is not None and failover_priority <= 0
```

patroni/tags.py
```diff
@@ -45,8 +47,29 @@ def clonefrom(self) -> bool:
...
     @property
     def nofailover(self) -> bool:
-        """``True`` if ``nofailover`` is ``True``, else ``False``."""
-        return bool(self.tags.get('nofailover', False))
+        """Common logic for obtaining the value of ``nofailover`` from ``tags`` if defined.
+
+        If ``no_failover is not defined, this methods returns True if ``failover_priority`` in
+        :attr:`Member`.tags`` is non-positive, else defaults to ``False``.
+        """
+        from_tags = self.tags.get('nofailover', None)
```
CyberDem0n, line 55:
```suggestion
        from_tags = self.tags.get('nofailover')
```

patroni/tags.py
```diff
@@ -45,8 +47,29 @@ def clonefrom(self) -> bool:
...
+
+    @property
+    def failover_priority(self) -> int:
+        """Common logic for obtaining the value of ``failover_priority`` from ``tags`` if defined.
+
+        If ``nofailover`` is defined as True, this will return 0. Otherwise, it will return the value of
+        ``failover_priority``, defaulting to 1 if it's not defined.
+        """
+        from_tags = self.tags.get('nofailover', None)
+        failover_priority = parse_int(self.tags.get('failover_priority', 1))
```
CyberDem0n, line 70:
```suggestion
        from_tags = self.tags.get('nofailover')
        failover_priority = parse_int(self.tags.get('failover_priority'))
```

patroni/tags.py
```diff
@@ -45,8 +47,29 @@ def clonefrom(self) -> bool:
...
+            return bool(from_tags)
+        failover_priority = parse_int(self.tags.get('failover_priority', 1))
+        return failover_priority <= 0 if failover_priority is not None else False
+
+    @property
+    def failover_priority(self) -> int:
+        """Common logic for obtaining the value of ``failover_priority`` from ``tags`` if defined.
+
+        If ``nofailover`` is defined as True, this will return 0. Otherwise, it will return the value of
+        ``failover_priority``, defaulting to 1 if it's not defined.
```
CyberDem0n, line 67:
```suggestion
        If ``nofailover`` is defined as ``True``, this will return ``0``. Otherwise, it will return the value of
        ``failover_priority``, defaulting to ``1`` if it's not defined or invalid.
```

features/priority_failover.feature
```diff
@@ -0,0 +1,20 @@
+Feature: priority replication
+  We should check that we can give nodes priority during failover
+
+  Scenario: check failover priority 0 prevents leaderships
+    Given I configure and start postgres0 with a tag failover_priority 1
+    And I configure and start postgres1 with a tag failover_priority 0
+    When I shut down postgres0
```
CyberDem0n, line 7:
```suggestion
    Then replication works from postgres0 to postgres1 after 20 seconds
    When I shut down postgres0
```

features/priority_failover.feature
```diff
@@ -0,0 +1,20 @@
...
+    And I sleep for 5 seconds
+    Then postgres1 role is the secondary after 10 seconds
+    And there is a "following a different leader because I am not allowed to promote" INFO in the postgres1 patroni log
+    Given I start postgres0
+    Then postgres0 role is the primary after 10 seconds
+
+  Scenario: check higher failover priority is respected
+    Given I configure and start postgres3 with a tag failover_priority 2
+    And I configure and start postgres2 with a tag failover_priority 1
+    And I sleep for 10 seconds
```
CyberDem0n, line 17:
```suggestion
    Then replication works from postgres0 to postgres3 after 20 seconds
    And replication works from postgres0 to postgres2 after 20 seconds
```

patroni/config.py
```diff
@@ -895,3 +896,36 @@ def get_global_config(self, cluster: Optional[Cluster]) -> GlobalConfig:
...
+          To preserve sanity (and backwards compatibility) the ``nofailover`` tag will still exist. A contradictory
+          configuration is one where ``nofailover`` is ``True`` but ``failover_priority > 0``, or where
+          ``nofailover`` is ``False``, but ``failover_priority <= 0``. Essentially, ``nofailover`` and
+          ``failover_priority`` are communicating different things.
+          This checks for this edge case (which is a misconfiguration on the part of the user) and warns them.
+          The behaviour is as if ``failover_priority`` were not provided (i.e ``nofailover`` is the
+          bedrock source of truth)
+        """
+        tags = self.get('tags', {})
+        nofailover_tag = tags.get('nofailover', None)
```
CyberDem0n, line 913:
```suggestion
        nofailover_tag = tags.get('nofailover')
```

patroni/config.py
```diff
@@ -895,3 +896,36 @@ def get_global_config(self, cluster: Optional[Cluster]) -> GlobalConfig:
...
+        failover_priority_tag = tags.get('failover_priority', None)
```
CyberDem0n, line 914:
```suggestion
        failover_priority_tag = parse_int(tags.get('failover_priority'))
```

patroni/config.py
```diff
@@ -895,3 +896,36 @@ def get_global_config(self, cluster: Optional[Cluster]) -> GlobalConfig:
...
+        # First handle case where `failover_priority` is str, not int
+        if isinstance(failover_priority_tag, str):
+            self['tags']['failover_priority'] = parse_int(failover_priority_tag)
+            failover_priority_tag = tags.get('failover_priority', None)
```
CyberDem0n, line 918: Please don't change anything. str->int casting is already handled in the `Tags` class.
```suggestion
```

CyberDem0n (commented):
> It seems to randomly have a small chance of failing, and I've seen it across OS/DCS

Well... sleep 10 isn't the right tool to make sure that replication is already working. See suggestions.

features/priority_failover.feature
mfpekala-nutanix: This helps, I also think that sometimes postgres3 would become leader so fast that postgres2 would see itself as having a inferior WAL status, which would lead to a different log message. Fixed the behave step to include accepting either of these two cases (either node 2 sees node 3 with higher priority _or_ higher wal status because it's already become the leader)

mfpekala-nutanix: Looks like it's still failing behave, but on a different test that seems unrelated

features/priority_failover.feature
```diff
@@ -0,0 +1,23 @@
+Feature: priority replication
+  We should check that we can give nodes priority during failover
+
+  Scenario: check failover priority 0 prevents leaderships
+    Given I configure and start postgres0 with a tag failover_priority 1
+    And I configure and start postgres1 with a tag failover_priority 0
+    Then replication works from postgres0 to postgres1 after 20 seconds
+    When I shut down postgres0
+    And I sleep for 5 seconds
```
CyberDem0n, line 9: Such sleep calls are not very reliable. Sometimes we get so slow MacOS instances that tests are still failing.
In general it is better to check some condition in a loop instead.
For example:
```cucumber
    And there is a "following a different leader because I am not allowed to promote" INFO in the postgres1 patroni log after 5 seconds
```
And perform the check every second in a loop (example): https://github.com/zalando/patroni/blob/3333e78500204159332e032e5d07727dfbb7129d/features/steps/patroni_api.py#L136-L147

That is., fail on timeout rather than on a single attempt after sleep. The advantage is that on success we interrupt the loop much earlier. Also, there is a possibility on increase the timeout when it is known that tests are executed on CI: https://github.com/zalando/patroni/blob/3333e78500204159332e032e5d07727dfbb7129d/features/environment.py#L1060-L1062

mfpekala-nutanix: This makes a lot of sense, thanks for the in-depth explanation. Should I rewrite parts of this test to avoid the sleep here?

CyberDem0n: @mfpekala-nutanix yes, please rewrite them using to use 1s polling with configurable timeout before fail instead of using sleep.

patroni/config.py
```diff
@@ -897,3 +898,31 @@ def get_global_config(self, cluster: Optional[Cluster]) -> GlobalConfig:
...
+          ``nofailover`` is ``False``, but ``failover_priority <= 0``. Essentially, ``nofailover`` and
+          ``failover_priority`` are communicating different things.
+          This checks for this edge case (which is a misconfiguration on the part of the user) and warns them.
+          The behaviour is as if ``failover_priority`` were not provided (i.e ``nofailover`` is the
+          bedrock source of truth)
+        """
+        tags = self.get('tags', {})
+        nofailover_tag = tags.get('nofailover')
+        failover_priority_tag = parse_int(tags.get('failover_priority'))
+        if nofailover_tag is not None and failover_priority_tag is not None:
```
hughcapet, line 917: can we decrease the number of nested levels with 
```
if nofailover_tag is None or failover_priority_tag is None:
    return
```
or by integrating into one condition (which is probably more difficult to read)
?

patroni/tags.py
```diff
@@ -45,8 +47,29 @@ def clonefrom(self) -> bool:
...
+    @property
+    def failover_priority(self) -> int:
+        """Common logic for obtaining the value of ``failover_priority`` from ``tags`` if defined.
+
+        If ``nofailover`` is defined as ``True``, this will return ``0``. Otherwise, it will return the value of
+        ``failover_priority``, defaulting to ``1`` if it's not defined or invalid.
+        """
+        from_tags = self.tags.get('nofailover')
+        failover_priority = parse_int(self.tags.get('failover_priority'))
+        failover_priority = 1 if failover_priority is None else failover_priority
```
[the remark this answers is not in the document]
CyberDem0n:
```python
parse_int(0) or 1
```

will break it.

docs/yaml_configuration.rst
```diff
@@ -384,11 +384,15 @@ Watchdog
...
 -  **noloadbalance**: ``true`` or ``false``. If set to ``true`` the node will return HTTP Status Code 503 for the ``GET /replica`` REST API health-check and therefore will be excluded from the load-balancing. Defaults to ``false``.
 -  **replicatefrom**: The IP address/hostname of another replica. Used to support cascading replication.
 -  **nosync**: ``true`` or ``false``. If set to ``true`` the node will never be selected as a synchronous replica.
+-  **nofailover**: ``true`` or ``false``, controls whether this node is allowed to participate in the leader race and become a leader. Defaults to ``false``, meaning this node _can_ participate in leader races. 
+-  **failover_priority**: integer, controls the priority that this node should have during failover. Nodes with higher priority will be preferred over lower priority nodes. However, it's important to be aware that during failover, Patroni will select the best promotion candidates based on how up-to-date they are with the current primary node, and that this setting is only used to break ties between replicas that are considered the best candidates by Patroni, if 2 or more happen to be selected.
```
CyberDem0n, line 392:
```suggestion
-  **failover_priority**: integer, controls the priority that this node should have during failover. Nodes with higher priority will be preferred over lower priority nodes if they received/replayed the same amount of WAL. But, nodes with higher values of receive/replay LSN are preferred even if their priority is lower. If the ``failover_priority`` is 0 or negative - such node is not allowed to participate in the leader race and become a leader (similar to ``nofailover: true``).
```

tests/test_ha.py
```diff
@@ -1033,9 +1035,17 @@ def test__is_healthiest_node(self):
...
+        # accessible, not in_recovery
+        self.ha.fetch_node_status = get_node_status(in_recovery=False)
         self.assertFalse(self.ha._is_healthiest_node(self.ha.old_cluster.members))
-        self.ha.fetch_node_status = get_node_status(wal_position=11)  # accessible, in_recovery, wal position ahead
+        # accessible, in_recovery, higher priority
+        self.ha.fetch_node_status = get_node_status(failover_priority=2)
+        self.assertFalse(self.ha._is_healthiest_node(self.ha.old_cluster.members))
+        # if there is a higher-priority node but it has a lower WAL position then this node should race
+        self.ha.fetch_node_status = get_node_status(failover_priority=6, wal_position=9)
+        self.assertTrue(self.ha._is_healthiest_node)
```
hughcapet, line 1046:
```suggestion
        self.assertTrue(self.ha._is_healthiest_node(self.ha.old_cluster.members))
```

patroni/validator.py
```diff
@@ -379,21 +396,37 @@ def __init__(self, *args: Any) -> None:
         self.args = args
 
 
-class Optional(object):
-    """Mark a configuration option as optional.
+class OneOf(object):
```
hughcapet, line 399: maybe `AtMostOne`? (no idea why I thought it fits `*available_dcs`. I guess, "less than one" should not pass there)

hughcapet:
👍

CyberDem0n:
:+1:

PR #2781, opened 2023-07-27
Set permissions for files and directories created in PGDATA

### LINKED ISSUE #1899: allow-group-access clause in initdb seems not woking correctly

CyberDem0n:
Postgres supports two types of permissions:
1. owner only
2. group readable

By default the first one is used because it provides better security. But, sometimes people want to run a backup tool with the user that is different from postgres. In this case the second option becomes very useful. Unfortunately it didn't work correctly because Patroni was creating files with owner access only permissions.

This PR changes the behaviour and permissions on files and directories that are created by Patroni will be calculated based on permissions of PGDATA. I.e., they will get group readable access when it is necessary.

Close #1899
Close #1901

### DIFF
--- a/patroni/file_perm.py
+++ b/patroni/file_perm.py
@@ -11,7 +11,11 @@ logger = logging.getLogger(__name__)
 
 
 class __FilePermissions:
-    """Class that helps with figuring out file and directory permissions based on permissions of PGDATA. """
+    """Helper class for managing permissions of directories and files under PGDATA.
+    
+    Execute :meth:`set_permissions_from_data_directory` to figure out which permissions should be used for files and
+    directories under PGDATA based on permissions of PGDATA root directory.
+    """
 
     # Mode mask for data directory permissions that only allows the owner to read/write directories and files -- mask 077.
     __PG_MODE_MASK_OWNER = stat.S_IRWXG | stat.S_IRWXO
--- a/patroni/file_perm.py
+++ b/patroni/file_perm.py
@@ -12,12 +12,13 @@ logger = logging.getLogger(__name__)
 
 class __FilePermissions:
     """Helper class for managing permissions of directories and files under PGDATA.
-    
+
     Execute :meth:`set_permissions_from_data_directory` to figure out which permissions should be used for files and
     directories under PGDATA based on permissions of PGDATA root directory.
     """
 
-    # Mode mask for data directory permissions that only allows the owner to read/write directories and files -- mask 077.
+    # Mode mask for data directory permissions that only allows the owner to
+    # read/write directories and files -- mask 077.
     __PG_MODE_MASK_OWNER = stat.S_IRWXG | stat.S_IRWXO
 
     # Mode mask for data directory permissions that also allows group read/execute -- mask 027.
@@ -42,9 +43,10 @@ class __FilePermissions:
 
     def __set_umask(self) -> None:
         """Set umask value based on calculations.
-        
+
         .. note::
-            Should only be called once either :meth:`__set_owner_permissions` or :meth:`__set_group_permissions` has been executed.
+            Should only be called once either :meth:`__set_owner_permissions`
+            or :meth:`__set_group_permissions` has been executed.
         """
         try:
             os.umask(self.__pg_mode_mask)
--- a/patroni/file_perm.py
+++ b/patroni/file_perm.py
@@ -13,31 +13,35 @@ logger = logging.getLogger(__name__)
 class __FilePermissions:
     """Class that helps with figuring out file and directory permissions based on permissions of PGDATA. """
 
-    # Mode mask for data directory permissions that only allows the owner to read/write directories and files.
+    # Mode mask for data directory permissions that only allows the owner to read/write directories and files -- mask 077.
     __PG_MODE_MASK_OWNER = stat.S_IRWXG | stat.S_IRWXO
 
-    # Mode mask for data directory permissions that also allows group read/execute.
+    # Mode mask for data directory permissions that also allows group read/execute -- mask 027.
     __PG_MODE_MASK_GROUP = stat.S_IWGRP | stat.S_IRWXO
 
-    # Default mode for creating directories
+    # Default mode for creating directories -- mode 700.
     __PG_DIR_MODE_OWNER = stat.S_IRWXU
 
-    # Mode for creating directories that allows group read/execute
+    # Mode for creating directories that allows group read/execute -- mode 750.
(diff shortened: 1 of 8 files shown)


patroni/postgresql/config.py
```diff
@@ -438,7 +448,9 @@ def write_postgresql_conf(self, configuration: Optional[CaseInsensitiveDict] = N
     def append_pg_hba(self, config: List[str]) -> bool:
         if not self.hba_file and not self._config.get('pg_hba'):
             with open(self._pg_hba_conf, 'a') as f:
+                self.set_file_permissions(self._pg_hba_conf)
                 f.write('\n{}\n'.format('\n'.join(config)))
+            self.set_file_permissions(self._pg_hba_conf)
```
barthisrael, line 453:
```suggestion
                f.write('\n{}\n'.format('\n'.join(config)))
            self.set_file_permissions(self._pg_hba_conf)
```

I guess we can remove one of the apparently redundant calls.

patroni/postgresql/config.py
```diff
@@ -410,6 +419,7 @@ def write_postgresql_conf(self, configuration: Optional[CaseInsensitiveDict] = N
             configuration['hot_standby_feedback'] = 'on'
 
         with ConfigWriter(self._postgresql_conf) as f:
+            self.set_file_permissions(self._postgresql_conf)
```
barthisrael, line 422: Just a side thought...

I was wondering if somehow we could use `os.fchmod` inside `ConfigWriter.__enter__` to set permissions of the file descriptor as we are always calling `self.set_file_permissions` right after entering the context.

However that class is not "PGDATA aware", and we would somehow copy the logic of `set_file_permissions`, so...

tests/test_file_perm.py
```diff
@@ -0,0 +1,25 @@
...
+        mock_stat.return_value.st_mode = stat.S_IRWXU | stat.S_IRGRP | stat.S_IXGRP
+        pg_perm.set_permissions_from_data_directory('test')
+        self.assertEqual(mock_umask.call_args[0][0], stat.S_IWGRP | stat.S_IRWXO)
+        self.assertEqual(mock_logger.call_args[0][0], 'Can not set umask to %03o: %r')
+
+    @patch('os.stat', Mock(side_effect=FileNotFoundError))
+    @patch('patroni.file_perm.logger.error')
+    def test_set_permissions_from_data_directory(self, mock_logger):
+        pg_perm.set_permissions_from_data_directory('test')
+        self.assertEqual(mock_logger.call_args[0][0], 'Can not check permissions on %s: %r')
```
barthisrael, line 25: What about adding a couple tests to check if `__set_group_permissions` or `__set_owner_permissions` is being called?

barthisrael (changes requested):
The PR looks good overall. Posted a few comments about docstrings, and a few side thoughts.

patroni/file_perm.py
```diff
@@ -0,0 +1,85 @@
...
+            self.__set_umask()
+
+    @property
+    def dir_create_mode(self) -> int:
+        """Directory permissions."""
+        return self.__pg_dir_create_mode
+
+    @property
+    def file_create_mode(self) -> int:
+        """File permissions"""
```
hughcapet, line 81:
```suggestion
        """File permissions."""
```

tests/test_postgresql.py
```diff
@@ -523,8 +523,9 @@ def test_sysid(self):
     def test_save_configuration_files(self):
         self.p.config.save_configuration_files()
 
-    @patch('os.path.isfile', Mock(side_effect=[False, True]))
-    @patch('shutil.copy', Mock(side_effect=IOError))
+    @patch('os.path.isfile', Mock(side_effect=[False, True, False, True]))
+    @patch('shutil.copy', Mock(side_effect=[None, IOError]))
+    @patch('os.chmod', Mock())
```
hughcapet, line 528: I guess, perfectly, the tests should also check that `chmod`  was actually called with the proper permissions everywhere it is required (where it was added in this PR) 😇

tests/test_file_perm.py
CyberDem0n: Well, `umask` could only called when `os.stat()` didn't raise any exception.
Therefore one or another is successfully called :)

barthisrael: Ah, sure. I guess I didn't express myself clearly, sorry.
I meant based on a few sample mock values for `st_mode`, to check if it's taking the expected code path.

CyberDem0n: At least one path is checked here: https://github.com/zalando/patroni/pull/2781/files#diff-6251748fa4c45a038329a6f9eb925a57f7a1c07b50b64823ca36ea20804748efR16-R18

barthisrael: Right, the group one.
Your call on adding a test for the other or not.

hughcapet:
👍

PR #2782, opened 2023-07-27
Avoid unnecessary updates of /status key

CyberDem0n:
When we don't have permanent logical slots Patroni was updating the `/status` on every heart-beat loop even when LSN on the primary isn't moving forward.

The issue was introduced in the #2485


hughcapet:
👍

CyberDem0n:
:+1:

Issue #2779, opened 2023-07-26
Ambiguous behavior


glushakov:
### What happened?

Hi.
I have encountered a behavior that I can not understand.

On the host server_2, an out of memory occurred as a result of which the postgres process was killed
`2023-07-26 18:01:37.786 MSK    2307  0 LOG:  server process (PID 1326001) was terminated by signal 9: Killed`

But the switch to the standby (server_1) happened earlier, I think when the master was already overloaded and could not update its leader key.
`2023-07-26 18:00:45,518 INFO: promoted self to leader by acquiring session lock`

But after a successful promotion, an incomprehensible message appears after 1 minute on a new master (server_1)
```
2023-07-26 18:01:37,179 INFO: Lock owner: server_2; I am server_1
2023-07-26 18:01:37,180 INFO: Demoting self (immediate-nolock)
2023-07-26 18:01:37,371 INFO: demoting self because I do not have the lock and I was a leader
```

At the same time, server_2 did not know that the master had been changed and continued to work without changing the role:
```
2023-07-26 17:59:09,144 INFO: no action. I am (server_2), the leader with the lock
2023-07-26 17:59:19,145 INFO: no action. I am (server_2), the leader with the lock
2023-07-26 17:59:29,144 INFO: no action. I am (server_2), the leader with the lock
2023-07-26 17:59:39,145 INFO: no action. I am (server_2), the leader with the lock
2023-07-26 17:59:49,144 INFO: no action. I am (server_2), the leader with the lock
2023-07-26 18:00:12,385 INFO: no action. I am (server_2), the leader with the lock
2023-07-26 18:00:12,423 WARNING: Loop time exceeded, rescheduling immediately.
2023-07-26 18:00:14,027 INFO: no action. I am (server_2), the leader with the lock
   <I suppose this moment of high load on the server>
2023-07-26 18:01:36,990 INFO: Lock owner: server_2; I am server_2
2023-07-26 18:01:37,243 INFO: Updating synchronous privilege temporarily from ['server_1'] to []
2023-07-26 18:01:37,303 INFO: Synchronous replication key updated by someone else.
2023-07-26 18:01:37,349 INFO: no action. I am (server_2), the leader with the lock
2023-07-26 18:01:37,350 WARNING: Loop time exceeded, rescheduling immediately.
2023-07-26 18:01:37,606 INFO: no action. I am (server_2), the leader with the lock
2023-07-26 18:01:37,695 INFO: no action. I am (server_2), the leader with the lock
[...]

PR #2783, opened 2023-07-27
Invalidate cache if txn failed due to revision mismatch

CyberDem0n:
It was reported in #2779 that the primary was constantly logging messages like `Synchronous replication key updated by someone else`.

It happened after Patroni was stuck due to resource starvation. Key updates are performed using create_revision/mod_revision field, which value is taken from the internal cached. Hence, it is a clear symptom of stale cache.

Similar issues in K8s implementation were addressed by invalidating the cache and restarting watcher connections every time when update failed due to resource_version mismatch, so we do the same for Etcd3.

### DIFF
--- a/patroni/dcs/etcd3.py
+++ b/patroni/dcs/etcd3.py
@@ -630,6 +630,16 @@ class PatroniEtcd3Client(Etcd3Client):
 
         return ret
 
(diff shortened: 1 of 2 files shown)


hughcapet:
👍

CyberDem0n:
:+1:

PR #2784, opened 2023-07-27
Ensure strict failover/switchover definition difference

hughcapet:
- allow manual failover to an async node in sync mode
- `patronictl` fixes
  - don't set leader in the `/failover` key for the failover command
  - ask confirmation for an async node failover in the sync mode when no force option provided
  - don't add `scheduled_at` field to the `/failover` key  if not provided
- add REST API check that no `scheduled_at` is passed for failover
- ensure consistency of the terms in the comments, log lines and tests
- refactor and cleanup the failover/switchover unit tests

### TODO
- [x] Failover/switchover docs
- [x] More test cases?
    - [x] https://github.com/zalando/patroni/pull/2784/files#diff-9dada8f509128404d7027246ac9622d2484fc77758c6a4af9dc7cc8c081d9fffR1261-R1266 
- [x] Beautify+extend ctl failover/switchover tests
- [x] Beautify+extend api failover/switchover tests


patroni/ha.py
```diff
@@ -1206,28 +1221,34 @@ def process_manual_failover_from_leader(self) -> Optional[str]:
...
+        # then do nothing
         if not failover or (self.is_paused() and not self.state_handler.is_primary()):
             return
 
+        action = self._get_failover_action_name()
+
+        # it is not the time for the the scheduled failover yet, do nothing
         if (failover.scheduled_at and not
-            self.should_run_scheduled_action("failover", failover.scheduled_at, lambda:
+            self.should_run_scheduled_action(action, failover.scheduled_at, lambda:
```
CyberDem0n, line 1234: I guess this line will look strange: https://github.com/zalando/patroni/blob/82d2ef4878cd327e63fdb25e6faaf47dca3cebbe/patroni/ha.py#L1203

patroni/ha.py
```diff
@@ -921,11 +932,12 @@ def is_failover_possible(self, *, cluster_lsn: int = 0, exclude_failover_candida
...
 
+        action = self._get_failover_action_name()
         if self.is_synchronous_mode() and self.cluster.failover and self.cluster.failover.candidate and not candidates:
-            logger.warning('Failover candidate=%s does not match with sync_standbys=%s',
-                           self.cluster.failover.candidate, self.cluster.sync.sync_standby)
+            logger.warning('%s candidate=%s does not match with sync_standbys=%s',
+                           action.title(), self.cluster.failover.candidate, self.cluster.sync.sync_standby)
         elif not candidates:
-            logger.warning('manual failover: candidates list is empty')
+            logger.warning('%s: candidates list is empty', action)
```
CyberDem0n, line 940: Hmm, but this method is also called from `shutdown()`, and message could be misleading: `Manual failover: candidates list is empty`.

patroni/ctl.py
```diff
@@ -1271,20 +1281,24 @@ def _do_failover_or_switchover(obj: Dict[str, Any], action: str, cluster_name: s
...
     logging.debug(failover_value)
 
     # By now we have established that the leader exists and the candidate exists
     if not force:
-        demote_msg = ', demoting current leader ' + leader if leader else ''
+        demote_msg = f', demoting current leader {cluster.leader.name}' if cluster.leader else ''
         if scheduled_at_str:
-            if not click.confirm('Are you sure you want to schedule {0} of cluster {1} at {2}{3}?'
-                                 .format(action, cluster_name, scheduled_at_str, demote_msg)):
+            assert action == 'switchover'
```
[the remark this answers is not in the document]
hughcapet: we have already agreed on this. But I am still concerned by the fact this assumption is only in our heads and as soon as preceding code is to be changed we can skip a regression 😢 will at least add a comment then

patroni/ha.py
hughcapet: https://github.com/zalando/patroni/pull/2784/files/ba204884d874b1805f48ada50578ae7255761c73#diff-b6a3a0ef64783a6d32feb908d74a3a009e9f25008b8ca5313c9c9747e3b74a63R217-R218 ("manual" can only appear inappropriately if there is an existing failover key during a shutdown/recover/demote/pg staring too long)

hughcapet:
@CyberDem0n do you have any idea about the decreased coverage?

tests/test_ctl.py
```diff
@@ -96,91 +97,123 @@ def test_switchover(self, mock_get_dcs):
...
+        self.assertIn('This cluster has no leader', result.output)
+
+        # Citus cluster, no group number specified
+        result = self.runner.invoke(ctl, ['switchover', 'dummy', '--force'], input='\n')
+        self.assertEqual(result.exit_code, 1)
+        self.assertIn('For Citus clusters the --group must me specified', result.output)
 
     @patch('patroni.ctl.get_dcs')
     @patch.object(PoolManager, 'request', Mock(return_value=MockResponse()))
+    @patch('patroni.ctl.request_patroni', Mock(return_value=urllib3.HTTPResponse()))
```
CyberDem0n, line 196:
```suggestion
    @patch('patroni.ctl.request_patroni', Mock(return_value=MockResponse()))
```

tests/test_ctl.py
```diff
@@ -2,6 +2,7 @@
 import mock
 import os
 import unittest
+import urllib3
```
CyberDem0n, line 5:
```suggestion
```

patroni/ctl.py
```diff
@@ -46,6 +46,7 @@
 except ImportError:  # pragma: no cover
     from cdiff import markup_to_pager, PatchStream  # pyright: ignore [reportMissingModuleSource]
 
+from .config import get_global_config
```
CyberDem0n, line 49:
```suggestion
from .config import Config, get_global_config
```

and remove import from the load_config() method.

patroni/ctl.py
```diff
@@ -1266,25 +1273,28 @@ def _do_failover_or_switchover(obj: Dict[str, Any], action: str, cluster_name: s
...
     # By now we have established that the leader exists and the candidate exists
     if not force:
-        demote_msg = ', demoting current leader ' + leader if leader else ''
+        demote_msg = f', demoting current leader {cluster.leader.name}' if cluster.leader else ''
         if scheduled_at_str:
-            if not click.confirm('Are you sure you want to schedule {0} of cluster {1} at {2}{3}?'
-                                 .format(action, cluster_name, scheduled_at_str, demote_msg)):
+            # only switchover can be scheduled
+            if not click.confirm(f'Are you sure you want to schedule switchover of cluster \
+{cluster_name} at {scheduled_at_str}{demote_msg}?'):
```
matthbakeredb, line 1294:
```suggestion
            if not click.confirm(f'Are you sure you want to schedule switchover of cluster '
                                 f'{cluster_name} at {scheduled_at_str}{demote_msg}?'):
```

patroni/ctl.py
```diff
@@ -1253,7 +1253,14 @@ def _do_failover_or_switchover(obj: Dict[str, Any], action: str, cluster_name: s
...
     if candidate and candidate not in candidate_names:
-        raise PatroniCtlException('Member {0} does not exist in cluster {1}'.format(candidate, cluster_name))
+        raise PatroniCtlException(
+            f'Member {candidate} does not exist in cluster {cluster_name} or is tagged as nofailover')
+
+    if not force and action == 'failover':
+        if global_config.is_synchronous_mode and not cluster.sync.is_empty\
+           and not cluster.sync.matches(candidate, True)\
+           and not click.confirm(f'Are you sure you want to failover to the asynchronous node {candidate}'):
+            raise PatroniCtlException('Aborting ' + action)
```
matthbakeredb, line 1263:
```suggestion
    if not force and action == 'failover':
        if all((
                global_config.is_synchronous_mode,
                not cluster.sync.is_empty,
                not cluster.sync.matches(candidate, True),
                not click.confirm(f'Are you sure you want to failover to the asynchronous node {candidate}')
        )):
            raise PatroniCtlException('Aborting ' + action)
```

matthbakeredb: Or do all the non-interactive checks first:
```suggestion
    if all((
            not force,
            action == 'failover',
            global_config.is_synchronous_mode,
            not cluster.sync.is_empty,
            not cluster.sync.matches(candidate, True),
    )):
        if not click.confirm(f'Are you sure you want to failover to the asynchronous node {candidate}'):
            raise PatroniCtlException('Aborting ' + action)

```

matthbakeredb (changes requested):
Just a few docs and style suggestions.

patroni/ctl.py
```diff
@@ -1318,19 +1330,15 @@ def _do_failover_or_switchover(obj: Dict[str, Any], action: str, cluster_name: s
 @ctl.command('failover', help='Failover to a replica')
 @arg_cluster_name
 @option_citus_group
-@click.option('--leader', '--primary', '--master', 'leader', help='The name of the current leader', default=None)
```
matthbakeredb, line 1321: This is a non-backwards compatible breaking change, so I wonder if we should instead leave the option in but print a warning that it will be removed in a future version. Then remove it when we bump the major version of Patroni to 4. Perhaps we can add a warning message like this?

```
if leader:
  logger.warning('Supplying a leader name using this command is deprecated and will be removed in a future version of Patroni, change your scripts to use `switchover` instead.')
```

patroni/ctl.py
```diff
@@ -1266,25 +1275,28 @@ def _do_failover_or_switchover(obj: Dict[str, Any], action: str, cluster_name: s
...
     if not force:
-        demote_msg = ', demoting current leader ' + leader if leader else ''
+        demote_msg = f', demoting current leader {cluster.leader.name}' if cluster.leader else ''
         if scheduled_at_str:
-            if not click.confirm('Are you sure you want to schedule {0} of cluster {1} at {2}{3}?'
-                                 .format(action, cluster_name, scheduled_at_str, demote_msg)):
+            # only switchover can be scheduled
+            if not click.confirm(f'Are you sure you want to schedule switchover of cluster '
+                                 f'{cluster_name} at {scheduled_at_str}{demote_msg}?'):
                 raise PatroniCtlException('Aborting scheduled ' + action)
```
barthisrael, line 1297:
```suggestion
                raise PatroniCtlException('Aborting scheduled switchover')
```

patroni/ctl.py
```diff
@@ -1266,25 +1275,28 @@ def _do_failover_or_switchover(obj: Dict[str, Any], action: str, cluster_name: s
...
         else:
-            if not click.confirm('Are you sure you want to {0} cluster {1}{2}?'
-                                 .format(action, cluster_name, demote_msg)):
+            if not click.confirm(f'Are you sure you want to {action} cluster {cluster_name}{demote_msg}?'):
```
barthisrael, line 1299:
```suggestion
            if not click.confirm(f'Are you sure you want to perform a {action} on cluster {cluster_name}{demote_msg}?'):
```

patroni/ha.py
```diff
@@ -1980,8 +2003,9 @@ def get_failover_candidates(self, exclude_failover_candidate: bool) -> List[Memb
         exclude = [self.state_handler.name] + ([failover.candidate] if failover and exclude_failover_candidate else [])
 
         def is_eligible(node: Member) -> bool:
-            # TODO: allow manual failover (=no leader specified) to async node
-            if self.sync_mode_is_active() and not self.cluster.sync.matches(node.name):
+            # in synchronous mode we allow failover (not switchover!) to async node
+            if self.sync_mode_is_active() and not self.cluster.sync.matches(node.name)\
+                    and not (failover and not failover.leader):
```
barthisrael, line 2008: Maybe we could implement a couple helper methods on `Failover` class to help deciding what is going on? I mean something like:

* `def is_switchover...`
* `def is_failover...`

Or some other standardized way that would help whenever we want to check these stuffs?

barthisrael (changes requested):
Nice PR!
Put a few comments.

docs/rest_api.rst
```diff
@@ -554,39 +554,111 @@ The above call removes ``postgresql.parameters.max_connections`` from the dynami
...
 
-Depending on the situation the request might finish with a different HTTP status code and body. The status code **200** is returned when the switchover or failover successfully completed. If the switchover was successfully scheduled, Patroni will return HTTP status code **202**. In case something went wrong, the error status code (one of **400**, **412** or **503**) will be returned with some details in the response body. For more information please check the source code of ``patroni/api.py:do_POST_failover()`` method.
+- be reachable via Patroni API,
+- not have ``nofailover`` tag set to ``true``,
+- have watchdog fully functional (if required by the configuration),
+- in case of a switchover or a failover in a healthy cluster, not exceed maximum replication lag (``maximum_lag_on_failover`` :ref:`configuration parameter <dynamic_configuration>`),
+- in case of a switchover or a failover in a healthy cluster, not have a timeline number smaller than the cluster timeline,
+- in :ref:`synchronous mode <synchronous_mode>`:
```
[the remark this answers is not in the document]
hughcapet: can not find the proper punctuation rule for that, wanted to ask @matthbakeredb, actually:)

patroni/ctl.py
hughcapet: I didn't want to change output messages in this PR (otherwise I would change not only this 😅 )

patroni/ha.py
hughcapet: probably, but I am already working on some refactoring of the switchover/failover api in another branch, so maybe we move this suggestion there?

docs/rest_api.rst
```diff
@@ -554,39 +554,111 @@ The above call removes ``postgresql.parameters.max_connections`` from the dynami
...
+.. _switchover_api:
 
-1. The failover endpoint allows to perform a manual failover when there are no healthy nodes, but at the same time it will not allow you to schedule a switchover.
+Switchover
+^^^^^^^^^^
 
-2. The switchover endpoint is the opposite. It works only when the cluster is healthy (there is a leader) and allows to schedule a switchover at a given time.
+``/switchover`` endpoint only works when cluster is healthy (there is a leader). It allows to schedule a switchover at a given time.
 
+When calling ``/switchover`` endpoint candidate can be specified but is not required, in contrast to ``/failover`` endpoint. If candidate is not provided, all the healthy nodes that are allowed to failover participate in the leader race.
```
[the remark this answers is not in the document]
hughcapet: maybe you suggest a new version of the sentence then? I want it to be easy to understand:)

docs/rest_api.rst
```diff
@@ -554,39 +554,111 @@ The above call removes ``postgresql.parameters.max_connections`` from the dynami
...
 .. code-block:: bash
 
-    $ curl -s http://localhost:8008/switchover -XPOST -d \
-	    '{"leader":"postgresql0","scheduled_at":"2019-09-24T12:00+00"}'
-    Switchover scheduled
+	$ curl -s http://localhost:8008/failover -XPOST -d '{"candidate":"postgresql1"}'
+	Successfully failed over to "postgresql1"
+
+.. warning::
+	:ref:`Be very careful <failover_healthcheck>` using this endpoint, as this can cause data loss in certain situations. In most cases, :ref:`the switchover endpoint <switchover_api>` satisfies the administrator's needs.
```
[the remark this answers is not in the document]
hughcapet: on? isn't it "when"/"about"?

patroni/ctl.py
```diff
@@ -1176,8 +1173,8 @@ def reinit(obj: Dict[str, Any], cluster_name: str, group: Optional[int],
 
 
 def _do_failover_or_switchover(obj: Dict[str, Any], action: str, cluster_name: str,
-                               group: Optional[int], leader: Optional[str], candidate: Optional[str],
-                               force: bool, scheduled: Optional[str] = None) -> None:
+                               group: Optional[int], candidate: Optional[str], force: bool,
+                               leader: Optional[str] = None, scheduled: Optional[str] = None) -> None:
```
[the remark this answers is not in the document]
hughcapet: leader now has the default value?

docs/rest_api.rst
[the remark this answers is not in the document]
barthisrael: I put a suggestion above. I was just wondering if the suggestion made sense or not 😆 

The thing that I found a bit weird was "allowed to failover participate".

docs/rest_api.rst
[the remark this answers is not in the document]
barthisrael: Right, I thought of "when" and typed "on" -- #BlameIsrael

patroni/ctl.py
hughcapet: idk, honestly. wanted to have it at least somewhere here to be able to catch a regression in the future. maybe formatting with "action" var everywhere then...

docs/rest_api.rst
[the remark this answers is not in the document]
barthisrael: Ah, I see. As Matt lives in England, maybe he knows better how that works in English 😆 
What I mentioned above is very usual in Portuguese -- people write even laws that way.
I could find a reference about that here too -- https://www.grammarbook.com/blog/semicolons/semicolon-use-in-lists/#:~:text=To%20do%20this,of%20each%20line.
But I'm not sure how reliable that reference is 😆

patroni/ctl.py
barthisrael: Makes sense, we can submit that as a separate change.

patroni/ha.py
barthisrael: Makes sense too, wfm.

patroni/ctl.py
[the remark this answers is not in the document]
barthisrael: Oh, I didn't notice you add a ` = None`, sorry.

patroni/ctl.py
barthisrael: Hmm, right. Either that or put a comment in the code, IDK.
I made the above suggestion just to follow the pattern of your changes.

docs/rest_api.rst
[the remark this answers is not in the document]
hughcapet: Ok, got it. Well, the idea of the sentence is: 
only the nodes that are:
- healthy (don't lag, are reachable, don't have watchdog failed if one requested)
- allowed to promote (have no limitations like 'nofailover' tag, are in the /sync key if in sync mode)

Maybe then
`If candidate is not provided, all the healthy nodes of the cluster that are allowed to promote participate in the leader race.`

This also makes me think we could have two separate methods for "health" and "allowance" evaluations but it is more for the refactoring

patroni/ctl.py
hughcapet: yes, I can do that but I am not sure if we need 4 for this. is 3.2 not enough?
And to be honest, just the option removal is not the most breaking change here... We also have "allow manual failover to an async node in sync mode" and it basically will not be possible to use `patronictl failover` as a replacement for unscheduled `patronictl switchover` anymore. So I wonder if the warning with the leader value discarding will actually hide the extent of the changes... 
@CyberDem0n wdyt

docs/rest_api.rst
```diff
@@ -554,39 +554,111 @@ The above call removes ``postgresql.parameters.max_connections`` from the dynami
...
+   * - Can be scheduled
+     - no
+     - yes (if not in pause)
+
+.. _failover_healthcheck:
+
+Healthy standby
+^^^^^^^^^^^^^^^
 
+There are a couple of checks that a member of a cluster should pass to be able to participate in the leader race during a switchover or to become a leader as a failover/switchover candidate:
```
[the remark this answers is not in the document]
CyberDem0n: > I know the primary is not well but Patroni doesn't see it

It doesn't fully describe the situation.
Usually Patroni knows that there is no primary, but sometimes it can't do anything about it because there are no members that satisfy all conditions to be automatically selected as a new primary.

patroni/ctl.py
CyberDem0n: Actually, I like the idea about showing a warning. Not sure whether we should discard the provided leader value or not.
Waiting for 4 shouldn't be a problem. There is another "breaking" change pending - replacing "master" with primary everywhere.

hughcapet: > Not sure whether we should discard the provided leader value or not.

if we don't - the whole PR doesn't make much sense

hughcapet: I agree adding the warning is a good idea but it should more descriptive about "it is not be possible to use patronictl failover as a replacement for unscheduled patronictl switchover anymore" Basically, big part of the `patronictl failover` behaviour has changed

docs/rest_api.rst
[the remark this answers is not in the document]
matthbakeredb: Haha, being English is not a qualification of the proper use of the language! We're generally terrible at it. However, when I first saw the suggestion it seemed odd to me, at first. I think that it makes sense where each item is a continuation of an initial sentence and breaks something that could have sub-lists within the sentence. Like how it's shown in the example.

This consists a list of related things:

* this one and this one, and another; or
* these, and this one too.

I also think that commas and periods are fine, with the latter more suitable where the line forms a full sentence. If there is one word or phrase in the list it probably would be fine with it being a comma. However, this is informal language, as the link Israel provided mentions it being for formal text, it might be more appropriate to use a semicolon.

In short, I am not 100% sure on the use of the semicolon, I wouldn't get too hung up on it though. I don't have a problem with this suggestion.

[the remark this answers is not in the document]
matthbakeredb: Perhaps we should employ a grammar and style linter so that we have an automated rules system to save us from needing to worry about it. Like vale https://github.com/errata-ai/vale/#usage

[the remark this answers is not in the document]
matthbakeredb: This is a good resource https://docs.gitlab.com/ee/development/documentation/testing.html#vale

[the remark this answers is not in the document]
matthbakeredb: With the gitlab style guide vale reports this for the docs/ directory after running sphinx (so it includes the api docs)

```
67 errors, 2197 warnings, and 511 suggestions found.
```

patroni/ctl.py
matthbakeredb: > > Not sure whether we should discard the provided leader value or not.
> 
> if we don't - the whole PR doesn't make much sense

I think that it's sometimes good to introduce a bunch of code to support a new way of doing things which introduces a breaking change, which will do all the work but defaults to not making that change at first. Either the code is not called or is available as an "opt-in" configuration. A warning shows that said code will become the new standard and allow people time to switch. Then finally it becomes active by switching the config option to be the default or just deactivating the old way but not removing the code. If the old way is still needed then it can be activated by a user. Then in a future version that feature can be removed when the user has had time to migrate.

You might not need all that song-and-dance for this, but I don't think the change is wasted.

matthbakeredb: > yes, I can do that but I am not sure if we need 4 for this. is 3.2 not enough?

That would not be using semantic versioning. Major version change is needed when the code is no longer compatible with a user's use of it. I say wait for 4 as you might want to queue up a few changes to do together so we're not on v143 by Christmas ;-)

docs/rest_api.rst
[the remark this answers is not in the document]
hughcapet: yes, was also thinking of adopting some linter for that, thanks for the link - will take a look

docs/pause.rst
```diff
@@ -19,7 +19,7 @@ When Patroni runs in a paused mode, it does not change the state of PostgreSQL,
 
 - For the Postgres primary with the leader lock Patroni updates the lock. If the node with the leader lock stops being the primary (i.e. is demoted manually), Patroni will release the lock instead of promoting the node back.
 
-- Manual unscheduled restart, reinitialize and manual failover are allowed. Manual failover is only allowed if the node to failover to is specified. In the paused mode, manual failover does not require a running primary node.
+- Manual unscheduled restart, manual unscheduled failover/switchover and reinitialize are allowed. No scheduled action is allowed. Manual switchover is only allowed if the node to switchover to is specified.
```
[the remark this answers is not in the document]
matthbakeredb: My inclination is to use `to`, as it implies the movement towards something.

patroni/ctl.py
hughcapet: Ok, giving up, you've convinced me:) 
so something like this? https://github.com/zalando/patroni/pull/2784/commits/cc336fc1fe08d1514fb2026faf516b2c87a0a779

matthbakeredb: Looks good! Added a comment.

docs/rest_api.rst
```diff
@@ -554,39 +554,111 @@ The above call removes ``postgresql.parameters.max_connections`` from the dynami
...
 
-``POST /switchover`` or ``POST /failover``. These endpoints are very similar to each other. There are a couple of minor differences though:
+.. _switchover_api:
 
-1. The failover endpoint allows to perform a manual failover when there are no healthy nodes, but at the same time it will not allow you to schedule a switchover.
+Switchover
+^^^^^^^^^^
 
-2. The switchover endpoint is the opposite. It works only when the cluster is healthy (there is a leader) and allows to schedule a switchover at a given time.
+``/switchover`` endpoint only works when the cluster is healthy (there is a leader). It allows to schedule a switchover at a given time.
```
CyberDem0n, line 562:
```suggestion
``/switchover`` endpoint only works when the cluster is healthy (there is a leader). It also allows to schedule a switchover at a given time.
```

docs/rest_api.rst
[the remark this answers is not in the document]
CyberDem0n: What about the following?
```suggestion
When calling ``/switchover`` endpoint a candidate can be specified but is not required, in contrast to ``/failover`` endpoint. If a candidate is not provided, all eligible nodes of the cluster will participate in the leader race after the leader stepped down.
```

docs/rest_api.rst
```diff
@@ -554,39 +554,111 @@ The above call removes ``postgresql.parameters.max_connections`` from the dynami
...
 
+When calling ``/switchover`` endpoint candidate can be specified but is not required, in contrast to ``/failover`` endpoint. If candidate is not provided, all the healthy nodes that are allowed to failover participate in the leader race.
 
-In the JSON body of the ``POST`` request you must specify at least the ``leader`` or ``candidate`` fields and optionally the ``scheduled_at`` field if you want to schedule a switchover at a specific time.
+In the JSON body of the ``POST`` request, you must specify at least the ``leader`` field and, optionally, the ``candidate`` and ``scheduled_at`` field if you want to schedule a switchover at a specific time.
```
CyberDem0n, line 566:
```suggestion
In the JSON body of the ``POST`` request, you must specify the ``leader`` field, optionally the ``candidate``, and optionally the ``scheduled_at`` field if you want to schedule a switchover at a specific time.
```

docs/rest_api.rst
```diff
@@ -554,39 +554,111 @@ The above call removes ``postgresql.parameters.max_connections`` from the dynami
...
+Healthy standby
+^^^^^^^^^^^^^^^
 
+There are a couple of checks that a member of a cluster should pass to be able to participate in the leader race during a switchover or to become a leader as a failover/switchover candidate:
 
-Depending on the situation the request might finish with a different HTTP status code and body. The status code **200** is returned when the switchover or failover successfully completed. If the switchover was successfully scheduled, Patroni will return HTTP status code **202**. In case something went wrong, the error status code (one of **400**, **412** or **503**) will be returned with some details in the response body. For more information please check the source code of ``patroni/api.py:do_POST_failover()`` method.
+- be reachable via Patroni API;
+- not have ``nofailover`` tag set to ``true``;
+- have watchdog fully functional (if required by the configuration);
+- in case of a switchover or a failover in a healthy cluster, not exceed maximum replication lag (``maximum_lag_on_failover`` :ref:`configuration parameter <dynamic_configuration>`);
```
CyberDem0n, line 650:
```suggestion
- in case of a switchover in a healthy cluster or automatic failover, not exceed maximum replication lag (``maximum_lag_on_failover`` :ref:`configuration parameter <dynamic_configuration>`);
```

docs/rest_api.rst
```diff
@@ -554,39 +554,111 @@ The above call removes ``postgresql.parameters.max_connections`` from the dynami
...
+- in case of a switchover or a failover in a healthy cluster, not have a timeline number smaller than the cluster timeline;
```
CyberDem0n, line 651:
```suggestion
- in case of a switchover in a healthy cluster or automatic failover, not have a timeline number smaller than the cluster timeline if ``check_timeline`` :ref:`configuration parameter <dynamic_configuration>` is set to ``true``;
```

docs/rest_api.rst
```diff
@@ -554,39 +554,111 @@ The above call removes ``postgresql.parameters.max_connections`` from the dynami
...
+- in :ref:`synchronous mode <synchronous_mode>`:
 
-- ``DELETE /switchover``: delete the scheduled switchover
+  - In case of a switchover (both with and without a candidate): be listed in the ``/sync`` key members;
+  - For a failover in both healthy and unhealthy clusters, this check is omitted.
 
-The ``POST /switchover`` and ``POST failover`` endpoints are used by ``patronictl switchover`` and ``patronictl failover``, respectively.
-The ``DELETE /switchover`` is used by ``patronictl flush <cluster-name> switchover``.
+.. warning::
+    In case of a failover in a cluster without a leader, a candidate will be allowed to promote even if:
```
CyberDem0n, line 658:
```suggestion
    In case of a manual failover in a cluster without a leader, a candidate will be allowed to promote even if:
```

docs/rest_api.rst
```diff
@@ -554,39 +554,111 @@ The above call removes ``postgresql.parameters.max_connections`` from the dynami
...
+	- it is not in the ``/sync`` key members when synchronous mode is enabled;
+	- its lag exceeds the maximum replication lag allowed;
+	- it has the timeline number smaller than the cluster timeline.
```
CyberDem0n, line 661:
```suggestion
	- it has the timeline number smaller than the last known cluster timeline.
```

docs/rest_api.rst
hughcapet: don't fully get it. you want to emphasise that both are optional?

patroni/ctl.py
hughcapet: extended the docstring

docs/rest_api.rst
CyberDem0n: Yes, they are both optional and I want to emphasize it.

tests/test_ha.py
```diff
@@ -687,110 +689,289 @@ def test_restart_in_progress(self):
...
-        self.ha.cluster = get_cluster_initialized_without_leader(failover=Failover(0, '', self.p.name, None))
-        self.ha.cluster = get_cluster_initialized_without_leader(failover=Failover(0, '', 'leader', None))
         self.p.set_role('replica')
-        self.assertEqual(self.ha.run_cycle(), 'promoted self to leader by acquiring session lock')
-        self.ha.fetch_node_status = get_node_status()  # accessible, in_recovery
-        self.assertEqual(self.ha.run_cycle(), 'following a different leader because i am not the healthiest node')
-        self.ha.cluster = get_cluster_initialized_without_leader(failover=Failover(0, self.p.name, '', None))
-        self.assertEqual(self.ha.run_cycle(), 'following a different leader because i am not the healthiest node')
-        self.ha.fetch_node_status = get_node_status(reachable=False)  # inaccessible, in_recovery
+
```
matthbakeredb, line 873: Are the separate commented sections in this test (and others) steps, i.e. the previous step is necessary before the next step can be done, or is each the same operation but with a different configuration? It looks like they're similar. I don't worry too much about tests normally but if it's the latter then perhaps the test method could be done with [mark parameterize](https://docs.pytest.org/en/7.3.x/how-to/parametrize.html). Then you could have the unique code written once and have a clear indication of the different configurations at the top of the test.

matthbakeredb (approved):
It looks good to me. I didn't go too deep into a review this time but have read a few times already.

I made a comment on the test structure but that could be a job for another time if you think it's relevant/useful.

tests/test_ha.py
hughcapet: hm, I doubt it will be possible to implement at all in most cases and, more importantly, will be more "readable" but I will try to adopt this next time

hughcapet:
👍

tests/test_ha.py
hughcapet: https://docs.pytest.org/en/stable/how-to/unittest.html#pytest-features-in-unittest-testcase-subclasses

PR #2786, opened 2023-07-27
[WIP] Implement generate-config

hughcapet:
new `patroni.py` option that allows to
- generate `patroni.yml` configuration file with the values from a running cluster
- generate a sample `patroni.yml` configuration file

Documentation will follow in a separate PR

Resolve https://github.com/zalando/patroni/issues/391

### DIFF
--- a/patroni/__main__.py
+++ b/patroni/__main__.py
@@ -7,6 +7,7 @@ import time
 from argparse import Namespace
 from typing import Any, Dict, Optional, TYPE_CHECKING
 
+from patroni.config_generator import generate_config
 from patroni.daemon import AbstractPatroniDaemon, abstract_main, get_base_arg_parser
 
 if TYPE_CHECKING:  # pragma: no cover
@@ -164,10 +165,23 @@ def patroni_main(configfile: str) -> None:
 
 def process_arguments() -> Namespace:
     parser = get_base_arg_parser()
-    parser.add_argument('--validate-config', action='store_true', help='Run config validator and exit')
+    group = parser.add_mutually_exclusive_group()
+    group.add_argument('--validate-config', action='store_true', help='Run config validator and exit')
+    group.add_argument('--generate-sample-config', action='store_true',
+                       help='Generate a sample Patroni yaml configuration file')
+    group.add_argument('--generate-config', action='store_true',
+                       help='Generate a Patroni yaml configuration file for a running instance')
+    parser.add_argument('--dsn', help='Optional DSN string of the instance to be used as a source \
+                                    for config generation. Superuser connection is required.')
     args = parser.parse_args()
 
-    if args.validate_config:
+    if args.generate_sample_config:
+        generate_config(args.configfile or '/tmp/patroni.yml', True, None)
+        sys.exit(0)
+    elif args.generate_config:
+        generate_config(args.configfile or '/tmp/patroni.yml', False, args.dsn)
+        sys.exit(0)
+    elif args.validate_config:
         from patroni.validator import schema
         from patroni.config import Config, ConfigParseError
 
--- a/patroni/collections.py
+++ b/patroni/collections.py
@@ -3,7 +3,7 @@
 Provides a case insensitive :class:`dict` and :class:`set` object types.
 """
 from collections import OrderedDict
-from typing import Any, Collection, Dict, Iterator, MutableMapping, MutableSet, Optional
+from typing import Any, Collection, Dict, Iterator, KeysView, MutableMapping, MutableSet, Optional
 
 
 class CaseInsensitiveSet(MutableSet[str]):
@@ -186,6 +186,9 @@ class CaseInsensitiveDict(MutableMapping[str, Any]):
         :return: a new dict object with the same keys and values of this dict.
         """
         return CaseInsensitiveDict({v[0]: v[1] for v in self._values.values()})
+
+    def keys(self) -> KeysView[str]:
+        return self._values.keys()
 
     def __repr__(self) -> str:
         """Get a string representation of the dict.
--- a/patroni/config.py
+++ b/patroni/config.py
@@ -32,6 +32,20 @@ _AUTH_ALLOWED_PARAMETERS = (
     'channel_binding'
 )
 
+AUTH_ALLOWED_PARAMETERS_MAPPING = {
+    'user': 'PGUSER',
+    'password': 'PGPASSWORD',
+    'sslmode': 'PGSSLMODE',
+    'sslcert': 'PGSSLCERT',
+    'sslkey': 'PGSSLKEY',
+    'sslpassword': '',
+    'sslrootcert': 'PGSSLROOTCERT',
+    'sslcrl': 'PGSSLCRL',
+    'sslcrldir': 'PGSSLCRLDIR',
+    'gssencmode': 'PGGSSENCMODE',
+    'channel_binding': 'PGCHANNELBINDING'
+}
+
 
 def default_validator(conf: Dict[str, Any]) -> List[str]:
     if not conf:
@@ -193,10 +207,9 @@ class Config(object):
             'recovery_min_apply_delay': ''
         },
         'postgresql': {
-            'bin_dir': '',
             'use_slots': True,
             'parameters': CaseInsensitiveDict({p: v[0] for p, v in ConfigHandler.CMDLINE_OPTIONS.items()
-                                               if p not in ('wal_keep_segments', 'wal_keep_size')})
+                                               if v[0] is not None and p not in ('wal_keep_segments', 'wal_keep_size')})
         }
     }
 
@@ -233,6 +246,14 @@ class Config(object):
     @property
     def dynamic_configuration(self) -> Dict[str, Any]:
         return deepcopy(self._dynamic_configuration)
+
+    @property
+    def local_configuration(self) -> Dict[str, Any]:
+        return deepcopy(dict(self._local_configuration))
+
+    @classmethod
+    def get_default_config(cls) -> Dict[str, Any]:
+        return deepcopy(cls.__DEFAULT_CONFIG)
 
     def _load_config_path(self, path: str) -> Dict[str, Any]:
         """
@@ -334,7 +355,7 @@ class Config(object):
                 or not is_local and ConfigHandler.CMDLINE_OPTIONS[name][1](value)}
 
     def _safe_copy_dynamic_configuration(self, dynamic_configuration: Dict[str, Any]) -> Dict[str, Any]:
-        config = deepcopy(self.__DEFAULT_CONFIG)
+        config = self.get_default_config()
 
         for name, value in dynamic_configuration.items():
             if name == 'postgresql':
--- a/patroni/config_generator.py
+++ b/patroni/config_generator.py
@@ -1,0 +1,286 @@
+"""patroni --generate-config machinery."""
+import os
+import socket
+import sys
+import yaml
+
+from typing import Any, Dict, Optional
+
+from .config import Config
+from .exceptions import PatroniException
+from .postgresql.config import ConfigHandler
+from .postgresql.misc import postgres_major_version_to_int
+from .utils import get_major_version, patch_config
+
+
+def get_ip() -> str:
+    """
+    Try to get an ip address of the hostname returned by :func:`gethostname`.
+
+    .. note::
+            Can also return local ip
+
+    :returns: the first element in the sorted list of the addresses returned by :func:`getaddrinfo`.
+              Sorting guarantees it will prefer ipv4.
+    """
+    hostname = None
+    try:
+        hostname = socket.gethostname()
+        return sorted(socket.getaddrinfo(hostname, 0, socket.AF_UNSPEC, socket.SOCK_STREAM, 0),
+                      key=lambda x: x[0])[0][4][0]
+    except OSError as e:
+        sys.exit(f'Failed to define ip address: {e}')
+    except IndexError:
+        sys.exit(f'Failed to define ip address. No address returned by getaddrinfo for {hostname}')
+
+
+def get_int_major_version(config: Optional[Dict[str, Any]] = None) -> int:
+    """Get major version of PostgreSQL from the binary as an integer.
+
+    :param config: optional dictionary representing Patroni config. If contains bin_dir and/or the custom
+    bin_name for postgres, the values are used for the version retrieval.
+    """
+    config = config or {}
+    postgres_bin = ((config.get('postgresql') or {}).get('bin_name') or {}).get('postgres', 'postgres')
+    return postgres_major_version_to_int(get_major_version(config['postgresql'].get('bin_dir') or None, postgres_bin))
+
+
+def get_bin_dir_from_running_instance(data_dir: str) -> str:
+    """Define the directory postgres binaries reside using postmaster's pid executable."""
+    postmaster_pid = None
+    try:
+        with open(f"{data_dir}/postmaster.pid", 'r') as f:
+            postmaster_pid = f.readline()
+            if not postmaster_pid:
+                sys.exit('Failed to obtain postmaster pid from postmaster.pid file')
+            postmaster_pid = int(postmaster_pid.strip())
+    except OSError as e:
+        sys.exit(f'Error while reading postmaster.pid file: {e}')
+    import psutil
+    try:
+        return os.path.dirname(psutil.Process(postmaster_pid).exe())
+    except psutil.NoSuchProcess:
+        sys.exit('Obtained postmaster pid doesn\'t exist')
+
+
+def enrich_config_from_running_instance(config: Dict[str, Any], no_value_msg: str, dsn: Optional[str] = None) -> None:
+    """Extend the passed ``config`` dictionary with the values gathered from a running instance.
+
+    Get
+    - non-internal GUC values having configuration file, postmaster command line or environment variable as a source
+    - postgresql.connect_address, postgresql.listen,
+    - postgresql.pg_hba and postgresql.pg_ident if hba_file/ident_file is set to the default value
+    - superuser auth parameters (from the options used for connection)
+    And redefine scope with the clister_name GUC value if set
+
+    :param config: configuration parameters dict to be enriched
+    :param no_value_msg: str value to be used when a parameter value is not available
+    :param dsn: optional DSN string for the source running instance
+    """
+    from getpass import getuser, getpass
+    from patroni.postgresql.config import parse_dsn
+    from patroni.config import AUTH_ALLOWED_PARAMETERS_MAPPING
+
+    su_params: Dict[str, str] = {}
+    parsed_dsn = {}
+
+    if dsn:
+        parsed_dsn = parse_dsn(dsn)
+        if not parsed_dsn:
+            sys.exit('Failed to parse DSN string')
+
+    # gather auth parameters for the superuser config
+    for conn_param, env_var in AUTH_ALLOWED_PARAMETERS_MAPPING.items():
+        val = parsed_dsn.get(conn_param, os.getenv(env_var))
+        if val:
+            su_params[conn_param] = val
+    patroni_env_su_username = ((config.get('authentication') or {}).get('superuser') or {}).get('username')
+    patroni_env_su_pwd = ((config.get('authentication') or {}).get('superuser') or {}).get('password')
+    # because we use "username" in the config for some reason
+    su_params['username'] = su_params.pop('user', patroni_env_su_username) or getuser()
+    su_params['password'] = su_params.get('password', patroni_env_su_pwd) or getpass('Please enter the user password:')
+
+    from . import psycopg
+    try:
+        conn = psycopg.connect(dsn=dsn, password=su_params['password'])
+    except psycopg.Error as e:
+        sys.exit(f'Failed to establish PostgreSQL connection: {e}')
+
+    with conn.cursor() as cur:
+        cur.execute("SELECT 1 FROM pg_roles WHERE rolname=%s AND rolsuper='t';", (su_params['username'],))
+        if cur.rowcount < 1:
+            sys.exit('The provided user does not have superuser privilege')
+
+        required_params = ['hba_file', 'ident_file', 'config_file',
+                           'data_directory'] + list(ConfigHandler.CMDLINE_OPTIONS.keys())
+        cur.execute("SELECT name, current_setting(name) FROM pg_settings \
+                     WHERE context <> 'internal' \
+                     AND source IN ('configuration file', 'command line', 'environment variable') \
+                     AND category <> 'Write-Ahead Log / Recovery Target' \
+                     AND setting <> '(disabled)' \
+                     OR name = ANY(%s);", (required_params,))
+
+        helper_dict = dict.fromkeys(['port', 'listen_addresses'])
+        # adjust values
+        for p, v in cur.fetchall():
+            if p == 'data_directory':
+                config['postgresql']['data_dir'] = v
+            elif p == 'cluster_name' and v:
+                config['scope'] = v
+            elif p in ('archive_command', 'restore_command', 'archive_cleanup_command',
+                       'recovery_end_command', 'ssl_passphrase_command',
+                       'hba_file', 'ident_file', 'config_file'):
+                # write commands to the local config due to security implications
+                # write hba/ident/config_file to local config to ensure they are not removed later
+                config['postgresql'].setdefault('parameters', {})
+                config['postgresql']['parameters'][p] = v
+            elif p in helper_dict:
+                helper_dict[p] = v
+            else:
+                config['bootstrap']['dcs']['postgresql']['parameters'][p] = v
+
+    conn.close()
+
+    connect_port = parsed_dsn.get('port', os.getenv('PGPORT', helper_dict['port']))
+    config['postgresql']['connect_address'] = f'{get_ip()}:{connect_port}'
+    config['postgresql']['listen'] = f'{helper_dict["listen_addresses"]}:{helper_dict["port"]}'
+
+    # it makes sense to define postgresql.pg_hba/pg_ident only if hba_file/ident_file are set to defaults
+    default_hba_path = os.path.join(config['postgresql']['data_dir'], 'pg_hba.conf')
+    if config['postgresql']['parameters']['hba_file'] == default_hba_path:
+        try:
+            allowed_records = ['local', 'host', 'hostssl', 'hostnossl', 'hostgssenc', 'hostnogssenc']
+            if get_int_major_version(config) >= 16:
+                allowed_records += ['include', 'include_if_exists', 'include_dir']
+            with open(default_hba_path, 'r') as f:
+                config['postgresql']['pg_hba'] = list(filter(lambda i: i and i.split()[0] in allowed_records,
+                                                             (li.strip() for li in f.readlines())))
+        except OSError as e:
+            sys.exit(f'Failed to read pg_hba.conf: {e}')
+
+    default_ident_path = os.path.join(config['postgresql']['data_dir'], 'pg_ident.conf')
+    if config['postgresql']['parameters']['ident_file'] == default_ident_path:
+        try:
+            with open(default_ident_path, 'r') as f:
+                config['postgresql']['pg_ident'] = [i.strip() for i in f.readlines()
+                                                    if i.strip() and not i.startswith('#')]
+        except OSError as e:
+            sys.exit(f'Failed to read pg_ident.conf: {e}')
+        if not config['postgresql']['pg_ident']:
+            del config['postgresql']['pg_ident']
+
+    config['postgresql']['authentication'] = {
+        'superuser': su_params,
+        'replication': {'username': no_value_msg, 'password': no_value_msg}
+    }
+
+
+def generate_config(file: str, sample: bool, dsn: Optional[str]) -> None:
+    """Generate Patroni configuration file.
+
+    Gather all the available non-internal GUC values having configuration file, postmaster command line or environment
+    variable as a source and store them in the appropriate part of Patroni configuration (``postgresql.parameters`` or
+    ``bootsrtap.dcs.postgresql.parameters``). Either the provided DSN (takes precedence) or PG ENV vars will be used
+    for the connection. If password is not provided, it should be entered via prompt.
+
+    The created configuration contains:
+    - ``scope``: cluster_name GUC value or PATRONI_SCOPE ENV variable value if available
+    - ``name``: PATRONI_NAME ENV variable value if set, otherewise hostname
+    - ``bootsrtap.dcs``: section with all the parameters (incl. the majority of PG GUCs) set to their default values
+      defined by Patroni and adjusted by the source instances's configuration values.
+    - ``postgresql.parameters``: the source instance's archive_command, restore_command, archive_cleanup_command,
+      recovery_end_command, ssl_passphrase_command, hba_file, ident_file, config_file GUC values
+    - ``postgresql.bin_dir``: path to Postgres binaries gathered from the running instance or, if not available,
+      the value of PATRONI_POSTGRESQL_BIN_DIR ENV variable. Otherwise, an empty string.
+    - ``postgresql.datadir``: the value gathered from the corresponding PG GUC
+    - ``postgresql.listen``: source instance's listen_addresses and port GUC values
+    - ``postgresql.connect_address``: if possible, generated from the connection params
+    - ``postgresql.authentication``:
+        - superuser and replication users defined (if possible, usernames are set from the respective Patroni ENV vars,
+          otherwise the default 'postgres' and 'replicator' values are used).
+          If not a sample config, either DSN or PG ENV vars are used to define superuser authentication parameters.
+        - rewind user is defined for a sample config if PG version can be defined and PG version is 11+
+          (if possible, username is set from the respective Patroni ENV var)
+    - ``bootsrtap.dcs.postgresql.use_pg_rewind``
+    - ``postgresql.pg_hba`` defaults or the lines gathered from the source instance's hba_file
+    - ``postgresql.pg_ident`` the lines gathered from the source instance's ident_file
+
+    :param file: Full path to the configuration file to be created (/tmp/patroni.yml by default).
+    :param sample: Optional flag. If set, no source instance will be used - generate config with some sane defaults.
+    :param dsn: Optional DSN string for the local instance to get GUC values from.
+    """
+    no_value_msg = '#FIXME'
+    pg_version = None
+    local_ip = get_ip()
+
+    config = Config('', None).local_configuration  # Get values from env
+    dynamic_config = Config.get_default_config()
+    dynamic_config['postgresql']['parameters'] = dict(dynamic_config['postgresql']['parameters'])
+    config.setdefault('bootstrap', {})['dcs'] = dynamic_config
+    config.setdefault('postgresql', {})
+
+    template_config: Dict[str, Any] = {
+        'scope': no_value_msg,
+        'name': socket.gethostname(),
+        'postgresql': {
+            'data_dir': no_value_msg,
+            'connect_address': no_value_msg + ':5432',
+            'listen': no_value_msg + ':5432',
+            'bin_dir': '',
+            'authentication': {
+                'superuser': {
+                    'username': 'postgres',
+                    'password': no_value_msg
+                },
+                'replication': {
+                    'username': 'replicator',
+                    'password': no_value_msg
+                }
+            }
+        },
+        'restapi': {
+            'connect_address': local_ip + ':8008',
+            'listen': local_ip + ':8008'
+        }
+    }
+
+    if not sample:
+        enrich_config_from_running_instance(config, no_value_msg, dsn)
+        config['postgresql']['bin_dir'] = get_bin_dir_from_running_instance(config['postgresql']['data_dir'])
+
+    try:
+        pg_version = get_int_major_version(config)
+    except PatroniException as e:
+        sys.exit(str(e))
+
+    patch_config(template_config, config)
+    config = template_config
+
+    # generate sample config
+    if sample:
+        auth_method = 'scram-sha-256' if pg_version and pg_version >= 100000 else 'md5'
+        config['postgresql']['parameters'] = {'password_encryption': auth_method}
+        config['postgresql']['pg_hba'] = [
+            f'host all all all {auth_method}',
+            f'host replication {config["postgresql"]["authentication"]["replication"]["username"]} all {auth_method}'
+        ]
+
+        # add version-specific configuration
+        wal_keep_param = 'wal_keep_segments' if pg_version < 130000 else 'wal_keep_size'
+        config['bootstrap']['dcs']['postgresql']['parameters'][wal_keep_param] =\
+            ConfigHandler.CMDLINE_OPTIONS[wal_keep_param][0]
+
+        config['bootstrap']['dcs']['postgresql']['use_pg_rewind'] = True
+        if pg_version >= 110000:
+            config['postgresql']['authentication'].setdefault(
+                'rewind', {'username': 'rewind_user'}).setdefault(
+                    'password', no_value_msg)
+
+    # redundant values from the default config
+    del config['bootstrap']['dcs']['standby_cluster']
+
+    dir_path = os.path.dirname(file)
+    if dir_path and not os.path.isdir(dir_path):
+        os.makedirs(dir_path)
+    with open(file, 'w') as fd:
+        yaml.safe_dump(config, fd, default_flow_style=False)
--- a/patroni/utils.py
+++ b/patroni/utils.py
@@ -16,6 +16,7 @@ import platform
 import random
 import re
 import socket
+import subprocess
 import sys
 import tempfile
 import time
@@ -977,3 +978,33 @@ def unquote(string: str) -> str:
     except ValueError:
         ret = string
     return ret
+
+
+def get_major_version(bin_dir: Optional[str] = None, bin_name: str = 'postgres') -> str:
+    """Get the major version of PostgreSQL.
+
+    It is based on the output of ``postgres --version``.
+
+    :param bin_dir: path to PostgreSQL binaries directory. If ``None`` it will use the first ``postgres`` binary that
+        is found by subprocess in the ``PATH``.
+    :param bin_name: name of the postgres binary to call ('postgres' by default)
+    :returns: the PostgreSQL major version.
+    :raises :class:`patroni.exceptions.PatroniException`: if the postgres binary call failed due to OSError
+
+    :Example:
+
+        * Returns `9.6` for PostgreSQL 9.6.24
+        * Returns `15` for PostgreSQL 15.2
+    """
+    if not bin_dir:
+        binary = bin_name
+    else:
+        binary = os.path.join(bin_dir, bin_name)
+    try:
+        version = subprocess.check_output([binary, '--version']).decode()
+    except OSError as e:
+        raise PatroniException(f'Failed to get postgres version: {e}')
+    version = re.match(r'^[^\s]+ [^\s]+ (\d+)(\.(\d+))?', version)
+    if TYPE_CHECKING:  # pragma: no cover
+        assert version is not None
+    return '.'.join([version.group(1), version.group(3)]) if int(version.group(1)) < 10 else version.group(1)
--- a/tests/test_patroni.py
+++ b/tests/test_patroni.py
@@ -1,17 +1,22 @@
 import etcd
 import logging
 import os
+import psutil
 import signal
 import time
 import unittest
 
+from copy import deepcopy
+
+from . import MockCursor
 import patroni.config as config
 from http.server import HTTPServer
-from mock import Mock, PropertyMock, patch
+from mock import MagicMock, Mock, PropertyMock, mock_open, patch
 from patroni.api import RestApiServer
 from patroni.async_executor import AsyncExecutor
 from patroni.dcs import Cluster, Member
 from patroni.dcs.etcd import AbstractEtcdClientWithFailover
+from patroni.config import Config
 from patroni.exceptions import DCSError
 from patroni.postgresql import Postgresql
 from patroni.postgresql.config import ConfigHandler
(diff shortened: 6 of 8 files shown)


patroni/__main__.py
```diff
@@ -164,10 +165,23 @@ def patroni_main(configfile: str) -> None:
...
+                       help='Generate a sample Patroni yaml configuration file')
+    group.add_argument('--generate-config', action='store_true',
+                       help='Generate a Patroni yaml configuration file for a running instance')
+    parser.add_argument('--dsn', help='Optional DSN string of the instance to be used as a source \
+                                    for config generation. Superuser connection is required.')
     args = parser.parse_args()
 
-    if args.validate_config:
+    if args.generate_sample_config:
+        generate_config(args.configfile or '/tmp/patroni.yml', True, None)
```
CyberDem0n, line 179: If configfile isn't specified I would suggest to simply print it to stdout.
WDYT?

patroni/config_generator.py
```diff
@@ -0,0 +1,286 @@
...
+            sys.exit('The provided user does not have superuser privilege')
+
+        required_params = ['hba_file', 'ident_file', 'config_file',
+                           'data_directory'] + list(ConfigHandler.CMDLINE_OPTIONS.keys())
+        cur.execute("SELECT name, current_setting(name) FROM pg_settings \
+                     WHERE context <> 'internal' \
+                     AND source IN ('configuration file', 'command line', 'environment variable') \
+                     AND category <> 'Write-Ahead Log / Recovery Target' \
+                     AND setting <> '(disabled)' \
+                     OR name = ANY(%s);", (required_params,))
```
CyberDem0n, line 121: Can we have adjacent strings instead of multiline? :)

CyberDem0n, line 121: We don't need semicolon at the end of SQL.

patroni/__main__.py
hughcapet: hm, yes, nice improvement. will add this

patroni/__main__.py
```diff
@@ -164,10 +165,23 @@ def patroni_main(configfile: str) -> None:
...
     parser = get_base_arg_parser()
-    parser.add_argument('--validate-config', action='store_true', help='Run config validator and exit')
+    group = parser.add_mutually_exclusive_group()
+    group.add_argument('--validate-config', action='store_true', help='Run config validator and exit')
+    group.add_argument('--generate-sample-config', action='store_true',
+                       help='Generate a sample Patroni yaml configuration file')
+    group.add_argument('--generate-config', action='store_true',
+                       help='Generate a Patroni yaml configuration file for a running instance')
+    parser.add_argument('--dsn', help='Optional DSN string of the instance to be used as a source \
+                                    for config generation. Superuser connection is required.')
```
barthisrael, line 175: I feel like we should start creating some sub-commands in this CLI instead of creating several different flags. IDK, some random idea:

In the main `patroni` command we could have a `-c/--config-file` flag that would apply to all sub-commands. Then we could have some sub-commands like:

* `patroni run` to run the agent. Requires `-c` to be set.
* `patroni config`: to group config related commands:
    * `patroni config validate`: to validate a configuration file. Requires `-c` to be set.
    * `patroni config generate`: to generate a configuration file. If `-c` is set, write to that file, otherwise write to stdout. It could have a couple flags:
        * `--sample`: if the user wants to create a sample file. If not given would require connecting to a running instance.
        * `--dsn`: optional DSN to connect to the running instance. Mutually exclusive with `--sample`.

What do you think?

patroni/config.py
```diff
@@ -32,6 +32,20 @@
...
+    'sslcert': 'PGSSLCERT',
+    'sslkey': 'PGSSLKEY',
+    'sslpassword': '',
+    'sslrootcert': 'PGSSLROOTCERT',
+    'sslcrl': 'PGSSLCRL',
+    'sslcrldir': 'PGSSLCRLDIR',
+    'gssencmode': 'PGGSSENCMODE',
+    'channel_binding': 'PGCHANNELBINDING'
+}
+
```
barthisrael, line 48: I wonder if this should be here or in `config_generator.py` as it is only used there. If we intend to keep it here as a "public" variable in the module, I'd say to add that to the docstring.

I guess this PR will be merged before https://github.com/zalando/patroni/pull/2708. In that case, if we decide to keep this variable as it is, I can later rebase that PR and document this variable there.

barthisrael:
```suggestion
AUTH_ALLOWED_PARAMETERS_MAPPING = {k: f"PG{k.upper().replace('_', '')}" if k != 'sslpassword' else ''
                                   for k in _AUTH_ALLOWED_PARAMETERS}
```
Something else that came to my mind is that we can use dict comprehension above.

barthisrael: Hmm, actually that dict comprehension would have a problem with `username`.

patroni/utils.py
```diff
@@ -977,3 +978,33 @@ def unquote(string: str) -> str:
...
+def get_major_version(bin_dir: Optional[str] = None, bin_name: str = 'postgres') -> str:
+    """Get the major version of PostgreSQL.
+
+    It is based on the output of ``postgres --version``.
+
+    :param bin_dir: path to PostgreSQL binaries directory. If ``None`` it will use the first ``postgres`` binary that
+        is found by subprocess in the ``PATH``.
+    :param bin_name: name of the postgres binary to call ('postgres' by default)
+    :returns: the PostgreSQL major version.
+    :raises :class:`patroni.exceptions.PatroniException`: if the postgres binary call failed due to OSError
```
barthisrael, line 992:
```suggestion
    :raises :class:`patroni.exceptions.PatroniException`: if the postgres binary call failed due to :exc:`OSError`.
```

patroni/config_generator.py
```diff
@@ -0,0 +1,289 @@
...
+from .config import Config
+from .exceptions import PatroniException
+from .postgresql.config import ConfigHandler
+from .postgresql.misc import postgres_major_version_to_int
+from .utils import get_major_version, patch_config
+
+
+def get_ip() -> str:
+    """
+    Try to get an ip address of the hostname returned by :func:`gethostname`.
```
barthisrael, line 18:
```suggestion
    """Try to get an ip address of the hostname returned by :func:`~socket.gethostname`.
```

Just to keep the pattern of the other docstrings in the module.
About `gethostname`, when generating API documentation IIRC it requires the fully qualified name -- although we can show only the function name in the actual generates docs by using `~`.

patroni/config_generator.py
```diff
@@ -0,0 +1,289 @@
...
+
+    .. note::
+            Can also return local ip
```
barthisrael, line 21: Maybe it's worth adding a note that this function can quit execution when facing `OSError` or `IndexError` exceptions.

patroni/config_generator.py
```diff
@@ -0,0 +1,289 @@
...
+        sys.exit(f'Failed to define ip address: {e}')
+    except IndexError:
+        sys.exit(f'Failed to define ip address. No address returned by getaddrinfo for {hostname}')
+
+
+def get_int_major_version(config: Optional[Dict[str, Any]] = None) -> int:
+    """Get major version of PostgreSQL from the binary as an integer.
+
+    :param config: optional dictionary representing Patroni config. If contains bin_dir and/or the custom
+    bin_name for postgres, the values are used for the version retrieval.
```
barthisrael, line 41: What about adding a couple examples of expected outputs?

patroni/config_generator.py
```diff
@@ -0,0 +1,289 @@
...
+    """
+    config = config or {}
+    postgres_bin = ((config.get('postgresql') or {}).get('bin_name') or {}).get('postgres', 'postgres')
+    return postgres_major_version_to_int(get_major_version(config['postgresql'].get('bin_dir') or None, postgres_bin))
+
+
+def get_bin_dir_from_running_instance(data_dir: str) -> str:
+    """Define the directory postgres binaries reside using postmaster's pid executable."""
```
barthisrael, line 49: Missing to describe parameter and return value.

barthisrael: Maybe it's worth it adding a note that this function can quit execution,

patroni/config_generator.py
```diff
@@ -0,0 +1,289 @@
...
+        sys.exit(f'Error while reading postmaster.pid file: {e}')
+    import psutil
+    try:
+        return os.path.dirname(psutil.Process(postmaster_pid).exe())
+    except psutil.NoSuchProcess:
+        sys.exit('Obtained postmaster pid doesn\'t exist')
+
+
+def enrich_config_from_running_instance(config: Dict[str, Any], no_value_msg: str, dsn: Optional[str] = None) -> None:
+    """Extend the passed ``config`` dictionary with the values gathered from a running instance.
```
barthisrael, line 67:
```suggestion
    """Extend the passed *config* dictionary with the values gathered from a running instance.
```

patroni/config_generator.py
```diff
@@ -0,0 +1,289 @@
...
+
+    Get
+    - non-internal GUC values having configuration file, postmaster command line or environment variable as a source
```
barthisrael, line 70:
```suggestion
    Get:

    - non-internal GUC values having configuration file, postmaster command line or environment variable as a source
```

IIRC sphinx docs would look broken if we don't add a whitespace between text and a list.

patroni/config_generator.py
```diff
@@ -0,0 +1,289 @@
...
+    - postgresql.connect_address, postgresql.listen,
+    - postgresql.pg_hba and postgresql.pg_ident if hba_file/ident_file is set to the default value
+    - superuser auth parameters (from the options used for connection)
+    And redefine scope with the clister_name GUC value if set
```
barthisrael, line 74:
```suggestion
    - superuser auth parameters (from the options used for connection)

    And redefine scope with the ``cluster_name`` GUC value if set.
```

We also need a space after the list.

patroni/config_generator.py
```diff
@@ -0,0 +1,289 @@
...
+        sys.exit('Obtained postmaster pid doesn\'t exist')
+
+
+def enrich_config_from_running_instance(config: Dict[str, Any], no_value_msg: str, dsn: Optional[str] = None) -> None:
+    """Extend the passed ``config`` dictionary with the values gathered from a running instance.
+
+    Get
+    - non-internal GUC values having configuration file, postmaster command line or environment variable as a source
+    - postgresql.connect_address, postgresql.listen,
+    - postgresql.pg_hba and postgresql.pg_ident if hba_file/ident_file is set to the default value
```
barthisrael, line 72:
```suggestion
    - ``postgresql.connect_address``, postgresql.listen``,
    - ``postgresql.pg_hba`` and ``postgresql.pg_ident`` if hba_file/ident_file is set to the default value
```

barthisrael (changes requested):
Posting a few thoughts that I had so far.

I intend to resume checking this PR tomorrow.

patroni/config_generator.py
hughcapet: hm, maybe execution quitting actually should not happen here...

patroni/config.py
hughcapet: > Hmm, actually that dict comprehension would have a problem with username.

Exactly 🥲

matthbakeredb:
```suggestion
AUTH_ALLOWED_PARAMETERS_MAPPING = {
    k: f"PG{k.upper().replace('_', '')}" if k != 'sslpassword' else ''
    for k in _AUTH_ALLOWED_PARAMETERS + ('user',)
    if k != 'username'
}
```
:laughing:

patroni/config_generator.py
```diff
@@ -0,0 +1,290 @@
...
+      defined by Patroni and adjusted by the source instances's configuration values.
+    - ``postgresql.parameters``: the source instance's archive_command, restore_command, archive_cleanup_command,
+      recovery_end_command, ssl_passphrase_command, hba_file, ident_file, config_file GUC values
+    - ``postgresql.bin_dir``: path to Postgres binaries gathered from the running instance or, if not available,
+      the value of PATRONI_POSTGRESQL_BIN_DIR ENV variable. Otherwise, an empty string.
+    - ``postgresql.datadir``: the value gathered from the corresponding PG GUC
+    - ``postgresql.listen``: source instance's listen_addresses and port GUC values
+    - ``postgresql.connect_address``: if possible, generated from the connection params
+    - ``postgresql.authentication``:
+        - superuser and replication users defined (if possible, usernames are set from the respective Patroni ENV vars,
```
matthbakeredb, line 200: Indented lists need to be separated by a blank line for Sphinx/rst docs

patroni/config_generator.py
```diff
@@ -0,0 +1,290 @@
...
+                    'password', no_value_msg)
+
+    # redundant values from the default config
+    del config['bootstrap']['dcs']['standby_cluster']
+
+    if file:
+        dir_path = os.path.dirname(file)
+        if dir_path and not os.path.isdir(dir_path):
+            os.makedirs(dir_path)
+        with open(file, 'w') as fd:
```
matthbakeredb, line 287: `open` should have encoding set.

matthbakeredb: I would consider having os modifying code separate from configuration generation and construction code.

tests/test_patroni.py
```diff
@@ -236,3 +241,257 @@ def test_ensure_unique_name(self):
...
+
+
+@patch('patroni.psycopg.connect', psycopg_connect)
+@patch('socket.getaddrinfo', Mock(return_value=[(0, 0, 0, 0, ('1.9.8.4', 1984))]))
+@patch('builtins.open', MagicMock())
+@patch('socket.gethostname', Mock(return_value='testhost'))
+@patch('subprocess.check_output', Mock(return_value=b"postgres (PostgreSQL) 13.2"))
+@patch('psutil.Process.exe', Mock(return_value='/bin/dir/from/running/postgres'))
+@patch('psutil.Process.__init__', Mock(return_value=None))
+class TestGenerateConfig(unittest.TestCase):
```
matthbakeredb, line 253: Should the tests for `config_generator.py` be in its own `test_config_generator.py` file?

tests/test_patroni.py
```diff
@@ -236,3 +241,257 @@ def test_ensure_unique_name(self):
...
+            },
+            'restapi': {
+                'connect_address': self.environ['PATRONI_RESTAPI_CONNECT_ADDRESS'],
+                'listen': self.environ['PATRONI_RESTAPI_LISTEN']
+            }
+        }
+
+    def _set_running_instance_config_vals(self):
+        # values are taken from tests/__init__.py
+        self.config['scope'] = 'my_cluster'  # cluster_name
```
matthbakeredb, line 310: Would be nice if there was a `deep_merge` func in utils that would allow

```
deep_merge(self.conf, {
  "scope": "my_cluster",
  ...
})
```

:wink:

matthbakeredb (changes requested):
Made a bunch of comments about code style more than anything else.

Very cool idea!

patroni/config_generator.py
```diff
@@ -0,0 +1,290 @@
...
+
+    - non-internal GUC values having configuration file, postmaster command line or environment variable as a source
+    - ``postgresql.connect_address``, postgresql.listen``,
+    - ``postgresql.pg_hba`` and ``postgresql.pg_ident`` if hba_file/ident_file is set to the default value
+    - superuser auth parameters (from the options used for connection)
+
+    And redefine scope with the ``cluster_name`` GUC value if set.
+
+    :param config: configuration parameters dict to be enriched
+    :param no_value_msg: str value to be used when a parameter value is not available
```
barthisrael, line 78:
```suggestion
    :param no_value_msg: :class:`str` value to be used when a parameter value is not available
```

Maybe like this if we are making a reference to the class.

patroni/config_generator.py
```diff
@@ -0,0 +1,290 @@
...
+    :param dsn: optional DSN string for the source running instance
+    """
+    from getpass import getuser, getpass
+    from patroni.postgresql.config import parse_dsn
+    from patroni.config import AUTH_ALLOWED_PARAMETERS_MAPPING
```
barthisrael, line 83: I wonder if we should move this to the top.

patroni/config_generator.py
```diff
@@ -0,0 +1,290 @@
...
+        val = parsed_dsn.get(conn_param, os.getenv(env_var))
+        if val:
+            su_params[conn_param] = val
+    patroni_env_su_username = ((config.get('authentication') or {}).get('superuser') or {}).get('username')
+    patroni_env_su_pwd = ((config.get('authentication') or {}).get('superuser') or {}).get('password')
+    # because we use "username" in the config for some reason
+    su_params['username'] = su_params.pop('user', patroni_env_su_username) or getuser()
+    su_params['password'] = su_params.get('password', patroni_env_su_pwd) or getpass('Please enter the user password:')
+
+    from . import psycopg
```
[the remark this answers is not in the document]
barthisrael: Same comment about moving imports to top.

patroni/config_generator.py
```diff
@@ -0,0 +1,290 @@
...
+                       'recovery_end_command', 'ssl_passphrase_command',
+                       'hba_file', 'ident_file', 'config_file'):
+                # write commands to the local config due to security implications
+                # write hba/ident/config_file to local config to ensure they are not removed later
+                config['postgresql'].setdefault('parameters', {})
+                config['postgresql']['parameters'][p] = v
+            elif p in helper_dict:
+                helper_dict[p] = v
+            else:
+                config['bootstrap']['dcs']['postgresql']['parameters'][p] = v
```
barthisrael, line 141: I was wondering if shouldn't everything other than what is listed in [PostgreSQL parameters controlled by Patroni](https://patroni.readthedocs.io/en/latest/patroni_configuration.html#postgresql-parameters-controlled-by-patroni) go to local configuration?

I mean, I see "dynamic configuration" as the configuration that applies to the whole cluster, and local configuration the configuration that applies only to the local node.

With that in mind, I'd say to write the settings that Patroni thinks we should enforce globally into dynamic configuration, and everything else go to local configuration by default. Then it's up to the user to check what he wants to be applied globally or not.

What do you think of that?

patroni/config_generator.py
```diff
@@ -0,0 +1,290 @@
...
+            if p == 'data_directory':
+                config['postgresql']['data_dir'] = v
+            elif p == 'cluster_name' and v:
+                config['scope'] = v
+            elif p in ('archive_command', 'restore_command', 'archive_cleanup_command',
+                       'recovery_end_command', 'ssl_passphrase_command',
+                       'hba_file', 'ident_file', 'config_file'):
+                # write commands to the local config due to security implications
+                # write hba/ident/config_file to local config to ensure they are not removed later
+                config['postgresql'].setdefault('parameters', {})
```
barthisrael, line 136: Couldn't we move this out of the loop?

patroni/config_generator.py
```diff
@@ -0,0 +1,290 @@
...
+                                                             (li.strip() for li in f.readlines())))
+        except OSError as e:
+            sys.exit(f'Failed to read pg_hba.conf: {e}')
+
+    default_ident_path = os.path.join(config['postgresql']['data_dir'], 'pg_ident.conf')
+    if config['postgresql']['parameters']['ident_file'] == default_ident_path:
+        try:
+            with open(default_ident_path, 'r') as f:
+                config['postgresql']['pg_ident'] = [i.strip() for i in f.readlines()
+                                                    if i.strip() and not i.startswith('#')]
```
barthisrael, line 167: This code can bring unexpected lines because of `i.startswith('#')` with no stripping, e.g.:

```python
>>> with open('pg_ident.conf', 'r') as f:
...     print(f.read())
...
test1
 test2
# test3
 # test4
test5
>>> with open('pg_ident.conf', 'r') as f:
...     print([i.strip() for i in f.readlines() if i.strip() and not i.startswith('#')])
...
['test1', 'test2', '# test4', 'test5']
```

patroni/config_generator.py
```diff
@@ -0,0 +1,290 @@
...
+            else:
+                config['bootstrap']['dcs']['postgresql']['parameters'][p] = v
+
+    conn.close()
+
+    connect_port = parsed_dsn.get('port', os.getenv('PGPORT', helper_dict['port']))
+    config['postgresql']['connect_address'] = f'{get_ip()}:{connect_port}'
+    config['postgresql']['listen'] = f'{helper_dict["listen_addresses"]}:{helper_dict["port"]}'
+
+    # it makes sense to define postgresql.pg_hba/pg_ident only if hba_file/ident_file are set to defaults
```
barthisrael, line 149: For sake of curiosity, do you know why Patroni ignores `postgresql.pg_hba` if `hba_file` in Postgres is not default?

patroni/config_generator.py
```diff
@@ -0,0 +1,290 @@
...
+    except OSError as e:
+        sys.exit(f'Error while reading postmaster.pid file: {e}')
+    import psutil
+    try:
+        return os.path.dirname(psutil.Process(postmaster_pid).exe())
+    except psutil.NoSuchProcess:
+        sys.exit('Obtained postmaster pid doesn\'t exist')
+
+
+def enrich_config_from_running_instance(config: Dict[str, Any], no_value_msg: str, dsn: Optional[str] = None) -> None:
```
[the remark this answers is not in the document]
barthisrael: Maybe `no_value_msg` could be a module constant, so we can simplify this signature. It is not modified anywhere, and is used in a couple functions.

patroni/config_generator.py
barthisrael, line 200:
```suggestion
    The created configuration contains:

    - ``scope``: ``cluster_name`` GUC value or ``PATRONI_SCOPE`` ENV variable value if available
    - ``name``: ``PATRONI_NAME`` ENV variable value if set, otherwise hostname
    - ``bootsrtap.dcs``: section with all the parameters (incl. the majority of PG GUCs) set to their default values
      defined by Patroni and adjusted by the source instances's configuration values.
    - ``postgresql.parameters``: the source instance's ``archive_command``, ``restore_command``, ``archive_cleanup_command``,
      ``recovery_end_command``, ``ssl_passphrase_command``, ``hba_file``, ``ident_file``, ``config_file`` GUC values
    - ``postgresql.bin_dir``: path to Postgres binaries gathered from the running instance or, if not available,
      the value of ``PATRONI_POSTGRESQL_BIN_DIR`` ENV variable. Otherwise, an empty string.
    - ``postgresql.datadir``: the value gathered from the corresponding PG GUC
    - ``postgresql.listen``: source instance's ``listen_addresses`` and ``port`` GUC values
    - ``postgresql.connect_address``: if possible, generated from the connection params
    - ``postgresql.authentication``:

        - superuser and replication users defined (if possible, usernames are set from the respective Patroni ENV vars,
```

patroni/config_generator.py
```diff
@@ -0,0 +1,290 @@
...
+        - rewind user is defined for a sample config if PG version can be defined and PG version is 11+
+          (if possible, username is set from the respective Patroni ENV var)
+    - ``bootsrtap.dcs.postgresql.use_pg_rewind``
+    - ``postgresql.pg_hba`` defaults or the lines gathered from the source instance's hba_file
+    - ``postgresql.pg_ident`` the lines gathered from the source instance's ident_file
+
+    :param file: Full path to the configuration file to be used. If not provided, result is sent to stdout.
+    :param sample: Optional flag. If set, no source instance will be used - generate config with some sane defaults.
+    :param dsn: Optional DSN string for the local instance to get GUC values from.
+    """
```
barthisrael, line 212:
```suggestion
    The created configuration contains:

    - ``scope``: ``cluster_name`` GUC value or ``PATRONI_SCOPE`` ENV variable value if available
    - ``name``: ``PATRONI_NAME`` ENV variable value if set, otherwise hostname
    - ``bootsrtap.dcs``: section with all the parameters (incl. the majority of PG GUCs) set to their default values
      defined by Patroni and adjusted by the source instances's configuration values.
    - ``postgresql.parameters``: the source instance's ``archive_command``, ``restore_command``, ``archive_cleanup_command``,
      ``recovery_end_command``, ``ssl_passphrase_command``, ``hba_file``, ``ident_file``, ``config_file`` GUC values
    - ``postgresql.bin_dir``: path to Postgres binaries gathered from the running instance or, if not available,
      the value of ``PATRONI_POSTGRESQL_BIN_DIR`` ENV variable. Otherwise, an empty string.
    - ``postgresql.datadir``: the value gathered from the corresponding PG GUC
    - ``postgresql.listen``: source instance's ``listen_addresses`` and ``port`` GUC values
    - ``postgresql.connect_address``: if possible, generated from the connection params
    - ``postgresql.authentication``:

        - superuser and replication users defined (if possible, usernames are set from the respective Patroni ENV vars,
          otherwise the default ``postgres`` and ``replicator`` values are used).
          If not a sample config, either DSN or PG ENV vars are used to define superuser authentication parameters.
        - rewind user is defined for a sample config if PG version can be defined and PG version is 11+
          (if possible, username is set from the respective Patroni ENV var)
    - ``bootsrtap.dcs.postgresql.use_pg_rewind``
    - ``postgresql.pg_hba`` defaults or the lines gathered from the source instance's ``hba_file``
    - ``postgresql.pg_ident`` the lines gathered from the source instance's ``ident_file``

    :param file: Full path to the configuration file to be used. If not provided, result is sent to stdout.
    :param sample: Optional flag. If set, no source instance will be used - generate config with some sane defaults.
    :param dsn: Optional DSN string for the local instance to get GUC values from.
    """
```

patroni/config_generator.py
```diff
@@ -0,0 +1,290 @@
...
+    no_value_msg = '#FIXME'
```
barthisrael, line 213: I mentioned it before, but I guess this could be a module constant.

patroni/config_generator.py
```diff
@@ -0,0 +1,290 @@
...
+    pg_version = None
+    local_ip = get_ip()
```
barthisrael, line 215: Maybe the helper function could return both hostname and IP? Just suggesting because we perform calls to `socket.hostname` in 2 different places for but for a similar purpose.

tests/test_patroni.py
barthisrael: Isn't that kind of what `patch_config` does?

patroni/config_generator.py
hughcapet: oh, seems we should fix it in a lot of places then
e.g. https://github.com/zalando/patroni/blob/master/patroni/api.py#L207-L233

patroni/config_generator.py
```diff
@@ -0,0 +1,290 @@
...
+        if cur.rowcount < 1:
+            sys.exit('The provided user does not have superuser privilege')
+
+        required_params = ['hba_file', 'ident_file', 'config_file',
+                           'data_directory'] + list(ConfigHandler.CMDLINE_OPTIONS.keys())
+        cur.execute("SELECT name, current_setting(name) FROM pg_settings "
+                    "WHERE context <> 'internal' "
+                    "AND source IN ('configuration file', 'command line', 'environment variable') "
+                    "AND category <> 'Write-Ahead Log / Recovery Target' "
+                    "AND setting <> '(disabled)' "
```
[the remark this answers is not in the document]
hughcapet: yes

patroni/config_generator.py
hughcapet: >  It is strange to me that all imports are done in methods in Patroni.

In Patroni it is usually done not to import when it is actually not required (some branches are not that often accessed)
As I understand it (at least that is what guides me to do it like that) @CyberDem0n?

patroni/config_generator.py
hughcapet: > open should have encoding set.

had a deep dive into this, thanks:)
https://docs.python.org/3/library/os.html#utf8-mode
still not sure we need to explicitly specify it now, but definitely won't hurt. 

the more serious problem, I guess, is with `safe_dump`, we definitely need to use `allow_unicode=True` there

patroni/config_generator.py
matthbakeredb: I wonder how efficient that is. If you need to run import machinery to load a module at time of execution vs once when the application starts it might actually add a time hit. An imported module is a single instance in memory so once it's imported the same module is accessed on each import, rather than read from disk every import statement. That could potentially still cause a performance hit in some cases, although negligible. Obviously, there's different visible effects between the user cli, api and the daemon. Just out of interest, has any benchmarking/profiling been done for this?

CyberDem0n: The problem exists only with `patroni/__main__.py`, which could fork Patroni when running with the PID=1 (in docker).

That is, https://github.com/zalando/patroni/pull/2786/files#diff-48810aca58351596dee0cb6815e887053509d54b64929c9f1fd8a2c1c538a752R10 should be imported from process_arguments(). Inside the config_generator we can do anything you like.

hughcapet: 🤦‍♀️ how short memory I have

patroni/config_generator.py
hughcapet: because it may happen it is located outside of PGDATA and we don't have write permissions for it

patroni/collections.py
```diff
@@ -187,6 +187,13 @@ def copy(self) -> 'CaseInsensitiveDict':
         """
         return CaseInsensitiveDict({v[0]: v[1] for v in self._values.values()})
 
+    def keys(self) -> KeysView[str]:
+        """Return a new view of the dict's keys
+
+        :return: a set-like object providing a view on the dict's keys
```
barthisrael, line 193:
```suggestion
        :returns: a set-like object providing a view on the dict's keys
```

patroni/config_generator.py
```diff
@@ -0,0 +1,353 @@
...
+    * ``bootsrtap.dcs.postgresql.use_pg_rewind``
+    * ``postgresql.pg_hba`` defaults or the lines gathered from the source instance's hba_file
+    * ``postgresql.pg_ident`` the lines gathered from the source instance's ident_file
+
+    :param file: Full path to the configuration file to be used. If not provided, result is sent to stdout.
+    :param sample: Optional flag. If set, no source instance will be used - generate config with some sane defaults.
+    :param dsn: Optional DSN string for the local instance to get GUC values from.
+    """
+    pg_version = None
+    hotname, local_ip = get_address()
```
barthisrael, line 279:
```suggestion
    hostname, local_ip = get_address()
```

patroni/config_generator.py
```diff
@@ -0,0 +1,353 @@
...
+
+    config = Config('', None).local_configuration  # Get values from env
+    dynamic_config = Config.get_default_config()
+    dynamic_config['postgresql']['parameters'] = dict(dynamic_config['postgresql']['parameters'])
+    config.setdefault('bootstrap', {})['dcs'] = dynamic_config
+    config.setdefault('postgresql', {})
+
+    template_config: Dict[str, Any] = {
+        'scope': _NO_VALUE_MSG,
+        'name': hotname,
```
barthisrael, line 289:
```suggestion
        'name': hostname,
```

barthisrael (commented):
I can see you turned this PR back to a draft, but I'm putting 3 small comments based on what I saw earlier.

patroni/__main__.py
hughcapet: even if we find a way to implement it backward-compatible, let's not do it in this PR:)

patroni/config_generator.py
hughcapet: What is your use case of having different configuration on different nodes?

patroni/__main__.py
barthisrael: These were just some side thoughts because I felt we are starting to pile up commands through flags, and personally I think we should adjust things sooner than later, otherwise the impact might be higher later 😆 

There is the counter part of existing deployments, yes. Maybe increasing a more significant digit in the version when releasing could pull some attention that the upgrade may break things. I mean, that's what in see in Postgres. We have the minor releases that in theory should not break things (although I already saw a minor breaking the application once 😆 ), and the major releases that have more structural changes that introduce new things and breaking changes.

I guess this goes more or less in the same direction of some thoughts that came up in Essen during the meeting.

patroni/config_generator.py
barthisrael: Maybe we should just keep it as it is. It was just some random thought based on the conception that I got about how these settings apply to nodes.

I already saw people with different settings, like `work_mem`, because queries that were run in the standbys had different aspects than the ones being run in the primary. But in fact, as they are part of a replication scenario, and standbys could take the primary role at any moment, I would prefer the standbys to be similar in terms of hardware and configuration.

In practice it would not change the result -- unless the existing cluster has a divergence in configuration among the nodes. Also, I guess it could make it harder to manage the cluster after that if one attempts to change configuration through `patronictl edti-config`, as the local configuration would take precedence.

Just forget about it 😄

patroni/config_generator.py
barthisrael: Oh, I see. Thanks!

matthbakeredb (changes requested):
Looks great!

Made some docstring comments mostly. I might have made the line too long in some cases when using the fully-qualified name needed for sphinx to create links, e.g. :attr:`~RunningClusterConfigGenerator.config`. So might need to check the final result. Sorry about that.

patroni/config_generator.py
```diff
@@ -0,0 +1,459 @@
...
+        allowed_types = ('local', 'host', 'hostssl', 'hostnossl', 'hostgssenc', 'hostnogssenc')
+        if self.pg_major and self.pg_major >= 160000:
+            allowed_types += ('include', 'include_if_exists', 'include_dir')
+        return allowed_types
+
+    @property
+    def _required_pg_params(self) -> List[str]:
+        """PG configuration prameters that have to be always present in the generated config.
+
+        :returns:
```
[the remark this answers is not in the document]
hughcapet: When you leave something halfway saying „will just fix that other line and return here“ - you never do 🤦‍♀️

matthbakeredb:
> @matthbakeredb thank you for your patient reviews. My attentiveness is the worst see_no_evil

You should see Israel's internal reviews of my code :laughing:

patroni/config_generator.py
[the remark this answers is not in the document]
matthbakeredb: I do it all the time!

patroni/config_generator.py
```diff
@@ -0,0 +1,466 @@
+"""patroni --generate-config machinery."""
```
barthisrael, line 1:
```suggestion
"""patroni ``--generate-config`` machinery."""
```

patroni/config_generator.py
```diff
@@ -0,0 +1,466 @@
...
+_NO_VALUE_MSG = '#FIXME'
+
+
+def get_address() -> Tuple[str, str]:
+    """Try to get hostname and the ip address for it returned by :func:`~socket.gethostname`.
+
+    .. note::
+        Can also return local ip.
+
+    :returns: tuple consisting of the hostname returned by `~socket.gethostname`
```
barthisrael, line 51:
```suggestion
    :returns: tuple consisting of the hostname returned by :func:`~socket.gethostname`
```

patroni/config_generator.py
```diff
@@ -0,0 +1,466 @@
...
+        and the first element in the sorted list of the addresses returned by :func:`~socket.getaddrinfo`.
+        Sorting guarantees it will prefer IPv4.
+
+    :raises:
+        :class:`PatroniException`: if :exc:`OSError` occured
```
barthisrael, line 56:
```suggestion
        :exc:`~patroni.exceptions.PatroniException`: if :exc:`OSError` occured
```

patroni/config_generator.py
```diff
@@ -0,0 +1,466 @@
...
+    try:
+        hostname = socket.gethostname()
+        return hostname, sorted(socket.getaddrinfo(hostname, 0, socket.AF_UNSPEC, socket.SOCK_STREAM, 0),
+                                key=lambda x: x[0])[0][4][0]
+    except OSError as err:
+        raise PatroniException(f'Failed to define ip address: {err}')
+
+
+class AbstractConfigGenerator(abc.ABC):
+    """Object representing the generated Patroni config."""
```
barthisrael, line 68:
```suggestion
    """Object representing the generated Patroni config.
    
    :ivar output_file: ...
    :ivar pg_major: ...
    :ivar config: ...
    """
```

Add documentation of public instance attributes.

patroni/config_generator.py
```diff
@@ -0,0 +1,466 @@
...
+            See :func:`~patroni.postgresql.misc.postgres_major_version_to_int` and
+            :func:`~patroni.utils.get_major_version`.
+        """
+        postgres_bin = ((self.config.get('postgresql') or {}).get('bin_name') or {}).get('postgres', 'postgres')
+        return postgres_major_version_to_int(get_major_version(self.config['postgresql'].get('bin_dir') or None,
+                                                               postgres_bin))
+
+    @abc.abstractmethod
+    def generate(self) -> None:
+        """Generate config and store in :attr:`~AbstractConfigGenerator.self.config`."""
```
barthisrael, line 124:
```suggestion
        """Generate config and store in :attr:`~AbstractConfigGenerator.config`."""
```

patroni/config_generator.py
```diff
@@ -0,0 +1,466 @@
...
+        :param output_file: full path to the output file to be used
+        """
+        super().__init__(output_file)
+
+        self.pg_major = self._get_int_major_version()
+        self.generate()
+
+    @property
+    def get_auth_method(self) -> str:
+        """Return the preferred authentication method for a specific PG version if provided or the default 'md5'.
```
barthisrael, line 162:
```suggestion
        """Return the preferred authentication method for a specific PG version if provided or the default ``md5``.
```

patroni/config_generator.py
```diff
@@ -0,0 +1,466 @@
...
+    """Object representing the Patroni config generated using information gathered from a running instance."""
+
+    def __init__(self, output_file: Optional[str] = None, dsn: Optional[str] = None) -> None:
+        """Additionally store the passed dsn (if any) in both original and parsed version and run config generation.
+
+        :param output_file: full path to the output file to be used
+        :param dsn: DSN string for the local instance to get GUC values from
+
+        :raises:
+            :class:`PatroniException`: if DSN parsing failed
```
barthisrael, line 202:
```suggestion
            :exc:`~patroni.exceptions.PatroniException`: if DSN parsing failed
```

patroni/config_generator.py
```diff
@@ -0,0 +1,466 @@
...
+            self.config['postgresql']['authentication'].setdefault(
+                'rewind', {'username': 'rewind_user'}).setdefault('password', _NO_VALUE_MSG)
+
+        del self.config['bootstrap']['dcs']['standby_cluster']
+
+        self.merge_with_template()
+
+
+class RunningClusterConfigGenerator(AbstractConfigGenerator):
+    """Object representing the Patroni config generated using information gathered from a running instance."""
```
barthisrael, line 193:
```suggestion
    """Object representing the Patroni config generated using information gathered from a running instance.
    
    :ivar dsn: ...
    :ivar parsed_dsn: ...
    """
```

patroni/config_generator.py
```diff
@@ -0,0 +1,466 @@
...
+            if not self.parsed_dsn:
+                raise PatroniException('Failed to parse DSN string')
+
+        self.generate()
+
+    @property
+    def _get_hba_conn_types(self) -> Tuple[str, ...]:
+        """Return the connection types allowed.
+
+        If pg_major is defined, adds additional params for PostgreSQL version >=16.
```
barthisrael, line 219:
```suggestion
        If :attr:`~RunningClusterConfigGenerator.pg_major` is defined, adds additional params for PostgreSQL version >=16.
```

patroni/config_generator.py
```diff
@@ -0,0 +1,466 @@
...
+
+    def _get_bin_dir_from_running_instance(self) -> str:
+        """Define the directory postgres binaries reside using postmaster's pid executable.
+
+        :param data_dir: the PostgreSQL data directory to search for postmaster.pid file in.
+
+        :returns: path to the PostgreSQL binaries directory
+
+        :raises:
+            :class:`PatroniException`: if:
```
barthisrael, line 245:
```suggestion
            :exc:`~patroni.exceptions.PatroniException`: if:
```

patroni/config_generator.py
```diff
@@ -0,0 +1,466 @@
...
+
+                * pid could not be obtained from the `postmaster.pid` file; or
+                * :exc:`OSError` occured during `postmaster.pid` file handling; or
```
barthisrael, line 248:
```suggestion
                * pid could not be obtained from the ``postmaster.pid`` file; or
                * :exc:`OSError` occured during ``postmaster.pid`` file handling; or
```

patroni/config_generator.py
```diff
@@ -0,0 +1,466 @@
...
+            return os.path.dirname(psutil.Process(postmaster_pid).exe())
+        except psutil.NoSuchProcess:
+            raise PatroniException("Obtained postmaster pid doesn't exist.")
+
+    @contextmanager
+    def _get_connection_cursor(self) -> Iterator[Union['cursor', 'Cursor[Any]']]:
+        """Get cursor for the PG connection established based on the stored information.
+
+        :raises:
+            :class:`PatroniException`: if :exc:`psycopg.Error` occured.
```
barthisrael, line 271:
```suggestion
            :exc:`~patroni.exceptions.PatroniException`: if :exc:`psycopg.Error` occured.
```

patroni/config_generator.py
```diff
@@ -0,0 +1,466 @@
...
+        return cur.rowcount == 1
+
+    def _set_pg_params(self, cur: Union['cursor', 'Cursor[Any]']) -> None:
+        """Extend :attr:`~RunningClusterConfigGenerator.config` with the actual PG GUCs values.
+
+        THe following GUC values are set:
+
+            * Non-internal having configuration file, postmaster command line or environment variable
+              as a source.
+            * List of the always required parameters (see :func:`~RunningClusterConfigGenerator._required_pg_params`)
```
barthisrael, line 298:
```suggestion
            * Non-internal having configuration file, postmaster command line or environment variable
              as a source.

            * List of the always required parameters (see :func:`~RunningClusterConfigGenerator._required_pg_params`)
```

Do it doesn't merge when building with sphinx.

patroni/config_generator.py
```diff
@@ -0,0 +1,466 @@
...
+    def _set_conf_files(self) -> None:
+        """Extend :attr:`~RunningClusterConfigGenerator.config` with the information from ``pg_hba.conf`` and
+        ``pg_ident.conf`` files.
+
+        .. note::
+            This function only defines ``postgresql.pg_hba`` and ``postgresql.pg_ident`` when
+            ``hba_file`` and ``ident_file`` are set to the defaults.
+
+        :raises:
+            :class:`PatroniException`: if :exc:`OSError` occured during the conf files handling.
```
barthisrael, line 361:
```suggestion
            :exc:`~patroni.exceptions.PatroniException`: if :exc:`OSError` occured during the conf files handling.
```

patroni/config_generator.py
```diff
@@ -0,0 +1,466 @@
...
+            try:
+                self.config['postgresql']['pg_ident'] = [i for i in read_stripped(default_ident_path)
+                                                         if i and not i.startswith('#')]
+            except OSError as err:
+                raise PatroniException(f'Failed to read pg_ident.conf: {err}')
+            if not self.config['postgresql']['pg_ident']:
+                del self.config['postgresql']['pg_ident']
+
+    def _enrich_config_from_running_instance(self) -> None:
+        """Extend `self.config` dictionary with the values gathered from a running instance.
```
barthisrael, line 382:
```suggestion
        """Extend :attr:`~RunningClusterConfigGenerator.config` dictionary with the values gathered from a running instance.
```

patroni/config_generator.py
```diff
@@ -0,0 +1,466 @@
...
+
+        * superuser auth parameters (see :func:`~RunningClusterConfigGenerator._set_su_params`)
+        * some GUC values (see :func:`~RunningClusterConfigGenerator._set_pg_params`)
+        * ``postgresql.connect_address``, ``postgresql.listen``
+        * ``postgresql.pg_hba`` and ``postgresql.pg_ident`` (see :func:`~RunningClusterConfigGenerator._set_conf_files`)
+
+        And redefine ``scope`` with the ``cluster_name`` GUC value if set.
+
+        :raises:
+            :class:`PatroniException`: if the provided user doesn't have superuser privileges.
```
barthisrael, line 394:
```suggestion
            :exc:`~patroni.exceptions.PatroniException`: if the provided user doesn't have superuser privileges.
```

patroni/config_generator.py
```diff
@@ -0,0 +1,466 @@
...
+            if not self._is_superuser(cur, self.config['postgresql']['authentication']['superuser']['username']):
+                raise PatroniException('The provided user does not have superuser privilege')
+
+            self._set_pg_params(cur)
+
+        self._set_conf_files()
+
+    def generate(self) -> None:
+        """Generate config using the info gathered from the specified running PG instance and update
+        `~RunningClusterConfigGenerator.config`."""
```
barthisrael, line 410:
```suggestion
        :attr:`~RunningClusterConfigGenerator.config`."""
```

patroni/config_generator.py
```diff
@@ -0,0 +1,466 @@
...
+
+        self.merge_with_template()
+
+
+def generate_config(output_file: str, sample: bool, dsn: Optional[str]) -> None:
+    """Generate Patroni configuration file.
+
+    Gather all the available non-internal GUC values having configuration file, postmaster command line or environment
+    variable as a source and store them in the appropriate part of Patroni configuration (``postgresql.parameters`` or
+    ``bootsrtap.dcs.postgresql.parameters``). Either the provided DSN (takes precedence) or PG ENV vars will be used
```
barthisrael, line 423:
```suggestion
    ``bootstrap.dcs.postgresql.parameters``). Either the provided DSN (takes precedence) or PG ENV vars will be used
```

patroni/config_generator.py
```diff
@@ -0,0 +1,466 @@
...
+
+        * superuser and replication users defined (if possible, usernames are set from the respective Patroni ENV vars,
+          otherwise the default 'postgres' and 'replicator' values are used).
+          If not a sample config, either DSN or PG ENV vars are used to define superuser authentication parameters.
+        * rewind user is defined only for sample config, if PG version can be defined and PG version is >=11
+          (if possible, username is set from the respective Patroni ENV var)
+
+    * ``bootsrtap.dcs.postgresql.use_pg_rewind`` set to ``True`` for a sample config only.
+    * ``postgresql.pg_hba`` defaults or the lines gathered from the source instance's ``hba_file``.
+    * ``postgresql.pg_ident`` the lines gathered from the source instance's ``ident_file``.
```
barthisrael, line 449:
```suggestion
* ``scope``: ``cluster_name`` GUC value or ``PATRONI_SCOPE ENV`` variable value if available
    * ``name``: ``PATRONI_NAME`` ENV variable value if set, otherwise hostname

    * ``bootsrtap.dcs``: section with all the parameters (incl. the majority of PG GUCs) set to their default values
      defined by Patroni and adjusted by the source instances's configuration values.

    * ``postgresql.parameters``: the source instance's ``archive_command``, ``restore_command``,
      ``archive_cleanup_command``, ``recovery_end_command``, ``ssl_passphrase_command``, ``hba_file``, ``ident_file``,
      ``config_file`` GUC values.

    * ``postgresql.bin_dir``: path to Postgres binaries gathered from the running instance or, if not available,
      the value of ``PATRONI_POSTGRESQL_BIN_DIR`` ENV variable. Otherwise, an empty string.

    * ``postgresql.datadir``: the value gathered from the corresponding PG GUC.
    * ``postgresql.listen``: source instance's ``listen_addresses`` and port GUC values.
    * ``postgresql.connect_address``: if possible, generated from the connection params
    * ``postgresql.authentication``:

        * superuser and replication users defined (if possible, usernames are set from the respective Patroni ENV vars,
          otherwise the default ``postgres`` and ``replicator`` values are used).
          If not a sample config, either DSN or PG ENV vars are used to define superuser authentication parameters.

        * rewind user is defined only for sample config, if PG version can be defined and PG version is >=11
          (if possible, username is set from the respective Patroni ENV var)

    * ``bootsrtap.dcs.postgresql.use_pg_rewind`` set to ``True`` for a sample config only.
    * ``postgresql.pg_hba`` defaults or the lines gathered from the source instance's ``hba_file``.
    * ``postgresql.pg_ident`` the lines gathered from the source instance's ``ident_file``.
```

patroni/config_generator.py
```diff
@@ -0,0 +1,466 @@
...
+
+    .. note::
+        In case :class:`PatroniException` is raised by any of the called methods, execution is terminated.
```
barthisrael, line 452:
```suggestion
        In case :exc:`~patroni.exceptions.PatroniException` is raised by any of the called methods, execution is terminated.
```

barthisrael (changes requested):
Very nice work in the refactoring :)

Some sphinx related comments, basically:

- Fix broken links
- Use appropriate `exc` for exceptions
- Adjust lists that would render weirdly
- Missing instance attributes docs
- Typos

patroni/config_generator.py
```diff
@@ -0,0 +1,466 @@
...
+from typing import Any, Dict, Iterator, List, Optional, Tuple, TYPE_CHECKING, Union
+if TYPE_CHECKING:  # pragma: no cover
+    from psycopg import Cursor
+    from psycopg2 import cursor
+
+from . import psycopg
+from .config import Config
+from .exceptions import PatroniException
+from .postgresql.config import parse_dsn
+from .postgresql.config import ConfigHandler
```
CyberDem0n, line 21:
```suggestion
from .postgresql.config import ConfigHandler, parse_dsn
```

patroni/config_generator.py
```diff
@@ -0,0 +1,466 @@
...
+        return hostname, sorted(socket.getaddrinfo(hostname, 0, socket.AF_UNSPEC, socket.SOCK_STREAM, 0),
+                                key=lambda x: x[0])[0][4][0]
+    except OSError as err:
+        raise PatroniException(f'Failed to define ip address: {err}')
+
+
+class AbstractConfigGenerator(abc.ABC):
+    """Object representing the generated Patroni config."""
+
+    _HOSTNAME, _IP = get_address()
```
CyberDem0n, line 70: If something is wrong get_address() will raise an exception that will not be handled in the generate_config() function: https://github.com/zalando/patroni/pull/2786/files#diff-cfee4a0506f1e512ce4c1fb59d7f7bac2f7924ad20e7d6a679044f4c5f4d6c88R465-R466

patroni/utils.py
```diff
@@ -977,3 +990,33 @@ def unquote(string: str) -> str:
...
         ret = string
     return ret
+
+
+def get_major_version(bin_dir: Optional[str] = None, bin_name: str = 'postgres') -> str:
+    """Get the major version of PostgreSQL.
+
+    It is based on the output of ``postgres --version``.
+
+    :param bin_dir: path to PostgreSQL binaries directory. If ``None`` it will use the first ``postgres`` binary that
```
CyberDem0n, line 1000:
```suggestion
    :param bin_dir: path to PostgreSQL binaries directory. If ``None`` or empty string it will use the first *bin_name* binary that
```

patroni/config_generator.py
```diff
@@ -0,0 +1,466 @@
...
+    def _get_int_major_version(self) -> int:
+        """Get major PostgreSQL version from the binary as an integer.
+
+        :returns: an integer PostgreSQL major version representation gathered from the PostgreSQL binary.
+            See :func:`~patroni.postgresql.misc.postgres_major_version_to_int` and
+            :func:`~patroni.utils.get_major_version`.
+        """
+        postgres_bin = ((self.config.get('postgresql') or {}).get('bin_name') or {}).get('postgres', 'postgres')
+        return postgres_major_version_to_int(get_major_version(self.config['postgresql'].get('bin_dir') or None,
+                                                               postgres_bin))
```
CyberDem0n, line 120:
```suggestion
        return postgres_major_version_to_int(get_major_version(self.config['postgresql'].get('bin_dir'), postgres_bin))
```

patroni/config_generator.py
```diff
@@ -0,0 +1,466 @@
...
+
+        :raises:
+            :class:`PatroniException`: if the provided user doesn't have superuser privileges.
+        """
+        self._set_su_params()
+
+        with self._get_connection_cursor() as cur:
+            self.pg_major = getattr(cur.connection, 'server_version', 0)
+
+            if not self._is_superuser(cur, self.config['postgresql']['authentication']['superuser']['username']):
```
CyberDem0n, line 401:
```suggestion
            if not parse_bool(cur.connection.info.parameter_status('is_superuser')):
```

And we can get rid of `_is_superuser()` method.

patroni/config_generator.py
```diff
@@ -0,0 +1,466 @@
...
+        self._set_conf_files()
+
+    def generate(self) -> None:
+        """Generate config using the info gathered from the specified running PG instance and update
+        `~RunningClusterConfigGenerator.config`."""
+        self._enrich_config_from_running_instance()
+        self.config['postgresql']['bin_dir'] = self._get_bin_dir_from_running_instance()
+        del self.config['bootstrap']['dcs']['standby_cluster']
+
+        self.merge_with_template()
```
CyberDem0n, line 415: This two lines are exactly the same as here https://github.com/zalando/patroni/pull/2786/files#diff-cfee4a0506f1e512ce4c1fb59d7f7bac2f7924ad20e7d6a679044f4c5f4d6c88R187-R189

Also, self.generate() is called from both constructors independently. We probably can do something about it.

patroni/config_generator.py
```diff
@@ -0,0 +1,459 @@
...
+        """
+        self.output_file = file
+
+        self.pg_major = 0
+
+        self.config: Dict[str, Any] = Config('', None).local_configuration  # Get values from env
+        dynamic_config = Config.get_default_config()
+        dynamic_config['postgresql']['parameters'] = dict(dynamic_config['postgresql']['parameters'])
+        self.config.setdefault('bootstrap', {})['dcs'] = dynamic_config
+        self.config.setdefault('postgresql', {})
```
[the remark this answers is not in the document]
hughcapet: sorry, I don't get: why is it better?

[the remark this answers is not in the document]
matthbakeredb: Prepare a dict and perform modifications to it and set the class attr self.config once with the final result. It's a more atomic approach. I think of the treatment of the data handled by the `config` dict as internal to the method, whilst the writing to the class attr as external. Although, there's not much than can go wrong in between the first write here, it makes me uneasy as "what if" something happened and your class attr only had partial information written. :-)

patroni/config_generator.py
CyberDem0n: Example:

```python
$ cat test.py 
class TestException(Exception):
    pass


def test():
    raise TestException


class Test:
    foo = test()


if __name__ == '__main__':
    print('main')

$ python test.py 
  File "test.py", line 9, in <module>
    class Test:
  File "test.py", line 11, in Test
    foo = test()
  File "test.py", line 6, in test
    raise TestException
__main__.TestException
```

patroni/config_generator.py
```diff
@@ -0,0 +1,465 @@
...
+    Sane defults are used based on the gathered PG version.
+    """
+
+    def __init__(self, output_file: Optional[str] = None) -> None:
+        """Additionally set the PG major version from the binary and run config generation.
+
+        :param output_file: full path to the output file to be used.
+        """
+        super().__init__(output_file)
+
```
CyberDem0n, line 166:
```suggestion
```

patroni/config_generator.py
```diff
@@ -0,0 +1,465 @@
...
+    def __init__(self, output_file: Optional[str], dsn: Optional[str] = None) -> None:
+        """Set up the output file (if passed), helper vars and the minimal config structure.
+
+        :param output_file: full path to the output file to be used.
+        :param dsn: DSN string for the local instance to get GUC values from.
+        """
+        self.output_file = output_file
+        self.pg_major = 0
+        self.dsn = dsn
+        self.parsed_dsn = {}
```
CyberDem0n, line 109: SampleConfigGenerator doesn't use these attributes.

They should be moved to `RunningClusterConfigGenerator.__init__()`, _before_ calling `super().__init__()`.

patroni/config_generator.py
```diff
@@ -0,0 +1,462 @@
...
+    :ivar config: dictionary used for the generated configuration storage.
+    """
+
+    _HOSTNAME, _IP = get_address()
+
+    def __init__(self, output_file: Optional[str], dsn: Optional[str] = None) -> None:
+        """Set up the output file (if passed), helper vars and the minimal config structure.
+
+        :param output_file: full path to the output file to be used.
+        :param dsn: DSN string for the local instance to get GUC values from.
```
CyberDem0n, line 79:
```suggestion
    def __init__(self, output_file: Optional[str]) -> None:
        """Set up the output file (if passed), helper vars and the minimal config structure.

        :param output_file: full path to the output file to be used.
```

patroni/config_generator.py
```diff
@@ -0,0 +1,462 @@
...
+
+        :raises:
+            :exc:`~patroni.exceptions.PatroniException`: if DSN parsing failed.
+        """
+        self.dsn = dsn
+        self.parsed_dsn = {}
+
+        super().__init__(output_file, dsn)
```
CyberDem0n, line 210:
```suggestion
        super().__init__(output_file)
```

patroni/utils.py
```diff
@@ -467,6 +468,18 @@ def _sleep(interval: Union[int, float]) -> None:
...
 
+def read_stripped(file_path: str) -> Iterator[str]:
+    """Iterate over stripped lines in the given file.
+
+    :param file_path: path to the file to read from
+
+    :yields: each line from the given filed stripped
+    """
+    with open(file_path) as f:
+        for line in f.readlines():
```
CyberDem0n, line 479:
```suggestion
        for line in f:
```

patroni/utils.py
```diff
@@ -977,3 +990,33 @@ def unquote(string: str) -> str:
...
     return ret
+
+
+def get_major_version(bin_dir: Optional[str] = None, bin_name: str = 'postgres') -> str:
+    """Get the major version of PostgreSQL.
+
+    It is based on the output of ``postgres --version``.
+
+    :param bin_dir: path to the PostgreSQL binaries directory. If ``None`` or an empty string, it will use the first
+    ``postgres`` binary that is found by the subprocess in the ``PATH``.
```
barthisrael, line 1001:
```suggestion
    *bin_name* binary that is found by the subprocess in the ``PATH``.
```

patroni/utils.py
```diff
@@ -977,3 +990,33 @@ def unquote(string: str) -> str:
...
+    :param bin_name: name of the postgres binary to call ('postgres' by default)
+    :returns: the PostgreSQL major version.
+    :raises :exc:`~patroni.exceptions.PatroniException`: if the postgres binary call failed due to :exc:`OSError`.
```
barthisrael, line 1004:
```suggestion
    :raises:
        :exc:`~patroni.exceptions.PatroniException`: if the postgres binary call failed due to :exc:`OSError`.
```

hughcapet:
@matthbakeredb @barthisrael any other suggestions/objections?

patroni/config_generator.py
```diff
@@ -0,0 +1,461 @@
...
+def get_address() -> Tuple[str, str]:
+    """Try to get hostname and the ip address for it returned by :func:`~socket.gethostname`.
+
+    .. note::
+        Can also return local ip.
+
+    :returns: tuple consisting of the hostname returned by :func:`~socket.gethostname`
+        and the first element in the sorted list of the addresses returned by :func:`~socket.getaddrinfo`.
+        Sorting guarantees it will prefer IPv4.
+        If an exception occured, hostname and ip values are equal to `~patroni.config_generator._NO_VALUE_MSG`.
```
barthisrael, line 53:
```suggestion
        If an exception occured, hostname and ip values are equal to :var:`~patroni.config_generator._NO_VALUE_MSG`.
```

If I recall correctly we can use `var` for module level variable reference.

patroni/config_generator.py
```diff
@@ -0,0 +1,461 @@
...
+        self.config = self.get_template_config()
+
+        self.generate()
+
+    @classmethod
+    def get_template_config(cls) -> Dict[str, Any]:
+        """Generate a template config for further extension (e.g. in the inherited classes).
+
+        :returns: dictionary with the values gathered from Patroni env, hopefully defined hostname and ip address
+        (otherwise set to `~patroni.config_generator._NO_VALUE_MSG`), and some sane defaults.
```
barthisrael, line 91:
```suggestion
        (otherwise set to :var:`~patroni.config_generator._NO_VALUE_MSG`), and some sane defaults.
```

patroni/config_generator.py
```diff
@@ -0,0 +1,461 @@
...
+                }
+            },
+            'restapi': {
+                'connect_address': cls._IP + ':8008',
+                'listen': cls._IP + ':8008'
+            }
+        }
+
+        dynamic_config = Config.get_default_config()
+        dynamic_config['postgresql']['parameters'] = dict(dynamic_config['postgresql']['parameters'])
```
barthisrael, line 119: I guess we do this so we can later dump as YAML, right?
If so, do you think it is worth it a comment about that?

patroni/config_generator.py
```diff
@@ -0,0 +1,461 @@
...
+        config = Config('', None).local_configuration  # Get values from env
+        config.setdefault('bootstrap', {})['dcs'] = dynamic_config
+        config.setdefault('postgresql', {})
+        del config['bootstrap']['dcs']['standby_cluster']
+
+        patch_config(template_config, config)
+        return template_config
+
+    def _get_int_major_version(self) -> int:
```
barthisrael, line 128: Maybe we can move this to `SampleConfigGenerator`? I mean, it is only used in the subclass.

patroni/config_generator.py
```diff
@@ -0,0 +1,461 @@
...
+
+
+class SampleConfigGenerator(AbstractConfigGenerator):
+    """Object representing the generated sample Patroni config.
+
+    Sane defults are used based on the gathered PG version.
+    """
+
+    @property
+    def get_auth_method(self) -> str:
```
barthisrael, line 161:
```suggestion
    def _get_auth_method(self) -> str:
```

I wonder if this should be a "private" property.

patroni/config_generator.py
```diff
@@ -0,0 +1,461 @@
...
+
+    def generate(self) -> None:
+        """Generate sample config using some sane defaults and update :attr:`~AbstractConfigGenerator.config`."""
+        self.pg_major = self._get_int_major_version()
+
+        self.config['postgresql']['parameters'] = {'password_encryption': self.get_auth_method}
+        username = self.config["postgresql"]["authentication"]["replication"]["username"]
+        self.config['postgresql']['pg_hba'] = [
+            f'host all all all {self.get_auth_method}',
+            f'host replication {username} all {self.get_auth_method}'
```
barthisrael, line 176:
```suggestion
        self.config['postgresql']['parameters'] = {'password_encryption': self._get_auth_method}
        username = self.config["postgresql"]["authentication"]["replication"]["username"]
        self.config['postgresql']['pg_hba'] = [
            f'host all all all {self._get_auth_method}',
            f'host replication {username} all {self._get_auth_method}'
```

Just in case you decide to apply the previous suggestion.

patroni/collections.py
```diff
@@ -187,6 +187,13 @@ def copy(self) -> 'CaseInsensitiveDict':
         """
         return CaseInsensitiveDict({v[0]: v[1] for v in self._values.values()})
 
+    def keys(self) -> KeysView[str]:
+        """Return a new view of the dict's keys
```
matthbakeredb, line 191:
```suggestion
        """Return a new view of the dict's keys.
```

patroni/config_generator.py
```diff
@@ -0,0 +1,461 @@
...
+            'superuser': su_params,
+            'replication': {'username': _NO_VALUE_MSG, 'password': _NO_VALUE_MSG}
+        }
+
+    def _set_conf_files(self) -> None:
+        """Extend :attr:`~RunningClusterConfigGenerator.config` with ``pg_hba.conf`` and ``pg_ident.conf`` content.
+
+        .. note::
+            This function only defines ``postgresql.pg_hba`` and ``postgresql.pg_ident`` when
+            ``hba_file`` and ``ident_file`` are set to the defaults.
```
barthisrael, line 345: Maybe it is worth it adding the reason why it's done like that.

patroni/config_generator.py
matthbakeredb, line 91:
```suggestion
        :returns: dictionary with the values gathered from Patroni env, hopefully defined hostname and ip address
                  (otherwise set to `~patroni.config_generator._NO_VALUE_MSG`), and some sane defaults.
```

patroni/config_generator.py
```diff
@@ -0,0 +1,461 @@
...
+    Gather all the available non-internal GUC values having configuration file, postmaster command line or environment
+    variable as a source and store them in the appropriate part of Patroni configuration (``postgresql.parameters`` or
+    ``bootstrap.dcs.postgresql.parameters``). Either the provided DSN (takes precedence) or PG ENV vars will be used
+    for the connection. If password is not provided, it should be entered via prompt.
+
+    The created configuration contains:
+    * ``scope``: ``cluster_name`` GUC value or ``PATRONI_SCOPE ENV`` variable value if available.
+    * ``name``: ``PATRONI_NAME`` ENV variable value if set, otherwise hostname.
+
+    * ``bootsrtap.dcs``: section with all the parameters (incl. the majority of PG GUCs) set to their default values
```
barthisrael, line 421:
```suggestion
    * ``bootstrap.dcs``: section with all the parameters (incl. the majority of PG GUCs) set to their default values
```

patroni/config_generator.py
```diff
@@ -0,0 +1,461 @@
...
+
+        patch_config(template_config, config)
+        return template_config
+
+    def _get_int_major_version(self) -> int:
+        """Get major PostgreSQL version from the binary as an integer.
+
+        :returns: an integer PostgreSQL major version representation gathered from the PostgreSQL binary.
+            See :func:`~patroni.postgresql.misc.postgres_major_version_to_int` and
+            :func:`~patroni.utils.get_major_version`.
```
matthbakeredb, line 133:
```suggestion
        :returns: an integer PostgreSQL major version representation gathered from the PostgreSQL binary.
                  See :func:`~patroni.postgresql.misc.postgres_major_version_to_int` and
                  :func:`~patroni.utils.get_major_version`.
```

patroni/config_generator.py
```diff
@@ -0,0 +1,461 @@
...
+    * ``postgresql.authentication``:
+
+        * superuser and replication users defined (if possible, usernames are set from the respective Patroni ENV vars,
+          otherwise the default ``postgres`` and ``replicator`` values are used).
+          If not a sample config, either DSN or PG ENV vars are used to define superuser authentication parameters.
+
+        * rewind user is defined only for sample config, if PG version can be defined and PG version is >=11
+          (if possible, username is set from the respective Patroni ENV var).
+
+    * ``bootsrtap.dcs.postgresql.use_pg_rewind`` set to ``True`` for a sample config only.
```
barthisrael, line 443:
```suggestion
    * ``bootstrap.dcs.postgresql.use_pg_rewind`` set to ``True`` for a sample config only.
```

patroni/config_generator.py
```diff
@@ -0,0 +1,461 @@
...
+
+    def _set_pg_params(self, cur: Union['cursor', 'Cursor[Any]']) -> None:
+        """Extend :attr:`~RunningClusterConfigGenerator.config` with the actual PG GUCs values.
+
+        THe following GUC values are set:
+
+            * Non-internal having configuration file, postmaster command line or environment variable
+              as a source.
+
+            * List of the always required parameters (see :func:`~RunningClusterConfigGenerator._required_pg_params`).
```
barthisrael, line 285:
```suggestion
            * List of the always required parameters (see :meth:`~RunningClusterConfigGenerator._required_pg_params`).
```

patroni/config_generator.py
```diff
@@ -0,0 +1,461 @@
...
+
+    def _enrich_config_from_running_instance(self) -> None:
+        """Extend :attr:`~RunningClusterConfigGenerator.config` with the values gathered from the running instance.
+
+        Retrieve the following information from the running PostgreSQL instance:
+
+        * superuser auth parameters (see :func:`~RunningClusterConfigGenerator._set_su_params`);
+        * some GUC values (see :func:`~RunningClusterConfigGenerator._set_pg_params`);
+        * ``postgresql.connect_address``, ``postgresql.listen``;
+        * ``postgresql.pg_hba`` and ``postgresql.pg_ident`` (see :func:`~RunningClusterConfigGenerator._set_conf_files`)
```
barthisrael, line 376:
```suggestion
        * superuser auth parameters (see :meth:`~RunningClusterConfigGenerator._set_su_params`);
        * some GUC values (see :meth:`~RunningClusterConfigGenerator._set_pg_params`);
        * ``postgresql.connect_address``, ``postgresql.listen``;
        * ``postgresql.pg_hba`` and ``postgresql.pg_ident`` (see :meth:`~RunningClusterConfigGenerator._set_conf_files`)
```

barthisrael (changes requested):

matthbakeredb (approved):
Looks great! Made some trivial docstring suggestions.

barthisrael:
> @matthbakeredb @barthisrael any other suggestions/objections?

Turns out I started a review a couple minutes before your comment 😆 
I think it is in a good shape. I posted a set of suggestions with some tweaks.

patroni/config_generator.py
hughcapet: is there actually such a directive? 🤔

matthbakeredb: yep!  https://www.sphinx-doc.org/en/master/usage/restructuredtext/domains.html#role-py-meth

patroni/config_generator.py
hughcapet: well, I am not sure. now it is not used outside of the class methods (like anything else except `write_config()`. But will maybe makes sense as an object attr in the future usage.

patroni/config_generator.py
barthisrael: Actually it was not `var`, but `data` -- https://www.sphinx-doc.org/en/master/usage/restructuredtext/domains.html#role-py-data.

I wonder if we should use `const`, though -- https://www.sphinx-doc.org/en/master/usage/restructuredtext/domains.html#role-py-const.

patroni/config_generator.py
barthisrael: your call :)

barthisrael (approved):

CyberDem0n:
:+1:

hughcapet:
👍

PR #2788, opened 2023-07-28
Fix `replicatefrom` tag in `postgres2.yml`

barthisrael:
The node `postgresql2` in the repository is apparently supposed to be a cascading standby.

However, if that is the case, there is a typo in the name of its upstream node.

This commit fixes that issue.

### DIFF
--- a/postgres2.yml
+++ b/postgres2.yml
@@ -121,4 +121,4 @@ tags:
     nofailover: false
     noloadbalance: false
     clonefrom: false
-    replicatefrom: postgres1
+    replicatefrom: postgresql1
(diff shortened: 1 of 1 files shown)


postgres2.yml
```diff
@@ -121,4 +121,4 @@ tags:
     nofailover: false
     noloadbalance: false
     clonefrom: false
-    replicatefrom: postgres1
+    replicatefrom: postgresql1
```
CyberDem0n, line 124:
```suggestion
#    replicatefrom: postgresql1
```

Lets comment it out, just in case.

CyberDem0n:
:+1:

hughcapet:
👍

PR #2790, opened 2023-07-31
Refactor replica_list

CyberDem0n:
As suggested in
https://github.com/zalando/patroni/pull/2668/files#r1276115738, introduce a couple of classes that represent a single replica and collection of replicas.

### DIFF
--- a/patroni/postgresql/sync.py
+++ b/patroni/postgresql/sync.py
@@ -194,7 +194,7 @@ class _ReplicaList(List[_Replica]):
         super().__init__()
 
         # We want to prioritize candidates based on `write_lsn``, ``flush_lsn``, or ``replay_lsn``.
-        # Which exactly column to pick depends on the values of ``synchronous_commit`` GUC.
+        # Which column exactly to pick depends on the values of ``synchronous_commit`` GUC.
         sort_col = {
             'remote_apply': 'replay',
             'remote_write': 'write'
@@ -274,7 +274,7 @@ END;$$""")
         :param replica_list: collection of replicas that we want to evaluate.
         """
         for replica in replica_list:
-            # if standby name is listed in the /sync key we can count it as synchronous, otherwice
+            # if standby name is listed in the /sync key we can count it as synchronous, otherwise
             # it becomes really synchronous when sync_state = 'sync' and it is known that it managed to catch up
             if replica.application_name not in self._ready_replicas\
                     and replica.application_name in self._ssn_data.members\
--- a/patroni/postgresql/sync.py
+++ b/patroni/postgresql/sync.py
@@ -153,6 +153,72 @@ def parse_sync_standby_names(value: str) -> _SSN:
     return _SSN(sync_type, has_star, num, members)
 
 
+class _Replica(NamedTuple):
+    """Class representing a single replica that is eligible to be synchronous.
+
+    Attributes are taken from ``pg_stat_replication`` view and respective ``Cluster.members``.
+
+    :ivar pid: PID of walsender process.
+    :ivar application_name: matches with the ``Member.name``.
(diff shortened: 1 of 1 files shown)


patroni/postgresql/sync.py
```diff
@@ -153,6 +153,72 @@ def parse_sync_standby_names(value: str) -> _SSN:
...
+    def __init__(self, postgresql: 'Postgresql', cluster: Cluster) -> None:
+        """Create :class:``_ReplicaList`` object.
+
+        :param postgresql: reference to :class:``Postgresql`` object.
+        :param cluster: currently known cluster state from DCS.
+        """
+        super().__init__()
+
+        # We want to prioritize candidates based on `write_lsn``, ``flush_lsn``, or ``replay_lsn``.
+        # Which exactly column to pick depends on the values of ``synchronous_commit`` GUC.
```
matthbakeredb, line 197:
```suggestion
        # Which exactly column exactly to pick depends on the values of ``synchronous_commit`` GUC.
```

patroni/postgresql/sync.py
```diff
@@ -153,6 +153,72 @@ def parse_sync_standby_names(value: str) -> _SSN:
...
+            #   a. ``Member`` doesn't have ``nosync`` tag set;
+            #   b. PostgreSQL on the member is known to be running and accepting client connections.
+            if member and row[sort_col] is not None and member.is_running and not member.tags.get('nosync', False):
+                self.append(_Replica(row['pid'], row['application_name'],
+                                     row['sync_state'], row[sort_col], bool(member.nofailover)))
+
+        # Prefer replicas that are in state ``sync`` and with higher values of ``write``/``flush``/``replay`` LSN.
+        self.sort(key=lambda r: (r.sync_state, r.lsn), reverse=True)
+
+        self.max_lsn = max(self, key=lambda x: x.lsn).lsn if len(self) > 1 else postgresql.last_operation()
```
matthbakeredb, line 219: Is this OK to be set once during class init and not refreshed when called? Would a property be better?

```suggestion
    @property
    def max_lsn(self):
        return = max(self, key=lambda x: x.lsn).lsn if len(self) > 1 else self._postgresql.last_operation()
```

(will need to make _postgresql and attr of the class)

patroni/postgresql/sync.py
```diff
@@ -201,6 +267,21 @@ def _handle_synchronous_standby_names_change(self) -> None:
...
         self._postgresql.reset_cluster_info_state(None)  # Reset internal cache to query fresh values
 
+    def _process_replica_readiness(self, cluster: Cluster, replica_list: _ReplicaList) -> None:
+        """Flags replicas as truly "synchronous" when they have caught up with ``_primary_flush_lsn``.
+
+        :param cluster: current cluster topology from DCS
+        :param replica_list: collection of replicas that we want to evaluate.
+        """
+        for replica in replica_list:
+            # if standby name is listed in the /sync key we can count it as synchronous, otherwice
```
matthbakeredb, line 277:
```suggestion
            # if standby name is listed in the /sync key we can count it as synchronous, otherwise
```

matthbakeredb (commented):
Awesome, looks very cool. Couple of minor spelling things and a question/comment about max_lsn.

patroni/postgresql/sync.py
CyberDem0n: It doesn't make a lot of sense, because:
1. We need to add _postgresql as an attribute
2. In worst case the property will execute `max()` for every access.
3. It will always return the same value.

matthbakeredb (approved):
lgtm

hughcapet:
👍

CyberDem0n:
:+1:

PR #2793, opened 2023-07-31
Resolve simple TODO in api

mfpekala-nutanix:
Tries and catches writing response to the api socket and logs ConnectionResetErrors


CyberDem0n (commented):
It is not that simple as it looks like.
This (or similar) exception could be raised from anywhere:
- SSL handshake
- end_headers()
- any self.wfile.write() call
- SSL shutdown

They all indicate that something went unexpectedly...

CyberDem0n:
The only question is - what are trade offs?
Yes, we can make an effort and silence these errors, but as a result users will miss something important.

PR #2794, opened 2023-08-01
Fix bug in the Cluster class

CyberDem0n:
The `workers` attribute when not passed explicitly was set to the same mutable `dict` object every time.

Problem was introduced in #2652


matthbakeredb (approved):
lgtm

hughcapet:
👍

CyberDem0n:
:+1:

PR #2795, opened 2023-08-01
Fix behaviour of replicas in standby cluster in pause

CyberDem0n:
When the leader key expires replicas should not follow the remote node but keep primary_conninfo as it is.


hughcapet:
👍

CyberDem0n:
:+1:

vitabaks:
Thanks

PR #2796, opened 2023-08-01
Refactor docs about migration to Patroni

barthisrael:
This PR is an attempt of refactoring the docs about migration to Patroni.

These are a few enhancements that we propose through this PR:

* Docs used to mention the procedure can only be performed in a single-node cluster. We changed that so the procedure considers a cluster composed of primary and standbys;
* Teach how to deal with pre-existing replication slots;
* Explain how to create the user for `pg_rewind`, if user intends to enable `use_pg_rewind`.

References: PAT-143.


docs/existing_data.rst
```diff
@@ -10,18 +10,51 @@ To deploy a Patroni cluster without using a pre-existing PostgreSQL instance, se
...
 
-.. code-block:: sql
+    #. If you are running Postgres through systemd, then disable the Postgres systemd unit. This is performed as Patroni manages starting and stopping the Postgres daemon.
+    #. Create a YAML configuration file for Patroni. If you have replication slots being used for replication between cluster members, then it is recommended that you enable ``use_slots`` and configure the existing replication slots as permanent via the ``slots`` configuration item. Be aware that Patroni automatically creates replication slots for replication between members, and drops replication slots that it does not recognize, when ``use_slots`` is enabled. The idea of using permanent slots here is to allow your existing slots to persist while the migration to Patroni is in progress. See :ref:`YAML Configuration Settings <yaml_configuration>` for details.
+    #. Create the Postgres users as defined in the :ref:`authentication <postgresql_settings>` section of the Patroni configuration. You can find sample SQL commands to create the users in the code block below. Replace the usernames and passwords as per your configuration.
```
CyberDem0n, line 19: Most of it is only for the primary:
- only the leader (primary) writes `bootstrap.dcs` to config key, and therefore other nodes don't even need to have it
- it is possible to create users only on primary

docs/existing_data.rst
```diff
@@ -10,18 +10,51 @@ To deploy a Patroni cluster without using a pre-existing PostgreSQL instance, se
...
+
+           -- Patroni replication user
+           CREATE USER PATRONI_REPLICATION_USERNAME WITH REPLICATION ENCRYPTED PASSWORD '$PATRONI_REPLICATION_PASSWORD';
+
+           -- Patroni rewind user, if you enabled use_pg_rewind in your configuration
+           CREATE USER PATRONI_REWIND_USERNAME WITH ENCRYPTED PASSWORD '$PATRONI_REWIND_PASSWORD';
+           GRANT EXECUTE ON function pg_catalog.pg_ls_dir(text, boolean, boolean) TO PATRONI_REWIND_USERNAME;
+           GRANT EXECUTE ON function pg_catalog.pg_stat_file(text, boolean) TO PATRONI_REWIND_USERNAME;
+           GRANT EXECUTE ON function pg_catalog.pg_read_binary_file(text) TO PATRONI_REWIND_USERNAME;
+           GRANT EXECUTE ON function pg_catalog.pg_read_binary_file(text, bigint, bigint, boolean) TO PATRONI_REWIND_USERNAME;
```
CyberDem0n, line 34: would be nice to explicitly tell that they should replace user names with whatever they use.

docs/existing_data.rst
```diff
@@ -10,18 +10,51 @@ To deploy a Patroni cluster without using a pre-existing PostgreSQL instance, se
...
+
+    #. Start Patroni using the `patroni` systemd service unit. It automatically detects that Postgres is already running and starts monitoring the instance.
+
+#. Hand over Postgres "start up procedure" to Patroni. In order to do that you need to restart the cluster members through ``patronictl restart cluster-name member-name`` command. For minimal downtime you might want to split this step into:
+
+    #. Immediate restart of the standby nodes.
+    #. Scheduled restart of the primary node within a maintenance window.
```
CyberDem0n, line 41: 1. Restart order depends on mismatch of actual postgres configuration and desired. For example if postgres runs with max_connections = 1000 and you set it to 500, the procedure will not work. I.e., Patroni will start replica back with the original value (1000).
2. Patroni might automatically restart replica nodes in case if primary_conninfo, primary_slot_name, restore_command or similar doesn't match, but only for Postgres older than 13.

CyberDem0n (commented):
somehow GH renders strangely...
image

docs/existing_data.rst
```diff
@@ -10,18 +10,51 @@ To deploy a Patroni cluster without using a pre-existing PostgreSQL instance, se
...
+
+#. If you configured permanent slots in step ``1.2.``, then you should remove them from ``slots`` configuration once the ``restart_lsn`` of the slots created by Patroni is able to catch up with the ``restart_lsn`` of the original slots for the corresponding members. By removing the slots from ``slots`` configuration you will allow Patroni to drop the original slots from your cluster once they are not needed anymore. You can find below an example query to check the ``restart_lsn`` of a couple slots, so you an compare them:
```
hughcapet, line 43:
```suggestion
#. If you configured permanent slots in step ``1.2.``, then you should remove them from ``slots`` configuration once the ``restart_lsn`` of the slots created by Patroni is able to catch up with the ``restart_lsn`` of the original slots for the corresponding members. By removing the slots from ``slots`` configuration you will allow Patroni to drop the original slots from your cluster once they are not needed anymore. You can find below an example query to check the ``restart_lsn`` of a couple slots, so you can compare them:
```

hughcapet: if I got it right 😅

docs/existing_data.rst
barthisrael: Good point, let me amend it.

docs/existing_data.rst
barthisrael: I initially put a note about that in the step itself:

```rst
#. Create the Postgres users as defined in the :ref:`authentication <postgresql_settings>` section of the Patroni configuration. You can find sample SQL commands to create the users in the code block below. Replace the usernames and passwords as per your configuration.
```

But I can add that as comments in the code-block too if you think that is valuable.

docs/existing_data.rst
barthisrael: My idea for this documentation was that users would migrate an existing cluster and *do not* change configuration.
Maybe we could a put an explicit note about that assumption in the docs? What do you think?

BTW this PR https://github.com/zalando/patroni/pull/2786 could come handy here in regards to creating an YAML file, huh? :)

barthisrael:
I just pushed a commit with a few changes based on your comments.
What do you think of this new list of steps?
About the link in your screenshot, weird, I can see the proper links when running `tox -m docs`.

hughcapet:
> About the link in your screenshot, 

I think 3 spaces instead of 4 for the nested level of a list will work for both rst rendering in gh and for sphinx 🤔

barthisrael:
> > About the link in your screenshot,
> 
> I think 3 spaces instead of 4 for the nested level of a list will work for both rst rendering in gh and for sphinx 🤔

I got it wrong, I thought the issue was with links, not with the block quote. Let me adjust it.

barthisrael:
> > > About the link in your screenshot,
> > 
> > 
> > I think 3 spaces instead of 4 for the nested level of a list will work for both rst rendering in gh and for sphinx 🤔
> 
> I got it wrong, I thought the issue was with links, not with the block quote. Let me adjust it.

Actually it is now using 4 spaces from what I can see -- it used to be 3 in previous commits. 🤔

barthisrael:
> > it used to be 3 in previous commits.
> 
> which commits?

This one for example: https://github.com/zalando/patroni/pull/2796/commits/54df5fbdfc80aae660d5019a6c6a0fa2299b6499#diff-5a02bb6c63e019ee11ac16bf9e81be2daf115b349c44808365b961955a55af75R24-R35.

But I just noticed it had 3 leading spaces only in the SQL code block, not in the actual list.
I'm not sure why GitHub renders the file with block quotes to show the indented lists.

hughcapet:
Yes, it is strange, but I guess 3 spaces also don't break sphinx? Can we switch to it here?

barthisrael:
> Yes, it is strange, but I guess 3 spaces also don't break sphinx? Can we switch to it here?

Sure, just submitted a commit with that change. I intended to have already submitted, but got side tracked here.

The file now looks nice in GitHub as you mentioned. Also looks fine through sphinx, I tested it locally.

hughcapet:
👍

CyberDem0n:
:+1:

PR #2797, opened 2023-08-01
Enhance docs of `slots` dynamic configuration

barthisrael:
This PR is an attempt to enhance the documentation of `slots` dynamic configuration in terms of:

* Expanding on the following mention, and explaining how the usage of member slots names in permanent slots can be useful -- this makes a reference to the work performed through #2796:

> Everything else is the responsibility of the operator to make sure that there are no clashes in names between replication slots automatically created by Patroni for members and permanent replication slots.

* Add missing information on how Patroni deals with slots:
    * It creates non-existing slots
    * It maintains physical slots only in the primary

References: PAT-144.


docs/dynamic_configuration.rst
```diff
@@ -46,9 +46,9 @@ In order to change the dynamic configuration you can use either ``patronictl edi
...
+   -  **my\_slot\_name**: the name of the permanent replication slot. If the permanent slot name matches with the name of the current primary it will not be created. Please note that Patroni does not make checks for permanent slot names added to this configuration matching those that Patroni creates automatically for members. If those names are added, Patroni will ensure that any slots that were created are not removed even if the member becomes unresponsive, situation which would normally result in the slot's removal by Patroni. Although this can be useful in some situations, such as when importing existing members to a new Patroni cluster (see :ref:`Convert a Standalone to a Patroni Cluster <_existing_data>` for details), caution should be exercised by the operator that these clashes in names are not persisted in the DCS due to its effect on normal functioning of Patroni.
```
CyberDem0n, line 51:
```suggestion
   -  **my\_slot\_name**: the name of the permanent replication slot. If the permanent slot name matches with the name of existing member of the Patroni cluster it will not be created. Please note that Patroni does not make checks for permanent slot names added to this configuration matching those that Patroni creates automatically for members. If those names are added, Patroni will ensure that any slots that were created are not removed even if the member becomes unresponsive, situation which would normally result in the slot's removal by Patroni. Although this can be useful in some situations, such as when importing existing members to a new Patroni cluster (see :ref:`Convert a Standalone to a Patroni Cluster <_existing_data>` for details), caution should be exercised by the operator that these clashes in names are not persisted in the DCS due to its effect on normal functioning of Patroni.
```

barthisrael: I took a new read at the code and it seem the behavior is:

- For logical slots: do not create if it matches an existing member name
- For physical slots: do not create if it matches the current leader name

Is that accurate? If so we can adjust the above a bit.

CyberDem0n: Sorry, my bad, I misread it.

There is just one rule: leader doesn't create a permanent slot (doesn't matter whether it logical or physical) that matches with its own name

The rest just follows from it. I.e., if there is no permanent logical slot on the primary, replicas will not be able to copy it.

barthisrael, line 51:
```suggestion
   -  **my\_slot\_name**: the name of the permanent replication slot. If the permanent slot name matches with the name of the current leader it will not be created. Please note that Patroni does not make checks for permanent slot names added to this configuration matching those that Patroni creates automatically for members. If those names are added, Patroni will ensure that any slots that were created are not removed even if the member becomes unresponsive, situation which would normally result in the slot's removal by Patroni. Although this can be useful in some situations, such as when importing existing members to a new Patroni cluster (see :ref:`Convert a Standalone to a Patroni Cluster <_existing_data>` for details), caution should be exercised by the operator that these clashes in names are not persisted in the DCS due to its effect on normal functioning of Patroni.
```

Cool! What about this then?

CyberDem0n: lgtm

docs/dynamic_configuration.rst
```diff
@@ -46,9 +46,9 @@ In order to change the dynamic configuration you can use either ``patronictl edi
...
+   -  **my\_slot\_name**: the name of the permanent replication slot. If the permanent slot name matches with the name of the current leader it will not be created. Please note that Patroni does not make checks for permanent slot names added to this configuration matching those that Patroni creates automatically for members. If those names are added, Patroni will ensure that any slots that were created are not removed even if the member becomes unresponsive, situation which would normally result in the slot's removal by Patroni. Although this can be useful in some situations, such as when importing existing members to a new Patroni cluster (see :ref:`Convert a Standalone to a Patroni Cluster <_existing_data>` for details), caution should be exercised by the operator that these clashes in names are not persisted in the DCS due to its effect on normal functioning of Patroni.
```
hughcapet, line 51:
```suggestion
   -  **my\_slot\_name**: the name of the permanent replication slot. If the permanent slot name matches with the name of the current leader it will not be created. Please note that Patroni does not make checks for permanent slot names added to this configuration matching those that Patroni creates automatically for members. If those names are added, Patroni will ensure that any slots that were created are not removed even if the member becomes unresponsive, situation which would normally result in the slot's removal by Patroni. Although this can be useful in some situations, such as when importing existing members to a new Patroni cluster (see :ref:`Convert a Standalone to a Patroni Cluster <existing_data>` for details), caution should be exercised by the operator that these clashes in names are not persisted in the DCS due to its effect on normal functioning of Patroni.
```

CyberDem0n:
:+1:

PR #2802, opened 2023-08-03
Mock request() method when running tests

CyberDem0n:
1. Unit tests should not really try accessing any resources.
2. Not doing so results in significant execution time of unit tests on Windows

In addition to that perform a request with timeout 3s. Usually this is more than enough to figure out whether resource is accessible.

Followup on #2724


hughcapet:
👍

CyberDem0n:
:+1:

PR #2803, opened 2023-08-04
reset failsafe state when promote

ChenChangAo:
consider the scenario(enable failsafe_mode):
0. node1(primary) - node2(replica)
1. stop all etcd nodes; wait ttl seconds; start all etcd nodes; (node2's failsafe will contain the info about node1)
2. switchover to node2; (node2's failsafe still contain the info about node1)
3. stop all etcd nodes; wait ttl seconds; start all etcd nodes;
4. node2 will demote because it consider node1 as primary
reset failsafe state when promote could solve this problem.

### DIFF
--- a/patroni/ha.py
+++ b/patroni/ha.py
@@ -83,11 +83,7 @@ class Failsafe(object):
     def __init__(self, dcs: AbstractDCS) -> None:
         self._lock = RLock()
         self._dcs = dcs
-        self._last_update = 0
-        self._name = None
-        self._conn_url = None
-        self._api_url = None
-        self._slots = None
+        self._reset_state()
 
     def update(self, data: Dict[str, Any]) -> None:
         with self._lock:
@@ -97,13 +93,12 @@ class Failsafe(object):
             self._api_url = data['api_url']
             self._slots = data.get('slots')
 
-    def reset_state(self) -> None:
-        with self._lock:
-            self._last_update = 0
-            self._name = None
-            self._conn_url = None
-            self._api_url = None
-            self._slots = None
+    def _reset_state(self) -> None:
+        self._last_update = 0
+        self._name = None
+        self._conn_url = None
+        self._api_url = None
+        self._slots = None
 
     @property
     def leader(self) -> Optional[Leader]:
@@ -138,6 +133,8 @@ class Failsafe(object):
     def set_is_active(self, value: float) -> None:
         with self._lock:
             self._last_update = value
+            if not value:
+                self._reset_state()
 
 
 class Ha(object):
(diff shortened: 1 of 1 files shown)


CyberDem0n:
It seems that you didn't took into an account that the failsafe state has limited lifetime expires after ttl: https://github.com/zalando/patroni/blob/84aac437c1d9437a71f26b81655292afe5124974/patroni/ha.py#L100-L114

Between 2 and 4 it takes more time than ttl and therefore it will not be possible to reproduce it.

ChenChangAo:
> It seems that you didn't took into an account that the failsafe state has limited lifetime expires after ttl:
> 
> https://github.com/zalando/patroni/blob/84aac437c1d9437a71f26b81655292afe5124974/patroni/ha.py#L100-L114
> 
> Between 2 and 4 it takes more time than ttl and therefore it will not be possible to reproduce it.

ChenChangAo:
image
primary will set it active in _handle_dcs_error

patroni/ha.py
```diff
@@ -770,6 +778,9 @@ def enforce_primary_role(self, message: str, promote_message: str) -> str:
                 self.state_handler.sync_handler.set_synchronous_standby_names(
                     CaseInsensitiveSet('*') if self.global_config.is_synchronous_mode_strict else CaseInsensitiveSet())
             if self.state_handler.role not in ('master', 'promoted', 'primary'):
+                # reset failsafe state when promote
+                self._failsafe.reset_state()
+
```
CyberDem0n, line 783:
```suggestion
```

patroni/ha.py
```diff
@@ -97,6 +97,14 @@ def update(self, data: Dict[str, Any]) -> None:
             self._api_url = data['api_url']
             self._slots = data.get('slots')
 
+    def reset_state(self) -> None:
+        with self._lock:
+            self._last_update = 0
+            self._name = None
+            self._conn_url = None
+            self._api_url = None
+            self._slots = None
```
CyberDem0n, line 106:
```suggestion
    def _reset_state(self) -> None:
        self._last_update = 0
        self._name = None
        self._conn_url = None
        self._api_url = None
        self._slots = None
```

CyberDem0n (changes requested):
Indeed...
Actually, `set_is_active()` seems to be a better place for it.
On top of suggested changes please apply the following:
```patch
diff --git a/patroni/ha.py b/patroni/ha.py
index 5603ba44..ddcc1fc2 100644
--- a/patroni/ha.py
+++ b/patroni/ha.py
@@ -83,11 +83,7 @@ class Failsafe(object):
     def __init__(self, dcs: AbstractDCS) -> None:
         self._lock = RLock()
         self._dcs = dcs
-        self._last_update = 0
-        self._name = None
-        self._conn_url = None
-        self._api_url = None
-        self._slots = None
+        self._reset_state()
 
     def update(self, data: Dict[str, Any]) -> None:
         with self._lock:
@@ -130,6 +133,8 @@ class Failsafe(object):
     def set_is_active(self, value: float) -> None:
         with self._lock:
             self._last_update = value
+            if not value:
+                self._reset_state()

```

CyberDem0n (approved):

hughcapet:
👍

CyberDem0n:
:+1:

CyberDem0n:
Merged, thank you @ChenChangAo!

PR #2807, opened 2023-08-07
[Refactor] Rename _is_leader to _leader_expiry

mfpekala-nutanix:
In `class Ha` the private variable `_is_leader` reads like a `bool` and is initialized as a `bool`, but in practice it is a numeric timestamp representing the the time that this node's leadership will expire (0 if explicitly not leader).

Renamed to `_leader_expiry` and updated initialization to reflect this without changing the behavior of the public `is_leader` or `set_is_leader` functions

### DIFF
--- a/patroni/ha.py
+++ b/patroni/ha.py
@@ -147,8 +147,8 @@ class Ha(object):
         self.cluster = Cluster.empty()
         self.global_config = self.patroni.config.get_global_config(None)
         self.old_cluster = Cluster.empty()
-        self._is_leader = False
-        self._is_leader_lock = RLock()
+        self._leader_expiry = 0
+        self._leader_expiry_lock = RLock()
         self._failsafe = Failsafe(patroni.dcs)
         self._was_paused = False
         self._leader_timeline = None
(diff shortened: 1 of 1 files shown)


patroni/ha.py
```diff
@@ -193,12 +193,12 @@ def is_standby_cluster(self) -> bool:
...
-        with self._is_leader_lock:
-            return self._is_leader > time.time()
+        with self._leader_expiry_lock:
+            return self._leader_expiry > time.time()
 
     def set_is_leader(self, value: bool) -> None:
-        with self._is_leader_lock:
-            self._is_leader = time.time() + self.dcs.ttl if value else 0
+        with self._leader_expiry_lock:
+            self._leader_expiry = time.time() + self.dcs.ttl if value else 0
```
CyberDem0n, line 201: Since you already changed these methods please also add docstrings.

patroni/ha.py
```diff
@@ -193,12 +193,18 @@ def is_standby_cluster(self) -> bool:
...
-        with self._is_leader_lock:
-            return self._is_leader > time.time()
+        """:returns: `True` if the current node is the leader, based on expiration set when it last held the key."""
+        with self._leader_expiry_lock:
+            return self._leader_expiry > time.time()
 
     def set_is_leader(self, value: bool) -> None:
-        with self._is_leader_lock:
-            self._is_leader = time.time() + self.dcs.ttl if value else 0
+        """Updates the current node's view of it's own leadership status. Will update the expiry timestamp to match
```
hughcapet, line 201: http://www.pydocstyle.org/en/2.1.1/error_codes.html

D205, D209, D400 (you can use flake8-docstrings to track if all issues are resolved)

patroni/ha.py
```diff
@@ -193,12 +193,20 @@ def is_standby_cluster(self) -> bool:
...
 
     def set_is_leader(self, value: bool) -> None:
-        with self._is_leader_lock:
-            self._is_leader = time.time() + self.dcs.ttl if value else 0
+        """Update the current node's view of it's own leadership status.
+
+        Will update the expiry timestamp to match the dcs ttl if setting leadership to true,
+        otherwise will set the expiry to the past to immediately invalidate.
+
+        :param value: Is the current node the leader.
```
hughcapet, line 206:
```suggestion
        :param value: is the current node the leader.
```

hughcapet:
👍

CyberDem0n:
:+1:

hughcapet:
👍

PR #2811, opened 2023-08-10
Make docs more clear on changes to the `bootstrap.dcs` section

matthbakeredb:
It seems that a common pitfall for new users of Patroni is that the `bootstrap.dcs` section is only used to initialize the configuration in DCS. This moves the comment about this to an info block so it is more visible to the reader.


hughcapet:
👍

PR #2812, opened 2023-08-10
[Docs] HA Docstrings

mfpekala-nutanix:
Takes a first stab at providing useful and correctly formatted dostrings in the `HA.py` class.

### DIFF
--- a/patroni/ha.py
+++ b/patroni/ha.py
@@ -147,8 +147,8 @@ class Ha(object):
         self.cluster = Cluster.empty()
         self.global_config = self.patroni.config.get_global_config(None)
         self.old_cluster = Cluster.empty()
-        self._is_leader = False
-        self._is_leader_lock = RLock()
+        self._leader_expiry = 0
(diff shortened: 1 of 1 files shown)


hughcapet:
Please, reference [this PR](https://github.com/zalando/patroni/pull/2777/files) as an example of how docstring PRs should look like (start with checking if params and return values descriptions are present everywhere)

matthbakeredb:
As you probably know, the doc style is reStructuredText (or reST). There's plenty of docs out there but the best starter is this for Sphinx as that's what builds the docs.

https://www.sphinx-doc.org/en/master/usage/restructuredtext/basics.html

There's a PR https://github.com/zalando/patroni/pull/2699 which has the tooling for you to generate the docs locally. Hopefully, that will be merged soon. ;-) We've been using it to test docs we write. You can also use `pydocstyle` http://www.pydocstyle.org/en/stable/ to lint docs.

hth

hughcapet:
Closing this PR due to inactivity

PR #2813, opened 2023-08-14
Use pg_current_wal_flush_lsn() starting from 9.6

CyberDem0n:
Due to historical reasons (not available before 9.6) we used `pg_current_wal_lsn()`/`pg_current_xlog_location()` functions to get current WAL LSN on the primary. But, this LSN is not necessarily synced to disk, and could be lost if the primary node crashed.

### DIFF
--- a/patroni/api.py
+++ b/patroni/api.py
@@ -448,7 +448,7 @@ class RestApiHandler(BaseHTTPRequestHandler):
         * ``patroni_postmaster_start_time``: epoch timestamp since Postmaster was started;
         * ``patroni_master``: ``1`` if this node holds the leader lock, else ``0``;
         * ``patroni_primary``: same as ``patroni_master``;
-        * ``patroni_xlog_location``: ``pg_wal_lsn_diff(pg_current_wal_lsn(), '0/0')`` if leader, else ``0``;
+        * ``patroni_xlog_location``: ``pg_wal_lsn_diff(pg_current_wal_flush_lsn(), '0/0')`` if leader, else ``0``;
         * ``patroni_standby_leader``: ``1`` if standby leader node, else ``0``;
         * ``patroni_replica``: ``1`` if a replica, else ``0``;
         * ``patroni_sync_standby``: ``1`` if a sync replica, else ``0``;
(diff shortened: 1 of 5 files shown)


patroni/api.py
```diff
@@ -1179,8 +1179,8 @@ def get_postgresql_status(self, retry: bool = False) -> Dict[str, Any]:
                     " application_name, client_addr, w.state, sync_state, sync_priority"
                     " FROM pg_catalog.pg_stat_get_wal_senders() w, pg_catalog.pg_stat_get_activity(pid)) AS ri")
 
-            row = self.query(stmt.format(postgresql.wal_name, postgresql.lsn_name), retry=retry)[0]
-
+            row = self.query(stmt.format(postgresql.wal_name, postgresql.lsn_name,
+                                         postgresql.wal_flush), retry=retry)[0]
```
hughcapet, line 1183: can we add a unit test that verifies the proper function call for all 3 cases?

patroni/postgresql/__init__.py
```diff
@@ -58,7 +58,7 @@ class Postgresql(object):
               "ELSE ('x' || pg_catalog.substr(pg_catalog.pg_{0}file_name("
               "pg_catalog.pg_current_{0}_{1}()), 1, 8))::bit(32)::int END, "  # primary timeline
               "CASE WHEN pg_catalog.pg_is_in_recovery() THEN 0 "
-              "ELSE pg_catalog.pg_{0}_{1}_diff(pg_catalog.pg_current_{0}_{1}(), '0/0')::bigint END, "  # write_lsn
+              "ELSE pg_catalog.pg_{0}_{1}_diff(pg_catalog.pg_current_{0}{2}_{1}(), '0/0')::bigint END, "  # flush_lsn
```
hughcapet, line 61: probably better to mention that it is flush for >=9.6 only

patroni/api.py
```diff
@@ -1141,7 +1141,7 @@ def get_postgresql_status(self, retry: bool = False) -> Dict[str, Any]:
             * ``server_version``: Postgres version without periods, e.g. ``150002`` for Postgres ``15.2``;
             * ``xlog``: dictionary. Its structure depends on ``role``:
                 * If ``master``:
-                    * ``location``: ``pg_current_wal_lsn()``
+                    * ``location``: ``pg_current_wal_flush_lsn()``
```
hughcapet, line 1144: I have a feeling, we have to reveal these details in the user documentation for API as well, along with the versions' peculiarities
`Current location of the Postgres transaction log` doesn't look like enough info.
Wdyt?

CyberDem0n: 9.6 and even 10 are EOL, I don't think it is worth time spend on it.

hughcapet: ok, regardless the older versions, there are still quite some versions of what can be the "Current location" here

hughcapet:
👍

CyberDem0n:
:+1:

PR #2818, opened 2023-08-16
Fix `IntValidator` regarding validation of value `0`

### LINKED ISSUE #2817: wrong validate for value = 0
### What happened?

if i set max_prepared_transaction = 0 i have warning:
`2023-08-16 10:51:35,219 - WARNING - postgresql parameter max_prepared_transactions=0 failed validation, defaulting to 0`

set max_prepared_transaction = 0 (or other parametr, max_connections for example)

if value 0 is correct, than don't report it

-   Patroni version: 3.1.0
-   PostgreSQL version: 13
-   DCS (and its version): etcd

```yaml
loop_wait: 10
master_start_timeout: 300
postgresql:
  parameters:
    max_connections: 100
    max_locks_per_transaction: 64
    max_prepared_transactions: 0
    max_replication_slots: 10
    max_wal_senders: 4
    max_worker_processes: 8
    wal_keep_segments: 50
    wal_log_hints: true
  use_pg_rewind: true
  use_slots: true
retry_timeout: 10
ttl: 30
```

### patronictl show-config

```yaml
loop_wait: 10
master_start_timeout: 300
postgresql:
  parameters:
    max_connections: 100
    max_locks_per_transaction: 64
    max_prepared_transactions: 0
    max_replication_slots: 10
    max_wal_senders: 4
    max_worker_processes: 8
    wal_keep_segments: 50
    wal_log_hints: true
  use_pg_rewind: true
  use_slots: true
retry_timeout: 10
ttl: 30
```

### Patroni log files

```shell
no
```

### PostgreSQL log files

```shell
no
```

- [X] Yes

_No response_

barthisrael:
Previous to this commit `IntValidator` would always consider the value `0` invalid, even if in the allowed range.

The problem was that `parse_int` was returning `0` in the following line:

```python
value = parse_int(value, self.base_unit) or ""
```

However the `or ""` was evaluating to an empty string.

As `parse_int` returns either an `int` if able to parse, or `None` otherwise, the `isinstance(value, int)` is enough to error out when not a valid `int`.

Closes #2817

### DIFF
--- a/patroni/validator.py
+++ b/patroni/validator.py
@@ -767,7 +767,7 @@ class IntValidator(object):
         :param value: value to be checked against the rules defined for this :class:`IntValidator` instance.
         :returns: ``True`` if *value* is valid and within the expected range.
         """
-        value = parse_int(value, self.base_unit) or ""
+        value = parse_int(value, self.base_unit)
         ret = isinstance(value, int)\
             and (self.min is None or value >= self.min)\
             and (self.max is None or value <= self.max)


CyberDem0n:
:+1:

hughcapet:
👍

PR #2823, opened 2023-08-17
Factor out tags handling into a dedicated class

CyberDem0n:
The same (almost) logic was used in three different places:
1. `Patroni` class
2. `Member` class
3. `_MemberStatus` class

Now they all inherit newly intoduced `Tags` class.

### DIFF
--- a/patroni/ha.py
+++ b/patroni/ha.py
@@ -20,31 +20,28 @@ from .postgresql.callback_executor import CallbackAction
 from .postgresql.misc import postgres_version_to_int
 from .postgresql.postmaster import PostmasterProcess
 from .postgresql.rewind import Rewind
+from .tags import Tags
 from .utils import polling_loop, tzutc
 
 logger = logging.getLogger(__name__)
 
 
-class _MemberStatus(NamedTuple):
-    """Node status distilled from API response:
+class _MemberStatus(Tags, NamedTuple('_MemberStatus',
+                                     [('member', Member),
+                                      ('reachable', bool),
+                                      ('in_recovery', Optional[bool]),
+                                      ('wal_position', int),
+                                      ('data', Dict[str, Any])])):
+    """Node status distilled from API response.
 
-        member - dcs.Member object of the node
-        reachable - `!False` if the node is not reachable or is not responding with correct JSON
-        in_recovery - `!True` if pg_is_in_recovery() == true
-        dcs_last_seen - timestamp from JSON of last succesful communication with DCS
-        timeline - timeline value from JSON
-        wal_position - maximum value of `replayed_location` or `received_location` from JSON
-        tags - dictionary with values of different tags (i.e. nofailover)
-        watchdog_failed - indicates that watchdog is required by configuration but not available or failed
+    Consists of the following fields:
+
+    :ivar member: :class:`Member` object of the node.
+    :ivar reachable: ``False`` if the node is not reachable or is not responding with correct JSON.
+    :ivar in_recovery: ``False`` if the node is running as a primary (`if pg_is_in_recovery() == true`).
+    :ivar wal_position: maximum value of ``replayed_location`` or ``received_location`` from JSON.
+    :ivar data: the whole JSON response for future usage.
     """
-    member: Member
-    reachable: bool
-    in_recovery: Optional[bool]
-    dcs_last_seen: int
-    timeline: int
-    wal_position: int
-    tags: Dict[str, Any]
-    watchdog_failed: bool
 
     @classmethod
     def from_api_response(cls, member: Member, json: Dict[str, Any]) -> '_MemberStatus':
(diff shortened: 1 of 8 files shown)


patroni/tags.py
```diff
@@ -0,0 +1,60 @@
...
+                if tag not in ('clonefrom', 'nofailover', 'noloadbalance', 'nosync') or value}
+
+    @property
+    @abc.abstractmethod
+    def tags(self) -> Dict[str, Any]:
+        raise NotImplementedError  # pragma: no cover
+
+    @property
+    def clonefrom(self) -> bool:
+        """``True`` if ``clonefrom`` tag is ``True``."""
```
barthisrael, line 39:
```suggestion
        """``True`` if ``clonefrom`` tag is ``True``, else ``False``."""
```

patroni/ha.py
```diff
@@ -20,31 +20,28 @@
...
-        reachable - `!False` if the node is not reachable or is not responding with correct JSON
-        in_recovery - `!True` if pg_is_in_recovery() == true
-        dcs_last_seen - timestamp from JSON of last succesful communication with DCS
-        timeline - timeline value from JSON
-        wal_position - maximum value of `replayed_location` or `received_location` from JSON
-        tags - dictionary with values of different tags (i.e. nofailover)
-        watchdog_failed - indicates that watchdog is required by configuration but not available or failed
+    Consists of the following fields:
+
+    :ivar member: :class:`Member` object of the node.
```
barthisrael, line 39:
```suggestion
    :ivar member: :class:`~patroni.dcs.Member` object of the node.
```

patroni/ha.py
```diff
@@ -20,31 +20,28 @@
...
 
-class _MemberStatus(NamedTuple):
-    """Node status distilled from API response:
+class _MemberStatus(Tags, NamedTuple('_MemberStatus',
+                                     [('member', Member),
+                                      ('reachable', bool),
+                                      ('in_recovery', Optional[bool]),
+                                      ('wal_position', int),
+                                      ('data', Dict[str, Any])])):
+    """Node status distilled from API response.
```
barthisrael, line 35: It seems we are refactoring not only tags, but also `_MemberStatus`.
I guess we should describe this in the PR so it's included in the commit, I mean, that we took advantage of the change to also change this class.

Also, it seems we got rid of `dcs_last_seen`. Was that intentional, or are we missing to add the corresponding property into the class?

barthisrael (approved):
The changes look good to me, nice work!
Left a few comments/suggestions along the way.
I only feel we are missing to replace this, if that makes sense -- I didn't check this piece deeply, had to go AFK :s:

```diff
diff --git a/patroni/postgresql/sync.py b/patroni/postgresql/sync.py
index c8a6ccd..9cff04e 100644
--- a/patroni/postgresql/sync.py
+++ b/patroni/postgresql/sync.py
@@ -209,7 +209,7 @@ class _ReplicaList(List[_Replica]):
             # 2. can be mapped to a ``Member`` of the ``Cluster``:
             #   a. ``Member`` doesn't have ``nosync`` tag set;
             #   b. PostgreSQL on the member is known to be running and accepting client connections.
-            if member and row[sort_col] is not None and member.is_running and not member.tags.get('nosync', False):
+            if member and row[sort_col] is not None and member.is_running and not member.nosync:
                 self.append(_Replica(row['pid'], row['application_name'],
                                      row['sync_state'], row[sort_col], bool(member.nofailover)))
```

patroni/ha.py
CyberDem0n: Refactoring of the `_MemberStatus` was unavoidable. `Tags` class dictates that child classes must implement `tags` property, but such an attribute already existed in the `_MemberStatus`. Renaming it to `_tags` wasn't possible (NamedTuple doesn't allow to have attributes that start with underscore) and therefore I just decided to keep the whole JSON response in the `data` attribute.
`tags`, `timeline`, and `watchdog_failed` attributes are returned from the JSON response as it is, and therefore I converted them to properties.
`dcs_last_seen` attribute is not used anywhere in the code. IIRC it was added to `_MemberStatus` just because it could be useful in the future. In two years it is still not used, therefore I decided to remove it. If we need it one day - adding a property would be just 4 lines of code.

`reachable`, `in_recovery`, and `wal_position` attributes are kept as it is because their values are calculated based on some logic. For example https://github.com/zalando/patroni/pull/2823/files#diff-b6a3a0ef64783a6d32feb908d74a3a009e9f25008b8ca5313c9c9747e3b74a63R54 will raise an exception is there is no `wal` or `xlog` in the JSON document and as a result the `fetch_node_status()` will return `_MemberStatus.unknown(member)`.

patroni/tags.py
```diff
@@ -0,0 +1,60 @@
+"""Tags handling"""
```
hughcapet, line 1:
```suggestion
"""Tags handling."""
```

patroni/tags.py
```diff
@@ -0,0 +1,60 @@
...
+
+        :returns: a dictionary of tags set for this node. The key is the tag name, and the value is the corresponding
+            tag value.
+        """
+        return {tag: value for tag, value in tags.items()
+                if tag not in ('clonefrom', 'nofailover', 'noloadbalance', 'nosync') or value}
+
+    @property
+    @abc.abstractmethod
+    def tags(self) -> Dict[str, Any]:
```
hughcapet, line 34: docstring missing

patroni/ha.py
```diff
@@ -57,21 +54,33 @@ def from_api_response(cls, member: Member, json: Dict[str, Any]) -> '_MemberStat
...
+        return self.data.get('tags', {})
+
+    @property
+    def timeline(self) -> int:
+        """Timeline value from JSON."""
+        return self.data.get('timeline', 0)
+
+    @property
+    def watchdog_failed(self) -> bool:
+        """indicates that watchdog is required by configuration but not available or failed."""
```
matthbakeredb, line 72:
```suggestion
        """Indicates that watchdog is required by configuration but not available or failed."""
```

patroni/ha.py
```diff
@@ -57,21 +54,33 @@ def from_api_response(cls, member: Member, json: Dict[str, Any]) -> '_MemberStat
...
+        return self.data.get('watchdog_failed', False)
 
     @classmethod
     def unknown(cls, member: Member) -> '_MemberStatus':
```
matthbakeredb, line 76:
```suggestion
    def unknown(cls, member: Member) -> '_MemberStatus':
        """Create a new class instance with empty or null values."""
```

matthbakeredb (changes requested):
Nice abstraction.

Couple small doc suggestions.

hughcapet:
👍

CyberDem0n:
:+1:

PR #2824, opened 2023-08-18
Update api docs examples

hughcapet:
* Add failsafe_mode_is_active to /patroni and /metrics
* Add examples showing that failsafe_mode_is_active and cluster_unlocked are only shown for /patroni when the value is "true"
* Update /patroni and /config examples


CyberDem0n:
:+1:

hughcapet:
👍

PR #2827, opened 2023-08-21
Updated HAProxy Go Template Yaml

mich1990:
I used the template but it generates the same config for read/write and read-only replicas. I updated the config to automatically generate leader dbs in primary section and replicas in the replicas section

### DIFF
--- a/extras/confd/templates/haproxy.tmpl
+++ b/extras/confd/templates/haproxy.tmpl
@@ -21,12 +21,12 @@ listen primary
 	option httpchk HEAD /primary
 	http-check expect status 200
 	default-server inter 3s fall 3 rise 2 on-marked-down shutdown-sessions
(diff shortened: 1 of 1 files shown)


CyberDem0n:
It is a clear :-1: from me. It seems you don't understand the idea of HAProxy TCP traffic routing based on health checks.

mich1990:
> It is a clear 👎 from me. It seems you don't understand the idea of HAProxy TCP traffic routing based on health checks.

I understand that during failover the server goes red on HAProxy GUI and it wont be used but why not being specific and updated it asap?

CyberDem0n:
confd updates config files on a schedule (lets say every 10s).
That means, it will take up to 10s to find the new leader and update the config file and on top of that it will take at least 3s (**inter 3s** fall 3 **rise 2**) to make newly added backend healthy.

On top of that you are doing yet another request to DCS.

That's a clear lose-lose situation:
- the number of DCS calls from confd is two times higher than before
- downtime is higher.

PR #2829, opened 2023-08-21
Introduce connection pool

CyberDem0n:
Make it hold connection kwargs for local connections and all `NamedConnection` objects use them automatically.

Also get rid of redundant `ConfigHandler.local_connect_kwargs`.

On top of that we will introduce a dedicated connection for the REST API thread.

### DIFF
--- a/patroni/postgresql/__init__.py
+++ b/patroni/postgresql/__init__.py
@@ -18,7 +18,7 @@ from .bootstrap import Bootstrap
 from .callback_executor import CallbackAction, CallbackExecutor
 from .cancellable import CancellableSubprocess
 from .config import ConfigHandler, mtime
-from .connection import Connection, get_connection_cursor
+from .connection import ConnectionPool, get_connection_cursor
 from .citus import CitusHandler
 from .misc import parse_history, parse_lsn, postgres_major_version_to_int
 from .postmaster import PostmasterProcess
@@ -79,7 +79,8 @@ class Postgresql(object):
         self.set_state('stopped')
 
         self._pending_restart = False
-        self._connection = Connection()
+        self.connection_pool = ConnectionPool()
+        self._connection = self.connection_pool.get('__main__')
         self.citus_handler = CitusHandler(self, config.get('citus'))
         self.config = ConfigHandler(self, config)
         self.config.check_directories()
@@ -277,7 +278,7 @@ class Postgresql(object):
 
         :returns: 'ok' if PostgreSQL is up, 'reject' if starting up, 'no_resopnse' if not up."""
 
-        r = self.config.local_connect_kwargs
+        r = self.connection_pool.conn_kwargs
         cmd = [self.pgcommand('pg_isready'), '-p', r['port'], '-d', self._database]
 
         # Host is not set if we are connecting via default unix socket
@@ -327,10 +328,6 @@ class Postgresql(object):
 
     def connection(self) -> Union['connection3', 'Connection3[Any]']:
         return self._connection.get()
-
-    def set_connection_kwargs(self, kwargs: Dict[str, Any]) -> None:
-        self._connection.set_conn_kwargs(kwargs.copy())
-        self.citus_handler.set_conn_kwargs(kwargs.copy())
 
     def _query(self, sql: str, *params: Any) -> List[Tuple[Any, ...]]:
         """Execute *sql* query with *params* and optionally return results.
@@ -694,7 +691,7 @@ class Postgresql(object):
         # the former node, otherwise, we might get a stalled one
         # after kill -9, which would report incorrect data to
         # patroni.
-        self._connection.close()
+        self.connection_pool.close()
 
         if self.is_running():
             logger.error('Cannot start PostgreSQL because one is already running.')
@@ -765,7 +762,7 @@ class Postgresql(object):
     def checkpoint(self, connect_kwargs: Optional[Dict[str, Any]] = None,
                    timeout: Optional[float] = None) -> Optional[str]:
         check_not_is_in_recovery = connect_kwargs is not None
-        connect_kwargs = connect_kwargs or self.config.local_connect_kwargs
+        connect_kwargs = connect_kwargs or self.connection_pool.conn_kwargs
         for p in ['connect_timeout', 'options']:
             connect_kwargs.pop(p, None)
         if timeout:
--- a/patroni/postgresql/__init__.py
+++ b/patroni/postgresql/__init__.py
@@ -80,7 +80,7 @@ class Postgresql(object):
 
         self._pending_restart = False
         self.connection_pool = ConnectionPool()
-        self._connection = self.connection_pool.get('__main__')
+        self._connection = self.connection_pool.get('heartbeat')
         self.citus_handler = CitusHandler(self, config.get('citus'))
         self.config = ConfigHandler(self, config)
         self.config.check_directories()
--- a/patroni/postgresql/config.py
+++ b/patroni/postgresql/config.py
@@ -942,23 +942,6 @@ class ConfigHandler(object):
                 return 'localhost'  # connection via localhost is preferred
         return listen_addresses[0].strip()  # can't use localhost, take first address from listen_addresses
 
-    @property
-    def local_connect_kwargs(self) -> Dict[str, Any]:
-        ret = self._local_address.copy()
-        # add all of the other connection settings that are available
-        ret.update(self._superuser)
-        # if the "username" parameter is present, it actually needs to be "user"
-        # for connecting to PostgreSQL
-        if 'username' in self._superuser:
-            ret['user'] = self._superuser['username']
-            del ret['username']
-        # ensure certain Patroni configurations are available
-        ret.update({'dbname': self._postgresql.database,
-                    'fallback_application_name': 'Patroni',
-                    'connect_timeout': 3,
-                    'options': '-c statement_timeout=2000'})
-        return ret
-
     def resolve_connection_addresses(self) -> None:
         port = self._server_parameters['port']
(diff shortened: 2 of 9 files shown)


patroni/postgresql/__init__.py
```diff
@@ -277,7 +278,7 @@ def pg_isready(self) -> str:
 
         :returns: 'ok' if PostgreSQL is up, 'reject' if starting up, 'no_resopnse' if not up."""
 
-        r = self.config.local_connect_kwargs
+        r = self.connection_pool.conn_kwargs
         cmd = [self.pgcommand('pg_isready'), '-p', r['port'], '-d', self._database]
```
matthbakeredb, line 282:
```suggestion
        connect_kwargs = self.connection_pool.conn_kwargs
        cmd = [self.pgcommand('pg_isready'), '-p', connect_kwargs['port'], '-d', self._database]
```

matthbakeredb: And the other references of this var in this method

patroni/postgresql/config.py
```diff
@@ -942,23 +942,6 @@ def _get_tcp_local_address(self) -> str:
...
-            ret['user'] = self._superuser['username']
-            del ret['username']
-        # ensure certain Patroni configurations are available
-        ret.update({'dbname': self._postgresql.database,
-                    'fallback_application_name': 'Patroni',
-                    'connect_timeout': 3,
-                    'options': '-c statement_timeout=2000'})
-        return ret
-
     def resolve_connection_addresses(self) -> None:
```
matthbakeredb, line 945: No docstring ;-)

patroni/postgresql/config.py
```diff
@@ -972,12 +955,25 @@ def resolve_connection_addresses(self) -> None:
...
         tcp_local_address = {'host': tcp_local_address, 'port': port}
 
-        self._local_address = unix_local_address if self._config.get('use_unix_socket') else tcp_local_address
         self.local_replication_address = unix_local_address\
             if self._config.get('use_unix_socket_repl') else tcp_local_address
 
         self._postgresql.connection_string = uri('postgres', netloc, self._postgresql.database)
-        self._postgresql.set_connection_kwargs(self.local_connect_kwargs)
+
+        local_address = unix_local_address if self._config.get('use_unix_socket') else tcp_local_address
```
matthbakeredb, line 963: I feel like this needs a comment to explain the intention here. Then that makes me think that this method is doing a number of different steps to build up different data points. There's 2 separate places where data is changed outside of this method. Should there be a single place where this updates? Also do we envisage a possibility of the below failing and if so do the previous data write(s) need rolling back or not be set?

patroni/postgresql/connection.py
```diff
@@ -76,12 +83,72 @@ def query(self, sql: str, *params: Any) -> List[Tuple[Any, ...]]:
...
             raise PostgresConnectionException('connection problems') from exc
 
-    def close(self) -> None:
-        """Close the psycopg connection to postgres."""
+    def close(self, silent: bool = False) -> bool:
+        """Close the psycopg connection to postgres.
+
+        :param silent: whether the method should write logs.
+
+        :returns: ``True`` if ``psycopg`` connection was closed, ``False``, otherwise.``
```
matthbakeredb, line 91:
```suggestion
        :returns: ``True`` if ``psycopg`` connection was closed, ``False`` otherwise.``
```

patroni/postgresql/__init__.py
CyberDem0n: This will be unrelated change, will bloat PR and make it harder to review.

patroni/postgresql/config.py
```diff
@@ -943,6 +943,30 @@ def _get_tcp_local_address(self) -> str:
...
 
     def resolve_connection_addresses(self) -> None:
+        """Calculates and sets local and remote connection urls and options.
+
+        This method sets:
+            * ``Postgresql.connection_string`` attribute, which is later written to the member key in DCS
+              as ``conn_url``.
+            * ``ConfigHandler.local_replication_address`` attribute, is used for replication connections
+              to local postgres.
+            * ``ConnectionPool.conn_kwargs`` attribute, is used for superuser connections to local postgres.
```
matthbakeredb, line 953:
```suggestion
            * :attr:``Postgresql.connection_string`` attribute, which is later written to the member key in DCS
              as ``conn_url``.
            * :attr:``ConfigHandler.local_replication_address`` attribute, is used for replication connections
              to local postgres.
            * :attr:``ConnectionPool.conn_kwargs`` attribute, is used for superuser connections to local postgres.
```

patroni/postgresql/config.py
```diff
@@ -943,6 +943,30 @@ def _get_tcp_local_address(self) -> str:
...
+
+        .. note::
+            If there is a valid directory in ``postgresql.parameters.unix_socket_directories``  in the Patroni
```
matthbakeredb, line 956:
```suggestion
            If there is a valid directory in ``postgresql.parameters.unix_socket_directories`` in the Patroni
```

patroni/postgresql/config.py
```diff
@@ -943,6 +943,30 @@ def _get_tcp_local_address(self) -> str:
...
+            configuration and there are ``postgresql.use_unix_socket`` and/or ``postgresql.use_unix_socket_repl``
```
matthbakeredb, line 957:
```suggestion
            configuration and ``postgresql.use_unix_socket`` and/or ``postgresql.use_unix_socket_repl``
```

patroni/postgresql/config.py
```diff
@@ -943,6 +943,30 @@ def _get_tcp_local_address(self) -> str:
...
+            are set to ``True``, we respectively use unix sockets for superuser and replication connections
+            to local postgres.
+
+            If there is a requirement to use unix sockets, but nothing is set in the
+            ``postgresql.parameters.unix_socket_directories`` we omit a ``host`` in connection parameters relying
```
matthbakeredb, line 962:
```suggestion
            ``postgresql.parameters.unix_socket_directories``, we omit a ``host`` in connection parameters relying
```

patroni/postgresql/config.py
```diff
@@ -943,6 +943,30 @@ def _get_tcp_local_address(self) -> str:
...
+            on the ability of ``libpq`` to connect via some default unix socket directory.
+
+            If unix sockets are not requested we "switch" to TCP, but prefer to use ``localhost`` if it is possible
+            to deduct that Postgres is listening on it.
```
matthbakeredb, line 966:
```suggestion
            If unix sockets are not requested we "switch" to TCP, prefering to use ``localhost`` if it is possible
            to deduce that Postgres is listening on a local interface address.
```

matthbakeredb (changes requested):
Thanks for adding the doc string. Added some suggestions to that

matthbakeredb (approved):
lgtm!

patroni/postgresql/config.py
```diff
@@ -942,24 +942,31 @@ def _get_tcp_local_address(self) -> str:
...
-
     def resolve_connection_addresses(self) -> None:
+        """Calculates and sets local and remote connection urls and options.
+
+        This method sets:
+            * :attr:``Postgresql.connection_string`` attribute, which is later written to the member key in DCS
+              as ``conn_url``.
+            * :attr:``ConfigHandler.local_replication_address`` attribute, is used for replication connections
+              to local postgres.
+            * :attr:``ConnectionPool.conn_kwargs`` attribute, is used for superuser connections to local postgres.
```
barthisrael, line 953:
```suggestion
            * :attr:`Postgresql.connection_string <patroni.postgresql.Postgresql.connection_string>` attribute, which
              is later written to the member key in DCS as ``conn_url``.
            * :attr:`ConfigHandler.local_replication_address` attribute, which is used for replication connections to
              local postgres.
            * :attr:`ConnectionPool.conn_kwargs <patroni.postgresql.connection.ConnectionPool.conn_kwargs>` attribute,
              which is used for superuser connections to local postgres.
```

We need the full path to create links to things from other modules. In order to keep the alias, I suggested the syntax:

```
:attr:`ALIAS <PATH_TO_LINK>`
```

patroni/postgresql/connection.py
```diff
@@ -76,12 +83,72 @@ def query(self, sql: str, *params: Any) -> List[Tuple[Any, ...]]:
...
+
+
+class ConnectionPool:
+    """Helper class to manage named connections from Patroni to PostgreSQL.
+
+    The instance keeps named :class:`NamedConnection` objects and parameters that must be used for new connections.
+    """
+
+    def __init__(self) -> None:
+        """Create and instance of :class:`ConnectionPool` class."""
```
barthisrael, line 110:
```suggestion
        """Create an instance of :class:`ConnectionPool` class."""
```

patroni/postgresql/connection.py
```diff
@@ -76,12 +83,72 @@ def query(self, sql: str, *params: Any) -> List[Tuple[Any, ...]]:
...
+
+    def get(self, name: str, kwargs_override: Optional[Dict[str, Any]] = None) -> NamedConnection:
+        """Get a new named :class:`NamedConnection` object from the pool.
+
+        .. note::
+            Creates a new :class:`NamedConnection` object if it doesn't yet exist in the pool.
+
+        :param name: name of the connection.
+        :param kwargs_override: :class:`dict` object with connection parameters that should be
+                                different from default values provided by ``conn_kwargs``.
```
barthisrael, line 138:
```suggestion
                                different from default values provided by :attr:`conn_kwargs`.
```

So we link to the property.

barthisrael (approved):
PR looks good.

Posted a few suggestions for better API docs.

CyberDem0n:
:+1:

PR #2831, opened 2023-08-24
Generate documentation of private members through sphinx docs

barthisrael:
With this commit we make sphinx build API docs for the following things, which were missing up to this point:

* `__init__` method of classes;
* "private" members (properties, functions, methods, attributes, etc., which name starts with an underscore);
* members that are missing a docstring, so we can still reference them with links in the documentation.

The third point can be removed later, if we wish, when we reach a point where everything has proper docstrings in the Patroni code base.

References: PAT-195.


hughcapet:
what's with the Tests / docs step failing here?

matthbakeredb:
> what's with the Tests / docs step failing here?

I expect sphinx wasn't parsing the private methods at all before so didn't detect formatting problems. I'll take a look as Israel is on leave this week.

patroni/config.py
```diff
@@ -257,7 +257,7 @@ def __init__(self, configfile: str,
 
               * file or directory path passed as command-line argument (*configfile*), if it exists and the file or
                 files found in the directory can be parsed (see :meth:`~Config._load_config_path`), otherwise
-              * YAML file passed via the environment variable (see :cvar:`PATRONI_CONFIG_VARIABLE`), if the referenced
+              * YAML file passed via the environment variable (see :attr:`PATRONI_CONFIG_VARIABLE`), if the referenced
```
CyberDem0n, line 260: Hmm, in the class docstring it is declared as `:cvar:`. Shouldn't it be here the same?

patroni/postgresql/validator.py
```diff
@@ -280,17 +280,18 @@ def _read_postgres_gucs_validators_file(file: str) -> Dict[str, Any]:
...
     """Load all Postgres GUC validators from YAML files.
 
+    Load validators from YAML files that were found.
+
     Recursively walk through ``available_parameters`` directory and load validators of each found YAML file into
     ``parameters`` and/or ``recovery_parameters`` variables.
 
     Walk through directories in top-down fashion and for each of them:
         * Sort files by name;
-        * Load validators from YAML files that were found.
```
CyberDem0n, line 288: Is it an expected change?

patroni/config.py
matthbakeredb: Apparently, it's not a valid "domain" according to sphinx command output. Docs are here for what can be used
https://www.sphinx-doc.org/en/master/usage/restructuredtext/domains.html#cross-referencing-python-objects

patroni/postgresql/validator.py
matthbakeredb: Hmmn yes that is odd. Let me revert it and see if there were errors.

matthbakeredb: Yep, unrelated. Reverted.

CyberDem0n:
:+1:

hughcapet:
👍

PR #2832, opened 2023-08-24
Fix bug with kubernetes.standby_leader_label_value

CyberDem0n:
When running with the leader lock Patroni was just setting the `role` label to `master` and effectively `kubernetes.standby_leader_label_value` feature never worked.

Now it is fixed, but in order to not introduce breaking changes we just update default value of the `standby_leader_label_value` to the `master`.


matthbakeredb (approved):
Would the breaking change be to not use the word `master`? Is that something that should be considered for a future change?

barthisrael (approved):

hughcapet:
👍

CyberDem0n:
@matthbakeredb, yes, if we currently use something different from master it would be a breaking change.

CyberDem0n:
:+1:

PR #2833, opened 2023-08-24
Introduce a dedicated postgres connection for REST API

CyberDem0n:
Sharing a single connection between REST API and the main thread (doing heartbeats) was working mostly fine, except when Postgres becomes so slow that REST API queries start blocking the main loop.

If the dedicated REST API connection isn't available we use the heartbeat connection as a fallback.

### DIFF
--- a/patroni/api.py
+++ b/patroni/api.py
@@ -1380,7 +1380,7 @@ class RestApiServer(ThreadingMixIn, HTTPServer, Thread):
             :class:`psycopg.Error`: if had issues while executing *sql*.
             :class:`~patroni.exceptions.PostgresConnectionException`: if had issues while connecting to the database.
         """
-        # We first try top get a heartbeat connection because it is always required for the main thread
+        # We first try to get a heartbeat connection because it is always required for the main thread.
         try:
             heartbeat_connection = self.patroni.postgresql.connection_pool.get('heartbeat')
             heartbeat_connection.get()  # try to open psycopg connection to postgres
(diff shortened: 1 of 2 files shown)


patroni/api.py
```diff
@@ -1377,10 +1380,21 @@ def query(self, sql: str, *params: Any) -> List[Tuple[Any, ...]]:
             :class:`psycopg.Error`: if had issues while executing *sql*.
             :class:`~patroni.exceptions.PostgresConnectionException`: if had issues while connecting to the database.
         """
+        # We first try top get a heartbeat connection because it is always required for the main thread
```
matthbakeredb, line 1383:
```suggestion
        # We first try to get a heartbeat connection because it is always required for the main thread.
```

matthbakeredb (changes requested):
Looks good. One tiny correction.

barthisrael (approved):

tests/test_api.py
```diff
@@ -659,3 +677,9 @@ def test_reload_local_certificate(self):
 
     def test_get_certificate_serial_number(self):
         self.assertIsNone(self.srv.get_certificate_serial_number())
+
+    def test_query(self):
+        with patch.object(MockConnection, 'get', Mock(side_effect=OperationalError)):
+            self.assertRaises(PostgresConnectionException, self.srv.query, 'SELECT 1')
+        with patch.object(MockConnection, 'get', Mock(side_effect=[MockConnect(), OperationalError])):
+            self.srv.query('SELECT 1')
```
hughcapet, line 685: Maybe something like 
```
self.assertEqual(self.srv.query('SELECT 1'), MockConnection.query('SELECT 1'))
```
or
```
        with patch.object(MockConnection, 'get', Mock(side_effect=[MockConnect(), OperationalError])), \
             patch.object(MockConnection, 'query') as mock_query:
            self.srv.query('SELECT 1')
            mock_query.assert_called_once_with('SELECT 1')
```
?

hughcapet:
👍

CyberDem0n:
:+1:

Issue #2837, opened 2023-08-30
Patroni crashes during shutdown in Kubernetes environment - crashing PostgreSQL as well


feikesteenbergen:
### What happened?

During the shutdown of a Pod, Patroni crashes:

```
  File "/usr/lib/python3/dist-packages/patroni/dcs/kubernetes.py", line 1127, in _write_status
    raise NotImplementedError  # pragma: no cover
NotImplementedError

```

### How can we reproduce it (as minimally and precisely as possible)?

Shutdown a pod while using Kubernetes as a DCS

### What did you expect to happen?

Patroni waiting for PostgreSQL to shutdown and then to shutdown itself

### Patroni/PostgreSQL/DCS version

-   Patroni version: 3.0.4
-   PostgreSQL version: 15.4
-   DCS (and its version):  Kubernetes v1.25.10

### Patroni configuration file

```yaml
NA
```

### patronictl show-config

```yaml
NA
```

### Patroni log files

```shell
2023-08-30 13:23:45,586 WARNING: manual failover: members list is empty
2023-08-30 13:23:45,587 ERROR: Exception during Ha.shutdown
Traceback (most recent call last):
  File "/usr/lib/python3/dist-packages/patroni/daemon.py", line 174, in abstract_main
    controller.run()
  File "/usr/lib/python3/dist-packages/patroni/__main__.py", line 133, in run
    super(Patroni, self).run()
  File "/usr/lib/python3/dist-packages/patroni/daemon.py", line 143, in run
    self._run_cycle()
  File "/usr/lib/python3/dist-packages/patroni/__main__.py", line 145, in _run_cycle
    self.schedule_next_run()
  File "/usr/lib/python3/dist-packages/patroni/__main__.py", line 127, in schedule_next_run
    elif self.ha.watch(nap_time):
  File "/usr/lib/python3/dist-packages/patroni/ha.py", line 1885, in watch
    return self.dcs.watch(leader_version, timeout)
  File "/usr/lib/python3/dist-packages/patroni/dcs/kubernetes.py", line 1357, in watch
    return super(Kubernetes, self).watch(None, timeout)
  File "/usr/lib/python3/dist-packages/patroni/dcs/__init__.py", line 1134, in watch
    self.event.wait(timeout)
  File "/usr/lib/python3.10/threading.py", line 607, in wait
    signaled = self._cond.wait(timeout)
  File "/usr/lib/python3.10/threading.py", line 324, in wait
[...]

PR #2838, opened 2023-08-30
Troubleshooting: Make _write_status a noop

feikesteenbergen:
Tentative troubleshooting fix for https://github.com/zalando/patroni/issues/2837


CyberDem0n (commented):
Oh... Good catch.
Actually at this moment Postgres is already cleanly shut down (according to pg_controldata), so there is no PG crash involved.
But, this method on K8s is supposed to update the `optime` annotation on the leader endpoint/configmap. Making it noop isn't the right thing.

PR #2840, opened 2023-08-30
Return system id to the ctl list title

### LINKED ISSUE #2839: patronictl list doesn't show the cluster system identifier anymore
### What happened?

`patronictl list` doesn't show the cluster's Postgres System Identifier anymore.

```
$ python3 ./patronictl.py -c postgres0.yml list
+ Cluster: batman --------+--------+---------+----+-----------+
| Member      | Host      | Role   | State   | TL | Lag in MB |
+-------------+-----------+--------+---------+----+-----------+
| postgresql0 | 127.0.0.1 | Leader | running |  1 |           |
+-------------+-----------+--------+---------+----+-----------+

$ python3 ./patronictl.py -c postgres0.yml version
patronictl version 3.1.0
```

```
 patronictl version
patronictl version 2.1.4

$ patronictl -c postgres0.yml list
+-------------+-----------+--------+---------+----+-----------+
| Member      | Host      | Role   | State   | TL | Lag in MB |
+ Cluster: batman (7273083582773762774) -----+----+-----------+
| postgresql0 | 127.0.0.1 | Leader | running |  1 |           |
+-------------+-----------+--------+---------+----+-----------+
```

Note `7273083582773762774`, it's the cluster system identifier

-   Patroni version: 

First noticed on v3.04, can't reproduce on 2.1.4

-   PostgreSQL version: 

Doesn't matter

-   DCS (and its version): 

Doesn't matter

```yaml
Doesn't apply.
```

### patronictl show-config

```yaml
Doesn't apply.
```

### Patroni log files

```shell
Doesn't apply.
```

### PostgreSQL log files

```shell
Doesn't apply.
```

- [X] Yes

_No response_

hughcapet:
Since https://github.com/zalando/patroni/commit/4872ac51e00abefc98678f1771dd93aead59f4ce `patronictl list` was only showing system id for a Citus cluster with the `--group` option specified

Close #2839

### DIFF
--- a/tests/test_ctl.py
+++ b/tests/test_ctl.py
@@ -406,10 +406,14 @@ class TestCtl(unittest.TestCase):
         result = self.runner.invoke(ctl, ['list'])
         assert '127.0.0.1' in result.output
         assert result.exit_code == 0
-        assert 'Citus cluster: alpha (12345678901)' in result.output
+        assert 'Citus cluster: alpha -' in result.output
 
         result = self.runner.invoke(ctl, ['list', '--group', '0'])
-        assert 'Citus cluster: alpha (group: 0, 12345678901)' in result.output
+        assert 'Citus cluster: alpha (group: 0, 12345678901) -' in result.output
+
+        with patch('patroni.ctl.load_config', Mock(return_value={'scope': 'alpha'})):
+            result = self.runner.invoke(ctl, ['list'])
+            assert 'Cluster: alpha (12345678901) -' in result.output
 
         with patch('patroni.ctl.load_config', Mock(return_value={})):
             self.runner.invoke(ctl, ['list'])
--- a/tests/test_ctl.py
+++ b/tests/test_ctl.py
@@ -402,9 +402,15 @@ class TestCtl(unittest.TestCase):
     @patch('patroni.ctl.get_dcs')
     def test_members(self, mock_get_dcs):
         mock_get_dcs.return_value.get_cluster = get_cluster_initialized_with_leader
+
         result = self.runner.invoke(ctl, ['list'])
         assert '127.0.0.1' in result.output
         assert result.exit_code == 0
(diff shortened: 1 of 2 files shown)


tests/test_ctl.py
```diff
@@ -402,9 +402,15 @@ def test_get_all_members(self):
     @patch('patroni.ctl.get_dcs')
     def test_members(self, mock_get_dcs):
         mock_get_dcs.return_value.get_cluster = get_cluster_initialized_with_leader
+
         result = self.runner.invoke(ctl, ['list'])
         assert '127.0.0.1' in result.output
         assert result.exit_code == 0
+        assert 'Citus cluster: alpha (12345678901)' in result.output
```
CyberDem0n, line 409: It shows a system id for coordinator cluster if patronictl is executed without `--group` argument.
This is not the behavior we want, because users might get a false impression that all nodes in the citus cluster have the same sysid, while in fact it is different for different groups.

hughcapet: 🤦‍♀️ sure

tests/test_ctl.py
```diff
@@ -402,9 +402,19 @@ def test_get_all_members(self):
     @patch('patroni.ctl.get_dcs')
     def test_members(self, mock_get_dcs):
         mock_get_dcs.return_value.get_cluster = get_cluster_initialized_with_leader
+
         result = self.runner.invoke(ctl, ['list'])
         assert '127.0.0.1' in result.output
         assert result.exit_code == 0
+        assert 'Citus cluster: alpha -' in result.output
```
CyberDem0n, line 409: Citus case was actually working fine. What was broken - classic Patroni clusters.
Could we add a test for it?

hughcapet: but I have added a test for a classic Patroni setup...

CyberDem0n:
:+1:

hughcapet:
👍

PR #2842, opened 2023-08-31
Disable WAL streaming on standby node via new boolean tag "nostream"

gsmolk:
Hello!
I would like to propose the support of standby with explicitly disabled WAL streaming via tag 'nostream' (so node obtain WAL only via `restore_command` and `pg_wal` polling).
It works already really well in our production system, consisting of several hundred patroni clusters.
It allowed us to move all the heavy OLAP-stuff on "nostream" nodes, so they do not affect OLTP and vacuum on master (due to `hot_standby_feedback`).
I would love to receive any feedback on the proposed patch (which by no means is commit-ready), or whether you consider the work on this one worth pursuing at all.


patroni/__main__.py
```diff
@@ -155,6 +155,10 @@ def reload_config(self, sighup: bool = False, local: Optional[bool] = False) ->
         except Exception:
             logger.exception('Failed to reload config_file=%s', self.config.config_file)
 
+    @property
```
hughcapet, line 158: without going deep into the whole idea and the implementation - we have a dedicated abstract class for tags handling (see `tags.py`)

gsmol: @hughcapet, thank you, I was rebasing this patch from 2.1.7 version (the one we use in production) to current master, and some mistakes must have slipped through the cracks.
I will fix it.

gsmol: fixed, thank you for a feedback

patroni/postgresql/__init__.py
```diff
@@ -1067,7 +1067,7 @@ def get_history(self, timeline: int) -> List[Union[Tuple[int, int, str], Tuple[i
         return history
 
     def follow(self, member: Union[Leader, Member, None], role: str = 'replica',
-               timeout: Optional[float] = None, do_reload: bool = False) -> Optional[bool]:
+               is_archive: bool = False, timeout: Optional[float] = None, do_reload: bool = False) -> Optional[bool]:
```
hughcapet, line 1070: what is this new param here for?

hughcapet:
Please also watch out flake8 issues and the already failing tests

gsmol:
> Most of the changes (may be all of them) in patroni/postgresql/*.py seems to be unnecessary.
They would behave exactly as you want (postgres will not stream from anywhere but use only restore_command) if you just pass member=None.
The only thing that needs to be changed - [Ha._get_node_to_follow()](https://github.com/zalando/patroni/blob/6b7f914da7fd4974ca091c8f55c769b2ad0f58a3/patroni/ha.py#L566C16-L594) method. It simply should return None if this is the archive-only node

Neat, I will look into it.

> Another question - how this tag should be called. The archive doesn't seem to be a right name for it. archiveonly, archive-only, or archive_only better describe the intention

Considering that it is entirely possible for user not to have archive and WAL files can just be shoveled into pg_wal by some external machinery (and Postgres documentation totally accepts the validity of such use-case), _maybe_ the correct naming should be `nostream`.

> And the last one - should such nodes even be allowed to become the primary? Or used as cascading nodes

I think, it should, and that if user don't is not ok with that - he can use `nofailover`\`nosync` flags.

> Please also watch out flake8 issues and the already failing tests

Sure

gsmolk:
@CyberDem0n, @hughcapet, Hello!
I`m sorry for such a long delay with this feature.
I`ve refactored the PR (additional thanks for refactoring around `Cluster`), added test and documentation, and renamed tag to `nostream`.
Any feedback will be much appreciated, thank you in advance.

CyberDem0n (commented):
@gsmolk your changes look good.
But, there is one thing forgotten: we don't want to "create" logical replication slots on "nostream" nodes, and also all nodes that replicating from it, directly or indirectly.
Example:
```bash
$ patronictl topology
+ Cluster: batman (7333181567713260675) -----+---------------------+----+-----------+------------------------------+
| Member          | Host           | Role    | State               | TL | Lag in MB | Tags                         |
+-----------------+----------------+---------+---------------------+----+-----------+------------------------------+
| postgresql0     | 127.0.0.1:5432 | Leader  | running             | 40 |           |                              |
| + postgresql1   | 127.0.0.1:5433 | Replica | in archive recovery | 40 |         0 | {nostream: true}             |
|   + postgresql2 | 127.0.0.1:5434 | Replica | streaming           | 40 |         0 | {replicatefrom: postgresql1} |
+-----------------+----------------+---------+---------------------+----+-----------+------------------------------+
```

```yaml
$ patronictl show-config
loop_wait: 10
postgresql:
  parameters:
    archive_command: mkdir -p ../wal_archive && test ! -f ../wal_archive/%f && cp %p ../wal_archive/%f
    archive_mode: 'on'
    archive_timeout: 1800s
    restore_command: cp ../wal_archive/%f %p
    wal_level: logical
  use_pg_rewind: true
retry_timeout: 10
slots:
  foo:
    database: postgres
    plugin: test_decoding
    type: logical
ttl: 30
```

With such configuration the logical slot `foo` should be "created" only on `postgresql0`.
Having logical failover slots on postgresql1 and postgresql2 is technically possible, but it will be unsafe to use them after failover, because nodes can't give a feedback to the primary about `catalog_xmin` and the primary can clean up tuples in pg_catalog that are required for logical decoding to work.

Here are the changes that should solve it:
```diff
diff --git a/patroni/dcs/__init__.py b/patroni/dcs/__init__.py
index f9da6dd7..c26780ad 100644
--- a/patroni/dcs/__init__.py
+++ b/patroni/dcs/__init__.py
@@ -1054,7 +1054,7 @@ class Cluster(NamedTuple('Cluster',
         if not global_config.use_slots or tags.nofailover:
             return {}
 
-        if global_config.is_standby_cluster:
+        if global_config.is_standby_cluster or self.get_slot_name_on_primary(postgresql.name, tags) is None:
             return self.__permanent_physical_slots \
                 if postgresql.major_version >= SLOT_ADVANCE_AVAILABLE_VERSION or role == 'standby_leader' else {}
 
@@ -1177,7 +1177,7 @@ class Cluster(NamedTuple('Cluster',
             return any(self.should_enforce_hot_standby_feedback(postgresql, m) for m in members)
         return False
 
-    def get_slot_name_on_primary(self, name: str, tags: Tags) -> str:
+    def get_slot_name_on_primary(self, name: str, tags: Tags) -> Optional[str]:
         """Get the name of physical replication slot for this node on the primary.
 
         .. note::
@@ -1191,6 +1191,8 @@ class Cluster(NamedTuple('Cluster',
 
         :returns: the slot name on the primary that is in use for physical replication on this node.
         """
+        if tags.nostream:
+            return None
         replicatefrom = self.get_member(tags.replicatefrom, False) if tags.replicatefrom else None
         return self.get_slot_name_on_primary(replicatefrom.name, replicatefrom) \
             if isinstance(replicatefrom, Member) else slot_name_from_member_name(name)
```

Please test them, and maybe add some comments to the code and update docstrings.


I am also thinking that nodes in a chain under `nostream` should be less favorable during the leader race in case if there are permanent logical slots defined in the configuration, but for now it could probably be solved by `failover_priority` tags.

gsmolk:
> Having logical failover slots on postgresql1 and postgresql2 is technically possible, but it will be unsafe to use them after failover, because nodes can't give a feedback to the primary about catalog_xmin and the primary can clean up tuples in pg_catalog that are required for logical decoding to work.

I've missed that nuisance, thank you. I'm working on it.

gsmolk:
@CyberDem0n, Hello!
I`ve added the your patch and tested it, improved documentation a bit and added new test for around slot-handling.

gsmolk:
@CyberDem0n, hello and sorry for bothering.
Is there anything else that I can do for this PR is to be considered for merging?

patroni/tags.py
```diff
@@ -89,3 +89,8 @@ def nosync(self) -> bool:
     def replicatefrom(self) -> Optional[str]:
         """Value of ``replicatefrom`` tag, if any."""
         return self.tags.get('replicatefrom')
+
+    @property
+    def nostream(self) -> bool:
+        """``True`` if ``nostream`` is ``True``, else ``False``."""
+        return bool(self.tags.get('nostream', False))
```
CyberDem0n, line 96: I guess it was copied from `nosync` or similar, but behavior isn't really nice, because it treats any non-empty strings like `True`.
We probably better to use something like:
```suggestion
        return parse_bool(self.tags.get('nostream')) or False
```

The reason we didn't change other methods - just to be backward compatible in case if someone has configuration file in a weird format. But for the new methods we can handle it nicely.

P.S. validator.py is OK, it could be stricter.

tests/test_slots.py
```diff
@@ -124,6 +124,68 @@ def test_process_permanent_slots(self):
...
+            'tags': {'replicatefrom': 'test-2'}
+        })
+        stream_node = Member(0, 'test-4', 28, {
+            'state': 'running', 'conn_url': 'postgres://replicator:rep-pass@127.0.0.1:5436/postgres'})
+        cluster = Cluster(
+            True, config, self.leader, Status.empty(),
+            [self.leadermem, nostream_node, cascade_node, stream_node], None, SyncState.empty(), None, None)
+        global_config.update(cluster)
+
+        # sanity for master
```
CyberDem0n, line 145:
```suggestion
        # sanity for primary
```

patroni/tags.py
gsmolk: > The reason we didn't change other methods - just to be backward compatible in case if someone has configuration file in a weird format. But for the new methods we can handle it nicely.

oh, thats explains it

tests/test_slots.py
gsmolk: done

patroni/tags.py
gsmolk: done

patroni/dcs/__init__.py
```diff
@@ -1087,11 +1093,13 @@ def _get_members_slots(self, name: str, role: str) -> Dict[str, Dict[str, str]]:
         members = filter(lambda m: m.name != name, self.members)
```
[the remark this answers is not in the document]
gsmolk: done

patroni/dcs/__init__.py
```diff
@@ -1087,11 +1093,13 @@ def _get_members_slots(self, name: str, role: str) -> Dict[str, Dict[str, str]]:
...
-            members = [m for m in members if m.replicatefrom is None
-                       or m.replicatefrom == name or not self.has_member(m.replicatefrom)]
+            members = [m for m in members if (m.replicatefrom is None
+                       or m.replicatefrom == name or not self.has_member(m.replicatefrom))
+                       and not m.nostream]
         else:
             # only manage slots for replicas that replicate from this one, except for the leader among them
-            members = [m for m in members if m.replicatefrom == name and m.name != self.leader_name]
+            members = [m for m in members if (m.replicatefrom == name
+                                              and m.name != self.leader_name and not m.nostream)]
```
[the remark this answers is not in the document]
gsmolk: oh, thank you, that was sloppy

gsmolk:
@CyberDem0n, done!
As always, any feedback will be much appreciated.

hughcapet:
@gsmolk can you please adjust PR title and description to match the actual implementation? (e.g. the new tag name)

patroni/tags.py
```diff
@@ -3,13 +3,15 @@
...
 
-from patroni.utils import parse_int
+from patroni.utils import parse_int, parse_bool
 
 
 class Tags(abc.ABC):
     """An abstract class that encapsulates all the ``tags`` logic.
 
     Child classes that want to use provided facilities must implement ``tags`` abstract property.
+    .. note::
```
hughcapet, line 13:
```suggestion

    .. note::
```

gsmolk: thank you, fixed

gsmolk:
@hughcapet, sure thing, done.

docs/yaml_configuration.rst
```diff
@@ -399,6 +399,7 @@ Tags
...
 -  **nofailover**: ``true`` or ``false``, controls whether this node is allowed to participate in the leader race and become a leader. Defaults to ``false``, meaning this node _can_ participate in leader races. 
 -  **failover_priority**: integer, controls the priority that this node should have during failover. Nodes with higher priority will be preferred over lower priority nodes if they received/replayed the same amount of WAL. However, nodes with higher values of receive/replay LSN are preferred regardless of their priority. If the ``failover_priority`` is 0 or negative - such node is not allowed to participate in the leader race and to become a leader (similar to ``nofailover: true``).
+-  **nostream**: ``true`` or ``false``. If set to ``true`` the node will not use replication protocol to stream WAL. It will rely instead on archive recovery (if ``restore_command`` is configured) and ``pg_wal``/``pg_xlog`` polling. It also disables copying and synchronization of permanent replication slots on the node itself and all its cascading replicas. Setting this tag on primary node has no effect.
```
hughcapet, line 402:
```suggestion
-  **nostream**: ``true`` or ``false``. If set to ``true`` the node will not use replication protocol to stream WAL. It will rely instead on archive recovery (if ``restore_command`` is configured) and ``pg_wal``/``pg_xlog`` polling. It also disables copying and synchronization of permanent logical replication slots on the node itself and all its cascading replicas. Setting this tag on primary node has no effect.
```

hughcapet:
👍

CyberDem0n:
:+1:

hughcapet:
Merged.
Thank you @gsmolk for the contribution!

gsmolk:
@hughcapet, @CyberDem0n, thank you very much for review and help with this PR!

PR #2843, opened 2023-08-31
Document how to change Postgres settings that touch shared memory

barthisrael:
Some special handling is required when changing either of these settings in a Postgres cluster that has standby nodes:

* `max_connections`
* `max_prepared_transactions`
* `max_locks_per_transaction`
* `max_wal_senders`
* `max_worker_processes`

If one attempts to decrease `max_connections` dynamic setting and restart all nodes at the same time (primary and standbys), Patroni will refuse to apply the new value on the standbys and require the user to restart it again later, once replication catches up.

That behavior is correct, but it is not documented.

This commit adds information to documentation about that behavior and why it's required.

References: PAT-166.

### DIFF
--- a/docs/patroni_configuration.rst
+++ b/docs/patroni_configuration.rst
@@ -90,6 +90,40 @@ The parameters would be applied in the following order (run-time are given the h
 This allows configuration for all the nodes (2), configuration for a specific node using ``ALTER SYSTEM`` (3) and ensures that parameters essential to the running of Patroni are enforced (4), as well as leaves room for configuration tools that manage `postgresql.conf` directly without involving Patroni (1).
 
 
+PostgreSQL parameters that touch shared memory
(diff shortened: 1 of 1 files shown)


CyberDem0n (commented):
I like these suggestions, but I think we somehow need to emphasize that we are talking about Postgres restart, because very often I observed that people don't use `patronictl restart`, but just "restart" Patroni on the primary, which results in a failover.

barthisrael:
> I like these suggestions, but I think we somehow need to emphasize that we are talking about Postgres restart, because very often I observed that people don't use `patronictl restart`, but just "restart" Patroni on the primary, which results in a failover.

Makes sense! Just submitted a commit with your suggestion.

hughcapet:
👍

CyberDem0n:
👍

PR #2846, opened 2023-09-03
Enreach some endpoints with the scope and name

sskserk:
Hi Patroni Guys,

here is the PR for an enhancement mentioned in the thread https://app.slack.com/client/T0FS7GCKS/C9XPYG92A/thread/C9XPYG92A-1687875160.874589

The feature significantly simplifies automatic discovery of the patroni nodes. 

We have developed a product that discovers/discovers/configures the Patroni clusters. Also, it deeply monitors insights of the posgres and the patroni itself. Hopefully together we can  make the Postgres community bigger.

### DIFF
--- a/docs/rest_api.rst
+++ b/docs/rest_api.rst
@@ -364,7 +364,6 @@ Cluster status endpoints
     {
       "members": [
         {
-          "scope": "demo",
           "name": "patroni1",
           "role": "leader",
           "state": "running",
@@ -377,7 +376,6 @@ Cluster status endpoints
           }
         },
         {
-          "scope": "demo",
           "name": "patroni2",
           "role": "replica",
           "state": "streaming",
@@ -391,6 +389,7 @@ Cluster status endpoints
           "lag": 0
         }
       ],
+      "scope": "demo",
       "scheduled_switchover": {
         "at": "2023-09-24T10:36:00+02:00",
         "from": "patroni1",
--- a/docs/rest_api.rst
+++ b/docs/rest_api.rst
@@ -283,70 +283,70 @@ Retrieve the Patroni metrics in Prometheus format through the ``GET /metrics`` e
 	
 	# HELP patroni_version Patroni semver without periods. \
 	# TYPE patroni_version gauge
-	patroni_version{scope="batman"} 020103
+	patroni_version{scope="batman", name="patroni1"} 020103
 	# HELP patroni_postgres_running Value is 1 if Postgres is running, 0 otherwise.
 	# TYPE patroni_postgres_running gauge
-	patroni_postgres_running{scope="batman"} 1
+	patroni_postgres_running{scope="batman", name="patroni1"} 1
 	# HELP patroni_postmaster_start_time Epoch seconds since Postgres started.
 	# TYPE patroni_postmaster_start_time gauge
-	patroni_postmaster_start_time{scope="batman"} 1657656955.179243
+	patroni_postmaster_start_time{scope="batman", name="patroni1"} 1657656955.179243
 	# HELP patroni_master Value is 1 if this node is the leader, 0 otherwise.
 	# TYPE patroni_master gauge
-	patroni_master{scope="batman"} 1
+	patroni_master{scope="batman", name="patroni1"} 1
 	# HELP patroni_primary Value is 1 if this node is the leader, 0 otherwise.
 	# TYPE patroni_primary gauge
-	patroni_primary{scope="batman"} 1
+	patroni_primary{scope="batman", name="patroni1"} 1
 	# HELP patroni_xlog_location Current location of the Postgres transaction log, 0 if this node is not the leader.
 	# TYPE patroni_xlog_location counter
-	patroni_xlog_location{scope="batman"} 22320573386952
+	patroni_xlog_location{scope="batman", name="patroni1"} 22320573386952
 	# HELP patroni_standby_leader Value is 1 if this node is the standby_leader, 0 otherwise.
 	# TYPE patroni_standby_leader gauge
-	patroni_standby_leader{scope="batman"} 0
+	patroni_standby_leader{scope="batman", name="patroni1"} 0
 	# HELP patroni_replica Value is 1 if this node is a replica, 0 otherwise.
 	# TYPE patroni_replica gauge
-	patroni_replica{scope="batman"} 0
+	patroni_replica{scope="batman", name="patroni1"} 0
 	# HELP patroni_sync_standby Value is 1 if this node is a sync standby replica, 0 otherwise.
 	# TYPE patroni_sync_standby gauge
-	patroni_sync_standby{scope="batman"} 0
+	patroni_sync_standby{scope="batman", name="patroni1"} 0
 	# HELP patroni_xlog_received_location Current location of the received Postgres transaction log, 0 if this node is not a replica.
 	# TYPE patroni_xlog_received_location counter
-	patroni_xlog_received_location{scope="batman"} 0
+	patroni_xlog_received_location{scope="batman", name="patroni1"} 0
 	# HELP patroni_xlog_replayed_location Current location of the replayed Postgres transaction log, 0 if this node is not a replica.
 	# TYPE patroni_xlog_replayed_location counter
-	patroni_xlog_replayed_location{scope="batman"} 0
+	patroni_xlog_replayed_location{scope="batman", name="patroni1"} 0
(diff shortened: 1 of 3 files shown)


patroni/api.py
```diff
@@ -1263,6 +1263,7 @@ def get_postgresql_status(self, retry: bool = False) -> Dict[str, Any]:
             row = self.query(stmt.format(postgresql.wal_name, postgresql.lsn_name,
                                          postgresql.wal_flush), retry=retry)[0]
             result = {
+                'name': postgresql.name,
```
CyberDem0n, line 1266: If we go this way maybe we should also add the `scope`?

CyberDem0n (commented):
I think it would also be worth adding the name as a tag to the `GET /metrics` endpoint: 
https://github.com/zalando/patroni/blob/89d794facc3e1fa728ccfbe6ac0323d668c8e021/patroni/api.py#L529

docs/rest_api.rst
```diff
@@ -482,6 +486,7 @@ Let's check that the node processed this configuration. First of all it should s
 
 	$ curl -s http://localhost:8008/patroni | jq .
 	{
+          "name": "patroni1",
 	  "pending_restart": true,
```
hughcapet, line 490:
```suggestion
      "name": "patroni1",
	  "pending_restart": true,
```

hughcapet, line 490:
```suggestion
       "name": "patroni1",
	  "pending_restart": true,
```

patroni/api.py
sskserk: I'm not sure what you are mentioning. I'm chasing the "name" in the output of the "/patroni". Let's do it with a separate PR

sskserk (commented):
QnA

patroni/api.py
CyberDem0n, line 1266: It doesn't deserve a separate PR.
```suggestion
                'scope': postgresql.scope,
                'name': postgresql.name,
```

sskserk (commented):
Addressed comment

CyberDem0n:
@sskserk what about https://github.com/zalando/patroni/pull/2846#pullrequestreview-1608775982?
This is a standard prometheus endpoint...

sskserk:
@CyberDem0n  prometheus endpoint is enhanced

hughcapet:
Now `/metrics` endpoint docs also need to be extended. Also please check pyright errors

CyberDem0n:
@sskserk this is not what I meant.
We need to add `name` to `labels`:
```diff
-        scope_label = '{{scope="{0}"}}'.format(patroni.postgresql.scope)
+        labels = f'{{scope="{patroni.postgresql.scope}",name="{patroni.postgresql.name}"}}'
```

And of course rename all occurrences of `scope_label` to `labels` down the method.

CyberDem0n:
A few more changes are needed:
1. Add scope to all examples of monitoring endpoints
2. update labels (add name) in the example of Prometheus output: https://github.com/zalando/patroni/blob/eb6bbea33853d241509ec67216296e65d1303f7f/docs/rest_api.rst?plain=1#L286-L349

And when looking at the [documentation](https://github.com/zalando/patroni/blob/eb6bbea33853d241509ec67216296e65d1303f7f/docs/rest_api.rst?plain=1#L351-L360?) I also wonder if we should add scope to the `GET /cluster` endpoint [here](https://github.com/zalando/patroni/blob/eb6bbea33853d241509ec67216296e65d1303f7f/patroni/utils.py#L802)?

docs/rest_api.rst
```diff
@@ -279,70 +283,70 @@ Retrieve the Patroni metrics in Prometheus format through the ``GET /metrics`` e
 	
 	# HELP patroni_version Patroni semver without periods. \
 	# TYPE patroni_version gauge
-	patroni_version{scope="batman"} 020103
+	patroni_version{scope="batman", name="patroni1"} 020103
```
CyberDem0n, line 290: no space character needed
```suggestion
	patroni_version{scope="batman",name="patroni1"} 020103
```
please remove it from the remaining lines.

CyberDem0n (commented):
1. you already added "name" to monitoring endpoints examples: https://github.com/zalando/patroni/pull/2846/files#diff-c10e80ec2654546cb5d06ed2aee8a7de61371f731c065d2afc623c7c13728291R101. Similarly you need to add "scope"
2. add "scope" here: https://github.com/zalando/patroni/blob/eb6bbea33853d241509ec67216296e65d1303f7f/patroni/utils.py#L802 and update respective [documentation](https://github.com/zalando/patroni/blob/eb6bbea33853d241509ec67216296e65d1303f7f/docs/rest_api.rst?plain=1#L351-L360).

sskserk:
I'm not sure how to obtain the 'scope' in the function which you mention in relation to the code ' ret: Dict[str, Any] = {'members': []} '

patroni/utils.py
```diff
@@ -809,7 +809,7 @@ def cluster_as_json(cluster: 'Cluster', global_config: Optional['GlobalConfig']
             role = 'replica'
 
         state = (m.data.get('replication_state', '') if role != 'leader' else '') or m.data.get('state', '')
-        member = {'name': m.name, 'role': role, 'state': state, 'api_url': m.api_url}
+        member = {'scope', global_config.scope, 'name': m.name, 'role': role, 'state': state, 'api_url': m.api_url}
```
CyberDem0n, line 812: Oh, you are right. Here the scope isn't available :(
We need to do it here: https://github.com/zalando/patroni/blob/6b7f914da7fd4974ca091c8f55c769b2ad0f58a3/patroni/api.py#L452

sskserk: Not sure what you are mentioning. This function just sends the response.

sskserk: Please suggest the change. It will take significantly less time and efforts.

CyberDem0n: Yes, but we can always modify it before sending:
```python
response = cluster_as_json(cluster, global_config)
response['scope'] = self.server.patroni.postgresql.scope
self._write_json_response(200, response) 
```

docs/rest_api.rst
```diff
@@ -356,6 +364,7 @@ Cluster status endpoints
     {
       "members": [
         {
+          "scope": "demo",
```
CyberDem0n, line 367: scope should be on the cluster level

CyberDem0n:
@sskserk sometimes they fail, MacOS VMs could be very slow :(

sskserk:
What are the next actions? @CyberDem0n

sskserk:
All tests passed

CyberDem0n:
Oh, @sskserk I am very sorry, but we'll have to redo some of the things :(
I totally missed the fact that the `scope` is already available in the response since many years: https://github.com/zalando/patroni/commit/704b29e6868eff28ffdea12d044e962b90d177e3 in the `{"patroni":{"scope":"$SCOPE"}}` and I falsely insisted on adding yet another one :(
Thanks @hughcapet for pointing it out. As a result I start thinking whether the "name" should also be put under "patroni"?
Any thoughts about it?

sskserk:
@CyberDem0n,
Do you want to move the "name" into the "patroni" struct? No problem at all. I'm chasing for the "name". It doesn't make a big difference where it will be available

sskserk (commented):
format

sskserk:
These is something wrong with the runner

patroni/api.py
```diff
@@ -1262,7 +1269,7 @@ def get_postgresql_status(self, retry: bool = False) -> Dict[str, Any]:
 
             row = self.query(stmt.format(postgresql.wal_name, postgresql.lsn_name,
                                          postgresql.wal_flush), retry=retry)[0]
-            result = {
+            result =
```
CyberDem0n, line 1272:
```suggestion
            result = {
```

hughcapet:
👍

CyberDem0n:
:+1:

CyberDem0n:
Merged, thank you @sskserk!

sskserk:
@CyberDem0n,
 thanks for merging the PR.
When will it be incorporated into a release?

We will start  recommending it to our customers.

sskserk:
Hi @CyberDem0n,

When do you plan to merge the feature into a release? We are are just looking forward to installing it into production.

adolsalamanca:
hello
Was this PR skipped on purpose on release v3.1.2?
We're interested on this.

Thanks!

adolsalamanca:
hello
Was this PR skipped on purpose on release `v3.1.2`?
We're interested on this.

Thanks!

hughcapet:
This feature will be included into 3.2, 3.1.2 is a bugfix only release

PR #2849, opened 2023-09-04
Debian docker image pip error

matthbakeredb:
The current Dockerfile uses pip to install global packages which throws an error on latest base image of Debian.


CyberDem0n:
:+1:

hughcapet:
👍

PR #2851, opened 2023-09-05
Refactor replication slots handling

CyberDem0n:
1. make _get_members_slots() method return data in the same format as _get_permanent_slots() method
2. move conflicting name handling from get_replication_slots() to _get_members_slots() method
3. enrich structure returned by get_replication_slots() with the LSN of permanent logical slots reported by primary
4. use the added information in the SlotsHandler instead of fetching it from the Cluster.slots
5. bugfix: don't try to advance logical slot that doesn't match required configuration

### DIFF
--- a/patroni/dcs/__init__.py
+++ b/patroni/dcs/__init__.py
@@ -903,8 +903,16 @@ class Cluster(NamedTuple('Cluster',
 
     @property
     def __permanent_slots(self) -> Dict[str, Union[Dict[str, Any], Any]]:
-        """Dictionary of permanent replication slots."""
-        return self.config and self.config.permanent_slots or {}
+        """Dictionary of permanent replication slots with their known LSN."""
+        ret = deepcopy(self.config.permanent_slots if self.config else {})
+        # If primary reported flush LSN for permanent slots we want to enrich our structure with it
+        for name, lsn in (self.slots or {}).items():
+            if name in ret:
+                if not ret[name]:
+                    ret[name] = {}
+                if isinstance(ret[name], dict):
+                    ret[name]['lsn'] = lsn
+        return ret
 
     @property
     def __permanent_physical_slots(self) -> Dict[str, Any]:
@@ -929,7 +937,6 @@ class Cluster(NamedTuple('Cluster',
 
         Will log an error if:
 
-            * Conflicting slot names between members are found
             * Any logical slots are disabled, due to version compatibility, and *show_error* is ``True``.
 
         :param my_name: name of this node.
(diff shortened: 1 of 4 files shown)


patroni/dcs/__init__.py
```diff
@@ -1042,35 +1037,48 @@ def _get_permanent_slots(self, is_standby_cluster: bool, role: str, nofailover:
 
         return self.__permanent_slots if role in ('master', 'primary') else self.__permanent_logical_slots
 
-    def _get_slot_members(self, my_name: str, role: str) -> List[str]:
+    def _get_members_slots(self, my_name: str, role: str) -> Dict[str, Dict[str, str]]:
```
matthbakeredb, line 1040: I'm not convinced this makes sense. You are changing the meaning of this method, whilst making it do the same as the old, adding a new purpose.

I would either:
1. Create a new method called `_get_members_slots`, which to me means "get the slots belonging to a member", and reference this after the call to `_get_slot_members`. This makes sure that `_get_members_slots` follows the "do one thing, well" principle.
2. (and honestly I much prefer 1 :smile:) update the docstring so that it better describe the new role of this method.

Just to expand on 1, I think that following the "one thing" rule will help to keep the complexity to a reasonable level whilst encouraging a flatter structure. It makes it easier to follow code, without needing to read the depth of it. Docstrings become easier to write. Method signatures need fewer args. Modules can be shorter...

matthbakeredb (changes requested):
Added a few of comments

patroni/dcs/__init__.py
```diff
@@ -903,8 +903,16 @@ def get_clone_member(self, exclude_name: str) -> Union[Member, Leader, None]:
...
     def __permanent_slots(self) -> Dict[str, Union[Dict[str, Any], Any]]:
-        """Dictionary of permanent replication slots."""
-        return self.config and self.config.permanent_slots or {}
+        """Dictionary of permanent replication slots with their known LSN."""
+        ret = deepcopy(self.config.permanent_slots if self.config else {})
+        # If primary reported flush LSN for permanent slots we want to enrich our structure with it
+        for name, lsn in (self.slots or {}).items():
+            if name in ret:
+                if not ret[name]:
+                    ret[name] = {}
```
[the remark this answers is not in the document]
CyberDem0n: This is not the same and will not work.

```yaml
slots:
  ph_slot:
```
Above is a valid configuration of a permanent physical replication slot, which will look like `{'slots':{'ph_slot': None}}`. setdefault in this case will do nothing.

patroni/dcs/__init__.py
CyberDem0n: There are very good reasons why I changed the method.
I am planning to create physical replication slots on all members that could become the leader and use pg_replication_slot_advance() function to move them forward based on data from `/member/*` keys.

For that I need to add `lsn` value to the `{'type': 'physical'}` (similarly to https://github.com/zalando/patroni/pull/2851/files#diff-247edd60473c0933e528a1b4621cc612da54b6dc964720719d7623135d312474R914). Doing it in the get_replication_slots() method seems totally wrong, therefore I renamed _get_slot_members() to _get_members_slots() and moved piece of the code that builds a structure there. But, forgot to update `Get a list of member names that have replication slots sourcing from this node` docstring.

patroni/dcs/__init__.py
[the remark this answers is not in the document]
matthbakeredb: Ah yes, that makes sense.

patroni/dcs/__init__.py
matthbakeredb: Thanks for the extra info. So you're intending on using `_get_members_slots()` from other methods?

I didn't mean to suggest that you didn't have good reason to make the change. Apologies if that's how it came across. I wasn't saying don't move the code out of `__permanent_physical_slots`, I meant have 2 separate methods for each of "getting a list of members" and "getting the slots of a member". If you think a separate `_get_slot_members` is overkill as they will always be together, that's good.

Updating the docstring is good too. :smile:

CyberDem0n: `_get_members_slots()` will be called as before, only from `get_replication_slots()`.

The new method isn't big and IMO it is way to early for splitting it.

matthbakeredb (approved):
:+1:

CyberDem0n:
:+1:

PR #2852, opened 2023-09-07
Permanent physical slots on standby nodes

CyberDem0n:
Create permanent physical replication slots on standby nodes and use `pg_replication_slot_advance()` function to move them forward.

The `restart_lsn` is advanced based on values stored in the `/status` key by the primary node.

When slot is created on a replica it could be ahead the same slot on the primary and therefore there is some period of time when it doesn't protect WAL files from being recycled.


tests/test_slots.py
```diff
@@ -178,3 +183,10 @@ def test_slots_advance_thread(self):
         with patch.object(SlotsHandler, 'get_local_connection_cursor', Mock(side_effect=Exception)):
             self.s.schedule_advance_slots({'foo': {'bar': 100}})
             self.s._advance.sync_slots()
+
+    @patch.object(Postgresql, 'is_primary', Mock(return_value=False))
+    def test_advance_physical_slots(self):
```
hughcapet, line 188: `patroni/postgresql/slots.py:389-390` (exception) is not covered
Can we also have the query call's content assertion here? 🥺

features/steps/slots.py
```diff
@@ -38,23 +44,46 @@ def does_not_have_logical_replication_slot(context, pg_name, slot_name):
         assert False, "Error looking for slot {0} on {1}".format(slot_name, pg_name)
 
 
-@step('Logical slot {slot_name:w} is in sync between {pg_name1:w} and {pg_name2:w} after {time_limit:d} seconds')
-def logical_slots_in_sync(context, slot_name, pg_name1, pg_name2, time_limit):
+@step('{slot_type:w} slot {slot_name:w} is in sync between {pg_name1:w} and {pg_name2:w} after {time_limit:d} seconds')
```
hughcapet, line 47: Is a new scenario for physical slots yet to be committed?

hughcapet:
👍

patroni/dcs/__init__.py
```diff
@@ -1026,16 +1028,19 @@ def _get_permanent_slots(self, is_standby_cluster: bool, role: str, nofailover:
...
+        :param major_version: postgresql major version.
 
         :returns: dictionary of permanent slot names mapped to attributes.
         """
         if not self.use_slots or nofailover:
             return {}
 
         if is_standby_cluster:
-            return self.__permanent_physical_slots if role == 'standby_leader' else {}
+            return self.__permanent_physical_slots\
```
matthbakeredb, line 1039:
```suggestion
            return self.__permanent_physical_slots \
```

patroni/dcs/__init__.py
```diff
@@ -1026,16 +1028,19 @@ def _get_permanent_slots(self, is_standby_cluster: bool, role: str, nofailover:
...
+                if major_version >= _SLOT_ADVANCE_AVAILABLE_VERSION or role == 'standby_leader' else {}
 
-        return self.__permanent_slots if role in ('master', 'primary') else self.__permanent_logical_slots
+        return self.__permanent_slots if major_version >= _SLOT_ADVANCE_AVAILABLE_VERSION\
```
matthbakeredb, line 1042:
```suggestion
        return self.__permanent_slots if major_version >= _SLOT_ADVANCE_AVAILABLE_VERSION \
```

patroni/postgresql/slots.py
```diff
@@ -260,16 +262,19 @@ def load_replication_slots(self) -> None:
         """
         if self._postgresql.major_version >= 90400 and self._schedule_load_slots:
             replication_slots: Dict[str, Dict[str, Any]] = {}
-            extra = ", catalog_xmin, pg_catalog.pg_wal_lsn_diff(confirmed_flush_lsn, '0/0')::bigint" \
+            pg_wal_lsn_diff = f'pg_catalog.pg_{self._postgresql.wal_name}_{self._postgresql.lsn_name}_diff'
```
matthbakeredb, line 265: I wonder if you would consider running `black` on code and having a test that errors if black would make changes to edited files? It might cut down on trivial formatting comments by standardising formatting.

```suggestion
            pg_wal_lsn_diff = f"pg_catalog.pg_{self._postgresql.wal_name}_{self._postgresql.lsn_name}_diff"
```

patroni/postgresql/slots.py
```diff
@@ -371,6 +378,16 @@ def _ensure_physical_slots(self, slots: Dict[str, Any]) -> None:
                 except Exception:
                     logger.exception("Failed to create physical replication slot '%s'", name)
                 self._schedule_load_slots = True
+            elif not self._postgresql.is_primary() and self._postgresql.major_version >= 110000\
```
matthbakeredb, line 381:
```suggestion
            elif not self._postgresql.is_primary() and self._postgresql.major_version >= 110000 \
```

patroni/postgresql/slots.py
```diff
@@ -371,6 +378,16 @@ def _ensure_physical_slots(self, slots: Dict[str, Any]) -> None:
...
+                    and self._replication_slots[name]['type'] == 'physical':
+                value['restart_lsn'] = self._replication_slots[name]['restart_lsn']
+                lsn = parse_int(value.get('lsn'))
+                if lsn and lsn > value['restart_lsn']:  # The slot has feedback in DCS and needs to be advanced
+                    try:
+                        lsn = format_lsn(lsn)
+                        self._query("SELECT pg_catalog.pg_replication_slot_advance(%s, %s)", name, lsn)
+                    except Exception as exc:
+                        logger.error("Error while advancing replication slot %s to position '%s': %r", name, lsn, exc)
```
matthbakeredb, line 390: Would `logger.exception` or `logger.error(..., exc_info=True)` work here?

matthbakeredb: Ah, I see now this is done throughout the code. I expect there's good reason for that...

features/steps/slots.py
```diff
@@ -15,17 +15,25 @@ def create_logical_replication_slot(context, slot_name, pg_name, plugin):
...
-    try:
-        row = context.pctl.query(pg_name, ("SELECT slot_type, plugin FROM pg_replication_slots"
-                                           " WHERE slot_name = '{0}'").format(slot_name)).fetchone()
-        assert row, "Couldn't find replication slot named {0}".format(slot_name)
-        assert row[0] == "logical", "Found replication slot named {0} but wasn't a logical slot".format(slot_name)
-        assert row[1] == plugin, ("Found replication slot named {0} but was using plugin "
-                                  "{1} rather than {2}").format(slot_name, row[1], plugin)
-    except pg.Error:
-        assert False, "Error looking for slot {0} on {1} with plugin {2}".format(slot_name, pg_name, plugin)
+@step('{pg_name:w} has a logical replication slot named {slot_name}'
```
matthbakeredb, line 18: Always impressed by this stuff.

features/permanent_slots.feature
```diff
@@ -0,0 +1,34 @@
...
+  Scenario: check that permanent slots are advanced on the replica
+    Given I add the table replicate_me to postgres0
+    And I get all changes from physical slot test_physical on postgres0
+    When I get all changes from logical slot test_logical on postgres0
+    Then Logical slot test_logical is in sync between postgres0 and postgres1 after 10 seconds
+    And Physical slot test_physical is in sync between postgres0 and postgres1 after 10 seconds
+
+  Scenario: check permanent physical replication slot after failover
+    Given I shut down postgres0
+    Then postgres1 has a physical replication slot named test_physical after 10 seconds
```
matthbakeredb, line 34: Should there be a `And postgres1 has a logical replication slot...`?

features/permanent_slots.feature
```diff
@@ -0,0 +1,34 @@
...
+  Scenario: check that logical permanent slots are created
+    Given I run patronictl.py restart batman postgres0 --force
+    And I issue a PATCH request to http://127.0.0.1:8008/config with {"slots": {"test_logical": {"type": "logical", "database": "postgres", "plugin": "test_decoding"}}}
+    Then postgres0 has a logical replication slot named test_logical with the test_decoding plugin after 10 seconds
+
+  @slot-advance
+  Scenario: check that permanent slots are created on the replica
+    Given postgres1 has a logical replication slot named test_logical with the test_decoding plugin after 10 seconds
+    Then Logical slot test_logical is in sync between postgres0 and postgres1 after 10 seconds
+    And postgres1 has a physical replication slot named test_physical after 2 seconds
```
matthbakeredb, line 22: Is the creation of the physical slot directly related to the existence of the logical slot in the `Given` statement? I feel like it would exist as a result of starting postgres1? If that is right should there then be a `Scenario` between the first and second that has `Given I start postgres1` with this `postgres1 has a physical replication slot named test_physical`?

patroni/dcs/__init__.py
```diff
@@ -1080,42 +1085,50 @@ def _get_members_slots(self, my_name: str, role: str) -> Dict[str, Dict[str, str
                                    for k, v in slot_conflicts.items() if len(v) > 1))
         return slots
 
-    def has_permanent_logical_slots(self, my_name: str, nofailover: bool, major_version: int = 110000) -> bool:
+    def has_permanent_slots(self, my_name: str, nofailover: bool = False) -> bool:
```
matthbakeredb, line 1088: Not a critique but rather a query so I know how better to review in the future.

Are we concerned at all about anyone who might be using patroni as a standard library which they might implement their own extensions from? I am thinking if a customer has say an internal key/value store and they want to write their own DCS module.

matthbakeredb (changes requested):

features/permanent_slots.feature
CyberDem0n: Thats totally true, but we need to start postgres1 in the first scenario, because otherwise the last scenario will fail on 9.6 and 10 (they don't run scenarios tagged with `@slot-advance`).

features/permanent_slots.feature
CyberDem0n: No, it will not work on 9.6 and 10.
Also, we already checked that the logical slot existed in previous scenario while postgres1 was still a replica.
After promote it will still be there.

patroni/dcs/__init__.py
CyberDem0n: While in theory anyone could implement their own, on practice I don't expect in ever happen, so we support only DCS that are included in `patroni/dcs`.

patroni/postgresql/slots.py
CyberDem0n: There is no real good reason to show a full trace in this case.

patroni/dcs/__init__.py
matthbakeredb: Ok cool, thanks for confirming.

matthbakeredb (approved):
lgtm

features/permanent_slots.feature
barthisrael: Should we keep claiming to support Postgres versions in Patroni that are not supported by PGDG?
IMO that kind of incentives people that are running old versions to keep running them.
I'd say it would be a bit less relaxed if we support the 5 versions supported by PGDG + the one that just went EOL, so people can move out in the meantime.
What are your thoughts on that?

patroni/postgresql/__init__.py
```diff
@@ -456,7 +454,7 @@ def _cluster_info_state_get(self, name: str) -> Optional[Any]:
                                                'received_tli', 'slot_name', 'conninfo', 'receiver_state',
                                                'restore_command', 'slots', 'synchronous_commit',
                                                'synchronous_standby_names', 'pg_stat_replication'], result))
-                if self._has_permanent_logical_slots:
+                if self._has_permanent_slots and self.major_version >= 110000:
```
barthisrael, line 457: I wonder if we should reuse the variable created in `patroni.dcs`?

barthisrael (changes requested):
I like the changes.

I'd suggest adjusting the docs, e.g. we have the following mention:

> The physical slots are maintained only in the current primary.

I also noticed the following behavior on replicas when you clash the slot name with the primary member name:

```bash
$ PGPORT=5433 PGPASSWORD=zalando psql -h localhost -U postgres -c "TABLE pg_replication_slots"
  slot_name  | plugin | slot_type | datoid | database | temporary | active | active_pid | xmin | catalog_xmin | restart_lsn | confirmed_flush_lsn | wal_status | safe_wal_size | two_phase
-------------+--------+-----------+--------+----------+-----------+--------+------------+------+--------------+-------------+---------------------+------------+---------------+-----------
 postgresql0 |        | physical  |        |          | f         | f      |            |      |              | 0/B000B38   |                     | reserved   |               | f
 postgresql2 |        | physical  |        |          | f         | f      |            |      |              | 0/F050BB0   |                     | reserved   |               | f
(2 rows)
```

```bash
$ PGPORT=5434 PGPASSWORD=zalando psql -h localhost -U postgres -c "TABLE pg_replication_slots"
  slot_name  | plugin | slot_type | datoid | database | temporary | active | active_pid | xmin | catalog_xmin | restart_lsn | confirmed_flush_lsn | wal_status | safe_wal_size | two_phase
-------------+--------+-----------+--------+----------+-----------+--------+------------+------+--------------+-------------+---------------------+------------+---------------+-----------
 postgresql0 |        | physical  |        |          | f         | f      |            |      |              | 0/B000B38   |                     | reserved   |               | f
 postgresql1 |        | physical  |        |          | f         | f      |            |      |              | 0/F050BB0   |                     | reserved   |               | f
(2 rows)
```

But we already warn in the documentation that users should take care when clashing with the nodes' names.

Maybe in the future we could think of a way of keeping slots around for a while if a member goes unresponsive.

CyberDem0n:
> it won't advance the LSN of a slot which matches the name of the current primary because the slot doesn't exist in the primary:

Nice catch! Fixed in 852205e6856f302d2519cfc2125ab04d39dd6822.

features/permanent_slots.feature
barthisrael: I see, ok! TBH I wouldn't expect people to be running on 9.3 e.g. -- although there are always weird cases 😆 .

hughcapet: ~i can hear greenplum users crying~

patroni/dcs/__init__.py
```diff
@@ -912,13 +926,22 @@ def __permanent_slots(self) -> Dict[str, Union[Dict[str, Any], Any]]:
...
+
+        # there is no slot on the leader for itself, use `lsn` from the member key.
+        leader = self.leader and self.leader.member
+        if leader and leader.lsn:
+            name = slot_name_from_member_name(leader.name)
+            if name in ret and self.is_physical_slot(ret[name]):
+                if not isinstance(ret[name], dict):
+                    ret[name] = {}
+                if 'lsn' not in ret[name]:
+                    ret[name]['lsn'] = leader.lsn
```
barthisrael, line 938: Why do we update only if `'lsn' not in ret[name]'`?
In the previous part, which handles the non-leader slot, we always update the LSN.
The above code could face issues if by chance ones does something like this in the config:

```
slots:
  postgresql0:
    lsn: 123
    type: physical
  postgresql1:
    type: physical
  postgresql2:
    type: physical
```

Unlikely to occur, but an open door 😆

barthisrael (changes requested):
> Nice catch! Fixed in https://github.com/zalando/patroni/commit/852205e6856f302d2519cfc2125ab04d39dd6822.

Nice stuff!

I think we should now amend the documentation of permanent replication slots to reflect its new behavior.

Besides that I think we can fix the description ":ivar slots: state of permanent logical replication slots on the primary in the format: {"slot_name": int}." in "class Cluster" docstring.

Other than that I'm really happy about this PR 😄

docs/dynamic_configuration.rst
```diff
@@ -46,7 +46,7 @@ In order to change the dynamic configuration you can use either ``patronictl edi
...
...all nodes and their position is advanced every **loop_wait** seconds. For older PostgreSQL sversions older than 11 permanent physical replication slots are maintained only on the current primary. The logical slots are copied from the primary to a standby with restart, and after that their position advanced every **loop_wait** seconds (if necessary). Copying logical slot files performed via ``libpq`` connection and using either rewind or superuser credentials (see **postgresql.authentication** section). There is always a chance that the logical slot position on the replica is a bit behind the former primary, therefore application should be prepared that some messages could be received the second time after the failover. The easiest way of doing so - tracking ``confirmed_flush_lsn``. Enabling permanent replication slots requires **postgresql.use_slots** to be set to ``true``. If there are permanent logical replication slots defined Patroni will automatically enable the ``hot_standby_feedback``. Since the failover of logical replication slots is unsafe on PostgreSQL 9.6 and older and PostgreSQL version 10 is missing some important functions, the feature only works with PostgreSQL 11+.
```
barthisrael, line 49:
```suggestion
-  **slots**: define permanent replication slots. These slots will be preserved during switchover/failover. Permanent slots that don't exist will be created by Patroni. With PostgreSQL 11 onwards permanent physical slots are created on all nodes and their position is advanced every **loop_wait** seconds. For PostgreSQL versions older than 11 permanent physical replication slots are maintained only on the current primary. The logical slots are copied from the primary to a standby with restart, and after that their position advanced every **loop_wait** seconds (if necessary). Copying logical slot files performed via ``libpq`` connection and using either rewind or superuser credentials (see **postgresql.authentication** section). There is always a chance that the logical slot position on the replica is a bit behind the former primary, therefore application should be prepared that some messages could be received the second time after the failover. The easiest way of doing so - tracking ``confirmed_flush_lsn``. Enabling permanent replication slots requires **postgresql.use_slots** to be set to ``true``. If there are permanent logical replication slots defined Patroni will automatically enable the ``hot_standby_feedback``. Since the failover of logical replication slots is unsafe on PostgreSQL 9.6 and older and PostgreSQL version 10 is missing some important functions, the feature only works with PostgreSQL 11+.
```

Fixing typos.

barthisrael (approved):
Looks good!
Just a small suggestion to fix a couple typos.

hughcapet:
👍

CyberDem0n:
:+1:

PR #2854, opened 2023-09-11
Extend documentation with package installation and upgrade process

matthbakeredb:
Adds instructions for installing Patroni using packages and upgrading Patroni safely. To save the README from growing ever larger this moves the installation instructions to its own page.


hughcapet:
👍

PR #2858, opened 2023-09-13
Env var of PATRONI_RESTAPI_CONNECT_ADDRESS and PATRONI_POSTGRESQL_CON…

devopsy-ir:
…NECT_ADDRESS to make available on diffrent vms and posrts


CyberDem0n:
@devopsy-ir have you seen the comment? https://github.com/zalando/patroni/blob/b31a4d55c9c628c6489a2c2bd7be1d0611577ead/Dockerfile#L1-L2

devopsy-ir:
> @devopsy-ir have you seen the comment?
> 
> https://github.com/zalando/patroni/blob/b31a4d55c9c628c6489a2c2bd7be1d0611577ead/Dockerfile#L1-L2

Yes, you are true. But what about multi node deployment?

CyberDem0n:
> Yes, you are true. But what about multi node deployment?

These Dockerfile, entrypoint, and docker-compose are to play with Patroni _locally_ and never meant for anything other than that.

devopsy-ir:
> > Yes, you are true. But what about multi node deployment?
> 
> These Dockerfile, entrypoint, and docker-compose are to play with Patroni _locally_ and never meant for anything other than that.

What is the prolem of reading this variables from env var of os?!!

PR #2861, opened 2023-09-14
Always store integer CMDLINE_OPTIONS config values as int

hughcapet:
In addition to the validation of the integer values defined in `CMDLINE_OPTIONS`, always cast and store them as int (to be on the safe side when e.g. performing arithmetical operations)

There is at least one bug caused by this now: `max_prepared_transactions` value for a Citus cluster [can become a doubled string](https://github.com/zalando/patroni/blob/master/patroni/postgresql/citus.py#L401)

### DIFF
--- a/patroni/config.py
+++ b/patroni/config.py
@@ -16,6 +16,7 @@ from .dcs import ClusterConfig, Cluster
 from .exceptions import ConfigParseError
 from .file_perm import pg_perm
 from .postgresql.config import ConfigHandler
+from .validator import IntValidator
 from .utils import deep_compare, parse_bool, parse_int, patch_config
 
 logger = logging.getLogger(__name__)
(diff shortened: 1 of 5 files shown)


CyberDem0n:
:+1:

hughcapet:
👍

jorsol:
Does this handle parameters with units?

I'm getting the following error that should be correct:
```
2023-09-22 15:25:51,367 ERROR: Exception when setting dynamic_configuration
  File "/usr/lib/python3.9/site-packages/patroni/config.py", line 312, in set_dynamic_configuration
    self.__effective_configuration = self._build_effective_configuration(configuration,
  File "/usr/lib/python3.9/site-packages/patroni/config.py", line 537, in _build_effective_configuration
    config = self._safe_copy_dynamic_configuration(dynamic_configuration)
  File "/usr/lib/python3.9/site-packages/patroni/config.py", line 359, in _safe_copy_dynamic_configuration
    config['postgresql'][name].update(self._process_postgresql_parameters(value))
  File "/usr/lib/python3.9/site-packages/patroni/config.py", line 345, in _process_postgresql_parameters
    pg_params[name] = int(value) if isinstance(validator, IntValidator) else value
ValueError: invalid literal for int() with base 10: '1536MB'
```

Should I open a new bug?

hughcapet:
@jorsol what is the parameter name?

jorsol:
`wal_keep_size`, which is actually here:
https://github.com/zalando/patroni/blob/6507ec7b82e1eae1bcb1cbedb406483f2714949d/patroni/postgresql/config.py#L296C10-L296C23

I'm not familiar with python so I don't know the logic that is being executed.

hughcapet:
@jorsol yes, please, open an issue for this bug, we will fix it next week (you can meanwhile set the value without units)

jorsol:
Thanks @hughcapet , I have opened the issue for this bug.

This parameter is set automatically by StackGres operator and is transformed to a value with a unit, so is not possible right now to set the value without units as a workaround, so we need to wait for the next release of Patroni with the fix.

PR #2863, opened 2023-09-14
Don't rely on pg_stat_wal_receiver when deciding on pg_rewind

CyberDem0n:
As was reported by @ants on Slack it could happen that `received_tli` is ahead of replayed timeline, therefore we should stop using it when deciding on pg_rewind if postgres is running and use only `IDENTIFY_SYSTEM` via replication connection.


hughcapet:
👍

CyberDem0n:
:+1:

Issue #2865, opened 2023-09-16
Syntax error when creating postgres user


GuanqunYang193:
### What happened?

Patroni enables the inclusion of specific options when creating a user, to be executed within the context of the CREATE statement. Internally, it performs a check to determine if the user exists. If the user does exist, it employs the CREATE statement; otherwise, it will use the ALTER statement. Nevertheless, regarding PostgreSQL, some options are exclusively supported by the CREATE statement (membership-related). This implies that attempting to utilize these options will result in a syntax error.

### How can we reproduce it (as minimally and precisely as possible)?

add "in role xx" under options and run any behave test.

### What did you expect to happen?

user can be created and the role can be granted.

### Patroni/PostgreSQL/DCS version

-   Patroni version: 3
-   PostgreSQL version: 13

### Patroni configuration file

```yaml
template
```

### patronictl show-config

```yaml
template
```

### Patroni log files

```shell
template
```

### PostgreSQL log files

```shell
template
```

### Have you tried to use GitHub issue search?

- [X] Yes

Issue #667, opened 2018-04-19
Ordering of user creation and post_bootstrap execution


bradnicholson:
Is there a reason why we create all of the non-superuser roles after running the post_bootstrap?

What I'd ideally like to do is create a set of users via Patroni, then be able to create DB objects owned by those users are part of a post_bootstrap script.  While this can be worked around by creating the users in the postbootstrap script, that seems wrong to me.

If there are no objections to creating the users prior to running the post_bootstrap script, I'll put a PR up for the change.

alexeyklyukin:
While I see nothing wrong in creating users in the `post_bootstrap` script I also see no reason why do we create users mentioned in the Patroni configuration before and not after running the post_bootstrap script.

I'd expect people would use one of those two, i.e. either do everything via the `post_bootstrap` or create users with the Patroni configuration. Mixing both looks a bit weird, you will probably end up with a script that assigns privileges to a given set of users while there is no definition of those users anywhere in the script. So creating 'normal' users before and not after the post_bootstrap phase looks like a nuance to me.

PR #2866, opened 2023-09-16
Fix syntax error when creating users

GuanqunYang193:
fix the [issue](https://github.com/zalando/patroni/issues/2865)

* supported add membership of the specified existing roles when creating the user
* added step in behave test to check the membership created

### DIFF
--- a/features/basic_replication.feature
+++ b/features/basic_replication.feature
@@ -4,6 +4,7 @@ Feature: basic replication
   Scenario: check replication of a single table
     Given I start postgres0
     Then postgres0 is a leader after 10 seconds
+    Then admin in postgres0 is a member of pg_monitor after 10 seconds
     And there is a non empty initialize key in DCS after 15 seconds
     When I issue a PATCH request to http://127.0.0.1:8008/config with {"ttl": 20, "synchronous_mode": true}
     Then I receive a response code 200
--- a/features/environment.py
+++ b/features/environment.py
@@ -308,6 +308,23 @@ class PatroniController(AbstractController):
                     return True
             time.sleep(1)
         return False
+    
+    def check_user_member_of(self, username, role, timeout=10):
+        bound_time = time.time() + timeout
+        while time.time() < bound_time:
+            cur = self.query('''
+            SELECT * 
+            FROM pg_auth_members a, 
+                 pg_roles b, 
+                 pg_roles c  
+            WHERE a.roleid = b.oid 
+            AND a.member = c.oid 
+            AND b.rolname = '{0}' 
+            AND c.rolname = '{1}' '''.format(username,role), fail_ok=True)
+            if cur:
+                    return True
+            time.sleep(1)
+        return False
 
     def get_watchdog(self):
         return self.watchdog
@@ -832,7 +849,7 @@ class PatroniPoolController(object):
         self._processes[name].start(max_wait_limit)
 
     def __getattr__(self, func):
-        if func not in ['stop', 'query', 'write_label', 'read_label', 'check_role_has_changed_to',
+        if func not in ['stop', 'query', 'write_label', 'read_label', 'check_role_has_changed_to', 'check_user_member_of',
                         'add_tag_to_config', 'get_watchdog', 'patroni_hang', 'backup', 'read_patroni_log']:
             raise AttributeError("PatroniPoolController instance has no attribute '{0}'".format(func))
 
--- a/features/environment.py
+++ b/features/environment.py
@@ -308,21 +308,21 @@ class PatroniController(AbstractController):
                     return True
             time.sleep(1)
         return False
-    
+
     def check_user_member_of(self, username, role, timeout=10):
         bound_time = time.time() + timeout
         while time.time() < bound_time:
             cur = self.query('''
-            SELECT * 
-            FROM pg_auth_members a, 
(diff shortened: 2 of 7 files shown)


CyberDem0n:
To be honest, it was one of the wrong decisions we made 7 years ago when implemented this feature.
The thing is that once `in role` support is fixed someone will try to use it for roles that are also defined in Patroni config, we we will have to resolve dependencies between roles.
It's not a Patroni task to manage users (replication user is an exception) and I would rather remove this feature entirely than spend time on supporting it.

GuanqunYang193:
> > We create users based on the order set by the user.
> 
> That's the point, users are defined in a dictionary and there is no ordering.

So, during the process of reading the config, it won't preserve the order in which users are defined in the config?

CyberDem0n:
> So, during the process of reading the config, it won't preserve the order in which users are defined in the config?

Config is a YAML file, it is processed by a library. The only we have is a data structure that returned from it. And `users` is a dictionary. Even if the order of elements is preserved like in the YAML file that would be an implementation detail that we shouldn't rely on.

> from my perspective, removing these features would require significant changes and could lead to version incompatibility.

removing it is fairly simple:
1. https://github.com/zalando/patroni/blob/25ceb6825741fb3ac6f9e535af4f3336301edbb0/patroni/postgresql/bootstrap.py#L403-L405
2. https://github.com/zalando/patroni/blob/25ceb6825741fb3ac6f9e535af4f3336301edbb0/patroni/config.py#L752-L767
3. Maybe a few lines in unit tests.

From the Patroni side it all backward compatible. Just the `bootstrap.users` and some environment variables will be ignored.

Really, user-management is a complex task and it is not only about creating users when the cluster is created, but also about possibility to add new users during the cluster life time, about removing unused users, and maybe updating passwords. It is not something that Patroni should do.

Last, but not least, there is a post_bootstrap hook, which you can use to do whatever you like when the new cluster is created.

GuanqunYang193:
1. from the official [doc](https://yaml.org/spec/1.2-old/spec.html#id2802662) the order of list in YAML would be kept.
2. It is not a easy change to not support creating users since:
* although we can ignore the config client create, but when they upgrade to our latest version, it would be a huge change for them since they have to find some where to implement the logic.
* how backup restore would be affected? If we ignore the users in backup files, does that means we can not restore from the file anymore?
* yes I notice there is a script can be use, but from [here](https://github.com/zalando/patroni/issues/667), actually is also the issue we have met, the order of running this script limits some functionalities.

Discontinuing support for user creation is a significant change compared to this PR, as it would affect a large number of users. However, this PR is unlikely to have any impact on existing users. I would still recommend that if we are considering discontinuing user creation, we should further discuss what else should be discontinued in terms of user-defined configurations and then implement those changes together. Personally, I really like this feature, as it provides added security and reliability by keeping this logic in the script rather than relying solely on Patroni.

CyberDem0n:
> from the official [doc](https://yaml.org/spec/1.2-old/spec.html#id2802662) the order of list in YAML would be kept.

You are referring to a list, while `bootstrap.users` is a mapping (python dict).

> how could they deal with existing database instances?

Existing? The `bootstrap` section works only when the new cluster is initialized.

> how backup restore would be affected? If we ignore the users in backup files, does that means we can not restore from the file anymore?

Patroni doesn't interfere with backups.

> yes I notice there is a script can be used, but from https://github.com/zalando/patroni/issues/667, actually is also the issue we have met, the order of running this script limits some functionalities.

The issue also says to stick with one of those, users or post_bootstrap.

> Discontinuing support for user creation is a significant change compared to this PR

How removing ~20 lines of code would be a significant change comparing to this PR?

> Personally, I really like this feature, as it is more reliable to rely on Patroni comparing to implementing and running a script.

It is not about running a script, but about having a proper system that manages users. Patroni shouldn't be solving this problem on the first place.

GuanqunYang193:
> > from the official [doc](https://yaml.org/spec/1.2-old/spec.html#id2802662) the order of list in YAML would be kept.
> 
> You are referring to a list, while `bootstrap.users` is a mapping (python dict).
> 
> > how could they deal with existing database instances?
> 
> Existing? The `bootstrap` section works only when the new cluster is initialized.
> 
> > how backup restore would be affected? If we ignore the users in backup files, does that means we can not restore from the file anymore?
> 
> Patroni doesn't interfere with backups.
> 
> > yes I notice there is a script can be used, but from #667, actually is also the issue we have met, the order of running this script limits some functionalities.
> 
> The issue also says to stick with one of those, users or post_bootstrap.
> 
> > Discontinuing support for user creation is a significant change compared to this PR
> 
> How removing ~20 lines of code would be a significant change comparing to this PR?
> 
> > Personally, I really like this feature, as it is more reliable to rely on Patroni comparing to implementing and running a script.
> 
> It is not about running a script, but about having a proper system that manages users. Patroni shouldn't be solving this problem on the first place.

fine, can you create an issue about this and I can implement and submit a PR.

GuanqunYang193:
@hughcapet do you think it is ok to stop supporting user creation or use this PR change to enrich the functionality

hughcapet:
@GuanqunYang193 yes, I do think user management is not Patorni's task and we can remove it in the next major release (preliminarily adding a deprecation warning to the docs and maybe also log warning to the next minor release)

PR #2871, opened 2023-09-18
Bump pyright version

CyberDem0n:
and fix all reported issues.

We aren't sticking to the latest version this time because it has [a bug](https://github.com/microsoft/pyright/issues/5968).


hughcapet:
👍

CyberDem0n:
:+1:

PR #2874, opened 2023-09-20
Add a documentation page for `patronictl`

barthisrael:
This PR introduces a documentation page for `patronictl` application.

We adopted a top-down approach when writing this document. We start by describing the outer most parts, and then keep writing new sections that specialize the knowledge.

We basically added a section called `patronictl` to the left menu. Inside that section we created a page with this structure:

- `patronictl`: describes what it is
    - `Configuraiton`: how to configure `patronictl`
    - `Usage`: how to use the CLI. Inside this section, there are subsections for each of the subcommands exposed by `patronictl`, and each of them are described using the following subsubsections:
        - `Synopsis`: syntax of the command and its positional and optional arguments
        - `Description`: a description of what the command does
        - `Parameters`: a detailed description of the arguments and how to use them
        - `Examples`: one or more examples of execution of the command

References: PAT-200.

### DIFF
--- a/docs/patronictl.rst
+++ b/docs/patronictl.rst
@@ -18,7 +18,7 @@ Configuration
 
 Those configuration options can come either from environment variables or from a configuration file. Look for the above sections in :ref:`Environment Configuration Settings <environment>` or :ref:`YAML Configuration Settings <yaml_configuration>` to understand how you can set the options for them through environment variables or through a configuration file.
 
-If you opt for using environment variables, it's a straight forward approach. Patroni will read the environment variables and use their values.
+If you opt for using environment variables, it's a straight forward approach. Patronictl will read the environment variables and use their values.
 
 If you opt for using a configuration file, you have different ways to inform ``patronictl`` about the file to be used. By default ``patronictl`` will attempt to load a configuration file named ``patronictl.yaml``, which is expected to be found under either of these paths, according to your system:
 
@@ -129,7 +129,6 @@ Parameters
     - ``replica``: a replica of a Patroni cluster; or
     - ``standby``: same as ``replica``; or
     - ``any``: any role. Same as omitting this parameter; or
-    - ``master``: same as ``primary``.
 
 ``-m`` / ``--member``
     Choose a member of the cluster with the given name.
@@ -148,9 +147,9 @@ Examples
 
 Get DSN of the primary node:
 
-.. code:: text
+.. code:: bash
 
-    patronictl -c postgres0.yml dsn batman -r primary
+    $ patronictl -c postgres0.yml dsn batman -r primary
     host=127.0.0.1 port=5432
 
 Get DSN of the node named ``postgresql1``:
@@ -203,6 +202,8 @@ Parameters
 
 ``--group``
     Change dynamic configuration of the given Citus group.
+    
+    If not given, ``patronictl`` will attempt to fetch that from the ``citus.group`` configuration, if it exists.
 
     ``CITUS_GROUP`` is the ID of the Citus group.
 
@@ -249,7 +250,7 @@ Examples
 
 Change ``max_connections`` Postgres GUC:
 
-.. code:: text
+.. code:: diff
 
     patronictl -c postgres0.yml edit-config batman --pg max_connections="150" --force
     ---
@@ -335,6 +336,7 @@ It is designed to be used when the cluster is not healthy, e.g.:
 - There is no leader; or
 - There is no synchronous standby available in a synchronous cluster.
 
+It also allows to fail over to asynchronous node is synchronous mode is enabled.
 .. note::
     Nothing prevents you from running ``patronictl failover`` in a healthy cluster. However, we recommend using ``patronictl switchover`` in those cases.
 
@@ -572,6 +574,8 @@ Parameters
     Show history of events from the given Citus group.
 
     ``CITUS_GROUP`` is the ID of the Citus group.
+    
+    If not given, ``patronictl`` will attempt to fetch that from the ``citus.group`` configuration, if it exists.
 
 ``-f`` / ``--format``
     How to format the list of events in the output.
@@ -910,6 +914,8 @@ Parameters
     Pause the given Citus group.
 
     ``CITUS_GROUP`` is the ID of the Citus group.
+    
+    If not given, ``patronictl`` will attempt to fetch that from the ``citus.group`` configuration, if it exists.
 
 ``--wait``
     Wait until all Patroni members are paused before returning control to the caller.
@@ -1526,6 +1532,8 @@ Parameters
     Resume the given Citus group.
 
     ``CITUS_GROUP`` is the ID of the Citus group.
+    
+    If not given, ``patronictl`` will attempt to fetch that from the ``citus.group`` configuration, if it exists.
 
 ``--wait``
     Wait until all Patroni members are unpaused before returning control to the caller.
@@ -1580,6 +1588,8 @@ Parameters
     Show dynamic configuration of the given Citus group.
 
     ``CITUS_GROUP`` is the ID of the Citus group.
+    
+    If not given, ``patronictl`` will attempt to fetch that from the ``citus.group`` configuration, if it exists.
 
 .. _patronictl_show_config_examples:
 
--- a/docs/patronictl.rst
+++ b/docs/patronictl.rst
@@ -5,6 +5,8 @@ Patronictl
 
 Patroni has a command-line interface named ``patronictl``, which is used basically to interact with Patroni's REST API and with the DCS. It is intended to make it easier to perform operations in the cluster, and can easily be used by humans or scripts.
 
+.. _patronictl_configuration:
+
 Configuration
 -------------
 
@@ -34,6 +36,8 @@ You can override that behavior either by:
 
 .. note::
     If you are running ``patronictl`` in the same host as ``patroni`` daemon is running, you may just use the same configuration file if it contains all the configuration sections required by ``patronictl``.
+
+.. _patronictl_usage:
 
 Usage
 -----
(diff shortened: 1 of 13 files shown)


hughcapet:
one thought straight away - should we add references to the new page for the existing occurrences of patronictl in the other parts of the docs?

barthisrael:
> one thought straight away - should we add references to the new page for the existing occurrences of patronictl in the other parts of the docs?

Hmm, good point, I guess so.

barthisrael:
I just submitted a commit that will allow any cross-reference to sections/subsections/subsubsections under `patronictl` page.

However, there is a behavior that I was not able to implement properly regarding the `patronictl subcommands` subsubsections: the first command, `patronictl dsn`, doesn't use the same anchor name format than others. I mean:

* I have these labes for `patronictl dsn` subsubsections:
    * `.. _patronictl_dsn_synopsis:`
    * `.. _patronictl_dsn_description:`
    * `.. _patronictl_dsn_parameters:`
    * `.. _patronictl_dsn_examples:`

But the HTML anchor IDs are `#synopsis`, `#description`, `#parameters`, and `#examples`, respectively.

That seems to occur only for the first occurrences of the duplicated subsubections names.

From the second subcommand onward (`patronictl edit-config` onward), I get anchor IDs named as I would expect: `#patronictl-edit-config-synopsis`, `#patronictl-edit-config-description`, `#patronictl-edit-config-parameters`, `#patronictl-edit-config-examples`, respectively.

Any idea on how to fix that on Sphinx, and make it use the same anchor ID naming convention for the first occurrences of duplicated subsubsections names too?

barthisrael:
_I just noticed that each commit is triggering a new workflow run. I'll submit the next ones in a single batch_

barthisrael:
I'm done with the commits. My only pending issues is the one described here (if that's an issue): https://github.com/zalando/patroni/pull/2874#issuecomment-1727941755.

I was not able to find a built-in way or resolving that. Do you happen to know any?

docs/patronictl.rst
```diff
@@ -0,0 +1,1963 @@
...
+Parameters
+""""""""""
+
+``CLUSTER_NAME``
+    Name of the Patroni cluster.
+
+    If not given, ``patronictl`` will attempt to fetch that from the ``scope`` configuration, if it exists.
+
+``--group``
+    Change dynamic configuration of the given Citus group.
```
CyberDem0n, line 205:
```suggestion
    Change dynamic configuration of the given Citus group.
    
    If not given, ``patronictl`` will attempt to fetch that from the ``citus.group`` configuration, if it exists.
```

docs/patronictl.rst
```diff
@@ -0,0 +1,1963 @@
...
+
+``CLUSTER_NAME``
+    Name of the Patroni cluster.
+
+    If not given, ``patronictl`` will attempt to fetch that from the ``scope`` configuration, if it exists.
+
+``--group``
+    Pause the given Citus group.
+
+    ``CITUS_GROUP`` is the ID of the Citus group.
```
CyberDem0n, line 912:
```suggestion
    ``CITUS_GROUP`` is the ID of the Citus group.
    
    If not given, ``patronictl`` will attempt to fetch that from the ``citus.group`` configuration, if it exists.
```

docs/patronictl.rst
```diff
@@ -0,0 +1,1963 @@
...
+
+``CLUSTER_NAME``
+    Name of the Patroni cluster.
+
+    If not given, ``patronictl`` will attempt to fetch that from the ``scope`` configuration, if it exists.
+
+``--group``
+    Resume the given Citus group.
+
+    ``CITUS_GROUP`` is the ID of the Citus group.
```
CyberDem0n, line 1528:
```suggestion
    ``CITUS_GROUP`` is the ID of the Citus group.
    
    If not given, ``patronictl`` will attempt to fetch that from the ``citus.group`` configuration, if it exists.
```

docs/patronictl.rst
```diff
@@ -0,0 +1,1963 @@
...
+
+``CLUSTER_NAME``
+    Name of the Patroni cluster.
+
+    If not given, ``patronictl`` will attempt to fetch that from the ``scope`` configuration, if it exists.
+
+``--group``
+    Show history of events from the given Citus group.
+
+    ``CITUS_GROUP`` is the ID of the Citus group.
```
CyberDem0n, line 574:
```suggestion
    ``CITUS_GROUP`` is the ID of the Citus group.
    
    If not given, ``patronictl`` will attempt to fetch that from the ``citus.group`` configuration, if it exists.
```

docs/patronictl.rst
```diff
@@ -0,0 +1,1963 @@
...
+
+``CLUSTER_NAME``
+    Name of the Patroni cluster.
+
+    If not given, ``patronictl`` will attempt to fetch that from the ``scope`` configuration, if it exists.
+
+``--group``
+    Show dynamic configuration of the given Citus group.
+
+    ``CITUS_GROUP`` is the ID of the Citus group.
```
CyberDem0n, line 1582:
```suggestion
    ``CITUS_GROUP`` is the ID of the Citus group.
    
    If not given, ``patronictl`` will attempt to fetch that from the ``citus.group`` configuration, if it exists.
```

docs/patronictl.rst
```diff
@@ -0,0 +1,1963 @@
...
+    Perform a failover in the given Citus group.
+
+    ``CITUS_GROUP`` is the ID of the Citus group.
+
+``--leader`` / ``--primary`` / ``--master``
+    Indicate who is the expected leader at failover time.
+
+    If given, a switchover is performed instead of a failover.
+
+    ``LEADER_NAME`` should match the name of the current leader in the cluster.
```
CyberDem0n, line 364: It is getting deprecated, should we still document it?

docs/patronictl.rst
```diff
@@ -0,0 +1,1963 @@
...
+Description
+"""""""""""
+
+``patronictl failover`` performs a manual failover in the cluster.
+
+It is designed to be used when the cluster is not healthy, e.g.:
+
+- There is no leader; or
+- There is no synchronous standby available in a synchronous cluster.
+
```
CyberDem0n, line 337:
```suggestion

It also allows to fail over to asynchronous node is synchronous mode is enabled.
```

docs/patronictl.rst
```diff
@@ -0,0 +1,1963 @@
...
+      [ --group CITUS_GROUP ]
+      [ { -r | --role } { leader | primary | standby-leader | replica | standby | any | master } ]
+      [ --force ]
+
+.. _patronictl_reload_description:
+
+Description
+"""""""""""
+
+``patronictl reload`` requests a reload of local configuration for one or more Patroni members.
```
CyberDem0n, line 1221: It also triggers `pg_ctl reload` even if nothing have been changed.

docs/patronictl.rst
```diff
@@ -0,0 +1,1963 @@
...
+
+Description
+"""""""""""
+
+``patronictl switchover`` performs a switchover in the cluster.
+
+It is designed to be used when the cluster is healthy, e.g.:
+
+- There is a leader; or
+- There are synchronous standbys available in a synchronous cluster.
```
CyberDem0n, line 1634: I read it as `we could have synchronous standbys but no leader`, which is wrong.

CyberDem0n (commented):
Looks super nice!
I left a few comments:
- maybe we can sometimes use different style for code blocks (mainly for patronictl edit-config)
- maybe we should prepend all commands with `$` character to emphasize that this is is supposed to be executed from shell
- for come commands Citus group could be taken from the configuration file

docs/patronictl.rst
barthisrael: I guess so. I thought of adding a deprecation notice. We can remove it later, once we remove this altogether.

docs/patronictl.rst
barthisrael: Removed the "or". I guess that's enough, right?

barthisrael:
I guess I covered all points.
Thanks for your thoughts!

CyberDem0n:
> But the HTML anchor IDs are #synopsis, #description, #parameters, and #examples, respectively.

@barthisrael https://github.com/GeeTransit/sphinx-better-subsection seems to do what you want.
I wonder though if it will introduce some side effects on other parts of docs and how critical they would be.

barthisrael:
> @barthisrael https://github.com/GeeTransit/sphinx-better-subsection seems to do what you want.
> I wonder though if it will introduce some side effects on other parts of docs and how critical they would be.

Ah, you found something :)

Right! And I wonder how much of a problem the described issue would be:

* If someone tries to infere the links for `patronictl dsn` subsections based on the pattern of the other commands, they will reach the head of the document;
* If someone saved the `#synopsis` link for `patronictl dsn` and we introduce a newer command in the future and put it before `patronictl dsn`, the link will go to the new command.

IAC I think the former is not much of an issue, and the second is very unlikely to occur.

Maybe we should just leave it as it is?

CyberDem0n:
The problem is not with `patronictl.rst`, but with the rest of documentation.
If someone was referring to anchors they might stop working. At the moment there are ~80 such places.
From another side if we change titles it also changes anchors and links also break.

barthisrael:
> The problem is not with `patronictl.rst`, but with the rest of documentation. If someone was referring to anchors they might stop working. At the moment there are ~80 such places. From another side if we change titles it also changes anchors and links also break.

Ah, yes, I got it when you mentioned about the extension.

I was just wondering that by keeping the current behavior we might not cause much problems to people consulting patronictl doc page other than what I described above.

docs/patronictl.rst
```diff
@@ -0,0 +1,1975 @@
...
+Patroni has a command-line interface named ``patronictl``, which is used basically to interact with Patroni's REST API and with the DCS. It is intended to make it easier to perform operations in the cluster, and can easily be used by humans or scripts.
+
+.. _patronictl_configuration:
+
+Configuration
+-------------
+
+``patronictl`` uses 3 sections of the configuration:
+
+- **ctl**: how to authenticate against the REST API server, and how to validate the server identity;
```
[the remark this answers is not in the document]
barthisrael: Good point!

hughcapet:
👍

CyberDem0n:
:+1:

PR #2876, opened 2023-09-21
Read GUC's values when joining running Postgres

CyberDem0n:
If restarted in pause Patroni was discarding `synchronous_standby_names` from `postgresql.conf` because in the internal cache this values was set to `None`. As a result synchronous replication transitioned to a broken state, with no synchronous replicas according to the `synchronous_standby_names` and Patroni not selecting/setting the new synchronous replicas (another bug).

To solve the problem of broken initial state and to avoid similar issues with other GUC's we will read GUC's value if Patroni is joining running Postgres.

### DIFF
--- a/patroni/config_generator.py
+++ b/patroni/config_generator.py
@@ -287,7 +287,7 @@ class RunningClusterConfigGenerator(AbstractConfigGenerator):
 
         :param cur: connection cursor to use.
         """
-        cur.execute("SELECT name, current_setting(name) FROM pg_settings "
+        cur.execute("SELECT name, pg_catalog.current_setting(name) FROM pg_catalog.pg_settings "
                     "WHERE context <> 'internal' "
                     "AND source IN ('configuration file', 'command line', 'environment variable') "
                     "AND category <> 'Write-Ahead Log / Recovery Target' "
--- a/patroni/postgresql/config.py
+++ b/patroni/postgresql/config.py
@@ -326,13 +326,23 @@ class ConfigHandler(object):
                                         .format(self._pgpass))
         self._passfile = None
         self._passfile_mtime = None
-        self._synchronous_standby_names = None
         self._postmaster_ctime = None
         self._current_recovery_params: Optional[CaseInsensitiveDict] = None
         self._config = {}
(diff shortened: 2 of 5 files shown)


patroni/config_generator.py
```diff
@@ -287,7 +287,7 @@ def _set_pg_params(self, cur: Union['cursor', 'Cursor[Any]']) -> None:
 
         :param cur: connection cursor to use.
         """
-        cur.execute("SELECT name, current_setting(name) FROM pg_settings "
+        cur.execute("SELECT name, pg_catalog.current_setting(name) FROM pg_catalog.pg_settings "
```
hughcapet, line 290: can we avoid this unrelated change?

CyberDem0n: No, I want to explicitly call `schema.func_name()` and want to reuse mocks :)

patroni/postgresql/config.py
```diff
@@ -326,14 +326,24 @@ def __init__(self, postgresql: 'Postgresql', config: Dict[str, Any]) -> None:
...
+    def load_current_server_parameters(self) -> None:
+        """Read GUC's values from ``pg_settings`` when Patroni is joining the the postgres that is already running."""
+        exclude = [name.lower() for name, value in self.CMDLINE_OPTIONS.items() if value[1] == _false_validator] \
+            + [name.lower() for name in self._RECOVERY_PARAMETERS]
+        self._server_parameters = CaseInsensitiveDict({r[0]: r[1] for r in self._postgresql.query(
+            "SELECT name, pg_catalog.current_setting(name) FROM pg_catalog.pg_settings"
+            " WHERE (source IN ('command line', 'environment variable') OR sourcefile = %s)"
+            " AND pg_catalog.lower(name) != ALL(%s)", self._postgresql_conf, exclude)})
+        self._config.pop('pg_hba', None)
+        self._config.pop('pg_ident', None)
```
CyberDem0n, line 345: I think that I forgot to remove this lines...

hughcapet:
👍

CyberDem0n:
:+1:

PR #2878, opened 2023-09-22
Silence annoying warnings when checking for node uniqueness

CyberDem0n:
WARNING messages are produced by `urllib3` if Patroni is quickly restarted.
Instead we will check that the node is listen on a given port. This fact is actually enough to detect names clashes, while HTTP request could raise an exception is a few other cases, what might case false negatives.

Close https://github.com/zalando/patroni/issues/2881

### DIFF
--- a/patroni/__main__.py
+++ b/patroni/__main__.py
@@ -106,6 +106,8 @@ class Patroni(AbstractPatroniDaemon, Tags):
 
     def ensure_unique_name(self) -> None:
         """A helper method to prevent splitbrain from operator naming error."""
+        from urllib.parse import urlparse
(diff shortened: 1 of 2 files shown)


barthisrael (approved):
Looks good!

CyberDem0n:
@barthisrael actually, after thinking a bit more about it I decided to change the check so it fails as soon as it detects that the port is open.

barthisrael (approved):
> @barthisrael actually, after thinking a bit more about it I decided to change the check so it fails as soon as it detects that the port is open.

Ah, makes sense! Looks better!

rishiraj88:
:+1:

Issue #2879, opened 2023-09-22
Error setting the parameter `wal_keep_size` with unit in Patroni 3.1.1


jorsol:
### What happened?

Error setting the PostgreSQL parameter `wal_keep_size` with unit, this is related to the PR https://github.com/zalando/patroni/pull/2861

Not sure if other parameters with units are affected.

### How can we reproduce it (as minimally and precisely as possible)?

.

### What did you expect to happen?

The parameter is valid but the conversion to int is not correct as is not interpreting the unit.

### Patroni/PostgreSQL/DCS version

-   Patroni version: 3.1.1
-   PostgreSQL version:  15.2
-   DCS (and its version): kubernetes 

### Patroni configuration file

```yaml
.
```

### patronictl show-config

```yaml
.
```

### Patroni log files

```shell
2023-09-22 15:25:51,367 ERROR: Exception when setting dynamic_configuration
Traceback (most recent call last):
  File "/usr/lib/python3.9/site-packages/patroni/config.py", line 312, in set_dynamic_configuration
    self.__effective_configuration = self._build_effective_configuration(configuration,
  File "/usr/lib/python3.9/site-packages/patroni/config.py", line 537, in _build_effective_configuration
    config = self._safe_copy_dynamic_configuration(dynamic_configuration)
  File "/usr/lib/python3.9/site-packages/patroni/config.py", line 359, in _safe_copy_dynamic_configuration
    config['postgresql'][name].update(self._process_postgresql_parameters(value))
  File "/usr/lib/python3.9/site-packages/patroni/config.py", line 345, in _process_postgresql_parameters
    pg_params[name] = int(value) if isinstance(validator, IntValidator) else value
ValueError: invalid literal for int() with base 10: '1536MB'
```

### PostgreSQL log files

```shell
.
```

### Have you tried to use GitHub issue search?

- [X] Yes

PR #2883, opened 2023-09-25
Bugfix for GUC's values with units

CyberDem0n:
Despite being validated by `IntValidator` some GUC's couldn't be casted directly to `int` because they include suffix. Example: `128MB`.

Close https://github.com/zalando/patroni/issues/2879


hughcapet:
👍

CyberDem0n:
:+1:

PR #2886, opened 2023-09-26
Don't call socket functions from tests

hughcapet:
We used to call `socket` module's functions from the config_generator tests to later compare with the output produced by --generate-config. That however sometimes ends up with the whole test module failure if `gethostname()` returned `None`.
Also includes a little code deduplication (`NO_VALUE_MSG` imported directly from the `config_generator` module)

### DIFF
--- a/tests/test_config_generator.py
+++ b/tests/test_config_generator.py
@@ -1,6 +1,5 @@
 import os
 import psutil
-import socket
 import unittest
 
 from . import MockConnect, MockCursor, MockConnectionInfo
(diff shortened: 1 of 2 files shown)


tests/test_config_generator.py
```diff
@@ -9,24 +8,22 @@
...
 @patch('psutil.Process.__init__', Mock(return_value=None))
 class TestGenerateConfig(unittest.TestCase):
-
-    no_value_msg = '#FIXME'
-    _HOSTNAME = socket.gethostname()
-    _IP = sorted(socket.getaddrinfo(_HOSTNAME, 0, socket.AF_UNSPEC, socket.SOCK_STREAM, 0), key=lambda x: x[0])[0][4][0]
+    _HOSTNAME = 'tests_hostname'
+    _IP = '1.9.8.4'
+    setattr(AbstractConfigGenerator, '_HOSTNAME', _HOSTNAME)
+    setattr(AbstractConfigGenerator, '_IP', _IP)
```
CyberDem0n, line 26: Why not use patch.object() for that?
Of course it would require to make _HOSTNAME and _IP global.

hughcapet: right, no reason not to do that:)

CyberDem0n:
:+1:

hughcapet:
👍

PR #2888, opened 2023-09-27
Don't fail if pg_hba/pg_ident contain comment lines

hughcapet:
yaml parser interprets such lines as null and stores it as None into the array of the parsed values, which can not be handled by wrtie() function and crashes the whole bootstrap process.

It is relevant for pg_hba/pg_ident params. Like:
```
postgres:
  pg_hba:
        - #this is a comment
        - local   all             postgres                                peer
```
Even though it is not the proper value, it won't hurt if we just ignore it instead of failing completely.


CyberDem0n:
:+1:

PR #2890, opened 2023-09-28
Speed up dcs failsafe behave tests

CyberDem0n:
- get rid from sleeps
- reduce retry_timeout
- avoid graceful Patroni shut down while DCS is "paused", just kill Patroni and after that gracefully stop postgres
- don't try to delete Pod when Patroni is killed. If K8s API is paused it takes ages

The run time on my laptop is reduced from 2m to 1m28s.


hughcapet:
👍

CyberDem0n:
:+1:

PR #2891, opened 2023-09-28
Further work on permanent physical slots

CyberDem0n:
- Fixed issues with has_permanent_slots() method. It didn't took into account the case of permanent physical slots for members, falsely concluding that there are no permanent slots.
- Write to the status key only LSNs for permanent slots (not just for slots that exist on the primary).
  - Include pg_current_wal_flush_lsn() to slots feedback, so that slots on standby nodes could be advanced
- Improved behave tests:
  - Verify that permanent slots are properly created on standby nodes
  - Verify that permanent slots are properly advanced, including DCS failsafe mode
  - Verify that only permanent slots are written to the `/status`


patroni/ha.py
```diff
@@ -277,7 +277,10 @@ def update_lock(self, write_leader_optime: bool = False) -> bool:
         if write_leader_optime:
             try:
                 last_lsn = self.state_handler.last_operation()
-                slots = self.state_handler.slots()
+                slots = self.cluster.filter_permanent_slots(
+                    {**self.state_handler.slots(), slot_name_from_member_name(self.state_handler.name): last_lsn},
+                    self.is_standby_cluster(),
+                    self.state_handler.major_version)
```
barthisrael, line 283: Maybe we could add a docstring to this method, taking advantage that we are "modifying" it? :)

patroni/ha.py
```diff
@@ -907,7 +910,10 @@ def check_failsafe_topology(self) -> bool:
             'api_url': self.patroni.api.connection_string,
         }
         try:
-            data['slots'] = self.state_handler.slots()
+            data['slots'] = {
+                **self.state_handler.slots(),
+                slot_name_from_member_name(self.state_handler.name): self.state_handler.last_operation()
+            }
```
barthisrael, line 916: Same comment as previous about this.

patroni/dcs/__init__.py
```diff
@@ -994,9 +1001,6 @@ def __permanent_slots(self) -> Dict[str, Union[Dict[str, Any], Any]]:
                 if name in slots:
                     # If primary reported flush LSN for permanent slots we want to enrich our structure with it
                     value['lsn'] = slots[name]
-                elif self.is_physical_slot(value) and name == leader_name and leader and leader.lsn:
-                    # there is no slot on the leader for itself, use `lsn` from the member key.
-                    value['lsn'] = leader.lsn
```
barthisrael, line 999: Am I right in thinking that we are now writing all permanent slots to the `status` key in DCS, including the one for the leader, if it has been configured as a permanent slot?

patroni/ha.py
```diff
@@ -901,6 +918,26 @@ def call_failsafe_member(self, data: Dict[str, Any], member: Member) -> bool:
...
+            * ``conn_url`` - connection URL to the postgres, which is reachable from other nodes;
+            * ``api_url`` - connection URL to Patroni REST API on this node reachable from other nodes;
+            * ``slots`` - a :class:`dict` with replication slots that exist on the leader node, including the primary
+              itself with the last known LSN, because there could be a permanent physical slot on standby nodes.
+
+            Standby nodes are using information from the ``slots`` dict to advance position of permanent
+            replication slots while DCS is not accessible in order to avoid indefinite growth of ``pg_wal``.
+
+        :returns: ``True`` if all members from the ``/failsafe`` topology agree that this node could continue to
+                  run as a ``primary``, or ``False`` if some of standby nodes are not accessible of don't agree.
```
barthisrael, line 939:
```suggestion
                  run as a ``primary``, or ``False`` if some of standby nodes are not accessible or don't agree.
```

barthisrael (approved):
Looks good!

CyberDem0n:
@barthisrael there is still one problem :(
If we have `{tags: {replicatefrom: postgresql1}}` on `postgresql2` and `postgresql2` is declared as a permanent physical slot, then `postgresql0` will create the physical slot for it, but it will never move forward :(
I already have a fix for it (and of course behave tests), but it requires https://github.com/zalando/patroni/pull/2909 to be merged.

features/permanent_slots.feature
```diff
@@ -23,15 +25,21 @@ Feature: permanent slots
...
     And Logical slot test_logical is in sync between postgres0 and postgres2 after 10 seconds
+    And Logical slot test_logical is in sync between postgres0 and postgres3 after 10 seconds
     And postgres1 has a physical replication slot named test_physical after 2 seconds
     And postgres2 has a physical replication slot named test_physical after 2 seconds
+    And postgres3 has a physical replication slot named test_physical after 2 seconds
 
   @slot-advance
   Scenario: check permanent physical slots that match with member names
-    Given postgres1 has a physical replication slot named postgres0 after 2 seconds
+    Given postgres0 has a physical replication slot named postgres3 after 2 seconds
```
barthisrael, line 35: Maybe we could check that `postgres0` has `postgres1` too?

features/permanent_slots.feature
```diff
@@ -42,18 +50,25 @@ Feature: permanent slots
...
     And "status" key in DCS does not have postgres2 in slots
+    And "status" key in DCS has postgres3 in slots
 
   Scenario: check permanent physical replication slot after failover
-    Given I shut down postgres2
+    Given I shut down postgres3
+    And I shut down postgres2
     And I shut down postgres0
     Then postgres1 has a physical replication slot named test_physical after 10 seconds
     Then postgres1 has a physical replication slot named postgres0 after 10 seconds
```
barthisrael, line 74: Maybe we should check that `postgres` has a slot `postgres3` too?

barthisrael:
> @barthisrael there is still one problem :( If we have `{tags: {replicatefrom: postgresql1}}` on `postgresql2` and `postgresql2` is declared as a permanent physical slot, then `postgresql0` will create the physical slot for it, but it will never move forward :( I already have a fix for it (and of course behave tests), but it requires #2909 to be merged.

Nice catch!

Added a couple suggestions regarding behave tests.

A side note: it seems the PR got bloated, and now it touches 40 files. Maybe some issue with rebase or cherry-pick?

features/permanent_slots.feature
CyberDem0n: We are already doing that here: https://github.com/zalando/patroni/pull/2891/files#diff-6dffdf09b22cfc35032e5209953e42980b37c1b696a53598d73f61a1fba1dd1aR13

barthisrael:
I've been testing this locally, and the changes seem to behave well :D

I just think we need to slightly change the documentation of permanent slots to reflect the new behavior, for example:

* This is no longer true "If the permanent slot name matches with the name of the current leader it will not be created. Please note that Patroni does not make checks for permanent slot names added to this configuration matching those that Patroni creates automatically for members.". It actually does create all permanent physical slots except for the one that clashes with the own node name;
* I think we should also add a note that although permanent physical slots are created on all nodes, they are only synched from the primary to the replicas. In other words: if you create a permanent physical slot to use on external applications, make sure to use the slot on the current primary. The only exception to this rule are the physical slots for the members, which are maintained and synched among themselves according to what the primary sets in the status key.

docs/dynamic_configuration.rst
```diff
@@ -48,7 +48,7 @@ In order to change the dynamic configuration you can use either ``patronictl edi
...
+   -  **my\_slot\_name**: the name of the permanent replication slot. If the permanent slot name matches with the name of the current node it will not be created on this node. Please note that Patroni does not make checks for permanent slot names added to this configuration matching those that Patroni creates automatically for members. If those names are added, Patroni will ensure that any slots that were created are not removed even if the member becomes unresponsive, situation which would normally result in the slot's removal by Patroni. Although this can be useful in some situations, such as when importing existing members to a new Patroni cluster (see :ref:`Convert a Standalone to a Patroni Cluster <existing_data>` for details), caution should be exercised by the operator that these clashes in names are not persisted in the DCS due to its effect on normal functioning of Patroni.
```
barthisrael, line 51:
```suggestion
   -  **my\_slot\_name**: the name of the permanent replication slot. If the permanent slot name matches with the name of the current node it will not be created on this node. If you add a permanent physical replication slot which name matches the name of a Patroni member, Patroni will ensure that the slot that was created is not removed even if the corresponding member becomes unresponsive, situation which would normally result in the slot's removal by Patroni. Although this can be useful in some situations, such as when you want replication slots used by members to persist during temporary failures or when importing existing members to a new Patroni cluster (see :ref:`Convert a Standalone to a Patroni Cluster <existing_data>` for details), caution should be exercised by the operator that these clashes in names are not persisted in the DCS, when the slot is no longer required, due to its effect on normal functioning of Patroni.
```

What about this?

barthisrael:
👍

CyberDem0n:
:+1:

XiuhuaRuan:
@CyberDem0n @barthisrael Hi, I have a question regarding the behavior of permanent physical replication slots on standby nodes. Patroni creates physical replication slots on standby nodes and periodically calls pg_replication_slot_advance() on these standby nodes to advance the LSN of these "inactive" slots.
My question is: Is this pg_replication_slot_advance() activity on standby nodes expected, especially if the slots are not actively being consumed for replication on that standby?

CyberDem0n:
@XiuhuaRuan of course these slots on standby are not used by anyone.
They exists for the case when standby will be promoted to the primary.
Consider the following:
- A(primary), with slots: B(active), C(active)
- B(standby), with slots: A(inactive), C(inactive)
- C(standby), with slots A(inactive), B(inactive)

B advances slots A(when LSN on A is moving forward), and C(when restart_lsn on slot C on node A is moving forward).

If B is promoted, node C will be immediately able to connect and use slot B, because it guaranties that WAL is preserved.

PR #2894, opened 2023-10-02
Remove user creation

GuanqunYang193:
remove user creation since it is not patroni's task to manage user creation.


patroni/postgresql/bootstrap.py
```diff
@@ -400,9 +400,9 @@ def post_bootstrap(self, config: Dict[str, Any], task: CriticalTask) -> Optional
 END;$$""".format(f, quote_ident(rewind['username'], postgresql.connection()))
                         postgresql.query(sql)
 
-                for name, value in (config.get('users') or {}).items():
-                    if all(name != a.get('username') for a in (superuser, replication, rewind)):
-                        self.create_or_update_role(name, value.get('password'), value.get('options', []))
+                if config.get('users'):
+                    logger.error('user creation is not be supported after v4.0.0')
+                    raise Exception('user creation is not be supported after v4.0.0')
```
CyberDem0n, line 405: I don't think we need to raise an exception.
```suggestion
```

patroni/postgresql/bootstrap.py
```diff
@@ -400,9 +400,9 @@ def post_bootstrap(self, config: Dict[str, Any], task: CriticalTask) -> Optional
 END;$$""".format(f, quote_ident(rewind['username'], postgresql.connection()))
                         postgresql.query(sql)
 
-                for name, value in (config.get('users') or {}).items():
-                    if all(name != a.get('username') for a in (superuser, replication, rewind)):
-                        self.create_or_update_role(name, value.get('password'), value.get('options', []))
+                if config.get('users'):
+                    logger.error('user creation is not be supported after v4.0.0')
```
CyberDem0n, line 404:
```suggestion
                    logger.error('User creation via "bootstrap.users" is not supported since v4.0.0')
                    logger.info('Please use "boostrap.post_bootstrap" script to create users.')
```

patroni/postgresql/bootstrap.py
```diff
@@ -400,9 +400,9 @@ def post_bootstrap(self, config: Dict[str, Any], task: CriticalTask) -> Optional
 END;$$""".format(f, quote_ident(rewind['username'], postgresql.connection()))
                         postgresql.query(sql)
 
-                for name, value in (config.get('users') or {}).items():
-                    if all(name != a.get('username') for a in (superuser, replication, rewind)):
-                        self.create_or_update_role(name, value.get('password'), value.get('options', []))
+                if config.get('users'):
```
CyberDem0n, line 403: When the feature is removed we probably better complain only if there are any users different from superuser, replication, rewind:
```suggestion
                if any(user not in (superuser, replication, rewind) for user in (config.get('users') or {}).keys()):
```

CyberDem0n (commented):
I wonder if we should remove `bootstrap.users` from sample config files and from the documentation even before merging this PR.
Docs:
1. https://github.com/zalando/patroni/blob/60d8bc3a70e95a183da9bba1917062642a0171f8/docs/ENVIRONMENT.rst?plain=1#L27-L34
2. https://github.com/zalando/patroni/blob/60d8bc3a70e95a183da9bba1917062642a0171f8/docs/yaml_configuration.rst?plain=1#L57-L68

Besides that, there is also code that handles ENV variables: https://github.com/zalando/patroni/blob/60d8bc3a70e95a183da9bba1917062642a0171f8/patroni/config.py#L814-L829

GuanqunYang193:
> I wonder if we should remove `bootstrap.users` from sample config files and from the documentation even before merging this PR. Docs:
> 
> 1. https://github.com/zalando/patroni/blob/60d8bc3a70e95a183da9bba1917062642a0171f8/docs/ENVIRONMENT.rst?plain=1#L27-L34
> 2. https://github.com/zalando/patroni/blob/60d8bc3a70e95a183da9bba1917062642a0171f8/docs/yaml_configuration.rst?plain=1#L57-L68
> 
> Besides that, there is also code that handles ENV variables:
> 
> https://github.com/zalando/patroni/blob/60d8bc3a70e95a183da9bba1917062642a0171f8/patroni/config.py#L814-L829

That's a good idea, I will submit a separate PR for the doc first.

patroni/postgresql/bootstrap.py
GuanqunYang193: can an exception remind the user of Patroni that their Postgres users are not created?

patroni/postgresql/bootstrap.py
GuanqunYang193: We still need to prevent some users from initializing those three users using the 'users' option. They might provide other initialization options, which will be ignored after we have removed this feature.

GuanqunYang193: For example:
replication:
    options:
          - createdb (which will be ignored but not complained)

CyberDem0n: These three users are already created earlier.  I am not suggesting to create them once again (it didn't happen before). My suggestion it to show not error if there are no other users.

Also, probably I am overthinking it.

patroni/postgresql/bootstrap.py
CyberDem0n: This exception may interrupt a log running custom bootstrap (for example restoring 10TB database from the backup).

GuanqunYang193: make sense

patroni/postgresql/bootstrap.py
GuanqunYang193: got you, let's complain everything about deleted "users" options when people tries to use it.

hughcapet (approved):

PR #2895, opened 2023-10-04
Take IP from the pod if `kubernetes.pod_ip` is missing

CyberDem0n:
It used to work before #2652

Besides that fix a couple of more problems:
- make sure `_patch_or_create()` method isn't instantiating the `k8s_client.V1ConfigMap` object instead of `k8s_client.V1Endpoints` for non leader endpoints. The only reason it worked is that the JSON serialization for both object types is the same and doesn't include the object type name.
- `attempt_to_acquire_leader()` should immediately put the IP address of the primary to the leader endpoint. It didn't happen because of the oversight in the https://github.com/zalando/patroni/pull/1820.


Jan-M:
:+1:

hughcapet:
I think we need to document the subsets definition logic in the user docs. i.e. fix the `kubernetes.pod_ip` description

hughcapet:
👍

CyberDem0n:
:+1:

PR #2896, opened 2023-10-06
Ignore synchronous_mode setting in a standby cluster

hughcapet:
is_synchronous_mode() should always return False in standby clusters


CyberDem0n:
:+1:

hughcapet:
👍

PR #2900, opened 2023-10-06
Bugfix: standby cluster switchover

CyberDem0n:
1.  Enforce `_load_cluster()` after acquisition for the leader lock in ZooKeeper. Sometimes the notification from ZooKeeper was arriving too late and Patroni wasn't setting the `role=standby_leader`.

2. The `_get_node_to_follow()` method was falsely assuming that we still own the leader lock and returning the remote node instead of the new standby leader. While not a big issue per se, because the next HA loop usually fixes it, such behavior was causing flakiness of behave tests with Postgres 12 and older, where restart is required to update `primary_conninfo` GUC.

### DIFF
--- a/patroni/ha.py
+++ b/patroni/ha.py
@@ -586,7 +586,7 @@ class Ha(object):
         # the remote member, except when there is no standby leader in pause.
         if self.is_standby_cluster() \
                 and (cluster.leader and cluster.leader.name and cluster.leader.name == self.state_handler.name
-                     or self.cluster.is_unlocked() and not self.is_paused()):
+                     or cluster.is_unlocked() and not self.is_paused()):
             node_to_follow = self.get_remote_member()
         # If replicatefrom tag is set, try to follow the node mentioned there, otherwise, follow the leader.
         elif self.patroni.replicatefrom and self.patroni.replicatefrom != self.state_handler.name:
--- a/patroni/ha.py
+++ b/patroni/ha.py
@@ -584,7 +584,9 @@ class Ha(object):
         """
         # The standby leader or when there is no standby leader we want to follow
         # the remote member, except when there is no standby leader in pause.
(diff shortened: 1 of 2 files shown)


hughcapet:
The [test](https://github.com/zalando/patroni/actions/runs/6432349470/job/17467019075) fails with
```
And there is a postgres2_cb.log with "on_start replica batman1\non_role_change standby_leader batman1" in postgres2 data directory # features/steps/cascading_replication.py:12 0.000s
Assertion Failed: "on_start replica batman1\non_restart replica batman1\non_role_change standby_leader batman1" in postgres2_cb.log doesn't contain on_start replica batman1\non_role_change standby_leader batman1
```
after a switchover from 1 to 2

it means the on_restart hook was eventually fired on the newly promoted standby leader (2), not the old leader, isn't it? So I am not really getting how this fixes the problem in that run.

patroni/ha.py
```diff
@@ -584,7 +584,9 @@ def _get_node_to_follow(self, cluster: Cluster) -> Union[Leader, Member, None]:
         """
         # The standby leader or when there is no standby leader we want to follow
         # the remote member, except when there is no standby leader in pause.
-        if self.is_standby_cluster() and (self.has_lock(False) or self.cluster.is_unlocked() and not self.is_paused()):
+        if self.is_standby_cluster() \
+                and (cluster.leader and cluster.leader.name and cluster.leader.name == self.state_handler.name
```
hughcapet, line 588: Btw, why do we use `cluster.leader` here but `self.cluster.is_unlocked()` immediately after that ?

CyberDem0n: Addressed in https://github.com/zalando/patroni/pull/2900/commits/47e590dc153f0078203d9da57e6df8894575e731

hughcapet:
👍

CyberDem0n:
:+1:

PR #2902, opened 2023-10-09
Make it easier to debug callbacks

CyberDem0n:
1. Introduce DEBUG logs for callbacks
2. Configure log format in behave tests to include filename, line, and method name that triggered the callback and enable DEBUG logs for `patroni.postgresql.callback_executor` module.

P.S. unfortunately it works only starting from python 3.8, but it should be good enough for debug purpose because 3.7 is already EOL.


hughcapet:
👍

CyberDem0n:
:+1:

PR #2903, opened 2023-10-10
Set citus.local_hostname

CyberDem0n:
There are cases when Citus wants to have a connection to the local postgres. By default it uses `localhost` for that, which is not alwasy available. To solve it we will set `citus.local_hostname` GUC to custom value, which is the same as Patroni uses to connect to Postgres.

### DIFF
--- a/docs/citus.rst
+++ b/docs/citus.rst
@@ -38,14 +38,18 @@ After that you just need to start Patroni and it will handle the rest:
 2. If ``max_prepared_transactions`` isn't explicitly set in the global
    :ref:`dynamic configuration <dynamic_configuration>` Patroni will
    automatically set it to ``2*max_connections``.
-3. The ``citus.database`` will be automatically created followed by ``CREATE EXTENSION citus``.
-4. Current superuser :ref:`credentials <postgresql_settings>` will be added to the ``pg_dist_authinfo``
+3. Will adjust ``citus.local_hostname`` GUC value from ``localhost`` to the
+   value that Patroni is using in order to connect to the local PostgreSQL
+   instance. The value sometimes should be different from the ``localhost``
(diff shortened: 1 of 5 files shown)


patroni/postgresql/citus.py
```diff
@@ -403,6 +403,9 @@ def adjust_postgres_gucs(self, parameters: Dict[str, Any]) -> None:
         # Resharding in Citus implemented using logical replication
         parameters['wal_level'] = 'logical'
 
+        # Sometimes Citus needs to connect to the local postgres. We will do it the same way as Patroni does.
+        parameters['citus.local_hostname'] = self._postgresql.connection_pool.conn_kwargs.get('host', 'localhost')
```
hughcapet, line 407: shouldn't we mention this guc value definition logic in "Citus support"?

docs/citus.rst
```diff
@@ -38,14 +38,18 @@ After that you just need to start Patroni and it will handle the rest:
 2. If ``max_prepared_transactions`` isn't explicitly set in the global
    :ref:`dynamic configuration <dynamic_configuration>` Patroni will
    automatically set it to ``2*max_connections``.
-3. The ``citus.database`` will be automatically created followed by ``CREATE EXTENSION citus``.
-4. Current superuser :ref:`credentials <postgresql_settings>` will be added to the ``pg_dist_authinfo``
+3. Will adjust ``citus.local_hostname`` GUC value from ``localhost`` to the
```
hughcapet, line 41:
```suggestion
3.  The ``citus.local_hostname`` GUC value will be adjusted from ``localhost`` to the
```

hughcapet:
👍

CyberDem0n:
:+1:

PR #2907, opened 2023-10-11
Allow to specify psycopg* in extras and switch to `build`

CyberDem0n:
* remove check_psycopg() call from the setup.py, when installing from wheel it doesn't work anyway.
* call check_psycopg() function before process_arguments(), because the last one is trying to import psycopg and fails with the stacktrace, while the first one shows a nice human-readable error message.
* add psycopg2, psycopg2-binary, and psycopg3 extras, that will install psycopg2>=2.5.4, psycopg2-binary, or psycopg[binary]>=3.0.0 modules respectively.
* move check_psycopg() function to the __main__.py.
* introduce the new extra called `all`, it will allow to install all dependencies at once (except psycopg related).
* use the `build` module in order to create sdist bdist_wheel packages.
* update the documentation regarding psycopg and extras (dependencies).

### DIFF
--- a/docs/installation.rst
+++ b/docs/installation.rst
@@ -30,23 +30,10 @@ There are a few options available:
     sudo apt-get install python3-psycopg2  # install psycopg2 module on Debian/Ubuntu
     sudo yum install python3-psycopg2      # install psycopg2 on RedHat/Fedora/CentOS
 
-2. Install psycopg2 from the binary package
+2. Specify one of `psycopg`, `psycopg2`, or `psycopg2-binary` in the :ref:`list of dependencies <extras>` when installing Patroni with pip.
 
-.. code-block:: shell
 
-    pip install psycopg2-binary
-
-3. Install psycopg2 from source
-
-.. code-block:: shell
-
-    pip install psycopg2>=2.5.4
-
-4. Use psycopg 3.0 instead of psycopg2
-
-.. code-block:: shell
-
-    pip install psycopg[binary]>=3.0.0
+.. _extras:
 
 General installation for pip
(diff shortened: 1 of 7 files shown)


docs/installation.rst
```diff
@@ -73,12 +60,20 @@ raft
     `pysyncobj` module in order to use python Raft implementation as DCS
 aws
     `boto3` in order to use AWS callbacks
+all
+    all of the above (except psycopg family)
+psycopg
+    `psycopg[binary]>=3.0.0` module
```
barthisrael, line 66: This line differs from the corresponding one in `README.rst`.
I guess we should adjust either of them, probably in `README.rst` given the commit message.

setup.py
```diff
@@ -139,15 +145,22 @@ def setup_package(version):
                     deps[i] = r
                     EXTRAS_REQUIRE[e] = deps
                     extra = True
-                    break
-            if extra:
-                break
         if not extra:
             install_requires.append(r)
 
+    # Just for convinience, if someone wants to install dependencies for all extras
```
barthisrael, line 151:
```suggestion
    # Just for convenience, if someone wants to install dependencies for all extras
```

barthisrael (approved):
Looks good!
I only have a couple small pointers.

docs/installation.rst
barthisrael: The code in `setup.py` and the psycopg3 docs page also suggest the one to be fixed is `README.rst`.

barthisrael:
👍

CyberDem0n:
:+1:

PR #2909, opened 2023-10-12
Don't cache anything in Zookeeper implementation

CyberDem0n:
Cache creates a lot of problems and prevents implementing a feature of automatic retention of physical replication slots for members with configurable retention policy.

Just read the entire cluster from Zookeeper instead and use watchers only for the `/leader` and `/config` keys.


barthisrael (approved):
I don't have knowledge on ZooKeeper, but the code changes seem aligned with the PR proposal and make sense to me.

barthisrael:
👍

CyberDem0n:
:+1:

PR #2911, opened 2023-10-13
Authenticate and retry when etcd returns "revision of auth store is old" error

kennydo:
From https://etcd.io/docs/v3.5/learning/design-auth-v3/#notes-on-the-implementation-of-authenticate-rpc, when etcd creates an auth token, it has an associated auth store revision number. If the auth store changes (ex: a new user is added), then the auth store revision is updated. Attempts to use auth tokens with old auth store revisions return an error (https://github.com/etcd-io/etcd/blob/6d68ab092d30e539308e6c6d9e286b3cd81a8f04/server/auth/store.go#L874).

# Current behavior

Currently, the etcd client does not re-authenticate if it gets an error about the auth revision being old.

To replicate:
1. Start an etcd server:
```bash
echo "my private key" > priv_key.txt
etcd \
  --log-level debug \
  --auth-token 'jwt,sign-method=HS512,priv-key=priv_key.txt' \
  --auth-token-ttl 999999999
```
2. Enable auth:
```bash
etcdctl role add root
etcdctl user add root --new-user-password="rootpw"
etcdctl user grant-role root root

etcdctl role add role1
etcdctl role grant-permission role1 readwrite --prefix=true /
etcdctl user add user1 --new-user-password "user1pw"
etcdctl user grant-role user1 role1

etcdctl auth enable
```
3. Create an `Etcd3Client` and verify it can `put`:
```bash
$ venv/bin/python
Python 3.11.6 (main, Oct  2 2023, 13:45:54) [Clang 15.0.0 (clang-1500.0.40.1)] on darwin
>>> from patroni.dcs import etcd3
>>> from patroni.dcs.etcd import DnsCachingResolver
>>> client = etcd3.Etcd3Client({
...   'host': '127.0.0.1',
...   'port': 2379,
...   'username': 'user1',
...   'password': 'user1pw',
...   'retry_timeout': 5,
...   'version_prefix': '/v3',
... }, DnsCachingResolver())
>>> client.put('/patronikey', 'patronivalue')
{'header': {'cluster_id': '14841639068965178418', 'member_id': '10276657743932975437', 'revision': '2', 'raft_term': '2'}}
```
4. Add a new etcd user
```bash
etcdctl user --user root --password rootpw add user2 --new-user-password "user2pw"
```
5. Attempting to `put` again throws an error:
```python
>>> client.put('/patronikey', 'patronivalue')
  File "/Users/kedo/Workspace/kennydo/patroni/patroni/dcs/etcd3.py", line 198, in wrapper
    return self.handle_auth_errors(func, *args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/kedo/Workspace/kennydo/patroni/patroni/dcs/etcd3.py", line 315, in handle_auth_errors
    return func(self, *args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/kedo/Workspace/kennydo/patroni/patroni/dcs/etcd3.py", line 364, in put
    return self.call_rpc('/kv/put', fields, retry)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/kedo/Workspace/kennydo/patroni/patroni/dcs/etcd3.py", line 283, in call_rpc
    return self.api_execute(self.version_prefix + method, self._MPOST, fields)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/kedo/Workspace/kennydo/patroni/patroni/dcs/etcd.py", line 283, in api_execute
    return self._handle_server_response(response)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/kedo/Workspace/kennydo/patroni/patroni/dcs/etcd3.py", line 241, in _handle_server_response
    raise _raise_for_data(ret or data, response.status)
patroni.dcs.etcd3.InvalidArgument: <InvalidArgument error: 'etcdserver: revision of auth store is old', code: 3>
```

I expected the etcd3 client to re-authenticate and retry.

# This change

This change makes it so that the etcd client re-authenticates and retries if it receives the old auth store revision error.

Between these two `put`s, I added a new user. It logs about the old auth store revision and then retries:
```
>>> client.put('/patronikey', 'patronivalue')
{'header': {'cluster_id': '14841639068965178418', 'member_id': '10276657743932975437', 'revision': '2', 'raft_term': '2'}}

>>> client.put('/patronikey', 'patronivalue')
Auth token is for old revision of auth store
{'header': {'cluster_id': '14841639068965178418', 'member_id': '10276657743932975437', 'revision': '3', 'raft_term': '2'}}
```

### DIFF
--- a/patroni/dcs/etcd3.py
+++ b/patroni/dcs/etcd3.py
@@ -138,6 +138,10 @@ class InvalidAuthToken(Etcd3ClientError):
     error = "etcdserver: invalid auth token"
 
 
+class RevisionOfAuthStoreIsOld(InvalidArgument):
+    error = "etcdserver: revision of auth store is old"
+
+
 errStringToClientError = {getattr(s, 'error'): s for s in Etcd3ClientError.get_subclasses() if hasattr(s, 'error')}
 errCodeToClientError = {getattr(s, 'code'): s for s in Etcd3ClientError.__subclasses__()}
 
@@ -320,6 +324,9 @@ class Etcd3Client(AbstractEtcdClientWithFailover):
                 raise UnsupportedEtcdVersion('Authentication is required by Etcd cluster but not '
                                              'supported on version lower than 3.3.0. Cluster version: '
                                              '{0}'.format('.'.join(map(str, self._cluster_version))))
+            return retry(e)
+        except RevisionOfAuthStoreIsOld as e:
+            logger.error('Auth token is for old revision of auth store')
             return retry(e)
         except InvalidAuthToken as e:
             logger.error('Invalid auth token: %s', self._token)
--- a/patroni/dcs/etcd3.py
+++ b/patroni/dcs/etcd3.py
@@ -199,7 +199,7 @@ def build_range_request(key: str, range_end: Union[bytes, str, None] = None) -> 
 
 def _handle_auth_errors(func: Callable[..., Any]) -> Any:
     def wrapper(self: 'Etcd3Client', *args: Any, **kwargs: Any) -> Any:
-        return self.handle_auth_errors(func, *args, **kwargs)
+        return self.create_auth_retry()(lambda: self.handle_auth_errors(func, *args, **kwargs))
     return wrapper
 
 
@@ -306,8 +306,11 @@ class Etcd3Client(AbstractEtcdClientWithFailover):
             self._token = response.get('token')
         return old_token != self._token
 
+    def create_auth_retry(self) -> Retry:
+        return Retry(max_tries=-1, retry_exceptions=(RevisionOfAuthStoreIsOld,))
+
     def handle_auth_errors(self: 'Etcd3Client', func: Callable[..., Any], *args: Any, **kwargs: Any) -> Any:
-        def retry(ex: Exception) -> Any:
+        def authenticate_and_retry_once(ex: Exception) -> Any:
             if self.username and self.password:
                 self.authenticate()
                 return func(self, *args, **kwargs)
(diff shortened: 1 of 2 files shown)


kennydo:
On further thought, I realized that we actually need to retry more than once.

A common procedure is making a new etcd role, granting the role permissions, making a new user, and then granting the role to the user.

There could be an unfortunate timeline where:
T=0 patroni authenticates and gets a valid auth token
T=1 admin creates a new role
T=2 patroni tries to do an operation, but it fails due to the auth revision being old. It authenticates and gets a valid auth token
T=3 admin grants the new role some permissions
T=4 patroni tries again to do an operation, but it fails due to the auth revision being old. It authenticates and gets a valid auth token
T=5 admin creates a new user
T=6 patroni tries again to do an operation, but it fails due to the auth revision being old. It authenticates and gets a valid auth token
...and so on

Unlike the other errors in this `handle_auth_errors` where retrying once is fine, the "revision of auth store is old" error might need to be retried multiple times.

patroni/dcs/etcd3.py
```diff
@@ -302,8 +306,11 @@ def authenticate(self) -> bool:
             self._token = response.get('token')
         return old_token != self._token
 
+    def create_auth_retry(self) -> Retry:
+        return Retry(max_tries=-1, retry_exceptions=(RevisionOfAuthStoreIsOld,))
```
CyberDem0n, line 310: Thanks for detailed description of the issue! 
Unfortunately we can't afford infinite retries, and there are chances of it to happen if someone will modify users with the high rate.

In the `kwargs` of the `handle_auth_errors()` method there could be a `retry` object, that should be used:
1. If there is no `retry` argument we can just give up immediately and don't perform reauthentication at all. The retry will happen naturally, on the next hear beat cycle.
2. If there is a `retry` object, it should be used to make sure that the total execution time doesn't exceed the `Retry.deadline` (e.g. `retry_timeout`).

Also, as of now the successful `authenticate()` method restarts a watcher thread, but in case of `RevisionOfAuthStoreIsOld` we don't have to do it, because it continues to work normally.

I'll try to come up with the proper fix and retry policies today.

CyberDem0n: Something like this: https://github.com/zalando/patroni/pull/2913/commits/47628a7f1d8a35e51a147911a6bc7a8ccdc84af0

kennydo: Yea, that makes sense. Thank you for making this fix! I'll close this PR since https://github.com/zalando/patroni/pull/2913 exists then.

PR #2912, opened 2023-10-13
Update README.md to reflect changes in etcd v3

anli-xsigns:
In etcdctl v3 the ls command isn't present anymore, it has to be changed to etcdctl get --keys-only --prefix (as already used in the citus paragraph)


barthisrael:
👍

PR #2913, opened 2023-10-13
Handle AuthOldRevision error

CyberDem0n:
If retries are requested - try to get a new new token and repeat the request. Repeat it in a loop until request is successfully executed or until `retry_timeout` is exhausted. This is the only way of solving a race condition, because between authentication and executing the request yet another modification of the user database in Etcd might happen.

In case if the request doesn't have to be immediately retried - set a flag that the next API request should perform the authentication first and let Patroni to naturally repeat the request on the next heartbeat
loop.

Co-authored-by: Kenny Do <kedo@render.com>


patroni/dcs/etcd3.py
barthisrael, line 1: I can see you turned `retry` into a keyword-only argument in all methods, except for `call_rpc`. Any reason for that?

patroni/dcs/etcd3.py
```diff
@@ -193,6 +197,12 @@ def build_range_request(key: str, range_end: Union[bytes, str, None] = None) ->
     return fields
 
 
+class ReAuthenticateReason(IntEnum):
```
barthisrael, line 200: I wonder if we should call it "mode" instead of "reason"?

barthisrael (approved):
The PR looks good. Posted a couple comments.

kennydo:
I think I can still reproduce the behavior where the old revision bubbles up as an exception and doesn't get retried:

Here's my repro steps:
1. Start the etcd server (same as in https://github.com/zalando/patroni/pull/2911#issue-1941011436)
2. Enable auth (same as in above)
3. Create an etcd client and verify it can put:
```python
In [1]: from patroni.dcs import etcd3

In [2]: from patroni.dcs.etcd import DnsCachingResolver

In [3]: client = etcd3.Etcd3Client({'host': '127.0.0.1', 'port': 2379, 'username': 'user1', 'password': 'user1pw', 'retry_timeout': 5, 'version_prefix': '/v3'}, DnsCachingResolver())

In [4]: from patroni.utils import Retry

In [5]: retry = Retry(deadline=10, max_delay=1, max_tries=-1)

In [6]: client.put('/patronikey', 'patronivalue', retry=retry.copy())
Out[6]:
{'header': {'cluster_id': '14841639068965178418',
  'member_id': '10276657743932975437',
  'revision': '2',
  'raft_term': '2'}}
```
4. Add a new user (same command as in the other PR)
5. When I try to `put` again, even with retry enabled, it still throws an exception:
```python
In [7]: client.put('/patronikey', 'patronivalue', retry=retry.copy())
Auth token is for old revision of auth store
---------------------------------------------------------------------------
AuthOldRevision                           Traceback (most recent call last)
Cell In[7], line 1
----> 1 client.put('/patronikey', 'patronivalue', retry=retry.copy())

File ~/Workspace/kennydo/patroni/patroni/dcs/etcd3.py:208, in _handle_auth_errors.<locals>.wrapper(self, *args, **kwargs)
    207 def wrapper(self: 'Etcd3Client', *args: Any, **kwargs: Any) -> Any:
--> 208     return self.handle_auth_errors(func, *args, **kwargs)

File ~/Workspace/kennydo/patroni/patroni/dcs/etcd3.py:353, in Etcd3Client.handle_auth_errors(self, func, retry, *args, **kwargs)
    351 if not retry:
    352     raise exc
--> 353 retry.ensure_deadline(0.5, exc)

File ~/Workspace/kennydo/patroni/patroni/utils.py:581, in Retry.ensure_deadline(self, timeout, raise_ex)
    579 if self.deadline < timeout:
    580     if raise_ex:
--> 581         raise raise_ex
    582     return False
    583 return True

File ~/Workspace/kennydo/patroni/patroni/dcs/etcd3.py:334, in Etcd3Client.handle_auth_errors(self, func, retry, *args, **kwargs)
    331         raise exc or Etcd3Exception(msg)
    333 try:
--> 334     return func(self, *args, retry=retry, **kwargs)
    335 except (UserEmpty, PermissionDenied) as e:  # no token provided
    336     # PermissionDenied is raised on 3.0 and 3.1
    337     if self._cluster_version < (3, 3) and (not isinstance(e, PermissionDenied)
    338                                            or self._cluster_version < (3, 2)):

File ~/Workspace/kennydo/patroni/patroni/dcs/etcd3.py:392, in Etcd3Client.put(self, key, value, lease, create_revision, mod_revision, retry)
    390     compare = {'target': 'MOD', 'mod_revision': mod_revision}
    391 else:
--> 392     return self.call_rpc('/kv/put', fields, retry)
    393 compare['key'] = fields['key']
    394 return self.txn(compare, {'request_put': fields}, retry=retry)

File ~/Workspace/kennydo/patroni/patroni/dcs/etcd3.py:294, in Etcd3Client.call_rpc(self, method, fields, retry)
    292 def call_rpc(self, method: str, fields: Dict[str, Any], retry: Optional[Retry] = None) -> Dict[str, Any]:
    293     fields['retry'] = retry
--> 294     return self.api_execute(self.version_prefix + method, self._MPOST, fields)

File ~/Workspace/kennydo/patroni/patroni/dcs/etcd.py:283, in AbstractEtcdClientWithFailover.api_execute(self, path, method, params, timeout)
    281 try:
    282     response = self._do_http_request(retry, machines_cache, request_executor, method, path, **kwargs)
--> 283     return self._handle_server_response(response)
    284 except etcd.EtcdWatchTimedOut:
    285     raise

File ~/Workspace/kennydo/patroni/patroni/dcs/etcd3.py:252, in Etcd3Client._handle_server_response(self, response)
    250         raise etcd.EtcdException('Server response was not valid JSON: %r' % e)
    251     ret = {}
--> 252 raise _raise_for_data(ret or data, response.status)

AuthOldRevision: <AuthOldRevision error: 'etcdserver: revision of auth store is old', code: 3>
```

CyberDem0n:
@kennydo that's because you need to do it slightly different:
```python
retry(client.put, '/patronikey', 'patronivalue', retry=retry)
```

patroni/dcs/etcd3.py
CyberDem0n: `call_rpc` is not supposed to be decorated.

barthisrael (approved):

barthisrael:
👍

CyberDem0n:
:+1:

PR #2917, opened 2023-10-17
Attempt to reduce flakiness for recovery behave test on K8s

CyberDem0n:
- wait until Postgres is properly started after the first crash before changing `primary_start_timeout` and killing it once again.


hughcapet:
👍

CyberDem0n:
:+1:

PR #2922, opened 2023-10-19
get rid of stale comment of get_cluster

zhjwpku:
PR #2909 remove the cache in Zookeeper implementation of DCS, so the comment of get_cluster should be changed to 'Retrieve a fresh view of DCS' since every implementation does so.


CyberDem0n:
:+1:

hughcapet:
👍

PR #2925, opened 2023-10-23
Add warning to the sample config about bootstrap section

CyberDem0n:
it often people are trying to change it and coming with the questions why it doesn't work.


hughcapet:
👍

CyberDem0n:
:+1:

barthisrael:
I wonder if we should also add such a warning when generating a sample config file through the feature that @hughcapet developed?

hughcapet:
seems adding comment lines with pyyaml is not straightforward enough to do it right now:)

PR #2926, opened 2023-10-23
Add capability of specifying namespace through `--dcs` argument

barthisrael:
This commit changes the `patronictl` application in such a way its `--dcs` argument is now able to receive a namespace.

Previous to this commit this was the format of that argument's value: `DCS://HOST:PORT`.

From now on it accepts this format: `DCS://HORT:PORT/NAMESPACE`. As all previous parts of the argument value, `NAMESPACE` is optional, and if not given `patronictl` will fallback to the value from the configuration file, if any, or to `service`.

This change is specifically useful when you are running a cluster in a custom namespace, and from a machine where you don't have a configuration file for Patroni or `patronictl`. It can avoid that you would have to create a configuration file only with `namespace` filed in that case.

References: PAT-219.

Issue reported by: Shaun Thomas <shaun@bonesmoses.org>

### DIFF
--- a/docs/patronictl.rst
+++ b/docs/patronictl.rst
@@ -52,9 +52,9 @@ Before jumping into each of the sub-commands of ``patronictl``, be aware that ``
 ``-d`` / ``--dcs-url`` / ``--dcs``
     Provide a connection string to the DCS used by Patroni.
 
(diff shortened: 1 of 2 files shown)


CyberDem0n:
:+1:

hughcapet:
👍

PR #2927, opened 2023-10-24
Add support for additional parameters on custom bootstrap

barthisrael:
Previous to this commit, if a user would ever like to add parameters to the custom bootstrap script call, they would need to configure Patroni like this:

```
bootstrap:
  method: custom_method_name
  custom_method_name:
    command: /path/to/my/custom_script --arg1=value1 --arg2=value2 ...
```

This commit extends that so we achieve a similar behavior that is seen when using `create_replica_methods`, i.e., we also allow the following syntax:

```
bootstrap:
  method: custom_method_name
  custom_method_name:
    command: /path/to/my/custom_script
    arg1: value1
    arg2: value2
```

All keys in the mapping which are not recognized by Patroni, will be dealt with as if they were additional named arguments to be passed down to the `command` call.

References: PAT-218.

### DIFF
--- a/features/backup_restore.py
+++ b/features/backup_restore.py
@@ -6,6 +6,7 @@ if __name__ == "__main__":
     parser = argparse.ArgumentParser()
     parser.add_argument("--datadir", required=True)
     parser.add_argument("--sourcedir", required=True)
+    parser.add_argument("--test-argument", required=True)
     args, _ = parser.parse_known_args()
 
     shutil.copytree(args.sourcedir, args.datadir)
--- a/features/environment.py
+++ b/features/environment.py
@@ -892,12 +892,12 @@ class PatroniPoolController(object):
         }
         self.start(to_name, custom_config=custom_config)
 
-    def backup_restore_config(self, params={}):
+    def backup_restore_config(self, params=None):
         return {
             'command': (self.BACKUP_RESTORE_SCRIPT
                         + ' --sourcedir=' + os.path.join(self.patroni_path, 'data', 'basebackup')).replace('\\', '/'),
             'test-argument': 'test-value',  # test config mapping approach on custom bootstrap/replica creation
-            **params,
+            **(params or {}),
         }
 
     def bootstrap_from_backup(self, name, cluster_name):
--- a/features/environment.py
+++ b/features/environment.py
@@ -892,14 +892,19 @@ class PatroniPoolController(object):
         }
         self.start(to_name, custom_config=custom_config)
 
+    def backup_restore_config(self, params={}):
+        return {
+            'command': (self.BACKUP_RESTORE_SCRIPT + ' --sourcedir=' + os.path.join(self.patroni_path, 'data', 'basebackup')).replace('\\', '/'),
+            'test-argument': 'test-value', # only used for testing config mapping approach on custom bootstrap/replica creation
+            **params,
+        }
+
     def bootstrap_from_backup(self, name, cluster_name):
         custom_config = {
             'scope': cluster_name,
             'bootstrap': {
                 'method': 'backup_restore',
-                'backup_restore': {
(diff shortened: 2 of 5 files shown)


features/backup_restore.py
```diff
@@ -6,6 +6,7 @@
     parser = argparse.ArgumentParser()
     parser.add_argument("--datadir", required=True)
     parser.add_argument("--sourcedir", required=True)
+    parser.add_argument("--test-argument", required=True)
```
CyberDem0n, line 9: It is also used for custom replica creation: https://github.com/zalando/patroni/blob/f5ee67fa1c683cacfcc398375e1c55d6a8839c21/features/environment.py#L930-L935
This is the reason why behave tests are failing.

barthisrael: 🤦 I didn't notice the other behave tests were failing. Thanks for the heads up!

CyberDem0n: Standby cluster behave tests are still failing. In logs I've found the following error: 
```
usage: backup_restore.py [-h] --datadir DATADIR --sourcedir SOURCEDIR
                         --test-argument TEST_ARGUMENT
backup_restore.py: error: the following arguments are required: --test-argument
2023-10-24 16:05:36,677 ERROR [/home/runner/work/patroni/patroni/patroni/postgresql/bootstrap.py:317 - create_replica]: Error creating replica using method backup_restore: /opt/hostedtoolcache/Python/3.7.17/x64/bin/python features/backup_restore.py --sourcedir=/home/runner/work/patroni/patroni/data/basebackup exited with code=2
```
There is at least one more place left: https://github.com/zalando/patroni/blob/f5ee67fa1c683cacfcc398375e1c55d6a8839c21/features/steps/standby_cluster.py#L18-L20

From the first look in all three places there is exactly the same thing happening, the `command` is built from from some class variables in `PatroniPoolController` and from `PatroniPoolController.patroni_path`. But, right now I'm not sure why it is done a bit differently in two places. It could be related to some Windows quirks (need to double check). Although... the thing it does is `.replace('\\', '/')`, which is hard to screw up.

IMO, we should create a new method that will return the entire dict.
something like:
```python
def backup_restore_config(self, **kwargs):
  return {
    'command': (self.BACKUP_RESTORE_SCRIPT + ' --sourcedir=' + os.path.join(self.patroni_path, 'data', 'basebackup')).replace('\\', '/'),
    'test-argument': 'test-value',
    **kwargs
  }
```

and call it from appropriate places.

barthisrael: Let's see how it goes, I just submitted a commit. I replaced `**kwargs` with a `params={}` to allow passing things like `test-argument`, which would be an invalid parameter name.

barthisrael:
I'll fix an issue with `flake8` soon, once the behave tests finish.

features/environment.py
```diff
@@ -892,22 +892,28 @@ def clone(self, from_name, cluster_name, to_name):
         }
         self.start(to_name, custom_config=custom_config)
 
+    def backup_restore_config(self, params={}):
+        return {
+            'command': (self.BACKUP_RESTORE_SCRIPT
+                        + ' --sourcedir=' + os.path.join(self.patroni_path, 'data', 'basebackup')).replace('\\', '/'),
+            'test-argument': 'test-value',  # test config mapping approach on custom bootstrap/replica creation
+            **params,
```
CyberDem0n, line 900: Lets also make codacy happy:
```suggestion
    def backup_restore_config(self, params=None):
        return {
            'command': (self.BACKUP_RESTORE_SCRIPT
                        + ' --sourcedir=' + os.path.join(self.patroni_path, 'data', 'basebackup')).replace('\\', '/'),
            'test-argument': 'test-value',  # test config mapping approach on custom bootstrap/replica creation
            **(params or {}),
```

hughcapet:
👍

CyberDem0n:
:+1:

PR #2928, opened 2023-10-24
Improve formatting of generated config

CyberDem0n:
- order sections similar to sample configs
- add warnings and comments to `bootstrap.dcs` section.
- add `tags` and `log` sections.
- use discovered IPs in `postgresql.connect_address` and `postgresql.listen`
- set `wal_level` to `replica` for PostgreSQL 9.6+
- make unit tests pass with python 3.6
- improve config validator so it doesn't complain when some ints are strings in YAML file.

### DIFF
--- a/patroni/config_generator.py
+++ b/patroni/config_generator.py
@@ -9,7 +9,7 @@ import yaml
 
 from getpass import getuser, getpass
 from contextlib import contextmanager
-from typing import Any, Dict, Iterator, List, Optional, Tuple, TYPE_CHECKING, Union
+from typing import Any, Dict, Iterator, List, Optional, TextIO, Tuple, TYPE_CHECKING, Union
 if TYPE_CHECKING:  # pragma: no cover
     from psycopg import Cursor
     from psycopg2 import cursor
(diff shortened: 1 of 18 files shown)


patroni/config_generator.py
```diff
@@ -130,16 +136,82 @@ def get_template_config(cls) -> Dict[str, Any]:
...
+            if isinstance(self.config[section_name], dict):
+                yield ''
+            yield self._format_block({section_name: self.config[section_name]})
+
+    def _format_config(self) -> Iterator[str]:
+        """Format current :attr:`~AbstractConfigGenerator.config` and enrich it with some comments.
+
+        :yields: formatted lines or blocks that represent a text output of the YAML document.
+        """
+        for name in ('scope', 'namespace', 'name', 'restapi', 'ctl' 'citus',
```
hughcapet, line 174: log section?

CyberDem0n: BTW, should we add it do the generated config?

hughcapet: definitely won't hurt

hughcapet:
👍

CyberDem0n:
:+1:

PR #2932, opened 2023-10-27
Add ability to pass auth_data to zk client

mumcuyan:
Hi! This PR adds ability to specify ensemble authentication value for the client initiated in Zookeeper.
It's a ZK option available starting from 3.6.0: https://zookeeper.apache.org/doc/r3.6.0/zookeeperAdmin.html, allowing the client to make sure it connects to right Zookeeper server.

### DIFF
--- a/docs/ENVIRONMENT.rst
+++ b/docs/ENVIRONMENT.rst
@@ -86,6 +86,7 @@ ZooKeeper
 -  **PATRONI\_ZOOKEEPER\_VERIFY**: (optional) Whether to verify certificate or not. Defaults to ``true``.
 -  **PATRONI\_ZOOKEEPER\_SET\_ACLS**: (optional) If set, configure Kazoo to apply a default ACL to each ZNode that it creates. ACLs will assume 'x509' schema and should be specified as a dictionary with the principal as the key and one or more permissions as a list in the value.  Permissions may be one of ``CREATE``, ``READ``, ``WRITE``, ``DELETE`` or ``ADMIN``.  For example, ``set_acls: {CN=principal1: [CREATE, READ], CN=principal2: [ALL]}``.
 -  **PATRONI\_ZOOKEEPER\_AUTH\_DATA**: (optional) Authentication credentials to use for the connection. Should be a dictionary in the form that `scheme` is the key and `credential` is the value. Defaults to empty dictionary.
+
 .. note::
     It is required to install ``kazoo>=2.6.0`` to support SSL.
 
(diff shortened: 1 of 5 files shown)


patroni/dcs/zookeeper.py
```diff
@@ -126,6 +126,9 @@ def __init__(self, config: Dict[str, Any]) -> None:
         self._client._connection._connect = self._kazoo_connect
 
         self._client.start()
+        ensemble_auth = config.get('ensemble_auth', '')
+        if ensemble_auth:
+            self._client.add_auth("ensemble", ensemble_auth)
```
CyberDem0n, line 131: The thing is that the python-kazoo module doesn't have a single mention of "ensemble" in this context.
Therefore lets be general enough: I would prefer passing `auth_data` parameter to the `KazooClient` constructor, similar to `set_acls`, instead of hardcoding "ensemble_auth".
In the config it may look like:
```yaml
zookeeper:
  auth_data:
    ensemble: blabla
```
or env variable:
```
PATRONI_ZOOKEEPER_AUTH_DATA='{auth_data: {ensemble: blabla}}'
```

mumcuyan: 👍  done

docs/ENVIRONMENT.rst
```diff
@@ -85,7 +85,7 @@ ZooKeeper
 -  **PATRONI\_ZOOKEEPER\_KEY\_PASSWORD**: (optional) The client key password.
 -  **PATRONI\_ZOOKEEPER\_VERIFY**: (optional) Whether to verify certificate or not. Defaults to ``true``.
 -  **PATRONI\_ZOOKEEPER\_SET\_ACLS**: (optional) If set, configure Kazoo to apply a default ACL to each ZNode that it creates. ACLs will assume 'x509' schema and should be specified as a dictionary with the principal as the key and one or more permissions as a list in the value.  Permissions may be one of ``CREATE``, ``READ``, ``WRITE``, ``DELETE`` or ``ADMIN``.  For example, ``set_acls: {CN=principal1: [CREATE, READ], CN=principal2: [ALL]}``.
-
+-  **PATRONI\_ZOOKEEPER\_AUTH\_DATA**: (optional) Authentication credentials to use for the connection. Should be a dictionary in the form that `scheme` is the key and `credential` is the value. Defaults to empty dictionary.
```
CyberDem0n, line 88: You also need to update https://github.com/zalando/patroni/blob/ce10e5fccc8eb6fc3e0e6bd177f6958d463904fd/docs/yaml_configuration.rst?plain=1#L126-L138

mumcuyan: missed that, thanks, updated.

CyberDem0n, line 88: > /home/runner/work/patroni/patroni/docs/ENVIRONMENT.rst:89: WARNING: Bullet list ends without a blank line; unexpected unindent.

```suggestion
-  **PATRONI\_ZOOKEEPER\_AUTH\_DATA**: (optional) Authentication credentials to use for the connection. Should be a dictionary in the form that `scheme` is the key and `credential` is the value. Defaults to empty dictionary.

```

hughcapet:
👍

CyberDem0n:
:+1:

PR #2933, opened 2023-10-27
Add a FAQ page to the docs

barthisrael:
This PR introduces a FAQ page to the docs. The idea is to get most frequently asked questions answered before-hand, so the user is able to get them answered quickily without going into detail in the docs or having to go to Slack/GitHub to clarify questions.

References: PAT-221.

### DIFF
--- a/docs/faq.rst
+++ b/docs/faq.rst
@@ -208,11 +208,19 @@ Postgres management
 -------------------
 
 Can I change Postgres GUCs directly in Postgres configuration?
-    You should **not** do that!
+    You can, but you should avoid that.
 
-    Postgres configuration is managed by Patroni, and attempts to edit the configuration files will end up being frustrated by Patroni as it will eventually overwrite them.
+    Postgres configuration is managed by Patroni, and attempts to edit the configuration files may end up being frustrated by Patroni as it may eventually overwrite them.
 
-    You need to manage all the Postgres configuration through Patroni!
+    There are a few options available to overcome the management performed by Patroni:
+
+    * Change Postgres GUCs through ``$PGDATA/postgresql.base.conf``; or
+    * Define a ``postgresql.custom_conf`` which will be used instead of ``postgresql.base.conf`` so you can manage that externally; or
+    * Change GUCs using ``ALTER SYSTEM`` / ``ALTER DATABASE`` / ``ALTER USER``.
+
+    You can find more information about that in the section :ref:`important_configuration_rules`.
+
+    In any case we recommend that you manage all the Postgres configuration through Patroni. That will centralize the management and make it easier to debug Patroni when needed.
 
 Can I restart Postgres nodes directly?
     No, you should **not** attempt to manage Postgres directly!
--- a/docs/faq.rst
+++ b/docs/faq.rst
@@ -139,7 +139,7 @@ How can I change my environment configuration?
 
     With that in mind, if you change the environment configuration you will need to restart the corresponding Patroni agent.
 
-    Take care to not cause a failover in the cluster!
+    Take care to not cause a failover in the cluster! You might be interested in checking :ref:`patronictl_pause`.
 
 What occurs if I change a Postgres GUC that requires a reload?
     When you change the dynamic or the local configuration as explained in the previous questions, Patroni will take care of reloading the Postgres configuration for you.
@@ -220,6 +220,13 @@ Does Patroni require a minimum number of Postgres nodes in the cluster?
     No, you can run Patroni with any number of Postgres nodes.
 
     Remember: Patroni is decoupled from the DCS.
+
+What does ``pause`` mean in Patroni?
+    Pause is an operation exposed by Patroni so the user can ask Patroni to step back in regards to Postgres management.
+
+    That is mainly useful when you want to perform maintenance on the cluster, and would like to avoid that Patroni takes decisions related with HA, like failing over to a standby when you stop the primary.
+
+    You can find more information about that in :ref:`pause`.
 
 Automatic failover
 ------------------
--- a/docs/faq.rst
+++ b/docs/faq.rst
@@ -141,6 +141,7 @@ How can I change my local configuration?
     * Send a ``POST`` request to the REST API :ref:`reload_endpoint`; or
     * Run :ref:`patronictl_reload`; or
     * Locally signal the Patroni process with ``SIGHUP``:
+
         * If you started Patroni through systemd, you can use the command ``systemctl reload PATRONI_UNIT.service``, ``PATRONI_UNIT`` being the name of the Patroni service; or
         * If you started Patroni through other means, you will need to identify the ``patroni`` process and run ``kill -s HUP PID``, ``PID`` being the process ID of the ``patroni`` process.
 
--- a/docs/faq.rst
+++ b/docs/faq.rst
@@ -167,6 +167,8 @@ What occurs if I change a Postgres GUC that requires a restart?
     * :ref:`patronictl_restart`; or
     * A ``POST`` request to :ref:`restart_endpoint`.
 
+    **Note:** some Postgres GUCs require a special management in terms of the order for restarting the Postgres nodes. Refer to :ref:`shared_memory_gucs` for more details.
+
 What is the difference between ``etcd`` and ``etcd3`` in Patroni configuration?
     ``etcd`` uses the API version 2 of ``etcd``, while ``etcd3`` uses the API version 3 of ``etcd``.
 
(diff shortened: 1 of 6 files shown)


docs/faq.rst
```diff
@@ -0,0 +1,287 @@
...
+    ``etcd`` uses the API version 2 of ``etcd``, while ``etcd3`` uses the API version 3 of ``etcd``.
+
+    Be aware that information stored by the API version 2 is not manageable by API version 3 and vice-versa.
+
+    The recommended one is ``etcd3`` taking into consideration that ``etcd`` API version 2 end of life was in 2017.
+
+I have ``use_slots`` enabled in my Patroni configuration, but when a cluster member goes offline for some time, the replication slot used by that member is dropped on the upstream node. What can I do to avoid that issue?
+    You can configure a permanent physical replication slot for the members.
+
+    Since Patroni ``3.2.0`` it is now possible to have member slots as permanent slots managed by Patroni.
```
CyberDem0n, line 165: This is not entirely true. Permanent physical slots that match with member names was possible to have to a few years, but they existed only on the leader. That is, in case of failover/switchover these slots were created on the new leader, but that didn't guaranty that it had all WAL segments for the absent node.

docs/faq.rst
```diff
@@ -0,0 +1,287 @@
...
+
+    Patroni will create the permanent physical slots on all nodes, and make sure to not remove the slots, as well as to advance the slots' LSN on all nodes according to the LSN that has been consumed by the member.
```
CyberDem0n, line 167: There is a small race condition. In the very beginning, when the slot is created on the replica it could be ahead of the same slot on the leader and in case if nobody is consuming from the slot there is still a chance that some files could be missing after failover. But in any case this is a good improvement comparing to the old state.

P.S., maybe also recommend to run continues archiving / PITR?

docs/faq.rst
```diff
@@ -0,0 +1,287 @@
...
+
+    Patroni stores the cluster's status in the DCS, among them a ``leader`` lock which holds the name of the Patroni member which is the current ``leader`` of the cluster.
+
+    That ``leader`` lock has a time-to-live associated with it. If the leader node fails to update the lease of the ``leader`` lock in time, the key will eventually expire from the DCS.
+
+    When the ``leader`` lock expires, it triggers what Patroni calls a ``leader race``: all nodes start performing checks to determine if they are the best candidates for taking over the ``leader`` role.
+    Some of these checks include calls to the REST API of all other Patroni members.
+
+    All Patroni members that find themselves as the best candidate for taking over the ``leader`` lock will attempt to do so.
+    The first Patroni member that is able to take the ``leader`` lock will promote itself to a read/write node, and the others will be configured to follow it.
```
CyberDem0n, line 254:
```suggestion
    The first Patroni member that is able to take the ``leader`` lock will promote itself to a read/write node (or ``standby leader``), and the others will be configured to follow it.
```

docs/faq.rst
barthisrael: Added a commit with this info.

docs/faq.rst
barthisrael: Added a commit with this info.

docs/faq.rst
```diff
@@ -0,0 +1,291 @@
...
+What is the difference between ``loop_wait``, ``retry_timeout`` and ``ttl``?
+    Patroni performs what we call a HA cycle from time to time. On each HA cycle it takes care of performing a series of checks on the cluster to determine its healthiness, and depending on the status it may take actions, like failing over to a standby.
+
+    ``loop_wait`` determines for long, in seconds, Patroni should sleep before performing a new cycle of HA checks.
+
+    ``retry_timeout`` sets the timeout for retry operations on the DCS and on Postgres. For example: if the DCS is unresponsive for more than ``retry_timeout`` seconds, Patroni might demote the primary node as a security action.
+
+    ``ttl`` sets the lease time on the ``leader`` lock in the DCS. If the current leader of the cluster is not able to renew the lease during its HA cycles for longer than ``ttl``, then the lease will expire and that will trigger a ``leader race`` in the cluster.
+
+Postgres management
```
CyberDem0n, line 184: Please also add an item that postgresql.service **MUST** be disabled.
And explanation: if the node running as a primary failed, the postgresql.service shouldn't start postgres automatically because it may cause a split-brain. Only Patroni is allowed to start/stop/promote Postgres.

docs/faq.rst
```diff
@@ -0,0 +1,291 @@
...
+    Be aware that that is not a rule, though. Some operations performed by Patroni cause it to immediately update the DCS information.
+
+Configuration
+-------------
+
+What is the difference between dynamic configuration and local configuration?
+    Dynamic configuration (or global configuration) is the configuration stored in the DCS, and which is applied to all members of the Patroni cluster.
+    This is primarily where you should store your configuration.
+
+    Settings that are specific to a node, or settings that you would like the global configuration overwritten, you should set only on the desired Patroni member as a local configuration.
```
hughcapet, line 94:
```suggestion
    Settings that are specific to a node, or settings that you would like to overwrite the global configuration with, you should set only on the desired Patroni member as a local configuration.
```
but maybe I don't get the idea of the sentence

docs/faq.rst
```diff
@@ -0,0 +1,291 @@
...
+
+    **Note:** even with Patroni ``3.2.0`` there might be a small race condition. In the very beginning, when the slot is created on the replica it could be ahead of the same slot on the leader and in case if nobody is consuming the slot there is still a chance that some files could be missing after failover. With that in mind, it is recommended that you configure continuous archiving, which makes it possible to restore required WALs or perform PITR.
+
+What is the difference between ``loop_wait``, ``retry_timeout`` and ``ttl``?
+    Patroni performs what we call a HA cycle from time to time. On each HA cycle it takes care of performing a series of checks on the cluster to determine its healthiness, and depending on the status it may take actions, like failing over to a standby.
+
+    ``loop_wait`` determines for long, in seconds, Patroni should sleep before performing a new cycle of HA checks.
```
hughcapet, line 178:
```suggestion
    ``loop_wait`` determines for how long, in seconds, Patroni should sleep before performing a new cycle of HA checks.
```

docs/faq.rst
barthisrael: That's it, yours looks better :)

docs/faq.rst
```diff
@@ -0,0 +1,291 @@
...
+What occurs if I change a Postgres GUC that requires a reload?
+    When you change the dynamic or the local configuration as explained in the previous questions, Patroni will take care of reloading the Postgres configuration for you.
+
+What occurs if I change a Postgres GUC that requires a restart?
+    Patroni will mark the affected members with a flag of ``pending restart``.
+
+    It is up to you to determine when and how to restart the members. That can be accomplished either through:
+
+    * :ref:`patronictl_restart`; or
+    * A ``POST`` request to :ref:`restart_endpoint`.
```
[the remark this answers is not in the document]
barthisrael: Cool, good point, I'll add a reference to the relevant docs that explain that :)

docs/faq.rst
```diff
@@ -0,0 +1,291 @@
...
+
+    **Note:** on Patroni older than ``3.2.0`` you could still have member slots configured as permanent physical slots, however they would be managed only on the current leader. That is, in case of failover/switchover these slots would be created on the new leader, but that wouldn't guarantee that it had all WAL segments for the absent node.
+
+    **Note:** even with Patroni ``3.2.0`` there might be a small race condition. In the very beginning, when the slot is created on the replica it could be ahead of the same slot on the leader and in case if nobody is consuming the slot there is still a chance that some files could be missing after failover. With that in mind, it is recommended that you configure continuous archiving, which makes it possible to restore required WALs or perform PITR.
+
+What is the difference between ``loop_wait``, ``retry_timeout`` and ``ttl``?
+    Patroni performs what we call a HA cycle from time to time. On each HA cycle it takes care of performing a series of checks on the cluster to determine its healthiness, and depending on the status it may take actions, like failing over to a standby.
```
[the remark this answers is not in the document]
barthisrael: Right, I'll add a reference to the corresponding docs page.

docs/faq.rst
```diff
@@ -0,0 +1,291 @@
...
+
+Postgres management
+-------------------
+
+Can I change Postgres GUCs directly in Postgres configuration?
+    You should **not** do that!
+
+    Postgres configuration is managed by Patroni, and attempts to edit the configuration files will end up being frustrated by Patroni as it will eventually overwrite them.
+
+    You need to manage all the Postgres configuration through Patroni!
```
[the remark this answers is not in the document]
barthisrael: Thanks, added that info and a link to the "important rules" section.

hughcapet:
👍

hughcapet:
Thank you, @barthisrael! 🥳 ❤️

barthisrael:
> Thank you, @barthisrael! 🥳 ❤️

Glad to work with both of you and contribute with the project :)
I hope that page will be helpful to users🤞

PR #2935, opened 2023-11-01
Make GlobalConfig really global

CyberDem0n:
1. extract `GlobalConfig` class to its own module
2. make the module instantiate the `GlobalConfig` object on load and replace sys.modules with the this instance
3. don't pass `GlobalConfig` object around, but use `patroni.global_config` module everywhere.
4. move `ignore_slots_matchers`, `max_timelines_history`,  and `permanent_slots` from `ClusterConfig` to `GlobalConfig`.
5. add `use_slots` property to global_config and remove duplicated code from `Cluster` and `Postgresql.ConfigHandler`.

Besides that improve readability of couple of checks in ha.py and formatting of `/config` key when saved from patronictl.


patroni/global_config.py
```diff
@@ -0,0 +1,224 @@
...
+        """
+        return bool(cluster and cluster.config and cluster.config.modify_version)
+
+    def update(self, cluster: Optional['Cluster']) -> None:
+        """Update with the new global configuration from the :class:`Cluster` object view.
+
+        .. note::
+            Global configuration is updated only when configuration in the *cluster* view is valid.
+
+            Update happens in-pleace and executed only from the main heartbeat Thread.
```
hughcapet, line 53:
```suggestion
            Update happens in-place and executed only from the main heartbeat Thread.
```

patroni/global_config.py
```diff
@@ -0,0 +1,224 @@
...
+
+    @property
+    def use_slots(self) -> bool:
+        """``True`` if cluster is configured to use replication slots."""
+        return bool(parse_bool((self.get('postgresql') or {}).get('use_slots', True)))
+
+    @property
+    def permanent_slots(self) -> Dict[str, Any]:
+        """Dictionary of permanent slots information from the global configuration."""
+        return (self.get('permanent_replication_slots') or self.get('permanent_slots') or self.get('slots') or {})
```
barthisrael, line 221: I wonder if we should return a "deep copy" in this case too?

barthisrael (commented):
I'll need to switch context, but I'm posting some thoughts that I had so far.

barthisrael (approved):
Nice work!

barthisrael:
👍

hughcapet:
👍

hughcapet:
👍

CyberDem0n:
:+1:

PR #2937, opened 2023-11-03
update Dockerfiles

rockdrilla:
- better cleanup for vim
- adjust dumb-init usage


Dockerfile
```diff
@@ -143,6 +143,7 @@ ARG PGBIN=/usr/lib/postgresql/$PG_MAJOR/bin
 
 ENV LC_ALL=$LC_ALL LANG=$LANG EDITOR=/usr/bin/editor
 ENV PGDATA=$PGDATA PATH=$PATH:$PGBIN
+ENV DUMB_INIT_SETSID=0
```
CyberDem0n, line 146: Currently dump-init is only used for haproxy/confd container.
I am not sure if this is a good thing to have this option for it.

I would suggest instead to introduce dump-init for patroni containers and have this option directly in the entrypoint.sh file: https://github.com/zalando/patroni/blob/552e8643d9d5555ecaf7a882638bcc8302dc6933/docker/entrypoint.sh#L75

rockdrilla:
@CyberDem0n PTAL

docker/entrypoint.sh
```diff
@@ -13,6 +13,23 @@ readonly PATRONI_NAMESPACE="${PATRONI_NAMESPACE%/}"
...
+    while [ "$USE_DUMB_INIT" = 1 ] ; do
+        command -v dumb-init >/dev/null 2>&1 || break
+        # verify $DUMB_INIT_OPTS with /bin/true
+        # shellcheck disable=SC2086
+        dumb-init $DUMB_INIT_OPTS -- true || break
+        # shellcheck disable=SC2086
+        exec dumb-init $DUMB_INIT_OPTS -- "$@"
+    done
+    exec "$@"
+}
```
CyberDem0n, line 31:
```suggestion
export DUMB_INIT_SETSID=0
```

docker/entrypoint.sh
```diff
@@ -33,13 +50,13 @@ case "$1" in
 $(echo "$ETCDCTL_ENDPOINTS" | sed 's/,/\n/g')
 EOT
         fi
-        exec dumb-init "$@"
+        exec_init "$@"
```
CyberDem0n, line 53:
```suggestion
        exec dumb-init "$@"
```

docker/entrypoint.sh
```diff
@@ -33,13 +50,13 @@ case "$1" in
...
         ;;
     etcd)
-        exec "$@" -advertise-client-urls "http://$DOCKER_IP:2379"
+        exec_init "$@" -advertise-client-urls "http://$DOCKER_IP:2379"
```
CyberDem0n, line 56:
```suggestion
        exec "$@" -advertise-client-urls "http://$DOCKER_IP:2379"
```

docker/entrypoint.sh
```diff
@@ -33,13 +50,13 @@ case "$1" in
...
         ;;
     zookeeper)
-        exec /usr/share/zookeeper/bin/zkServer.sh start-foreground
+        exec_init /usr/share/zookeeper/bin/zkServer.sh start-foreground
```
CyberDem0n, line 59:
```suggestion
        exec /usr/share/zookeeper/bin/zkServer.sh start-foreground
```

docker/entrypoint.sh
```diff
@@ -72,4 +89,4 @@ export PATRONI_SUPERUSER_SSLKEY="${PATRONI_SUPERUSER_SSLKEY:-$PGSSLKEY}"
 export PATRONI_SUPERUSER_SSLCERT="${PATRONI_SUPERUSER_SSLCERT:-$PGSSLCERT}"
 export PATRONI_SUPERUSER_SSLROOTCERT="${PATRONI_SUPERUSER_SSLROOTCERT:-$PGSSLROOTCERT}"
 
-exec python3 /patroni.py postgres0.yml
+exec_init python3 /patroni.py postgres0.yml
```
CyberDem0n, line 92:
```suggestion
exec dumb-init python3 /patroni.py postgres0.yml
```

CyberDem0n (commented):
That's way more changes than I anticipated.
We need dumb-init only for patroni and haproxy containers. Everything else doesn't create child processes and works just fine.

Besides that I see no reason for supporting `DUMP_INIT_OPTS`. The only goal of these Dockerfile and docker-compose.yml is to be able to play with Patroni locally. It is not for production and we don't need unnecessary flexibility.

CyberDem0n:
:+1:

PR #2939, opened 2023-11-03
Verify that replica nodes received checkpoint LSN on shutdown

CyberDem0n:
In case if archiving is enabled the `Postgresql.latest_checkpoint_location()` method returns LSN of the prev (SWITCH) record, which points to the beginning of the WAL file. It is done in order to make it possible to safely promote replica which recovers WAL files from the archive and wasn't streaming when the primary was stopped (primary doesn't archive this WAL file).

But, in certain cases using the LSN pointing to SWITCH record was causing unnecessary pg_rewind, if replica didn't managed to replay shutdown checkpoint record before it was promoted.

In order to mitigate the problem we need to check that replica received/replayed exactly the shutdown checkpoint LSN. But, at the same time we will still write LSN of the SWITCH record to the `/status` key when releasing the leader lock.

### DIFF
--- a/patroni/postgresql/__init__.py
+++ b/patroni/postgresql/__init__.py
@@ -596,7 +596,7 @@ class Postgresql(object):
 
         :param data: :class:`dict` object with values returned by `pg_controldata` tool.
 
-        :returns: a tuple of checkpoint LSN for the cleanly shut down primary, and and LSN of prev wal record (SWITCH)
+        :returns: a tuple of checkpoint LSN for the cleanly shut down primary, and LSN of prev wal record (SWITCH)
                   if we know that the checkpoint was written to the new WAL file due to the archive_mode=on.
         """
         timeline = data.get("Latest checkpoint's TimeLineID")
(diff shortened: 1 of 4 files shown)


patroni/postgresql/__init__.py
```diff
@@ -591,14 +591,17 @@ def parse_wal_record(self, timeline: str,
...
-        """Returns checkpoint location for the cleanly shut down primary.
-           But, if we know that the checkpoint was written to the new WAL
-           due to the archive_mode=on, we will return the LSN of prev wal record (SWITCH)."""
+    def _checkpoint_locations_from_controldata(self, data: Dict[str, str]) -> Optional[Tuple[int, int]]:
+        """Get shutdown checkpoint location.
+
+        :param data: :class:`dict` object with values returned by `pg_controldata` tool.
 
-        data = self.controldata()
+        :returns: a tuple of checkpoint LSN for the cleanly shut down primary, and and LSN of prev wal record (SWITCH)
```
hughcapet, line 599:
```suggestion
        :returns: a tuple of checkpoint LSN for the cleanly shut down primary, and LSN of prev wal record (SWITCH)
```

hughcapet:
👍

CyberDem0n:
:+1:

PR #2942, opened 2023-11-08
Do a real http request when performing name uniqueness check

CyberDem0n:
When running in containers it is possible that the traffic is routed using `docker-proxy`, which listens on the port and accepting incoming connections.

This commit effectively sticks to the original solution from #2878


hughcapet:
👍

barthisrael:
👍

PR #2944, opened 2023-11-09
Fix pg_rewind behavior with Postgres v16+

CyberDem0n:
The error message format was changed in
https://github.com/postgres/postgres/commit/4ac30ba4f29d4b586b131404b0d514f16501272a, what caused `pg_rewind` being called by Patroni even when it was not necessary.


barthisrael:
👍

hughcapet:
👍

CyberDem0n:
:+1:

PR #2945, opened 2023-11-10
Get rid of pass_obj() in most of patronictl commands

CyberDem0n:
The `obj` could be easily obtained with the help of `click.get_current_context().obj`.

Introduced function `is_citus_cluster()` will simplify future refactoring to add support of other MPP databases.

In addition to that refactor ctl.py unit tests by moving most of mocks to the global scope.,

### DIFF
--- a/patroni/ctl.py
+++ b/patroni/ctl.py
@@ -825,7 +825,6 @@ def query(
     for _ in watching(w, watch, clear=False):
         if cluster is None:
             cluster = dcs.get_cluster()
-#            cursor = get_cursor(cluster, group, connect_parameters, role=role, member=member)
 
         output, header = query_member(cluster, group, cursor, member, role, sql, connect_parameters)
         print_output(header, output, fmt=fmt, delimiter=delimiter)
--- a/patroni/ctl.py
+++ b/patroni/ctl.py
@@ -306,10 +306,18 @@ def ctl(ctx: click.Context, config_file: str, dcs_url: Optional[str], insecure: 
     ctx.obj.setdefault('ctl', {})['insecure'] = ctx.obj.get('ctl', {}).get('insecure') or insecure
 
 
-def get_dcs(config: Dict[str, Any], scope: str, group: Optional[int]) -> AbstractDCS:
+def is_citus_cluster() -> bool:
+    """Check if we are working with Citus cluster.
+
+    :returns: ``True`` if configuration has ``citus`` section, otherwise ``False``.
+    """
+    ctx = click.get_current_context()
+    return bool(ctx.obj.get('citus'))
+
+
+def get_dcs(scope: str, group: Optional[int]) -> AbstractDCS:
     """Get the DCS object.
 
-    :param config: Patroni configuration.
     :param scope: cluster name.
     :param group: if *group* is defined, use it to select which alternative Citus group this DCS refers to. If *group*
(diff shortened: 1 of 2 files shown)


barthisrael (commented):
I think it is good!
Just a question: did you keep the `pass_obj` in `members` because we only touch `obj` in that function, I mean, the `obj` was not directly used by any other function only in `members`? I wonder if we should leave it that way or use the `click.get_current_context()` in `members` just to keep the pattern of the other functions in terms of not using the decorator?

CyberDem0n:
The `members()` function is kind of special, it supports multiple cluster names, and if none is specified it tries to explicitly get `scope` from the config.

A  bunch of other functions get the default value of cluster name by calling a `default` callback on the `argument` object:
```python
arg_cluster_name = click.argument('cluster_name', required=False,
                                  default=lambda: click.get_current_context().obj.get('scope'))
```

barthisrael:
> The `members()` function is kind of special, it supports multiple cluster names, and if none is specified it tries to explicitly get `scope` from the config.
> 
> A bunch of other functions get the default value of cluster name by calling a `default` callback on the `argument` object:
> 
> ```python
> arg_cluster_name = click.argument('cluster_name', required=False,
>                                   default=lambda: click.get_current_context().obj.get('scope'))
> ```

Wouldn't it be possible to do something like this, for example:

```diff
diff --git a/patroni/ctl.py b/patroni/ctl.py
index 75f6260..2883203 100644
--- a/patroni/ctl.py
+++ b/patroni/ctl.py
@@ -1604,14 +1604,12 @@ def output_members(cluster: Cluster, name: str, extended: bool = False,
 @option_format
 @option_watch
 @option_watchrefresh
-@click.pass_obj
-def members(obj: Dict[str, Any], cluster_names: List[str], group: Optional[int],
+def members(cluster_names: List[str], group: Optional[int],
             fmt: str, watch: Optional[int], w: bool, extended: bool, ts: bool) -> None:
     """Process ``list`` command of ``patronictl`` utility.

     Print information about the Patroni cluster through :func:`output_members`.

-    :param obj: Patroni configuration.
     :param cluster_names: name of clusters that should be printed. If ``None`` consider only the cluster present in
         ``scope`` key of *obj*.
     :param group: filter which Citus group we should get members from. Refer to the module note for more details.
@@ -1622,9 +1620,10 @@ def members(obj: Dict[str, Any], cluster_names: List[str], group: Optional[int],
         more details.
     :param ts: if timestamp should be included in the output.
     """
+    config = click.get_current_context().obj
     if not cluster_names:
-        if 'scope' in obj:
-            cluster_names = [obj['scope']]
+        if 'scope' in config:
+            cluster_names = [config['scope']]
         if not cluster_names:
             return logging.warning('Listing members: No cluster names were provided')
```

However I'm not sure how much that would be different from the current approach other than removing the `obj` argument.

CyberDem0n:
At least it will make the fiele one loc shorter :)

barthisrael:
> At least it will make the fiele one loc shorter :)

Heh heh, I'm fine with the PR.
There is only a check failing on codacy. Are we going to make it happy before proceeding?

CyberDem0n:
@barthisrael codacy complains about code complexity in a couple of functions, I don't want to spend my time on it.

barthisrael:
👍

hughcapet:
👍

PR #2950, opened 2023-11-14
Abstract CitusHandler and decouple it from configuration

CyberDem0n:
the main issue was that the configuration for Citus handler and for DCS existed in two places, while ideally AbstractDCS should not know many details about what kind of MPP is in use.

To solve the problem we first dynamically create an object implementing AbstractMPP interfaces, which is a configuration for DCS. Later this object is used to instantiate the class implementing AbstractMPPHandler interface.

This is just a starting point, which does some heavy lifting. As a next steps all kind of variables named after Citus in files different from patroni/postgres/mpp/citus.py should be renamed.

In other words this commit takes over the most complex part of #2940, which was never implemented.

PR #2940, opened 2023-11-03
refactor and make it possible to integrate more formation dbs

zhjwpku:
Citus is one successful mpp architecture, besides it, there are many other mpp dbs which are based on postgresql, so it might be a good idea to make Patroni extendable for other mpp dbs.

The two key points for Patroni to support Citus is:

1. a DCS hierarchy with an extra ''group'' level for citus
2. CitusHandler to update pg_dist_node when failover happens

To make Patroni extenable, I'd first like to clarify the 5 meanings of *cluster* in Patroni.

1. Postgresql cluster: a cluster of postgresql instances which have the same system identifier.
2. Formation cluster: a cluster of Postgresql cluster which contains different data in each Postgresql cluster, usually one Postgresql cluster acts as coordinator, and other Postgresql clusters act as workers.
3. Coordinator cluster: a Postgresql cluster which has the role 'coordinator' within a Formation cluster.
4. Worker cluster: a Postgresql cluster which has the role 'worker' within a Formation cluster.
5. Patroni cluster: all cluster managed by Patroni are called Patroni cluster, including Postgresql cluster and Formation cluster.

Then the two main efforts of this patch are:

1. dcs: get rid of Citus specific code(like the cluster loader) and instead introduce formation group, cluter type when initializing concrete dcs class.
2. create a AbstractMPPHandler, define some hooks that module `Postgresql` and `HA` should call when things happens. Let CitusHandler inherit it, and implement the interfaces by defining what citus should do in each phase. Other mpp dbs can integrate themselves by implement another xxxHandler.

Co-author-by: @CyberDem0n

PR #2954, opened 2023-11-20
Factor out dynamic class finder/loader to a dedicated file

CyberDem0n:
It could be reused to do the same for MPP modules/classes.
Ref: #2940 and #2950

### DIFF
--- a/patroni/dcs/__init__.py
+++ b/patroni/dcs/__init__.py
@@ -98,7 +98,7 @@ def iter_dcs_classes(
     :param config: configuration information with possible DCS names as keys. If given, only attempt to import DCS
                    modules defined in the configuration. Else, if ``None``, attempt to import any supported DCS module.
 
-    :yields: a tuple containing the module ``name`` and the imported DCS class object.
+    :returns: an iterator of tuples, each containing the module ``name`` and the imported DCS class object.
     """
     return iter_classes(__package__, AbstractDCS, config)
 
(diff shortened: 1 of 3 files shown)


patroni/dynamic_loader.py
```diff
@@ -0,0 +1,96 @@
...
+logger = logging.getLogger(__name__)
+
+
+def iter_modules(package: str) -> List[str]:
+    """Get names of modules from *package*, depending on execution environment.
+
+    .. note::
+        If being packaged with PyInstaller, modules aren't discoverable dynamically by scanning source directory because
+        :class:`importlib.machinery.FrozenImporter` doesn't implement :func:`iter_modules`. But it is still possible to
+        find all potential DCS modules by iterating through ``toc``, which contains list of all "frozen" resources.
```
hughcapet, line 24:
```suggestion
        find all potential modules by iterating through ``toc``, which contains list of all "frozen" resources.
```

patroni/dcs/__init__.py
```diff
@@ -124,42 +100,14 @@ def iter_dcs_classes(
 
     :yields: a tuple containing the module ``name`` and the imported DCS class object.
```
barthisrael, line 101:
```suggestion
    :returns: an iterator of tuples, each containing the module ``name`` and the imported DCS class object.
```

I wonder if that describes the new situation more precisely? I mean, it's not generating values, but returning a generator now.

barthisrael (approved):
Looks good to me -- just a couple small comments.

barthisrael:
👍

hughcapet:
👍

Issue #2955, opened 2023-11-20
REST API returns unknown after postgres restart


XiuhuaRuan:
### What happened?

First, setup a patroni cluster with three postgres nodes. After restart postgres node, REST API returns "unknown" state

```
[root@sophia-pghost5 patroni]# curl http://localhost:8008
{"state": "unknown", "role": "master", "dcs_last_seen": 1700478557, "database_system_identifier": "7303378219179270632", "patroni": {"version": "3.2.0", "scope": "postgres-cluster", "name": "postgres_01"}}
```

### How can we reproduce it (as minimally and precisely as possible)?

Running patronictl restart cluster can reproduce it.

```
[root@sophia-pghost5 patroni]# patronictl -c patroni.yml restart postgres-cluster
+ Cluster: postgres-cluster (7303378219179270632) ---+----+-----------+
| Member      | Host           | Role    | State     | TL | Lag in MB |
+-------------+----------------+---------+-----------+----+-----------+
| postgres_01 | 192.168.61.105 | Leader  | running   |  1 |           |
| postgres_02 | 192.168.61.106 | Replica | streaming |  1 |         0 |
| postgres_03 | 192.168.61.107 | Replica | streaming |  1 |         0 |
+-------------+----------------+---------+-----------+----+-----------+
When should the restart take place (e.g. 2023-11-20T20:13)  [now]:
Are you sure you want to restart members postgres_01, postgres_02, postgres_03? [y/N]: y
Restart if the PostgreSQL version is less than provided (e.g. 9.5.2)  []:
Success: restart on member postgres_01
Success: restart on member postgres_02
Success: restart on member postgres_03
```

### What did you expect to happen?

REST API should return correct "running" state, even after restart postgres nodes.

### Patroni/PostgreSQL/DCS version

-   Patroni version: 3.2.0
-   PostgreSQL version: 14.0
-   DCS (and its version): etcd3.5.9

### Patroni configuration file

```yaml
scope: postgres-cluster
namespace: /service/
name: postgres_01

restapi:
  listen: 192.168.61.105:8008
  connect_address: 192.168.61.105:8008

etcd:
  hosts: 192.168.61.105:2379,192.168.61.106:2379,192.168.61.107:2379

log:
  level: INFO
  traceback_level: INFO
  dir: /home/postgres/patroni
  file_num: 10
  file_size: 104857600

bootstrap:
  dcs:
    ttl: 30
    loop_wait: 10
    retry_timeout: 10
[...]

PR #2956, opened 2023-11-21
Fix the issue that REST API returns unknown after postgres restart

XiuhuaRuan:
**Problem:**
Currenctly REST API could establish connection to postgres at the first time and return correct query result. But after postgres node restarted, restapi connection returned exception as the previous restapi connection was not closed properly. 
**Solution:**
For the close function in ConnectionPool, we need to close all named connections including heartbeat and restapi connections. So in this pull request, I change function any() to all() to close all connections in the pool. This will fix issue #2955

### DIFF
--- a/patroni/postgresql/connection.py
+++ b/patroni/postgresql/connection.py
@@ -147,7 +147,7 @@ class ConnectionPool:
     def close(self) -> None:
         """Close all named connections from Patroni to PostgreSQL registered in the pool."""
         with self._lock:
-            if all(conn.close(True) for conn in self._connections.values()):
+            if all(conn.close(False) for conn in self._connections.values()):
                 logger.info("closed patroni connections to postgres")
 
 
--- a/patroni/postgresql/connection.py
+++ b/patroni/postgresql/connection.py
@@ -147,7 +147,8 @@ class ConnectionPool:
     def close(self) -> None:
         """Close all named connections from Patroni to PostgreSQL registered in the pool."""
         with self._lock:
-            if all(conn.close(False) for conn in self._connections.values()):
+            closed_connections = [conn.close(True) for conn in self._connections.values()]
+            if any(closed_connections):
                 logger.info("closed patroni connections to postgres")
 
 
--- a/patroni/postgresql/connection.py
+++ b/patroni/postgresql/connection.py
@@ -147,7 +147,7 @@ class ConnectionPool:
     def close(self) -> None:
         """Close all named connections from Patroni to PostgreSQL registered in the pool."""
         with self._lock:
-            if any(conn.close(True) for conn in self._connections.values()):
+            if all(conn.close(True) for conn in self._connections.values()):
                 logger.info("closed patroni connections to postgres")
 
 


patroni/postgresql/connection.py
```diff
@@ -147,7 +147,7 @@ def get(self, name: str, kwargs_override: Optional[Dict[str, Any]] = None) -> Na
     def close(self) -> None:
         """Close all named connections from Patroni to PostgreSQL registered in the pool."""
         with self._lock:
-            if any(conn.close(True) for conn in self._connections.values()):
+            if all(conn.close(True) for conn in self._connections.values()):
```
CyberDem0n, line 150: Well, indeed `any` stops after the first success, but the fix isn't correct.
We want to log if _any_ of the connections were closed.

XiuhuaRuan: But we need to close all connections in this function for class ConnectionPool, right?
```
    def close(self) -> None:
        """**Close all named connections** from Patroni to PostgreSQL registered in the pool."""
        with self._lock:
            if any(conn.close(True) for conn in self._connections.values()):
                logger.info("**closed patroni connections to postgres**")
```

If we want to log if any of the connections were closed, I think we can set 'silent' to False to log it.

```
    def close(self, silent: bool = False) -> bool:
        """Close the psycopg connection to postgres.

        :param silent: whether the method should not write logs.

        :returns: ``True`` if ``psycopg`` connection was closed, ``False`` otherwise.``
        """
        ret = False
        if self._connection and self._connection.closed == 0:
            self._connection.close()
            if not silent:
                logger.info("closed patroni %s connection to postgres", self._name)
            ret = True
        self._connection = None
        return ret
```

Please check if below change is OK. Please tell me if there is any misunderstanding. 
```
    def close(self) -> None:
        """**Close all named connections** from Patroni to PostgreSQL registered in the pool."""
        with self._lock:
            if **all**(conn.close(**True**) for conn in self._connections.values()):
                logger.info("**closed patroni connections to postgres**")
```

patroni/postgresql/connection.py
```diff
@@ -147,7 +147,7 @@ def get(self, name: str, kwargs_override: Optional[Dict[str, Any]] = None) -> Na
     def close(self) -> None:
         """Close all named connections from Patroni to PostgreSQL registered in the pool."""
         with self._lock:
-            if any(conn.close(True) for conn in self._connections.values()):
+            if all(conn.close(False) for conn in self._connections.values()):
```
CyberDem0n, line 150: We don't want to have a separate log line for every closed connection,  but a single line indicating that one or more connections have been closed.
It could only be achieved by first executing close() method on all connections, accumulating results to a list, and checking if any of the elements in the list contain True.

CyberDem0n, line 150:
```suggestion
            closed_connections = [conn.close(True) for conn in self._connections.values()]
            if any(closed_connections):
```

hughcapet:
👍

CyberDem0n:
:+1:

XiuhuaRuan:
Sorry for the late response. I got your point and verified the latest fix is OK. Thanks.

PR #2960, opened 2023-11-25
Abstract MPP handler and decouple it from DCS

zhjwpku:
Ideally DCS should not know much details about what kind of MPP is in use, this PR solves the issue, so that we can introduce more MPP dbs in the future.

This PR contains two commits:
1. Abstract CitusHandler and decouple it from configuration: to solve the chicken-egg initialization problem, first dynamically create an object implementing AbstractMPP interfaces, which is a configuration for DCS. Later this object is used to instantiate the class implementing AbstractMPPHandler interface.
2. Introduce mpp package: this commit introduces patroni.postgresql.mpp package with some unittests.

This PR takes over #2940 and #2950 and it's the first step, following PRs would take care of:

1. renaming citus_handler to mpp_handler 
2. polish ctl.py

This work is co-authored by @CyberDem0n and @zhjwpku

### DIFF
--- a/patroni/__main__.py
+++ b/patroni/__main__.py
@@ -68,7 +68,7 @@ class Patroni(AbstractPatroniDaemon, Tags):
         self.watchdog = Watchdog(self.config)
         self.load_dynamic_configuration()
 
-        self.postgresql = Postgresql(self.config['postgresql'])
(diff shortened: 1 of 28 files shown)


zhjwpku:
Close this in favor of #2950

PR #2967, opened 2023-11-30
Only call postgres --describe-config once

waynerv:
When patroni write postgresql.conf, each configuration parameter will cause `postgres --describe-config` called once, resulting in a large amount of log printing for orphan process reaped. 
this small change might improve this situation.


CyberDem0n:
:+1:

PR #2969, opened 2023-12-01
Prevent HA nap_time longer than loop_wait

waynerv:
General there is time synchronization mechanism on each server node. 
In extreme scenarios, the system time may be rolled back several ten seconds, then master will sleep for more than the TTL  period in a loop, causing leader lock expiration and triggering a failover:
image

This issue is unrelated to Patroni, but I think limiting `nap_time` to not exceed `loop_wait` may make HA loop mechanism more robust.

### DIFF
--- a/patroni/__main__.py
+++ b/patroni/__main__.py
@@ -172,7 +172,7 @@ class Patroni(AbstractPatroniDaemon, Tags):
         """
         self.next_run += self.dcs.loop_wait
         current_time = time.time()
-        nap_time = self.next_run - current_time
+        nap_time = min(self.next_run - current_time, self.dcs.loop_wait)
         if nap_time <= 0:
             self.next_run = current_time
             # Release the GIL so we don't starve anyone waiting on async_executor lock


CyberDem0n:
I don't think in this specific case the time was moved back.
It looks like that the system is under huge stress and Patroni wasn't able to run for about 1 minute.
I bet that some other node was promoted while this node was wedged :(

waynerv:
There is indeed the possibility you mentioned. Currently, I have checked the CPU and memory monitor dashboard during failover but did not find any issues. I will further investigate other state of OS.

CyberDem0n:
@waynerv Did the failover happen at around 08:51:15?

CyberDem0n:
So yes, indeed, the Patroni process on the old primary was sort of frozen and not able to do anything.
In my experience usually it happens when there is a high memory pressure due to a lot of dirty pages (Dirty field from /proc/meminfo) due to slow writes. We were able to solve it by setting some reasonable values of `vm.dirty_bytes` and `vm.dirty_background_bytes`. By default Linux is configured to use `vm.dirty_ratio` and `vm.dirty_background_ratio`, which are in % of memory, and pretty bad for containers, where the amount of memory is lower than on the entire machine.

But, certainly Patroni could improve on that and implement some additional checks before starting the leader race. I even created a card for it: https://github.com/orgs/zalando/projects/2?pane=issue&itemId=42772671

waynerv:
Log of the database as a supplement. A strange phenomenon is that the timestamps recorded in the pglogs are no longer monotonically increasing.(we have csvlog and pgaudit enabled)
image

waynerv:
It seems the issue was addressed after we change  io scheduler algorithm from `cfq` to `deadline`. Considering low `Dirty` value, It appears that read occupies a larger proportion in database.

During peak business hours, the `await` and `%util` values in `iostat` have significantly decreased, and the number of blocked processes in `vmstat` results has also become very small.

@CyberDem0n I think this PR can be closed now, hope not bother you.

PR #2970, opened 2023-12-01
Don't let replica restore initialize key when DCS was wiped

CyberDem0n:
It was happening from the branch where Patroni was supposed to be complain about converting standalone PG cluster to be governed by Patroni and exit.


hughcapet:
👍

CyberDem0n:
:+1:

PR #2975, opened 2023-12-06
Reload postgres config if a server param was reset

hughcapet:
Fix the case when a parameter value was changed and then reset back to the initial value without restart - before this fix, the second change was not reflected in the Postgres config.
This commit also includes the related unit test refactoring.


CyberDem0n:
:+1:

hughcapet:
👍

PR #2976, opened 2023-12-06
Disable SSL for MacOS GH action runners

CyberDem0n:
Latest runners release (20231127.1) somehow broke our tests. Connections to postgres somehow failing with strange error:
```
could not accept SSL connection: Socket operation on non-socket
```


hughcapet:
👍

CyberDem0n:
:+1:

PR #2978, opened 2023-12-11
Add pending restart reason information

hughcapet:
Show info about the PG parameters that caused "pending restart"
flag to be set in both `patronictl list` and `/patroni` REST API endpoint response

### DIFF
--- a/patroni/ctl.py
+++ b/patroni/ctl.py
@@ -1576,8 +1576,8 @@ def output_members(cluster: Cluster, name: str, extended: bool = False,
             def format_diff(param: str, values: Dict[str, str], hide_long: bool):
                 full_diff = param + ': ' + values['old_value'] + '->' + values['new_value']
                 return full_diff if not hide_long or len(full_diff) <= 50 else param + ': [hidden - too long]'
-            restart_reason = '\n'.join(
-                [format_diff(k, v, fmt == 'pretty') for k, v in member.get('pending_restart_reason', {}).items()]) or ''
+            restart_reason = '\n'.join([format_diff(k, v, fmt in ('pretty', 'topology'))
+                                        for k, v in member.get('pending_restart_reason', {}).items()]) or ''
 
             member.update(cluster=name, member=member['name'], group=g,
                           host=member.get('host', ''), tl=member.get('timeline', ''),
--- a/patroni/ctl.py
+++ b/patroni/ctl.py
@@ -50,7 +50,6 @@ from . import global_config
 from .config import Config
 from .dcs import get_dcs as _get_dcs, AbstractDCS, Cluster, Member
 from .exceptions import PatroniException
-from .postgresql.config import ParamDiff
 from .postgresql.misc import postgres_version_to_int
 from .postgresql.mpp import get_mpp
 from .utils import cluster_as_json, patch_config, polling_loop
@@ -1574,8 +1573,11 @@ def output_members(cluster: Cluster, name: str, extended: bool = False,
 
             lag = member.get('lag', '')
 
-            restart_reason = '\n'.join([ParamDiff.as_string(k, v, fmt == 'pretty')
-                                        for k, v in member.get('pending_restart_reason', {}).items()]) or ''
+            def format_diff(param: str, values: Dict[str, str], hide_long: bool):
+                full_diff = param + ': ' + values['old_value'] + '->' + values['new_value']
+                return full_diff if not hide_long or len(full_diff) <= 50 else param + ': [hidden - too long]'
+            restart_reason = '\n'.join(
+                [format_diff(k, v, fmt == 'pretty') for k, v in member.get('pending_restart_reason', {}).items()]) or ''
 
             member.update(cluster=name, member=member['name'], group=g,
                           host=member.get('host', ''), tl=member.get('timeline', ''),
--- a/patroni/ctl.py
+++ b/patroni/ctl.py
@@ -1573,7 +1573,7 @@ def output_members(cluster: Cluster, name: str, extended: bool = False,
                 full_diff = param + ': ' + values[0] + '->' + values[1]
                 return full_diff if len(full_diff) <= 50 else param + ': [hidden - too long]'
             restart_reason = '\n'.join(
-                [format_diff(k,v) for k, v in member.get('pending_restart_reason', {}).items()]) or ''
+                [format_diff(k, v) for k, v in member.get('pending_restart_reason', {}).items()]) or ''
 
             member.update(cluster=name, member=member['name'], group=g,
                           host=member.get('host', ''), tl=member.get('timeline', ''),
--- a/patroni/ctl.py
+++ b/patroni/ctl.py
@@ -1569,8 +1569,8 @@ def output_members(cluster: Cluster, name: str, extended: bool = False,
 
             lag = member.get('lag', '')
 
-            def format_diff(param: str, values: Tuple[str, str], hide_long: bool):
-                full_diff = param + ': ' + values[0] + '->' + values[1]
+            def format_diff(param: str, values: Dict[str, str], hide_long: bool):
+                full_diff = param + ': ' + values['old_value'] + '->' + values['new_value']
                 return full_diff if not hide_long or len(full_diff) <= 50 else param + ': [hidden - too long]'
             restart_reason = '\n'.join(
                 [format_diff(k, v, fmt == 'pretty') for k, v in member.get('pending_restart_reason', {}).items()]) or ''
--- a/patroni/ctl.py
+++ b/patroni/ctl.py
@@ -1569,11 +1569,11 @@ def output_members(cluster: Cluster, name: str, extended: bool = False,
 
             lag = member.get('lag', '')
 
-            def format_diff(param: str, values: Tuple[str, str]):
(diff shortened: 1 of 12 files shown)


patroni/postgresql/__init__.py
```diff
@@ -78,6 +78,7 @@ def __init__(self, config: Dict[str, Any]) -> None:
         self.set_state('stopped')
 
         self._pending_restart = False
```
CyberDem0n, line 80: I think we can get rid of `self._pending_restart` in Postgresql.

patroni/postgresql/config.py
```diff
@@ -1154,16 +1156,27 @@ def reload_config(self, config: Dict[str, Any], sighup: bool = False) -> None:
             if self._postgresql.major_version >= 90500:
                 time.sleep(1)
                 try:
-                    pending_restart = self._postgresql.query(
-                        'SELECT COUNT(*) FROM pg_catalog.pg_settings'
-                        ' WHERE pg_catalog.lower(name) != ALL(%s) AND pending_restart',
-                        [n.lower() for n in self._RECOVERY_PARAMETERS])[0][0] > 0
-                    self._postgresql.set_pending_restart(pending_restart)
+                    settings_diff = {
+                        p: (v, (lambda x: '?' if x is None else x)(self._postgresql.get_guc_value(p)))
```
CyberDem0n, line 1160: What is `(lambda x: '?' if x is None else x)`?

patroni/postgresql/config.py
```diff
@@ -1154,16 +1156,27 @@ def reload_config(self, config: Dict[str, Any], sighup: bool = False) -> None:
...
+                            'SELECT name, setting FROM pg_catalog.pg_settings'
+                            ' WHERE pg_catalog.lower(name) != ALL(%s) AND pending_restart',
+                            [n.lower() for n in self._RECOVERY_PARAMETERS])
+                    }
+                    pending_restart = len(settings_diff) > 0
+                    external_change = settings_diff.keys() - param_diff.keys()
+                    if external_change:
+                        logger.info("PostgreSQL configuration parameters requiring restart"
+                                    " (%s) seem to be changed bypassing Patroni config."
+                                    " Setting 'Pending restart' flag", ', '.join(external_change))
```
CyberDem0n, line 1171: What about the case when parameter is set in Patroni config and in some externally managed file with a different value?

patroni/ha.py
```diff
@@ -363,6 +363,7 @@ def touch_member(self) -> bool:
                 data['tags'] = tags
             if self.state_handler.pending_restart:
                 data['pending_restart'] = True
+                data['pending_restart_reason'] = self.state_handler.pending_restart_reason
```
CyberDem0n, line 366: Why do we use different format for the /member key and REST API?
```yaml
  "pending_restart_reason": {
    "max_connections": [
      "100",
      "101"
    ]
  }
```

vs.

```yaml
  "pending_restart_reason": {
    "max_connections": {
      "old_value": "100",
      "new_value": "101"
    }
  },
```

CyberDem0n (commented):
I think we should also port convert_int_from_base_unit() func to python, it is not super complex, but will make output for parameters like shared_buffers human friendly:

https://github.com/postgres/postgres/blob/4908c5872059c409aa647bcde758dfeffe07996e/src/backend/utils/misc/guc.c#L2711-L2745

patroni/ha.py
hughcapet: the idea was to make rest api response more user-friendly, while /member key value is more for the quick access in Patroni code. But maybe it makes very little if any sense and I can make it unified

patroni/postgresql/config.py
hughcapet: `get_guc_value()` returns `None` if it failed to execute the cmd:)

patroni/ha.py
CyberDem0n: IMO it would be better to have it unified. I'm pretty sure there are tools that could check both sources.

tests/test_bootstrap.py
```diff
@@ -235,7 +235,7 @@ def test_post_bootstrap(self):
         self.assertTrue(task.result)
 
         self.b.bootstrap(config)
-        with patch.object(Postgresql, 'pending_restart', PropertyMock(return_value=True)), \
+        with patch.object(Postgresql, 'pending_restart_reason', PropertyMock({'max_connections': ('200', '100')})), \
```
CyberDem0n, line 238: Shouldn't it be ParamDiff()?

patroni/postgresql/config.py
```diff
@@ -270,6 +271,19 @@ def _bool_is_true_validator(value: Any) -> bool:
...
+
+    def __init__(self, old_value: Any, new_value: Any):
+        super(ParamDiff, self).__init__()
+        self['old_value'] = str(old_value)
+        self['new_value'] = str(new_value)
+
+    @staticmethod
+    def as_string(param: str, diff_dict: Dict[str, str], hide_long: bool) -> str:
+        full_diff = param + ': ' + diff_dict['old_value'] + '->' + diff_dict['new_value']
+        return full_diff if not hide_long or len(full_diff) <= 50 else param + ': [hidden - too long]'
```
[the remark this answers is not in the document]
hughcapet: Moved it here just logically. But can make it a bit uglier inner function in ctl.py again:)

patroni/postgresql/config.py
```diff
@@ -271,6 +271,31 @@ def _bool_is_true_validator(value: Any) -> bool:
...
+            :func:`~patroni.utils.maybe_convert_from_base_unit`.
+        :param unit: unit of *old/new_value*. See ``base_unit`` argument of
+            :func:`~patroni.utils.maybe_convert_from_base_unit`.
+        """
+        super().__init__()
+        str_value: Callable[[Any], str] = lambda x: '' if x is None else str(x)
+        self['old_value'] = (maybe_convert_from_base_unit(str_value(old_value), vartype, unit)
+                             if vartype else str_value(old_value))
+        self['new_value'] = (maybe_convert_from_base_unit(str_value(new_value), vartype, unit)
+                             if vartype else str_value(new_value))
```
CyberDem0n, line 296: It seems to be overengineered.
Instead it could be just a function, that returns a dict with two key-value pairs.

patroni/postgresql/config.py
```diff
@@ -1170,16 +1195,27 @@ def reload_config(self, config: Dict[str, Any], sighup: bool = False) -> None:
...
-                        'SELECT COUNT(*) FROM pg_catalog.pg_settings'
-                        ' WHERE pg_catalog.lower(name) != ALL(%s) AND pending_restart',
-                        [n.lower() for n in params_skip_changes])[0][0] > 0
-                    self._postgresql.set_pending_restart(pending_restart)
+                    settings_diff: CaseInsensitiveDict = CaseInsensitiveDict()
+                    for param, value, unit, vartype in self._postgresql.query(
+                            'SELECT name, pg_catalog.current_setting(name), unit, vartype FROM pg_catalog.pg_settings'
+                            ' WHERE pg_catalog.lower(name) != ALL(%s) AND pending_restart',
+                            [n.lower() for n in params_skip_changes]):
+                        new_value = (lambda v: '?' if v is None else v)(self._postgresql.get_guc_value(param))
```
CyberDem0n, line 1203: Every time it takes a few minutes to understand what is happening here.
Let's make life of @barthisrael easier :)
```suggestion
                        new_value = self._postgresql.get_guc_value(param)
                        new_value = '?' if new_value is None else new_value
```

patroni/postgresql/config.py
```diff
@@ -1233,7 +1269,7 @@ def effective_configuration(self) -> CaseInsensitiveDict:
                 effective_configuration[name] = cvalue
                 logger.info("%s value in pg_controldata: %d, in the global configuration: %d."
                             " pg_controldata value will be used. Setting 'Pending restart' flag", name, cvalue, value)
-                self._postgresql.set_pending_restart(True)
+                self._postgresql.set_pending_restart_reason(CaseInsensitiveDict({name: ParamDiff(cvalue, value)}), True)
```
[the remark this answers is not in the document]
hughcapet: so you mean, we can rely on the knowledge that `set_pending_restart_reason` is always reset before an `effective_configuration` invocation?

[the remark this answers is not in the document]
CyberDem0n: Yes, `effective_configuration` is called when we starting postgres up. That is, we know that pending restart is cleared.

patroni/ctl.py
```diff
@@ -1572,11 +1572,19 @@ def output_members(cluster: Cluster, name: str, extended: bool = False,
             logging.debug(member)
 
             lag = member.get('lag', '')
+
+            def format_diff(param: str, values: Dict[str, str], hide_long: bool):
+                full_diff = param + ': ' + values['old_value'] + '->' + values['new_value']
+                return full_diff if not hide_long or len(full_diff) <= 50 else param + ': [hidden - too long]'
+            restart_reason = '\n'.join(
+                [format_diff(k, v, fmt == 'pretty') for k, v in member.get('pending_restart_reason', {}).items()]) or ''
```
barthisrael, line 1580: Maybe we should "hide longs" on `topology` too?

barthisrael:
👍

CyberDem0n:
:+1:

PR #2980, opened 2023-12-13
Actually allow failover to an async candidate in sync mode

hughcapet:
An oversight in https://github.com/zalando/patroni/pull/2784

I can not explain this:)


CyberDem0n:
:+1:

hughcapet:
👍

PR #2983, opened 2023-12-15
Exclude leader from failover candidates in ctl

hughcapet:
Exclude the actual leader (not the passed leader argument) from the candidates list in the `patronictl failover` prompt. Abort `patronictl failover` execution if candidate specified is the same as the current cluster leader.


CyberDem0n:
:+1:

hughcapet:
👍

PR #2985, opened 2023-12-17
Fix Docker entrypoint to handle multiple ZooKeeper nodes and provide sample docker-compose files for ZK DCS

mkedwards:
I've preserved the quoting convention in the existing documentation, though it seems unlikely that it ever worked:

docs/ENVIRONMENT.rst:-  **PATRONI\_ZOOKEEPER\_HOSTS**: Comma separated list of ZooKeeper cluster members: "'host1:port1','host2:port2','etc...'". It is important to quote every single entity!

I'm using this setup day-to-day with x86_64 Docker images on an aarch64 Mac running Sonoma (14.2), via docker-compose / Colima / VZ / Rosetta.  It's kinda amazing that it works at all; but in fact it seems to work reasonably well.

Here's what I do from a cold start (my needs do not presently include a persistent data volume):

* Start the Colima container hosting environment with `colima start --network-address`

* Watch the Colima logs in another terminal with `tail -F ~/.colima/_lima/colima/serial*.log`

* From a local clone of the patroni github repo: ** In another terminal, fire up the ZooKeeper quorum with `docker compose -f docker-compose-zk-quorum.yml up -d` ** Watch the ZK logs with `docker compose -f docker-compose-zk-quorum.yml logs -f` ** Wait for ZK to spin up (verify via ZooNavigator at http://localhost:9000/editor/data?path=%2Fzookeeper%2Fconfig) ** In another terminal, fire up the Patroni cluster with `docker compose -f docker-compose-zk-patroni.yml up -d` ** Watch the Patroni logs with `docker compose -f docker-compose-zk-patroni.yml logs -f` ** Wait for Patroni to spin up (verify via HAProxy stats at http://localhost:7880/)

* Create Postgres roles, databases, etc.; I habitually use Postico 2 pointed at localhost:5432
* Go to town (note that the Patroni HAProxy container also exposes port 5433 for access to the follower replicas)

### DIFF
--- a/docker/entrypoint.sh
+++ b/docker/entrypoint.sh
@@ -20,10 +20,13 @@ case "$1" in
         haproxy -f /etc/haproxy/haproxy.cfg -p /var/run/haproxy.pid -D
         set -- confd "-prefix=$PATRONI_NAMESPACE/$PATRONI_SCOPE" -interval=10 -backend
         if [ -n "$PATRONI_ZOOKEEPER_HOSTS" ]; then
-            while ! /usr/share/zookeeper/bin/zkCli.sh -server "$PATRONI_ZOOKEEPER_HOSTS" ls /; do
+            while ! /usr/share/zookeeper/bin/zkCli.sh -server `echo "$PATRONI_ZOOKEEPER_HOSTS" | tr -d "'"` ls /; do
                 sleep 1
             done
-            set -- "$@" zookeeper -node "$PATRONI_ZOOKEEPER_HOSTS"
+            set -- "$@" zookeeper
+            for z in `echo "$PATRONI_ZOOKEEPER_HOSTS" | tr -d "'" | tr "," " "`; do
+                set -- "$@" -node "$z"
+            done
         else
             while ! etcdctl member list 2> /dev/null; do
                 sleep 1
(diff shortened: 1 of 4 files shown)


docker/entrypoint.sh
```diff
@@ -20,10 +20,13 @@ case "$1" in
...
         set -- confd "-prefix=$PATRONI_NAMESPACE/$PATRONI_SCOPE" -interval=10 -backend
         if [ -n "$PATRONI_ZOOKEEPER_HOSTS" ]; then
-            while ! /usr/share/zookeeper/bin/zkCli.sh -server "$PATRONI_ZOOKEEPER_HOSTS" ls /; do
+            while ! /usr/share/zookeeper/bin/zkCli.sh -server `echo "$PATRONI_ZOOKEEPER_HOSTS" | tr -d "'"` ls /; do
                 sleep 1
             done
-            set -- "$@" zookeeper -node "$PATRONI_ZOOKEEPER_HOSTS"
+            set -- "$@" zookeeper
+            for z in `echo "$PATRONI_ZOOKEEPER_HOSTS" | tr -d "'" | tr "," " "`; do
+                set -- "$@" -node "$z"
```
CyberDem0n, line 28: It will not pass shellcheck validation at least because of legacy backticks.

CyberDem0n (commented):
I am sorry @mkedwards that I wasn't clear enough saying that we have no interest in making the container more "universal". We are also not really interested in keeping two new docker-compose files in the repo.
Reasons for that are pretty simple:
1. Adding zookeeper (java!) and util-linux bloats the image
2. All this stuff should be properly maintained, and with limited resources we want to avoid it.
3. Especially because in more than 8 years you are the only one who wanted to try Patroni with Zookeeper and docker-compose.

PR #2986, opened 2023-12-18
Update PG version in a couple of places

hughcapet:
* All dockerfiles to use PG16 by default
* PGVERSION env in the test pipelines to 16.1-1 by default
* 11->14 in the dcs-pg mapping for test pipelines
* Code comments fixes


CyberDem0n:
:+1:

hughcapet:
👍

PR #2989, opened 2023-12-19
Implement validation of the `log` section

CyberDem0n:
Somehow it was always forgotten.

### DIFF
--- a/patroni/validator.py
+++ b/patroni/validator.py
@@ -937,6 +937,18 @@ validate_etcd = {
 schema = Schema({
     "name": str,
     "scope": str,
+    Optional("log"): {
+        Optional("level"): EnumValidator(('DEBUG', 'INFO', 'WARN', 'WARNING', 'ERROR', 'FATAL', 'CRITICAL'),
+                                         case_sensitive=True, raise_assert=True),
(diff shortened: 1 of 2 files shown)


patroni/validator.py
```diff
@@ -937,6 +937,18 @@ def validate_watchdog_mode(value: Any) -> None:
 schema = Schema({
     "name": str,
     "scope": str,
+    Optional("log"): {
+        Optional("level"): EnumValidator(('DEBUG', 'INFO', 'WARN', 'WARNING', 'ERROR', 'FATAL', 'CRITICAL'),
+                                         case_sensitive=True, raise_assert=True),
+        Optional("traceback_level"): EnumValidator(('DEBUG', 'ERROR'), raise_assert=True),
```
hughcapet, line 943: why is this one case-insensitive?

CyberDem0n: Because it performs case-insensitive checks here: https://github.com/zalando/patroni/blob/c1ee99d81da8b35e370350a433c5f5c4889baff5/patroni/log.py#L253-L258

hughcapet:
👍

CyberDem0n:
:+1:

PR #2991, opened 2023-12-21
rename citus_handler to mpp_handler

zhjwpku:
This patch rename citus_handler to mpp_handler, along with some renaming stuff in DCS module, obey the following 5 meanings of terminology _cluster_ in Patroni.

1. PostgreSQL cluster: a cluster of postgresql instances which have the same system identifier.
2. MPP cluster: a cluster of PostgreSQL clusters that one of them acts as Coodinator and others act as workers.
3. Coordinator cluster: a PostgreSQL cluster which act the role of 'coordinator' within a MPP cluster.
4. Worker cluster: a PostgreSQL cluster which act the role 'worker' within a MPP cluster.
5. Patroni cluster: all cluster managed by Patroni can be called Patroni cluster, but we usually use this term to refering a single PostgreSQL cluster or an MPP cluster.

### DIFF
--- a/patroni/api.py
+++ b/patroni/api.py
@@ -1153,6 +1153,13 @@ class RestApiHandler(BaseHTTPRequestHandler):
     def do_POST_citus(self) -> None:
         """Handle a ``POST`` request to ``/citus`` path.
 
+        Deprecated, dispatch to do_POST_mpp
+        """
+        self.do_POST_mpp()
+
+    def do_POST_mpp(self) -> None:
+        """Handle a ``POST`` request to ``/mpp`` path.
+
         Call :func:`~patroni.postgresql.mpp.AbstractMPPHandler.handle_event` to handle the request,
         then write a response with HTTP status code ``200``.
 
@@ -1164,9 +1171,9 @@ class RestApiHandler(BaseHTTPRequestHandler):
             return
 
         patroni = self.server.patroni
-        if patroni.postgresql.citus_handler.is_coordinator() and patroni.ha.is_leader():
+        if patroni.postgresql.mpp_handler.is_coordinator() and patroni.ha.is_leader():
             cluster = patroni.dcs.get_cluster()
-            patroni.postgresql.citus_handler.handle_event(cluster, request)
+            patroni.postgresql.mpp_handler.handle_event(cluster, request)
         self.write_response(200, 'OK')
 
     def parse_request(self) -> bool:
--- a/patroni/api.py
+++ b/patroni/api.py
@@ -1153,7 +1153,8 @@ class RestApiHandler(BaseHTTPRequestHandler):
     def do_POST_citus(self) -> None:
         """Handle a ``POST`` request to ``/citus`` path.
 
-        We keep this entrypoint and dispatch the request to do_POST_mpp.
+        .. note::
+            We keep this entrypoint for backward compatibility and simply dispatch the request to :meth:`do_POST_mpp`.
         """
         self.do_POST_mpp()
 
--- a/patroni/api.py
+++ b/patroni/api.py
@@ -1153,7 +1153,7 @@ class RestApiHandler(BaseHTTPRequestHandler):
     def do_POST_citus(self) -> None:
         """Handle a ``POST`` request to ``/citus`` path.
 
-        Deprecated, dispatch to do_POST_mpp
+        We keep this entrypoint and dispatch the request to do_POST_mpp.
         """
         self.do_POST_mpp()
 
--- a/patroni/dcs/__init__.py
+++ b/patroni/dcs/__init__.py
@@ -1532,8 +1532,8 @@ class AbstractDCS(abc.ABC):
     def get_mpp_coordinator(self) -> Optional[Cluster]:
         """Load the PostgreSQL cluster for the MPP Coordinator.
 
-          .. note::
-              This method is only executed on the worker nodes to find the coordinator.
+        .. note::
+            This method is only executed on the worker nodes to find the coordinator.
 
         :returns: Select :class:`Cluster` instance associated with the MPP Coordinator group ID.
         """
--- a/patroni/dcs/__init__.py
+++ b/patroni/dcs/__init__.py
@@ -799,7 +799,7 @@ class Cluster(NamedTuple('Cluster',
     :ivar sync: reference to :class:`SyncState` object, last observed synchronous replication state.
     :ivar history: reference to `TimelineHistory` object.
     :ivar failsafe: failsafe topology. Node is allowed to become the leader only if its name is found in this list.
-    :ivar workers: dictionary of workers of the MPP cluster, optional.  Each key representing the group and the
+    :ivar workers: dictionary of workers of the MPP cluster, optional. Each key representing the group and the
                    corresponding value is a :class:`Cluster` instance.
     """
 
--- a/patroni/dcs/__init__.py
+++ b/patroni/dcs/__init__.py
@@ -1475,7 +1475,7 @@ class AbstractDCS(abc.ABC):
     def _postgresql_cluster_loader(self, path: Any) -> Cluster:
         """Load and build the :class:`Cluster` object from DCS, which represents a single PostgreSQL cluster.
 
-        :param path: the path in DCS where to load Cluster(s) from.
+        :param path: the path in DCS where to load :class:`Cluster` from.
 
         :returns: :class:`Cluster` instance.
         """
@@ -1486,8 +1486,7 @@ class AbstractDCS(abc.ABC):
 
         :param path: the path in DCS where to load Cluster(s) from.
 
-        :returns: all MPP groups as :class:`dict`, with group IDs as keys and :class:`Cluster` objects as values or a
-                  :class:`Cluster` object representing the coordinator with filled `Cluster.workers` attribute.
+        :returns: all MPP groups as :class:`dict`, with group IDs as keys and :class:`Cluster` objects as values.
         """
 
     @abc.abstractmethod
--- a/patroni/dcs/__init__.py
+++ b/patroni/dcs/__init__.py
@@ -782,7 +782,7 @@ class Cluster(NamedTuple('Cluster',
                           ('history', Optional[TimelineHistory]),
                           ('failsafe', Optional[Dict[str, str]]),
                           ('workers', Dict[int, 'Cluster'])])):
-    """Immutable object (namedtuple) which represents PostgreSQL or Citus cluster.
+    """Immutable object (namedtuple) which represents PostgreSQL or MPP cluster.
 
     .. note::
         We are using an old-style attribute declaration here because otherwise it is not possible to override `__new__`
@@ -799,8 +799,8 @@ class Cluster(NamedTuple('Cluster',
     :ivar sync: reference to :class:`SyncState` object, last observed synchronous replication state.
     :ivar history: reference to `TimelineHistory` object.
     :ivar failsafe: failsafe topology. Node is allowed to become the leader only if its name is found in this list.
-    :ivar workers: dictionary of workers of the Citus cluster, optional. Each key is an :class:`int` representing
-                   the group, and the corresponding value is a :class:`Cluster` instance.
+    :ivar workers: dictionary of workers of the MPP cluster, optional.  Each key representing the group and the
+                   corresponding value is a :class:`Cluster` instance.
     """
 
     def __new__(cls, *args: Any, **kwargs: Any):
@@ -1263,11 +1263,11 @@ class AbstractDCS(abc.ABC):
     Functional methods that are critical in their timing, required to complete within ``retry_timeout`` period in order
     to prevent the DCS considered inaccessible, each perform construction of complex data objects:
 
-        * :meth:`~AbstractDCS._cluster_loader`:
+        * :meth:`~AbstractDCS._postgresql_cluster_loader`:
             method which processes the structure of data stored in the DCS used to build the :class:`Cluster` object
             with all relevant associated data.
-        * :meth:`~AbstractDCS._citus_cluster_loader`:
-            Similar to above but specifically representing Citus group and workers information.
+        * :meth:`~AbstractDCS._mpp_cluster_loader`:
+            Similar to above but specifically representing MPP group and workers information.
         * :meth:`~AbstractDCS._load_cluster`:
             main method for calling specific ``loader`` method to build the :class:`Cluster` object representing the
             state and topology of the cluster.
@@ -1337,7 +1337,7 @@ class AbstractDCS(abc.ABC):
     _FAILSAFE = 'failsafe'
 
     def __init__(self, config: Dict[str, Any], mpp: 'AbstractMPP') -> None:
-        """Prepare DCS paths, Citus group ID, initial values for state information and processing dependencies.
+        """Prepare DCS paths, MPP object, initial values for state information and processing dependencies.
 
         :ivar config: :class:`dict`, reference to config section of selected DCS.
                       i.e.: ``zookeeper`` for zookeeper, ``etcd`` for etcd, etc...
@@ -1472,8 +1472,8 @@ class AbstractDCS(abc.ABC):
         return self._last_seen
 
     @abc.abstractmethod
-    def _cluster_loader(self, path: Any) -> Cluster:
-        """Load and build the :class:`Cluster` object from DCS, which represents a single Patroni or Citus cluster.
+    def _postgresql_cluster_loader(self, path: Any) -> Cluster:
+        """Load and build the :class:`Cluster` object from DCS, which represents a single PostgreSQL cluster.
 
         :param path: the path in DCS where to load Cluster(s) from.
 
@@ -1481,12 +1481,12 @@ class AbstractDCS(abc.ABC):
         """
 
     @abc.abstractmethod
-    def _citus_cluster_loader(self, path: Any) -> Dict[int, Cluster]:
-        """Load and build all Patroni clusters from a single Citus cluster.
+    def _mpp_cluster_loader(self, path: Any) -> Dict[int, Cluster]:
+        """Load and build all PostgreSQL clusters from a single MPP cluster.
 
         :param path: the path in DCS where to load Cluster(s) from.
 
-        :returns: all Citus groups as :class:`dict`, with group IDs as keys and :class:`Cluster` objects as values or a
+        :returns: all MPP groups as :class:`dict`, with group IDs as keys and :class:`Cluster` objects as values or a
                   :class:`Cluster` object representing the coordinator with filled `Cluster.workers` attribute.
         """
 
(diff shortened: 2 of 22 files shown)


patroni/ha.py
```diff
@@ -326,20 +326,20 @@ def get_effective_tags(self) -> Dict[str, Any]:
...
                 try:
                     data = {'type': event,
-                            'group': self.state_handler.citus_handler.group,
+                            'group': self.state_handler.mpp_handler.group,
                             'leader': self.state_handler.name,
                             'timeout': self.dcs.ttl,
                             'cooldown': self.patroni.config['retry_timeout']}
                     timeout = self.dcs.ttl if event == 'before_demote' else 2
-                    self.patroni.request(coordinator.leader.member, 'post', 'citus', data, timeout=timeout, retries=0)
+                    self.patroni.request(coordinator.leader.member, 'post', 'mpp', data, timeout=timeout, retries=0)
```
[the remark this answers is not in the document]
zhjwpku: Yeah, that's a problem. Should we leave this as 'citus' and leave a comment describing this for now, and change this later when major version changes?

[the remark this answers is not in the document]
CyberDem0n: There are two options:
1. Check Patroni version from the `leader.member`, and make a decision based on it.
2. Make a decision based on mpp_handler class name, and call /citus endpoint in case of Citus.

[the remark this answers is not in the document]
zhjwpku: Thanks for the advice, I kind of prefer the first option, since finally we would get rid of /citus endpoint. So I will implement option 1 for now and wait for more comments.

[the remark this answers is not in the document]
CyberDem0n: > finally we would get rid of /citus endpoint

We will not. Keeping it costs nothing, but make code backward compatible.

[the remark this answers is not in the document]
zhjwpku: > > finally we would get rid of /citus endpoint
> 
> We will not. Keeping it costs nothing, but make code backward compatible.

Make sense, then I guess option 2 should be more reasonable, citus use /citus endpoint,
all other mpp clusters use /mpp.

patroni/dcs/__init__.py
```diff
@@ -1517,39 +1518,40 @@ def __get_patroni_cluster(self, path: Optional[str] = None) -> Cluster:
...
+              This method is only executed on the worker nodes to find the coordinator.
 
         :returns: Select :class:`Cluster` instance associated with the MPP Coordinator group ID.
         """
         try:
-            return self.__get_patroni_cluster(f'{self._base_path}/{self._mpp.coordinator_group_id}/')
+            return self.__get_postgresql_cluster(f'{self._base_path}/{self._mpp.coordinator_group_id}/')
         except Exception as e:
-            logger.error('Failed to load Citus coordinator cluster from %s: %r', self.__class__.__name__, e)
+            logger.error('Failed to load %s coordinator cluster from %s: %r' %
```
zhjwpku, line 1544: @CyberDem0n I'm confused when adding another %s to the error str, if I just add a param seperated by the comma, the error string won't rendered as expected, so I add a % with parentheses(Chatgpt recomment using format or f-strings, but I'm not sure if there will be some compatibility problems). Can you share some experiences?

patroni/postgresql/mpp/__init__.py
```diff
@@ -65,6 +65,14 @@ def group(self) -> Any:
     def coordinator_group_id(self) -> Any:
         """The group id of the coordinator PostgreSQL cluster."""
 
+    @property
+    @abc.abstractmethod
+    def type(self) -> str:
+        """The type of the MPP cluster.
+
+        :returns: A string representation of the type of a given MPP implementation.
+        """
```
CyberDem0n, line 74: I would suggest doing something like:
```suggestion
    @property
    def type(self) -> str:
        """The type of the MPP cluster.

        :returns: A string representation of the type of a given MPP implementation.
        """
        for base in self.__class__.__bases__:
            if not base.__name__.startswith('Abstract'):
                return base.__name__
        return self.__class__.__name__
```

This way we don't have to create additional properties in subclasses.

patroni/ha.py
```diff
@@ -326,21 +326,26 @@ def get_effective_tags(self) -> Dict[str, Any]:
...
                             'leader': self.state_handler.name,
                             'timeout': self.dcs.ttl,
                             'cooldown': self.patroni.config['retry_timeout']}
                     timeout = self.dcs.ttl if event == 'before_demote' else 2
-                    self.patroni.request(coordinator.leader.member, 'post', 'citus', data, timeout=timeout, retries=0)
+                    if mpp_handler.type == 'Citus':
+                        self.patroni.request(coordinator.leader.member, 'post', 'citus',
+                                             data, timeout=timeout, retries=0)
+                    else:
+                        self.patroni.request(coordinator.leader.member, 'post', 'mpp', data, timeout=timeout, retries=0)
```
CyberDem0n, line 345:
```suggestion
                    endpoint = 'citus' if mpp_handler.type == 'Citus' else 'mpp'
                    self.patroni.request(coordinator.leader.member, 'post', endpoint, data, timeout=timeout, retries=0)
```

zhjwpku:
> I assume changes in ctl.py will follow in a separate PR?

Yeah, that's the third step of the refactor ;)

tests/test_kubernetes.py
```diff
@@ -264,11 +264,11 @@ def test__get_citus_cluster(self):
 
     @patch('patroni.dcs.kubernetes.logger.error')
     def test_get_citus_coordinator(self, mock_logger):
```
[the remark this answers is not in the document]
zhjwpku: good catch, fixed

tests/test_zookeeper.py
```diff
@@ -186,9 +186,9 @@ def test__get_citus_cluster(self):
             self.assertIsInstance(cluster.workers[1], Cluster)
 
     @patch('patroni.dcs.zookeeper.logger.error')
-    @patch.object(ZooKeeper, '_cluster_loader', Mock(side_effect=Exception))
+    @patch.object(ZooKeeper, '_postgresql_cluster_loader', Mock(side_effect=Exception))
     def test_get_citus_coordinator(self, mock_logger):
```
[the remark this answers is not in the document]
zhjwpku: Yeah, good point, I added the assertion and refactored the code a bit.

hughcapet:
👍

patroni/dcs/kubernetes.py
```diff
@@ -746,7 +746,8 @@ def is_ready(self) -> bool:
 
 class Kubernetes(AbstractDCS):
 
-    _CITUS_LABEL = 'citus-group'
+    # For backward compatibility, we keep citus-group here.
+    _MPP_GROUP_LABEL = 'citus-group'
```
CyberDem0n, line 750: Should we maybe introduce another property in AbstractMPP for that?
```python
@property
def k8s_group_label(self):
    return self.type.lower() + '-group'.
```

zhjwpku: I think this makes perfect sense, thanks for pointing out.

zhjwpku:
@CyberDem0n @hughcapet @barthisrael Hey guys, I'd appreciate if you take a look at this PR again :)

patroni/api.py
```diff
@@ -1153,6 +1153,13 @@ def do_POST_switchover(self) -> None:
     def do_POST_citus(self) -> None:
         """Handle a ``POST`` request to ``/citus`` path.
 
+        We keep this entrypoint and dispatch the request to do_POST_mpp.
```
barthisrael, line 1156:
```suggestion
        .. note::
            We keep this entrypoint for backward compatibility and simply dispatch the request to :meth:`do_POST_mpp`.
```

patroni/dcs/__init__.py
barthisrael: What issue are you getting?
The way it was before (only tuple of args instead of `%` formatting) is the usual way of using `logging.error` and similar functions.

patroni/dcs/kubernetes.py
```diff
@@ -979,14 +977,14 @@ def _load_cluster(
...
+    def get_mpp_coordinator(self) -> Optional[Cluster]:
         try:
-            ret = self.__load_cluster(str(self._mpp.coordinator_group_id), self._cluster_loader)
+            ret = self.__load_cluster(str(self._mpp.coordinator_group_id), self._postgresql_cluster_loader)
             if TYPE_CHECKING:  # pragma: no cover
                 assert isinstance(ret, Cluster)
             return ret
         except Exception as e:
-            logger.error('Failed to load Citus coordinator cluster from Kubernetes: %r', e)
+            logger.error('Failed to load %s coordinator cluster from Kubernetes: %r' % (self._mpp.type, e))
```
barthisrael, line 987: Same comment as previous.

patroni/ha.py
```diff
@@ -326,21 +326,23 @@ def get_effective_tags(self) -> Dict[str, Any]:
...
                     data = {'type': event,
-                            'group': self.state_handler.citus_handler.group,
+                            'group': mpp_handler.group,
                             'leader': self.state_handler.name,
                             'timeout': self.dcs.ttl,
                             'cooldown': self.patroni.config['retry_timeout']}
                     timeout = self.dcs.ttl if event == 'before_demote' else 2
-                    self.patroni.request(coordinator.leader.member, 'post', 'citus', data, timeout=timeout, retries=0)
+                    endpoint = 'citus' if mpp_handler.type == 'Citus' else 'mpp'
+                    self.patroni.request(coordinator.leader.member, 'post', endpoint, data, timeout=timeout, retries=0)
```
barthisrael, line 342: Do we need to do this?
It seems like an overkill to me as `/citus` is a simple redirect to `/mpp`.
I'd say to keep the `/citus` available for backward compatibility for the users, but we (Patroni) don't need to make that distinction.

patroni/ha.py
```diff
@@ -326,21 +326,23 @@ def get_effective_tags(self) -> Dict[str, Any]:
...
                 except Exception as e:
-                    logger.warning('Request to Citus coordinator leader %s %s failed: %r',
-                                   coordinator.leader.name, coordinator.leader.member.api_url, e)
+                    logger.warning('Request to %s coordinator leader %s %s failed: %r' % (mpp_handler.type,
+                                   coordinator.leader.name, coordinator.leader.member.api_url, e))
```
barthisrael, line 345: Same comment about formatting on logging.

patroni/postgresql/mpp/__init__.py
```diff
@@ -65,6 +65,21 @@ def group(self) -> Any:
...
+        for base in self.__class__.__bases__:
+            if not base.__name__.startswith('Abstract'):
+                return base.__name__
+        return self.__class__.__name__
+
+    @property
+    def k8s_group_label(self):
```
barthisrael, line 80: Missing docstring and type annotation.

patroni/dcs/consul.py
```diff
@@ -420,7 +420,7 @@ def _cluster_from_nodes(self, nodes: Dict[str, Any]) -> Cluster:
     def _consistency(self) -> str:
         return 'consistent' if self._ctl else self._client.consistency
 
-    def _cluster_loader(self, path: str) -> Cluster:
+    def _postgresql_cluster_loader(self, path: str) -> Cluster:
```
barthisrael, line 423: What about adding a docstring if you are touching this code? :)

patroni/dcs/consul.py
```diff
@@ -431,7 +431,7 @@ def _cluster_loader(self, path: str) -> Cluster:
 
         return self._cluster_from_nodes(nodes)
 
-    def _citus_cluster_loader(self, path: str) -> Dict[int, Cluster]:
+    def _mpp_cluster_loader(self, path: str) -> Dict[int, Cluster]:
```
barthisrael, line 434: What about adding a docstring if you are touching this code? :)

patroni/dcs/etcd.py
```diff
@@ -710,15 +710,15 @@ def _cluster_from_nodes(self, etcd_index: int, nodes: Dict[str, etcd.EtcdResult]
 
         return Cluster(initialize, config, leader, status, members, failover, sync, history, failsafe)
 
-    def _cluster_loader(self, path: str) -> Cluster:
+    def _postgresql_cluster_loader(self, path: str) -> Cluster:
```
barthisrael, line 713: What about adding a docstring if you are touching this code? :)

patroni/dcs/etcd.py
```diff
@@ -710,15 +710,15 @@ def _cluster_from_nodes(self, etcd_index: int, nodes: Dict[str, etcd.EtcdResult]
...
         try:
             result = self.retry(self._client.read, path, recursive=True, quorum=self._ctl)
         except etcd.EtcdKeyNotFound:
             return Cluster.empty()
         nodes = {node.key[len(result.key):].lstrip('/'): node for node in result.leaves}
         return self._cluster_from_nodes(result.etcd_index, nodes)
 
-    def _citus_cluster_loader(self, path: str) -> Dict[int, Cluster]:
+    def _mpp_cluster_loader(self, path: str) -> Dict[int, Cluster]:
```
barthisrael, line 721: What about adding a docstring if you are touching this code? :)

patroni/dcs/etcd3.py
```diff
@@ -733,7 +733,7 @@ def create_lease(self) -> None:
 
     @property
     def cluster_prefix(self) -> str:
-        return self._base_path + '/' if self.is_citus_coordinator() else self.client_path('')
+        return self._base_path + '/' if self.is_mpp_coordinator() else self.client_path('')
```
barthisrael, line 736: What about adding a docstring if you are touching this code? :)

patroni/dcs/zookeeper.py
```diff
@@ -258,11 +258,11 @@ def _cluster_loader(self, path: str) -> Cluster:
 
         return Cluster(initialize, config, leader, status, members, failover, sync, history, failsafe)
 
-    def _citus_cluster_loader(self, path: str) -> Dict[int, Cluster]:
+    def _mpp_cluster_loader(self, path: str) -> Dict[int, Cluster]:
```
barthisrael, line 261: What about adding a docstring if you are touching this code? :)

patroni/ha.py
```diff
@@ -326,21 +326,23 @@ def get_effective_tags(self) -> Dict[str, Any]:
             tags['nosync'] = True
         return tags
 
-    def notify_citus_coordinator(self, event: str) -> None:
-        if self.state_handler.citus_handler.is_worker():
-            coordinator = self.dcs.get_citus_coordinator()
+    def notify_mpp_coordinator(self, event: str) -> None:
```
barthisrael, line 329: What about adding a docstring if you are touching this code? :)

barthisrael (changes requested):
Some comments.
Most of them some "copy and paste" comments regarding taking the opportunity to add docstrings. Some with proposed changes.

patroni/dcs/__init__.py
zhjwpku: Here is what I got when using the way before:
image

The placeholders seems not rendered, which is really confusing to me.

patroni/dcs/etcd3.py
zhjwpku: I'd appreciate if you can help with the wording ;)

patroni/ha.py
zhjwpku: Wording help needed here to :)

patroni/dcs/etcd3.py
```diff
@@ -733,7 +733,11 @@ def create_lease(self) -> None:
 
     @property
     def cluster_prefix(self) -> str:
-        return self._base_path + '/' if self.is_citus_coordinator() else self.client_path('')
+        """Construct the cluster prefix for the cluster.
+
+        :returns: A string representation the cluster prefix.
```
barthisrael, line 738:
```suggestion
        :returns: path in the DCS under which we store information about this Patroni cluster.
```

patroni/dcs/etcd3.py
barthisrael: Put a suggestion.

patroni/ha.py
barthisrael: I like it, thanks!

patroni/ha.py
barthisrael: Oh, that's a good point actually. Thanks for clarifying!

patroni/dcs/__init__.py
barthisrael: When you use the `%` string it formats the string before calling the mocked function (`logger.error`). In that case you are able to assert against the expected string (`Failed to load Null coordinator cluster from ZooKeeper`) as the first argument, because that is indeed the first argument (the output of a formatted string).

If you call the `logger.error` with arguments and delegate the formatting task to the `logging` module, then you need to adjust the mock assertions accordingly. For example:

```diff
diff --git a/patroni/dcs/__init__.py b/patroni/dcs/__init__.py
index da2d8dcc..481a8e6c 100644
--- a/patroni/dcs/__init__.py
+++ b/patroni/dcs/__init__.py
@@ -1540,8 +1540,8 @@ class AbstractDCS(abc.ABC):
         try:
             return self.__get_postgresql_cluster(f'{self._base_path}/{self._mpp.coordinator_group_id}/')
         except Exception as e:
-            logger.error('Failed to load %s coordinator cluster from %s: %r' %
-                         (self._mpp.type, self.__class__.__name__, e))
+            logger.error('Failed to load %s coordinator cluster from %s: %r',
+                         self._mpp.type, self.__class__.__name__, e)
         return None

     def _get_mpp_cluster(self) -> Cluster:
diff --git a/tests/test_zookeeper.py b/tests/test_zookeeper.py
index 797330d5..301925d0 100644
--- a/tests/test_zookeeper.py
+++ b/tests/test_zookeeper.py
@@ -191,7 +191,10 @@ class TestZooKeeper(unittest.TestCase):
         with patch.object(ZooKeeper, '_postgresql_cluster_loader', Mock(side_effect=Exception)):
             self.assertIsNone(self.zk.get_mpp_coordinator())
             mock_logger.assert_called_once()
-            self.assertIn('Failed to load Null coordinator cluster from ZooKeeper', mock_logger.call_args[0][0])
+            self.assertEqual(mock_logger.call_args[0][0], 'Failed to load %s coordinator cluster from %s: %r')
+            self.assertEqual(mock_logger.call_args[0][1], 'Null')
+            self.assertEqual(mock_logger.call_args[0][2], 'ZooKeeper')
+            self.assertIsInstance(mock_logger.call_args[0][3], ZooKeeperError)

     @patch('patroni.dcs.logger.error')
     def test_get_citus_coordinator(self, mock_logger):
```

I mean, with your approach you are formatting the string before calling the `logger.error`, and with the previous approach the code was delegating the formatting to `logging` module, so it would occur inside of `logging` module.

I suggest to use the approach that delegates to `logging`, which from what I can see is used through the Patroni code base more often -- and actually what the `logging` module suggests in their examples.

barthisrael: The same thought applies to other similar changes in the PR.

barthisrael (changes requested):
Some more comments.

barthisrael:
👍

CyberDem0n:
:+1:

PR #2992, opened 2023-12-21
Not filter out contradictory nofailover tag

hughcapet:
* Ensure that `nofailover` will always be used if both `nofailover` and
`failover_priority` tags are provided
* Call `_validate_failover_tags` from  `reload_local_configuration()` as well
* Properly check values in the `_validate_failover_tags()`: `nofailover` value should be casted to boolean like it is done when accessed in other places

### DIFF
--- a/features/steps/patroni_api.py
+++ b/features/steps/patroni_api.py
@@ -128,6 +128,11 @@ def scheduled_restart(context, url, in_seconds, data):
     context.execute_steps(u"""Given I issue a POST request to {0}/restart with {1}""".format(url, json.dumps(data)))
 
 
+@step('I set {tag:w} tag in {pg_name:w} config')
+def add_bool_tag_to_config(context, tag, pg_name):
+    context.pctl.add_tag_to_config(pg_name, tag, True)
+
+
 @step('I add tag {tag:w} {value:w} to {pg_name:w} config')
 def add_tag_to_config(context, tag, value, pg_name):
     context.pctl.add_tag_to_config(pg_name, tag, value)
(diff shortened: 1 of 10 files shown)


patroni/config.py
```diff
@@ -814,10 +815,12 @@ def _validate_failover_tags(self) -> None:
           bedrock source of truth)
         """
         tags = self.get('tags', {})
+        if 'nofailover' not in tags:
+            return
         nofailover_tag = tags.get('nofailover')
```
CyberDem0n, line 820:
```suggestion
        nofailover_tag = bool(tags.get('nofailover'))
```

patroni/config.py
```diff
@@ -814,10 +815,12 @@ def _validate_failover_tags(self) -> None:
...
         failover_priority_tag = parse_int(tags.get('failover_priority'))
         if failover_priority_tag is not None \
-                and (nofailover_tag is True and failover_priority_tag > 0
-                     or nofailover_tag is False and failover_priority_tag <= 0):
+                and (bool(nofailover_tag) is True and failover_priority_tag > 0
+                     or bool(nofailover_tag) is False and failover_priority_tag <= 0):
```
CyberDem0n, line 824:
```suggestion
                and (nofailover_tag is True and failover_priority_tag > 0
                     or nofailover_tag is False and failover_priority_tag <= 0):
```

features/steps/patroni_api.py
```diff
@@ -128,6 +128,12 @@ def scheduled_restart(context, url, in_seconds, data):
     context.execute_steps(u"""Given I issue a POST request to {0}/restart with {1}""".format(url, json.dumps(data)))
 
 
+@step('I {action:w} {tag:w} tag in {pg_name:w} config')
+def add_bool_tag_to_config(context, action, tag, pg_name):
+    value = True if action == 'set' else False
```
CyberDem0n, line 133:
```suggestion
    value = action == 'set'
```

patroni/config.py
hughcapet: we need the actual value for the log

CyberDem0n:
:+1:

hughcapet:
👍

Issue #2995, opened 2023-12-28
Pyinstaller bundled patroni started with multiprocessing-fork error


XiuhuaRuan:
### What happened?

When started patroni bundled by pyinstaller , it reported below multiprocessing related error.
```
usage: patroni [-h] [--version] [--validate-config] [configfile]
patroni: error: unrecognized arguments: --multiprocessing-fork pipe_handle=11
```

### How can we reproduce it (as minimally and precisely as possible)?

1.  Execute mkbinary in source code to reproduce a patroni binary
2.  Starting the patroni binary could reproduce it 

### What did you expect to happen?

Starting patroni should succeed without error.

### Patroni/PostgreSQL/DCS version

-   Patroni version: 3.1.0, 3.2.1
-   PostgreSQL version: 14.0
-   DCS (and its version): etcd3.5.9

### Patroni configuration file

```yaml
scope: postgres-cluster
namespace: /service/
name: postgres_01

restapi:
  listen: 192.168.61.105:8008
  connect_address: 192.168.61.105:8008

etcd:
  hosts: 192.168.61.105:2379,192.168.61.106:2379,192.168.61.107:2379

log:
  level: INFO
  traceback_level: INFO
  dir: /home/postgres/patroni
  file_num: 10
  file_size: 104857600

bootstrap:
  dcs:
    ttl: 30
    loop_wait: 10
    retry_timeout: 10
    maximum_lag_on_failover: 1048576
    master_start_timeout: 300
    synchronous_mode: false
    postgresql:
      use_pg_rewind: true
      parameters:
        listen_addresses: "*"
        port: 5432
        wal_level: replica
        hot_standby: "on"
        wal_keep_size: 100
        max_wal_senders: 10
        max_replication_slots: 10
        wal_log_hints: "on"
        archive_mode: "off"
        archive_timeout: 1800s
        logging_collector: on
        log_destination: 'stderr'
        log_truncate_on_rotation: on
        log_checkpoints: on
        log_connections: on
        log_disconnections: on
        log_error_verbosity: default
        log_lock_waits: on
        log_temp_files: 0
        log_autovacuum_min_duration: 0
        log_min_duration_statement: 50
        log_timezone: 'PRC'
        log_filename: postgresql-%Y-%m-%d_%H.log
        log_line_prefix: '%t [%p]: db=%d,user=%u,app=%a,client=%h '


postgresql:
  database: postgres
  listen: 0.0.0.0:5432
[...]

PR #2996, opened 2023-12-28
call freeze_support in main module to solve pyinstaller frozen issue

XiuhuaRuan:
To solve issue #2995, we need to call multiprocessing.freeze_support() straight after the if __name__ == __main__ line of the main module. Refer to https://docs.python.org/3/library/multiprocessing.html#multiprocessing.freeze_support

In the fix for issue #1441, freeze_support was added to patroni_main() but it doesn't work now since v3.1.0. I moved this part of code to follow the if __name__ == __main__ line and verified OK. Besides, as this issue also occured on linux I removed 'Windows' from the comment.

### DIFF
--- a/patroni/__main__.py
+++ b/patroni/__main__.py
@@ -229,11 +229,6 @@ def patroni_main(configfile: str) -> None:
 
     :param configfile: path to Patroni configuration file.
     """
-    from multiprocessing import freeze_support
-
-    # Windows executables created by PyInstaller are frozen, thus we need to enable frozen support for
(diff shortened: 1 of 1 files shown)


hughcapet:
Let's move `freeze_support()` inside `main()` so that we avoid decrease of the test coverage

XiuhuaRuan:
@hughcapet Thanks for your comment. I have moved freeze_support() inside main() and verified OK.

CyberDem0n:
:+1:

hughcapet:
👍

PR #3000, opened 2024-01-03
Close the doors for a possible future bug in the config generator

barthisrael:
The `AbstractConfigGenerator._format_config` method was missing a comma in the declaration of a tuple. As a consequence it was concatenating the strings `ctl` and `citus` instead of creating two separate items in the tuple.

There is currently no observed bug from that issue in the code because the template configuration created by the method `AbstractConfigGenerator.get_template_config` doesn't include either of `ctl` or `citus` keys.

However, it is still important that we close the doors for possible future bugs that would come up if we ever attempt to use either of those keys in the template, for example.

References: PAT-231.

### DIFF
--- a/patroni/config_generator.py
+++ b/patroni/config_generator.py
@@ -178,7 +178,7 @@ class AbstractConfigGenerator(abc.ABC):
 
         :yields: formatted lines or blocks that represent a text output of the YAML document.
         """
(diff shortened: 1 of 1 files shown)


hughcapet:
👍

CyberDem0n:
:+1:

PR #3003, opened 2024-01-11
Fixes around pending_restart flag

hughcapet:
* Do not set `pending_restart` flag if `hot_standby` is set to 'off' during a custom bootstrap (even though we will have this flag actually set in PG, this configuration parameter is irrelevant on primary and there is no actual need for restart)
* Skip `hot_standby` and `wal_log_hints` when querying parameters pending restart on config reload. They actually can be changed manually (e.g. via `ALTER SYSTEM`) and it will cause the pending_restart state in PG but Patroni anyway always passes those params to postmaster as command line options. And there they only can have one value - 'on' (except on primary when performing custom bootstrap)

### DIFF
--- a/patroni/postgresql/config.py
+++ b/patroni/postgresql/config.py
@@ -645,10 +645,6 @@ class ConfigHandler(object):
                                           'recovery_target_action', 'standby_mode', self._triggerfile_wrong_name})
         return CaseInsensitiveSet(self._RECOVERY_PARAMETERS - skip_params)
 
-    @property
-    def _parameters_skip_changes(self) -> CaseInsensitiveSet:
-        return CaseInsensitiveSet((*self._RECOVERY_PARAMETERS, 'hot_standby', 'wal_log_hints'))
-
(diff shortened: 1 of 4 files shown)


patroni/postgresql/config.py
```diff
@@ -1169,7 +1170,7 @@ def reload_config(self, config: Dict[str, Any], sighup: bool = False) -> None:
                     pending_restart = self._postgresql.query(
                         'SELECT COUNT(*) FROM pg_catalog.pg_settings'
                         ' WHERE pg_catalog.lower(name) != ALL(%s) AND pending_restart',
-                        [n.lower() for n in self._RECOVERY_PARAMETERS])[0][0] > 0
+                        [n for n in params_skip_changes])[0][0] > 0
```
CyberDem0n, line 1173: Here we actually do need lower case comparison.
```suggestion
                        [n.lower() for n in params_skip_changes])[0][0] > 0
```

hughcapet: 🤦‍♀️  went too far

CyberDem0n:
:+1:

hughcapet:
👍

PR #3004, opened 2024-01-15
Optimize priority failover behave tests

CyberDem0n:
1. get rid of useless sleep calls
2. call `POST /failover` on the node where we want to failover to


CyberDem0n:
:+1:

PR #3006, opened 2024-01-16
Fix bugs introduced in the jsonlog implementation

CyberDem0n:
1. RotatingFileHandler is a child of StreamHandler, therefore we can't rely on `not isinstance(handler, logging.StreamHandler)`.
2. If the legacy version of `python-json-logger` is installed (that doesn't support rename_fields or static_fields), we want do what is possible rather than fail with the exception.

Besides that:
1. improve code coverage
2. make unit tests pass without python-json-logger installed or if only some old version is installed.

### DIFF
--- a/tests/test_log.py
+++ b/tests/test_log.py
@@ -3,13 +3,21 @@ import os
 import sys
 import unittest
 import yaml
-import json
 from io import StringIO
 
 from mock import Mock, patch
 from patroni.config import Config
 from patroni.log import PatroniLogger
 from queue import Queue, Full
+
+try:
+    from pythonjsonlogger import jsonlogger
+    import json
+    json_formatter_is_available = True
+    static_fields_supported = ':param static_fields:' in jsonlogger.JsonFormatter.__init__.__doc__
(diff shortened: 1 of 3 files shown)


tests/test_log.py
```diff
@@ -3,14 +3,22 @@
...
 from io import StringIO
 
 from mock import Mock, patch
 from patroni.config import Config
 from patroni.log import PatroniLogger
 from queue import Queue, Full
 
+try:
+    from pythonjsonlogger import jsonlogger
+    import json
```
hughcapet, line 15: are we also checking for a possible(?) json module import error here? or only import when `json_formatter_is_available = True`? if latter, then maybe move it to the end of the "try" block?

hughcapet (commented):
Can we also fix this here?

https://github.com/zalando/patroni/blob/7d2bb08bb39c96eeeba902d6e2031b800f3356f1/patroni/validator.py#L55

hughcapet:
👍

CyberDem0n:
:+1:

PR #3015, opened 2024-02-01
Patroni doesn't filter out some not allowed options from `pg_basebackup`

barthisrael:
When running `pg_basebackup` to bootstrap a replica, Patroni sanitizes the custom user options that come from `postgresql.basebackup` configuration section using the `process_user_options` method.

However, there is a bug in that method: it filters out not allowed options that are in the format `- setting`, but not the ones in the format `- setting: value` from `postgresql.basebackup`.

An example of that issue is the `dbname` setting. If you specify something like this in the configuration file:

```yaml
postgresql:
  basebackup:
    - dbname: "host=RANDOM"
```

You end up with `--dbname` being specified twice for `pg_basebackup`, with `--dbname='host=RANDOM'` taking precedence as it comes up later in the command.

This commit fixes that issue by adding a `continue` statement when the setting in format `- setting: value` is not allowed, thus skipping it.

References: PAT-235.

### DIFF
--- a/patroni/postgresql/bootstrap.py
+++ b/patroni/postgresql/bootstrap.py
@@ -103,6 +103,7 @@ class Bootstrap(object):
                     if len(keys) != 1 or not isinstance(opt[keys[0]], str) or not option_is_allowed(keys[0]):
                         error_handler('Error when parsing {0} key-value option {1}: only one key-value is allowed'
                                       ' and value should be a string'.format(tool, opt[keys[0]]))
+                        continue
                     user_options.append('--{0}={1}'.format(keys[0], unquote(opt[keys[0]])))
                 else:
                     error_handler('Error when parsing {0} option {1}: value should be string value'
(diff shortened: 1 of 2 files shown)


patroni/postgresql/bootstrap.py
```diff
@@ -103,6 +103,7 @@ def option_is_allowed(name: str) -> bool:
                     if len(keys) != 1 or not isinstance(opt[keys[0]], str) or not option_is_allowed(keys[0]):
                         error_handler('Error when parsing {0} key-value option {1}: only one key-value is allowed'
                                       ' and value should be a string'.format(tool, opt[keys[0]]))
+                        continue
                     user_options.append('--{0}={1}'.format(keys[0], unquote(opt[keys[0]])))
```
CyberDem0n, line 107: I would prefer to get rid of `continue`:
```suggestion
                    if len(keys) == 1 and isinstance(opt[keys[0]], str) and option_is_allowed(keys[0]):
                        user_options.append('--{0}={1}'.format(keys[0], unquote(opt[keys[0]])))
                    else:
                        error_handler('Error when parsing {0} key-value option {1}: only one key-value is allowed'
                                      ' and value should be a string'.format(tool, opt[keys[0]]))
```

hughcapet:
👍

CyberDem0n:
:+1:

PR #3024, opened 2024-02-24
Instruct etcd to delete old revisions

hedleyroos:
Etcd keeps old revisions unless instructed to delete them. If we don't delete old revisions then etcd memory usage will keep growing forever due to keepalive updates. Since Patroni does not really need to roll back to older revisions we can safely delete them.

### DIFF
--- a/docker/entrypoint.sh
+++ b/docker/entrypoint.sh
@@ -38,7 +38,7 @@ EOT
         exec dumb-init "$@"
         ;;
     etcd)
-        exec "$@" --auto-compaction-retention=6 -advertise-client-urls "http://$DOCKER_IP:2379"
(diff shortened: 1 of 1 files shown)


docker/entrypoint.sh
```diff
@@ -38,7 +38,7 @@ EOT
         exec dumb-init "$@"
         ;;
     etcd)
-        exec "$@" -advertise-client-urls "http://$DOCKER_IP:2379"
+        exec "$@" --auto-compaction-retention=6 -advertise-client-urls "http://$DOCKER_IP:2379"
```
hughcapet, line 41:
```suggestion
        exec "$@" --auto-compaction-retention=1 -advertise-client-urls "http://$DOCKER_IP:2379"
```

CyberDem0n:
:+1:

hughcapet:
👍

PR #3029, opened 2024-03-03
add missing busybox install command

zhjwpku:
before this patch, when execute `ps` inside container, we see the following error:

```
postgres@a97c9e438eae:~$ ps
bash: ps: command not found
```


CyberDem0n:
:+1:

PR #3031, opened 2024-03-04
add available_parameters

valiyvvv:
Add ac_ignore_maclabel parameter for Astra postgresql with Mandatory Access Control


CyberDem0n:
Sorry, but NO. Such change will never be accepted. We don't aim to support Postgres forks directly. 0_postgres.yml list only GUCs that are available in PostgreSQL core.
The idea is that you may create another YAML file for your distribution and put it to `available_parameters` directory when you are packaging Patroni or after installing it.

PR #3036, opened 2024-03-20
Extend behave tests with nostream feature

hughcapet:
Check state and permanent logical replication slots behaviour


CyberDem0n:
👍

hughcapet:
👍

PR #3038, opened 2024-03-26
Set global_config from dynamic_config if DCS data is empty

hughcapet:
Fix the oversight of https://github.com/zalando/patroni/commit/193c73f6b80c42b316d790f9b40321da2ee55a31
We need to set global config from the local cache if `cluster.config` is not initialized or doesn't have a valid config.
If there is nothing written into the DCS (yet), we need the setup info for the decision making (e.g., if it is a standby cluster)


CyberDem0n:
:+1:

hughcapet:
👍

Issue #3039, opened 2024-03-28
FATAL:  could not connect to the primary server: connection to server at "x.x.x.x", port 5432 failed: session is read-only


novanbramantya:
### What happened?

When i setup the new patroni cluster for rollback needed, i want to replicate it from standby patroni cluster which is not promoted yet. after basebackup section somehow it not continuing the replication process and got fatal said session is read-only

### How can we reproduce it (as minimally and precisely as possible)?

i have existing patroni postgres cluster in 1 Cloud provider, i need to migrate it to another cloud provider and the state is done replication and set to standby leader. after that i need to setup the rollback patroni cluster. But when i start the patroni service somehow got this error
FATAL:  could not connect to the primary server: connection to server at "x.x.x.x", port 5432 failed: session is read-only
most likely patroni doesn't allow replication from replica instead of leader

### What did you expect to happen?

i need to replicate the data from standby patroni cluster to rollback cluster which is still in read-only session until it got promoted

### Patroni/PostgreSQL/DCS version

-   Patroni version: patroni 3.2.2
-   PostgreSQL version: 12.18
-   DCS (and its version): 3.4

### Patroni configuration file

```yaml
namespace: patroni
scope: test_rollback_postgres
name: test-rollback-postgres-n1
etcd3:
  url: https://test-patroni-etcd.service.consul:2379
  cert: /etc/etcd/etcd.pem
  key: /etc/etcd/etcd-key.pem
  cacert: /etc/etcd/root-ca.pem
postgresql:
  bin_dir: /usr/lib/postgresql/12/bin
  use_unix_socket: true
  listen: 0.0.0.0:5432
  config_dir: /etc/postgresql/patroni
  data_dir: /data/postgres/data/
  pgpass: /var/lib/postgresql/.pgpass
  pg_ctl_timeout: 60
  connect_address: x.x.x.x:5432
  authentication:
    superuser:
      username: admin_user
      password: admin_pass
    replication:
      username: repl_user
      password: repl_pass
    rewind:
      username: repl_user
      password: repl_pass
  basebackup:
    - progress
    - slot: "rollback_master"
    - verbose
  parameters:
    work_mem: 8MB
    archive_timeout: 60
    archive_command: "/usr/local/bin/wal-g-push-wal.sh %p"
    checkpoint_completion_target: 0.7
    checkpoint_timeout: 15min
    hot_standby_feedback: on
    log_checkpoints: on
    log_destination: 'stderr'
[...]

PR #3040, opened 2024-03-28
Use `target_session_attrs` only when multiple hosts in standby_cluster

CyberDem0n:
Actually comment in the code was already saying that, but on practice it didn't happen.

It should help #3039


docs/standby_cluster.rst
```diff
@@ -0,0 +1,72 @@
...
+your responsibility to copy it to PGDATA.
+
+If you use replication slots on the standby cluster, you must also create the
+corresponding replication slot on the primary cluster.  It will not be done
+automatically by the standby cluster implementation.  You can use Patroni's
+permanent replication slots feature on the primary cluster to maintain a
+replication slot with the same name as ``primary_slot_name``, or its default
+value if ``primary_slot_name`` is not provided.
+
+In case if the remote site doesn't provide a single endpoint that connects to a
```
hughcapet, line 63:
```suggestion
In case the remote site doesn't provide a single endpoint that connects to a
```

docs/standby_cluster.rst
```diff
@@ -0,0 +1,72 @@
...
+primary one could lists all hosts of the source cluster in the
```
hughcapet, line 64:
```suggestion
primary, one could list all hosts of the source cluster in the
```

docs/standby_cluster.rst
```diff
@@ -0,0 +1,72 @@
...
+``standby_cluster.host`` section.  When ``standby_cluster.host`` contains
+multiple hosts speareted by comma Patroni will:
```
hughcapet, line 66:
```suggestion
multiple hosts separated by comma Patroni will:
```

docs/standby_cluster.rst
```diff
@@ -0,0 +1,72 @@
...
+
+* add  to the ``primary_conninfo`` on the
```
hughcapet, line 68:
```suggestion
* add to the ``primary_conninfo`` on the
```

docs/standby_cluster.rst
```diff
@@ -0,0 +1,72 @@
...
+  standby leader node.
+* use ``target_session_attrs=read-write`` when trying to detemine whether we
```
hughcapet, line 70:
```suggestion
* use ``target_session_attrs=read-write`` when trying to determine whether we
```

docs/standby_cluster.rst
```diff
@@ -0,0 +1,72 @@
...
+In case the remote site doesn't provide a single endpoint that connects to a
+primary, one could list all hosts of the source cluster in the
+``standby_cluster.host`` section.  When ``standby_cluster.host`` contains
+multiple hosts separated by commas, Patroni will:
+
+* add ``target_session_attrs=read-write`` to the ``primary_conninfo`` on the
+  standby leader node.
+* use ``target_session_attrs=read-write`` when trying to determine whether we
+  need to run ``pg_rewind`` or when executing ``pg_rewind`` on all nodes of the
+  standby cluster.
```
hughcapet, line 72:
```suggestion
  standby cluster.
  
There is also a possibility to replicate the standby cluster from a standby member of the primary cluster: for that, you need to define the standby host in the ``standby_cluster.host`` section. However, you need to beware that in this case ``pg_rewind`` will fail to execute on the standby cluster.

```

hughcapet:
👍

CyberDem0n:
:+1:

PR #3042, opened 2024-04-02
Compatibility with ydiff>=1.3

CyberDem0n:
PatchStream class was removed as it was effectively doing nothing useful.


hughcapet:
👍

CyberDem0n:
:+1:

PR #3057, opened 2024-05-06
Add date to every released version

CyberDem0n:
Going to GH releases and/or tags to get it is not very convinient.


hughcapet:
👍

hughcapet:
👍

Issue #3061, opened 2024-05-10
 List order of sync_standby is  inconsistent with synchronous_standby_names


XiuhuaRuan:
### What happened?

We want to bind standby VIP to the sync standby with highest sync_priority according to the sync_standby list in DCS. But sync_standby list order is inconsistent with synchronous_standby_names. It is sorted by application name not by sync_ priority. Besides, synchronous_standby_names is not stable enough when changing synchronous_node_count.

To make sync_standby and synchronous_standby_names order more stable, I suggest to consider sync_priority as the third sort condition following sync_state and lsn when collecting ReplicaList from pg_stat_replication. Please help evaluate this proposal. Thanks.

### How can we reproduce it (as minimally and precisely as possible)?

1. Set up a cluster with 3 nodes
2. Set synchronous_mode to true via patronictl edit-config, ee_03 was selected as sync standby
3. Set synchronous_node_count to 2 via patronictl edit-config, ee_03 and ee_02 were selected as sync standby
4. check sync standby list 
postgresql.conf:
synchronous_standby_names = '2 (ee_03,ee_02)'
/sync key in etcd:
{"leader":"ee_01","sync_standby":"ee_02,ee_03"}
5. set synchronous_node_count to 1, sometimes ee_02 was selected as sync standby, sometimes ee_03 was selected as sync standby

### What did you expect to happen?

1. sync_standby list order is consistent with synchronous_standby_names 
2. synchronous_standby_names is changed consistently when changing synchronous_node_count. For example, when decrease synchronous_node_count, sync standby with higher sync_priority is expected to remain in the list.

### Patroni/PostgreSQL/DCS version

-   Patroni version: 3.1.0
-   PostgreSQL version: 13.0
-   DCS (and its version): etcd3.5.9

### Patroni configuration file

```yaml
scope: postgres-cluster
namespace: /service/
name: ee_01

restapi:
  listen: 192.168.61.105:8008
  connect_address: 192.168.61.105:8008

etcd:
  hosts: 192.168.61.105:2379,192.168.61.106:2379,192.168.61.107:2379

log:
  level: INFO
  traceback_level: INFO
  dir: /home/postgres/patroni
  file_num: 10
  file_size: 104857600

bootstrap:
  dcs:
    ttl: 30
    loop_wait: 10
    retry_timeout: 10
    maximum_lag_on_failover: 1048576
    master_start_timeout: 300
    synchronous_mode: false
    postgresql:
      use_pg_rewind: true
      parameters:
[...]

PR #3062, opened 2024-05-10
add sync_priority as sorting criteria to sync ReplicaList

XiuhuaRuan:
To make sync_standby list more stable when changing synchronous_node_count, I add sync_priority as sorting criteria to sync ReplicaList following sync_state and lsn. This PR is to solve issue #3061. Please help review if such change is OK and has no potential issues. Thanks.

### DIFF
--- a/patroni/dcs/__init__.py
+++ b/patroni/dcs/__init__.py
@@ -1870,7 +1870,7 @@ class AbstractDCS(abc.ABC):
 
         :returns: dictionary that later could be serialized to JSON or saved directly to DCS.
         """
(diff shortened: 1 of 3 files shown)


CyberDem0n (commented):
There is no value in ordering by `sync_priority`.
The thing is, that Patroni enforces the number of nodes required to be synchronous to always match with the number of nodes listed in the `synchronous_standby_names`. Effectively these nodes are treated equally, no matter what `sync_priority` value they have (which is IIRC assigned from the position in the `synchronous_standby_names`).

Issue #1942, opened 2021-05-28
"wal_log_hints" parameter  is not configurable 


ckh0618:
Hello. 

I just found "wal_log_hints" parameter is not configurable. 

```
    CMDLINE_OPTIONS = CaseInsensitiveDict({
        'listen_addresses': (None, lambda _: False, 90100),
        'port': (None, lambda _: False, 90100),
        'cluster_name': (None, lambda _: False, 90500),
        'wal_level': ('hot_standby', lambda v: v.lower() in ('hot_standby', 'replica', 'logical'), 90100),
        'hot_standby': ('on', lambda _: False, 90100),
        'max_connections': (100, lambda v: int(v) >= 25, 90100),
        'max_wal_senders': (10, lambda v: int(v) >= 3, 90100),
        'wal_keep_segments': (8, lambda v: int(v) >= 1, 90100),
        'wal_keep_size': ('128MB', lambda v: parse_int(v, 'MB') >= 16, 130000),
        'max_prepared_transactions': (0, lambda v: int(v) >= 0, 90100),
        'max_locks_per_transaction': (64, lambda v: int(v) >= 32, 90100),
        'track_commit_timestamp': ('off', lambda v: parse_bool(v) is not None, 90500),
        'max_replication_slots': (10, lambda v: int(v) >= 4, 90400),
        'max_worker_processes': (8, lambda v: int(v) >= 2, 90400),
        'wal_log_hints': ('on', lambda _: False, 90400)
    })
```

Is there any reason or intention for this ? 
I want to set "wal_log_hints" to off because I don't want  to produce too much WALs.

CyberDem0n:
It is required for pg_rewind to work.

ckh0618:
thanks for the reply. 

How about "wal_log_hints" parameter to be configurable if a user set "use_pg_rewind" to "off" ?

CyberDem0n:
> How about "wal_log_hints" parameter to be configurable if a user set "use_pg_rewind" to "off" ?

Are you volunteering to implement it?
Please also keep in mind that enabled data_checksums effectively producing the same impact as wal_log_hints...

mbanck:
Right, but I think not having `wal_log_hints` is a valid usecase, if you don't want/need checksums and rewind - it might drastically cut down on the size of WAL.

I do think having checksums is probably better, but others might disagree - and I'm not volunteering to implement it either

PR #3063, opened 2024-05-12
Make wal_log_hints configurable

sundongkim-dev:
resolve https://github.com/zalando/patroni/issues/1942


pol-o:
Could you give me an advice for test codes?
All test cases are assuming that 'wal_log_hints' is 'on'. Pretty hard to naming test cases

patroni/config_generator.py
```diff
@@ -268,7 +268,8 @@ def generate(self) -> None:
         wal_level = 'hot_standby' if self.pg_major < 90600 else 'replica'
         self.config['bootstrap']['dcs']['postgresql']['parameters']['wal_level'] = wal_level
 
-        self.config['bootstrap']['dcs']['postgresql']['use_pg_rewind'] = True
+        self.config['bootstrap']['dcs']['postgresql']['use_pg_rewind'] = True \
+            if self.config['bootstrap']['dcs']['postgresql']['parameters']['wal_log_hints'] == "on" else False
```
CyberDem0n, line 272:
```suggestion
        self.config['bootstrap']['dcs']['postgresql']['use_pg_rewind'] = \
            parse_bool(self.config['bootstrap']['dcs']['postgresql']['parameters']['wal_log_hints']) is True
```
But, TBH there is no need to do that, because we are not changing the default value of `wal_log_hints`.

CyberDem0n (commented):
Tests are making an assumption that wal_log_hints are always enabled because it might be required for pg_rewind to work (if `Data page checksum` are not enabled) and we want to test that pg_rewind works correctly.
The code implementing pg_rewind support can already handle if wal_log_hints is not set, there is no additional testing required: https://github.com/zalando/patroni/blob/ff99d29e6d96923117d684a22f4c4e388c552c36/patroni/postgresql/rewind.py#L41

What is actually required - fixing the documentation and writing in all places where pg_rewind is mentioned, that either the cluster must be initialized with `Data page checksums` (`--data-checksums` option for `initdb`) and/or `wal_log_hints` must be set to `on`, or pg_rewind will not work. Having `wal_log_hints` enabled in addition to `Data page checksums` doesn't create any overhead.

P.S. it is highly recommended to have `Data page checksums` enabled.

patroni/config_generator.py
pol-o: Do you want me to add the following contents as comments?

- "Either the cluster must be initialized with Data page checksums (--data-checksums option for initdb) and/or wal_log_hints must be set to on, or pg_rewind will not work."

I will commit an example for your request below. Check plz.

pol-o:
Do you want me to add the following contents as comments?

- "Either the cluster must be initialized with Data page checksums (--data-checksums option for initdb) and/or wal_log_hints must be set to on, or pg_rewind will not work."

I will commit an example for your request below. Check plz.

patroni/config_generator.py
pol-o: I will fix it 👍

pol-o:
@CyberDem0n 
It seems like I've fixed everything as you mentioned, could you check it for me?

tests/test_config_generator.py
```diff
@@ -56,7 +56,8 @@ def setUp(self):
         dynamic_config['postgresql']['parameters'] = dict(dynamic_config['postgresql']['parameters'])
         del dynamic_config['standby_cluster']
         dynamic_config['postgresql']['parameters']['wal_keep_segments'] = 8
-        dynamic_config['postgresql']['use_pg_rewind'] = True
+        dynamic_config['postgresql']['use_pg_rewind'] = True \
```
CyberDem0n, line 59: Please use the construction similar to https://github.com/zalando/patroni/pull/3063/files#diff-cfee4a0506f1e512ce4c1fb59d7f7bac2f7924ad20e7d6a679044f4c5f4d6c88R271-R272

pol-o:
@CyberDem0n 
Thx for review. I hope these commits full fill the intended purpose :)

pol-o (commented):
Check please :)

patroni/postgresql/config.py
```diff
@@ -325,7 +325,7 @@ class ConfigHandler(object):
         'track_commit_timestamp': ('off', _bool_validator, 90500),
         'max_replication_slots': (10, IntValidator(min=4), 90400),
         'max_worker_processes': (8, IntValidator(min=2), 90400),
-        'wal_log_hints': ('on', _bool_is_true_validator, 90400)
+        'wal_log_hints': ('on', _bool_validator, 90400)
```
hughcapet, line 328: https://github.com/zalando/patroni/blob/ff31f45226f03df70e2f4243e84308b5e7138994/patroni/postgresql/config.py#L1104

we should also remove it from here, as it is now configurable

pol-o: Thx for review :)

pol-o:
@CyberDem0n @hughcapet 
Check again please :D

hughcapet:
👍

CyberDem0n:
:+1:

PR #3069, opened 2024-05-28
Refactor format_dsn() method

CyberDem0n:
so that it doesn't take any decisions about which keywords should appear in the connection string and just uses a provided dict.


hughcapet:
👍

CyberDem0n:
:+1:

Issue #3071, opened 2024-05-31
Unable to connect to external etcdv3 cluster with client TLS authentication (with CommonName in cert)


ILikePhysics:
### What happened?

I am setting up a pg cluster consisting of 1 master and 2 replicas. I am trying to connect patroni to my existing external etcd v3 cluster which has client TLS auth enabled.

Unfortunately, the only I was able to have patroni connected to it was by using username/password config, with use_proxies enabled and client TLS auth turned off in the etcd cluster. I know that if I supply both client TLS certs and username/password in patroni, etcd will prioritize username/password authentication.

The problem is that regardless of what I do, I get the following error that 
`ERROR: Failed to get list of machines from https://etcd-1.xx:2379/v3: <Unknown error: 'CommonName of client sending a request against gateway will be ignored and not used as expected', code: 2>`.

My etcd cluster is not using any gateway or proxy. Am I misunderstanding something about etcd?

### How can we reproduce it (as minimally and precisely as possible)?

N/A

### What did you expect to happen?

Have patroni authenticate using its client TLS certs (using the CommonName as its username for etcd) or at least, be able to authenticate with username/password while still having client TLS auth enabled in etcd (because the docs says its possible).

### Patroni/PostgreSQL/DCS version

-   Patroni version: 3.3.0
-   PostgreSQL version:  16
-   DCS (and its version):  external etcd 3.5 (3 nodes cluster)

### Patroni configuration file

```yaml
etcd3:
  hosts: etcd-1.xx:2379,etcd-2.xx:2379,etcd-3.xx:2379
  cert: /etc/postgresql/certs/tls.crt
  key: /etc/postgresql/certs/tls.key
    #use_proxies: true
  protocol: https
    #username: pg-1-test
    #password: xxxxx
```

### patronictl show-config

```yaml
Error: Can not find suitable configuration of distributed configuration store
Available implementations: consul, etcd, etcd3, kubernetes, raft
```

### Patroni log files

```shell
With the patroni config above for etcd3, I get:

2024-05-31 02:24:34,476 DEBUG: New etcd client created for https://127.0.0.1:4001
2024-05-31 02:24:34,609 ERROR: Failed to get list of machines from https://etcd-3.xx:2379/v3: <Unknown error: 'CommonName of client sending a request against gateway will be ignored and not used as expected', code: 2>
[...]

Issue #2036, opened 2021-08-17
Certificate authentication doesn't work with etcd v3


chobostar:
**Describe the bug**
Patroni 2.1.0 tries to work with etcd using http requests. So if one tries to use TLS with Common Name authentication it will not work.

https://etcd.io/docs/v3.4/dev-guide/api_grpc_gateway/#notes
```
gRPC-gateway does not support authentication using TLS Common Name.
```

Here is discussion: https://github.com/etcd-io/etcd/issues/11970#issuecomment-687875315

**To Reproduce**
1. have a cert with CN:
```
$ openssl x509 -in /etc/ssl/patroni.pem  -text | grep Subject:
        Subject: L = <redacted>, O = <redacted>, OU = <redacted>, CN = patroni-client
```
2. enable client-tls-auth in etcd
3. add user and role `patroni-client` in etcd server, then enable auth
4. try to run patroni with:
```
etcd3:
  protocol: https
  cert: "/etc/ssl/patroni.pem"
  key: "/etc/ssl/patroni.key"
  cacert: "/etc/ssl/ca.pem"  
```

**Expected behavior**
patroni agent authenticated using etcd v3 as `patroni-client` user

**Screenshots**
no screenshots

**Environment**
- Patroni version: 2.1.0
- PostgreSQL version: 12.3
- DCS (and its version): etcd 3.4.16

**Patroni configuration file**
```
---
name: test123-1
namespace: /patroni/
scope: test123
restapi:
  listen: 0.0.0.0:8008
  connect_address: 10.11.12.3:8008
  authentication:
    username: patroni
    password: "123"
bootstrap:
  dcs:
    ttl: 60
    loop_wait: 10
    retry_timeout: 20
    maximum_lag_on_failover: 1048576
    master_start_timeout: 60
    postgresql:
      use_pg_rewind: True
      use_slots: True
      parameters:
        wal_level: logical
        hot_standby: 'on'
        wal_keep_segments: 100
        max_wal_senders: 12
        max_replication_slots: 12
        wal_log_hints: 'on'
        max_connections: '500'
  initdb: ['encoding=UTF8', 'data-checksums', 'auth-local=trust', 'locale=en_US.UTF-8', 'auth-host=md5', 'auth-local=trust']
  pg_hba: ['local all all trust', 'local replication all trust', 'host replication replication 127.0.0.1/32 trust', 'host replication replication 127.0.0.1/32 md5']
postgresql:
  listen: 0.0.0.0:5432
[...]

PR #3073, opened 2024-06-01
Doc improvement: mention use of gRPC gateway with EtcdV3

jostaub:
I recently tried to use TLS common name authentication with patroni, which is currently not possible, as discussed in #3071 and #2036. In my opinion, the documentation should mention the use of the gRPC gateway (proxy) in the configuration section to better help new users (like me) set up patroni and etcdv3.  Currently this is only reflected in the release notes (as far as I can see).

So I suggest adding an additional note to the warning section in the EtcdV3 Configuration (YML and Environment) documentation. However, I am not married to my current proposed wording and placement of this note. Also, I am not sure if my direct mention of TLS common name authentication should be included directly in the Patroni documentation, as this is already discussed in the official etcd documentation.


hughcapet:
👍

CyberDem0n:
:+1:

PR #3074, opened 2024-06-09
drop dependency on ancient mock

a-detiste:
https://github.com/testing-cabal/mock

mock is now part of the Python standard library, available as unittest.mock in Python 3.3 onwards.

### DIFF
--- a/.github/workflows/install_deps.py
+++ b/.github/workflows/install_deps.py
@@ -21,7 +21,7 @@ def install_requirements(what):
         from setup import EXTRAS_REQUIRE, read
     finally:
         sys.path = old_path
-    requirements = ['mock>=2.0.0', 'flake8', 'pytest', 'pytest-cov'] if what == 'all' else ['behave']
+    requirements = ['flake8', 'pytest', 'pytest-cov'] if what == 'all' else ['behave']
     requirements += ['coverage']
     # try to split tests between psycopg2 and psycopg3
     requirements += ['psycopg[binary]'] if sys.version_info >= (3, 8, 0) and\
--- a/requirements.dev.txt
+++ b/requirements.dev.txt
@@ -2,7 +2,6 @@ psycopg2-binary
 behave
 coverage
 flake8>=3.0.0
-mock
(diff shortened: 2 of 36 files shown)


CyberDem0n (changes requested):
Thanks for PR @a-detiste.
I think you didn't try to run tests.
In a few places mock is explicitly imported, because for python3.6 we need to use `mock.call(...)`: https://github.com/zalando/patroni/blob/0a6c09e252ecbee6f5184c1e1e13075527c1b3f0/tests/test_ctl.py#L646-L647

If we just import call() and use it directly, doctests are failing with the exception:
```python
/usr/lib/python3.6/doctest.py:933: in find
    self._find(tests, obj, name, module, source_lines, globs, {})
/usr/lib/python3.6/doctest.py:992: in _find
    if ((inspect.isroutine(inspect.unwrap(val))
/usr/lib/python3.6/inspect.py:516: in unwrap
    raise ValueError('wrapper loop when unwrapping {!r}'.format(f))
E   ValueError: wrapper loop when unwrapping call
```

Instead of using `import mock` we need to do `from unittest import mock`, otherwise tests will fail.
However, there could be place where we import mock, and use it for something different from `mock.call()`.
E.g., `mock.patch()`, and those better fix by doing `import patch from unittest.mock`.

a-detiste:
Hi,

I admit I didn't ran the tests at first attempt.
I'm sending in parralel a huge number of these similar PR and most times Pyflakes will find my mistakes.

>  better fix by doing **import patch from unittest.mock**.

I'm bot sure, different tests have different import sorting style.
A tiny bit confusing, but nothing really bothersome. I tried to keep my diff minimal
and not change the body of the tests.

image

Cheers, tests are running

image

CyberDem0n:
:+1:

hughcapet:
👍

a-detiste:
You might want to run `isort` on the whole codebase.

Le mer. 12 juin 2024 à 10:47, Polina Bungina ***@***.***> a
écrit :

PR #3076, opened 2024-06-11
Compatibility with 17-beta1

CyberDem0n:
- updated list of GUCs
- updated regex for filtering backend processes by name
- `primary_conninfo` will contain `dbname` parameter

The last one is required for synchronizing logical replication slots by slotsync worker and doesn't create problems on older versions.


hughcapet:
👍

CyberDem0n:
:+1:

PR #3082, opened 2024-06-14
Don't socket.getaddrinfo() from config_generator.py

CyberDem0n:
the get_address() function was called when config_generator.py is loaded because it was required to initialize _HOSTNAME and _IP properties of AbstractConfigGenerator and in some cases making unit tests very slow.


hughcapet:
👍

CyberDem0n:
:+1:

PR #3108, opened 2024-07-19
Make replication slots for members behave like permanent slots

CyberDem0n:
That is, create slots on all nodes that could potentially become a new leader and use `pg_replication_slot_advance()` to move them forward.


hughcapet:
https://github.com/patroni/patroni/blob/8cd9690fa47083d4a58eff9139180479dc1dcd23/docs/dynamic_configuration.rst?plain=1#L94-L110

needs an update

patroni/dcs/__init__.py
```diff
@@ -1075,7 +1076,8 @@ def _get_permanent_slots(self, postgresql: 'Postgresql', tags: Tags, role: str)
         return self.__permanent_slots if postgresql.can_advance_slots or role in ('master', 'primary') \
             else self.__permanent_logical_slots
 
-    def _get_members_slots(self, name: str, role: str) -> Dict[str, Dict[str, str]]:
+    def _get_members_slots(self, name: str, role: str, nofailover: bool,
+                           can_advance_slots: int) -> Dict[str, Dict[str, Any]]:
```
hughcapet, line 1080:
```suggestion
                           can_advance_slots: bool) -> Dict[str, Dict[str, Any]]:
```

patroni/dcs/__init__.py
```diff
@@ -1092,9 +1094,9 @@ def _get_members_slots(self, name: str, role: str) -> Dict[str, Dict[str, str]]:
             * Conflicting slot names between members are found
 
         :param name: name of this node.
-        :param role: role of this node, if this is a ``primary`` or ``standby_leader`` return list of members
-                     replicating from this node. If not then return a list of members replicating as cascaded
-                     replicas from this node.
+        :param role: role of this node, ``primary``, ``standby_leader``, or ``replica``.
+        :param nofailover: ``True`` if this node is tagged to not be a failover candidate, ``False`` otherwise.
+        :param can_advance_slots: ``True`` if PostgreSQL major version is v11+, ``False`` otherwise.
```
hughcapet, line 1099: i would better use the same docstring as in `Cluster._merge_permanent_slots()`

```
``True`` if ``pg_replication_slot_advance()`` function is available,
                                  ``False`` otherwise.
 ```

**Comment by hughcapet** [comment]
https://github.com/patroni/patroni/blob/2fb39616aa0cee310570b40fcf54fa3587cac30b/docs/dynamic_configuration.rst?plain=1#L108-L110
->
```
An exception to that rule are permanent physical slots that match the Patroni member names (created and maintained by Patroni).
```


i think it also makes sense to mention automatically created physical permanent slots here (because we used to give recommendations to manually configure member slots for some cases which are now covered):
https://github.com/patroni/patroni/blob/2fb39616aa0cee310570b40fcf54fa3587cac30b/docs/dynamic_configuration.rst?plain=1#L58


and mention here that since version 4 (?), these slots are created automatically, but you still need to define them as permanent if you want to avoid the deletion. 
don't forget to add an upcoming functionality in later prs:)

https://github.com/patroni/patroni/blob/2fb39616aa0cee310570b40fcf54fa3587cac30b/docs/faq.rst?plain=1#L188

**Comment by hughcapet** [comment]
What about the slots for `nostream` members? 🤔

File: patroni/dcs/__init__.py [hunk 3]
```diff
@@ -1703,6 +1744,50 @@ def failsafe(self) -> Optional[Dict[str, str]]:
         """Stored value of :attr:`~AbstractDCS._last_failsafe`."""
         return self._last_failsafe
 
+    def _build_retain_slots(self, cluster: Cluster, slots: Optional[Dict[str, int]]) -> Optional[List[str]]:
+        """Handle retention policy of physical replication slots for cluster members.
+
+        When the member key is missing we want to keep its replication slot for a while, so that WAL sengments
```
hughcapet, line 1750:
```suggestion
        When the member key is missing we want to keep its replication slot for a while, so that WAL segments
```

patroni/dcs/__init__.py
```diff
@@ -1703,6 +1744,50 @@ def failsafe(self) -> Optional[Dict[str, str]]:
...
+
+        This method handles retention policy by keeping the list of such replication slots in memory
+        and removing names when they were observed longer than ``member_slots_ttl`` ago.
+
+        :param cluster: :class:`Cluster` object with information about the current cluster state.
+        :param slots: slot names with LSN values.
+
+        :returns: the list of replication slots to be written to ``/status`` key or ``None``.
+        """
+        timestemp = time.time()
```
hughcapet, line 1762:
```suggestion
        timestamp = time.time()
```
and further

hughcapet, line 1762:
```suggestion
        timestamp = time.time()
```
and further

patroni/global_config.py
```diff
@@ -129,7 +129,7 @@ def is_standby_cluster(self) -> bool:
         return isinstance(config, dict) and\
             bool(config.get('host') or config.get('port') or config.get('restore_command'))
 
-    def get_int(self, name: str, default: int = 0) -> int:
+    def get_int(self, name: str, default: int = 0, base_unit: Optional[str] = None) -> int:
```
hughcapet, line 132: new param in the docstring?

patroni/global_config.py
```diff
@@ -225,5 +225,13 @@ def permanent_slots(self) -> Dict[str, Any]:
                         or self.get('slots')
                         or EMPTY_DICT.copy())
 
+    @property
+    def member_slots_ttl(self) -> int:
+        """Currently configured value of ``member_slots_ttl`` from the global configuration converted to seconds.
+
+        Assume ``1800`` is it is not set or invalid.
```
hughcapet, line 232:
```suggestion
        Assume ``1800`` if it is not set or invalid.
```

patroni/ha.py
```diff
@@ -372,11 +372,9 @@ def update_lock(self, update_status: bool = False) -> bool:
         if update_status:
             try:
                 last_lsn = self._last_wal_lsn = self.state_handler.last_operation()
-                slots = self.cluster.filter_permanent_slots(
-                    self.state_handler,
-                    {**self.state_handler.slots(), slot_name_from_member_name(self.state_handler.name): last_lsn})
+                slots = self.cluster.maybe_filter_permanent_slots(self.state_handler, self.state_handler.slots())
             except Exception:
-                logger.exception('Exception when called state_handler.last_operation()')
+                logger.exception('Exception when called state_handler.slots()')
```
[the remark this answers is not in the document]
CyberDem0n: Yes, last_operation() can error out.  slots() will just already cached results and can't raise an exception.

docs/dynamic_configuration.rst
```diff
@@ -56,6 +56,7 @@ In order to change the dynamic configuration you can use either :ref:`patronictl
    -  **archive\_cleanup\_command**: cleanup command for standby leader
    -  **recovery\_min\_apply\_delay**: how long to wait before actually apply WAL records on a standby leader
 
+-  **member_slots_ttl**: retention time of physical replication slots for replicas whey they are shut down. Default value: `30min`, set it to `0` if you want to keep the old behavior (when the member key expires from DCS slot is immediately removed). The feature works only starting from PostgreSQL 11.
```
barthisrael, line 59:
```suggestion
-  **member_slots_ttl**: retention time of physical replication slots for replicas when they are shut down. Default value: `30min`. Set it to `0` if you want to keep the old behavior (when the member key expires from DCS, the slot is immediately removed). The feature works only starting from PostgreSQL 11.
```

docs/dynamic_configuration.rst
```diff
@@ -92,7 +93,7 @@ Note: **slots** is a hashmap while **ignore_slots** is an array. For example:
             type: physical
           ...
 
-Note: if cluster topology is static (fixed number of nodes that never change their names) you can configure permanent physical replication slots with names corresponding to names of nodes to avoid recycling of WAL files while replica is temporary down:
+Note: When running PostgreSQL v11 or newer Patroni maintains physical replication slots on all nodes that could potentially become a leader, so that replica nodes keep WAL segments reserved if they potentially required by other nodes. In case if the node is absent and its member key in DCS expired, corresponding replication slot is dropped after ``member_slots_ttl`` (default value is `30min`). You can increase or decrease retention based on your needs. Alternatively, if your cluster topology is static (fixed number of nodes that never change their names) you can configure permanent physical replication slots with names corresponding to names of nodes to avoid slots removal and recycling of WAL files while replica is temporary down:
```
barthisrael, line 96:
```suggestion
Note: When running PostgreSQL v11 or newer Patroni maintains physical replication slots on all nodes that could potentially become a leader, so that replica nodes keep WAL segments reserved if they are potentially required by other nodes. In case the node is absent and its member key in DCS gets expired, the corresponding replication slot is dropped after ``member_slots_ttl`` (default value is `30min`). You can increase or decrease retention based on your needs. Alternatively, if your cluster topology is static (fixed number of nodes that never change their names) you can configure permanent physical replication slots with names corresponding to the names of the nodes to avoid slots removal and recycling of WAL files while replica is temporarily down:
```

docs/faq.rst
```diff
@@ -181,7 +181,10 @@ What is the difference between ``etcd`` and ``etcd3`` in Patroni configuration?
     * API version 2 will be completely removed on Etcd v3.6.
 
 I have ``use_slots`` enabled in my Patroni configuration, but when a cluster member goes offline for some time, the replication slot used by that member is dropped on the upstream node. What can I do to avoid that issue?
-    You can configure a permanent physical replication slot for the members.
+    There are two options:
+
+    1. You can tune ``member_slots_ttl`` (default value ``30min``, available since Patroni ``4.0.0`` and PostgreSQL 11 onwards) and replication slots for absent members will not be removed when member downtime is shorter than the configured threshold.
```
barthisrael, line 186:
```suggestion
    1. You can tune ``member_slots_ttl`` (default value ``30min``, available since Patroni ``4.0.0`` and PostgreSQL 11 onwards) and replication slots for absent members will not be removed when the members downtime is shorter than the configured threshold.
```

docs/faq.rst
```diff
@@ -181,7 +181,10 @@ What is the difference between ``etcd`` and ``etcd3`` in Patroni configuration?
...
+    2. You can configure a permanent physical replication slot for the members.
```
barthisrael, line 187:
```suggestion
    2. You can configure permanent physical replication slots for the members.
```

patroni/global_config.py
```diff
@@ -135,16 +135,17 @@ def is_standby_cluster(self) -> bool:
         return isinstance(config, dict) and\
             bool(config.get('host') or config.get('port') or config.get('restore_command'))
 
-    def get_int(self, name: str, default: int = 0) -> int:
+    def get_int(self, name: str, default: int = 0, base_unit: Optional[str] = None) -> int:
         """Gets current value of *name* from the global configuration and try to return it as :class:`int`.
 
         :param name: name of the parameter.
         :param default: default value if *name* is not in the configuration or invalid.
+        :param base_unit: an optional base unit to convert *value* to. Not used if *value* does not contain a unit.
```
barthisrael, line 143:
```suggestion
        :param base_unit: an optional base unit to convert value of *name* parameter to. Not used if the value does not contain a unit.
```

barthisrael (commented):
I still need to continue the review later, but I'm posting some suggestions that I had so far.

patroni/dcs/__init__.py
```diff
@@ -1729,6 +1770,50 @@ def failsafe(self) -> Optional[Dict[str, str]]:
...
+                members.add(slot_name_from_member_name(self._name))
+
+            permanent_slots = cluster.permanent_physical_slots
+            # we want to have in ``retain_slots`` only non-permanent member slots
+            self._last_retain_slots.update({name: timestamp for name in slots
+                                            if name in members and name not in permanent_slots})
+        # retention
+        for name, value in list(self._last_retain_slots.items()):
+            if value + global_config.member_slots_ttl <= timestamp:
+                del self._last_retain_slots[name]
```
barthisrael, line 1813: I think this deserves a log message so we have a clear indication of what is causing the slot to get dropped.

barthisrael (commented):
Most of the PR looks good.

I ran some tests and it seems to behave as expected.

I posted some questions and suggestions along the way. Maybe there is room for adding some more information to comments / docstrings.

patroni/dcs/__init__.py
```diff
@@ -1131,50 +1146,67 @@ def _get_members_slots(self, name: str, role: str) -> Dict[str, Dict[str, str]]:
         # also exlude members with disabled WAL streaming
         members = filter(lambda m: m.name != name and not m.nostream, self.members)
 
-        if role in ('master', 'primary', 'standby_leader'):
+        if can_advance_slots and global_config.member_slots_ttl > 0:
+            # if the node does only cascading and can't become the leader, we
+            # want only to have slots for members that could connect to it.
+            members = [m for m in members if not nofailover or m.replicatefrom == name]
+        elif role in ('master', 'primary', 'standby_leader'):  # PostgreSQL is older than 11
+            # on the leader want to have slots only for the nodes that are supposed to be replicating from it.
```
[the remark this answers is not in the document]
CyberDem0n: > I was wondering about that because the primary can also fall into the current first if (about slots and ttl), can't it?

This is exactly the reason.
Normally (old behavior), primary should have only replication slots for nodes that are streaming from it (exclude nodes with `tags.replicatefrom: othernode` if the `othernode` is healthy):
```
primary<-othernode<-node2
```

But, with the new feature it needs to be different:
```
primary: slots(othernode, node2)
othernode: slots(primary, node2)
node2: slots(primary, othernode)
```

That is, `primary` will have active slot for `othernode` and inactive slot for `node2`, which needs periodically to move forward using `pg_replication_slot_advance()`.
The same story for remaining nodes.

Nodes with `tags.nofailover` that are doing cascading are different. They need to have slots only for nodes that are supposed to replicate from them (`if not nofailover or m.replicatefrom == name`).

patroni/dcs/__init__.py
```diff
@@ -1131,50 +1146,67 @@ def _get_members_slots(self, name: str, role: str) -> Dict[str, Dict[str, str]]:
...
 
-        slots = {slot_name_from_member_name(m.name): {'type': 'physical'} for m in members}
-        if len(slots) < len(members):
+        slots: Dict[str, int] = self.slots
+        ret: Dict[str, Dict[str, Any]] = {}
+        for member in members:
+            slot_name = slot_name_from_member_name(member.name)
+            lsn = slots.get(slot_name, 0)
+            if member.replicatefrom:
+                lsn = max(member.lsn or 0, lsn)
```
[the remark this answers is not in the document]
CyberDem0n: It is again about cascading setups:
```
primary<-node1<-node2(replicatefrom: node1)
```

`/status: {"slots":{...}}` in DCS is updated only by the primary.
But, replication slot for `node2` on the primary is not active, therefore as a workaround we want to take `received/replayed` LSN reported by the `node2`. `max` is used just in case, for example if there is no feedback from the `node2` because it is restarting.
Later, `primary` will advance LSN of the `node2` slot and update "slots" value in DCS.

patroni/dcs/__init__.py
```diff
@@ -1131,50 +1146,67 @@ def _get_members_slots(self, name: str, role: str) -> Dict[str, Dict[str, str]]:
...
+            ret[slot_name] = {'type': 'physical', 'lsn': lsn}
+        ret.update({slot: {'type': 'physical'} for slot in self.status.retain_slots
+                    if slot not in ret and slot != name})
```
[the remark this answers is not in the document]
CyberDem0n: It is a way of keeping replication sots when /member/$name key is temporary absent.

patroni/ha.py
```diff
@@ -173,7 +173,7 @@ def update_cluster(self, cluster: Cluster) -> Cluster:
         leader = self.leader
         if leader:
             # We rely on the strict order of fields in the namedtuple
-            status = Status(cluster.status.last_lsn, leader.member.data['slots'])
+            status = Status(cluster.status[0], leader.member.data['slots'], *cluster.status[2:])
```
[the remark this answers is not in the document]
CyberDem0n: If `Status` namedtuple will be extended by adding new fields, this line will continue to work.

patroni/dcs/__init__.py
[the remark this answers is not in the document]
barthisrael: Got it, thanks for clarifying!

patroni/dcs/__init__.py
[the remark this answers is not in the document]
barthisrael: Got it, thanks!

patroni/ha.py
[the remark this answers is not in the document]
barthisrael: Makes sense!

I thought of that possibility after closing the computer yesterday 😄 but thanks for confirming!

patroni/dcs/__init__.py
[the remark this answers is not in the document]
barthisrael:
```suggestion
            if member.replicatefrom:
                # `/status` key is maintained by the leader, but `member` may be connected to some other node. In that
                # case, the slot in the leader is inactive and doesn't advance, so we use the LSN reported by the member
                # `max` is only a fallback so we take the LSN from the slot when there is no feedback from the member
                lsn = max(member.lsn or 0, lsn)
```

I wonder if this would be worth it?

barthisrael (approved):
I put a final suggestion in regards to a code comment, if you'd like to include it.

I'm putting my approval as the PR looks fine.

hughcapet (dismissed):

hughcapet (approved):

barthisrael (approved):
Thanks, looks good!

PR #3113, opened 2024-07-29
Ignore restapi.allowlist_include_members for POST /failsafe

### LINKED ISSUE #3096: Failsafe mode when master doesn't have access to DCS
### What happened?

_I am very sorry to create a bug issue but i just don't have access to slack in my country and even vpns seem to not let me in. And i am in a need of an answer. I cannot provide logs as i didn't use this patroni but I examined docs and code and this came to mind_

**Decsription:**
When failsafe_mode is enabled and master fails to update leader, it sends requests to all other nodes and they must accept it in in order for master to continue mastering. Every request is checked to come from an allowed url. In case when
- allowlist parameter is empty
- allowlist_include_members is True

incoming url must be checked to be in a cluster's members, which is maintained in DCS.

**The issue itself:**
If master lost access to DCS in cannot update its' members key in DCS. Therefore the key expires after ttl time passes, request from master is rejected as it comes from a not allowed url, then master demotes.

I know i can fill allowlist parameter with all cluster members and it will solve the problem, but i think it is not intended to do so. Are there any other ways to avoid demotion in such case?

1/ enable failsafe_mode
2/ block all master connections with DCS but remain all master-replica connections
3/ wait ttl seconds

master continue to being master as it can connect with replicas via REST

-   Patroni version: 
-   PostgreSQL version: 
-   DCS (and its version): 

```yaml
None
```

### patronictl show-config

```yaml
None
```

### Patroni log files

```shell
None
```

### PostgreSQL log files

```shell
None
```

- [X] Yes

Again very sorry for creating a bug, i cant seem to find how to remove the label

CyberDem0n:
If only the leader can't access DCS its member key will expire and `POST /failsafe` requests might be rejected because of that.

Close #3096

### DIFF
--- a/patroni/api.py
+++ b/patroni/api.py
@@ -43,7 +43,7 @@ def check_access(*args: Any, **kwargs: Any) -> Callable[..., Any]:
     .. note::
         The actual logic to check access is implemented through :func:`RestApiServer.check_access`.
 
-        Optionally it is possible to skip source ip check by specifying `allowlist_check_members`=False``.
+        Optionally it is possible to skip source ip check by specifying ``allowlist_check_members=False``.
 
     :returns: a decorator that executes *func* only if :func:`RestApiServer.check_access` returns ``True``.
 
--- a/patroni/api.py
+++ b/patroni/api.py
@@ -37,13 +37,13 @@ from .utils import deep_compare, enable_keepalive, parse_bool, patch_config, Ret
 logger = logging.getLogger(__name__)
 
 
-def check_access(func: Callable[..., None]) -> Callable[..., None]:
+def check_access(*args: Any, **kwargs: Any) -> Callable[..., Any]:
     """Check the source ip, authorization header, or client certificates.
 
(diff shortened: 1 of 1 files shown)


patroni/api.py
```diff
@@ -37,13 +37,13 @@
...
 
-def check_access(func: Callable[..., None]) -> Callable[..., None]:
+def check_access(*args: Any, **kwargs: Any) -> Callable[..., Any]:
     """Check the source ip, authorization header, or client certificates.
 
     .. note::
         The actual logic to check access is implemented through :func:`RestApiServer.check_access`.
 
-    :param func: function to be decorated.
+        Optionally it is possible to skip source ip check by specifying `allowlist_check_members`=False``.
```
hughcapet, line 46:
```suggestion
        Optionally it is possible to skip source ip check by specifying ``allowlist_check_members``=``False``.
```

hughcapet (approved):

PR #3126, opened 2024-08-14
Fix request URL in failsafe handling log

waynerv:
When addressing a DCS error, the URLs in the log do not reflect the actual ones.

```
2024-08-13 16:16:03,091 ERROR: Error communicating with DCS
2024-08-13 16:16:03,163 INFO: Got response from postgres-15add5e6-0-0 https://245.0.0.119:8009/patroni: Accepted
2024-08-13 16:16:03,163 INFO: Got response from postgres-15add5e6-1-0 https://245.0.1.149:8009/patroni: Accepted
2024-08-13 16:16:03,165 INFO: continue to run as a leader because failsafe mode is enabled and all members are accessible
```

We called `/failsafe`, not `/patroni`.

### DIFF
--- a/patroni/request.py
+++ b/patroni/request.py
@@ -2,7 +2,6 @@
 import json
 
 from typing import Any, Dict, Optional, Union
-from urllib.parse import urlparse, urlunparse
 
 import urllib3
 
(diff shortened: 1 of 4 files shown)


patroni/request.py
```diff
@@ -136,6 +136,21 @@ def reload_config(self, config: Union[Config, Dict[str, Any]]) -> None:
         cacert = self._get_ctl_value(config, 'cacert') or self._get_restapi_value(config, 'cafile')
         self._apply_pool_param('ca_certs', cacert)
 
+    @staticmethod
+    def get_restapi_url(base: Optional[str], endpoint: Optional[str]) -> str:
```
CyberDem0n, line 140: It would be more logical to make it a method of the Member class.
For example, `get_endpoint_url(self, endpoint: Optional[str] = None) -> str`

CyberDem0n (changes requested):
With failing pyright we can't accept it.

CyberDem0n (approved):

PR #3129, opened 2024-08-15
Remove patronictl failover --leader option

hughcapet:
Option has been deprecated and should be removed in the new major release

### DIFF
--- a/patroni/ctl.py
+++ b/patroni/ctl.py
@@ -1193,9 +1193,8 @@ def reinit(cluster_name: str, group: Optional[int], member_names: List[str], for
                 wait_on_members.remove(member)
 
 
-def _do_failover_or_switchover(action: str, cluster_name: str, group: Optional[int],
-                               switchover_leader: Optional[str], candidate: Optional[str],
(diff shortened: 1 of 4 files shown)


patroni/ctl.py
```diff
@@ -1193,9 +1193,8 @@ def reinit(cluster_name: str, group: Optional[int], member_names: List[str], for
                 wait_on_members.remove(member)
 
 
-def _do_failover_or_switchover(action: str, cluster_name: str, group: Optional[int],
-                               switchover_leader: Optional[str], candidate: Optional[str],
-                               force: bool, scheduled: Optional[str] = None) -> None:
+def _do_failover_or_switchover(action: str, cluster_name: str, group: Optional[int], candidate: Optional[str],
+                               force: bool, leader: Optional[str] = None, scheduled: Optional[str] = None) -> None:
```
CyberDem0n, line 1197: Maybe we shouldn't rename switchover_leader -> leader?
After all, it is used exclusively for a switchover.

hughcapet: well, scheduled is also only for switchover

CyberDem0n (approved):

PR #3130, opened 2024-08-19
Remove postgres --describe-config validation

hughcapet:
Due to `postgres --describe-config` not showing GUCs defined as `GUC_NO_SHOW_ALL | GUC_NOT_IN_SAMPLE | GUC_DISALLOW_IN_FILE`, Patroni was always ignoring some GUCs that a user might want to have configured with non-default values.

Also defined minor versions for availability bounds for some back-patched GUCs

### DIFF
--- a/patroni/postgresql/config.py
+++ b/patroni/postgresql/config.py
@@ -16,8 +16,9 @@ from ..collections import CaseInsensitiveDict, CaseInsensitiveSet, EMPTY_DICT
 from ..dcs import Leader, Member, RemoteMember, slot_name_from_member_name
 from ..exceptions import PatroniFatalException, PostgresConnectionException
 from ..file_perm import pg_perm
-from ..utils import compare_values, is_subpath, maybe_convert_from_base_unit, \
-    parse_bool, parse_int, split_host_port, uri, validate_directory
+from ..postgresql.misc import postgres_version_to_int
+from ..utils import compare_values, get_postgres_version, is_subpath, \
+    maybe_convert_from_base_unit, parse_bool, parse_int, split_host_port, uri, validate_directory
 from ..validator import EnumValidator, IntValidator
 from .validator import recovery_parameters, transform_postgresql_parameter_value, transform_recovery_parameter_value
 
@@ -404,6 +405,15 @@ class ConfigHandler(object):
         return self._config_dir
 
     @property
+    def full_pg_version(self) -> int:
+        if self._postgresql.state == 'running':
+            try:
+                return self._postgresql.server_version
+            except AttributeError:
+                pass
+        return postgres_version_to_int(get_postgres_version(bin_name=self._postgresql.pgcommand('postgres')))
+
+    @property
     def _configuration_to_save(self) -> List[str]:
         configuration = [os.path.basename(self._postgresql_conf)]
         if 'custom_conf' not in self._config:
@@ -487,7 +497,7 @@ class ConfigHandler(object):
             include = self._config.get('custom_conf') or self._postgresql_base_conf_name
             f.writeline("include '{0}'\n".format(ConfigWriter.escape(include)))
             for name, value in sorted((configuration).items()):
-                value = transform_postgresql_parameter_value(self._postgresql.major_version, name, value)
+                value = transform_postgresql_parameter_value(self.full_pg_version, name, value)
                 if value is not None and\
                         (name != 'hba_file' or not self._postgresql.bootstrap.running_custom_bootstrap):
                     f.write_param(name, value)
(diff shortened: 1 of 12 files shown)


barthisrael (commented):
At a first glance the changes look good.

_I didn't check into details about the back-patched GUCs._

patroni/postgresql/config.py
```diff
@@ -487,8 +497,7 @@ def write_postgresql_conf(self, configuration: Optional[CaseInsensitiveDict] = N
             include = self._config.get('custom_conf') or self._postgresql_base_conf_name
             f.writeline("include '{0}'\n".format(ConfigWriter.escape(include)))
             for name, value in sorted((configuration).items()):
-                value = transform_postgresql_parameter_value(self._postgresql.major_version, name, value,
-                                                             self._postgresql.available_gucs)
+                value = transform_postgresql_parameter_value(self.full_pg_version, name, value)
```
CyberDem0n, line 500: Lets introduce a new variable outside the loop to avoid calling `postgres --version` for every parameter in the worst case.

patroni/postgresql/config.py
```diff
@@ -403,6 +404,15 @@ def check_directories(self) -> None:
...
         return self._config_dir
 
+    @property
+    def full_pg_version(self) -> int:
+        if self._postgresql.state == 'running':
+            try:
+                return self._postgresql.server_version
+            except AttributeError:
+                pass
+        return postgres_version_to_int(get_postgres_version(bin_name=self._postgresql.pgcommand('postgres')))
```
CyberDem0n, line 414: There is a race condition possible, when major version in PG_VERSION doesn't match the binary.
In this case we better take `self._postgresql.major_version`.

patroni/utils.py
```diff
@@ -1211,3 +1211,36 @@ def get_major_version(bin_dir: Optional[str] = None, bin_name: str = 'postgres')
     if TYPE_CHECKING:  # pragma: no cover
         assert version is not None
     return '.'.join([version.group(1), version.group(3)]) if int(version.group(1)) < 10 else version.group(1)
+
+
+def get_postgres_version(bin_dir: Optional[str] = None, bin_name: str = 'postgres') -> str:
+    """Get full PostgreSQL version.
+
+    Like func:`get_major_version` but extended with minor version.
```
CyberDem0n, line 1219: IMO `get_major_version()` should simply call this function and remove last `r'\.\d+'` from the returned value.

patroni/utils.py
```diff
@@ -1211,3 +1211,36 @@ def get_major_version(bin_dir: Optional[str] = None, bin_name: str = 'postgres')
...
+    else:
+        binary = os.path.join(bin_dir, bin_name)
+    try:
+        version = subprocess.check_output([binary, '--version']).decode()
+    except OSError as e:
+        raise PatroniException(f'Failed to get postgres version: {e}')
+    version = re.match(r'^[^\s]+ [^\s]+ (\d+(\.\d+)*)', version)
+    if TYPE_CHECKING:  # pragma: no cover
+        assert version is not None
+    return version.group(1)
```
CyberDem0n, line 1246: For `17beta3` it will return just `17`, and `postgres_version_to_int` will raise `PostgresException`.

patroni/utils.py
```diff
@@ -1207,7 +1207,35 @@ def get_major_version(bin_dir: Optional[str] = None, bin_name: str = 'postgres')
...
+    :raises:
+        :exc:`~patroni.exceptions.PatroniException`: if the postgres binary call failed due to :exc:`OSError`.
+
+    :Example:
+
+        * Returns `9.6` for PostgreSQL 9.6.24
+        * Returns `15` for PostgreSQL 15.2
+    """
+    full_version = get_postgres_version(bin_dir, bin_name)
+    return re.sub(r'(\.\d+)(?!\.)', '', full_version)
```
CyberDem0n, line 1241: Hmm....
Why not?
```suggestion
    return re.sub(r'\.\d+$', '', full_version)
```

patroni/utils.py
```diff
@@ -1180,24 +1180,24 @@ def unquote(string: str) -> str:
     return ret
 
 
-def get_major_version(bin_dir: Optional[str] = None, bin_name: str = 'postgres') -> str:
-    """Get the major version of PostgreSQL.
+def get_postgres_version(bin_dir: Optional[str] = None, bin_name: str = 'postgres') -> str:
+    """Get full PostgreSQL version.
 
-    It is based on the output of ``postgres --version``.
+    Like func:`get_major_version` but extended with minor version.
```
CyberDem0n, line 1186: I think this line should be swapped with https://github.com/patroni/patroni/pull/3130/files#diff-1b3584607a3bcb7f81de8462e86c83f6ec566b0f0d2b58eb969f1ec4a778e7c0R1224 (and slightly modified)

CyberDem0n (approved):

PR #3142, opened 2024-08-30
Fix bug in member slots retention feature

CyberDem0n:
If `name` contains upper case or special characters the node was creating unused replication slot for itself.

### DIFF
--- a/docs/releases.rst
+++ b/docs/releases.rst
@@ -3,6 +3,18 @@
 Release notes
 =============
 
+Version 4.0.1
+-------------
+
+Released 2024-08-30
+
+**Bugfix**
+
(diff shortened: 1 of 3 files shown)


docs/releases.rst
```diff
@@ -3,6 +3,18 @@
...
+Version 4.0.1
+-------------
+
+Released 2024-08-30
+
+**Bugfix**
+
+- Patroni was creating unnecessary replication slots for itself (Alexander Kukushkin)
+
+  It was happening if ``name`` contains upper case or sepcial characters.
```
hughcapet, line 15:
```suggestion
  It was happening if ``name`` contains upper case or special characters.
```

hughcapet (approved):

PR #3149, opened 2024-09-06
Fix unhandled DCSError cause patroni exit during startup phase

waynerv:
DCSError in startup phase will cause patroni exit immediately because `Patroni.ensure_unique_name()` didn't catch exception raised by `Patroni.dcs.get_cluster()`

### DIFF
--- a/patroni/__main__.py
+++ b/patroni/__main__.py
@@ -18,6 +18,7 @@ from patroni.tags import Tags
 
 if TYPE_CHECKING:  # pragma: no cover
     from .config import Config
+    from .dcs import Cluster
 
 logger = logging.getLogger(__name__)
 
@@ -63,10 +64,11 @@ class Patroni(AbstractPatroniDaemon, Tags):
         self.dcs = get_dcs(self.config)
         self.request = PatroniRequest(self.config, True)
 
-        self.ensure_unique_name()
+        cluster = self.ensure_dcs_access()
+        self.ensure_unique_name(cluster)
 
         self.watchdog = Watchdog(self.config)
-        self.load_dynamic_configuration()
+        self.apply_dynamic_configuration(cluster)
 
         self.postgresql = Postgresql(self.config['postgresql'], self.dcs.mpp)
         self.api = RestApiServer(self, self.config['restapi'])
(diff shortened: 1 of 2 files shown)


patroni/__main__.py
```diff
@@ -88,28 +89,21 @@ def load_dynamic_configuration(self) -> None:
         .. note::
             This method is called only once, at the time when Patroni is started.
         """
```
CyberDem0n, line 91: Docstring doesn't correspond to what this method is doing and must be updated

patroni/__main__.py
```diff
@@ -88,28 +89,21 @@ def load_dynamic_configuration(self) -> None:
...
-                        self.watchdog.reload_config(self.config)
-                elif not self.config.dynamic_configuration and 'bootstrap' in self.config:
-                    if self.config.set_dynamic_configuration(self.config['bootstrap']['dcs']):
-                        self.dcs.reload_config(self.config)
-                        self.watchdog.reload_config(self.config)
-                break
-            except DCSError:
-                logger.warning('Can not get cluster from dcs')
-                time.sleep(5)
+        cluster = self.dcs.cluster
```
CyberDem0n, line 92: It is better explicitly pass `cluster` object as an argument instead of using cached value, because the last one may expire.

CyberDem0n: Also, probably this method should be renamed to `apply_dynamic_configuration`.

patroni/__main__.py
```diff
@@ -88,28 +89,21 @@ def load_dynamic_configuration(self) -> None:
...
+            if self.config.set_dynamic_configuration(self.config['bootstrap']['dcs']):
+                self.dcs.reload_config(self.config)
+                self.watchdog.reload_config(self.config)
 
     def ensure_unique_name(self) -> None:
         """A helper method to prevent splitbrain from operator naming error."""
         from patroni.dcs import Member
 
-        cluster = self.dcs.get_cluster()
+        cluster = self.dcs.cluster
```
CyberDem0n, line 106: The same, please don't use cached value but pass the result of get_cluster() as an argument.

patroni/dcs/__init__.py
```diff
@@ -1698,6 +1698,16 @@ def reset_cluster(self) -> None:
             self._cluster = None
             self._cluster_valid_till = 0
 
+    def ensure_cluster_access(self, sleep_time: int = 5) -> Optional[Cluster]:
```
CyberDem0n, line 1701:
```suggestion
    def ensure_dcs_access(self, sleep_time: int = 5) -> Cluster:
```

patroni/dcs/__init__.py
```diff
@@ -1698,6 +1698,16 @@ def reset_cluster(self) -> None:
...
+        """Continuously attempt to retrieve cluster from DCS, retrying on :exc:`DCSError` with delay."""
```
CyberDem0n, line 1702: The docstring must include method arguments and returned value.
Please take other methods as an example.

patroni/dcs/__init__.py
CyberDem0n: And why this method is in AbstractDCS and not in Patroni class?

CyberDem0n (changes requested):

patroni/dcs/__init__.py
waynerv: I originally thought that all relevant access methods should be provided by the DCS class, but it seems more appropriate for the Patroni class which is the only user to provide it.

hughcapet (dismissed):

patroni/__main__.py
```diff
@@ -14,6 +14,7 @@
 
 from patroni import MIN_PSYCOPG2, MIN_PSYCOPG3, parse_version
 from patroni.daemon import abstract_main, AbstractPatroniDaemon, get_base_arg_parser
+from patroni.dcs import Cluster
```
CyberDem0n, line 17: Please move it under `if TYPE_CHECKING:`:
```python
    from .config import Config
    from .dcs import Cluster
```

After that `Cluster` type should be referenced as string literal: `'Cluster'`.

CyberDem0n (approved):

Issue #3089, opened 2024-06-18
unreasonable ttl will cause all DCS connection raise [Errno 22] Invalid argument


waynerv:
### What happened?

can not establish new connection to DCS:
```
2024-06-18 11:35:58,695 ERROR: Request to server https://246.96.0.1:443 failed: MaxRetryError("HTTPSConnectionPool(host='246.96.0.1', port=443): Max retries exceeded with url: /api/v1/namespaces/qfusion-admin/pods?labelSelector=vendor%3Dqfusion%2Cqfusion-pgha-scope%3Dpostgres-2744242e (Caused by NewConnectionError('<urllib3.connection.HTTPSConnection object at 0x7f7d6564f4e0>: Failed to establish a new connection: [Errno 22] Invalid argument',))",)
```

patroni will use ((ttl - (loop_wait + retry_timeout)) / 3) to set TCP_KEEPINTVL socket option, and max allowed value  in Linux is 32767, invalid value will cause a OSError  [Errno 22] Invalid argument.

### How can we reproduce it (as minimally and precisely as possible)?

set ttl to 99999

### What did you expect to happen?

Setting ttl to an unreasonable value is the user's fault, but since it's difficult to link the ttl of the leader lock with the TCP_KEEPINTVL parameter of the underlying socket in the operating system, rejecting invalid parameters or including helpful hints in logs would be very helpful.

### Patroni/PostgreSQL/DCS version

-   Patroni version: 3.1.2
-   PostgreSQL version: 14 
-   DCS (and its version): Kubernetes

### Patroni configuration file

```yaml
use default loop_wai, retry_timeout & ttl to bootstrap
```

### patronictl show-config

```yaml
after setting ttl to 99999 patronictl show-config will fail

2024-06-18 11:34:23,226 - WARNING - Retrying (Retry(total=0, connect=None, read=None, redirect=0, status=None)) after connection broken by 'NewConnectionError('<urllib3.connection.HTTPSConnection object at 0x7f08b51694e0>: Failed to establish a new connection: [Errno 22] Invalid argument',)': /api/v1/namespaces/qfusion-admin/pods?labelSelector=vendor%3Dqfusion%2Cqfusion-pgha-scope%3Dpostgres-2744242e
[...]

PR #3158, opened 2024-09-11
Fix keepalive connection out of the range issue (#3089)

hadizamani021:
For issue #3089, I researched the possible ranges of keepalive options in Windows, Linux, and macOS. In this PR, I limit the value and check if the value for keepalive options exceeds the allowed range. If an exception is detected, we log a warning and change the value to the maximum possible value.

### DIFF
--- a/patroni/utils.py
+++ b/patroni/utils.py
@@ -1083,7 +1083,6 @@ def apply_keepalive_limit(option: str, value: int) -> int:
         'linux': {'TCP_USER_TIMEOUT': 2147483647, 'TCP_KEEPIDLE': 32767, 'TCP_KEEPINTVL': 32767, 'TCP_KEEPCNT': 127},
         'darwin': {'TCP_KEEPIDLE': 4294967, 'TCP_KEEPINTVL': 4294967, 'TCP_KEEPCNT': 2147483647},
     }
-    max_possible_value = None
     platform = 'linux' if sys.platform.startswith('linux') else sys.platform
     max_possible_value = max_of_options.get(platform, {}).get(option)
     if max_possible_value is not None and value > max_possible_value:


patroni/utils.py
```diff
@@ -1101,24 +1157,43 @@ def keepalive_socket_options(timeout: int, idle: int, cnt: int = 3) -> Iterator[
...
         yield (socket.IPPROTO_TCP, TCP_KEEPINTVL, intvl)
     TCP_KEEPCNT = getattr(socket, 'TCP_KEEPCNT', 0x102 if sys.platform.startswith('darwin') else None)
     if TCP_KEEPCNT is not None:
+        max_possible_value = max_keepalive_option_value("TCP_KEEPCNT")
+        if max_possible_value is not None and cnt > max_possible_value:
+            logger.warning(
+                "TCP_KEEPCNT must not exceed {0}, so it is changed from {1} to {0}."
+                .format(max_possible_value, cnt)
+            )
+            cnt = max_possible_value
```
CyberDem0n, line 1196: It all looks like overenginieering to me.
Why not just have a function `apply_keepalive_limit(option: str, value: int) -> int` which will make sure that the limit is not exceeded, or show the warning and adjust it.

Another question is whether we need to have a warning. IMO it could be a debug.

And the third problem - we prefer to use "%s" placeholders when working with logger instead of str.format or f-string.

hadizamani021: Thank you for your opinions; they are great. I pushed another commit and solved the problems.

patroni/utils.py
```diff
@@ -1064,6 +1064,60 @@ def data_directory_is_empty(data_dir: str) -> bool:
...
+        logger.debug(
+            '%s changed from %d to %d.', option, value, max_possible_value
+        )
+        value = max_possible_value
+
+    if min_possible_value > value:
+        logger.debug(
+            '%s changed from %d to %d.', option, value, min_possible_value
+        )
+        value = min_possible_value
```
CyberDem0n, line 1116: This is impossible situation.

patroni/utils.py
```diff
@@ -1064,6 +1064,60 @@ def data_directory_is_empty(data_dir: str) -> bool:
...
+        if option == "TCP_KEEPCNT":
+            max_possible_value = 2147483647
+
+    if sys.platform.startswith('win32'):
+        if option == "TCP_KEEPIDLE":
+            max_possible_value = 2147483647
+        if option == "TCP_KEEPINTVL":
+            max_possible_value = 2147483647
+        if option == "TCP_KEEPCNT":
+            max_possible_value = 255
```
CyberDem0n, line 1104: That's unknown territory, and has to be verified how it works.
So far only enable_keepalive() was handling the Windows case by using `socket.ioctl`.

CyberDem0n, line 1104: I don't like this spaghetti code, please convert it to a dict:
```
options = {
    'linux': {
        'TCP_USER_TIMEOUT': 2147483647,
        'TCP_KEEPIDLE': 32767,
...
    }
}
```

patroni/utils.py
```diff
@@ -1064,6 +1064,60 @@ def data_directory_is_empty(data_dir: str) -> bool:
...
+
+    if max_possible_value is not None and value > max_possible_value:
+        logger.debug(
+            '%s changed from %d to %d.', option, value, max_possible_value
+        )
```
CyberDem0n, line 1109: It perfectly fits to 120 characters.
```suggestion
        logger.debug('%s changed from %d to %d.', option, value, max_possible_value)
```

patroni/utils.py
hadizamani021: Ok, sure. For now i'll remove this section.

CyberDem0n: I mean, some people for sure running Patroni on Windows in production, and this change might affect them.
Maybe positive, but I don't know, it must be tested.

hadizamani021: I understand. However, I think we can proceed with this PR by temporarily excluding Windows and deferring the tests for now. Testing this scenario might be a bit complex, and there is currently a bug in Linux.

CyberDem0n (changes requested):

patroni/utils.py
```diff
@@ -1064,6 +1064,35 @@ def data_directory_is_empty(data_dir: str) -> bool:
...
+    :param option: The TCP keepalive option name.
+        Possible values are "TCP_USER_TIMEOUT", "TCP_KEEPIDLE", "TCP_KEEPINTVL", and "TCP_KEEPCNT".
+    :param value: The desired value for the keepalive option.
+
+    Returns:
+    :returns: The adjusted value for the keepalive option, ensuring it does not exceed the platform-specific maximum.
+    """
+    max_of_options = {
+        'linux': {'TCP_USER_TIMEOUT': 2147483647, 'TCP_KEEPIDLE': 32767, 'TCP_KEEPINTVL': 32767, 'TCP_KEEPCNT': 127},
+        'darwin': {'TCP_KEEPIDLE': 32767, 'TCP_KEEPINTVL': 32767, 'TCP_KEEPCNT': 127},
```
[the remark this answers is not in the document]
hadizamani021: Yes, sorry. It was my fault. during the previous commit I made the mistake. I correct them.

patroni/utils.py
```diff
@@ -1064,6 +1064,34 @@ def data_directory_is_empty(data_dir: str) -> bool:
...
+    max_of_options = {
+        'linux': {'TCP_USER_TIMEOUT': 2147483647, 'TCP_KEEPIDLE': 32767, 'TCP_KEEPINTVL': 32767, 'TCP_KEEPCNT': 127},
+        'darwin': {'TCP_KEEPIDLE': 4294967, 'TCP_KEEPINTVL': 4294967, 'TCP_KEEPCNT': 2147483647},
+    }
+    max_possible_value = None
+    platform = 'linux' if sys.platform.startswith('linux') else sys.platform
+    max_possible_value = max_of_options.get(platform, {}).get(option)
+    if max_possible_value is not None and value > max_possible_value:
+        logger.debug('%s changed from %d to %d.', option, value, max_possible_value)
+        value = max_possible_value
```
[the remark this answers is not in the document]
hadizamani021: Ok sure,  I added the test. If it has any problem please tell me to solve that.

tests/test_utils.py
```diff
@@ -1,9 +1,10 @@
+import sys
 import unittest
 
 from unittest.mock import Mock, patch
 
 from patroni.exceptions import PatroniException
-from patroni.utils import enable_keepalive, get_major_version, get_postgres_version, \
+from patroni.utils import apply_keepalive_limit, enable_keepalive, get_major_version, get_postgres_version, \
     polling_loop, Retry, RetryFailedError, unquote, validate_directory
```
hughcapet, line 8:
```suggestion
from patroni.utils import apply_keepalive_limit, enable_keepalive, get_major_version, \
    get_postgres_version, polling_loop, Retry, RetryFailedError, unquote, validate_directory
```

hadizamani021: Thank. I added it.

CyberDem0n (approved):

hughcapet (approved):

PR #3169, opened 2024-09-16
Don't retains member slots on nodes with nofailover tag

CyberDem0n:
Followup on #3142


hughcapet:
can we also adjust this docstring here?

https://github.com/patroni/patroni/blob/b31198bb8ff9dba7a7798d4c7378a27141fc06f7/patroni/dcs/__init__.py#L1121-L1124

hughcapet (approved):

PR #3173, opened 2024-09-26
Add support of sslnegotiation client-side connection option

CyberDem0n:
It is available in PostgreSQL 17

Besides that update docs and include PG17 to supported versions.


hughcapet:
shouldn't we also add it here?
https://github.com/patroni/patroni/blob/6b685036d01490c4a994c3616d916bfa7b0ef00d/patroni/config_generator.py#L31-L43

hughcapet (approved):

hughcapet (approved):

PR #3182, opened 2024-10-16
Pyright 1.1.385

CyberDem0n:
Declaring variables with `Union` and using `isinstance()` hack doesn't work anymore. Therefore the code is updated to use `Any` for variable and `cast` function after firguring out the correct type in order to avoid getting errors about `Unknown` types.


hughcapet:
tests failing due to https://github.com/psycopg/psycopg2/issues/1737

hughcapet (approved):

Issue #3189, opened 2024-10-19
Regression tests fail with python-consul 1.5.1


mbanck:
### What happened?

Debian unstable has upgraded python-consul from 0.7.1 to 1.5.1 recently, and since then the regression tests fail:
(see https://ci.debian.net/packages/p/patroni/testing/amd64/53163502/#S24):
```
2702s running test
2702s ============================= test session starts ==============================
2702s platform linux -- Python 3.12.6, pytest-8.3.3, pluggy-1.5.0 -- /usr/bin/python3
2702s cachedir: .pytest_cache
2702s rootdir: /tmp/autopkgtest-lxc.ji2ma5c6/downtmp/build.sB4/src
2702s plugins: cov-5.0.0, typeguard-4.3.0
2710s collecting ... collected 623 items / 2 errors
2710s 
2710s ==================================== ERRORS ====================================
2710s ____________________ ERROR collecting tests/test_consul.py _____________________
2710s tests/test_consul.py:88: in <module>
2710s     @patch.object(consul.Consul.KV, 'get', kv_get)
2710s E   AttributeError: type object 'Consul' has no attribute 'KV'
2710s ____________________ ERROR collecting tests/test_consul.py _____________________
2710s tests/test_consul.py:88: in <module>
2710s     @patch.object(consul.Consul.KV, 'get', kv_get)
2710s E   AttributeError: type object 'Consul' has no attribute 'KV'
```

### How can we reproduce it (as minimally and precisely as possible)?

Use python-consul 1.5.1 for the regression tests

### What did you expect to happen?

Regression tests pass

### Patroni/PostgreSQL/DCS version

-   Patroni version: n/a (at least 3.3.2)
-   PostgreSQL version:  n/a
-   DCS (and its version): consul, python3-consul 1.5.1

### Patroni configuration file

```yaml
n/a
```

### patronictl show-config

```yaml
n/a
```

### Patroni log files

```shell
n/a
```

### PostgreSQL log files

```shell
n/a
```

### Have you tried to use GitHub issue search?

- [X] Yes

PR #3191, opened 2024-10-21
Switch to py-consul

CyberDem0n:
python-consul is unmaintained for a long time and py-consul is an official replacement.
However, we still keep backward compatibility with python-consul.

Close: #3189

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -91,7 +91,7 @@ where dependencies can be either empty, or consist of one or more of the followi
 etcd or etcd3
     `python-etcd` module in order to use Etcd as DCS
 consul
-    `python-consul` module in order to use Consul as DCS
+    `py-consul` module in order to use Consul as DCS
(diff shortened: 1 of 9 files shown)


mbanck-ntap:
Ah ok, Debian (more or less) silently switched from python-consul to py-consul, without changing the source package name. I am currently away from my Debian notebook, but will pull/test this PR ASAP.

hughcapet:
should we also change it here?

https://github.com/patroni/patroni/blob/421d68ad4626a756a1ab334b7c0cc4f1f2d73ff4/README.rst?plain=1#L94

https://github.com/patroni/patroni/blob/421d68ad4626a756a1ab334b7c0cc4f1f2d73ff4/docs/installation.rst?plain=1#L52

hughcapet (approved):

PR #3233, opened 2024-12-15
Compatibility with python-json-logger>=3.1

CyberDem0n:
After refactoring the old API is still working, but producing warnings and pyright also fails.

Besides that improve coverage of watchdog/base.py and ctl.py

### DIFF
--- a/tests/test_ctl.py
+++ b/tests/test_ctl.py
@@ -404,6 +404,7 @@ class TestCtl(unittest.TestCase):
         mock_post.return_value.status = 200
         result = self.runner.invoke(ctl, ctl_args, input='y')
         assert result.exit_code == 0
+        assert 'might be different from the ones' in result.output
 
         # get restart with the non-200 return code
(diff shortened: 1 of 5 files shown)


tests/test_ctl.py
```diff
@@ -399,6 +399,12 @@ def test_restart_reinit(self, mock_post):
         result = self.runner.invoke(ctl, ctl_args, input='y')
         assert result.exit_code == 0
 
+        ctl_args = ['restart', 'alpha', '--pg-version', '99.0', '--pending', '--scheduled', '2300-10-01T14:30']
+        # normal restart, the schedule is actually parsed, but not validated in patronictl
+        mock_post.return_value.status = 200
+        result = self.runner.invoke(ctl, ctl_args, input='y')
+        assert result.exit_code == 0
```
hughcapet, line 406: maybe also assert `result.output`?

hughcapet (approved):

PR #3234, opened 2024-12-16
Fix FromAsCasing warning when building docker image due to uppercase

joaofoltran:
Docker build gives **FromAsCasing** warning due to uppercase mismatch on FROM/AS clauses.


joaofoltran:
Mistake, sorry!

PR #3244, opened 2025-01-14
Avoid dropping physical slots too early

CyberDem0n:
Consider a situation: there is a permanent logical slot and primary and replica are temporary down.
When Patroni is started on the former primary it starts Postgres in a standby mode, what leads to removal of physical replication slot for the replica because it has xmin.

We should postpone removal of such physical slots:
- on replica until there will be a leader in the cluster
- on primary until Postgres is promoted

### DIFF
--- a/patroni/postgresql/slots.py
+++ b/patroni/postgresql/slots.py
@@ -372,6 +372,8 @@ class SlotsHandler:
 
         :param slots: A dictionary mapping slot name to slot attributes. This method only considers a slot
                       if the value is a dictionary with the key ``type`` and a value of ``physical``.
+        :param clean_inactive_physical_slots: whether replication slots with ``xmin`` and not expected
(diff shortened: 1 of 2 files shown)


patroni/postgresql/slots.py
```diff
@@ -365,7 +365,7 @@ def _drop_incorrect_slots(self, cluster: Cluster, slots: Dict[str, Any]) -> None
                     logger.error("Failed to drop replication slot '%s'", name)
                     self._schedule_load_slots = True
 
-    def _ensure_physical_slots(self, slots: Dict[str, Any]) -> None:
+    def _ensure_physical_slots(self, slots: Dict[str, Any], clean_inactive_physical_slots: bool) -> None:
```
hughcapet, line 368: docstring update

hughcapet (approved):

PR #3245, opened 2025-01-14
Don't touch logical failover slots

CyberDem0n:
If logical replication slot is created with `failover => true` option, we get respective field set to true in `pg_replication_slots` view.

By avoiding interacting with such slots we make logical failover slots feature fully functional in PG17.


hughcapet:
behave tests

hughcapet (approved):

PR #3247, opened 2025-01-16
Replace forward slash in cluster_name

mib32:
If cluster name was defined with a "/" in it, which is actually normal if `pg_createcluster` was used, then trying to launch `patronictl edit-config` will result in an error that looks like this:

```
...
  File "/usr/lib/python3.12/tempfile.py", line 395, in _mkstemp_inner
    fd = _os.open(file, flags, 0o600)
         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
FileNotFoundError: [Errno 2] No such file or directory: '/tmp/14/main-config-guh3an57.yaml'
```

It took me a while to understand what's going on, that it was trying to save temporary file in `/tmp/14/` (my cluster name is `14/main`), so I propose this change that would not make it happen.

### DIFF
--- a/patroni/ctl.py
+++ b/patroni/ctl.py
@@ -2065,9 +2065,10 @@ def invoke_editor(before_editing: str, cluster_name: str) -> Tuple[str, Dict[str
     if not editor_cmd:
         raise PatroniCtlException('EDITOR environment variable is not set. editor or vi are not available')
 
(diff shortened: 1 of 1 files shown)


CyberDem0n (approved):

mib32:
Thank you 🤩

PR #3250, opened 2025-01-23
Readiness API checks for replication on replicas

ants:
Readiness probes on Kubernetes are used for a few different things. One is for determining pod disruption budget. Previous implementation considered replicas ready as soon as PostgreSQL was started. This could cause issues with async replication and rolling upgrades, where the primary would be shut down before replica had even the time to start replicating. Second important use is for determining which pods get included as endpoints in a service. For both cases we would like to see the pod only be considered ready when it is replicating and not too far behind.

Also changes readiness to consider replicas as ready when failsafe is active. In that case lag is ignored.

Open questions and notes:

* Currently the new code lacks test coverage. I would prefer to add tests that check that the response code is right, but got stuck figuring out how the mocking system works. Also it would be nice to cross verify that get_postgresql_status fields do not get renamed. Are any existing tests a good template for this type of testing?
* I am basing the lag on cluster.status.last_lsn, but should it perhaps be max(cluster.status.last_lsn, pg_stat_wal_receiver.latest_end_lsn)?
* For diagnosing I added a log entry that logs an info line when readiness check fails. Any opinions on whether it should be a debug message instead. I also considered logging only when the status changes, but that seemed too much complexity.
* I made the default mode check replay lag so people who are using a read-only replica service get somewhat useful behavior.

### DIFF
--- a/patroni/api.py
+++ b/patroni/api.py
@@ -445,27 +445,75 @@ class RestApiHandler(BaseHTTPRequestHandler):
         status_code = 200 if patroni.ha.is_paused() or patroni.next_run + liveness_threshold > time.time() else 503
         self._write_status_code_only(status_code)
 
+    def _readiness(self) -> Optional[str]:
+        """Check if readiness conditions are met
+
+        :returns: None if node can be considered ready or diagnostic message if not."""
+
+        patroni = self.server.patroni
+        if patroni.ha.is_leader():
+            # We only become leader after bootstrap or once up as a standby, so we are definitely ready.
+            return
+
+        # When postgres is not running we are not ready.
+        if patroni.postgresql.state != 'running':
+            return 'PostgreSQL is not running'
+
+        postgres = self.get_postgresql_status(True)
+
+        if postgres.get('replication_state') != 'streaming':
+            return 'PostgreSQL replication state is not streaming'
+
+        cluster = patroni.dcs.cluster
+
+        if not cluster:
+            if patroni.ha.failsafe_is_active():
+                return
+            return 'DCS is not accessible'
+
+        leader_optime = cluster.status.last_lsn
+
+        mode = 'write' if self.path_query.get('mode') == 'write' else 'apply'
+        location = 'received_location' if mode == 'write' else 'replayed_location'
+        lag = leader_optime - postgres.get('xlog', {}).get(location, 0)
+
+        logger.info("Lag: %s", self.path_query.get('lag', [None])[0])
+        max_replica_lag = parse_int(self.path_query.get('lag', [None])[0], 'B')
+        if max_replica_lag is None:
+            max_replica_lag = global_config.maximum_lag_on_failover
+
+        if lag > max_replica_lag:
+            return f'Replication {mode} lag {lag} exceeds maximum allowable {max_replica_lag}'
+
+        return
+
     def do_GET_readiness(self) -> None:
         """Handle a ``GET`` request to ``/readiness`` path.
+
+            * Query parameters:
+
+                * ``lag``: only accept replication lag up to ``lag``. Accepts either an :class:`int`, which
+                    represents lag in bytes, or a :class:`str` representing lag in human-readable format (e.g.
+                    ``10MB``).
+                * ``mode``: allowed values ``write``, ``apply``. Base replication lag off of received WAL or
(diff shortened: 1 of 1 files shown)


patroni/api.py
```diff
@@ -445,27 +445,75 @@ def do_GET_liveness(self) -> None:
...
+
+        logger.info("Lag: %s", self.path_query.get('lag', [None])[0])
+        max_replica_lag = parse_int(self.path_query.get('lag', [None])[0], 'B')
+        if max_replica_lag is None:
+            max_replica_lag = global_config.maximum_lag_on_failover
+
+        if lag > max_replica_lag:
+            return f'Replication {mode} lag {lag} exceeds maximum allowable {max_replica_lag}'
+
+        return
```
hughcapet, line 488:
```suggestion
```

CyberDem0n:
> Currently the new code lacks test coverage. I would prefer to add tests that check that the response code is right, but got stuck figuring out how the mocking system works. Also it would be nice to cross verify that get_postgresql_status fields do not get renamed. Are any existing tests a good template for this type of testing?

get_postgresql_status() mostly using results returned from https://github.com/patroni/patroni/blob/7db7dfd3c5ff1d04d1d575badfe2e98fc20b99ff/tests/test_api.py#L33-L37

I think the simplest would be doing something `patch.object(MockConnection, 'query', Mock(return_value=...some other query result...))` somewhere here: https://github.com/patroni/patroni/blob/7db7dfd3c5ff1d04d1d575badfe2e98fc20b99ff/tests/test_api.py#L342-L347.

This way you will not have to mock the entire get_postgresql_status() function and if fields are renamed tests will not break.

To check the response code you can to something like: https://github.com/patroni/patroni/blob/7db7dfd3c5ff1d04d1d575badfe2e98fc20b99ff/tests/test_api.py#L517-L519

> I am basing the lag on cluster.status.last_lsn, but should it perhaps be max(cluster.status.last_lsn, pg_stat_wal_receiver.latest_end_lsn)?

I think you can do something like `max(cluster.status.last_lsn, received_location, replayed_location)`, where received_location and replayed_location are values returned by get_postgresql_status(). pg_last_wal_receive_lsn() should return values which are effectively the same as pg_stat_wal_receiver.latest_end_lsn.
Actually, we should do the same in do_GET().

> For diagnosing I added a log entry that logs an info line when readiness check fails. Any opinions on whether it should be a debug message instead. I also considered logging only when the status changes, but that seemed too much complexity.

I think debug is good enough.

> I made the default mode check replay lag so people who are using a read-only replica service get somewhat useful behavior.

Health-check endpoints work the same way.

CyberDem0n:
> latest_end_lsn is the WAL flush position on the source host, if replication is not keeping up then that might be quite a bit ahead of pg_last_wal_receive_lsn().

Oh, I always forget that pg_last_wal_receive_lsn() returns flush lsn...
I just don't understand why docs are saying about latest_end_lsn "_Last write-ahead log location **reported to origin WAL sender**_", while [comment](https://github.com/postgres/postgres/blob/035f99cbebe5ffcaf52f8370394446cd59621ab7/src/backend/replication/walreceiver.c#L1260-L1266) says the opposite:
>  \* 'walEnd' and 'sendTime' are the end-of-WAL and timestamp of the latest
> \* message, **reported by primary.**

It seems to be a good idea to start using it for calculating lag.

On a side note, written_lsn returns written, but not yet flushed LSN, which we interested in when finding the most up-to-date node, and it is available since v13.

CyberDem0n:
> It seems to be a good idea to start using it for calculating lag.

Here we go: https://github.com/patroni/patroni/pull/3268/files#diff-13329cd3342d1757c2c9b4dd7a26c0170f4deee5d32eff5c2fd719cc9ff9375bR319

PTAL, @ants

ants:
Finally got back to this. Review feedback incorporated and tests added. Ready from my point of view.

hughcapet (approved):

hughcapet (approved):

dkapanidis:
Hello,

Regarding this PR it was merged to master branch on May 30, there was a release afterwards (v4.0.6 on June 6), but if I read correctly the git history it wasn't included on the release branch itself, so it didn't make it on v4.0.6.

Was this intentional, will it be included on next releases?

Issue #3251, opened 2025-01-23
Suppress duplicate heartbeat messages


MichaelMorrisEst:
### What happened?

Patroni logs a message for each loop iteration as follows:
INFO: no action. I am (...), the <leader/secondary>...

The messages are the same for each iteration except when the leader changes.

### How can we reproduce it (as minimally and precisely as possible)?

Setup a replicated database

### What did you expect to happen?

There have been issues opened in the past looking to change this behaviour so as INFO level messages are not logged every loop iteration, none of which of come to fruition (although https://github.com/patroni/patroni/pull/1955 was a good improvement)

I would like to propose the following:
- Introduce a new boolean configuration option named something like suppress_duplicate_hb_logs
- When set to true the "No action..." logs will only be output if they differ from the previous log
- When set to true the "Lock owner..." logs will only be output if the lock owner has changed
	- Suppressing the "No action..." messages would cause these "Lock owner..." message to start being logged every loop iteration otherwise
- The default shall be False so behaviour will continue as today unless the user decides to set the option to True

This would allow the exact same behaviour as today for users who would like to get these logs for every loop while also supporting users who only want to get a log when something has changed.

I will push a PR so you can see what the code could look like if there is agreement on the approach.

### Patroni/PostgreSQL/DCS version

-   Patroni version: 4.0.4
-   PostgreSQL version: 16
-   DCS (and its version): 

### Patroni configuration file

```yaml
N/A
```

### patronictl show-config

```yaml
N/A
```

### Patroni log files

```shell
patroni1  | 2025-01-23 11:05:48,332 INFO: no action. I am (patroni1), a secondary, and following a leader (patroni3)
patroni3  | 2025-01-23 11:05:52,729 INFO: no action. I am (patroni3), the leader with the lock
patroni2  | 2025-01-23 11:05:53,286 INFO: no action. I am (patroni2), a secondary, and following a leader (patroni3)
```

### PostgreSQL log files

```shell
N/A
```

### Have you tried to use GitHub issue search?

- [x] Yes

PR #3252, opened 2025-01-23
Add configuration option to suppress duplicate heartbeat logs

MichaelMorrisEst:
Proposed solution for #3251

### DIFF
--- a/patroni/__main__.py
+++ b/patroni/__main__.py
@@ -78,6 +78,8 @@ class Patroni(AbstractPatroniDaemon, Tags):
         self.next_run = time.time()
         self.scheduled_restart: Dict[str, Any] = {}
 
+        self.previous_heartbeat_msg = ''
+
     def ensure_dcs_access(self, sleep_time: int = 5) -> 'Cluster':
         """Continuously attempt to retrieve cluster from DCS with delay.
 
@@ -209,7 +211,13 @@ class Patroni(AbstractPatroniDaemon, Tags):
         the change and cache the new dynamic configuration values in ``patroni.dynamic.json`` file under Postgres data
         directory.
         """
-        logger.info(self.ha.run_cycle())
+        msg = self.ha.run_cycle()
+
+        suppress_duplicate_logs = self.config.get('log', {}).get('suppress_duplicate_hb_logs', False)
+        if not msg == self.previous_heartbeat_msg or not suppress_duplicate_logs:
+            logger.info(msg)
+
+        self.previous_heartbeat_msg = msg
 
         if self.dcs.cluster and self.dcs.cluster.config and self.dcs.cluster.config.data \
                 and self.config.set_dynamic_configuration(self.dcs.cluster.config):
--- a/patroni/ha.py
+++ b/patroni/ha.py
@@ -263,6 +263,8 @@ class Ha(object):
         # used only in backoff after failing a pre_promote script
         self._released_leader_key_timestamp = 0
 
+        self.previous_lock_owner = ''
+
         # Initialize global config
         global_config.update(None, self.patroni.config.dynamic_configuration)
 
@@ -400,8 +402,14 @@ class Ha(object):
 
     def has_lock(self, info: bool = True) -> bool:
         lock_owner = self.cluster.leader and self.cluster.leader.name
+
         if info:
-            logger.info('Lock owner: %s; I am %s', lock_owner, self.state_handler.name)
+            suppress_duplicate_logs = self.patroni.config.get('log', {}).get('suppress_duplicate_hb_logs', False)
+            if not lock_owner == self.previous_lock_owner or not suppress_duplicate_logs:
+                logger.info('Lock owner: %s; I am %s', lock_owner, self.state_handler.name)
+
+        self.previous_lock_owner = lock_owner
+
         return lock_owner == self.state_handler.name
 
     def get_effective_tags(self) -> Dict[str, Any]:


CyberDem0n (commented):
I don't like it.
You are suppressing all kinds of logs, not only heartbeat.
For example, 'bootstrap from {} {} in progress', 'waiting for leader to bootstrap', 'data directory is not accessible' and so on.
When something is wrong you better know whether it is moving rather then total silence.
Also, there is a better place to implement it: https://github.com/patroni/patroni/blob/7db7dfd3c5ff1d04d1d575badfe2e98fc20b99ff/patroni/log.py#L531-L540

ants:
Documentation also needs to come with a fat warning that you will lose capability to diagnose failovers due to resource exhaustion and similar problems. Information about when the HA loop was able to execute is very valuable in such cases.

MichaelMorrisEst:
Thanks for the feedback, I have pushed a new commit based on the comments

patroni/log.py
```diff
@@ -535,15 +536,25 @@ def run(self) -> None:
...
+                    if not record.msg == prev_hb_msg or not suppress_duplicate_logs:
+                        self.log_handler.handle(record)
+                    prev_hb_msg = record.msg
+                else:
+                    self.log_handler.handle(record)
 
             self._queue_handler.queue.task_done()
 
+    def isHeartBeat(self, record: logging.LogRecord) -> bool:
+        return record.msg.startswith('no action. ') or record.msg.startswith('PAUSE: no action')
```
CyberDem0n, line 556: We mostly use snake case. Also, this is static method, because it doesn't use self.
```suggestion
    @staticmethod
    def _is_heartbeat_msg(record: logging.LogRecord) -> bool:
        return record.msg.startswith('no action. ') or record.msg.startswith('PAUSE: no action')
```

patroni/log.py
```diff
@@ -535,15 +536,25 @@ def run(self) -> None:
...
                 else:
                     if prev_record and prev_record.thread == record.thread:
-                        if not (record.msg.startswith('no action. ') or record.msg.startswith('PAUSE: no action')):
+                        if not self.isHeartBeat(record):
                             self.log_handler.handle(prev_record)
                         prev_record = None
 
             if record:
-                self.log_handler.handle(record)
+                if self.isHeartBeat(record):
```
CyberDem0n, line 544:
```suggestion
                if self._is_heartbeat_msg(record):
```

patroni/log.py
```diff
@@ -535,15 +536,25 @@ def run(self) -> None:
                     prev_record, record = record, None
                 else:
                     if prev_record and prev_record.thread == record.thread:
-                        if not (record.msg.startswith('no action. ') or record.msg.startswith('PAUSE: no action')):
+                        if not self.isHeartBeat(record):
```
CyberDem0n, line 539:
```suggestion
                        if not self._is_heartbeat_msg(record):
```

patroni/log.py
```diff
@@ -535,15 +536,25 @@ def run(self) -> None:
...
             if record:
-                self.log_handler.handle(record)
+                if self.isHeartBeat(record):
+                    config = self._config or {}
+                    suppress_duplicate_logs = config.get('suppress_duplicate_hb_logs', False)
+                    if not record.msg == prev_hb_msg or not suppress_duplicate_logs:
+                        self.log_handler.handle(record)
+                    prev_hb_msg = record.msg
+                else:
+                    self.log_handler.handle(record)
```
CyberDem0n, line 551: I think in this case we also need to reset previous heartbeat message variable.

patroni/log.py
```diff
@@ -535,15 +536,25 @@ def run(self) -> None:
...
+                        if not self.isHeartBeat(record):
                             self.log_handler.handle(prev_record)
                         prev_record = None
 
             if record:
-                self.log_handler.handle(record)
+                if self.isHeartBeat(record):
+                    config = self._config or {}
+                    suppress_duplicate_logs = config.get('suppress_duplicate_hb_logs', False)
+                    if not record.msg == prev_hb_msg or not suppress_duplicate_logs:
```
CyberDem0n, line 547:
```suggestion
                    if record.msg != prev_hb_msg or not suppress_duplicate_logs:
```

docs/yaml_configuration.rst
```diff
@@ -36,6 +36,10 @@ Log
 
    -  **patroni.postmaster: WARNING**
    -  **urllib3: DEBUG**
+-  **suppress_duplicate_hb_logs**: If set to ``true``, successive heartbeat logs that are identical shall not be output. Default value is ``false``.
```
CyberDem0n, line 39: What about calling it `deduplicate_heartbeat_logs`?

patroni/log.py
MichaelMorrisEst: Done

docs/yaml_configuration.rst
MichaelMorrisEst: Done

patroni/log.py
CyberDem0n: After thinking a bit more on it, I have a strong feeling that there is something wrong.
The original filter that removes `Lock owner:` log messages works under assumptions that:
1. log level is INFO (if it is set to WARNING there will be no messages, and if it is set to DEBUG we want to see everything anyway)
2. Thread ID for `Lock owner:` and heartbeat messages is the same.

Now you moved deduplication outside of the original place, what at least may result in violation of the second condition.
If we start fixing it - it will make the code even more complex.

IMO, deduplication should happen somewhere inside https://github.com/patroni/patroni/pull/3252/files#diff-c8c8e070a28e83571297417a0d0bb7f4acff6f01a7eeba2c606f028de42eaaf7R538-R541

MichaelMorrisEst: I'm struggling to see the relevance of the thread ID - is there a scenario in which two different threads in the same process will be executing this code? Even if there is, are the variables prev_record, prev_hb_msg not thread safe anyway?

CyberDem0n: Logs could be written from different threads, but they will end up in the same queue.
While filtering messages we want to focus on messages produced from the main thread (only the main thread producing heartbeat messages). All messages from other threads should be just processed.

`Lock owner:` messages are produces only from the main thread, and we abuse comparison of its thread id with other messages and identify heartbeat messages (`no action.`) only if thread id matches.

patroni/config.py
```diff
@@ -541,7 +541,7 @@ def _set_section_values(section: str, params: List[str]) -> None:
         _set_section_values('postgresql', ['listen', 'connect_address', 'proxy_address',
                                            'config_dir', 'data_dir', 'pgpass', 'bin_dir'])
         _set_section_values('log', ['type', 'level', 'traceback_level', 'format', 'dateformat', 'static_fields',
-                                    'max_queue_size', 'dir', 'mode', 'file_size', 'file_num', 'loggers'])
+                                    'max_queue_size', 'dir', 'mode', 'file_size', 'file_num', 'loggers', 'deduplicate_heartbeat_logs'])
```
hughcapet, line 544: flake8:

```
patroni/config.py:544:121: E501 line too long (135 > 120 characters)
```

MichaelMorrisEst: Fixed

patroni/log.py
```diff
@@ -535,15 +536,27 @@ def run(self) -> None:
...
+                            prev_hb_msg = None
                         prev_record = None
 
             if record:
                 self.log_handler.handle(record)
 
             self._queue_handler.queue.task_done()
 
+    @staticmethod
+    def _is_heartbeat_msg(record: logging.LogRecord) -> bool:
```
CyberDem0n, line 557: Please also add a docstring.

patroni/config.py
```diff
@@ -541,7 +541,8 @@ def _set_section_values(section: str, params: List[str]) -> None:
         _set_section_values('postgresql', ['listen', 'connect_address', 'proxy_address',
                                            'config_dir', 'data_dir', 'pgpass', 'bin_dir'])
         _set_section_values('log', ['type', 'level', 'traceback_level', 'format', 'dateformat', 'static_fields',
-                                    'max_queue_size', 'dir', 'mode', 'file_size', 'file_num', 'loggers'])
+                                    'max_queue_size', 'dir', 'mode', 'file_size', 'file_num', 'loggers',
+                                    'deduplicate_heartbeat_logs'])
```
CyberDem0n, line 545: The value also should be converted from str to bool.
To do that you need to add it to https://github.com/patroni/patroni/blob/b4eab4897161eb15b37c6995991d76bf56de9ff7/patroni/config.py#L552-L553

patroni/log.py
```diff
@@ -535,15 +536,27 @@ def run(self) -> None:
...
                 else:
                     if prev_record and prev_record.thread == record.thread:
-                        if not (record.msg.startswith('no action. ') or record.msg.startswith('PAUSE: no action')):
+                        if self._is_heartbeat_msg(record):
+                            config = self._config or {}
+                            deduplicate_heartbeat_logs = config.get('deduplicate_heartbeat_logs', False)
+                            if record.msg == prev_hb_msg and deduplicate_heartbeat_logs:
+                                record = None
+                            else:
+                                prev_hb_msg = record.msg
```
CyberDem0n, line 545: Now it is much easier to understand. Thank you!
However, it would be also nice to add respective unit tests for it.

You can take as an example a test that covers `Lock owner` not being removed when it is not followed by heartbeat message: https://github.com/patroni/patroni/blob/b4eab4897161eb15b37c6995991d76bf56de9ff7/tests/test_log.py#L87-L94

Since it is multithreaded, it will not be easy to check whether expected log_handler.handle() calls are made, but it is not required. At least I want to see that all added lines are covered.

patroni/config.py
CyberDem0n: There is also config validator that needs to be updated: https://github.com/patroni/patroni/blob/b4eab4897161eb15b37c6995991d76bf56de9ff7/patroni/validator.py#L991-L1005

hughcapet (approved):

CyberDem0n:
Merged, thank you @MichaelMorrisEst!

MichaelMorrisEst:
Thanks @CyberDem0n, @hughcapet , @ants for taking the time to review and for the great feedback

PR #3256, opened 2025-01-24
Fall back to default iter_modules if pyinstaller toc is not present.

### LINKED ISSUE #3255: PyInstaller executable cannot find DCS implementations
### What happened?

Patroni packaged with the latest pyinstaller does not appear to be able to locate the DCS implementations.

There appears to be an incompatibility between the latest version of pyinstaller and the dynamic_loader, it is not finding any dcs modules.

1. Using windows
2. Install the latest pyinstaller(6.11.1)
3. pyinstaller --clean patroni.spec
4. execute patroni.exe with a proper config file.

Patroni to report and use the available DCS implementations.

-   Patroni version: 4.0.4
-   PostgreSQL version: 17.2
-   DCS (and its version): consul 1.20.2, though no implementations can be found

```yaml
...
consul:
  url: https://localhost:8501
  register_service: true
  cacert: 'C:\projects\ca.pem'
...
```

### patronictl show-config

```yaml
Error: Can not find suitable configuration of distributed configuration store
Available implementations:
```

### Patroni log files

```shell
Error: Can not find suitable configuration of distributed configuration store
Available implementations:
```

### PostgreSQL log files

```shell
N/A
```

- [x] Yes

There is a special case for pyinstaller in dynamic loader:
```
    if getattr(sys, 'frozen', False):
        toc: Set[str] = set()
        # dirname may contain a few dots, which causes pkgutil.iter_importers()
        # to misinterpret the path as a package name. This can be avoided
        # altogether by not passing a path at all, because PyInstaller's
        # FrozenImporter is a singleton and registered as top-level finder.
        for importer in pkgutil.iter_importers():
            if hasattr(importer, 'toc'):
                toc |= getattr(importer, 'toc')
        dots = module_prefix.count('.')  # search for modules only on the same level
        return [module for module in toc if module.startswith(module_prefix) and module.count('.') == dots]
```

If I remove this code entirely then the modules load correctly and patroni works.  PyInstaller seems to have added direct support for iter_modules back in v4.4, [see here](https://github.com/pyinstaller/pyinstaller/pull/5959/files)

We could also simply fall through if len(toc) == 0 to preserve the old behavior if someone is using an older version of pyinstaller.

joejensen:
fixes #3255 by eliminating a workaround that should not be necessary after pyinstaller 4.4 and appears to cause problems on the latest version of pyinstaller.

### DIFF
--- a/patroni/dynamic_loader.py
+++ b/patroni/dynamic_loader.py
@@ -38,8 +38,11 @@ def iter_modules(package: str) -> List[str]:
         for importer in pkgutil.iter_importers():
             if hasattr(importer, 'toc'):
                 toc |= getattr(importer, 'toc')
-        dots = module_prefix.count('.')  # search for modules only on the same level
-        return [module for module in toc if module.startswith(module_prefix) and module.count('.') == dots]
+        # If it found the pyinstaller toc then use it, otherwise fall through to default
+        # behavior which works in pyinstaller >= 4.4
+        if len(toc) > 0:
+            dots = module_prefix.count('.')  # search for modules only on the same level
+            return [module for module in toc if module.startswith(module_prefix) and module.count('.') == dots]
 
     # here we are making an assumption that the package which is calling this function is already imported
     pkg_file = sys.modules[package].__file__


joejensen:
Modified the commit to fall through to the default behavior only if the pyinstaller table of contents is not found, so old versions should continue the same as before.  New versions don't expose the toc in the same manner, but the default behavior should work.

CyberDem0n (approved):

PR #3261, opened 2025-01-31
Fix bug with slot for former leader not retained on failover

CyberDem0n:
the problem existed because _build_retain_slots() method was falsely relying on members being present in DCS, while on failover the member key for the former leader is expiring exactly at the same time.


mbanck:
Thanks, this fixes the problem I saw

hughcapet (approved):

PR #3268, opened 2025-02-05
Take advantage of written_lsn and latest_end_lsn from pg_stat_wal_receiver

CyberDem0n:
The first one if available starting from PostgreSQL v13 and contains the real write LSN. We will prefer it over value returned by pg_last_wal_receive_lsn(), which is in fact flush LSN.

The second one is available starting from PostgreSQL v9.6 and  points to WAL flush on the source host. In case of primary it will allow to better calculate the replay lag, because values stored in DCS are updated only every loop_wait seconds.


hughcapet:
patroni/api.py:1357 is not covered by tests

hughcapet (approved):

PR #3293, opened 2025-02-24
Convert states to enums

hughcapet:
- Postgresql._state
- pg_isready state

### DIFF
--- a/patroni/dcs/__init__.py
+++ b/patroni/dcs/__init__.py
@@ -295,8 +295,10 @@ class Member(Tags, NamedTuple('Member',
 
     @property
     def is_running(self) -> bool:
-        """``True`` if the member :attr:`~Member.state` is ``running``."""
-        return self.state == 'running'
+        """``True`` if the member :attr:`~Member.state` is ``PostgresqlState.RUNNING``."""
+        from ..postgresql.misc import PostgresqlState
+
+        return self.state == PostgresqlState.RUNNING
 
     @property
     def patroni_version(self) -> Optional[Tuple[int, ...]]:
--- a/patroni/dcs/__init__.py
+++ b/patroni/dcs/__init__.py
@@ -295,7 +295,7 @@ class Member(Tags, NamedTuple('Member',
 
     @property
     def is_running(self) -> bool:
-        """``True`` if the member :attr:`~Member.state` is ``PostgresqlState.RUNNING``."""
+        """``True`` if the member :attr:`~Member.state` is :class:`PostgresqlState.RUNNING`."""
         from ..postgresql.misc import PostgresqlState
 
         return self.state == PostgresqlState.RUNNING
--- a/patroni/dcs/__init__.py
+++ b/patroni/dcs/__init__.py
@@ -295,7 +295,7 @@ class Member(Tags, NamedTuple('Member',
 
     @property
     def is_running(self) -> bool:
-        """``True`` if the member :attr:`~Member.state` is :class:`PostgresqlState.RUNNING`."""
+        """``True`` if the member :attr:`~Member.state` is :class:`~patroni.postgresql.misc.PostgresqlState.RUNNING`."""
         from ..postgresql.misc import PostgresqlState
 
         return self.state == PostgresqlState.RUNNING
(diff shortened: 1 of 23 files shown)


patroni/postgresql/__init__.py
```diff
@@ -39,14 +40,24 @@
...
 
-STATE_RUNNING = 'running'
-STATE_REJECT = 'rejecting connections'
-STATE_NO_RESPONSE = 'not responding'
-STATE_UNKNOWN = 'unknown'
-
 STOP_POLLING_INTERVAL = 1
 
 
+class PgIsReadyStatus(str, Enum):
```
CyberDem0n, line 46: IMO, it should be `IntEnum`. String values aren't used anywhere in the code for reporting, only for comparing it with other strings.
As a result, we don't need to have mapping [here](https://github.com/patroni/patroni/pull/3293/files#diff-72030eab48ce77175c998a2fd1ee9e84fe950e132b66972aa72496532a3923e5R334-R338).
And, actually, `pg_isready(self)` should be returning `PgIsReadyStatus`, not `str`.

patroni/postgresql/__init__.py
```diff
@@ -1031,13 +1042,14 @@ def restart(self, timeout: Optional[float] = None, task: Optional[CriticalTask]
...
-        self.set_state('restarting')
+        self.set_state(PostgresqlState.RESTARTING)
         if not block_callbacks:
             self.__cb_pending = CallbackAction.ON_RESTART
         ret = self.stop(block_callbacks=True, before_shutdown=before_shutdown)\
             and self.start(timeout, task, True, role, after_start)
         if not ret and not self.is_starting():
-            self.set_state('restart failed ({0})'.format(self.state))
+            self.set_state(PostgresqlState.RESTART_FAILED)
+            logger.warning('restart failed (%r)', self.state)
```
CyberDem0n, line 1052: We want to write logs _before_ changing the state.
```suggestion
            logger.warning('restart failed (%r)', self.state)
            self.set_state(PostgresqlState.RESTART_FAILED)
```

patroni/postgresql/bootstrap.py
```diff
@@ -114,7 +115,9 @@ def option_is_allowed(name: str) -> bool:
         return user_options
 
     def _initdb(self, config: Any) -> bool:
-        self._postgresql.set_state('initializing new cluster')
+        # from ..postgresql.misc import PostgresqlState
+
```
CyberDem0n, line 119:
```suggestion
```

patroni/postgresql/bootstrap.py
```diff
@@ -240,8 +243,9 @@ def create_replica(self, clone_member: Union[Leader, Member, None]) -> Optional[
             defined by the user.  this is a list, so we need to
             loop through all methods the user supplies
         """
+        # from ..postgresql.misc import PostgresqlState
```
CyberDem0n, line 246:
```suggestion
```

patroni/postgresql/misc.py
```diff
@@ -2,13 +2,38 @@
...
+from enum import Enum
 from typing import Iterable, Tuple
 
 from ..exceptions import PostgresException
 
 logger = logging.getLogger(__name__)
 
 
+class PostgresqlState(str, Enum):
+    """Possible values of :attr:`Postgresql._state`."""
```
CyberDem0n, line 14: Lest better refer to a public attribute.
```suggestion
    """Possible values of :attr:`Postgresql.state`."""
```

patroni/dcs/__init__.py
```diff
@@ -295,8 +295,10 @@ def state(self) -> str:
 
     @property
     def is_running(self) -> bool:
-        """``True`` if the member :attr:`~Member.state` is ``running``."""
-        return self.state == 'running'
+        """``True`` if the member :attr:`~Member.state` is :class:`PostgresqlState.RUNNING`."""
```
CyberDem0n, line 298: I think it should be `:attr:` instead of `:class:`?
```suggestion
        """``True`` if the member :attr:`~Member.state` is :attr:`PostgresqlState.RUNNING`."""
```

hughcapet: although something is still off here

CyberDem0n (approved):

Issue #3300, opened 2025-03-03
Suppport systemd "notify" service


rdunklau:
### What happened?

As of now, it is unsafe to reload patroni unconditionally as the SIGHUP signal could hit it before it has setup it's signal handlers.

It could be argued that patroni sets up the SIGHUP signal handler too late,  but reducing the window would never make it disappear.

### How can we reproduce it (as minimally and precisely as possible)?

Run systemctl start patroni && systemctl reload patroni, and look at the systemctl status for the patroni unit:

Mar 03 15:35:00 patroni-2nodes-1 systemd[1]: Starting patroni.service - Runners to orchestrate a high-availability PostgreSQL...
Mar 03 15:35:00 patroni-2nodes-1 systemd[1]: Started patroni.service - Runners to orchestrate a high-availability PostgreSQL.
Mar 03 15:35:00 patroni-2nodes-1 systemd[1]: Reloading patroni.service - Runners to orchestrate a high-availability PostgreSQL...
Mar 03 15:35:00 patroni-2nodes-1 systemd[1]: Reloaded patroni.service - Runners to orchestrate a high-availability PostgreSQL.
Mar 03 15:35:00 patroni-2nodes-1 systemd[1]: patroni.service: Deactivated successfully.

Patroni receives the SIGHUP before it has a chance to install a signal handler, and exits. As far as systemd is concerned, the service as exited gracefully.

### What did you expect to happen?

Patroni to be running

### Patroni/PostgreSQL/DCS version

-   Patroni version: 4.0.5
-   PostgreSQL version: Not applicable
-   DCS (and its version):  Not applicable

### Patroni configuration file

```yaml
Not applicable
```

### patronictl show-config

```yaml
Not applicable
```

### Patroni log files

```shell
Not applicable
```

### PostgreSQL log files

```shell
Not applicable
```

### Have you tried to use GitHub issue search?

- [x] Yes

### Anything else we need to know?

Workaround is to insert some sleep time to let patroni actually start before sending any reload command. 
But the proper way would be just use `sd_notify("READY=1")` and have a `Type=notify` systemd unit.

PR #3301, opened 2025-03-04
Add support for systemd "notify" unit type

rdunklau:
Add support for systemd "notify" unit type
    
Without a notify unit type, it is possible to start patroni and
immediately send it a SIGHUP signal using systemd, effectively killing
it before it had time to setup it's signal handlers.

See issue #3300

### DIFF
--- a/patroni/daemon.py
+++ b/patroni/daemon.py
@@ -7,6 +7,7 @@ from __future__ import print_function
 
 import abc
 import argparse
+import logging
 import os
 import signal
 import sys
@@ -16,6 +17,8 @@ from typing import Any, Optional, Type, TYPE_CHECKING
 
 if TYPE_CHECKING:  # pragma: no cover
     from .config import Config
+
+logger = logging.getLogger(__name__)
 
 
 def get_base_arg_parser() -> argparse.ArgumentParser:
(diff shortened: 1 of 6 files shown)


patroni/daemon.py
```diff
@@ -134,6 +134,11 @@ def run(self) -> None:
         Start the logger thread and keep running execution cycles until a SIGTERM is eventually received. Also reload
         configuration upon receiving SIGHUP.
         """
+        try:
+            from systemd import daemon
+            daemon.notify("READY=1")
+        except ImportError:
+            logger.warn("Systemd integration is not supported")
```
CyberDem0n, line 141: I am not sure if it should be a warning.
IMO, info is enough.

setup.py
```diff
@@ -26,6 +26,7 @@
 
 EXTRAS_REQUIRE = {'aws': ['boto3'], 'etcd': ['python-etcd'], 'etcd3': ['python-etcd'],
                   'consul': ['py-consul'], 'exhibitor': ['kazoo'], 'zookeeper': ['kazoo'],
+                  'systemd': ['systemd-python'],
```
CyberDem0n, line 29: please also add it to requirements.txt, maybe with specifying a minimal version.

patroni/daemon.py
rdunklau: Done !

setup.py
rdunklau: I did not add a minimal version, since the feature has been available since the first versions of the package.

rdunklau:
Thank you for this quick review and showing interest in the proposal.
I added  minimal documentation per your request.
This is solving an immediate need, but would more (optional) systemd integration be interesting ? I'm thinking about supporting notify-reload as well.

requirements.txt
```diff
@@ -12,3 +12,4 @@ cryptography>=1.4
 psutil>=2.0.0
 ydiff>=1.2.0,<1.5,!=1.4.0,!=1.4.1
 python-json-logger>=2.0.2
+systemd-python
```
CyberDem0n, line 15:
```suggestion
```

CyberDem0n:
Actually, yes, it would be interesting to have better integration.
The only problem is that GH Actions is failing to install `systemd-python`.
I removed it from the requirements.txt for now, to see if tests will pass.

rdunklau:
Ok ! I'll try to make another PR for that then. Let me know if there is anything to change to this one.

hughcapet (approved):

PR #3303, opened 2025-03-12
Convert roles to enums

hughcapet:
- Postgresql.role
- roles used in ctl

### DIFF
--- a/patroni/postgresql/config.py
+++ b/patroni/postgresql/config.py
@@ -1247,7 +1247,7 @@ class ConfigHandler(object):
                     and config.get('pg_hba'):
                 hba_changed = self._config.get('pg_hba', []) != config['pg_hba']
 
-            if (not server_parameters.get('ident_file') or server_parameters['ident_file'] == self._pg_hba_conf) \
+            if (not server_parameters.get('ident_file') or server_parameters['ident_file'] == self._pg_ident_conf) \
                     and config.get('pg_ident'):
                 ident_changed = self._config.get('pg_ident', []) != config['pg_ident']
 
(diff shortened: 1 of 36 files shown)


patroni/postgresql/config.py
```diff
@@ -477,19 +477,16 @@ def _configuration_to_save(self) -> List[str]:
...
-            permissions otherwise.
+            Do nothing if the file is not under PGDATA.
 
         :param filename: path to a file which permissions might need to be adjusted.
         """
         if is_subpath(self._postgresql.data_dir, filename):
             pg_perm.set_permissions_from_data_directory(self._postgresql.data_dir)
             os.chmod(filename, pg_perm.file_create_mode)
-        else:
-            os.chmod(filename, 0o666 & ~pg_perm.orig_umask)
```
CyberDem0n, line 490: Looks like you "reverted" #3308 by accident.

tests/test_callback_executor.py
```diff
@@ -38,5 +39,5 @@ def test_callback_executor(self, mock_popen):
         self.assertIsNone(ce.call(callback))
 
         mock_popen.side_effect = [Mock()]
-        self.assertIsNone(ce.call(['test.sh', 'on_reload', 'replica', 'foo']))
+        self.assertIsNone(ce.call(['test.sh', 'on_reload', PostgresqlRole.REPLICA, 'foo']))
```
CyberDem0n, line 42: Actually, here we need to check that callback is called with 'replica' as a string.

CyberDem0n (approved):

Issue #3302, opened 2025-03-11
Out-of-pgdata postgresql.conf gets created with 600 permissions


mbanck-ntap:
### What happened?

Even if the old `postgresql.conf`/new `postgresql.base.conf` has 660 permissions, Patroni creates the new `postgresql.conf` with 600 permissions. I guess what actually happens is that the umask Patroni sets is more restrictive than what is expected and as `set_file_permissions()` only touches files in `PGDATA`, it does not fix this up. This seems to be a behavior change since #2781.

This is especially problematic on Debian, as `pg_lsclusters` reads the `postgresql.conf` and for regular users only displays garbage for Patroni-created instances:

```
Ver Cluster         Port      Status Owner     Data directory              Log file
15  example_cluster <unknown> down   <unknown> <unknown>                   <unknown>
```

I guess `write_postgresql_conf()` should record the file permissions of `postgresql.base.conf` (maybe only if it is not in `PGDATA`?) and then set the file permissions of `postgresql.conf` accordingly. However, that latter part is pretty much abstracted away in the `pg_perm` class so not sure whether we could/should just paper over it from `patroni/postgresql/config.py`

### How can we reproduce it (as minimally and precisely as possible)?

On Debian:
1. Create a Patroni config file with `pg_createconfig_patroni <ver> <instance>` or set `config_dir: /etc/postgresql/<ver>/<instance>` manually
2. Run `systemctl start patroni@<ver>-<instance>`
3. Observe that `/etc/postgresql/<ver>/<instance>/postgresql.conf` has 600 permissions

### What did you expect to happen?

`postgresql.conf` is world-readable, like `postgresql.base.conf`.

### Patroni/PostgreSQL/DCS version

-   Patroni version: master
-   PostgreSQL version: n/a
-   DCS (and its version): n/a

### Patroni configuration file

```yaml
bootstrap:
  method: pg_createcluster
  pg_createcluster:
    command: /usr/share/patroni/pg_createcluster_patroni
postgresql:
  create_replica_method:
    - pg_clonecluster
  pg_clonecluster:
    command: /usr/share/patroni/pg_clonecluster_patroni
  data_dir: /var/lib/postgresql/15/example_cluster
  bin_dir: /usr/lib/postgresql/15/bin
  config_dir: /etc/postgresql/15/example_cluster
```

### patronictl show-config

```yaml
n/a
```

### Patroni log files

```shell
n/a
```

### PostgreSQL log files

```shell
n/a
```

[...]

PR #3308, opened 2025-03-13
Fix permissions of out-of-PGDATA created postgresql.conf

mbanck-ntap:
Since 01d07f86c, the permissions of `postgresql.conf` created in PGDATA was explicitly set. However, the umask of the Patroni process was adjusted as well and as a result of this, Patroni would write `postgresql.conf` with 600 permissions if the configuration files are outside PGDATA.

Fix this by using the original umask as mode for files created outside PGDATA.

Fixes: #3302

### DIFF
--- a/patroni/postgresql/config.py
+++ b/patroni/postgresql/config.py
@@ -477,16 +477,19 @@ class ConfigHandler(object):
         return configuration
 
     def set_file_permissions(self, filename: str) -> None:
-        """Set permissions of file *filename* according to the expected permissions if it resides under PGDATA.
+        """Set permissions of file *filename* according to the expected permissions.
 
         .. note::
-            Do nothing if the file is not under PGDATA.
+            Use original umask if the file is not under PGDATA, use PGDATA
+            permissions otherwise.
 
         :param filename: path to a file which permissions might need to be adjusted.
         """
         if is_subpath(self._postgresql.data_dir, filename):
             pg_perm.set_permissions_from_data_directory(self._postgresql.data_dir)
             os.chmod(filename, pg_perm.file_create_mode)
+        else:
+            os.chmod(filename, 0o666 & ~pg_perm.orig_umask)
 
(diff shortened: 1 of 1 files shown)


CyberDem0n (commented):
Well...
> \# move conffiles, setup permissions
> move_conffile "$datadir/postgresql.conf", $confdir, $version, $cluster, $owneruid, $ownergid, '644';
move_conffile "$datadir/pg_hba.conf", $confdir, $version, $cluster, $owneruid, $ownergid, '640', 'hba_file';
move_conffile "$datadir/pg_ident.conf", $confdir, $version, $cluster, $owneruid, $ownergid, '640', 'ident_file';

`pg_createcluster` sets `0644` on postgresql.conf which is very arguable permission, because this file is expected to contain data which better keep hidden from the public.

Therefore I don't like taking permissions from postgresql.base.conf.

My suggestion was about modifying `set_file_permissions()` method to respect original `umask` value if it creates/updates files outside of PGDATA.

mbanck-ntap:
> `pg_createcluster` sets `0644` on postgresql.conf which is very arguable permission, because this file is expected to contain data which better keep hidden from the public.

Hrm.

> Therefore I don't like taking permissions from postgresql.base.conf.
>
> My suggestion was about modifying `set_file_permissions()` method to respect original `umask` value if we create/update files outside of PGDATA.

That's what I did initially, but then I thought it would be better to use whatever the user had setup `postgresql.conf` with earlier. But ok, I have reverted to using the original umask now, works for me as well.

CyberDem0n (approved):

hughcapet:
can you please also update the description to reflect the actual implementation? @mbanck-ntap

mbanck-ntap:
> can you please also update the description to reflect the actual implementation? @mbanck-ntap

Good point, how about now?

hughcapet (approved):

PR #3315, opened 2025-03-18
WIP: chore(consul): add custom metadata to consul service

octo-dama:
Adding custom metadata to consul service declaration.

It will allow consul_exporter to generate metrics in the same way as tags


CyberDem0n (commented):
What problem are you trying to solve?
Please provide more details and examples.
Besides that, please also update the documentation.

octo-dama:
Hi!

Sorry, I created the PR too soon, it is still work in progress, I will do Doc and testing soon.

To add more context,
We are using consul service metadata on our machines to distribute consul alerts across teams with prometheus and consul_exporter.

It works well when we define the consul service ourselves because we can define the `meta` config for each one but for patroni, we can only customise tags and not meta at the moment.
We could use tags but we don't want to create unnecessary DNS entry in consul (and with our naming convention, it would duplicate an information already in the full domain name but it is not related).

We need this because we are using consul_exporter to get all metrics related to services on our consul servers and it generates metrics based on tags and metadatas of these services.

```
Eg.
Two patroni clusters
- db_app_a
- db_app_b

They would create in consul two dns + subdomain with tags
- primary.db_app_a.service... / replica.db_app_a.service...
- primary.db_app_b.service... / replica.db_app_b.service...
```
If I want to add any kind of info on the consul service, I can only do it with tags. In our case, it could be teams or product line ...
If I want to say that consul service for `db_app_a` is for `team_a` and respectively for `team_b` with `db_app_b`, I can't with meta right now since patroni does not allow configuration for them.
And If I use tag (with `service_tags`) `team_a` or `team_b`, it would create a dns entry which is useless on most cases (`team_a.db_app_a.service...`)

So I just to add the ability to customize patroni consul service metadata: https://developer.hashicorp.com/consul/docs/services/configuration/services-configuration-reference#meta

If you have others questions, feel free to ask.

Thank you

PR #3316, opened 2025-03-19
Provide receive and replay LSN/lag information in API and ctl

hughcapet:
```
postgres@patroni2:~$ patronictl list                                                                                              
+ Cluster: demo (7483920351917613079) --------+----+-------------+-----+------------+-----+                                       
| Member   | Host       | Role    | State     | TL | Receive LSN | Lag | Replay LSN | Lag |
+----------+------------+---------+-----------+----+-------------+-----+------------+-----+
| patroni1 | 172.18.0.7 | Replica | streaming |  1 |   0/4BF1900 |   0 |  0/46964D0 |   5 |
| patroni2 | 172.18.0.6 | Leader  | running   |  1 |             |     |            |     |
| patroni3 | 172.18.0.3 | Replica | streaming |  1 |   0/4BF1900 |   0 |  0/4BF1900 |   0 |
+----------+------------+---------+-----------+----+-------------+-----+------------+-----+
```

```
postgres@patroni1:~$ curl -s localhost:8008/cluster | jq .
{
  "members": [
    {
      "name": "patroni1",
      "role": "replica",
      "state": "streaming",
      "api_url": "http://172.18.0.5:8008/patroni",
      "host": "172.18.0.5",
      "port": 5432,
      "timeline": 1,
      "receive_lag": 0,
      "receive_lsn": "0/4000060",
      "replay_lag": 0,
      "replay_lsn": "0/4000060",
      "lag": 0,
      "lsn": "0/4000060"
    },
    {
      "name": "patroni2",
      "role": "replica",
      "state": "streaming",
      "api_url": "http://172.18.0.4:8008/patroni",
      "host": "172.18.0.4",
      "port": 5432,
      "timeline": 1,
      "receive_lag": 0,
      "receive_lsn": "0/4000060",
      "replay_lag": 0,
      "replay_lsn": "0/4000060",
      "lag": 0,
      "lsn": "0/4000060"
    },
 <..>
}
```


patroni/ctl.py
```diff
@@ -175,6 +175,13 @@ def _get_hline(self) -> str:
         self.__hline_num += 1
         return ret
 
+    def _validate_field_names(self, val: List[str]) -> None:
```
CyberDem0n, line 178: please add doctring

patroni/ctl.py
```diff
@@ -175,6 +175,13 @@ def _get_hline(self) -> str:
...
+        try:
+            super(PatronictlPrettyTable, self)._validate_field_names(val)
+        except ValueError as e:
+            if 'Field names must be unique' not in e.args:
```
CyberDem0n, line 182:
```suggestion
            if 'Field names must be unique' not in str(e):
```

patroni/utils.py
```diff
@@ -967,13 +975,18 @@ def cluster_as_json(cluster: 'Cluster') -> Dict[str, Any]:
...
-            elif cluster_lsn >= lsn:
-                member['lag'] = cluster_lsn - lsn
-            else:
-                member['lag'] = 0
+            for location in ('received_', 'replayed_', ''):
+                lsn_type, lag_type = f'{location}lsn', f'{location}lag'
+
+                lsn = getattr(m, lsn_type)
+                if lsn is None:
+                    member[lsn_type] = member[lag_type] = 'unknown'
```
CyberDem0n, line 983: I wonder if we should set `*_lag` to `unknown`. It looks ugly at least in patronictl output, maybe at least in patronictl we should replace it with empty strings, because LSNs already say `unknown`.

patroni/ha.py
```diff
@@ -462,6 +462,8 @@ def touch_member(self) -> bool:
                     timeline, wal_position, pg_control_timeline = self.state_handler.timeline_wal_position()
                     data['xlog_location'] = self._last_wal_lsn = wal_position
                     if not timeline:  # running as a standby
+                        data['received_location'], data['replayed_location'] =\
```
CyberDem0n, line 465: Should we better call them `receive_lsn` and `replay_lsn`?
After all, they were renamed in PG10, and unlike REST API we don't need to be backward compatible here.

CyberDem0n: Also, we can also rename respective properties/methods in Postgresql and Member classes.

patroni/ctl.py
```diff
@@ -1602,18 +1618,24 @@ def output_members(cluster: Cluster, name: str, extended: bool = False,
...
                 return full_diff if not hide_long or len(full_diff) <= 50 else param + ': [hidden - too long]'
             restart_reason = '\n'.join([format_diff(k, v, fmt in ('pretty', 'topology'))
                                         for k, v in member.get('pending_restart_reason', {}).items()]) or ''
 
+            receive_lag, replay_lag = member.get('received_lag', ''), member.get('replayed_lag', '')
+            receive_lsn, replay_lsn = member.get('received_lsn', ''), member.get('replayed_lsn', '')
+            receive_lag = round(receive_lag / 1024 / 1024) if isinstance(receive_lag, int) \
+                else '' if receive_lag == 'unknown' else receive_lag
+            replay_lag = round(replay_lag / 1024 / 1024) if isinstance(replay_lag, int) \
+                else '' if replay_lag == 'unknown' else replay_lag
```
CyberDem0n, line 1632: What about backward compatibility, when new patronictl is used with old Patroni?
I would suggest to pick values of `lsn` and `lag` and show them either in Receive or Replay columns depending on whether State is `streaming` or not.

patroni/postgresql/__init__.py
```diff
@@ -1259,26 +1259,41 @@ def promote(self, wait_seconds: int, task: CriticalTask,
...
+        """
+        # This method could be called from different threads (simultaneously with some other `_query` calls).
+        # If it is called not from main thread we will create a new cursor to execute statement.
+        if current_thread().ident == self.__thread_ident:
+            replay_lsn = self.replay_lsn()
+            receive_lsn = self.receive_lsn()
+        else:
+            _, _, replay_lsn, receive_lsn = self._query(self.cluster_info_query)[0][:4]
+
+        return receive_lsn, replay_lsn
```
CyberDem0n, line 1295: I think we better not have this method, because when called not from the main thread, on standby, it will result in two separate queries :(
It would be better to return everything what is required from timeline_wal_position() method. It is used only in one place, in ha.py and it shouldn't create any problem.

CyberDem0n (approved):

PR #3324, opened 2025-03-26
Update patroni_configuration.rst

javicane:
Opening a pr. Related issue: https://github.com/patroni/patroni/issues/3323

### DIFF
--- a/docs/patroni_configuration.rst
+++ b/docs/patroni_configuration.rst
@@ -40,20 +40,20 @@ PostgreSQL parameters controlled by Patroni
 
 Some of the PostgreSQL parameters **must hold the same values on the primary and the replicas**. For those, **values set either in the local patroni configuration files or via the environment variables take no effect**. To alter or set their values one must change the shared configuration in the DCS. Below is the actual list of such parameters together with the default and minimal values:
 
-- **max_connections**: default value 100, minimal value 25.
-- **max_locks_per_transaction**: default value 64, minimal value 32.
-- **max_worker_processes**: default value 8, minimal value 2.
-- **max_prepared_transactions**: default value 0, minimal value 0.
-- **wal_level**: default value hot_standby, accepted values: hot_standby, replica, logical.
-- **track_commit_timestamp**: default value off. 
+- **max_connections**: default value 100, minimal value 25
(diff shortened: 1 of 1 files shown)


docs/patroni_configuration.rst
```diff
@@ -41,6 +41,7 @@ PostgreSQL parameters controlled by Patroni
 Some of the PostgreSQL parameters **must hold the same values on the primary and the replicas**. For those, **values set either in the local patroni configuration files or via the environment variables take no effect**. To alter or set their values one must change the shared configuration in the DCS. Below is the actual list of such parameters together with the default values:
 
 - **max_connections**: 100
+  Note: In Patroni, the lower setting of max_connections is set to 25.
```
CyberDem0n, line 44: Formatting with this change will look weird:
image

Maybe just write everything on the same line...

docs/patroni_configuration.rst
```diff
@@ -41,6 +41,7 @@ PostgreSQL parameters controlled by Patroni
 Some of the PostgreSQL parameters **must hold the same values on the primary and the replicas**. For those, **values set either in the local patroni configuration files or via the environment variables take no effect**. To alter or set their values one must change the shared configuration in the DCS. Below is the actual list of such parameters together with the default values:
```
CyberDem0n, line 41:
```suggestion
Some of the PostgreSQL parameters **must hold the same values on the primary and the replicas**. For those, **values set either in the local patroni configuration files or via the environment variables take no effect**. To alter or set their values one must change the shared configuration in the DCS. Below is the actual list of such parameters together with the default and minimal values:
```

docs/patroni_configuration.rst
CyberDem0n, line 44:
```suggestion
- **max_connections**: default value 100, minimal value 25.
```

CyberDem0n (changes requested):
@javicane, since you started changing it, maybe also do it for other parameters: https://github.com/patroni/patroni/blob/1ba9e68b4b0812c4221c251f468340e4a6a69574/patroni/postgresql/config.py#L359-L370

hughcapet (approved):

CyberDem0n (approved):

Issue #3327, opened 2025-04-03
Patroni failing to elect a new leader with synchronous_mode_strict


Maithem:
### What happened?

Patroni is failing to elect a new leader in a three node cluster with synchronous_mode_strict enabled. Since strict mode replicates transactions to at least 1 other node, then the cluster should recover if at least 2 nodes (out of the 3) are up and one of them is a synchronous replica. 

### How can we reproduce it (as minimally and precisely as possible)?

1. docker-compose up -d (from https://github.com/patroni/patroni/blob/master/docker-compose.yml#L7)
2. patronictl  edit-config , add: 
`synchronous_commit: remote_apply
synchronous_mode_strict: true 
synchronous_mode: true
synchronous_node_count: 2`

3. patronictl  list
+ Cluster: demo (7488968659735597079) -+-----------+----+-----------+
| Member   | Host       | Role         | State     | TL | Lag in MB |
+----------+------------+--------------+-----------+----+-----------+
| patroni1 | 172.18.0.8 | Sync Standby | streaming |  1 |         0 |
| patroni2 | 172.18.0.7 | Sync Standby | streaming |  1 |         0 |
| patroni3 | 172.18.0.6 | Leader       | running   |  1 |           |
+----------+------------+--------------+-----------+----+-----------+

4. docker stop demo-patroni1
5. docker stop demo-patroni2
6.  docker exec -ti demo-patroni3 bash
postgres@patroni3:~$ patronictl  list
+ Cluster: demo (7488968659735597079) -----+----+-----------+
| Member   | Host       | Role   | State   | TL | Lag in MB |
+----------+------------+--------+---------+----+-----------+
| patroni3 | 172.18.0.6 | Leader | running |  1 |           |
+----------+------------+--------+---------+----+-----------+

7. docker stop demo-patroni3
8. docker start demo-patroni1
9. docker start demo-patroni2
10. docker exec -ti demo-patroni1 bash
postgres@patroni1:~$ patronictl list
+ Cluster: demo (7488968659735597079) ------+----+-----------+
| Member   | Host       | Role    | State   | TL | Lag in MB |
+----------+------------+---------+---------+----+-----------+
| patroni1 | 172.18.0.6 | Replica | running |  1 |         0 |
| patroni2 | 172.18.0.7 | Replica | running |  1 |         0 |
+----------+------------+---------+---------+----+-----------+

11. Cluster fails to elect a new leader

### What did you expect to happen?

[...]

PR #3328, opened 2025-04-03
Maintain New Invariant For synchronous_mode_strict

Maithem:
This patch adds a new DCS update rule to maintain at least one synchronous replica in /sync. Prevent updates with an empty set by keeping the last known synchronous replica. This increases the chances of being able to elect a new leader. For example, if there are three nodes A, B and C (where A is the leader) then the following sequence prevents electing a new leader:

1. Shutdown B
2. Shutdown C
3. Shutdown A
4. Start B
5. Start C

If /sync is set '*' after step #2 then a new leader cannot be elected when B and C are up despite one of them being a synchronous replica. This patch skips updating the DCS such that sync_standby will still contain C, which will allow the cluster to recover correctly by promoting node C if A never joins back.

Addresses #3327

### DIFF
--- a/patroni/ha.py
+++ b/patroni/ha.py
@@ -961,11 +961,17 @@ class Ha(object):
         # update synchronous standby list in dcs temporarily to point to common nodes in current and picked
         sync_common = voters & allow_promote
         if sync_common != voters:
-            logger.info("Updating synchronous privilege temporarily from %s to %s",
-                        list(voters), list(sync_common))
-            sync = self.dcs.write_sync_state(self.state_handler.name, sync_common, 0, version=sync.version)
-            if not sync:
-                return logger.info('Synchronous replication key updated by someone else.')
+            if global_config.is_synchronous_mode_strict and not sync_common:
+                logger.info("Keeping existing sync replicas in /sync to maintain strict mode invariant: %s",
+                            list(voters))
+                if not picked:
+                    return
+            else:
+                logger.info("Updating synchronous privilege temporarily from %s to %s",
+                            list(voters), list(sync_common))
+                sync = self.dcs.write_sync_state(self.state_handler.name, sync_common, 0, version=sync.version)
+                if not sync:
+                    return logger.info('Synchronous replication key updated by someone else.')
 
         # When strict mode and no suitable replication connections put "*" to synchronous_standby_names
         if global_config.is_synchronous_mode_strict and not picked:
--- a/tests/test_ha.py
+++ b/tests/test_ha.py
@@ -1511,15 +1511,67 @@ class TestHa(PostgresInit):
         self.ha.run_cycle()
         self.assertEqual(self.ha.dcs.write_sync_state.call_count, 2)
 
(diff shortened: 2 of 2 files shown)


patroni/ha.py
```diff
@@ -914,6 +914,13 @@ def _process_multisync_replication(self) -> None:
 
         # update synchronous standby list in dcs temporarily to point to common nodes in current and picked
         sync_common = voters & allow_promote
+
+        # Since synchronous strict mode requires at least one other replica, /sync should not be updated
+        # with an empty set
+        if global_config.is_synchronous_mode_strict and len(voters) != 0 and not sync_common:
```
CyberDem0n, line 920: What about the following case?
We have a primary A and sync replica B.
At the same time when B is gone C is connected.
It will lead to the situation that `synchronous_standby_names` and `/sync` key will not be updated.
I don't think it is desired behavior.

Maithem: Hi @CyberDem0n, hit this issue again, so coming back to it.

Your "B gone while C connects" case is covered. When picked is non-empty (a streaming replacement exists), we skip the early return, and the final write_sync_state call in _process_multisync_replication sets both synchronous_standby_names and /sync to {C} in one cycle. The new if not picked: return fires only when picked is also empty. That's the original #3327 reproducer.

Added a unit test for that scenario: tests/test_ha.py::TestHa::test__process_multisync_replication, the "/sync is updated (not blocked) when a new replica C replaces departed replica B" sub-case. Plus a non-strict-mode regression test and a multi-voter narrowing test. All green.

#3435 is a broader strict-mode rewrite. Could we land this as the targeted fix and treat #3435 as the larger refactor on its own track? I can also backport this to the 4.x patch line if that helps.

CyberDem0n: @Maithem it would be very helpful if you spend some time testing #3435.
However, I didn't plan to backport it to older major/minor releases because it is not a simple bugfux, but a big architectural change.

Maithem: @CyberDem0n we're on 4.0.x today and planning to move to the 4.x release carrying #3435 once it lands — happy to test #3435 along the way. In parallel, can we push this PR through? Without an upstream backport we'd have to carry it as an internal fork, which we'd really like to avoid. I'll own the rebase onto `REL_4_0` (and `REL_4_1`) and keep them in sync as #3435 evolves.

Maithem: @CyberDem0n I need to know how to proceed, please make a call ^^

CyberDem0n: This PR solves the same problem as #3435, however this solution is incomplete and doesn't cover many edge cases. Why should it be pushed?

Maithem: - Closes the referenced bug (#3327) with a small diff behind a strict-mode only guard
- Backportable to REL_4_0
- Scope is synchronous_mode: true only; quorum is intentionally left to #3435

If there's something specific inside the synchronous_mode: true window you want covered before this lands, happy to extend, otherwise asking to merge as scoped.

hughcapet:
resolved via https://github.com/patroni/patroni/pull/3435

PR #3330, opened 2025-04-08
Enable PATRONI_POSTGRESQL_CONNECT_ADDRESS

Gradlon:
Override DOCKER_IP with PATRONI_POSTGRESQL_CONNECT_ADDRESS

I ran into an issue where when I used network_mode: host, I got the wrong IPS.
Implemented a small fix so the DOCKER_IP can be overridden a described in the docs


CyberDem0n (commented):
Have you seen https://github.com/patroni/patroni/blob/master/Dockerfile#L1-L2?

Gradlon:
Oh my bad!
Sorry for the trubble.

PR #3331, opened 2025-04-09
Add 'from-leader' parameter for patronictl reinit

hanguannan:
This merge has add a parameter for `patronictl reinit` command ,which to resolve the question described by: https://github.com/patroni/patroni/issues/3320. Thus we can add the '--from-leader' option when executing the `patronictl reinit` command to synchronize the wal log from the leader node of patroni.

### DIFF
--- a/patroni/api.py
+++ b/patroni/api.py
@@ -1002,6 +1002,7 @@ class RestApiHandler(BaseHTTPRequestHandler):
         The request body may contain a JSON dictionary with the following key:
 
             * ``force``: ``True`` if we want to cancel an already running task in order to reinit a replica.
+            * ``from_leader``: ``True`` if we want to reinit a replica and sync wal from the leader node.
 
         Response HTTP status codes:
 
@@ -1014,8 +1015,9 @@ class RestApiHandler(BaseHTTPRequestHandler):
             logger.debug('received reinitialize request: %s', request)
 
         force = isinstance(request, dict) and parse_bool(request.get('force')) or False
+        from_leader = isinstance(request, dict) and parse_bool(request.get('from-leader')) or False
 
-        data = self.server.patroni.ha.reinitialize(force)
+        data = self.server.patroni.ha.reinitialize(force, from_leader)
         if data is None:
             status_code = 200
             data = 'reinitialize started'
--- a/patroni/ctl.py
+++ b/patroni/ctl.py
@@ -1168,8 +1168,9 @@ def restart(cluster_name: str, group: Optional[int], member_names: List[str],
 @option_citus_group
 @click.argument('member_names', nargs=-1)
 @option_force
+@click.option('--from-leader', is_flag=True, help='Set to sync wal from leader')
 @click.option('--wait', help='Wait until reinitialization completes', is_flag=True)
-def reinit(cluster_name: str, group: Optional[int], member_names: List[str], force: bool, wait: bool) -> None:
+def reinit(cluster_name: str, group: Optional[int], member_names: List[str], force: bool, from_leader: bool, wait: bool) -> None:
     """Process ``reinit`` command of ``patronictl`` utility.
 
     Reinitialize cluster members based on given filters.
@@ -1181,6 +1182,7 @@ def reinit(cluster_name: str, group: Optional[int], member_names: List[str], for
     :param group: filter which Citus group we should reinit members. Refer to the module note for more details.
     :param member_names: name of the members that should be reinitialized.
     :param force: perform the restart without asking for confirmations.
+    :param from-leader: perform the reinit to sync wal from the leader node.
     :param wait: wait for the operation to complete.
     """
     cluster = get_dcs(cluster_name, group).get_cluster()
@@ -1188,7 +1190,7 @@ def reinit(cluster_name: str, group: Optional[int], member_names: List[str], for
 
     wait_on_members: List[Member] = []
     for member in members:
-        body: Dict[str, bool] = {'force': force}
+        body: Dict[str, bool] = {'force': force, 'from_leader': True }
         while True:
             r = request_patroni(member, 'post', 'reinitialize', body)
             started = check_response(r, member.name, 'reinitialize')
--- a/patroni/dcs/__init__.py
+++ b/patroni/dcs/__init__.py
@@ -937,16 +937,17 @@ class Cluster(NamedTuple('Cluster',
         return next((m for m in self.members if m.name == member_name),
                     self.leader if fallback_to_leader else None)
 
-    def get_clone_member(self, exclude_name: str) -> Union[Member, Leader, None]:
+    def get_clone_member(self, exclude_name: str, from_leader: bool = False) -> Union[Member, Leader, None]:
         """Get member or leader object to use as clone source.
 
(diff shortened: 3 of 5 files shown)


patroni/ctl.py
```diff
@@ -1168,8 +1168,9 @@ def restart(cluster_name: str, group: Optional[int], member_names: List[str],
 @option_citus_group
 @click.argument('member_names', nargs=-1)
 @option_force
+@click.option('--from-leader', is_flag=True, help='Set to sync wal from leader')
```
CyberDem0n, line 1171: I think the `help` section is wrong.
It is not about syncing WAL, but about getting basebackup.

patroni/ctl.py
```diff
@@ -1181,14 +1182,15 @@ def reinit(cluster_name: str, group: Optional[int], member_names: List[str], for
     :param group: filter which Citus group we should reinit members. Refer to the module note for more details.
     :param member_names: name of the members that should be reinitialized.
     :param force: perform the restart without asking for confirmations.
+    :param from-leader: perform the reinit to sync wal from the leader node.
```
CyberDem0n, line 1185: Function param name is `from_leader`, not `from-leader`, and the comment is also wrong.

patroni/ctl.py
```diff
@@ -1181,14 +1182,15 @@ def reinit(cluster_name: str, group: Optional[int], member_names: List[str], for
...
     :param wait: wait for the operation to complete.
     """
     cluster = get_dcs(cluster_name, group).get_cluster()
     members = get_members(cluster, cluster_name, member_names, 'replica', force, 'reinitialize', group=group)
 
     wait_on_members: List[Member] = []
     for member in members:
-        body: Dict[str, bool] = {'force': force}
+        body: Dict[str, bool] = {'force': force, 'from_leader': True }
```
CyberDem0n, line 1193: Hmm, why is it always set to true?

patroni/dcs/__init__.py
```diff
@@ -937,16 +937,17 @@ def get_member(self, member_name: str, fallback_to_leader: bool = True) -> Union
         return next((m for m in self.members if m.name == member_name),
                     self.leader if fallback_to_leader else None)
 
-    def get_clone_member(self, exclude_name: str) -> Union[Member, Leader, None]:
+    def get_clone_member(self, exclude_name: str, from_leader: bool = False) -> Union[Member, Leader, None]:
```
CyberDem0n, line 940: Why do you even change this function?
It is enough to change only `Ha._do_reinitialize()`.

patroni/api.py
```diff
@@ -1014,8 +1015,9 @@ def do_POST_reinitialize(self) -> None:
             logger.debug('received reinitialize request: %s', request)
 
         force = isinstance(request, dict) and parse_bool(request.get('force')) or False
+        from_leader = isinstance(request, dict) and parse_bool(request.get('from-leader')) or False
```
CyberDem0n, line 1018: It is better to use snake case rather then kebab case.
```suggestion
        from_leader = isinstance(request, dict) and parse_bool(request.get('from_leader')) or False
```

CyberDem0n (changes requested):
I left a bunch of comments that should be addressed.
Besides that, you don't take into account that `basebackup` [create replica method](https://patroni.readthedocs.io/en/latest/replica_bootstrap.html#building-replicas) could be not the first one in the [list](https://github.com/patroni/patroni/blob/1ba9e68b4b0812c4221c251f468340e4a6a69574/patroni/postgresql/bootstrap.py#L253-L270), and as a result of that it will use something else instead of pg_basebackup from the leader.

patroni/ctl.py
hanguannan: oh, it's my mistake.

patroni/dcs/__init__.py
hanguannan: yes, it's possible to solve this without modifying get_clone_member().  Thanks.

patroni/ctl.py
```diff
@@ -1168,8 +1168,9 @@ def restart(cluster_name: str, group: Optional[int], member_names: List[str],
 @option_citus_group
 @click.argument('member_names', nargs=-1)
 @option_force
+@click.option('--', is_flag=True, help='Set to get basebackup from leader')
```
CyberDem0n, line 1171:
```suggestion
@click.option('--from-leader', is_flag=True, help='Get basebackup from leader')
```

CyberDem0n (changes requested):
You still don't take into account that `basebackup` [create replica method](https://patroni.readthedocs.io/en/latest/replica_bootstrap.html#building-replicas) could be not the first one in the [list](https://github.com/patroni/patroni/blob/1ba9e68b4b0812c4221c251f468340e4a6a69574/patroni/postgresql/bootstrap.py#L253-L270), and as a result of that it will use something else instead of pg_basebackup from the leader.

patroni/ctl.py
hanguannan: Thank you for your guidance. This time, I considered and implemented the changes that when the reinit command uses '--from-leader' parameter, the replica will only use  basebackup method to synchronize data from the leader.

patroni/ha.py
```diff
@@ -220,7 +220,9 @@ def set_is_active(self, value: float) -> None:
 
 
 class Ha(object):
-
+    # Marking for '--from-leader' parameter for patronictl reinit command
+    global_is_from_leader = False
```
CyberDem0n, line 224: No way this will be accepted.
Please introduce new arguments to Bootstrap.clone() and Bootstrap.create_replica() methods.

hanguannan: Thanks. I have completed changes.

patroni/ha.py
```diff
@@ -220,7 +220,7 @@ def set_is_active(self, value: float) -> None:
 
 
 class Ha(object):
-
+
```
hughcapet, line 223: unwanted changes

hughcapet (commented):
Please also extend the documentation

hanguannan:
> Please also extend the documentation



> and please run tests locally before pushing https://patroni.readthedocs.io/en/latest/contributing_guidelines.html#running-tests

I added docs and run tests locally,hope be ok this time.

patroni/postgresql/bootstrap.py
```diff
@@ -249,9 +250,11 @@ def create_replica(self, clone_member: Union[Leader, Member, None]) -> Optional[
...
         # get list of replica methods either from clone member or from
         # the config. If there is no configuration key, or no value is
-        # specified, use basebackup
-        replica_methods = (clone_member.create_replica_methods if is_remote_member
-                           else self._postgresql.create_replica_methods) or ['basebackup']
+        # specified, use basebackup. If '--from-leader' parameter is set
+        # when reinit, always use basebackup.
+        replica_methods = ['basebackup'] if clone_from_leader else (
+            clone_member.create_replica_methods if is_remote_member
+            else self._postgresql.create_replica_methods) or ['basebackup']
```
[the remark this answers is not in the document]
hanguannan: Completed modification and tested ok.

hughcapet:
👍

PR #3339, opened 2025-04-23
Supplementary Explanation of Configurable Parameters for synchronous_…

feeling0713:
Supplement the docs/dynamic_configuration.rst regarding the configuration values for the synchronous_mode parameter. 
In fact, synchronous_mode can be set to off or false, on or true, and quorum in dcs.


hughcapet:
these are not all the acceptable values if we want to go this way:

https://github.com/patroni/patroni/blob/3e783606a0e4b91760926c512d61be7b3ee263ae/patroni/global_config.py#L113-L117

https://github.com/patroni/patroni/blob/3e783606a0e4b91760926c512d61be7b3ee263ae/patroni/global_config.py#L93-L101

https://github.com/patroni/patroni/blob/3e783606a0e4b91760926c512d61be7b3ee263ae/patroni/utils.py#L159-L186

PR #3340, opened 2025-04-24
Supplement the configuration instructions for ttl and loop_wait param…

feeling0713:
If loop_wait is greater than ttl/2, the replica member information will disappear in the DCS for a period of time and then be added back to the DCS, which may cause some confusion when querying the cluster status.

### DIFF
--- a/docs/dynamic_configuration.rst
+++ b/docs/dynamic_configuration.rst
@@ -19,6 +19,11 @@ In order to change the dynamic configuration you can use either :ref:`patronictl
 
         loop_wait + 2 * retry_timeout <= ttl
 
+    and you'd better follow the rule:
+
+    .. code-block:: python
+
+        2 * loop_wait <= ttl
 
 -  **maximum\_lag\_on\_failover**: the maximum bytes a follower may lag to be able to participate in leader election.
 -  **maximum\_lag\_on\_syncnode**: the maximum bytes a synchronous follower may lag before it is considered as an unhealthy candidate and swapped by healthy asynchronous follower. Patroni utilize the max replica lsn if there is more than one follower, otherwise it will use leader's current wal lsn. Default is -1, Patroni will not take action to swap synchronous unhealthy follower when the value is set to 0 or below. Please set the value high enough so Patroni won't swap synchrounous follower frequently during high transaction volume.


docs/dynamic_configuration.rst
```diff
@@ -19,6 +19,11 @@ In order to change the dynamic configuration you can use either :ref:`patronictl
 
         loop_wait + 2 * retry_timeout <= ttl
 
+    and you'd better follow the rule:
+
+    .. code-block:: python
+
+        2 * loop_wait <= ttl
```
CyberDem0n, line 26: There is no value in it, the existing requirement is already stronger then this one.
Besides that, Patroni 3.2 already enforcing the rule.

feeling0713: loop_wait: 12
retry_timeout: 3
ttl: 20

I have set the parameter values as above, and when I execute patronictl list, I notice that the cluster information is changing.

1）At a certain moment, only leader information is available.
| Member | Host                 | Role   | State   | TL | Lag in MB |
+--------+----------------------+--------+---------+----+-----------+
| see_01 | 192.168.173.108:5866 | Leader | running |  4 |           |
+--------+----------------------+--------+---------+----+-----------+

2）At a certain moment, it is queried that the standby node is asynchronous.
| Member | Host                 | Role    | State     | TL | Lag in MB |
+--------+----------------------+---------+-----------+----+-----------+
| see_01 | 192.168.173.108:5866 | Leader  | running   |  4 |           |
| see_02 | 192.168.173.109:5866 | Replica | streaming |  4 |         0 |
| see_03 | 192.168.173.107:5866 | Replica | streaming |  4 |         0 |
+--------+----------------------+---------+-----------+----+-----------+

3）At a certain moment, it is queried that the standby node is synchronous.
| Member | Host                 | Role           | State     | TL | Lag in MB |
+--------+----------------------+----------------+-----------+----+-----------+
| see_01 | 192.168.173.108:5866 | Leader         | running   |  4 |           |
| see_02 | 192.168.173.109:5866 | Quorum Standby | streaming |  4 |         0 |
| see_03 | 192.168.173.107:5866 | Quorum Standby | streaming |  4 |         0 |
+--------+----------------------+----------------+-----------+----+-----------+

feeling0713: The leader is normal.
I think that since the replica updates in the DCS through 'touch_member()' every loop_wait interval, but is released upon ttl expiration. if the ttl (e.g., 20 seconds) is just renewed, during the next loop_wait (12 seconds) the member will not be updated due to the '_do_refresh_lease()'. Consequently, before the arrival of the subsequent loop_wait (24 seconds), the TTL will expire, resulting in a period of  2*loop_wait - ttl  where no replica information exists in the DCS.

CyberDem0n: @feeling0713 it would be cool if you test it

feeling0713: Yes, that's effective. 
And by observing the leases using `etcdctl lease list` , it's indeed being refreshed continuously.

PR #3361, opened 2025-05-20
Clean cluster demotion

hughcapet:


patroni/ha.py
```diff
@@ -1593,19 +1599,26 @@ def before_shutdown() -> None:
...
+            self._rewind.archive_shutdown_checkpoint_wal()
         if mode_control['offline']:
             node_to_follow, leader = None, None
         else:
+            if mode_control['release']:
+                if not status['released']:
+                    checkpoint_lsn = prev_location if mode == 'graceful' else None
+                    with self._async_executor:
+                        self.release_leader_key_voluntarily(checkpoint_lsn)
+                time.sleep(2)  # Give a time to somebody to take the leader lock
```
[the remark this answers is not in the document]
hughcapet: hm, this way shutdown checkpoint is only archived when on_shutdown callback is not called

patroni/postgresql/__init__.py
```diff
@@ -1414,3 +1404,13 @@ def _get_gucs(self) -> CaseInsensitiveSet:
...
+
+    def get_archive_command(self) -> Optional[str]:
+        """Get ``archive_command`` GUC value if defined and archiving is enabled.
+
+        :returns: ``archive_command`` defined in the Postgres configuration or None.
+        """
+        archive_mode = self.get_guc_value('archive_mode')
+        archive_cmd = self.get_guc_value('archive_command')
+        if archive_mode in ('on', 'always') and archive_cmd:
+            return archive_cmd
```
CyberDem0n, line 1416: Hmm, it is only used in a context of rewind.py, shouldn't it better be there?

patroni/ha.py
```diff
@@ -1571,6 +1575,17 @@ def demote(self, mode: str) -> Optional[bool]:
...
         status = {'released': False}
 
+        demote_cluster_with_archive = False
+        archive_cmd = self.state_handler.get_archive_command()
+        if mode == 'demote-cluster':
+            # We need to send the shutdown checkpoint WAL file to archive to eliminate the need of rewind
+            # from a promoted instance that was previously replicating from archive
+            # When doing this, we disable stop timeout and do not run on_shutdown callback
+            if archive_cmd is None:
+                logger.info('Not archiving latest checkpoint WAL file. Archiving is not configured.')
```
CyberDem0n, line 1585: I think this message shows up in logs to early.
Better to do it after shutdown.

patroni/ha.py
```diff
@@ -1591,18 +1608,27 @@ def before_shutdown() -> None:
...
-            time.sleep(2)  # Give a time to somebody to take the leader lock
+                    if mode == 'demote-cluster':
+                        # for cluster demotion we need shutdown checkpoint lsn to be written to optime, not the prev one
+                        self.dcs.update_leader(self.cluster, checkpoint_lsn, None, self._failsafe_config())
+                    else:
+                        self.release_leader_key_voluntarily(prev_lsn)
+            if demote_cluster_with_archive:
+                self._rewind.archive_shutdown_checkpoint_wal(cast(str, archive_cmd))
+            else:
+                time.sleep(2)  # Give a time to somebody to take the leader lock
```
CyberDem0n, line 1631: Something feels wrong here.
Do we really need to sleep if we demote to standby cluster, when there is no archive_command and leader key is not released?

PR #3365, opened 2025-05-29
Add leader annotation to prevent endpoints being cleaned

waynerv:
Endpoints-controller in kube-controller-manager[ checks for endpoints that are not associated with any service during program startup and deletes them](https://github.com/kubernetes/kubernetes/blob/b35c5c0a301d326fdfa353943fca077778544ac6/pkg/controller/endpoint/endpoints_controller.go#L557). 

We can add a specific annotation `control-plane.alpha.kubernetes.io/leader='{}'` to the config/sync/failover endpoints to totally prevent this kind of cleanup. Then there is no need to creating a config service.

### DIFF
--- a/patroni/dcs/kubernetes.py
+++ b/patroni/dcs/kubernetes.py
@@ -39,6 +39,7 @@ SERVICE_HOST_ENV_NAME = 'KUBERNETES_SERVICE_HOST'
 SERVICE_PORT_ENV_NAME = 'KUBERNETES_SERVICE_PORT'
 SERVICE_TOKEN_FILENAME = '/var/run/secrets/kubernetes.io/serviceaccount/token'
 SERVICE_CERT_FILENAME = '/var/run/secrets/kubernetes.io/serviceaccount/ca.crt'
+LEADER_ELECTION_RECORD_ANNOTATION_KEY = 'control-plane.alpha.kubernetes.io/leader'
 __temp_files: List[str] = []
 
 
@@ -1104,6 +1105,8 @@ class Kubernetes(AbstractDCS):
 
         :returns: the new :class:`V1Endpoints` or :class:`V1ConfigMap` object, that was created or updated.
         """
+        use_endpoints = self._api.use_endpoints
+
         metadata = {'namespace': self._namespace, 'name': name, 'labels': self._labels, 'annotations': annotations}
         if patch or resource_version:
             if resource_version is not None:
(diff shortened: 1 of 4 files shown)


patroni/dcs/kubernetes.py
```diff
@@ -816,6 +816,10 @@ def client_path(self, path: str) -> str:
     def leader_path(self) -> str:
         return super(Kubernetes, self).leader_path[:-7 if self._api.use_endpoints else None]
 
+    @property
+    def should_create_leader_annotation_paths(self) -> Set[str]:
+        return {self.config_path, self.sync_path, self.failover_path}
+
```
CyberDem0n, line 822: I don't think we need it, because effectively we want to protect all endpoints created by Patroni.

CyberDem0n (commented):
@waynerv may I ask you to split it to two commits?
The first one should carefully revert https://github.com/patroni/patroni/pull/958, and the second just add annotation.

patroni/dcs/kubernetes.py
waynerv: Since endpoints were planned to be deprecated a long time ago, the reason why current endpoints can still work is that there is an endpointslicemirroring_controller mirroring a truly working endpointslice for it.

If we add this annotation to leader endpoints, then endpointslicemirroring_controller will skip it thus it won't accept traffic anymore, users have to define its own ingress service with label selector like role=primary. this seems breaked current behavior.

patroni/dcs/kubernetes.py
```diff
@@ -1115,9 +1117,12 @@ def _patch_or_create(self, name: str, annotations: Dict[str, Any],
             func = functools.partial(self._api.create_namespaced_kind)
             # skip annotations with null values
             metadata['annotations'] = {k: v for k, v in annotations.items() if v is not None}
+        if use_endpoints and name != self.leader_path:
+            # prevent endpoints without traffic from being cleaned by K8s
+            metadata['annotations'][LEADER_ELECTION_RECORD_ANNOTATION_KEY] = '{}'
```
CyberDem0n, line 1122: What does `'{}'` mean?
Should we better write something more meaningful there?
For example, `'Patroni'`?

Besides that, it looks good.

waynerv: This is just a placeholder, as the value of this annotation is not important in our case. The [typical usage](https://kubernetes.io/docs/reference/labels-annotations-taints/#control-plane-alpha-kubernetes-io-leader) of the leader key is like the following:
`control-plane.alpha.kubernetes.io/leader={"holderIdentity":"controller-0","leaseDurationSeconds":15,"acquireTime":"2023-01-19T13:12:57Z","renewTime":"2023-01-19T13:13:54Z","leaderTransitions":1}`

I chose to use '{}' to maintain this convention but indicate an empty value

waynerv: After careful consideration, I have changed my mind. If we are not using leader annotations for leader election, then there is no need to maintain the convention of using a specific data structure. An empty '{}' may cause unnecessary confusion for experienced k8s users. 

Setting the value to Patroni would be more meaningful and simple.

waynerv:
@CyberDem0n Just wanted to gently follow up on this PR. I've addressed the requested changes.
Please let me know if there's anything else I can do,thanks for your time!

hughcapet:
I personally do not like this hacky solution relying on something not intended for custom use. Moreover, it is currently [in deprecated state](https://kubernetes.io/docs/reference/labels-annotations-taints/?utm_source=chatgpt.com#control-plane-alpha-kubernetes-io-leader).
So if we still want to make use of it, I would better make this configurable and keep old behaviour as the default one.

hughcapet:
closing, as endpoint API is currently deprecated in k8s

PR #3380, opened 2025-06-10
Fix potential issue around resolving localhost to IPv6

avandras:
When specifying `0.0.0.0` or `127.0.0.1` in `listen_addresses`, Postgres won't listen on IPv6 (in other words, these addresses imply IPv4 only).  However, the typical Windows system will have a preference to resolve `localhost` to `::1`.

While one can change these preferences, it's unclear if it has unwanted side effects.  Instead of this, the proposed small change fixes this behaviour.

### DIFF
--- a/patroni/postgresql/config.py
+++ b/patroni/postgresql/config.py
@@ -1102,15 +1102,11 @@ class ConfigHandler(object):
         listen_addresses = self._server_parameters['listen_addresses'].split(',')
 
         for la in listen_addresses:
-            if os.name == 'nt':
-                if la.strip().lower() in ('*', 'localhost'):  # we are listening on '*' or localhost
-                    return 'localhost'  # connection via localhost is preferred
-                if la.strip().lower() in ('0.0.0.0', '127.0.0.1'):  # Postgres listens only on IPv4
-                    return '127.0.0.1'  # localhost, but don't allow Windows to resolve to IPv6
-            else:
-                # we are listening on '*' or localhost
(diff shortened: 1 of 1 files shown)


patroni/postgresql/config.py
```diff
@@ -1102,8 +1102,15 @@ def _get_tcp_local_address(self) -> str:
...
-                return 'localhost'  # connection via localhost is preferred
+            if os.name == 'nt':
+                if la.strip().lower() in ('*', 'localhost'):  # we are listening on '*' or localhost
+                    return 'localhost'  # connection via localhost is preferred
+                if la.strip().lower() in ('0.0.0.0', '127.0.0.1'):  # Postgres listens only on IPv4
+                    return '127.0.0.1'  # localhost, but don't allow Windows to resolve to IPv6
+            else:
+                # we are listening on '*' or localhost
+                if la.strip().lower() in ('*', '0.0.0.0', '127.0.0.1', 'localhost'):
+                    return 'localhost'  # connection via localhost is preferred
```
CyberDem0n, line 1113: Lets rewrite it as following to avoid code duplication:
```suggestion
            if la.strip().lower() in ('*', 'localhost'):  # we are listening on '*' or localhost
                return 'localhost'  # connection via localhost is preferred
            if la.strip().lower() in ('0.0.0.0', '127.0.0.1'):  # Postgres listens only on IPv4
                # localhost, but don't allow Windows to resolve to IPv6
                return '127.0.0.1' if os.name == 'nt' else 'localhost'
```

CyberDem0n (approved):

Issue #3379, opened 2025-06-09
Patroni exposes underlying python runtime version in Server header


dgrierso:
### What happened?

Running patroni 4.0.4 accessing the port 8008 API port, the `Server` header exposes information about the python runtime which could potentially be used as "useful" information for attack.

Patroni should not surface this information under normal execution.

### How can we reproduce it (as minimally and precisely as possible)?

1. Run patroni.
2. Access port 8008 via HTTP - e.g. with curl:
```
$ curl -i http://10.105.158.56:8008/
HTTP/1.0 200 OK
Server: BaseHTTP/0.6 Python/3.12.3
Date: Mon, 09 Jun 2025 10:03:26 GMT
Content-Type: application/json
```

### What did you expect to happen?

`Server:` header should not expose underlying software configuration:
```
Server: Patroni
```

### Patroni/PostgreSQL/DCS version

-   Patroni version: 4.0.4
-   PostgreSQL version: 16.9
-   DCS (and its version): etcd 3.5.21

### Patroni configuration file

```yaml
N/A
```

### patronictl show-config

```yaml
N/A
```

### Patroni log files

```shell
N/A
```

### PostgreSQL log files

```shell
N/A
```

### Have you tried to use GitHub issue search?

- [x] Yes

CyberDem0n:
Well. If someone isn't keeping their software up-to-date, potentially they have much bigger problem than "exposed" version of python.
For attackers it is enough to know that there is a python in order to try out different exploits.

TonyNoble:
@CyberDem0n - that doesn't mean you help them by giving them exact details of where to look.  And why would they know if's python if the server banner doesn't tell them? 

Server banner disclosure is considered a vulnerability by most security scanners and it's a simple fix to prevent in this instance.

CyberDem0n:
@TonyNoble do you want to work on it?
That is, implement a new config parameter that configures behavior of Server header.

TonyNoble:
@CyberDem0n , happy to do so.  I'll raise a PR in the next day or two.

PR #3384, opened 2025-06-18
Fix for #3379 - adding configuration for HTTP Server header

dgrierso:
### Adding support for configuring the HTTP `Server` header

Update to `RestApiHandler` class which adds a `version_string()` method to override the `BaseHTTPRequestHandler` base class' method to return the value of the `server.server_header` property. This is used to populate the `Server` HTTP header.

Update the `RestApiServer` class to add a method `__construct_server_tokens()` using the `restapi.server_tokens` configuration parameter. Valid values of the `restapi.server_tokens` parameter are:
* "Original" : will execute the original behaviour and print the "BaseHTTP/$version Python/$version" string
* "ProductOnly" : prints merely "Patroni"
* "Minimal" : prints "Patroni/$version"

An invalid or no value will result in the original behaviour being applied.

Addresses issue #3379

### DIFF
--- a/docs/ENVIRONMENT.rst
+++ b/docs/ENVIRONMENT.rst
@@ -208,6 +208,7 @@ REST API
 -  **PATRONI\_RESTAPI\_HTTP\_EXTRA\_HEADERS**: (optional) HTTP headers let the REST API server pass additional information with an HTTP response.
 -  **PATRONI\_RESTAPI\_HTTPS\_EXTRA\_HEADERS**: (optional) HTTPS headers let the REST API server pass additional information with an HTTP response when TLS is enabled. This will also pass additional information set in ``http_extra_headers``.
 -  **PATRONI\_RESTAPI\_REQUEST\_QUEUE\_SIZE**: (optional): Sets request queue size for TCP socket used by Patroni REST API.  Once the queue is full, further requests get a "Connection denied" error. The default value is 5.
+-  **PATRONI\_RESTAPI\_SERVER\_TOKENS**: (optional) Configures the value of the ``Server`` HTTP header.  ``Original`` (default) will expose the original behaviour and display the BaseHTTP and Python versions, e.g. ``BaseHTTP/0.6 Python/3.12.3``. ``Minimal``: The header will contain only the Patroni version, e.g. ``Patroni/4.0.0``. ``ProductOnly``: The header will contain only the product name, e.g. ``Patroni``.
 
 .. warning::
 


patroni/api.py
```diff
@@ -1480,6 +1494,38 @@ def __init__(self, patroni: Patroni, config: Dict[str, Any]) -> None:
...
+        logger.debug('restapi.server_tokens is set to "%s".', token_config)
+        if token not in ('original', 'full', 'productonly', 'minimal'):
+            logger.warning('restapi.server_tokens is set to "%s". Patroni will not modify the Server header. '
+                           'Valid values are: "Full", "Minimal", "ProductOnly".', token_config)
+            return ""
+        
+        # If 'original' is set, we do not modify the Server header.
+        # This is useful for compatibility with existing setups that expect the original header.
+        if token == 'original':
+            logger.info('restapi.server_tokens is set to "Original". Patroni will not modify the Server header.')
```
CyberDem0n, line 1515: It will log it every time when reload with default config happens.

CyberDem0n (commented):
You also need to:
1. Add unit tests.
2. Format code so that it passes flake8 check
3. Modify patroni/validator.py, so that config could be properly validated.
4. Modify paroni/config.py, to add support of `PATRONI_RESTAPI_SERVER_TOKENS`
5. Update documentation

For 1 and 2 you can check [docs](https://patroni.readthedocs.io/en/latest/contributing_guidelines.html#running-tests).

patroni/api.py
```diff
@@ -1480,6 +1494,38 @@ def __init__(self, patroni: Patroni, config: Dict[str, Any]) -> None:
...
+            return ""
+        
+        # If 'full', 'productonly', or 'minimal' is set, we construct the header accordingly.
+        if token == 'productonly': # Show only the product name, without versions.
+            return 'Patroni'
+        elif token == 'minimal': # Show only the product name and version, without PostgreSQL version.
+            return f'Patroni/{self.patroni.version}'
+        elif token == 'full': # Show everything, including Patroni and PostgreSQL versions.
+            return f'Patroni/{self.patroni.version} (PostgreSQL {self.patroni.postgresql.major_version})'
```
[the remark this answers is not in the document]
dgrierso: I was using the manner in which the Apache httpd strings together for a "Full" parameter to do this in a similar way - i.e.:

`Server: Apache/2.4.2 (Unix) PHP/4.2.2 MyMod/1.2`

You could argue that you don't need to see the "PHP/4.2.2 MyMod/1.2" part too ... but if you're specifying "Full" then you want to get all the details you can.

Normal operation would be "ProductOnly" or "Minimal" but if you're wanting to debug you might also want to expose the PostgreSQL details.

patroni/api.py
dgrierso: Good point - will remove the logging.

patroni/api.py
```diff
@@ -119,6 +119,16 @@ def __init__(self, request: Any,
         self.__start_time: float = 0.0
         self.path_query: Dict[str, List[str]] = {}
 
+    def version_string(self) -> str:
+        """Override the default version string to return the server header as specified in the configuration.
+
+        If the server header is not set, then it returns the default version string of the HTTP server.
+
+        :return: ``Server`` version string, which is either the server header or the default version string
+        from the BaseHTTPRequestHandler.
```
CyberDem0n, line 128: It fails: https://github.com/patroni/patroni/actions/runs/15759073435/job/44424269301?pr=3384#step:6:1843

tests/test_api.py
```diff
@@ -811,3 +811,18 @@ def test_query(self):
                 patch.object(MockConnection, 'query') as mock_query:
             self.srv.query('SELECT 1')
             mock_query.assert_called_once_with('SELECT 1')
+
+    def test__construct_server_tokens(self):
+        #
+        # Test cases (case insensitive values):
```
hughcapet, line 817:
```suggestion
```
you need to test not-all-lower-case also works. change at least one value in these use cases

patroni/api.py
```diff
@@ -119,6 +119,16 @@ def __init__(self, request: Any,
         self.__start_time: float = 0.0
         self.path_query: Dict[str, List[str]] = {}
 
+    def version_string(self) -> str:
+        """Override the default version string to return the server header as specified in the configuration.
+
+        If the server header is not set, then it returns the default version string of the HTTP server.
+
+        :return: ``Server`` version string, which is either the server header or the default version string
+            from the BaseHTTPRequestHandler.
```
hughcapet, line 128:
```suggestion
            from the :class:`BaseHTTPRequestHandler`.
```

docs/yaml_configuration.rst
```diff
@@ -349,6 +349,11 @@ REST API
    -  **http\_extra\_headers**: (optional): HTTP headers let the REST API server pass additional information with an HTTP response.
    -  **https\_extra\_headers**: (optional): HTTPS headers let the REST API server pass additional information with an HTTP response when TLS is enabled. This will also pass additional information set in ``http_extra_headers``.
    -  **request_queue_size**: (optional): Sets request queue size for TCP socket used by Patroni REST API.  Once the queue is full, further requests get a "Connection denied" error. The default value is 5.
+   -  **server_tokens**: (optional): Configures the value of the ``Server`` HTTP header.
+      - ``Full``: The header will contain the Patroni version and the PostgreSQL version, e.g. ``Patroni/4.0.0 (PostgreSQL 140000)``.
```
hughcapet, line 353: "full" is not fully removed

tests/test_api.py
dgrierso: Fair point - added tests for mixed cases of each valid value.

patroni/api.py
dgrierso: Updated.

tests/test_api.py
```diff
@@ -811,3 +811,21 @@ def test_query(self):
                 patch.object(MockConnection, 'query') as mock_query:
             self.srv.query('SELECT 1')
             mock_query.assert_called_once_with('SELECT 1')
+
+    def test__construct_server_tokens(self):
```
CyberDem0n, line 815:
```suggestion
    def test_construct_server_tokens(self):
```

docs/yaml_configuration.rst
```diff
@@ -349,6 +349,10 @@ REST API
    -  **http\_extra\_headers**: (optional): HTTP headers let the REST API server pass additional information with an HTTP response.
    -  **https\_extra\_headers**: (optional): HTTPS headers let the REST API server pass additional information with an HTTP response when TLS is enabled. This will also pass additional information set in ``http_extra_headers``.
    -  **request_queue_size**: (optional): Sets request queue size for TCP socket used by Patroni REST API.  Once the queue is full, further requests get a "Connection denied" error. The default value is 5.
+   -  **server_tokens**: (optional): Configures the value of the ``Server`` HTTP header.
```
CyberDem0n, line 352: You also need to update https://github.com/patroni/patroni/blob/master/docs/ENVIRONMENT.rst#rest-api

hughcapet (approved):

PR #3388, opened 2025-06-25
Skip advancing externally created physical slots without `restart_lsn`

barthisrael:
When a physical replication slot is defined in the DCS but created manually outside of Patroni, it may lack a reserved restart LSN. This causes `_ensure_physical_slots()` to fail during slot advancement due to unexpected `None` values.

The issue can be reproduced like this:

1. Start the primary
2. Start the standby
3. Create a physical replication slot in the standby without reserving the LSN
4. Add that slot to the permanent slots with patronictl edit-config

As soon as the standby attempts to synchronize the LSN, it logs a stack trace like this:

```
ERROR: Exception when changing replication slots
  File "/usr/lib/python3.9/site-packages/patroni/postgresql/slots.py", line 551, in sync_replication_slots
    self._drop_incorrect_slots(cluster, slots)
  File "/usr/lib/python3.9/site-packages/patroni/postgresql/slots.py", line 416, in _ensure_physical_slots
    value.get('expected_active') is not True and not value['xmin']:
TypeError: '>' not supported between instances of 'int' and 'NoneType'
```

Such failures can prevent all slot synchronization, including for healthy slots, affecting replication reliability.

This change improves resilience by skipping unmanaged slots without a `restart_lsn` and logging a clear warning to guide the user toward corrective action (in general, the user can fix the cluster by dropping the slot, and by letting Patroni recreate and manage it).

References: PAT-273.


patroni/postgresql/slots.py
```diff
@@ -417,8 +417,20 @@ def _ensure_physical_slots(self, slots: Dict[str, Any], clean_inactive_physical_
...
             elif self._postgresql.can_advance_slots and self._replication_slots[name]['type'] == 'physical' and\
                     value.get('expected_active') is not True and not value['xmin']:
+                restart_lsn = value.get('restart_lsn')
+                if not restart_lsn:
+                    # This slot either belongs to a member or was configured as a permanent slot. However, for some
+                    # reason the slot was created by an external agent instead of by Patroni, and it was created without
+                    # reserving the LSN. We ignore the slot here, as we cannot advance it. The user might want to drop
+                    # the slot and let Patroni recreate and manage it.
+                    logger.warning(
+                        "Physical replication slot '%s' has no restart_lsn, cannot advance it. "
```
hughcapet, line 427: Shouldn't we give users more context, what they should do?
Maybe take a step forward and delete this slot right away?

barthisrael: Maybe we can add information that they might want to drop the slots?

I was tempted to go ahead and remove, but I wonder if we should always take that action, or leave the option to the user to check if they did some mistake?

I guess the latter would be a bit more conservative 😆

hughcapet: Lgtm. But what would be your arguments against deletion?

barthisrael: I've been thinking about it, and I guess we can drop the slot on behalf of the user.

I was only afraid of wrongly taking some destructive action. But in theory, if this code is reached, that's either a permanent or a member physical slot, which are expected to be managed by Patroni. So I think we are safe to drop it through Patroni.

I can push a commit with that change later if you are also inclined to that approach :)

hughcapet: I would vote for it:) @CyberDem0n 

Btw, how did that happen in your case? Users thought they need to do it for Patroni or this slot was used for some other use case?

patroni/postgresql/slots.py
```diff
@@ -417,8 +417,24 @@ def _ensure_physical_slots(self, slots: Dict[str, Any], clean_inactive_physical_
...
+                    logger.warning('Dropping physical replication slot %s because it has no restart_lsn. '
+                                   'This slot was probably not created by Patroni, but by an external agent.',
+                                   name)
+                    _, dropped = self.drop_replication_slot(name)
+                    if dropped:
+                        self._replication_slots.pop(name)
+                    else:
+                        self._schedule_load_slots = True
+                        logger.error("Failed to drop replication slot '%s'", name)
+                    continue
```
CyberDem0n, line 435: These lines are not covered by tests:
```python
tests/test_slots.py ..
>>>>>>>>>>>>>>>>>>>>>>> PDB set_trace (IO-capturing turned off) >>>>>>>>>>>>>>>>>>>>>>>
> /home/akukushkin/git/patroni/patroni/postgresql/slots.py(431)_ensure_physical_slots()
-> _, dropped = self.drop_replication_slot(name)
(Pdb) n
ValueError: not enough values to unpack (expected 2, got 0)
> /home/akukushkin/git/patroni/patroni/postgresql/slots.py(431)_ensure_physical_slots()
-> _, dropped = self.drop_replication_slot(name)
```

Actually, this block is almost the same as https://github.com/patroni/patroni/blob/bc0c4234fedb5f4268b1f089208286b971e88ad2/patroni/postgresql/slots.py#L397-L405
I think we need to move it to a dedicated method `_drop_physical_slot()`.

barthisrael: I just sent a couple of commits:

* 1611d1d71f4b02e4f4e7fcea5f42337fe5e96516: abstracting the common logic in a new helper method `_drop_physical_slot`
* 0beeb12cce9f8eec3a5d677cf247d12f5f45a452: to modify the unit test suite. I left a note in that commit message, about a couple of patches to `drop_replication_slot` which I didn't modify yet. It seems to me those unit tests "will have no value" if I replace the mock, so we might want either to remove them or to keep as it is.

patroni/postgresql/slots.py
```diff
@@ -332,6 +332,26 @@ def drop_replication_slot(self, name: str) -> Tuple[bool, bool]:
...
+
+        .. note::
+            If not able to drop the slot, it will log a message and set the flag to reload slots.
+
+        :param name: name of the slot to be dropped.
+        :returns: ``True`` if the slot was successfully dropped, ``False`` otherwise.
+        """
+        active, dropped = self.drop_replication_slot(name)
+        if dropped:
+            logger.info("Dropped physical replication slot '%s'", name)
```
CyberDem0n, line 346:
```suggestion
            logger.info("Dropped physical replication slot '%s'", name)
            if name in self._replication_slots:
                del self._replication_slots[name]
```

it will make everything a bit nicer.

patroni/postgresql/slots.py
```diff
@@ -332,6 +332,26 @@ def drop_replication_slot(self, name: str) -> Tuple[bool, bool]:
                             ' FULL OUTER JOIN dropped ON true'), name)
         return (rows[0][0], rows[0][1]) if rows else (False, False)
 
+    def _drop_physical_slot(self, name: str) -> bool:
```
CyberDem0n, line 335:
```suggestion
    def _drop_physical_slot(self, name: str) -> None:
```

patroni/postgresql/slots.py
```diff
@@ -332,6 +332,26 @@ def drop_replication_slot(self, name: str) -> Tuple[bool, bool]:
...
+        """Drop a physical replication slot by name.
+
+        .. note::
+            If not able to drop the slot, it will log a message and set the flag to reload slots.
+
+        :param name: name of the slot to be dropped.
+        :returns: ``True`` if the slot was successfully dropped, ``False`` otherwise.
```
CyberDem0n, line 342:
```suggestion
```

barthisrael:
Thanks for the suggestions, I think they make sense.
I still need to update the unit tests -- will come back here once possible and get that done.

CyberDem0n:
Merged, thank you @barthisrael!

patroni/postgresql/slots.py
```diff
@@ -332,6 +332,26 @@ def drop_replication_slot(self, name: str) -> Tuple[bool, bool]:
                             ' FULL OUTER JOIN dropped ON true'), name)
         return (rows[0][0], rows[0][1]) if rows else (False, False)
 
+    def _drop_physical_slot(self, name: str) -> None:
```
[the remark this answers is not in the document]
CyberDem0n: @Uyoooo sorry, my bad, I will fix it separately.

barthisrael:
> Merged, thank you @barthisrael!

Thank you all too 🎉

PR #3435, opened 2025-09-15
Strict synchronous mode, take two

CyberDem0n:
Implement real `synchronous_mode_strict`.
If there are no good candidates to be synchronous nodes according to `pg_stat_replication` we will set/keep `synchronous_standby_names` value based on values stored in `/sync` key.
Examples:
1. `/sync` key has `leader=node1`, `sync_standby=node2,node3`, and both standby nodes suddenly stopped streaming. In this case we will set/keep `synchronous_standby_names` set to `X (node2,node3)`, because we know that these nodes are supposed to have latest commit.
2. `/sync` key has `leader=node`, `sync_standby=null`, and we manually failed over to asynchronous `node3`. In this case `node3` will set `synchronous_standby_names` to `node1`, because we know that the former primary supposed to have latest commits.
3. `/sync` key is empty, and we enable strict synchronous mode. In this case we will set `synchronous_standby_names` to `__patroni_strict_sync_replica_placeholder__`, so that all commits will hang until some eligible node start streaming from the primary.

At the same time, we implemented restriction of possible values for `name` parameter in `patroni.yaml` and disallow using `__patroni_strict_sync_replica_placeholder__`.


patroni/ha.py
```diff
@@ -855,13 +853,63 @@ def disable_synchronous_replication(self) -> None:
...
+                    numsync != replication_state.numsync or \
+                    sync_type != replication_state.sync_type:
+                self.state_handler.sync_handler.set_synchronous_standby_names(voters, numsync)
+            elif voters:
+                msg = 'Continue using old value of synchronous_standby_names="{0}". '.format(
+                    self.state_handler.synchronous_standby_names())
+
+            # We use self._synchronous_strict_mode_activated to show warning only once.
+            if not self._synchronous_strict_mode_activated:
+                logger.warning('No active replication connections and synchronous_mode_strict is requested.'
```
hughcapet: should it maybe be "no active replication connections from patroni members"?

tests/test_ha.py
```diff
@@ -1858,25 +1871,80 @@ def test_process_sync_replication_prepromote(self):
...
         self.assertEqual(mock_set_sync.call_count, 1)
         self.assertEqual(mock_set_sync.call_args_list[0][0], (None,))
 
+        mock_set_sync.reset_mock()
+        mock_write_sync.reset_mock()
+        self.p.set_role(PostgresqlRole.REPLICA)
+        # Postgres 9.5, with synchronous_mode_strict, our name is written to leader of the /sync key.
+        # Old leader name is written to the /sync key and synchronous_standby_names.
+        with patch.object(global_config.__class__, 'is_synchronous_mode_strict', PropertyMock(return_value=True)), \
+                patch.object(Postgresql, 'synchronous_standby_names', Mock(return_value='othe')):
```
hughcapet: which call is it mocking? 🤔

tests/test_ha.py
```diff
@@ -1858,25 +1871,79 @@ def test_process_sync_replication_prepromote(self):
...
+        self.assertEqual(mock_set_sync.call_args_list[0][0], ('other',))
+
         self.p._major_version = 90600
         mock_set_sync.reset_mock()
         mock_write_sync.reset_mock()
         self.p.set_role(PostgresqlRole.REPLICA)
+        # Postgres 9.6, with synchronous_mode_strict, our name is written to leader of the sync key.
+        # Voters and synchronous_standby_names are set based on the old value of /sync key.
+        with patch.object(global_config.__class__, 'is_synchronous_mode_strict', PropertyMock(return_value=True)), \
+                patch.object(Postgresql, 'synchronous_standby_names', Mock(return_value='2 (foo,other)')):
```
hughcapet: `patch.object(Postgresql, 'synchronous_standby_names', Mock()` does not seem to be needed in other places in this test as well

hughcapet (approved):

PR #3450, opened 2025-09-26
Add support for notify-reload systemd unit type

rdunklau:
This allows systemd-reload command to wait for the configuration to
actually have been processed.

Encapsulate the import logic in a separate function to be used by the
already present notify implementation.

The release reminded me I should have upstreamed this as well.

### DIFF
--- a/patroni/daemon.py
+++ b/patroni/daemon.py
@@ -20,6 +20,18 @@ if TYPE_CHECKING:  # pragma: no cover
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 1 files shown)


rdunklau:
@CyberDem0n sorry didn't realize my formatter had run before pushing. It should be ok now.

rdunklau:
Thanks, I incorporated your changes and squashed them in the base commit.

hughcapet (approved):

PR #3453, opened 2025-09-29
Primary race backoff - preventing false-positive failovers

CyberDem0n:
We often see false-positive failovers on clusters running *without* watchdog. Investigation of Patroni logs from the former primary usually shows that Patroni entered a **frozen** state and couldn’t renew the leader lock within the configured `ttl`. Meanwhile, PostgreSQL on that primary typically continued to operate normally and standbys continued to receive and replay WAL.
Because these frozen states are usually short (often < 60s), a simple mitigation is to increase `ttl` from the common default (30s) to 60–90s. That reduces spurious failovers but increases the time to failover in a real outage - a trade-off some deployments do not want.

However, such mitigation doesn't really help with Consul setups, when Consul agent on the primary is restarted for upgrade. Such restart invalidates all sessions and replica nodes start leader race quicker than Patroni on primary can recreate its session and keys.

The **Primary Race Backoff** feature attacks the same problem from the other side:
- Standby nodes detect an active replication stream by observing the WAL receiver process and the WAL progress.
- If replication is active and advancing, standbys **postpone** participating in the leader race (they back off from initiating a race).
- The decision is made by comparing the current `max(receive_lsn, replay_lsn)` with the last value observed in the heartbeat loop. If progress is observed, the standby assumes the primary is still running and delays the race.
- To avoid waiting forever in case the Patroni primary process dies while keeping a stale replication stream, the backoff is bounded by the cluster-level `primary_race_backoff` parameter. Once that timeout elapses, the normal leader race process proceeds.

Benefits:
- Reduces false-positive failovers when Patroni on the primary becomes unresponsive briefly but PostgreSQL and replication continue working.
- Keeps `ttl` values lower if you prefer faster automatic failover on real failure (because the backoff lets the cluster avoid premature races without lengthening `ttl`).

By default `primary_race_backoff` is set to `0` (i.e., feature is disabled). Consider to enable this feature for clusters that run without watchdog.

### DIFF
--- a/patroni/postgresql/bootstrap.py
+++ b/patroni/postgresql/bootstrap.py
@@ -258,9 +258,25 @@ class Bootstrap(object):
         # supports additional user-supplied options, those are not validated
         maxfailures = 2
         ret = 1
-        not_allowed_options = ('pgdata', 'format', 'wal-method', 'xlog-method', 'gzip',
-                               'version', 'compress', 'dbname', 'host', 'port', 'username', 'password')
-        user_options = process_user_options('basebackup', options, not_allowed_options, logger.error)
+        not_allowed_options = ['pgdata', 'format', 'wal-method', 'xlog-method', 'gzip',
+                               'version', 'dbname', 'host', 'port', 'username', 'password']
+        pg_version = self._postgresql.config.pg_version
+        if pg_version < 150000:
+            not_allowed_options.append('compress')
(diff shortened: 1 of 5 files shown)


hughcapet:
do not forget to document the new parameter:)

hughcapet (approved):

Issue #3463, opened 2025-10-08
Internal initdb bootstrap fails if superuser password: key is present but has no value


mbanck-cd:
### What happened?

Using internal default bootstrap, this:
```
    superuser:
      username: postgres
      password: 
```
results in this:
```
2025-10-07 09:26:40,186 INFO: trying to bootstrap a new cluster
2025-10-07 09:26:40,187 ERROR: Exception during execution of long running task bootstrap
Traceback (most recent call last):
  File "/usr/lib/python3.12/site-packages/patroni/async_executor.py", line 166, in run
    wakeup = func(*args) if args else func()
             ^^^^^^^^^^^     
  File "/usr/lib/python3.12/site-packages/patroni/postgresql/bootstrap.py", line 402, in bootstrap
    return do_initialize(config.get(method)) and self._postgresql.config.append_pg_hba(pg_hba) \
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.12/site-packages/patroni/postgresql/bootstrap.py", line 132, in _initdb 
    os.write(fd, self._postgresql.config.superuser['password'].encode('utf-8'))
                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
AttributeError: 'NoneType' object has no attribute 'encode'
```
Because there is a `password` key but it has no value. Removing the `password:` line fixes this, *but* just taking over a running cluster with a the above configuration works fine, this is just an issue on bootstrap.

Debian's sample config has been using the above syntax for a long time (see e.g. https://salsa.debian.org/postgresql/patroni/-/blob/master/debian/config.yml.in?ref_type=heads#L126) but it also uses a custom bootstrap method in order to integrate with Debian, so I never noticed until now. I think it might be worthwhile to catch this here? Will open a PR for it.

### How can we reproduce it (as minimally and precisely as possible)?

Bootstrap patroni with the following changes to `postgres0.yml` should suffice:

```
diff --git a/postgres0.yml b/postgres0.yml
index 1546045c..068d4343 100644
--- a/postgres0.yml
+++ b/postgres0.yml
@@ -100,7 +100,7 @@ bootstrap:
 postgresql:
   listen: 127.0.0.1:5432
   connect_address: 127.0.0.1:5432
-
+  use_unix_socket: true
[...]

PR #3464, opened 2025-10-08
Only write superuser password during initdb bootstrap if it is non-empty

mbanck-cd:
Close https://github.com/zalando/patroni/issues/3463


patroni/postgresql/bootstrap.py
```diff
@@ -127,7 +127,7 @@ def error_handler(e: str) -> None:
         if self._postgresql.config.superuser:
             if 'username' in self._postgresql.config.superuser:
                 options.append('--username={0}'.format(self._postgresql.config.superuser['username']))
-            if 'password' in self._postgresql.config.superuser:
+            if 'password' in self._postgresql.config.superuser and self._postgresql.config.superuser['password']:
```
CyberDem0n, line 130: I wonder if we should support the case when password is an empty string.
```suggestion
            if isinstance(self._postgresql.config.superuser.get('password') , str):
```

Actually, --validate-config successfully catching missing password value:
```
postgresql.authentication.superuser.password None is not a string
```

mbanck-cd: Hrm, I kept it that way since https://github.com/patroni/patroni/issues/928 cause not having `password:` there back then did not work, but I guess I should have changed it since #930 got merged. I will at least change it in the Debian package now.

I should really use `patroni --validate-config` more often...

PR #3467, opened 2025-10-09
Check pg_stat_replication to verify that synchronous_standby_names is applied

CyberDem0n:
After updating `synchronous_standby_names` and reloading we:
1. waited until `SHOW synchronous_standby_names` reported value change
2. memorized value returned by `pg_current_wal_insert_lsn()`
3. waited until standby from `synchronous_standby_names` confirmed write/flush/apply LSN from [2]

Unfortunately, this is not enough, because change of `SHOW synchronous_standby_names` value reported by current connection doesn't mean that Postgres already started using the new value.
It seems to be the best way to verify that the change is active by querying `pg_stat_replication`. Only when set nodes reported by `pg_stat_replication` as sync/priority/quorum matches set of nodes from `synchronous_standby_names` we will count that change of `synchronous_standby_names` is applied.

### DIFF
--- a/.github/workflows/mapping.py
+++ b/.github/workflows/mapping.py
@@ -1,1 +1,1 @@
-versions = {'etcd': '9.6', 'etcd3': '16', 'consul': '17', 'exhibitor': '12', 'raft': '14', 'kubernetes': '15'}
+versions = {'etcd': '9.6', 'etcd3': '17', 'consul': '18', 'exhibitor': '12', 'raft': '14', 'kubernetes': '16'}
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -13,12 +13,12 @@ jobs:
     permissions:
       id-token: write
     steps:
-    - uses: actions/checkout@v4
+    - uses: actions/checkout@v5
 
-    - name: Set up Python 3.11
+    - name: Set up Python 3.14
       uses: actions/setup-python@v5
       with:
-        python-version: 3.11
+        python-version: 3.14
 
     - name: Install dependencies
       run: python .github/workflows/install_deps.py
--- a/.github/workflows/tests.yaml
+++ b/.github/workflows/tests.yaml
@@ -20,7 +20,7 @@ jobs:
         os: [ubuntu, windows, macos]
 
     steps:
-    - uses: actions/checkout@v4
+    - uses: actions/checkout@v5
 
     - name: Set up Python 3.7
       uses: actions/setup-python@v5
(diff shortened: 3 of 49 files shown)


ants:
I am still to review the changes in this PR, but I wanted to get my thoughts down on the parts that are not in this PR but would also need to be solved to close all durability race conditions.

When we increase strictness of `synchronous_standby_names` (reduce sync set or increase num sync) we need to wait for the durability to match the new condition before we can adjust quorum correspondingly.

This means in cases like [removing a node]([https://github.com/patroni/patroni/blob/8c53ac205203ace050ca7ec9f78a03f0f4e62c6b/patroni/quorum.py#L330) or [changing replication factor](https://github.com/patroni/patroni/blob/8c53ac205203ace050ca7ec9f78a03f0f4e62c6b/patroni/quorum.py#L385) the change has to wait. The theoretical minimum requirement is that last confirmed transaction before synchronous_standby_name became active is replicated according to the new rules. That seems tricky to track, the easier option would be to wait for the change to become active, generate a new transaction and wait for that to be durable (synchronous_commit=on).

Because this wait can take abritrarily long we can't block HA loop while waiting, so if it is not applied quickly enough we should bail out and retry on next loop. But would have to remember the fact that we need to get durability on that new commit.

I think this would also affect recovery, if we boot up Patroni after a crash we can't assume that s_s_n is actually applied to existing transactions. So would need to conservatively use what is in DCS sync key, and then use regular quorum resolver from that. I didn't yet check if this is already done that way or not.

CyberDem0n:
> When we increase strictness of synchronous_standby_names (reduce sync set or increase num sync) we need to wait for the durability to match the new condition before we can adjust quorum correspondingly.

I am nearly done implementing wait on `reduce sync`.

> or increase num sync

It seems to be the problem only for quorum commit. QuorumStateResolver() takes into account how many standby nodes are confirmed to be synchronous (`pg_stat_replication.sync_state = quorum` + `confirmed LSN >= pg_current_wal_insert_lsn() `from the moment when pg_stat_replication on primary changed). To fix it we just need to reset [SyncHandler._ready_replicas](https://github.com/patroni/patroni/pull/3467/files#diff-d402f811e6cb3160cf7681ba1fa64e916284aa43b3a62db08792e51652e70d21R295) if we detect that numsync is increased.

CyberDem0n:
> To fix it we just need to reset [SyncHandler._ready_replicas](https://github.com/patroni/patroni/pull/3467/files#diff-d402f811e6cb3160cf7681ba1fa64e916284aa43b3a62db08792e51652e70d21R295) if we detect that numsync is increased.

Scratch it. When only numsync is changing we can't detect it by looking at pg_stat_replication :(

Update.
Actually, wake up of sync waiters is performed by walsender process.
Every walsender is has its own view of synchronous_standby_names, which is updated on SIGHUP.
However, the good thing is that walsender processes are usually not blocked so badly comparing to checkpointer, because typically they read/send not more than 128kB at once. And, after that they process SIGHUP and apply new configuration.
If we can detect that `sent_lsn` - it provides some guaranties that walsender performed reload.
Or, if as I said we reset `SyncHandler._ready_replicas` and wait until standby caught up to current LSN, it will give quite good guaranties that quorum could be reduced.

PR #3479, opened 2025-10-23
Support digest ACLs in Patroni ZooKeeper set_acls configuration

kvazimoda24:
### Summary
This PR extends Patroni's set_acls support for ZooKeeper. Previously, set_acls always applied ACLs with the x509 scheme. Now, users can specify other supported ZooKeeper schemes, such as digest, directly in the key.

### Motivation
Many deployments use digest authentication in ZooKeeper to separate access for different applications. Before this change, it was impossible to automatically set digest ACLs for Patroni-created ZNodes, forcing users to manually adjust ACLs after creation.

### Changes
* Updated patroni/dcs/zookeeper.py to parse the scheme from the key in set_acls.
* If a scheme is provided (e.g., digest:username:hash), it is applied; otherwise, x509 is used by default.
* Documentation updated to reflect support for multiple schemes.
* Added examples demonstrating digest ACLs usage.

### Example Usage
```yaml
zookeeper:
  hosts: 127.0.0.1:2181
  auth_data:
    digest: principal2:secretpassword
  set_acls:
    digest:principal2:+pjROuBuuwNNSujKyH8dGcEnFPQ=: [ALL]
    CN=principal1: [CREATE, READ]
```

### Backward compatibility
Existing configurations using x509 ACLs continue to work without changes.

### Testing
* Verified that Patroni can create ZNodes with digest ACLs applied.
* Verified backward compatibility with x509 ACLs.
* zkCli.sh getAcl /service/<scope> shows the expected ACLs.

### DIFF
--- a/tests/test_zookeeper.py
+++ b/tests/test_zookeeper.py
@@ -154,7 +154,9 @@ class TestZooKeeper(unittest.TestCase):
     @patch('patroni.dcs.zookeeper.PatroniKazooClient', MockKazooClient)
     def setUp(self):
         self.zk = get_dcs({'scope': 'test', 'name': 'foo', 'ttl': 30, 'retry_timeout': 10, 'loop_wait': 10,
-                           'zookeeper': {'hosts': ['localhost:2181'], 'set_acls': {'CN=principal2': ['ALL']}}})
+                           'zookeeper': {'hosts': ['localhost:2181'],
+                                         'set_acls': {'CN=principal2': ['ALL'],
+                                                      'digest:principal1:R6IlbH3lMF4qX9UfEp/rOy1fv/4=': ['ALL']}}})
         self.assertIsInstance(self.zk, ZooKeeper)
 
     def test_reload_config(self):


CyberDem0n (commented):
Please also update unit tests so that added lines are covered: https://github.com/patroni/patroni/blob/a50d6f406b73ee21125e646ef4ad86795b779b38/tests/test_zookeeper.py#L157

CyberDem0n (commented):
Please also update unit tests so that added lines are covered: https://github.com/patroni/patroni/blob/a50d6f406b73ee21125e646ef4ad86795b779b38/tests/test_zookeeper.py#L157

kvazimoda24:
> Please also update unit tests so that added lines are covered:

I added a line to test_zookeeper.py with the new permissions format. I'm not entirely sure this is sufficient, as I have a poor understanding of how these tests are structured.

CyberDem0n (approved):

CyberDem0n:
@kvazimoda24 LGTM.
Since you are on ZooKeeper topic, maybe you can take over https://github.com/patroni/patroni/discussions/3475 and implement it as a separate PR?

kvazimoda24:
> Since you are on ZooKeeper topic, maybe you can take over #3475 and implement it as a separate PR?

I'm not quite sure what you're asking of me. Should I open a PR with the code that fixes the Kerberos interaction issue? Or should I solve the permissions issue?
If you choose the latter, then this issue should already be resolved by the current PR. Replace "digest" with "sasl" and so on.

In any case, I don't think I'm the best choice for any more serious changes to Patroni's code. I'm far from a programmer, and even this change was helped by ChatGPT.
The changes ChatGPT suggested—my Python knowledge was enough to understand whether they were the right solution and whether they would break anything. I'm not ready for anything more serious at the moment.

PR #3484, opened 2025-11-11
Add support for custom `pg_rewind` options

barthisrael:
### Summary

This PR introduces the `postgresql.rewind` configuration parameter, allowing users to pass custom command-line options to `pg_rewind`.

This provides flexibility for the PostgreSQL users, and is also a necessary enhancement for supporting TDE (Transparent Data Encryption) on proprietary Postgres, like EPAS (EDB Postgres Advanced Server), which requires passing custom flags during the rewind process.

### Key Changes

* **Refactor:** The `process_user_options` function was first moved from `bootstrap.py` to `utils.py` to make it a generic helper. Its unit tests were moved, and a minor bug in its error reporting was fixed.
* **Feature:** `patroni/postgresql/rewind.py` now reads the new `postgresql.rewind` configuration parameter, parses it using the `process_user_options` helper, and appends the resulting flags to the `pg_rewind` command.
* **Testing:** New unit tests have been added to `test_rewind.py` to validate the new feature, including checking for disallowed options.
* **Docs:** The new `postgresql.rewind` parameter is documented in `yaml_configuration.rst` with a usage example.
* **Validation:** The configuration schema in `validator.py` is updated to include the new parameter.

References: PAT-284.

### DIFF
--- a/patroni/validator.py
+++ b/patroni/validator.py
@@ -1059,6 +1059,7 @@ schema = Schema({
                 },
                 Optional("use_pg_rewind"): bool,
                 Optional("rewind"): [Or(str, dict)],
+                Optional("basebackup"): [Or(str, dict)],
                 Optional("pg_hba"): [str],
                 Optional("pg_ident"): [str],
                 Optional("pg_ctl_timeout"): IntValidator(min=0, raise_assert=True),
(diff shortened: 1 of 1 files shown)


patroni/validator.py
```diff
@@ -1058,6 +1058,7 @@ def validate_watchdog_mode(value: Any) -> None:
                     Optional("max_worker_processes"): IntValidator(0, 262143, raise_assert=True),
                 },
                 Optional("use_pg_rewind"): bool,
+                Optional("rewind"): [Or(str, dict)],
```
CyberDem0n: You added it only to `bootstrap.dcs.postgresql`, but it also needs to be added to `postgresql` section.

Another thing, which is a bit related, we don't have validator for `basebackup` options, maybe you can add it.

Issue #3491, opened 2025-11-18
Add "/streaming" API endpoint (or fix "/replica" ;-) that does not include replicas in archive recovery


bluthg:
### What happened?

The "/replica?lag=" endpoint call will happily reply "200" while the node is still in archive recovery.
Now we all know that the catch-up to a high write-load primary can take quite some time. The other day, we had a failover event, and our HAProxy polling resulted in more than 300 "UP" and "DOWN" events over the course of 22 minutes or so.

We believe that either "/replica" should not return 200 when the node is still recovering from archive *or* there should be an additional endpoint "/streaming" that only returns 200 when the node is actually streaming. Obviously, that endpoint should also offer a "&lag=" option.

### How can we reproduce it (as minimally and precisely as possible)?

* Take a replica down
* Produce WAL
* Start the replica again (while still writing to the leader)
* Watch the "/replica?lag=XX" endpoint closely

Obviously, a somewhat slow archive and/or a tuneable writing process helps to reproduce the issue.

### What did you expect to happen?

"/replica?lag=" should be less optimistic regarding lag.

### Patroni/PostgreSQL/DCS version

-   Patroni version: 4.0.5
-   PostgreSQL version: 16
-   DCS (and its version): etcd 3.4

### Patroni configuration file

```yaml
Sorry, can't share, but should not be relevant anyway.
```

### patronictl show-config

```yaml
Sorry, can't share, but should not be relevant anyway.
```

### Patroni log files

```shell
Sorry, can't share, but should not be relevant anyway.
```

### PostgreSQL log files

```shell
Sorry, can't share, but should not be relevant anyway.
```

### Have you tried to use GitHub issue search?

- [x] Yes

CyberDem0n:
The simplest and backward compatible way of doing it would be enhancing:
https://github.com/patroni/patroni/blob/e0d43022013f6f47dc2631e67b186ed18ca01406/patroni/api.py#L334-L337

And start supporting additional filters like `GET /replica?lag=32MB&replication_state=streaming`.

@bluthg do you want to contribute a patch? :)

bluthg:
I guess so :-)

PR #3492, opened 2025-11-18
adds "replication_state" query parameter to REST api

bluthg:
See #3491


bluthg:
That should do.
Also added some (very basic) tests.

bluthg:
So where do we stand?

hughcapet (approved):

PR #3498, opened 2025-11-24
Fix bug with failover_priority with synchronous_mode=on

CyberDem0n:
`tag.failover_priority` values were ignored when `synchronous_node_count>1`.

Besides that document limitation regarding `failover_priority` with `synchronous_mode=quorum`.

Close https://github.com/patroni/patroni/issues/3496


docs/yaml_configuration.rst
```diff
@@ -412,7 +412,7 @@ Tags
...
--  **failover_priority**: integer, controls the priority this node should have during failover. Nodes with higher priority will be preferred over lower-priority nodes if they received/replayed the same amount of WAL. However, nodes with higher values of receive/replay LSN are preferred regardless of their priority. If the ``failover_priority`` is 0 or negative - such node is not allowed to participate in the leader race and to become a leader (similar to ``nofailover: true``).
+-  **failover_priority**: integer, controls the priority this node should have during failover. Nodes with higher priority will be preferred over lower-priority nodes if they received/replayed the same amount of WAL. However, nodes with higher values of receive/replay LSN are preferred regardless of their priority. If the ``failover_priority`` is 0 or negative - such node is not allowed to participate in the leader race and to become a leader (similar to ``nofailover: true``). Known limitation: ``failover_priority`` currently doesn't wokr with :ref:`quorum-based synchronous replication <quorum_mode>`.
```
Uyoooo:
```suggestion
-  **failover_priority**: integer, controls the priority this node should have during failover. Nodes with higher priority will be preferred over lower-priority nodes if they received/replayed the same amount of WAL. However, nodes with higher values of receive/replay LSN are preferred regardless of their priority. If the ``failover_priority`` is 0 or negative - such node is not allowed to participate in the leader race and to become a leader (similar to ``nofailover: true``). Known limitation: ``failover_priority`` currently doesn't work with :ref:`quorum-based synchronous replication <quorum_mode>`.
```

PR #3500, opened 2025-11-25
Patroni-online major version upgrade

rdunklau:
The documentation suggests stopping patroni, and wiping the DCS or removing the initialization key from the DCS.

There is currently no tooling for removing the initialization key, so this PR adds a DELETE /initialization endpoint for this.

The other problem with performing an upgrade while keeping patroni online is that it doesn't reload the data_dir / bin_dir configuration correctly, and as such doesn't update the system identifier as expected. 

So with those two patches it becomes possible to upgrade PG while keeping Patroni online following this procedure:

 - pause patroni
 - make sure it is actually paused
 - stop postgresql
 - perform a pg_upgrade
 - reconfigure patroni to point to the new binaries and the new data dir
 - reload patroni
 - it will start complaining about the database sysid having changed
 - remove the initialization key using the new endpoint
 - check that the new initialization key is populated and used
 - resume patroni: it will generate the configuration for the new PG cluster, and start it

A few questions: 

- is this desirable ?
- should this be split into two separate pull-requests ? 
- I considered adding a command to patronictl for the new endpoint, should I ?
- do we want to update the pg_upgrade documentation to suggest this method ? There are a few footguns along the way (eg, not waiting for the correct database sysid to be populated)

### DIFF
--- a/patroni/postgresql/__init__.py
+++ b/patroni/postgresql/__init__.py
@@ -338,6 +338,11 @@ class Postgresql(object):
             return PgIsReadyStatus.UNKNOWN
 
     def reload_config(self, config: Dict[str, Any], sighup: bool = False) -> None:
+        self._bin_dir = config.get('bin_dir') or ''
+        new_data_dir = str(config.get('data_dir', ''))
+        if new_data_dir != self._data_dir:
+            self._data_dir = new_data_dir
+            self._sysid = ''
         self.config.reload_config(config, sighup)
         self._is_leader_retry.deadline = self.retry.deadline = config['retry_timeout'] / 2.0
 


CyberDem0n (commented):
Before it is gone to far:
- I don't really like REST API endpoint that allows removing keys in DCS, even if it is just /initialize key :(
- Change of bin_dir on the fly is very dangerous and can suddenly break the running cluster which is not intended to be upgraded.

You may take a look how I implemented major upgrade with rsync here: https://github.com/zalando/spilo/blob/trigger/postgres-appliance/major_upgrade/inplace_upgrade.py
Of course, it was just for very specific containerized deployments, but works quite well.

Since you are interested in major upgrade topic I would suggest you to join forces with @ants, who has some good ideas how to implement it on Patroni side.

rdunklau:
Thanks. 
> I don't really like REST API endpoint that allows removing keys in DCS, even if it is just /initialize key :( 

Fair point, but then maybe the documentation shouldn't mentioning this is an option ? It feels a bit weird to have the procedure mention this alongside the "completely wipe the dcs" option. The approach you linked does basically the same thing, and is a good alternative but is it actually supported to import modules from patroni and use them directly ?

> Change of bin_dir on the fly is very dangerous and can suddenly break the running cluster which is not intended to be upgraded.

This should probably check that patroni is paused in that case. But I understand if the project doesn't want such a change. 

> Since you are interested in major upgrade topic I would suggest you to join forces with @ants, who has some good ideas how to implement it on Patroni side.

Good idea, maybe we can collaborate on this.

PR #3501, opened 2025-11-27
Don't start stopped postgres on replica with nofailover tag

CyberDem0n:
Close https://github.com/patroni/patroni/issues/3499


jobayer110:
###

PR #3510, opened 2025-12-05
ha: provide the desired shutdown mode for manual failovers

frouioui:
The default behavior of the manual failover is to shutdown the current primary using a graceful (`fast`) shutdown. This approach is more than valid and should be used in most cases by everyone using Patroni as it fully guarantees that all connections are correctly drained and all risks of split brain are eliminated before promoting the new primary.

However, this approach causes longer downtime period when failing over. The main loop in `_do_stop`, ensuring the primary is shutdown, sleeps every second between two iterations. A primary stop timeout of 1 second could be used, but this will also cause the downtime to be at the very least > 1000ms long.

For clusters with an extremely low downtime tolerance, and a good tolerance for slower replica restart post-demotion, using an `immediate` shutdown mode is more beneficial. The `immediate` shutdown mode in Postgres sends a `SIGQUIT` to the postmaster, and then to the backends, terminating opened connections near-instantaneously and by-passing the `HOLD_INTERRUPTS` protections that the `fast` shutdown mode respects. This however, will cause the demoted replica to go through a crash recovery upon restart.

This Pull Request introduces a new flag to `patronictl failover` and a new field in the DCS `/failover` key: `mode`. The value can be either `graceful` (default, `fast` shutdown, same behavior as today), or `immediate` (use an `immediate` shutdown, `SIGQUIT`).


CyberDem0n (commented):
@frouioui I don't see any changes in ha.py using `mode`. That is, this PR isn't doing anything useful at the moment.
Besides that, I would be strictly against immediate demote for non-sync replication modes.

frouioui:
@CyberDem0n, apologies, bad rebase. Branch should look more complete now.

> Besides that, I would be strictly against immediate demote for non-sync replication modes.

Understood. That follows the same reasoning as the primary-stop-timeout where we do not allow SIGQUIT/SIGTERM for non-sync replication modes. Would adding a validation to prevent `mode = immediate` for non-sync replication modes make sense, or documenting in the documentation be enough?

tests/test_ha.py
```diff
@@ -1967,3 +1967,26 @@ def test_is_failover_possible(self):
...
+        f = Failover.from_node(1, '{"leader": "a", "member": "b"}')
+        self.assertEqual(f.mode, 'graceful')
+
+        # Test explicit graceful mode
+        f = Failover.from_node(1, '{"leader": "a", "member": "b", "mode": "graceful"}')
+        self.assertEqual(f.mode, 'graceful')
+
+        # Test immediate mode
+        f = Failover.from_node(1, '{"leader": "a", "member": "b", "mode": "immediate"}')
+        self.assertEqual(f.mode, 'immediate')
```
CyberDem0n: This is not the best place for this test.
I would suggest moving it to test_kubernetes.py

CyberDem0n (commented):
The usual trick to reduce time required for graceful shutdown is to manually run a few times CHECKPOINT before running pg_ctl stop.

When I had a problem of reducing switchover downtime on busy clusters I just manually run CHECKPOINT until it started taking some reasonable time (lets say less than 1s) and only after that called `patronictl switchover`.

In case of switchover Patroni runs CHECKPOINT just once: https://github.com/patroni/patroni/blob/3e3a4e134699154eb6fa8fc63f5d54d8256f1536/patroni/postgresql/__init__.py#L857-L896

It is good enough for not very busy clusters, however for clusters with high write workload it easily happens that buffer pool accumulates yet another big chunk of dirty pages.
The first CHECKPOINT run on busy cluster may take 1 minute and even more.
The second run is usually a bit shorter. The third, is shorter than a second, the forth is shorter than a third, and so on, until it stabilizes and CHECKPOINT time becomes relatively small.

Have you actually considered improving this part?

hughcapet:
@frouioui are you planning to continue working on this pr?

frouioui:
@hughcapet, apologies no. We decided to no continue with this feature internally due to the various risks it can cause, including the high probability of dropping sync'ed logical slots during a failover.

Closing this PR for now.

PR #3521, opened 2025-12-21
[Draft] Patroni in-place upgrade feature

ants:
The goal of this feature is to provide full upgrade automation with the resiliency Patroni is known for. The intended user experience is that users make sure binaries for the new version are available on all cluster members and then just issue `patronictl upgrade`. No extra steps needed. The goal is to serve both full-automation and sysadmin managed systems equally well.

This is currently a draft patch to get feedback on the general design before going through the effort of polishing things to a mergeable state. Notably, testing is still missing to avoid adding extra inertia to shuffling the APIs around.

## Design

The general approach is lifted from inplace_upgrade.py in Spilo. The major modifications are to move the upgrade to happen in Patroni process, and to move things that were Spilo container image specific into a configuration/plugin based architecture.

The reason for doing the upgrade in process is to have greater control, ability to interact with the HA loop in a nicer way and generally better robustness.

### Adding version concept to Patroni

Right now Patroni has no facilities to talk about multiple PostgreSQL data directories and versions at the same time - the Postgresql class is a singleton. This patch adds the capability for the upgrade process to create a new Postgresql instance and swap the implementation in the Ha class as needed.

Right now I split the init function into two pieces, one that is general class initialization and the other is state discover performed on startup. Then added a new_version() method that uses __new__ to create a copy and only run the initialization part. The main entry point is still called __init__ to keep the API same for the test suite. Should we change the protocol here so that new_version() can also go through the __init__ path?

Ideally the PostgreSQL object should be split up to separate out instance configuration from data dir configuration and currently active state. However I did not wish to make this refactoring a prerequisite for the upgrade feature.

## Version specific paths handling

Right now the database version is implicitly encoded into the Patroni config file as the path to PostgreSQL binaries. The idea of Patroni changing its own config during the upgrade seemed bad, and having the user change the config during the upgrade even worse. 

The proposed solution is to generate the path to PostgreSQL binaries from the version number. During bootstrap the version number will be determined by a new `bootstrap.version` parameter. Like other bootstrap parameters this is only effective during bootstrap, after bootstrap the version to use is available on local disk. Currently implemented way is using `postgresql.bin_dir_template` where the substring `{major_version}` is replaced.

The other problem is version specific data directories. VM based installs commonly use distribution directory layouts that use a version specific data directory, and config directory in case of Debian derivatives. Generating those paths is easily solved by adding `data_dir_template` and `config_dir_template`.

With a templated data dir the issue becomes determining where is the currently active data directory. Ee can't determine it from the data dir because we don't know which one to look at. The version also can't come from DCS because DCS might not be available at startup. And if upgrades fail mid flight we need to keep track what is happening in the filesystem. For this I added `state_file` which is a JSON file containing version and status of the data dir (see patroni.postgresql.DataDirStatus). 

The version specific paths could also be explicitly configured using a map of version -> paths. But my hope is to be able to have the distribution packaging provide defaults for these paths out of the box. That seems simpler with a simple string template.

## Single data dir handling

If data dir is to remain in the same place after the upgrade then during the upgrade the data dirs need to be swapped. This is implemented by allowing the data dir to be changed in PostgreSQL object. The order of steps is to record the intent in status file 'swapping', then moving datadirs around, updating the Postgresql objects with new locations, and then updating the status file.

## Upgrade state handling

Instead of using maintenance mode, upgrade instead adds a new state stored in DCS. See `patroni.dcs.Upgrade` and the state diagram in `docs/upgrade-states.svg`. The primary will run the upgrade using the async executor, in case of failures either the primary or any of the replicas will have enough information to roll back or roll forward the upgrade. The upgrade process also records progress information in the upgrade state so clients can observe the process.

Upgrade state is stored under /status key as I didn't want to add yet another DCS key. The actual correct location seems to be the `/initialize` key because conceptually both are concerned about what the current cluster is storing. Changing that would be a headache because it's currently a plain string.

The upgrade is started by `POST /upgrade` on the REST API, this creates a new Upgrade state to store in the DCS and hands over the upgrade to the async executor and returns.  Progress can then be monitored by observing the DCS upgrade state. The DCS upgrade state also holds a copy of the upgrade configuration as it seemed prudent to not let the upgrade configuration change in the middle of an active upgrade.

## Upgrading replicas

Upgrading replicas is behind a plugin abstraction to allow for different methods. Right now only the rsyncd method is implemented. But others will definitely be following, including the simple option of just `reinit`-ing replicas. But for more secure environments, adding TLS to rsync would be very much needed, either by going over SSH or by implementing a TLS proxy for rsyncd.

With rsyncd the general process looks similar, but it makes use of being able to do API calls between members. This makes it unnecessary to do the `COPY TO PROGRAM` hack. Instead when it is time to upgrade replicas primary will do a `POST /rsync` call. This will start an async process that ends up running ReplicaRsync.perform_rsync().

Because we'd like the upgrade process to be resilient to something that might fail a long API call the completion is tracked via polling a `GET /rsync` call. This will return either, success, failure or still running after a small wait for completion.

## Open questions

* Should we disallow upgrade to proceed if some specific number/fraction of replicas fail to upgrade?
* Which parts of HA loop we would like to still execute on replicas while the primary is upgrading?
* How to make upgrades be a part of behave tests?

## Todo

* Sanitycheck all cluster members before the upgrade.
* Handle tablespaces and waldir.
* Reinit based replica upgrade.
* Automatic reinit of replicas that failed to upgrade, or after a partial upgrade rolled back.
* Handle recovering from a failed rsync. Old dir should always be usable before we start postgres.
* Sprinkle more timeouts everywhere. Can be way more liberal when a failed upgrade can always be recovered.
* The plugin API is a placeholder experiment right now. Should try to properly implement a few more plugins to figure out what works.

## Features not included in first version

* Schedule upgrades like switchovers and restarts can be scheduled.
* Optionally include required cluster members in upgrade to avoid members left behind.
* Make pg_upgrade cancellable.
* SSH based rsync.
* TLS-proxied rsync.


docs/upgrades.rst
```diff
@@ -0,0 +1,107 @@
...
+
+Example configuration:
+
+.. code:: YAML
+
+    bootstrap:
+        version: 15
+
+    postgresql:
+        data_dir_tempalte: /var/lib/pgsql/{major_version/data/
```
hughcapet:
```suggestion
        data_dir_tempalte: /var/lib/pgsql/{major_version}/data/
```

docs/upgrades.rst
```diff
@@ -0,0 +1,107 @@
...
+
+Example configuration:
+
+.. code:: YAML
+
+    bootstrap:
+        version: 15
+
+    postgresql:
+        data_dir_tempalte: /var/lib/pgsql/{major_version}/data/
```
hughcapet:
```suggestion
        data_dir_template: /var/lib/pgsql/{major_version}/data/
```

sundongkim-dev:
Hi @ants, @hughcapet — is this still active, or parked?

I've read through the diff and the design notes. It looks like the open questions in the description (the `Postgresql` refactoring, the DCS state location) need a maintainer call before more code is worth writing — so before I invest anywhere near this, I'd just like to know whether the direction is settled and whether help would be welcome.

PR #3527, opened 2026-01-16
Add dynamic synchronized_standby_slots management for PostgreSQL 17+

PavelZaytsev:
This feature automatically manages the synchronized_standby_slots parameter to keep it in sync with synchronous_standby_names, ensuring that logical replication slots are synchronized to the same physical standbys used for synchronous replication.

Changes:
- Add dynamic_synchronized_standby_slots_enabled property to global_config (defaults to True, can be disabled via configuration)
- Add _update_synchronized_standby_slots_from_ssn() method in config.py to parse synchronous_standby_names and update synchronized_standby_slots
- Integrate the new method into set_synchronous_standby_names() for PG 17+
- Add debug logging throughout ha.py for synchronous standby operations

This addresses the need to maintain consistency between physical synchronous replication and logical slot synchronization in PostgreSQL 17+.


CyberDem0n (commented):
Thanks for working on it @PavelZaytsev!

TBH, this implementation doesn't look nice/correct.
1. Parsing of `synchronous_standby_names` adds extra complexity and not really needed. ConfigHandler.set_synchronous_standby_names() is always called from SyncHandler.set_synchronous_standby_names(), and the last one is only called from ha.py, with `sync` argument containing what we need. However, hooking into SyncHandler.set_synchronous_standby_names() seems to be not enough, because of [2]
2. Consider scenario. We flip `dynamic_synchronized_standby_slots` to `on`, but cluster is stable and no updates of `synchronous_standby_names` happen. It will result it indefinite time before the configuration change is applied.
3. It looks like `synchronized_standby_slots` will not reset back to the value configured by user when feature is disabled.

How it could be improved/solved:
1. Still hook into SyncHandler.set_synchronous_standby_names() to automatically handle cases when `synchronous_standby_names` is for some reason updated.
2. Modify a _process_quorum_replication() and _process_multisync_replication() methods in ha.py. They should handle `dynamic_synchronized_standby_slots` changes when current synchronous configuration is stable and SyncHandler.set_synchronous_standby_names() isn't called.

PavelZaytsev:
@CyberDem0n I've addressed all your feedback:

- Moved `update_synchronized_standby_slots` from `SyncHandler` to `SlotsHandler` where it logically belongs
- Extracted a shared `_set_server_parameter()` method in `ConfigHandler` — both `set_synchronous_standby_names` and `set_synchronized_standby_slots` now delegate to it (no more direct `_server_parameters` access from outside)
- Single reload logic: `set_synchronized_standby_slots` only reloads when explicitly requested, avoiding double reloads
- Removed the redundant `test_dynamic_synchronized_standby_slots_enabled` test
- Removed debug logs from `config.py`
- Extracted `_handle_dynamic_sync_slots_toggle()` helper in `ha.py` to deduplicate the toggle logic
- Updated docs to mention `synchronous_mode` requirement and removed the incorrect `ignore_slots` warning
- Used `check_mode()` in `global_config.py`
- Rebased on latest master
- Added tests for coverage

Ready for another review when you get a chance. Thanks!

PavelZaytsev:
@CyberDem0n @hughcapet plz review when you get the chance

patroni/postgresql/sync.py
```diff
@@ -432,7 +432,14 @@ def set_synchronous_standby_names(self, sync: Collection[str], num: Optional[int
             prefix = 'ANY ' if global_config.is_quorum_commit_mode and self._postgresql.supports_quorum_commit else ''
             sync_param = f'{prefix}{num} ({sync_param})'
 
-        if not (self._postgresql.config.set_synchronous_standby_names(sync_param)
+        # Update synchronous_standby_names - this will reload if SSN changed
+        ssn_changed = self._postgresql.config.set_synchronous_standby_names(sync_param)
+
+        # Update synchronized_standby_slots to match sync members (PostgreSQL 17+)
+        # Only reload if SSN didn't change (to avoid double reload)
+        self._postgresql.slots_handler.update_synchronized_standby_slots(sync, reload=not ssn_changed)
```
CyberDem0n: Hmm, something isn't right here.
If set_synchronous_standby_names() did reload then update_synchronized_standby_slots() will not do it and therefore new synchronized_standby_slots value will not be applied.

CyberDem0n (commented):
There is a [thread](https://www.postgresql.org/message-id/flat/CAHg%2BQDfU7rOebrLDESPpHSgdiadKbpCOmBokcbmM6Gr%2BA5VobQ%40mail.gmail.com) in pgsql-hackers mailing list that makes a very good point.
Current handling of `synchronized_standby_slots` is inconsistent with quorum-based synchronous replication.

That is, in case of `synchronous_mode: on` we don't have any problem, because `synchronized_standby_slots` is guarantied to match `synchronous_standby_names`, however with `synchronous_mode: quorum` the requirement imposed by `synchronized_standby_slots` will be much stronger.
Example:
1. synchronous_standby_names = 'ANY 1 (node1, node2, node3)'
2. synchronized_standby_slots = 'node1, node2, node3'
3. node2 and node3 having some problems (for example lagging)

In this case there will be absolutely no issues with sync replication, but logical replication will be impacted.

I wonder, if we should stick to `synchronous_mode: on` only for now.

patroni/postgresql/sync.py
PavelZaytsev: @CyberDem0n 

You're right - the order was wrong. Fixed in `ec906d2b`:

The previous flow was:

`set_synchronous_standby_names()` updated SSN in `_server_parameters` and reloaded immediately - but `synchronized_standby_slots` still had the old value at that point, so the reload wrote the old slots value.
`update_synchronized_standby_slots(reload=False)` then updated the slots in `_server_parameters` but never reloaded - so the new value was never applied.

Now I update `synchronized_standby_slots` in `_server_parameters` first (no reload), then `set_synchronous_standby_names()` writes both values together when it reloads. If only the slots changed (SSN didn't), we trigger a reload manually. Still a single reload either way.

Also fixed a related bug in `_handle_dynamic_sync_slots_toggle()` that I noticed while looking at this: when the feature is toggled OFF, calling `update_synchronized_standby_slots(set())` was a no-op because that method bails out early when the feature is disabled. Switched to calling `set_synchronized_standby_slots(None, reload=True)` directly, so the override actually gets cleared.

patroni/postgresql/sync.py
```diff
@@ -432,7 +432,20 @@ def set_synchronous_standby_names(self, sync: Collection[str], num: Optional[int
...
 
-        if not (self._postgresql.config.set_synchronous_standby_names(sync_param)
+        # Update synchronized_standby_slots in _server_parameters first (no reload yet) so
+        # that the subsequent set_synchronous_standby_names() reload picks up both changes.
+        slots_changed = self._postgresql.slots_handler.update_synchronized_standby_slots(sync)
+
+        # Update synchronous_standby_names - this will reload if SSN changed,
+        # which also applies any pending synchronized_standby_slots change.
+        ssn_changed = self._postgresql.config.set_synchronous_standby_names(sync_param)
+
```
Maithem: Found a bug here while testing. For a Patroni member named patronidemo-1 (the k8s pod name):

- slot_name_from_member_name("patronidemo-1") returns patronidemo_1, the slot in pg_replication_slots and the name synchronized_standby_slots should list
- quote_standby_name("patronidemo-1") returns "patronidemo-1" for synchronous_standby_names. This PR feeds that quoted form back through slot_name_from_member_name, yielding u0034patronidemo_1u0034, which matches no slot in pg_replication_slots.

CyberDem0n: Nice catch @Maithem, thank you!

@PavelZaytsev behave tests would also have caught it, please add them.

PavelZaytsev: @CyberDem0n added them. Happy to add more if any specific case still feels uncovered - let me know.

patroni/global_config.py
```diff
@@ -249,5 +249,10 @@ def member_slots_ttl(self) -> int:
         """
         return self.get_int('member_slots_ttl', 1800, base_unit='s')
 
+    @property
+    def dynamic_synchronized_standby_slots_enabled(self) -> bool:
+        """``True`` if dynamic ``synchronized_standby_slots`` management is enabled."""
+        return self.check_mode('dynamic_synchronized_standby_slots')
```
CyberDem0n: Documentation declares that this feature depends on synchronous_mode, and it should be reflected here:
```suggestion
        return self.is_synchronous_mode() and self.check_mode('dynamic_synchronized_standby_slots')
```

patroni/ha.py
```diff
@@ -274,6 +274,24 @@ def __init__(self, patroni: Patroni):
         # used only in backoff after failing a pre_promote script
         self._released_leader_key_timestamp = 0
 
+        # Track the last known state of dynamic_synchronized_standby_slots feature to detect toggles
+        self._last_dynamic_sync_slots_enabled: Optional[bool] = None
+
+    def _handle_dynamic_sync_slots_toggle(self, sync_members: Collection[str]) -> bool:
```
CyberDem0n: This method should also be called from `Ha.disable_synchronous_replication()`.

patroni/ha.py
```diff
@@ -274,6 +274,24 @@ def __init__(self, patroni: Patroni):
...
+        feature_enabled = global_config.dynamic_synchronized_standby_slots_enabled
+        if feature_enabled == self._last_dynamic_sync_slots_enabled:
+            return False
+
+        logger.info("dynamic_synchronized_standby_slots changed to %s, updating slots", feature_enabled)
+        if feature_enabled:
+            self.state_handler.slots_handler.update_synchronized_standby_slots(sync_members, reload=True)
+        else:
+            # Feature was disabled - restore the user-configured value (or clear if none configured).
+            self.state_handler.config.restore_synchronized_standby_slots(reload=True)
```
CyberDem0n: I would prefer doing it the same way as here: https://github.com/patroni/patroni/blob/5d276234c542d4a16d5be57de3c61a3243943705/patroni/ha.py#L861-L863

That is, be introducing `ConfigHandler.synchronized_standby_slots` property and calling `set_synchronized_standby_slots(value, reload=True)`

patroni/postgresql/config.py
```diff
@@ -1077,6 +1077,16 @@ def get_server_parameters(self, config: Dict[str, Any]) -> CaseInsensitiveDict:
...
 
+        # When dynamic_synchronized_standby_slots is enabled, Patroni dynamically writes
+        # synchronized_standby_slots, so we must preserve our cached value here so that subsequent
+        # reload_config() calls don't reset it back to the user-configured value.
+        if global_config.dynamic_synchronized_standby_slots_enabled and self._postgresql.major_version >= 170000:
+            synchronized_standby_slots = self._server_parameters.get('synchronized_standby_slots')
+            if synchronized_standby_slots is None:
+                parameters.pop('synchronized_standby_slots', None)
+            else:
+                parameters['synchronized_standby_slots'] = synchronized_standby_slots
```
CyberDem0n, line 1097: These lines are reported as not covered by unit tests.

patroni/postgresql/sync.py
```diff
@@ -432,7 +435,20 @@ def set_synchronous_standby_names(self, sync: Collection[str], num: Optional[int
...
+        slots_changed = self._postgresql.slots_handler.update_synchronized_standby_slots(sync_members)
+
+        # Update synchronous_standby_names - this will reload if SSN changed,
+        # which also applies any pending synchronized_standby_slots change.
+        ssn_changed = self._postgresql.config.set_synchronous_standby_names(sync_param)
+
+        # If only synchronized_standby_slots changed (SSN didn't), trigger a reload to apply it.
+        if slots_changed and not ssn_changed and self._postgresql.state == PostgresqlState.RUNNING:
+            self._postgresql.config.write_postgresql_conf()
+            self._postgresql.reload()
```
CyberDem0n, line 457: Now these lines are not covered by unit tests.

CyberDem0n (commented):
I found a couple more problems.
Also, now there are a few conflicts caused by merge of #3435

patroni/global_config.py
PavelZaytsev: Done in [81435945](https://github.com/patroni/patroni/pull/3527/commits/81435945) — the property now reads `self.is_synchronous_mode and self.check_mode('dynamic_synchronized_standby_slots')` and the docstring spells out the dependency.

patroni/ha.py
PavelZaytsev: Added in [81435945](https://github.com/patroni/patroni/pull/3527/commits/81435945): `disable_synchronous_replication` now fires `_handle_dynamic_sync_slots_toggle(CaseInsensitiveSet())` right after restoring the user's SSN. With the new `is_synchronous_mode` gate the property is False once /sync is gone, so the off-edge triggers and the user-configured `synchronized_standby_slots` is restored in the same loop.

patroni/ha.py
PavelZaytsev: Refactored in [81435945](https://github.com/patroni/patroni/pull/3527/commits/81435945) to match the SSN pattern

patroni/postgresql/config.py
PavelZaytsev: Covered in [81435945](https://github.com/patroni/patroni/pull/3527/commits/81435945)

patroni/postgresql/sync.py
PavelZaytsev: Covered in [81435945](https://github.com/patroni/patroni/pull/3527/commits/81435945)

PavelZaytsev:
> I found a couple more problems. Also, now there are a few conflicts caused by merge of #3435

@CyberDem0n 

Thanks! All five comments addressed in [81435945](https://github.com/patroni/patroni/pull/3527/commits/81435945). The #3435 conflicts are resolved in the merge commit [3cbc69e3](https://github.com/patroni/patroni/pull/3527/commits/3cbc69e3) — kept the new `_handle_synchronous_strict_mode` and FIRST↔ANY transitions intact while threading the dyn-slots toggle through the call sites; all 689 unit tests pass and pyright shows no new errors vs. master.

tests/test_sync.py
```diff
@@ -195,6 +195,63 @@ def value_in_conf():
         mock_reload.assert_called()
         self.assertEqual(value_in_conf(), "synchronous_standby_names = '\"a-1\"'")
 
+        # PG17+ with dynamic_synchronized_standby_slots: ensure a manual reload happens
+        # when only synchronized_standby_slots changes (synchronous_standby_names unchanged).
+        from patroni.postgresql import PostgresqlState
+        from unittest.mock import PropertyMock
```
CyberDem0n: Please move it to the top and make sure that it passes isort check.

features/environment.py
```diff
@@ -975,6 +975,14 @@ def server_version(self):
         for p in self._processes.values():
             if p._conn:
                 return p._conn.server_version
+        
+        # Fallback to postgres binary version if no instances are started yet
+        from patroni.postgresql.misc import postgres_major_version_to_int
+        from patroni.utils import get_major_version
+        try:
+            return postgres_major_version_to_int(get_major_version())
```
CyberDem0n:
```suggestion
        from patroni.postgresql.misc import postgres_version_to_int
        from patroni.utils import get_postgres_version
        try:
            return postgres_version_to_int(get_postgres_version())
```

kylorend3r:
I think this feature will add complexity and can influence the user experience in a notorious way.  Because, the problem is still being discussed in the community which is about synchronized_standby_slots and synchronized_standby_names semantics. 

if the slot was already stalled before failover (because one of the servers couldn't keep up and blocked slot advancement via AND-semantics), then:

  - The slot position on the old primary = P (stalled, far behind current WAL position Q)
  - The slot copies on server1 and server2 = also P (synced to the stalled position)
  - After failover to server2: slot is at P, WAL is at Q

The distinction between synchronized_standby_slots and synchronous_standby_names (AND-semantics vs ANY-semantics) creates a gap where physical replication is healthy but logical replication is silently broken. 

The way I see it for now inside the Patroni we can rely on `maximum_lag_on_failover` already exists and implemented. If the lag is higher than maximum_lag_on_failover patroni excludes it and we can remove its slot from synchronized_standby_slots and reload config. 


| Mode | `synchronous_standby_names` | `synchronized_standby_slots` |
|------|-----------------------------|------------------------------|
| Sync (`FIRST N`) | unchanged | unchanged |
| Quorum — all standbys within `maximum_lag_on_failover` | full quorum pool | full quorum pool |
| Quorum — standby lag > `maximum_lag_on_failover` | full quorum pool | standby excluded(config reload) |
| Quorum — excluded standby lag < 50% of `maximum_lag_on_failover` | full quorum pool | standby re-added(config reload) |

To sum up, in addition to your work I suggest to implement the following changes. 

1. Quorum-aware filtering using maximum_lag_on_failover

A standby with lag > `maximum_lag_on_failover` is already excluded from
leader election (`ha.py:is_failover_possible()`). Syncing logical slots to it serves no purpose since the slot copy will never be used on failover and actively harms by blocking logical slot advancement on the primary. Once the node is deemed eligible again we can add it back to synchronized_standby_slots list. 

2. Hysteresis to prevent reload storms

A standby oscillating around maximum_lag_on_failover would trigger a pg_reload_conf() every HA loop without damping. A fixed 50% re-admission threshold is applied. 

A standby oscillating between 99 MB and 101 MB with a 100 MB threshold produces at most two reloads total (one eviction, one re-admission) regardless of how long it oscillates. Per-member exclusion state is tracked in HA._syncslots_excluded and resets on Patroni restart.

This also handles new nodes correctly: a freshly joined standby starts with high lag, stays excluded from synchronized_standby_slots while it catches up, and is admitted automatically once genuinely ready preventing the primary's logical slot from stalling during bootstrap.


3. wal_level Guard

We need to add both at config-load time (validator cross-check) and at runtime (update_synchronized_standby_slots() early return with a warning log) so the misconfiguration is surfaced clearly rather than silently.

I created a simple patch(some parts written with AI)&demo but wanted to ask your opinion first. 

This feature will require intensive unit tests also. 

@CyberDem0n @PavelZaytsev

CyberDem0n:
@kylorend3r it is not a news that existing semantic of `synchronized_standby_slots` is incompatible with quorum based semantic of `synchronous_standby_names`.

What you suggest as a workaround with filtering out standby based on their replication lag is already covered by [maximum_lag_on_syncnode](https://patroni.readthedocs.io/en/latest/replication_modes.html#maximum-lag-on-synchronous-node) parameter. Nodes with lag exceeding configured threshold will be excluded from  `synchronous_standby_names` and therefore from `synchronized_standby_slots`.

Regarding "reload storms" - the `maximum_lag_on_syncnode` feature exists for more than 5 years, and so far there were no complains about reload storms.

kylorend3r:
My question is can we maximum_lag_on_syncnode or maximum_lag_on_failover to manage slots dynamically missing in this MR or should we extend maximum_lag_on_syncnode to manage the slots ? 

If we extend it patroni will be able to manage slots in ANY semantics. 

Did I get it correctly ?

CyberDem0n:
@kylorend3r `maximum_lag_on_syncnode` will just work.
The way this feature is implemented - it maps `synchronous_standby_names` value to `synchronized_standby_slots`.
We can do it, because application_names for replication connection and slot names are mapped from node names.

If `maximum_lag_on_syncnode` excludes some nodes from `synchronous_standby_names` because of their replication log, replication slots for these nodes are also excluded from `synchronized_standby_slots`.

I just tried it locally:
1. started cluster with 3 nodes
```
2026-06-16 11:39:48,126 INFO: no action. I am (postgresql0), the leader with the lock
2026-06-16 11:39:58,103 INFO: Lock owner: postgresql0; I am postgresql0
2026-06-16 11:39:58,123 INFO: Enabled synchronous replication
2026-06-16 11:39:58,123 INFO: Assigning synchronous_standby_names to ANY 1 (postgresql1,postgresql2)
server signaled
2026-06-16 11:39:58.126 CEST [1579702] LOG:  received SIGHUP, reloading configuration files
2026-06-16 11:39:58.128 CEST [1579702] LOG:  parameter "synchronized_standby_slots" changed to "postgresql1,postgresql2"
2026-06-16 11:39:58.128 CEST [1579702] LOG:  parameter "synchronous_standby_names" changed to "ANY 1 (postgresql1,postgresql2)"
2026-06-16 11:39:58.129 CEST [1579885] LOG:  standby "postgresql2" is now a candidate for quorum synchronous standby
2026-06-16 11:39:58.129 CEST [1579885] STATEMENT:  START_REPLICATION SLOT "postgresql2" 0/05000000 TIMELINE 2
2026-06-16 11:39:58.129 CEST [1579874] LOG:  standby "postgresql1" is now a candidate for quorum synchronous standby
2026-06-16 11:39:58.129 CEST [1579874] STATEMENT:  START_REPLICATION SLOT "postgresql1" 0/03000000 TIMELINE 2
2026-06-16 11:39:59,235 INFO: Setting leader to postgresql0, quorum to 1 of (postgresql1, postgresql2)
2026-06-16 11:39:59,254 INFO: no action. I am (postgresql0), the leader with the lock
```

2. `kill -STOP <pidof walreceiver for postgresql2>`
3. `create table test as select * from generate_series(1, 1000000);` on primary.
```
2026-06-16 11:41:58,114 INFO: no action. I am (postgresql0), the leader with the lock
2026-06-16 11:42:08,104 INFO: Lock owner: postgresql0; I am postgresql0
2026-06-16 11:42:08,114 INFO: Assigning synchronous_standby_names to ANY 1 (postgresql1)
server signaled
2026-06-16 11:42:08.117 CEST [1579702] LOG:  received SIGHUP, reloading configuration files
2026-06-16 11:42:08.118 CEST [1579702] LOG:  parameter "synchronized_standby_slots" changed to "postgresql1"
2026-06-16 11:42:08.119 CEST [1579702] LOG:  parameter "synchronous_standby_names" changed to "ANY 1 (postgresql1)"
2026-06-16 11:42:08,219 INFO: Setting leader to postgresql0, quorum to 0 of (postgresql1)
2026-06-16 11:42:08,227 INFO: no action. I am (postgresql0), the leader with the lock
```
That is, postgresql2 was removed.

4. `kill -CONT <pidof walreceiver for postgresql2>`
```
2026-06-16 11:42:28,125 INFO: no action. I am (postgresql0), the leader with the lock
2026-06-16 11:42:38,104 INFO: Lock owner: postgresql0; I am postgresql0
2026-06-16 11:42:38,131 INFO: Setting leader to postgresql0, quorum to 1 of (postgresql1, postgresql2)
2026-06-16 11:42:38,135 INFO: Assigning synchronous_standby_names to ANY 1 (postgresql1,postgresql2)
server signaled
2026-06-16 11:42:38.138 CEST [1579702] LOG:  received SIGHUP, reloading configuration files
2026-06-16 11:42:38.140 CEST [1579702] LOG:  parameter "synchronized_standby_slots" changed to "postgresql1,postgresql2"
2026-06-16 11:42:38.140 CEST [1579702] LOG:  parameter "synchronous_standby_names" changed to "ANY 1 (postgresql1,postgresql2)"
2026-06-16 11:42:38.141 CEST [1580514] LOG:  standby "postgresql2" is now a candidate for quorum synchronous standby
2026-06-16 11:42:38.141 CEST [1580514] STATEMENT:  START_REPLICATION SLOT "postgresql2" 0/05000000 TIMELINE 2
2026-06-16 11:42:38,249 INFO: no action. I am (postgresql0), the leader with the lock
```
That is, postgresql2 caught up and was added back to synchronized_standby_slots and synchronous_standby_names.

patroni/global_config.py
```diff
@@ -249,5 +249,14 @@ def member_slots_ttl(self) -> int:
...
         return self.get_int('member_slots_ttl', 1800, base_unit='s')
 
+    @property
+    def dynamic_synchronized_standby_slots_enabled(self) -> bool:
+        """``True`` if dynamic ``synchronized_standby_slots`` management is enabled.
+
+        Depends on synchronous replication being enabled because Patroni derives the
+        ``synchronized_standby_slots`` value from the current set of synchronous standbys.
+        """
+        return self.is_synchronous_mode and self.check_mode('dynamic_synchronized_standby_slots')
```
CyberDem0n:
```suggestion
        return self.is_synchronous_mode and self.use_slots and self.check_mode('dynamic_synchronized_standby_slots')
```

features/steps/slots.py
```diff
@@ -136,3 +136,67 @@ def dcs_key_contains(context, name, subkey, key):
...
+            actual_slots = {row[0] for row in context.pctl.query(
+                name, "SELECT slot_name FROM pg_replication_slots WHERE slot_type = 'physical'"
+            ).fetchall()}
+            if declared.issubset(actual_slots):
+                return
+        except Exception:
+            pass
+        time.sleep(1)
+    assert False, (f"synchronized_standby_slots on {name} (={declared}) does not match existing "
+                   f"physical slots (={actual_slots}) after {time_limit} seconds")
```
CyberDem0n: IMO, there is no need to have timeouts here.
If node is listed in synchronous_standby_names and synchronized_standby_slots it already guarantees that replication slot should be there.

And, actually, after saying that I realized that global_config.dynamic_synchronized_standby_slots_enabled should also depend on global_config.use_slots (see suggestion).

features/steps/slots.py
```diff
@@ -136,3 +136,67 @@ def dcs_key_contains(context, name, subkey, key):
...
     assert key not in response or subkey not in response[key], f"{name} key in DCS has {subkey} in {key}"
+
+
+@then("synchronized_standby_slots on {name:2} is set to '{value}' after {time_limit:d} seconds")
+def check_synchronized_standby_slots(context, name, value, time_limit):
+    time_limit *= context.timeout_multiplier
+    max_time = time.time() + int(time_limit)
+    expected = set(value.split(',')) if value else set()
+    actual = None
+    while time.time() < max_time:
```
CyberDem0n: We are planning to use time.monotonic() everywhere. Ref https://github.com/patroni/patroni/pull/3612
```suggestion
    max_time = time.monotonic() + int(time_limit)
    expected = set(value.split(',')) if value else set()
    actual = None
    while time.monotonic() < max_time:
```

features/steps/slots.py
```diff
@@ -136,3 +136,67 @@ def dcs_key_contains(context, name, subkey, key):
...
+    assert False, (f"synchronized_standby_slots on {name} is not set to '{value}' "
+                   f"(found '{actual}') after {time_limit} seconds")
+
+
+@then('synchronized_standby_slots on {name:2} is empty after {time_limit:d} seconds')
+def check_synchronized_standby_slots_empty(context, name, time_limit):
+    time_limit *= context.timeout_multiplier
+    max_time = time.time() + int(time_limit)
+    actual = None
+    while time.time() < max_time:
```
CyberDem0n: We are planning to use time.monotonic() everywhere. Ref https://github.com/patroni/patroni/pull/3612
```suggestion
    max_time = time.monotonic() + int(time_limit)
    actual = None
    while time.monotonic() < max_time:
```

features/dynamic_synchronized_standby_slots.feature
```diff
@@ -0,0 +1,91 @@
...
+    When I add the table dyn_replicate1 to postgres-0
+    And I get all changes from logical slot test_dyn_sync on postgres-0
+    Then postgres-1 has a synced replication slot named test_dyn_sync with the test_decoding plugin after 60 seconds
+    When I add the table dyn_replicate2 to postgres-0
+    And I get all changes from logical slot test_dyn_sync on postgres-0
+    Then postgres-1 has a synced replication slot named test_dyn_sync with the test_decoding plugin after 30 seconds
+    When I add the table dyn_replicate3 to postgres-0
+    And I get all changes from logical slot test_dyn_sync on postgres-0
+    Then postgres-1 has a synced replication slot named test_dyn_sync with the test_decoding plugin after 30 seconds
+
```
CyberDem0n: It is unrelated to this PR and already tested here: https://github.com/patroni/patroni/blob/5ae67a7b845b017631996b29bc7df09b9771374d/features/ignored_slots.feature#L68-L91
```suggestion
```

tests/test_sync.py
```diff
@@ -195,6 +195,64 @@ def value_in_conf():
         mock_reload.assert_called()
         self.assertEqual(value_in_conf(), "synchronous_standby_names = '\"a-1\"'")
 
+        # PG17+ with dynamic_synchronized_standby_slots: ensure a manual reload happens
+        # when only synchronized_standby_slots changes (synchronous_standby_names unchanged).
+        from unittest.mock import PropertyMock
+
+        from patroni.postgresql import PostgresqlState
```
CyberDem0n: Please move imports to the top of the file.

docs/dynamic_configuration.rst
```diff
@@ -85,6 +85,24 @@ In order to change the dynamic configuration you can use either :ref:`patronictl
    -  **database**: the database name (when matching a ``logical`` slot).
    -  **plugin**: the logical decoding plugin (when matching a ``logical`` slot).
 
+-  **dynamic\_synchronized\_standby\_slots**: (PostgreSQL 17+ only) When enabled, Patroni automatically manages the ``synchronized_standby_slots`` parameter to keep it in sync with ``synchronous_standby_names``. This ensures that logical replication slots with ``failover=true`` are synchronized to the same physical standbys used for synchronous replication. Defaults to ``false`` (opt-in).
```
thedodd: A bit of a nit, but I think naming this `manage_synchronized_standby_slots` makes the intent of this parameter a bit more clear. It also more naturally matches the description / docs which you already have in place.

NBD either way.

PavelZaytsev: Whatever @CyberDem0n prefers to call this feature

thedodd: NBD either way.

PavelZaytsev:
@CyberDem0n @hughcapet please review when you get the chance.

patroni/ha.py
```diff
@@ -274,6 +274,31 @@ def __init__(self, patroni: Patroni):
         # used only in backoff after failing a pre_promote script
         self._released_leader_key_timestamp = 0
 
+        # Track the last known state of dynamic_synchronized_standby_slots feature to detect toggles
+        self._last_dynamic_sync_slots_enabled: Optional[bool] = None
+
+    def _handle_dynamic_sync_slots_toggle(self, sync_members: Collection[str]) -> bool:
```
hughcapet: please add the `:param:  sync_members` and return value description to the docsting

patroni/postgresql/config.py
```diff
@@ -1326,18 +1336,27 @@ def reload_config(self, config: Dict[str, Any], sighup: bool = False) -> None:
...
-            if self._postgresql.state == PostgresqlState.RUNNING:
+                self._server_parameters[name] = value
+            if reload and self._postgresql.state == PostgresqlState.RUNNING:
                 self.write_postgresql_conf()
                 self._postgresql.reload()
             return True
+        return False
+
+    def set_synchronous_standby_names(self, value: Optional[str]) -> Optional[bool]:
+        """Updates synchronous_standby_names and reloads if necessary.
```
hughcapet, line 1361: please add the `value` param desc

patroni/postgresql/slots.py
```diff
@@ -204,6 +204,20 @@ def _query(self, sql: str, *params: Any) -> List[Tuple[Any, ...]]:
         """
         return self._postgresql.query(sql, *params, retry=False)
 
+    def update_synchronized_standby_slots(self, sync_members: Collection[str], reload: bool = False) -> bool:
```
hughcapet: please write a proper docstring

patroni/postgresql/config.py
```diff
@@ -1326,18 +1336,27 @@ def reload_config(self, config: Dict[str, Any], sighup: bool = False) -> None:
 
         self._postgresql.set_pending_restart_reason(param_diff)
 
-    def set_synchronous_standby_names(self, value: Optional[str]) -> Optional[bool]:
-        """Updates synchronous_standby_names and reloads if necessary.
-        :returns: True if value was updated."""
-        if value != self._server_parameters.get('synchronous_standby_names'):
+    def _set_server_parameter(self, name: str, value: Optional[str], reload: bool = False) -> bool:
+        """Update a server parameter and optionally write config and reload PostgreSQL."""
```
hughcapet: please write a proper docstring

patroni/ha.py
```diff
@@ -274,6 +274,31 @@ def __init__(self, patroni: Patroni):
...
+            return False
+
+        logger.info("dynamic_synchronized_standby_slots changed to %s, updating slots", feature_enabled)
+        if feature_enabled:
+            self.state_handler.slots_handler.update_synchronized_standby_slots(sync_members, reload=True)
+        else:
+            # Feature was disabled - restore the user-configured value (or clear if none configured),
+            # mirroring the pattern used for synchronous_standby_names in disable_synchronous_replication.
+            value = self.state_handler.config.synchronized_standby_slots
+            self.state_handler.config.set_synchronized_standby_slots(value, reload=True)
```
hughcapet:
```suggestion
            self.state_handler.config.set_synchronized_standby_slots(self.state_handler.config.synchronized_standby_slots, reload=True)
```

patroni/postgresql/config.py
PavelZaytsev: fixed

patroni/postgresql/slots.py
PavelZaytsev: fixed

patroni/postgresql/config.py
PavelZaytsev: fixed

patroni/ha.py
PavelZaytsev: fixed

patroni/ha.py
```diff
@@ -274,6 +274,36 @@ def __init__(self, patroni: Patroni):
         # used only in backoff after failing a pre_promote script
         self._released_leader_key_timestamp = 0
 
+        # Track the last known state of dynamic_synchronized_standby_slots feature to detect toggles
+        self._last_dynamic_sync_slots_enabled: Optional[bool] = None
+
+    def _handle_dynamic_sync_slots_toggle(self, sync_members: Collection[str]) -> bool:
```
hughcapet:
```suggestion
    def _handle_dynamic_sync_slots_toggle(self, sync_members: CaseInsensitiveSet) -> bool:
```

patroni/postgresql/slots.py
```diff
@@ -204,6 +204,32 @@ def _query(self, sql: str, *params: Any) -> List[Tuple[Any, ...]]:
...
+                              ``synchronized_standby_slots`` GUC is cleared.
+        :param reload: whether to write ``postgresql.conf`` and reload PostgreSQL if the value changed.
+
+        :returns: ``True`` if the value of ``synchronized_standby_slots`` was updated, ``False`` otherwise.
+        """
+        if self._postgresql.major_version < 170000:
+            return False
+
+        if not global_config.dynamic_synchronized_standby_slots_enabled:
+            return False
```
hughcapet:
```suggestion
        if self._postgresql.major_version < 170000 or not global_config.dynamic_synchronized_standby_slots_enabled:
            return False
```

patroni/postgresql/config.py
```diff
@@ -1326,18 +1336,38 @@ def reload_config(self, config: Dict[str, Any], sighup: bool = False) -> None:
...
+    def set_synchronous_standby_names(self, value: Optional[str]) -> Optional[bool]:
+        """Updates synchronous_standby_names and reloads if necessary.
+
+        :param value: new value of ``synchronous_standby_names``, or ``None`` to remove it from the config.
+
+        :returns: ``True`` if value was updated, ``None`` otherwise.
+        """
+        return self._set_server_parameter('synchronous_standby_names', value, reload=True) or None
+
+    def set_synchronized_standby_slots(self, value: Optional[str], reload: bool = False) -> bool:
```
hughcapet, line 1369: docstring here as well, please

tests/test_ha.py
```diff
@@ -2051,6 +2053,54 @@ def test__process_quorum_replication(self):
         self.assertEqual(mock_set_sync.call_count, 1)
         self.assertEqual(mock_set_sync.call_args_list[0][0], ('ANY 1 (foo)',))
 
+    def test__handle_dynamic_sync_slots_toggle(self):
+        mock_update_slots = self.p.slots_handler.update_synchronized_standby_slots = Mock()
+        mock_set_slots = self.p.config.set_synchronized_standby_slots = Mock()
```
hughcapet:
```suggestion
    @patch.object(SlotsHandler, 'update_synchronized_standby_slots')
    @patch.object(ConfigHandler, 'set_synchronized_standby_slots')
    def test__handle_dynamic_sync_slots_toggle(self, mock_set_slots, mock_update_slots):
```

tests/test_ha.py
```diff
@@ -2051,6 +2053,54 @@ def test__process_quorum_replication(self):
...
+
+        # No change - should return False
+        self.ha._last_dynamic_sync_slots_enabled = False
+        with patch.object(type(global_config), 'dynamic_synchronized_standby_slots_enabled',
+                          new_callable=PropertyMock, return_value=False):
```
hughcapet:
```suggestion
        with patch.object(global_config.__class__, 'dynamic_synchronized_standby_slots_enabled',
                          PropertyMock(return_value=False)):
```

here and in other places just to be consistent with other tests

tests/test_postgresql.py
```diff
@@ -839,6 +839,57 @@ def test_get_server_parameters(self):
...
+
+            # Value cleared -> parameter removed even if present in user config.
+            self.p.config._server_parameters.pop('synchronized_standby_slots', None)
+            config_with_user_value = {'parameters': {**config['parameters'],
+                                                     'synchronized_standby_slots': 'user_slot'},
+                                      'listen': '0'}
+            params = self.p.config.get_server_parameters(config_with_user_value)
+            self.assertNotIn('synchronized_standby_slots', params)
+
+    def test_set_synchronized_standby_slots(self):
```
hughcapet: there is no need in this test. you are testing the general underlying _set_server_parameter function's logic. If you want, you can change it to test__set_server_parameter but i think it is already covered by the test_set_sync_standby

tests/test_slots.py
```diff
@@ -491,6 +491,60 @@ def test__drop_incorrect_failover_synced_slots(self):
...
+            slots = self.p.config._server_parameters['synchronized_standby_slots']
+            self.assertIn('node1', slots)
+            self.assertIn('node2', slots)
+
+        # Test: Feature enabled with reload - should trigger reload
+        with patch.object(Postgresql, 'major_version', 170000), \
+             patch.object(type(global_config), 'dynamic_synchronized_standby_slots_enabled',
+                          new_callable=PropertyMock, return_value=True):
+            self.p.config._server_parameters.pop('synchronized_standby_slots', None)
+            self.assertTrue(self.s.update_synchronized_standby_slots({'node3'}, reload=True))
```
hughcapet, line 537: please also assert reload is actually called

tests/test_postgresql.py
```diff
@@ -839,6 +839,57 @@ def test_get_server_parameters(self):
...
+        # Setting None again when already absent should return False
+        self.assertFalse(self.p.config.set_synchronized_standby_slots(None))
+
+        # Setting with reload=True when running should trigger reload
+        with patch.object(Postgresql, 'state', PropertyMock(return_value=PostgresqlState.RUNNING)), \
+             patch.object(Postgresql, 'reload', Mock()) as mock_reload:
+            self.assertTrue(self.p.config.set_synchronized_standby_slots('slot1', reload=True))
+            mock_reload.assert_called_once()
+
+    def test_synchronized_standby_slots_property(self):
```
hughcapet: no need in this test

tests/test_sync.py
```diff
@@ -195,6 +195,61 @@ def value_in_conf():
...
+                          new_callable=PropertyMock, return_value=True), \
+            patch.object(type(self.p), 'state',
+                         new_callable=PropertyMock,
+                         return_value=PostgresqlState.RUNNING):
+            self.s.set_synchronous_standby_names(CaseInsensitiveSet(['n1']))
+        # SSN didn't change, but slots did, so a manual reload is expected.
+        mock_reload.assert_called()
+        self.assertEqual(self.p.config._server_parameters.get('synchronized_standby_slots'), 'n1')
+
+        # Regression: member names containing characters that require quoting (e.g. dashes)
```
hughcapet, line 224: in this test we should only check update_synchronized_standby_slots, reload and write_postgresql_conf calls fact. how update_synchronized_standby_slots behaves you already tested in test_update_synchronized_standby_slots

patroni/postgresql/slots.py
hughcapet: maybe we also implement Postgresql.supports_synchronized_standby_slots property and use it here and in ConfigHandler.get_server_parameters?

hughcapet (changes requested):
we also need to extend validator.py for the new configuration parameter

docs/dynamic_configuration.rst
hughcapet: i vote for manage_synchronized_standby_slots

PavelZaytsev:
@hughcapet addressed
@CyberDem0n 
plz review

patroni/ha.py
```diff
@@ -274,6 +274,36 @@ def __init__(self, patroni: Patroni):
...
+
+        logger.info("manage_synchronized_standby_slots changed to %s, updating slots", feature_enabled)
+        if feature_enabled:
+            self.state_handler.slots_handler.update_synchronized_standby_slots(sync_members, reload=True)
+        else:
+            # Feature was disabled - restore the user-configured value (or clear if none configured),
+            # mirroring the pattern used for synchronous_standby_names in disable_synchronous_replication.
+            value = self.state_handler.config.synchronized_standby_slots
+            self.state_handler.config.set_synchronized_standby_slots(value, reload=True)
+        self._last_manage_sync_slots_enabled = feature_enabled
```
hughcapet: why do not we do this assignment after every set_synchronized_standby_slots? (e.g., in recover, demote, process_*_prepromote)
if it should be done after every update_synchronized_standby_slots, maybe it makes sense to move this property to SyncHandler and save the last value inside the set_synchronized_standby_slots itself?

patroni/postgresql/slots.py
```diff
@@ -204,6 +204,30 @@ def _query(self, sql: str, *params: Any) -> List[Tuple[Any, ...]]:
...
+
+        :param sync_members: set of currently active synchronous standby member names. If empty or
+                              containing ``*`` (any standby is allowed to be synchronous), the
+                              ``synchronized_standby_slots`` GUC is cleared.
+        :param reload: whether to write ``postgresql.conf`` and reload PostgreSQL if the value changed.
+
+        :returns: ``True`` if the value of ``synchronized_standby_slots`` was updated, ``False`` otherwise.
+        """
+        if not self._postgresql.supports_synchronized_standby_slots or \
+                not global_config.manage_synchronized_standby_slots_enabled:
```
hughcapet: what if `manage_synchronized_standby_slots` is set from `on` to `off` and we call `set_synchronous_standby_names` because of some transition. This way, the condition here will evaluate to False and we never reset `synchronized_standby_slots` because we set `_last_manage_sync_slots_enabled` to False and `_handle_manage_sync_slots_toggle` won't be able to catch the switch

patroni/ha.py
PavelZaytsev: Moved the bookkeeping into `SlotsHandler.update_synchronized_standby_slots` itself, so it updates on every call instead of being stamped from scattered `Ha` call sites. Removed `_handle_manage_sync_slots_toggle` and all the ad-hoc `_last_manage_sync_slots_enabled` assignments in `ha.py` (a2b4993c).

patroni/postgresql/slots.py
PavelZaytsev: Confirmed, that was a real bug, fixed by moving the toggle detection into this method itself: it now checks the enabled→disabled transition on every call regardless of caller, and restores the user-configured value when it just turned off, so `set_synchronous_standby_names` from any transition catches it (a2b4993c).

patroni/postgresql/slots.py
```diff
@@ -204,6 +206,45 @@ def _query(self, sql: str, *params: Any) -> List[Tuple[Any, ...]]:
         """
         return self._postgresql.query(sql, *params, retry=False)
 
+    def update_synchronized_standby_slots(self, sync_members: Collection[str], reload: bool = False) -> bool:
```
CyberDem0n: IMO, it would make sense to have a dedicated method handle_manage_sync_slots_toggle(), which will call update_synchronized_standby_slots() only if feature was just enabled/disabled. As @hughcapet correctly noticed, we also need to call it from Ha._handle_synchronous_strict_mode().

Besides that, it would be valuable to also pass `num` (similar to set_synchronous_standby_names()).
In a future it will allow to quickly adapt this method to new semantic of `synchronized_standby slots` (`ANY N (...)`).

Last, but not least, IMO `synchronized_standby` should follow the same rules for synchronous_mode_strict=on, that is, set it to `__patroni_strict_sync_replica_placeholder__` if there are no synchronous nodes available, and `num` parameter is also required for it.

CyberDem0n (commented):
I played a bit with it locally and did a few changes to make it work with sychronous_mode_strict.
Please see the diff:
```diff
diff --git a/patroni/ha.py b/patroni/ha.py
index 230752bf..92379be7 100644
--- a/patroni/ha.py
+++ b/patroni/ha.py
@@ -860,7 +860,7 @@ class Ha(object):
 
         # manage_synchronized_standby_slots depends on synchronous_mode, so when sync mode is off
         # the feature effectively goes off too - restore the user-configured value if needed.
-        self.state_handler.slots_handler.update_synchronized_standby_slots(CaseInsensitiveSet(), reload=True)
+        self.state_handler.slots_handler.handle_manage_sync_slots_toggle(CaseInsensitiveSet())
 
     def _handle_synchronous_strict_mode(self, dcs_state: SyncState, replication_state: Any) -> bool:
         """Handle strict synchronous mode.
@@ -892,9 +892,12 @@ class Ha(object):
                     numsync != replication_state.numsync or \
                     sync_type != replication_state.sync_type:
                 self.state_handler.sync_handler.set_synchronous_standby_names(voters, numsync)
-            elif voters:
-                msg = 'Continue using old value of synchronous_standby_names="{0}". '.format(
-                    self.state_handler.synchronous_standby_names())
+            else:
+                if voters:
+                    msg = 'Continue using old value of synchronous_standby_names="{0}". '.format(
+                        self.state_handler.synchronous_standby_names())
+
+                self.state_handler.slots_handler.handle_manage_sync_slots_toggle(voters, numsync)
 
             # We use self._synchronous_strict_mode_activated to show warning only once.
             if not self._synchronous_strict_mode_activated:
@@ -968,7 +971,7 @@ class Ha(object):
             if transition != 'restart' or _check_timeout(1):
                 # Check if manage_synchronized_standby_slots feature state changed even when no sync changes
                 if transition == 'break':
-                    self.state_handler.slots_handler.update_synchronized_standby_slots(sync_state.active, reload=True)
+                    self.state_handler.slots_handler.handle_manage_sync_slots_toggle(sync_state.active)
                 return
             # synchronous_standby_names was transitioned from empty to non-empty and it may take
             # some time for nodes to become synchronous. In this case we want to restart state machine
@@ -1019,7 +1022,7 @@ class Ha(object):
                 self.state_handler.sync_handler.set_synchronous_standby_names(picked)
             else:
                 # Check if manage_synchronized_standby_slots feature state changed
-                self.state_handler.slots_handler.update_synchronized_standby_slots(picked, reload=True)
+                self.state_handler.slots_handler.handle_manage_sync_slots_toggle(picked)
             return
 
         # update synchronous standby list in dcs temporarily to point to common nodes in current and picked
diff --git a/patroni/postgresql/slots.py b/patroni/postgresql/slots.py
index 62ad04ed..b72139c2 100644
--- a/patroni/postgresql/slots.py
+++ b/patroni/postgresql/slots.py
@@ -19,6 +19,7 @@ from ..psycopg import OperationalError
 from ..tags import Tags
 from .connection import get_connection_cursor
 from .misc import format_lsn, fsync_dir, PostgresqlRole
+from .sync import SYNC_STRICT_PLACEHOLDER
 
 if TYPE_CHECKING:  # pragma: no cover
     from psycopg import Cursor
@@ -206,7 +207,29 @@ class SlotsHandler:
         """
         return self._postgresql.query(sql, *params, retry=False)
 
-    def update_synchronized_standby_slots(self, sync_members: Collection[str], reload: bool = False) -> bool:
+    def handle_manage_sync_slots_toggle(self, sync_members: Collection[str], num: Optional[int] = None) -> None:
+        """Handle ``manage_synchronized_standby_slots`` feature toggle.
+
+        Detects edges between the feature being enabled and disabled and either pushes
+        a fresh dynamic value (when toggled on) or restores the user-configured value
+        (when toggled off).
+
+        :param sync_members: currently active synchronous standby members, used to compute
+                              the dynamic value of ``synchronized_standby_slots`` when the
+                              feature is being toggled on.
+        :param num: number of nodes to sync to, forwarded to :meth:`update_synchronized_standby_slots`.
+                    It is set only when quorum commit is enabled.
+        """
+        if not self._postgresql.supports_synchronized_standby_slots:
+            return
+
+        if self._last_manage_sync_slots_enabled == global_config.manage_synchronized_standby_slots_enabled:
+            return
+
+        self.update_synchronized_standby_slots(sync_members, num=num, reload=True)
+
+    def update_synchronized_standby_slots(self, sync_members: Collection[str],
+                                          num: Optional[int] = None, reload: bool = False) -> bool:
         """Update ``synchronized_standby_slots`` based on sync members for PostgreSQL 17+.
 
         Converts *sync_members* to their corresponding slot names and writes them to the
@@ -218,9 +241,11 @@ class SlotsHandler:
         This way a toggle-off is never missed, even when it is observed from a call that is otherwise
         a no-op (e.g. *sync_members* unchanged).
 
-        :param sync_members: set of currently active synchronous standby member names. If empty or
-                              containing ``*`` (any standby is allowed to be synchronous), the
-                              ``synchronized_standby_slots`` GUC is cleared.
+        :param sync_members: set of currently active synchronous standby member names.
+                             If it contains ``*`` (any standby is allowed to be synchronous), or is empty
+                             while *num* is >= 1, the ``synchronized_standby_slots`` GUC is set to the
+                             strict sync placeholder. If empty and *num* is not set, the GUC is cleared.
+        :param num: number of nodes to sync to. It is set only when quorum commit is enabled.
         :param reload: whether to write ``postgresql.conf`` and reload PostgreSQL if the value changed.
 
         :returns: ``True`` if the value of ``synchronized_standby_slots`` was updated, ``False`` otherwise.
@@ -239,11 +264,17 @@ class SlotsHandler:
         if not feature_enabled:
             return False
 
-        if not sync_members or '*' in sync_members:
-            return self._postgresql.config.set_synchronized_standby_slots(None, reload=reload)
+        # Special case. If sync_members contains the '*' wildcard, or the set is empty while the
+        # requested num of sync nodes is >= 1, we want to set synchronized_standby_slots to
+        # '__patroni_strict_sync_replica_placeholder__'
+        sync_strict = '*' in sync_members or num and num >= 1 and not sync_members
+        if sync_strict:
+            slot_names = [SYNC_STRICT_PLACEHOLDER]
+        else:
+            slot_names = [slot_name_from_member_name(member) for member in sorted(sync_members) if member != '*']
 
-        slot_names = [slot_name_from_member_name(member) for member in sync_members]
-        return self._postgresql.config.set_synchronized_standby_slots(','.join(sorted(slot_names)), reload=reload)
+        sync_param = None if not slot_names else ','.join(slot_names)
+        return self._postgresql.config.set_synchronized_standby_slots(sync_param, reload=reload)
 
     def copy_slot_items(self, src: Dict[str, Any], dst: Dict[str, Any], is_logical: bool = True) -> None:
         """Select values from *src* slots dictionary to update in *dst* slots dictionary.
diff --git a/patroni/postgresql/sync.py b/patroni/postgresql/sync.py
index 7106cc2c..727a2402 100644
--- a/patroni/postgresql/sync.py
+++ b/patroni/postgresql/sync.py
@@ -421,18 +421,17 @@ END;$$""")
         :param sync: set of nodes to sync to
         :param num: specifies number of nodes to sync to. The *num* is set only in case if quorum commit is enabled
         """
+        # Update synchronized_standby_slots in _server_parameters first (no reload yet) so
+        # that the subsequent set_synchronous_standby_names() reload picks up both changes.
+        slots_changed = self._postgresql.slots_handler.update_synchronized_standby_slots(sync, num)
+
         # Special case. If sync nodes set is empty but requested num of sync nodes >= 1
         # we want to set synchronous_standby_names to '__patroni_strict_sync_replica_placeholder__'
         sync_strict = '*' in sync or num and num >= 1 and not sync
-        # Keep the original (unquoted) member names for synchronized_standby_slots, which uses
-        # slot_name_from_member_name() and mustn't see the output of quote_standby_name() or
-        # the SYNC_STRICT_PLACEHOLDER. When sync_strict we treat it like the '*' wildcard for
-        # slots purposes so update_synchronized_standby_slots clears the value.
-        sync_members = ['*'] if sync_strict else sorted(sync)
         if sync_strict:
             sync = [SYNC_STRICT_PLACEHOLDER]
         else:
-            sync = [quote_standby_name(x) for x in sync_members]
+            sync = [quote_standby_name(x) for x in sorted(sync)]
 
         if self._postgresql.supports_multiple_sync and len(sync) > 1:
             if num is None:
@@ -448,10 +447,6 @@ END;$$""")
         if sync_param is not None:
             logger.info("Assigning synchronous_standby_names to %s", sync_param)
 
-        # Update synchronized_standby_slots in _server_parameters first (no reload yet) so
-        # that the subsequent set_synchronous_standby_names() reload picks up both changes.
-        slots_changed = self._postgresql.slots_handler.update_synchronized_standby_slots(sync_members)
-
         # Update synchronous_standby_names - this will reload if SSN changed,
         # which also applies any pending synchronized_standby_slots change.
         ssn_changed = self._postgresql.config.set_synchronous_standby_names(sync_param)
diff --git a/tests/test_slots.py b/tests/test_slots.py
index a8b887f9..a06bf138 100644
--- a/tests/test_slots.py
+++ b/tests/test_slots.py
@@ -511,11 +511,11 @@ class TestSlotsHandler(BaseTestPostgresql):
             self.assertFalse(self.s.update_synchronized_standby_slots(set()))
             self.assertNotIn('synchronized_standby_slots', self.p.config._server_parameters)
 
-        # Test: Feature enabled - wildcard clears the parameter (no-op when already absent)
+        # Test: Feature enabled - wildcard enables strict mode
         with patch.object(Postgresql, 'major_version', 170000), \
              patch.object(global_config.__class__, 'manage_synchronized_standby_slots_enabled',
                           PropertyMock(return_value=True)):
-            self.assertFalse(self.s.update_synchronized_standby_slots({'*'}))
+            self.assertTrue(self.s.update_synchronized_standby_slots({'*'}))
 
         # Test: Feature enabled - specific members sets the parameter
         with patch.object(Postgresql, 'major_version', 170000), \
diff --git a/tests/test_sync.py b/tests/test_sync.py
index 90db27b7..fc36ec43 100644
--- a/tests/test_sync.py
+++ b/tests/test_sync.py
@@ -216,7 +216,7 @@ class TestSync(BaseTestPostgresql):
                              Mock(return_value=True)) as mock_update_slots, \
                 patch.object(ConfigHandler, 'write_postgresql_conf', Mock()) as mock_write_conf:
             self.s.set_synchronous_standby_names(CaseInsensitiveSet(['n1']))
-        mock_update_slots.assert_called_once_with(['n1'])
+        mock_update_slots.assert_called_once_with(CaseInsensitiveSet(['n1']), None)
         # SSN didn't change, but slots did, so a manual reload (and config write) is expected.
         mock_write_conf.assert_called_once()
         mock_reload.assert_called()
@@ -232,7 +232,7 @@ class TestSync(BaseTestPostgresql):
                 patch.object(SlotsHandler, 'update_synchronized_standby_slots',
                              Mock(return_value=False)) as mock_update_slots:
             self.s.set_synchronous_standby_names(CaseInsensitiveSet(['postgres-1', 'postgres-2']))
-        mock_update_slots.assert_called_once_with(['postgres-1', 'postgres-2'])
+        mock_update_slots.assert_called_once_with(CaseInsensitiveSet(['postgres-1', 'postgres-2']), None)
         # synchronous_standby_names is built from QUOTED names, so we expect them quoted here:
         self.assertEqual(self.p.config._server_parameters.get('synchronous_standby_names'),
                          '2 ("postgres-1","postgres-2")')
@@ -247,7 +247,7 @@ class TestSync(BaseTestPostgresql):
                 patch.object(SlotsHandler, 'update_synchronized_standby_slots',
                              Mock(return_value=False)) as mock_update_slots:
             self.s.set_synchronous_standby_names(CaseInsensitiveSet(['postgres-1', 'postgres-2']), 1)
-        mock_update_slots.assert_called_once_with(['postgres-1', 'postgres-2'])
+        mock_update_slots.assert_called_once_with(CaseInsensitiveSet(['postgres-1', 'postgres-2']), 1)
         self.assertEqual(self.p.config._server_parameters.get('synchronous_standby_names'),
                          'ANY 1 ("postgres-1","postgres-2")')
 
@@ -268,7 +268,7 @@ class TestSync(BaseTestPostgresql):
         self.assertEqual(self.p.config._server_parameters.get('synchronized_standby_slots'), 'n1')
 
         # Feature toggled off without going through a dedicated toggle-check call site.
-        self.p.config._config.setdefault('parameters', {})['synchronized_standby_slots'] = 'user_slot'
+        self.p.config._config['parameters'] = {'synchronized_standby_slots': 'user_slot'}
         with patch.object(global_config.__class__, 'manage_synchronized_standby_slots_enabled',
                           PropertyMock(return_value=False)):
             self.s.set_synchronous_standby_names(CaseInsensitiveSet(['n1', 'n2']))
```

patroni/postgresql/slots.py
PavelZaytsev: Resolved

PavelZaytsev:
@CyberDem0n thanks, the `synchronous_mode_strict` handling works well now.

One semantics question about the strict branch in `SlotsHandler.update_synchronized_standby_slots` (patroni/postgresql/slots.py). When strict mode is on with no eligible standby, we set `synchronized_standby_slots` to `__patroni_strict_sync_replica_placeholder__`, mirroring `synchronous_standby_names`.

For `synchronous_standby_names` the placeholder is an unsatisfiable value that blocks commits — the intended strict behavior. But `synchronized_standby_slots` has different semantics: it is a list of *existing physical slots* that failover-enabled logical slots must wait on. A name with no backing slot makes PG17 log repeated `replication slot "..." does not exist` warnings and halt advancement of logical slots, on a slot that will never be created.

Since there is no standby to protect in that state anyway, would it be more correct to **clear** `synchronized_standby_slots` in the strict/`*` case (so logical decoding is not gated on a phantom slot) rather than write the placeholder? Or is halting logical-slot advancement here intentional?

hughcapet:
> Since there is no standby to protect in that state anyway, would it be more correct to clear synchronized_standby_slots in the strict/* case (so logical decoding is not gated on a phantom slot) rather than write the placeholder? Or is halting logical-slot advancement here intentional?

The whole idea of `synchronized_standby_slots` is to prevent logical replication slots from sending changes  that are not yet replicated to the given standbys. When there is no sync standby, and strict mode is enabled, changes will still be sent to a logical replica even though the commit is only committed locally on primary and is waiting for confirmation.

But it is true it is a bit ugly to see so many warnings, so it is worth mentioning this in the documentation

hughcapet:
@PavelZaytsev can you please add a note about the mentioned expected behavior (warning in the postgres logs) to the docs, so that we can merge this PR?

docs/dynamic_configuration.rst
```diff
@@ -85,6 +85,24 @@ In order to change the dynamic configuration you can use either :ref:`patronictl
...
+
+   .. note::
+       This feature requires ``synchronous_mode: true`` or ``synchronous_mode: quorum`` to be enabled, as it manages which standbys receive synchronized logical slots based on the current synchronous replication topology.
+
+   .. note::
+       In ``quorum`` mode Patroni lists **all** sync standbys in ``synchronized_standby_slots``. Because PostgreSQL blocks logical slot advancement until **every** listed physical slot has caught up, logical replication may wait longer than the commit semantics require (which only need ``synchronous_node_count`` standbys to confirm). This is a conservative default that guarantees no data loss for logical replication; a permanently lagging standby will however permanently block logical slot advancement.
+
+   .. note::
+       When this feature is disabled, Patroni restores ``synchronized_standby_slots`` to whatever value the user has configured under ``postgresql.parameters`` (or removes the parameter entirely if none is set).
```
le-duane, line 97:
```suggestion
   .. note::
       When this feature is disabled, Patroni restores ``synchronized_standby_slots`` to whatever value the user has configured under ``postgresql.parameters`` (or removes the parameter entirely if none is set).

   .. note::
       When ``synchronous_mode_strict`` is enabled and no synchronous standby is currently
       available, Patroni sets ``synchronized_standby_slots`` to the same internal placeholder
       used for ``synchronous_standby_names`` (``__patroni_strict_sync_replica_placeholder__``).
       Since no replication slot with that name exists, PostgreSQL will repeatedly write the
       following to the logs::

           WARNING:  replication slot "__patroni_strict_sync_replica_placeholder__" specified in parameter "synchronized_standby_slots" does not exist
           DETAIL:  Logical replication is waiting on the standby associated with replication slot "__patroni_strict_sync_replica_placeholder__".
           HINT:  Create the replication slot "__patroni_strict_sync_replica_placeholder__" or amend parameter "synchronized_standby_slots".

       This is expected. It prevents logical replication slots from advancing until a
       synchronous standby is available, so that logical subscribers never receive changes
       ahead of physical replication confirming them. The warnings stop as soon as a
       synchronous standby reconnects.

```

le-duane (commented):
Hope this helps unblock! We're really looking forward to this feature. Thanks!

docs/dynamic_configuration.rst
```diff
@@ -85,6 +85,42 @@ In order to change the dynamic configuration you can use either :ref:`patronictl
...
+-  **manage\_synchronized\_standby\_slots**: (PostgreSQL 17+ only) When enabled, Patroni automatically manages the ``synchronized_standby_slots`` parameter to keep it in sync with ``synchronous_standby_names``. This ensures that logical replication slots with ``failover=true`` are synchronized to the same physical standbys used for synchronous replication. Defaults to ``false`` (opt-in).
+
+   .. note::
+       This feature requires ``synchronous_mode: true`` or ``synchronous_mode: quorum`` to be enabled, as it manages which standbys receive synchronized logical slots based on the current synchronous replication topology.
+
+   .. note::
+       In ``quorum`` mode Patroni lists **all** sync standbys in ``synchronized_standby_slots``. Because PostgreSQL blocks logical slot advancement until **every** listed physical slot has caught up, logical replication may wait longer than the commit semantics require (which only need ``synchronous_node_count`` standbys to confirm). This is a conservative default that guarantees no data loss for logical replication; a permanently lagging standby will however permanently block logical slot advancement.
+
+   .. note::
+   .. note::
```
CyberDem0n, line 96:
```suggestion
```

CyberDem0n (approved):

CyberDem0n (approved):

hughcapet (approved):

PR #3528, opened 2026-01-20
Fix link for `slots` configuration in existing_data.rst

zaneduffield:
The link was to the YAML configuration, but `use_slots` and `slots` are dynamic configuration items.


zaneduffield:
From what I can tell, `postgresql.use_slots` is only available in DCS configuration.
https://github.com/patroni/patroni/blob/b3462eb4e9182f548de18ddebeeddd601e9ba74c/docs/dynamic_configuration.rst?plain=1#L35

Issue #3524, opened 2025-12-29
Feature: role-based pg_hba rules + postgres parameters


wjmorland:
### What happened?

Would like to open a feature request/proposal for adding role-based configurations within Patroni. This was previously brought up in #1925 - my use case is the same, defining different pg_hba rules on primary and replica nodes. As suggested in that issue, we currently handle this by using a custom callback script which swaps out pg_hba files based on the role, which works ok, but requires us to 1) manage pg_hba rules separately than we do the rest of our Patroni/Postgres config and 2) keep hba definitions in sync between hosts manually. I think there is some value in having this configuration live in the DCS as it improves auditability/makes it easier to keep things in sync. 

Furthermore, think there could be some value in extending this to other postgres parameters so they could be set differently on the primary vs the replica. For example, I could imagine it being useful to turn on more detailed logging on just replicas, or add additional tracking like `track_io_timing` which has some overhead which we may not want on the primary. Although probably want to restrict this to dynamic parameters only.

Just some initial brainstorming to help picture what this would look like, was thinking this could be implemented by either of these options:
1. defining role inline for each pg_hba rule/parameter, where there's an attribute for `role` which can be `any` (default if omitted), `primary`, or `replica`. So something like: 
```
postgresql:
  pg_hba:
  # gets applied to all nodes
  - host all all 127.0.0.1/32 md5
  # only applied to primaries
  - host all all 1.2.3.4/32 md5
    role: primary
  # only applied to replicas
  - host all all 1.2.3.4/24 md5
    role: replica
```
2. defining role based sections, with the same structure as `postgresql`. This is one way of doing it, or could have single `role_based` section with `primary` and `replica` subsection:
```
postgresql:
  pg_hba:
  # gets applied to all nodes
  - host all all 127.0.0.1/32 md5
  
postgresql_primary: 
  pg_hba:
  # only applied to primaries
  - host all all 1.2.3.4/32 md5

postgresql_replica: 
  pg_hba:
  # only applied to replicas
  - host all all 1.2.3.4/24 md5
```

Please let me know if this seems reasonable, I'm happy to contribute a patch, just wanted to get some feedback on the interface before spending time on it. Also sorry for opening this as a bug report, could not figure out how to create a normal issue 😅 

### How can we reproduce it (as minimally and precisely as possible)?

N/A

### What did you expect to happen?

N/A

### Patroni/PostgreSQL/DCS version

N/A

### Patroni configuration file

```yaml
N/A
```

[...]

PR #3530, opened 2026-01-22
Support for role-based configuration (#3524)

wjmorland:
This commit adds support for specifying role-based configuration for postgresql parameters, pg_hba, and pg_ident settings. These can be specified by appending `_primary`, `_replica`, or `_standby_leader` to the end of the configuration name (ex. `pg_hba_primary`).

For postgresql params, the role-based configs will merge with the existing config, with the role-based config taking precedence. For pg_hba and pg_ident, it will be a full replacement without merging.

This refactors the Config class to allow for passing in a role parameter when reloading/writing configs. The Config class then handles evaluation and implementation of role-based rules.

This also updates the HA and Postgresql init code to apply the role based configs when first initializing and when a node is promoted or demoted. If role-based configs are used, promotion will now trigger an additional SIGHUP sent to the Postgres process. Demotion already reloads configs, so the logic is mostly unchanged.

----

This is my first attempt at an implementation for role-based params, as proposed in #3524. 

I've done some basic manual testing and functionally, it seems to be working as I was expecting with updating the configs during role changes. However, I'm not 100% confident on this implementation from the maintainability/cleanliness point of view. It seems somewhat brittle to have to remember to call `apply_role_based_config_if_needed` in `ha.py`, and it doesn't feel quite right to be passing in role as a parameter in so many places. I did consider the alternative of trying to always call this after a role change is applied, but that felt even more intrusive (i.e after a promotion and role is set to primary, feels like `promote()` should immediately return, not do extra work). Would appreciate any ideas on a better approach or ways to simplify this!


wjmorland:
Thanks for getting back so quick, and really appreciate such clear, detailed feedback! 

That all makes sense, I'll try the new approach and update the PR when I have it working

Issue #3532, opened 2026-01-27
Compress option should be customisable in basebackup create replica method


zaneduffield:
### What happened?

The `compress` option cannot be customised in the `basebackup` options section.
This is the code that blocks it:
https://github.com/patroni/patroni/blob/28f00cc1da53c447a49ee8c5891a3b1561d2edfb/patroni/postgresql/bootstrap.py#L336-L337
which has been a blocked option since the original PR that added support for customising the basebackup options:
https://github.com/patroni/patroni/blob/4ce539ba1b729e96404d0c28582c77bcf9fc593a/patroni/postgresql.py#L1628-L1629

This error comes up in the logs when you try and use it:
> ERROR: compress option for basebackup is not allowed

I don't see why patroni can't allow the use of the `compress` option; `pg_basebackup` will decompress the files before writing them to the destination when using the plain format
From the [docs](https://www.postgresql.org/docs/18/app-pgbasebackup.html):
> When the plain format is used, client-side compression may not be specified, but it is still possible to request server-side compression. If this is done, the server will compress the backup for transmission, and the client will decompress and extract it.

If the network link is slow, compressing the backup server side may save a lot of time in replica creation.

I found this old issue on the topic: https://github.com/patroni/patroni/issues/2527.
This was quoted
> Not all parameters make sense, if you override a connection string or provide an option to created tar-ed or compressed base backups, patroni won’t be able to make a replica out of it.

but I don't understand what problem patroni could have with a 'compressed base backup', because `pg_basebackup` handles the decompression of the files transparently.

### How can we reproduce it (as minimally and precisely as possible)?

With this added to some valid patroni configuration
```yaml
postgresql:
  basebackup:
    - compress: 'server-zstd'
  create_replica_methods:
    - basebackup
```

attempt to create or reinitialise a replica

### What did you expect to happen?

The replica to be created by transferring compressed data using pg_basebackup.

### Patroni/PostgreSQL/DCS version

-   Patroni version: 4.1.0
-   PostgreSQL version: 17.7
-   DCS (and its version): n/a

### Patroni configuration file

```yaml
postgresql:
  basebackup:
    - compress: 'server-zstd'
  create_replica_methods:
    - basebackup
```

### patronictl show-config

```yaml
n/a
```

### Patroni log files

```shell
ERROR: compress option for basebackup is not allowed
ERROR: Error when parsing basebackup key-value option server-zstd: only one key-value is allowed and value should be a string
```

### PostgreSQL log files

```shell
[...]

PR #3537, opened 2026-01-28
Fix: validate user options in dict format for initdb/basebackup

### LINKED ISSUE #3533: User options for initdb and basebackup sections are only validated when passed as a list
### What happened?

You can bypass the validation of user options for the `initdb` and `pg_basebackup` commands by using a dictionary instead of a list in the configuration.

The inconsistency is here
https://github.com/patroni/patroni/blob/28f00cc1da53c447a49ee8c5891a3b1561d2edfb/patroni/postgresql/bootstrap.py#L93-L114
in the first branch, the validation function is never called.

Add this to your config
```yaml
postgresql:
  basebackup:
    compress: 'server-zstd'
  create_replica_methods:
    - basebackup
```

and observe that patroni fails to validate and reject the `compress` option (in #3532 I argue that this option should be valid), whereas if you were to specify the config as

```yaml
postgresql:
  basebackup:
    - compress: 'server-zstd'
  create_replica_methods:
    - basebackup
```
then patroni would reject the `compress` option.

I expected consistent parameter validation between the two methods.

-   Patroni version: 4.1.0
-   PostgreSQL version: 17.7

```yaml
postgresql:
  basebackup:
    compress: 'server-zstd'
  create_replica_methods:
    - basebackup
```

### patronictl show-config

```yaml
n/a
```

### Patroni log files

```shell
n/a
```

### PostgreSQL log files

```shell
n/a
```

- [x] Yes

_No response_

m4rrypro:
Fixes #3533

## Problem
When `initdb` or `basebackup` options are provided as a dict (instead of a list), the `option_is_allowed()` validation was bypassed, allowing blocked options like `compress` to be used.

## Solution
Added the `option_is_allowed(key)` check to the dict branch in `process_user_options()`.

## Testing
- Added test case for dict-format validation with blocked options
- All 717 tests pass
- flake8 and pyright pass
image
image
image

### DIFF
--- a/tests/test_bootstrap.py
+++ b/tests/test_bootstrap.py
@@ -108,24 +108,6 @@ class TestBootstrap(BaseTestPostgresql):
                 ['pg_basebackup', f'--pgdata={self.p.data_dir}', '-X', 'stream', '--dbname=', '--foo=bar'],
             )
 
-        # Test compress option: server-side compression should be allowed (issue #3532)
-        with patch('patroni.postgresql.bootstrap.logger.debug') as mock_debug:
-            self.p.cancellable.reset_is_cancelled()
-            self.b.basebackup("", None, {'compress': 'server-zstd'})
-            mock_debug.assert_called_with(
-                'calling: %r',
-                ['pg_basebackup', f'--pgdata={self.p.data_dir}', '-X', 'stream', '--dbname=', '--compress=server-zstd'],
-            )
-
-        # Test compress option: client-side compression should be rejected
-        with patch('patroni.postgresql.bootstrap.logger.error') as mock_error:
-            self.p.cancellable.reset_is_cancelled()
-            self.b.basebackup("", None, {'compress': 'gzip'})
-            mock_error.assert_called_with(
-                'compress option for basebackup must use server-side compression '
-                '(e.g., server-gzip, server-zstd). Client-side compression is not allowed.'
-            )
-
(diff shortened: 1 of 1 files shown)


patroni/postgresql/bootstrap.py
```diff
@@ -259,8 +259,20 @@ def basebackup(self, conn_url: str, env: Dict[str, str], options: Dict[str, Any]
...
+        for opt in user_options:
+            if opt.startswith('--compress='):
+                if opt.startswith('--compress=server'):
+                    validated_options.append(opt)
+                else:
+                    logger.error('compress option for basebackup must use server-side compression '
+                                 '(e.g., server-gzip, server-zstd). Client-side compression is not allowed.')
+            else:
+                validated_options.append(opt)
+        user_options = validated_options
```
CyberDem0n: I have feeling that this change shouldn't be a part of this PR.

m4rrypro: You're right, that code was added in the wrong PR. The compress validation is part of #3538 (fix for #3532), not this one.
This PR (#3537) only contains the fix for dict validation in `process_user_options()` in `utils.py`.

m4rrypro: Done. The branch now only contains changes to `patroni/utils.py` and `tests/test_utils.py` - no bootstrap.py changes. The compress validation code is in PR #3538.

PR #3538, opened 2026-01-28
Allow server-side compression for basebackup option

### LINKED ISSUE #3532: Compress option should be customisable in basebackup create replica method

m4rrypro:
Fixes #3532

## Problem
The `compress` option was completely blocked for basebackup, but since PostgreSQL 15, server-side compression is useful and works transparently with plain format.

## Solution
- Removed `compress` from blocked options list
- Added validation to allow only `--compress=server*` values (e.g., server-zstd, server-gzip)
- Reject client-side compression with helpful error message

## Testing
- Added tests for both valid and invalid compress options
- All 717 tests pass
- flake8 passes

### DIFF
--- a/tests/test_ha.py
+++ b/tests/test_ha.py
@@ -189,6 +189,7 @@ def run_async(self, func, args=()):
     'Latest checkpoint location': '0/12345678',
     "Latest checkpoint's TimeLineID": '2'}))
 @patch.object(SlotsHandler, 'load_replication_slots', Mock(side_effect=Exception))
+@patch.object(ConfigHandler, 'pg_version', PropertyMock(return_value=180000))
 @patch.object(ConfigHandler, 'append_pg_hba', Mock())
 @patch.object(ConfigHandler, 'write_pgpass', Mock(return_value={}))
 @patch.object(ConfigHandler, 'write_recovery_conf', Mock())


patroni/postgresql/bootstrap.py
```diff
@@ -259,8 +259,20 @@ def basebackup(self, conn_url: str, env: Dict[str, str], options: Dict[str, Any]
...
+        for opt in user_options:
+            if opt.startswith('--compress='):
+                if opt.startswith('--compress=server'):
+                    validated_options.append(opt)
+                else:
+                    logger.error('compress option for basebackup must use server-side compression '
+                                 '(e.g., server-gzip, server-zstd). Client-side compression is not allowed.')
+            else:
+                validated_options.append(opt)
+        user_options = validated_options
```
CyberDem0n: We need to do version dependent check.
You can get version from `self._postgresql.config.pg_version`.

PR #3540, opened 2026-02-05
allow etcd password to be parameterized

AndrewJackson2020:
This PR allows username and password to be specified as command line arguments for patronictl.

Currently the `--dcs-url` argument allows you to specify an etcd url but there is no way to supply a username and password. This means that the user basically just needs to add this in a config file which is not ideal if your goal is to not leave passwords lying around in plain text files unnecessarily. 

The implementation here adds 2 additional CLI arguments: `--etcd-username` and `--etcd-password`. I was originally going to name them `--dcs-username` and `--dcs-password` but at first glance it doesn't seem that this concept applies to any of the other DCS types. Alternatively this could be implemented in the dcs URL: thinking of a postgres type format like `etcd3://username:password@host:port/namespace`. If this is a better format and we dont want to add another 2 arguments that only apply to etcd I would be happy to rework this patch in order to accomodate that.

Thank you.

### DIFF
--- a/tests/test_ctl.py
+++ b/tests/test_ctl.py
@@ -65,11 +65,11 @@ class TestCtl(unittest.TestCase):
     def test_load_config(self, mock_logger_debug):
         runner = CliRunner()
         with runner.isolated_filesystem():
-            self.assertRaises(PatroniCtlException, load_config, './non-existing-config-file', None)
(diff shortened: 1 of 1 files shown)


AndrewJackson2020:
Closing this PR because I opened another one at #3541 that I think is a better implementation.

PR #3541, opened 2026-02-05
feat: add username/password parse to dcs-url

AndrewJackson2020:
This PR allows username and password to be specified as command line arguments for patronictl.

Currently the --dcs-url argument allows you to specify an etcd url but there is no way to supply a username and password. This means that the user basically just needs to add this in a config file which is not ideal if your goal is to not leave passwords lying around in plain text files unnecessarily.

This is an alternative implementation of the PR at #3540.

The implementation here uses the existing urlparse capabilities and allows the username and password to be input in a way similar to postgres connection strings (ie scheme://username:password@hostname:port/path). I think that this implementation is cleaner because it does not add any new CLI arguments and ends up only being 4 lines of code. 

I could possibly add validation to ensure that username/password is not set for schemes that do not support this concept, let me know if you think that is necessary.  

Thank you.

### DIFF
--- a/docs/patronictl.rst
+++ b/docs/patronictl.rst
@@ -54,7 +54,7 @@ Before jumping into each of the sub-commands of ``patronictl``, be aware that ``
 
     This argument can be used either to override the DCS and ``namespace`` settings from the ``patronictl`` configuration, or to define it if it's missing in the configuration.
 
-    The value should be in the format ``DCS://HOST:PORT/NAMESPACE``, e.g. ``etcd3://localhost:2379/service`` to connect to etcd v3 running on ``localhost`` with Patroni cluster stored under ``service`` namespace. Any part that is missing in the argument value will be replaced with the value present in the configuration or with its default.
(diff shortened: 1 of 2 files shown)


AndrewJackson2020:
> TBH, I wish `--dcs-url` should have never existed and don't see any good reasons in enhancing its usage.

Understood. If that is the case I'm fine closing this PR.

Would you be interested in any implementation that would allow configuration to be set/overridden via CLI flag/env var? Wondering if you want patroni be a config file driven application exclusively.

CyberDem0n:
Env variables are already quite well supported, please check docs: https://patroni.readthedocs.io/en/latest/ENVIRONMENT.html#etcd

AndrewJackson2020:
> Env variables are already quite well supported, please check docs: https://patroni.readthedocs.io/en/latest/ENVIRONMENT.html#etcd

Apologies, totally missed that. Closing PR, thank you.

PR #3542, opened 2026-02-09
allow "patronictl history" to display multiple clusters

AndrewJackson2020:
This PR allows the `patronictl history` command to take multiple cluster_names and display the history for all supplied cluster_names. This coincides with how the `patronictl topology` and `patronictl list` commands work.  Also I do not believe that it introduces any breaking changes because while some internal variable names are changed, from the callers perspective any old invocation of `patronictl history` should be backwards compatible with the new implementation.

Being able to view the history of multiple clusters is very convenient while troubleshooting dcs issues.

One thing that is a bit awkward about this implementation is that it prints out one table per cluster instead of one table with an extra cluster column. Despite this it is consistent with how `patronictl topology` and `patronictl list` function so I think having it function in the same way as those commands makes the most sense for this PR

### DIFF
--- a/patroni/ctl.py
+++ b/patroni/ctl.py
@@ -2266,10 +2266,10 @@ def version(cluster_name: str, group: Optional[int], member_names: List[str]) ->
 
 
 @ctl.command('history', help="Show the history of failovers/switchovers")
-@arg_cluster_name
+@click.argument('cluster_names', nargs=-1)
 @option_default_citus_group
 @option_format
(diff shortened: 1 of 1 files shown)


CyberDem0n (commented):
1. This change breaks existing behavior - `history` doesn't require specifying cluster name and takes one from the config
2. `history` is updated only eventually doesn't really help to troubleshoot DCS outages.

PR #3549, opened 2026-02-26
synchronous replica switchover feature

Fingolfin-Anekane:
## Hi, I would like to propose a new synchronous replica switchover feature. 

As master switchover it might be helpfull to control cluster topology. Feature is desined as much simillar to master switchover as possible.

### Why is it needed:
- to easily handle replicas affinity issues 
- to avoid hacks with manually set `synchronous_standby_names`

### Key changes:
- **API**: new `/sync_switchover` handler. 
- **ETCD**: API request stores new config key `sync_switchover` 
- **daemon**: handels `sync_switchover` key and performs synchronous replica switchover reusing existed methods
- **patronictl**: provides cli interface for the command

Guess tests and documentation are also needed, but want to discuss the feature first. Feature was tested manually and now succesfully works in environment with lots of clusters.

### DIFF
--- a/patroni/api.py
+++ b/patroni/api.py
@@ -1264,6 +1264,120 @@ class RestApiHandler(BaseHTTPRequestHandler):
         """
         self.do_POST_failover(action='switchover')
 
+    def poll_sync_switchover_result(self, candidates: Optional[List[str]]) -> Tuple[int, str]:
+        """Poll sync switchover operation until it finishes or times out.
+
+        :param candidates: names of the Patroni node to be promoted to synchronous replica.
+
+        :returns: a tuple composed of 2 items:
+
+            * Response HTTP status codes:
+
+                * ``200``: if the operation succeeded; or
+                * ``503``: if the operation failed or timed out.
+
+            * A status message about the operation.
+
+        """
+        timeout = max(10, self.server.patroni.dcs.loop_wait)
+        for _ in range(0, timeout * 2):
+            time.sleep(1)
+            try:
+                cluster = self.server.patroni.dcs.get_cluster()
+                sync = cluster.sync.sync_standby.split(',')
+                if sync and set(sync) == set(candidates):
+                    return 200, f'Successfully switched synchronous replicas to {candidates}'
+                if not cluster.sync_switchover:
+                    return 503, 'sync switchover failed'
+            except Exception as e:
+                logger.debug('Exception occurred during polling sync switchover result: %s', e)
+        return 503, 'sync switchover status unknown'
+
+    @check_access
+    def do_POST_sync_switchover(self) -> None:
+        """Handle a ``POST`` request to ``/sync_switchover`` path.
+        """
+        request = self._read_json_content()
+        if not request:
+            return
+
+        cluster = self.server.patroni.dcs.get_cluster()
+        leader = request.get('leader')
+        scheduled_at = request.get('scheduled_at')
+
+        err = self.validate_sync_switchover(request)
+        if err:
+            self.write_response(400, err)
+            return
+
+        candidates = list(set(request.get('candidates')))
+
+        if scheduled_at:
+            status_code, err, scheduled_at = self.parse_schedule(scheduled_at, 'sync_switchover')
+            if err:
+                self.write_response(status_code if status_code else 400, err)
+                return
+
+        err = self.is_sync_switchover_possible(cluster, leader, candidates)
+        if err:
+            self.write_response(412, err)
+            return
+
+        if self.server.patroni.dcs.manual_switch_sync(leader, candidates, scheduled_at):
+            self.server.patroni.ha.wakeup()
+            if scheduled_at:
+                self.write_response(202, 'sync switchover scheduled')
+                return
+            code, body = self.poll_sync_switchover_result(candidates)
+            self.write_response(code, body)
+            return
+        else:
+            self.write_response(503, 'failed to write sync_switchover key into DCS')
+            return
+
+    @staticmethod
+    def validate_sync_switchover(request: Dict) -> Optional[str]:
+
+        leader = request.get('leader')
+        candidates = request.get('candidates')
+
+        if not leader or type(leader) is not str:
+            return 'Sync switchover could be performed only from a specific leader'
+        if not candidates:
+            return 'Sync switchover could be performed only to a specific candidates'
+        if type(candidates) is not list:
+            if type(candidates) is str:
+                request['candidates'] = [candidates]
+            else:
+                return 'Sync switchover candidates type unknown'
+        if leader in candidates:
+            return 'Leader cannot be chosen as a synchronous replica'
+
+
+    @staticmethod
+    def is_sync_switchover_possible(cluster: Cluster, leader: Optional[str],
+                                    candidates: Optional[List[str]]) -> Optional[str]:
+        config = global_config.from_cluster(cluster)
+        if leader and (not cluster.leader or cluster.leader.name != leader):
+            return 'leader name does not match'
+        if config.is_paused:
+            # conservative way
+            return 'Sync switchover is unavailable is paused state'
+        if not config.is_synchronous_mode:
+            return 'Synchronous mode disabled'
+        if config.is_quorum_commit_mode:
+            return 'Sync switchover is unavailable is quorum commit mode'
+
+        if config.synchronous_node_count != len(candidates):
+            return (f'Synchronous node count is {config.synchronous_node_count}'
+                    f' but requested count is  {len(candidates)}.')
+
+        allowed_candidates = [str(m.name) for m in cluster.members if m.name != leader and not m.nofailover and
+                                not m.nosync and not m.nostream and not m.noloadbalance]
+        for candidate in candidates:
+            if candidate not in allowed_candidates:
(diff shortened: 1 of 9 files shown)


CyberDem0n (commented):
Sorry, but I am failing to understand what problem are you trying to solve

Fingolfin-Anekane:
> Sorry, but I am failing to understand what problem are you trying to solve

I want to add a new API handler which allows to explicitly set desired replica to synchrous standby.

CyberDem0n:
> I want to add a new API handler which allows to explicitly set desired replica to synchrous standby.

Sorry, but it still doesn't explain which problem you are trying to solve.

gsmolk:
> Sorry, but it still doesn't explain which problem you are trying to solve.

If cluster operates in multi-dc environment, it is desirable to provide a means for a user to set sync role to some member, located in another DC, so the loss of both master and sync standby can be avoided.

Fingolfin-Anekane:
ty @gsmolk 
that is what I wanted to say :)

CyberDem0n:
Please be more specific and provide concrete examples.

Fingolfin-Anekane:
Sure! Let's assume we have a patroni cluster
```
+ Cluster: exampledb (7551778309504793718) --------+--------------+--------------+---------------------+----+---------------+--------+---------------+--------+----------------------------+
| Member                                           | Host         | Role         | State               | TL |   Receive LSN |    Lag |    Replay LSN |    Lag | Tags                       |
+--------------------------------------------------+--------------+--------------+---------------------+----+---------------+--------+---------------+--------+----------------------------+
| test-host-3.zone_2.eu                | 1.1.2.3   | Replica      | streaming           | 39 | 1A11/95006910 |      0 |       unknown |        |                            |
+--------------------------------------------------+--------------+--------------+---------------------+----+---------------+--------+---------------+--------+----------------------------+
| test-host-1.zone_1.eu                | 1.1.2.1   | Leader       | running             | 39 |               |        |               |        |                            |
+--------------------------------------------------+--------------+--------------+---------------------+----+---------------+--------+---------------+--------+----------------------------+
| test-host-2.zone_1.eu 			   | 1.1.2.2   | Sync Standby | streaming           | 39 | 1A11/95006910 |      0 | 1A11/95006910 |      0 |                            |
+--------------------------------------------------+--------------+--------------+---------------------+----+---------------+--------+---------------+--------+----------------------------+
```
Where `zone_1` represents locality (server, datacenter, ...) for the replicas. In case when  `zone_1` is unavaliable, we might lose some data because `leader` and `sync` replicas were in the same locality. 

That's why we want to avoid `sync` and `leader` replicas being located in the same `zone_1`. To make that we want to set `test-host-3.zone_2.eu` as a `sync` replica. And this feature will allow to do that by simple API call instead of manually setting `synchronous_standby_names` to `test-host-3.zone_2.eu` and waiting while patroni will detect missconfiguration with dcs and patch it accordingly.

In case with 3 replicas and 2 zones it might be easy to set `nosync` tag to the  `test-host-2.zone_1.eu`. Yet it will be much more complicated to handle such affinity issues on bigger clusters

CyberDem0n:
Well, and how your PR helps with it?
It is better to make Patroni zone aware (zone could be added to tags).
When picking nodes for sync replication it should prefer nodes in different AZs.

What is important to note, it will help only with non-quorum replication configurations.
The problem is that synchronous_standby_names syntax is not flexible enough to describe complex typologies.
Example: `ANY 2 (ANY 1 (group1_1, group1_2, ...), ANY 1 (group2_1, group2_2, ...), ANY 1 (group3_1, group3_2, ...))`

Fingolfin-Anekane:
> Well, and how your PR helps with it?

I think I got you right. In this PR I'm not trying to manage those locality issues by patroni itself. I want to add functionality for DB administrators or for external automatic to request those sync switchovers by API and patroni only performs them. As I sad it works queit same with master switchover allowing DBA to set the exact topology

>  it will help only with non-quorum replication configurations

Yep, thats true. For quorum commit clusters an error will be returned
https://github.com/patroni/patroni/pull/3549/changes#diff-13329cd3342d1757c2c9b4dd7a26c0170f4deee5d32eff5c2fd719cc9ff9375bR1368

CyberDem0n:
IMO - what you do contradicts current Patroni goals.
We are supposed to automate synchronous based failver by selecting the best sync nodes and managing sychronous_standby_names and DCS /sync key.
What you suggest to do - set sycnhronous_standby_names to some fixed value, which could be quickly outdated.
Effectively it will break everything until the new value is set.
Even if it works for you, because you have a ton of orchestration on top - people will misuse it and complain.

On top that, I really dislike introducing new keys in DCS. One should really good reason for doing it. In 99.99% case existing keys could be used to store additional information.

PS. Making Patroni zone-aware will require order of magnitude less changes and will just work.

Fingolfin-Anekane:
Well, I got your point with zone-aware approach, will take a look on it.

However, can we discuss your points about the current changes.

> What you suggest to do - set sycnhronous_standby_names to some fixed value

That's what patroni is doing. The only difference is that we enforce particular candidate (or candidates) to `sync` role. Then the flow is simmilar to the regular patroni sync replication processing.

> which could be quickly outdated

Why should it be outdated? Patroni won't switch sync replica unless it violates lag constraint or replica is unavaliable

quote from comments
> _Current synchronous standby is always preferred, unless it has disconnected or does not want to be a
        synchronous standby any longer._ 


> Effectively it will break everything until the new value is set.

Didn't understand here. What and why should it break if appropriate replica was promoted to sync role by new functional and then it's gonna be maintained by HA loop

gsmolk:
> Even if it works for you, because you have a ton of orchestration on top - people will misuse it and complain.

But the same way people can misuse `nosync` tag or failover/switchover, any tool can used in harmful ways

CyberDem0n:
nosync tag doesn't introduce 500+ new LOC and new keys in DCS

yuvarajsm14:
We also faced the same issue with our postgres implementation [ we wanted sync-standby not to be in the same zone as our primary ]. With the help of the nosync tag, we could make it work. 

You can use a combination of zone identifiers on the nodes and the nosync tag to ensure that sync-standby and primary never sit in the same zone.

Fingolfin-Anekane:
This approach requires a lot of automation on top of patroni to manage tags. Besides logic inside this automation will be a complex one (you don't want to accidentally forget to unset `nosync` tag and for sure there will be a lot of corner cases for particular implementations). IMO - its much easier to delegate such task to patroni and only check the results.

Another case I forgot to mention is when you want to switch async replica to master (whatever reason you have).  For master switchover replica should be sync standby and there should be a way to perform that. This operation might even violate the zone constraint we discussed before because the current master and the desired master might be in the same zone.

Example:
You have a cluster, and for some reason you need to replace `host1.zone1` to the different server in the same zone (what ever reason)
```
host1.zone1 - master
host2.zone2 - sync
host3.zone3 - replica
```
You create another replica `host4.zone1` and want to promote it to master role and then terminate previous master `host1.zone1`.
```
host1.zone1 - master
host2.zone2 - sync
host3.zone3 - replica
host4.zone1 - replica
```
Now somehow you need to promote `host4.zone1` to sync standby and currently there is no way to do that unless managing `nosync` tags for the other replicas which I think is excessively complex. Zone tags approach will not work here because there will be a moment when we violate zone constraint.
```
host1.zone1 - master
host2.zone2 - replica
host3.zone3 - replica
host4.zone1 - sync
```
and then
```
host1.zone1 - replica
host2.zone2 - sync
host3.zone3 - replica
host4.zone1 - master
```

yuvarajsm14:
For sure, there are multiple scenarios that we are needed to be taken care of. For instance: 

- What if there is only one zone left
- What if synchronous_node_count is changed 
and many more, but I believe that is doable.

Also, for the example given, I believe you can 

1. Change the current sync-standby, i.e. host2.zone2, as master
2. Add one more node in zone1
3. By that time, one of the nodes in zone 1 will become a sync-standby [ given the solution you wrote around nosync works ]
4. Then make the sync-standby node in zone1 as master.

Fingolfin-Anekane:
> Change the current sync-standby, i.e. host2.zone2, as master

It should work however it seems that there should be a much simplier flow. Unfortunately it is not acceptable in my situation because it requires to switch master 2 times and it will cause a lot of `25006` (read-only error) to users. Intention is to make cluster maitenance as much invisible for users as it is possible.

yuvarajsm14:
> Unfortunately it is not acceptable in my situation because it requires to switch master 2 times and it will cause a lot of `25006` (read-only error) to users.

That is true. But in my case, I will not do switchover multiple times, as I already have a master running [zone2]. If having master in a specific zone is a hard requirement for your solution, then it is different.

Fingolfin-Anekane:
> If having master in a specific zone is a hard requirement for your solution, then it is different

It is. Master, sync and async roles usually consume different amounts of cpu and disk i/o, so in my particular case we have loadbalance automation controlling that stuff. We want to fully control cluster topology to put instances with higher resource consumption on fitting servers. Plus as I said there are affinity constraints on the top of that.

Guess I'm not the only one who faced such issues, that is why I've disided to contribute this patch.

hughcapet:
we are currently working on the zone/site-aware feature

PR #3551, opened 2026-02-27
Add DEBUG logging for pg_rewind decisions

### LINKED ISSUE #3544: Lack of logging around ability to use `pg_rewind`
### What happened?

When testing pg_rewind (related to #3543), I found that the logging provided around the decisions involving pg_rewind to be insufficient.

In particular, there is no logging to indicate any of the following error cases
1. pg_rewind is not being used because the command is failing or cannot be found (e.g. because of PATH issues or a configured `bin_name.pg_rewind` settings)
2. pg_rewind is not being used because `wal_log_hints` is not enabled and data checksums are not in use

What tends to happen in these scenarios is the replica starts normally as a standby, but fails to start streaming because it's on the wrong timeline. If `remove_data_directory_on_rewind_failure` and/or `remove_data_directory_on_diverged_timelines` are set, then patroni could act differently, of course.

I took a look at the relevant code in `rewind.Rewind.can_rewind`
https://github.com/patroni/patroni/blob/b247bfac17176b99bd24603c8d8aa28added2d02/patroni/postgresql/rewind.py#L49-L64

Logging could easily be added for the different error cases, but it would be pretty noisy, because this function is called many times. The function could be memoized, but that would probably create more bugs that it is worth. The logging could be demoted to the DEBUG level, making the duplication less important, but that would make them less useful.

1. start two-node cluster
  2. configure a custom `bin_path.pg_rewind` script that doesn't exist, or throws an error
  3. make sure `use_pg_rewind` is enabled
4. stop patroni on one node
5. `pg_ctl start && pg_ctl promote && pg_ctl stop` for the stopped node
6. start patroni on the node
7. Observe no logging about pg_rewind, and postgres errors like "highest timeline 4 of the primary is behind recovery timeline 5"

I expected logging from patroni indicating that pg_rewind wasn't used, and why.

-   Patroni version: 4.1.0
-   PostgreSQL version: 17.7

```yaml
n/a
```

### patronictl show-config

```yaml
n/a
```

### Patroni log files

```shell
n/a
```

### PostgreSQL log files

```shell
n/a
```

- [x] Yes

_No response_

marceloneppel:
When pg_rewind cannot be used, Patroni provides no indication of why. The replica silently fails to stream on the wrong timeline, and the user has to figure out on their own whether the issue is a missing binary, disabled use_pg_rewind, or lack of wal_log_hints/data checksums.

Added `can_rewind_reason()` method that returns a human-readable explanation. The `can_rewind` property now delegates to it, avoiding code duplication. DEBUG-level messages are logged at three decision points:

- in `trigger_check_diverged_lsn()` when diverged timeline check is skipped because neither pg_rewind nor `remove_data_directory_on_diverged_timelines` is available, including the specific reason why pg_rewind cannot be used — this is the main silent failure scenario from the issue
- in `_handle_rewind_or_reinitialize()` when rewind is needed but not possible
- in `execute()` when `can_rewind` flips to False after a failed rewind attempt

Close #3544

### DIFF
--- a/patroni/ha.py
+++ b/patroni/ha.py
@@ -601,6 +601,7 @@ class Ha(object):
 
             msg = 'running pg_rewind from ' + leader.name
             return self._async_executor.try_run_async(msg, self._rewind.execute, args=(leader,)) or msg
+
         if self._rewind.should_remove_data_directory_on_diverged_timelines and not self.is_standby_cluster():
             msg = 'reinitializing due to diverged timelines'
             return self._async_executor.try_run_async(msg, self._do_reinitialize, args=(self.cluster,)) or msg
--- a/patroni/postgresql/rewind.py
+++ b/patroni/postgresql/rewind.py
@@ -57,7 +57,7 @@ class Rewind(object):
 
         cmd = [self._postgresql.pgcommand('pg_rewind'), '--help']
         try:
-            ret = subprocess.call(cmd, stdout=subprocess.DEVNULL, stderr=subprocess.STDOUT)
+            ret = subprocess.call(cmd, stdout=open(os.devnull, 'w'), stderr=subprocess.STDOUT)
             if ret != 0:  # pg_rewind is not there, close up the shop and go home
                 logger.debug('pg_rewind is not possible: pg_rewind command returned non-zero exit code %s', ret)
                 return False


zaneduffield:
Thanks for working on this @marceloneppel!

I wonder whether the log level should be higher than DEBUG, but I can see how it could be perceived as noisy. However if the user has set `use_pg_rewind: true`, then I think it's safe to assume that they are interested in seeing all logs about why it couldn't be used. I'm not sure that it would make sense to log a message at different levels depending on configuration, I'm just trying to motivate why the user might expect to see the messages.

marceloneppel:
Thanks for the review @zaneduffield!

I agree that WARNING would make more sense when the user has explicitly set `use_pg_rewind: true` — if they opted in and it can't work, they should know about it. There's actually precedent for conditional log levels in the codebase, e.g. in `patroni/watchdog/base.py`:

```python
log = logger.warning if self.config.mode == MODE_REQUIRED else logger.debug
log("Could not activate %s: %s", self.impl.describe(), e)
```

We could do something similar here:

```python
log = logger.warning if self.enabled else logger.debug
log('not checking diverged timeline: ...')
```

However, @CyberDem0n mentioned in #3544 that the preference is for DEBUG level. So I'd like to keep it at DEBUG for now unless they approve the conditional approach.

If we do go with WARNING (or conditional WARNING/DEBUG), I'm also planning to address the noise concern you raised in the original issue ("this function is called many times"). The approach would be to add a `REWIND_STATUS.BLOCKED` state so that the message is logged only once — when transitioning to the blocked state — rather than on every HA loop iteration. Subsequent calls skip the log since `_state` is already `BLOCKED`. The state naturally resets when conditions change (e.g. if the user enables checksums) or after `reset_state()`.

This would also require changing the `executed` property from `self._state > REWIND_STATUS.NOT_NEED` to `self._state in (REWIND_STATUS.SUCCESS, REWIND_STATUS.FAILED)` to avoid the new enum value being incorrectly treated as "executed". I've verified this is the only ordering-based comparison on `REWIND_STATUS` in the codebase — all others use `==`, `!=`, or `in`.

Issue #3536, opened 2026-01-28
Instantaneous  403 Permission Denied caused primary node demote


XiuhuaRuan:
### What happened?

When Patroni received a 403 Permission Denied, the primary node immediately demoted. However,  the permission was restored within one second, the node promoted itself again, resulting in unnecessary demotion and promotion. 

### How can we reproduce it (as minimally and precisely as possible)?

This issue occurs occasionally, and the reproduction conditions are unknown.

### What did you expect to happen?

In Kubernetes environments, RBAC may occasionally experience intermittent unavailability due to various reasons. Could we implement retry logic upon receiving 403 errors to avoid these unnecessary demotion?

### Patroni/PostgreSQL/DCS version

-   Patroni version: 4.0.5
-   PostgreSQL version: 12
-   DCS (and its version): etcd-3.5.9

### Patroni configuration file

```yaml
bootstrap:
  dcs:
    postgresql:
      use_pg_rewind: true
restapi:
  connect_address: '10.40.0.3:8008'
postgresql:
  connect_address: '10.40.0.3:5866'
  database: postgres
  listen: 0.0.0.0:5432
  port: 5432
  logging_collector: on
  authentication:
    superuser:
      username:  xxx
      password: xxx
    replication:
      username:  xxx
      password: xxx
    rewind:
      username: xxx
      password: xxx
   
 tags:
    nofailover: false
    noloadbalance: false
    clonefrom: false
```

### patronictl show-config

```yaml
postgresql:
  use_pg_rewind: true
```

### Patroni log files

```shell
2025-12-13 03:50:43,055 ERROR: Permission denied
Traceback (most recent call last):
  File "patroni/dcs/kubernetes.py", line 1152, in _update_leader_with_retry
  File "patroni/dcs/kubernetes.py", line 1090, in _patch_or_create
  File "patroni/dcs/kubernetes.py", line 1149, in _retry
  File "patroni/utils.py", line 589, in __call__
  File "patroni/dcs/kubernetes.py", line 558, in wrapper
  File "patroni/dcs/kubernetes.py", line 479, in wrapper
  File "patroni/dcs/kubernetes.py", line 447, in call_api
  File "patroni/dcs/kubernetes.py", line 266, in _handle_server_response
patroni.dcs.kubernetes.K8sClient.rest.ApiException: (403)
Reason: Forbidden
[...]

PR #3559, opened 2026-03-15
add retry for 403 Permission denied when update leader

XiuhuaRuan:
**Problem Description：**
When the Kubernetes API returns a 403 Permission Denied error (e.g., due to temporary RBAC permission loss), the current code immediately logs an exception and returns False, which may cause the leader to be incorrectly demoted. However, in real-world scenarios, permission issues can be temporary (such as RBAC updates, network fluctuations), and the application should be given an opportunity to recover within a timeout period.

**Solution：**
Add a dedicated _handle_permission_denied method that, when encountering a 403 error:
1. Continuously verifies the leader status within the retry_timeout period
2. Checks the leader object every 0.5 seconds to confirm whether the current instance is still the leader
3. Returns one of three states based on the verification result:
'retry': Still the leader, continue retrying the update operation
'demote': Leadership has been lost, demote immediately（return False）
'timeout': Unable to confirm within the timeout period, handle as timeout（return False）

This PR is to fix issue #3536 .


XiuhuaRuan:
@CyberDem0n Alex，thanks for your review and suggestion. I understand your point. I also noticed the retriable_http_codes earlier, but considering that 403 may only occurs on this node, so adding confirmation whether it's still the leader before retrying update. However, even if this situation occurs, the retrying update operation won't succeed anyway, so your suggestion is indeed more concise and reasonable. I will modify the corresponding code, add some unit tests, perform some testing, and submit after the tests pass.

hughcapet:
@XiuhuaRuan please unarchive this PR when you think it is ready for review

XiuhuaRuan:
> @XiuhuaRuan please un-archive this PR when you think it is ready for review

@hughcapet Yes, It's ready for review. Thanks.

hughcapet (approved):

PR #3564, opened 2026-03-17
Improve switchover unknown-status messaging (#743)

### LINKED ISSUE #743: Improve error message on switchover
On a high load cluster, the shutdown of the master may take longer than patronictl waits.

This yields the following error:

```
Switchover failed, details: 503, Switchover status unknown
```

In fact at this state the master is still in shutdown mode and the switch over is still happening once shut down. Not yet sure what to write, but "failed" sounds like nothing is going to happen and master is fine.

m4rrypro:
## Summary
- clarify API timeout response for failover/switchover polling by including timeout duration and noting the operation may still be in progress
- adjust `patronictl` output to report `result unknown` (instead of `failed`) when the API returns a `status unknown` detail
- add/adjust tests in `tests/test_api.py` and `tests/test_ctl.py` for the updated messaging path

## Testing
- `.venv/bin/pytest tests/test_api.py::TestRestApiHandler::test_do_POST_switchover tests/test_ctl.py::TestCtl::test_switchover`

Fixes #743

### DIFF
--- a/patroni/ctl.py
+++ b/patroni/ctl.py
@@ -1399,10 +1399,8 @@ def _do_failover_or_switchover(action: str, cluster_name: str, group: Optional[i
             click.echo('{0} {1}'.format(timestamp(), r.data.decode('utf-8')))
         else:
             details = r.data.decode('utf-8')
-            message = '{0} failed, details: {1}, {2}'
-            if 'status unknown' in details.lower():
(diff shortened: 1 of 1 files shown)


m4rrypro:
This is a message-only behavior clarification; it does not change switchover/failover election logic.

PR #3566, opened 2026-03-18
Support role-based replicatefrom fallback (#422)

### LINKED ISSUE #422: replicatefrom role, rather than replicatefrom node
Hello,

Ideally I'd like to replicatefrom a standby (any standby) and if that doesn't work out bug the master.

Any thoughts on if this would be a good change to implement?

Cheers,
James

m4rrypro:
## Summary
- extend `replicatefrom` handling to accept role-style values (`any`, `standby`, `replica`) and pick an available standby source
- add fallback to the leader when the requested `replicatefrom` member is unavailable
- document the new `replicatefrom` behavior in YAML configuration docs
- add HA tests for both standby-role selection and leader fallback paths

## Testing
- `.venv/bin/pytest tests/test_ha.py`

Fixes #422

### DIFF
--- a/patroni/ha.py
+++ b/patroni/ha.py
@@ -730,8 +730,9 @@ class Ha(object):
         elif self.patroni.replicatefrom and self.patroni.replicatefrom != self.state_handler.name:
             if self.patroni.replicatefrom in ('any', 'standby', 'replica'):
                 leader_name = cluster.leader.name if cluster.leader else None
-                node_to_follow = next((m for m in cluster.members if m.name not in (self.state_handler.name,
-                                                                                     leader_name)), None)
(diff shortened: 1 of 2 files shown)


m4rrypro:
Thanks for the review, good catch. You are right: this approach breaks the replication-slot model because upstream slot creation depends on a concrete  target, not a role alias. I will drop this implementation.

m4rrypro:
Correction: upstream slot creation depends on a concrete `replicatefrom` member target, so a role alias (`any`/`standby`) is not sufficient.

m4rrypro:
Closing this one to avoid a misleading implementation. I’ll revisit #422 only if we can model slot management correctly for dynamic/role-based upstream selection.

PR #3596, opened 2026-05-03
Write pg_ident.conf during initdb post-bootstrap

The127:
### Problem

`Bootstrap.post_bootstrap` has two branches: custom bootstrap and initdb. The custom bootstrap branch calls `postgresql.config.replace_pg_ident()` to flush configured `pg_ident` entries to `pg_ident.conf`. The initdb branch does not.

As a result, on a freshly initdb-bootstrapped cluster, any `pg_ident` entries defined in the Patroni config are silently dropped. `pg_ident.conf` is never written until something else triggers a reload that calls `replace_pg_ident()`.

### Fix

Add the missing `replace_pg_ident()` call in the initdb branch, mirroring the custom bootstrap branch.

### Test

Added `test_post_bootstrap_initdb_writes_pg_ident` in `tests/test_bootstrap.py`, which asserts `replace_pg_ident()` is called on the initdb path. Verified it catches the regression by reverting the one-line fix locally: test fails without the fix, passes with it.

### DIFF
--- a/patroni/postgresql/bootstrap.py
+++ b/patroni/postgresql/bootstrap.py
@@ -423,6 +423,7 @@ END;$$""".format(f, quote_ident(rewind['username'], postgresql.connection()))
                             time.sleep(1)  # give a time to postgres to "reload" configuration files
                             postgresql.connection().close()  # close connection to reconnect with a new password
                 else:  # initdb
+                    postgresql.config.replace_pg_ident()
                     # We may want create database and extension for some MPP clusters
                     self._postgresql.mpp_handler.bootstrap()
         except Exception:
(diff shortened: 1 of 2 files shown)


CyberDem0n:
I can tell for sure, there is no bug in Patroni.
pg_ident.conf is always written on Postgres start: https://github.com/patroni/patroni/blob/abdd9e82e19a989505ae8128e634e6d5e0de23d2/patroni/postgresql/__init__.py#L752-L794

It works both, when `pg_ident` is set in `bootstrap.dcs.postgresql` or in `postgresql` sections.

The127:
Thank you for the super quick response! And you are of course correct.
I misdiagnosed something and wrote a bad reproducing setup. The issue is in another project entirely that uses patroni. I missed the actual issue. 
Sorry for taking your time and thank you again!

Issue #2433, opened 2022-10-17
SSL: CERTIFICATE_VERIFY_FAILED] certificate verify failed for reload using patronictl


jobinau:
**Describe the bug**
getting error
```
  File "/usr/local/lib/python3.6/site-packages/urllib3/request.py", line 170, in request_encode_body
    return self.urlopen(method, url, **extra_kw)
  File "/usr/local/lib/python3.6/site-packages/urllib3/poolmanager.py", line 376, in urlopen
    response = conn.urlopen(method, u.request_uri, **kw)
  File "/usr/local/lib/python3.6/site-packages/urllib3/connectionpool.py", line 828, in urlopen
    **response_kw
  File "/usr/local/lib/python3.6/site-packages/urllib3/connectionpool.py", line 828, in urlopen
    **response_kw
  File "/usr/local/lib/python3.6/site-packages/urllib3/connectionpool.py", line 828, in urlopen
    **response_kw
  File "/usr/local/lib/python3.6/site-packages/urllib3/connectionpool.py", line 788, in urlopen
    method, url, error=e, _pool=self, _stacktrace=sys.exc_info()[2]
  File "/usr/local/lib/python3.6/site-packages/urllib3/util/retry.py", line 592, in increment
    raise MaxRetryError(_pool, url, error or ResponseError(cause))
urllib3.exceptions.MaxRetryError: HTTPSConnectionPool(host='pg1', port=8008): Max retries exceeded with url: /reload (Caused by SSLError(SSLError(1, '[SSL: CERTIFICATE_VERIFY_FAILED] certificate verify failed (_ssl.c:877)'),))  
````
whenever I try using patronictl
```
patronictl -c /etc/patroni/patroni.yml reload <cluster>
```
But, the reload is working fine if I directly call the REST API
```
curl -X POST https://pg1:8008/reload --cacert /etc/patroni/rootCA.pem
```

**To Reproduce**
Self-signed certificate and keys with the following patroni configuration
```
restapi:
  listen: 0.0.0.0:8008
  connect_address: pg1:8008
  certfile: /etc/patroni/server.pem
  keyfile: /etc/patroni/server.key
  cafile: /etc/patroni/rootCA.pem
  verify_client: none

ctl:
  insecure: false
  certfile: /etc/patroni/server.pem
  keyfile: /etc/patroni/server.key
  cacert: /etc/patroni/rootCA.pem
```

**Expected behavior**
Expected to behave the same whether REST API is invoked or patronictl is used.

[...]

PR #3608, opened 2026-05-08
Add opt-in TLS verification flags for etcd/etcd3 DCS backends

chaitanyabodlapati:
## Summary

Patroni currently exposes `ctl.insecure: true` and `restapi.verify_client: none` for relaxing TLS verification on the REST API and control tool, but provides no equivalent for the DCS (etcd / etcd3) backend. Users with internal PKI, legacy SAN-less certificates, or etcd `--auto-tls` deployments have no in-config way to relax TLS verification short of patching site-packages directly.

This PR adds three opt-in flags under the `etcd:` and `etcd3:` config sections, mirroring the existing pattern for `restapi`/`ctl`:

- `verify` — disable CA chain validation (last resort, e.g. for `--auto-tls`)
- `verify_hostname` — keep CA validation, skip hostname/SAN check
- `hostname_checks_common_name` — accept CN-only certs (legacy PKI)

All flags default to the secure value; existing configs are unaffected. A `WARNING` is logged at startup whenever a flag is relaxed, so the security state is visible in operational logs.

## Why

Many production Patroni deployments connect to etcd over HTTPS using one of these configurations:
etcd --auto-tls — etcd generates a fresh self-signed certificate at every startup. There is no CA file to provide, so Patroni's cacert: parameter cannot help. The TLS handshake always fails certificate validation.
Internal PKI with legacy certificates — certificates issued by an internal CA but without Subject Alternative Name (SAN) entries. Modern Python (urllib3 v2+) dropped the CN fallback and only checks SAN, so these certificates fail hostname validation even when the CA is trusted.
Internal PKI with hostname/IP mismatch — certificates with valid SANs, but the SANs contain DNS names while Patroni connects via IP (or vice versa). Verification fails on identity, not on trust.
In all three cases, Patroni cannot connect to etcd. The DCS backend is unavailable, leader election cannot run, and the cluster cannot form. Because there is no configuration flag to relax verification for the DCS layer specifically, the only working option today is to modify Patroni's installed source files directly — a change that is invisible to configuration management, is silently lost on every upgrade, and affects urllib3 globally across the entire process.

## Design

The change is intentionally minimal:

1. **One new file:** `patroni/dcs/etcd_tls.py` (helper)
2. **Two added lines** in `patroni/dcs/etcd.py` (import + hook call)
3. **No changes** to `patroni/dcs/etcd3.py` — `Etcd3Client` inherits
   from `AbstractEtcdClientWithFailover` and the fix flows down via
   `super().__init__()`.

Scoping principles:
- No global urllib3 mutation. The helper modifies the etcd client's
  `PoolManager` in place via `connection_pool_kw`.
- No silenced warnings. urllib3's `InsecureRequestWarning` is left
  alone. The helper logs one WARNING at startup describing the
  relaxation.
- No retry-semantics changes. TLS configuration only.
- Backward-compatible: zero behavior change for users who don't
  set any new flag.

## FIPS compatibility

The helper uses only stdlib `ssl` module functions
(`create_default_context`, `_create_unverified_context`). It does not
override cipher selection or TLS minimum versions, and it does not
bypass OpenSSL's FIPS provider.

On FIPS-validated OpenSSL builds where
`SSLContext.hostname_checks_common_name` is read-only, the helper
detects this via a probe and raises `RuntimeError` at startup with a
clear message directing the user to `verify_hostname=false` instead.
This degraded behavior is tested.

## Related 

Multiple long-standing issues document this gap:
- #384 (2017) — self-signed certs, closed by pointing to `cacert:` workaround
- #2036 (2021) — etcd v3 + TLS Common Name, marked unsupported
- #2433 (2022) — `CERTIFICATE_VERIFY_FAILED` from `patronictl`, fixed
  for ctl/restapi but not DCS
- etcd-io/etcd#15945 (2023) — Patroni connecting to etcd with TLS


patroni/dcs/etcd_tls.py
```diff
@@ -0,0 +1,118 @@
...
+            "issuer will be accepted. Use only as a temporary measure while "
+            "migrating to a proper PKI."
+        )
+        return
+
+    # ---- Path 2: verification enabled, with optional narrowing ----
+    cafile = pool_kw.get('ca_certs') or config.get('ca_cert') or config.get('cacert')
+    ctx = ssl.create_default_context(cafile=cafile)
+
+    if cn_fallback:
```
CyberDem0n: Does it make sense to do it if verify_hostname is False?

patroni/dcs/etcd_tls.py
```diff
@@ -0,0 +1,118 @@
...
+        if not _cn_fallback_supported():
+            raise RuntimeError(
+                "hostname_checks_common_name=true was requested but this "
+                "Python/OpenSSL build does not support setting it (commonly "
+                "the case on FIPS-validated OpenSSL builds). Use "
+                "verify_hostname=false instead for the same practical effect "
+                "(CA chain still validated, only the SAN/CN identity check "
+                "is skipped)."
+            )
```
CyberDem0n: 1. By default `hostname_checks_common_name` is true: https://docs.python.org/3/library/ssl.html#ssl.SSLContext.hostname_checks_common_name
2. I don't really like raising RuntimeError.

patroni/dcs/etcd_tls.py
```diff
@@ -0,0 +1,118 @@
...
+        return
+
+    pool_kw = http_pool.connection_pool_kw
+
+    # ---- Path 1: verification fully disabled (last resort, opt-in) ----
+    if not verify:
+        ctx = ssl.SSLContext(ssl.PROTOCOL_TLS_CLIENT)
+        ctx.check_hostname = False
+        ctx.verify_mode = ssl.CERT_NONE
+        pool_kw['ssl_context'] = ctx
```
CyberDem0n: Why do we even need to create SSLContext in this case?

patroni/dcs/etcd_tls.py
```diff
@@ -0,0 +1,118 @@
...
+        pool_kw.pop('ca_certs', None)
+        logger.warning(
+            "etcd TLS certificate verification is DISABLED by configuration "
+            "(verify=false). This is insecure: any certificate from any "
+            "issuer will be accepted. Use only as a temporary measure while "
+            "migrating to a proper PKI."
+        )
+        return
+
+    # ---- Path 2: verification enabled, with optional narrowing ----
```
CyberDem0n: Have you thought about  etcd running in private network without TLS?
This warning doesn't make any sense this case.

CyberDem0n (commented):
I left a few comments.
Besides that, would like to avoid adding additional files to patroni/dcs in case if they don't implement DCS support

patroni/dcs/etcd_tls.py
```diff
@@ -0,0 +1,118 @@
...
+    verify_hostname: bool              # default True  (SAN/CN identity check)
+    hostname_checks_common_name: bool  # default False (allow legacy CN-only certs)
+
+All flags default to the secure value, so existing configs behave exactly
+as before. When a flag is relaxed, a single WARNING is logged at startup.
+No global urllib3 state is touched.
+"""
+
+import logging
+import ssl
```
CyberDem0n: If python is compiled without ssl everything will fail.

hughcapet:
@chaitanyabodlapati ping. Are you planning to follow up on this PR?

chaitanyabodlapati:
> @chaitanyabodlapati ping. Are you planning to follow up on this PR?

@hughcapet  Thanks for the review, and apologies for the delay — I got pulled into other work before I could follow up on the comments.

I agree with the direction of keeping this change as small and consistent with the existing Patroni structure as possible. I'll revisit the implementation and address your comments, especially:

1. avoiding a separate patroni/dcs/etcd_tls.py helper, since that directory is for DCS implementations;
2. simplifying the verify=false path and dropping the unnecessary SSLContext if urllib3 pool options are enough;
3. only emitting warnings for actual TLS verification relaxations, not for plain/private non-TLS etcd usage;
4. only building the CA-validation context where it's actually needed, rather than in paths where identity checking is already relaxed.
5. aligning hostname_checks_common_name with the Python/OpenSSL default (which is already True) and dropping the RuntimeError.

I'll rework the patch with these points in mind and push an updated version within the next couple of weeks.

chaitanyabodlapati:
> I left a few comments. Besides that, would like to avoid adding additional files to patroni/dcs in case if they don't implement DCS support

Done — moved the helper into patroni/dcs/etcd.py as a private function and removed the separate file. The tests moved into tests/test_etcd.py as well.

patroni/dcs/etcd_tls.py
chaitanyabodlapati: Good point — hostname_checks_common_name is now only applied when hostname verification is enabled, since it has no effect otherwise.

patroni/dcs/etcd_tls.py
chaitanyabodlapati: Agreed on both. Removed the RuntimeError. Since the attribute already defaults to True, on builds where the setter isn't available the desired behavior is already in effect, so setting it is now a best-effort try/except.

patroni/dcs/etcd_tls.py
chaitanyabodlapati: You're right, it wasn't needed. The verify=false path now just sets cert_reqs='CERT_NONE' and assert_hostname=False on the pool and lets urllib3 build its own context.

patroni/dcs/etcd_tls.py
chaitanyabodlapati: Added an early return when protocol is not https, so etcd on a private network without TLS is unaffected and no misleading warning is logged.

patroni/dcs/etcd_tls.py
chaitanyabodlapati: Moved import ssl inside the function, after the non-https and default-flag guards, so a Python built without ssl still works for plain-HTTP etcd.

chaitanyabodlapati:
Thanks for the review, @CyberDem0n. I've addressed all the comments:
1. Helper moved into etcd.py as a private function (no separate file); tests moved into tests/test_etcd.py.
2. verify=false no longer builds an SSLContext — it just sets cert_reqs='CERT_NONE' and assert_hostname=False on the pool and lets urllib3 build its own context.
3. Early return for non-https protocol, so etcd on a private network without TLS is unaffected and logs no misleading warning.
4. import ssl moved inside the function, after the non-https and default-flag guards, so a Python built without ssl still works for plain HTTP etcd.
5. hostname_checks_common_name only applied when hostname checking is on, and no longer raises RuntimeError — since it already defaults to True, setting it is now a best-effort try/except.

Would you mind taking another look when you get a chance? Thanks!

patroni/dcs/etcd.py
```diff
@@ -137,6 +137,81 @@ def _check_cluster_raft_term(self, cluster_id: Optional[str], value: Union[None,
...
+    if config.get('protocol') != 'https':
+        return
+
+    verify = config.get('verify', True)
+    verify_hostname = config.get('verify_hostname', True)
+    cn_fallback = config.get('hostname_checks_common_name', False)
+
+    # Nothing to do when all flags are at their secure defaults.
+    if verify and verify_hostname and not cn_fallback:
+        return
```
CyberDem0n: Default value of `SSLContext.hostname_checks_common_name` is True.
It goes against how cn_fallback is set.

patroni/dcs/etcd.py
```diff
@@ -137,6 +137,81 @@ def _check_cluster_raft_term(self, cluster_id: Optional[str], value: Union[None,
...
+                       'enabled.')
+
+    cert_file = pool_kw.get('cert_file')
+    key_file = pool_kw.get('key_file')
+    if cert_file:
+        ctx.load_cert_chain(certfile=cert_file, keyfile=key_file)
+
+    pool_kw['ssl_context'] = ctx
+    pool_kw['cert_reqs'] = 'CERT_REQUIRED'
+
```
CyberDem0n: Something is wrong.
Here is the table describing possible states:
| `verify` | `verify_hostname` | `hostname_checks_common_name` | Expected behavior |
|---:|---:|---:|---|
| `false` | any | any | No CA-chain or hostname verification |
| `true` | `false` | any | CA-chain validation only; hostname verification disabled |
| `true` | `true` | `false` | CA-chain validation plus SAN hostname validation; no CN fallback |
| `true` | `true` | `true` | CA-chain validation plus SAN hostname validation, with DNS CN fallback |

The implementation should explicitly set `ctx.hostname_checks_common_name` to the configured boolean whenever `verify_hostname` is true. If the platform cannot enforce `false`, silently continuing is unsafe; the code should either fail clearly or document that the platform cannot provide SAN-only verification.

patroni/dcs/etcd.py
```diff
@@ -137,6 +137,81 @@ def _check_cluster_raft_term(self, cluster_id: Optional[str], value: Union[None,
             raise StaleEtcdNode
         self._raft_term = raft_term
 
+def _apply_etcd_tls_config(http_pool: Any, config: Dict[str, Any]) -> None:
```
CyberDem0n: Please make it a method of AbstractEtcdClientWithFailover class.
This way you will not need to pass neither http_pool, nor config.

patroni/dcs/etcd.py
chaitanyabodlapati: Fixed — removed the fast-path that caused the default mismatch. When `verify_hostname` is enabled, `hostname_checks_common_name` is now explicitly set from the configured value.

patroni/dcs/etcd.py
chaitanyabodlapati: Thanks for the behavior table — updated the implementation to match it. When `verify_hostname` is enabled, `ctx.hostname_checks_common_name` is now explicitly set to the configured value:
`true` = DNS CN fallback allowed
`false` = SAN-only (no CN fallback)
If the platform cannot enforce `false`, it no longer silently continues — it logs a clear warning that SAN-only verification cannot be enforced.

patroni/dcs/etcd.py
chaitanyabodlapati: Done — converted it `to _configure_tls(self) `on `AbstractEtcdClientWithFailover`. It now uses `self._config` and `self.http `directly, so there's no need to pass http_pool or config.

patroni/dcs/etcd.py
```diff
@@ -154,13 +154,71 @@ def __init__(self, config: Dict[str, Any], dns_resolver: DnsCachingResolver, cac
...
+            ctx.check_hostname = False
+            pool_kw['assert_hostname'] = False
+            logger.warning('Etcd TLS hostname verification is disabled')
+        else:
+            # verify_hostname is True: explicitly set CN fallback to the
+            # configured value. True enables DNS Common Name fallback; False
+            # enforces SAN-only. The attribute defaults to True, so enforcing
+            # False depends on the platform supporting the setter.
+            try:
+                ctx.hostname_checks_common_name = cn_fallback
```
CyberDem0n, line 206: `hostname_checks_common_name` defaults to `False`, so the new default HTTPS path explicitly disables CN fallback for every existing configuration. **This is a behavior change despite the PR stating that existing configurations are unaffected**.

Python’s `SSLContext.hostname_checks_common_name` defaults to `True`; therefore, the default should preserve that behavior and only disable CN fallback when the user explicitly sets `hostname_checks_common_name: false`

chaitanyabodlapati: Good catch — you're right, defaulting it to False and always setting it changed behavior for existing configs. I Fixed it : the flag now defaults to unset, and `ctx.hostname_checks_common_name` is only assigned when the user explicitly configures it. When it's not set, the `SSLContext default (True)` is preserved, so existing configurations behave exactly as before. Explicit true/false still work per the table.

PR #3610, opened 2026-05-11
Do not parse the port as an int for replication connection

rdunklau:
Standby clusters support listing several hosts and several ports. If we do, we must accept the port not being an int but a comma separated list of ints.

This breaks when trying to open a replication connection using pg_rewind.


hughcapet:
We also need to update the port's validator then
https://github.com/patroni/patroni/blob/95e0af5c404326141d467659d52ffc5ee28768b1/patroni/validator.py#L1079-L1084

and add a unit test please C:

rdunklau:
Thanks ! 
Do we have a test for the multiple hosts for standby_clusters already ? I can't seem to find it.

hughcapet:
@rdunklau, no, i do not think we have anything. It was enabled without unit tests: https://github.com/patroni/patroni/commit/2d15e0dae65fb13c2a41c676e3a2ead223be1088 🙈

hughcapet:
@rdunklau gentle reminder:) Are you planning to work on this?

PR #3612, opened 2026-05-12
use monotonic for patroni

hanguannan:
https://github.com/patroni/patroni/issues/3560  
For this issue, due to project requirements, I modified the Patroni source code. The changes convert many places that used time.time (not all — some reasonable uses of time.time were left intact) to use monotonic; I also added logic in the schedule_next_run function to handle abnormally large nap_time values. I am now submitting these changes to the community in the hope they will be helpful. Thank you!


CyberDem0n (changes requested):
That's a huge change @hanguannan, and a lot of tests are failing.
I think we need to do it iteratively.
That is, start with the most important, get_monotonic_time() and than everything else, one by one.

hanguannan:
Thank you for your reply. Based on your suggestions, this time I have modified the code following the principle of minimizing changes, only altering the parts that require the use of a monotonic function.

After time.time() → time.monotonic() migration, timestamp variables initialized to 0 cause incorrect behavior on freshly booted systems where time.monotonic() returns small values (e.g., CI runners). With time.time(), 0 + offset > time.time() was always False since epoch is ~1.7B. With time.monotonic(), this becomes True on fresh boots.

So I Fixed: Use float('-inf') as the sentinel for "not yet set" / "expired" timestamps, which guarantees correct arithmetic results regardless of time.monotonic() value. Also use <= 0 instead of == 0 for guard checks as defensive programming.

patroni/dcs/kubernetes.py
```diff
@@ -908,6 +909,9 @@ def _cluster_from_nodes(self, group: str, nodes: Dict[str, K8sObject], pods: Col
         if leader_path == self.leader_path and (leader_record or self._leader_observed_record)\
                 and leader_record != self._leader_observed_record:
             self._leader_observed_record = leader_record
+            # Intentionally use wall-clock time: _leader_observed_time is compared with K8s
+            # renewTime/acquireTime annotations which are wall-clock timestamps, so it must
+            # be a real (system) timestamp for the comparison to be consistent.
             self._leader_observed_time = time.time()
```
CyberDem0n: This is not how it works.
renewTime/acquireTime are opaque. 
For `_leader_observed_time` we should also use time.monotonic().

patroni/ha.py
```diff
@@ -1000,10 +1000,10 @@ def process_sync_replication(self) -> None:
             # be postponed for `loop_wait` seconds, to give a chance to some replicas to start streaming.
             # In opposite case the /sync key will end up without synchronous nodes.
             if self.state_handler.is_primary():
-                if self._promote_timestamp == 0 or time.time() - self._promote_timestamp > self.dcs.loop_wait:
+                if self._promote_timestamp <= 0 or time.monotonic() - self._promote_timestamp > self.dcs.loop_wait:
                     self._process_quorum_replication()
-                if self._promote_timestamp == 0:
-                    self._promote_timestamp = time.time()
+                if self._promote_timestamp <= 0:
+                    self._promote_timestamp = time.monotonic()
```
CyberDem0n: According to [time.monotonic()](https://docs.python.org/3/library/time.html#time.monotonic) docs the reference point of the returned value is undefined.
That is, we can't replace `self._promote_timestamp == 0` with `self._promote_timestamp <= 0`.

patroni/ha.py
```diff
@@ -590,8 +590,8 @@ def bootstrap_standby_leader(self) -> Optional[bool]:
         return result
 
     def _handle_crash_recovery(self) -> Optional[str]:
-        if self._crash_recovery_started == 0 and (self.cluster.is_unlocked() or self._rewind.can_rewind):
-            self._crash_recovery_started = time.time()
+        if self._crash_recovery_started <= 0 and (self.cluster.is_unlocked() or self._rewind.can_rewind):
```
CyberDem0n: https://docs.python.org/3/library/time.html#time.monotonic:
> Return the value (in fractional seconds) of a monotonic clock, i.e. a clock that cannot go backwards. The clock is not affected by system clock updates. **The reference point of the returned value is undefined,** so that only the difference between the results of two calls is valid.

That is, we can't replace `self._crash_recovery_started == 0` with `self._crash_recovery_started <= 0`.

patroni/ha.py
```diff
@@ -1789,9 +1789,9 @@ def process_unhealthy_cluster(self) -> str:
         if not self.is_paused() and not self.is_standby_cluster() and \
                 not (self.cluster.failover and self.cluster.failover.candidate) and \
                 global_config.primary_race_backoff > 0 and self._prev_wal_lsn is not None:
-            if self._primary_race_backoff_timestamp == 0:
-                self._primary_race_backoff_timestamp = time.time()
-            time_left = self._primary_race_backoff_timestamp + global_config.primary_race_backoff - time.time()
+            if self._primary_race_backoff_timestamp <= 0:
```
CyberDem0n: The same problem. We can't use `self._primary_race_backoff_timestamp <= 0`.

tests/test_consul.py
```diff
@@ -197,15 +197,15 @@ def test_update_leader(self, mock_renew):
                 self.assertTrue(self.c.update_leader(cluster, 12345, failsafe={'foo': 'bar'}))
             with patch.object(KV, 'put', Mock(side_effect=ConsulException)):
                 self.assertFalse(self.c.update_leader(cluster, 12345))
-            mock_time = Mock(side_effect=[0, 0, 0, 0, 100, 200, 300])
-            with patch('time.time', mock_time), patch('time.time_ns', mock_time, create=True):
+            mock_time = Mock(side_effect=[0, 0, 100, 0, 0, 0, 0])
```
CyberDem0n: Why `side_effect` this changed?

tests/test_kubernetes.py
```diff
@@ -460,8 +460,8 @@ def test__update_leader_with_retry(self, mock_patch, mock_read):
         mock_patch.side_effect = [k8s_client.rest.ApiException(409, ''),
                                   k8s_client.rest.ApiException(409, ''), mock_namespaced_kind()]
         mock_read.return_value.metadata.resource_version = '2'
-        mock_time = Mock(side_effect=[0, 0, 100, 200, 0, 0, 0, 0, 0, 100, 200])
-        with patch('time.time', mock_time), patch('time.time_ns', mock_time, create=True):
+        mock_time = Mock(side_effect=[0, 0, 100, 0, 0, 0, 0, 100])
```
CyberDem0n: Why side_effect was changed?

patroni/ha.py
hanguannan: Since `_promote_timestamp` is set to `-inf` (line 236), it is compared with `<0` during the check here; simultaneously, for compatibility with the previous `==0` code logic, the comparison term is set to `<=0` here.

patroni/ha.py
hanguannan: Same as _promote_timestamp.

patroni/ha.py
hanguannan: Same as _promote_timestamp.

tests/test_consul.py
hanguannan: This modification changes `time.time()` to `time.monotonic()` in `_do_refresh_session`, causing `self._last_session_refresh = time.monotonic()` to consume an additional mocked value of `time.monotonic`. Originally, "100" was placed later (in the 5th position), but now it needs to be moved forward to the 3rd position. This ensures that `ensure_deadline(1, ...)` receives this large value and makes a decision that a timeout has occurred, thereby raising a `ConsulError`.

tests/test_kubernetes.py
hanguannan: In the _build_retain_slots() function (patroni/dcs/__init__.py:1889), time.time() was changed to time.monotonic(). This causes each call to update_leader to consume an additional mock value of time.monotonic(), thereby altering the allocation of the entire time series.

patroni/dcs/kubernetes.py
hanguannan: Yes,for _leader_observed_time we should also use time.monotonic(),and I will modify it.

patroni/ha.py
CyberDem0n: The logic breaks if time.monotonic() start with some negative value.

hanguannan: Yes,time.monotonic()‘s output may a  negative value,it's better to changed it for None. _promote_timestamp is delete for new code, I change _crash_recovery_started and _primary_race_backoff_timestamp to None.

CyberDem0n:
@hanguannan coverage for ha.py has decreased: https://github.com/patroni/patroni/actions/runs/27528330562/job/81360299358?pr=3612#step:26:790

More specifically, lines 952 and 2123.

hanguannan:
> @hanguannan coverage for ha.py has decreased: https://github.com/patroni/patroni/actions/runs/27528330562/job/81360299358?pr=3612#step:26:790
> 
> More specifically, lines 952 and 2123.

solved

patroni/ha.py
```diff
@@ -202,7 +202,7 @@ def is_active(self) -> bool:
         """
 
         with self._lock:
-            return self._last_update + self._dcs.ttl > time.time()
+            return self._last_update + self._dcs.ttl > time.monotonic()
 
     def set_is_active(self, value: float) -> None:
```
hughcapet, line 207: i guess we also need to rework the logic of this function and use float('-inf'), instead of 0

```
If the provided value is ``0``, it disables failsafe mode.
```

patroni/utils.py
```diff
@@ -727,7 +727,7 @@ def ensure_deadline(self, timeout: float, raise_ex: Optional[Exception] = None)
         :raises:
             :class:`Exception`: *raise_ex* if calculated deadline is smaller than provided *timeout*.
         """
-        if self.stoptime - time.time() < timeout:
+        if self.stoptime - time.monotonic() < timeout:
```
hughcapet, line 730:
```
    @property
    def stoptime(self) -> float:
        """Get the current stop time."""
        return self._cur_stoptime or 0
```

seems, it will not be working correctly if time.monotonic()<0
Should not we switch from 0 to `float('-int')` or maybe even None in this function?  Please double-check correctness in other places where `Retry.stoptime` is used

tests/test_kubernetes.py
hughcapet: additional call of time.time comes from LogRecord.__init__ (3rd and 7th)

```
patroni/patroni/dcs/kubernetes.py(1200)_update_leader_with_retry()
-> logger.warning('Concurrent update of %s', self.leader_path)
  /Library/Developer/CommandLineTools/Library/Frameworks/Python3.framework/Versions/3.9/lib/python3.9/logging/__init__.py(1458)warning()
-> self._log(WARNING, msg, args, **kwargs)
  /Library/Developer/CommandLineTools/Library/Frameworks/Python3.framework/Versions/3.9/lib/python3.9/logging/__init__.py(1587)_log()
-> record = self.makeRecord(self.name, level, fn, lno, msg, args,
  /Library/Developer/CommandLineTools/Library/Frameworks/Python3.framework/Versions/3.9/lib/python3.9/logging/__init__.py(1556)makeRecord()
-> rv = _logRecordFactory(name, level, fn, lno, msg, args, exc_info, func,
  /Library/Developer/CommandLineTools/Library/Frameworks/Python3.framework/Versions/3.9/lib/python3.9/logging/__init__.py(287)__init__()
-> ct = time.time()
```
it was actually using only 10 of the mock values

tests/test_consul.py
hughcapet: (4th call)
```
  patroni/patroni/dcs/consul.py(650)_update_leader()
-> logger.warning('Recreating the leader key due to session mismatch')
  /Library/Developer/CommandLineTools/Library/Frameworks/Python3.framework/Versions/3.9/lib/python3.9/logging/__init__.py(1458)warning()
-> self._log(WARNING, msg, args, **kwargs)
  /Library/Developer/CommandLineTools/Library/Frameworks/Python3.framework/Versions/3.9/lib/python3.9/logging/__init__.py(1587)_log()
-> record = self.makeRecord(self.name, level, fn, lno, msg, args,
  /Library/Developer/CommandLineTools/Library/Frameworks/Python3.framework/Versions/3.9/lib/python3.9/logging/__init__.py(1556)makeRecord()
-> rv = _logRecordFactory(name, level, fn, lno, msg, args, exc_info, func,
  /Library/Developer/CommandLineTools/Library/Frameworks/Python3.framework/Versions/3.9/lib/python3.9/logging/__init__.py(287)__init__()
-> ct = time.time()
```

it was actually using only 5 of the mock values

hughcapet:
```suggestion
            mock_time = Mock(side_effect=[0, 0, 0, 100, 200, 300])
```

tests/test_kubernetes.py
hughcapet:
```suggestion
        mock_time = Mock(side_effect=[0, 0, 200, 0, 0, 0, 0, 100])
```

hughcapet (commented):
plus add the patch to test__refresh_api_servers_cache, otherwise coverage drops

```
diff --git a/tests/test_kubernetes.py b/tests/test_kubernetes.py
index dde3363d..48d9c816 100644
--- a/tests/test_kubernetes.py
+++ b/tests/test_kubernetes.py
@@ -189,6 +189,7 @@ class TestApiClient(unittest.TestCase):
         mock_request.side_effect = [socket.timeout, socket.timeout, self.mock_get_ep]
         self.assertRaises(K8sConnectionFailed, retry, self.a.call_api, 'GET', 'f', _retry=retry)
 
+    @patch('time.monotonic', Mock(side_effect=[100, 101]))
     def test__refresh_api_servers_cache(self, mock_request):
         mock_request.side_effect = k8s_client.rest.ApiException(403, '')
         self.a.refresh_api_servers_cache()
```

tests/test_kubernetes.py
hanguannan: Modified

tests/test_consul.py
hanguannan: Modified

patroni/utils.py
hanguannan: yes, i modified and check other usage situation

patroni/ha.py
hanguannan: Modified

patroni/ha.py
```diff
@@ -229,15 +230,15 @@ def __init__(self, patroni: Patroni):
...
+        self._leader_expiry = float('-inf')
         self._leader_expiry_lock = RLock()
         self._failsafe = Failsafe(patroni.dcs)
         self._was_paused = False
         self._synchronous_strict_mode_activated = False
         self._leader_timeline = None
         self.recovering = False
         self._async_response = CriticalTask()
-        self._crash_recovery_started = 0
+        self._crash_recovery_started: Optional[float] = None
```
CyberDem0n: Hmm, why can't it have `float('-inf')` default value?
It will allow to simplify https://github.com/patroni/patroni/pull/3612/changes#diff-b6a3a0ef64783a6d32feb908d74a3a009e9f25008b8ca5313c9c9747e3b74a63R2123-R2126

self._leader_expiry set to float('-inf') seems to work, and I don't see a reason why here it shouldn't

patroni/ha.py
```diff
@@ -259,7 +260,7 @@ def __init__(self, patroni: Patroni):
         # receive/flush/replay LSN from last cycle, is used to detect false positives of dead primary
         self._prev_wal_lsn: Optional[int] = None
         # timestamp when primary_race_backoff was triggered
-        self._primary_race_backoff_timestamp = 0
+        self._primary_race_backoff_timestamp: Optional[float] = None
```
CyberDem0n: The same question as about `self._crash_recovery_started`.

tests/test_ha.py
```diff
@@ -1,6 +1,7 @@
 import datetime
 import os
 import sys
+import time
```
CyberDem0n: All changes in test_ha.py seems to be related to attempt to resolve merge conflict.
We don't need them (especially after applying suggestions I left in ha.py).

One minor thing - there is one mock seems to be forgotten:
```
diff --git a/tests/test_ha.py b/tests/test_ha.py
index 73b2e253..5b6f940f 100644
--- a/tests/test_ha.py
+++ b/tests/test_ha.py
@@ -2107,7 +2107,7 @@ class TestHa(PostgresInit):
 
         # Test that _process_quorum_replication doesn't take longer than loop_wait
         with patch.object(Postgresql, 'synchronous_standby_names', Mock(return_value='ANY 1 (foo)')), \
-                patch('time.time', Mock(side_effect=[30, 60, 90, 120, 150])):
+                patch('time.monotonic', Mock(side_effect=[30, 60, 90, 120, 150])):
             self.ha.process_sync_replication()
 
         # Test foo -> ANY 1 (foo) transition
```

tests/test_kubernetes.py
```diff
@@ -189,6 +189,7 @@ def test_request(self, mock_request):
         mock_request.side_effect = [socket.timeout, socket.timeout, self.mock_get_ep]
         self.assertRaises(K8sConnectionFailed, retry, self.a.call_api, 'GET', 'f', _retry=retry)
 
+    @patch('time.monotonic', Mock(side_effect=[100, 101]))
```
CyberDem0n: I don't think we need it.

tests/test_utils.py
```diff
@@ -10,9 +10,16 @@
...
+    @patch('time.monotonic', Mock(side_effect=[0, 0, 0.001]))
+    @patch('time.sleep', Mock())
     def test_polling_loop(self):
         self.assertEqual(list(polling_loop(0.001, interval=0.001)), [0])
 
+    @patch('time.monotonic', Mock(side_effect=[0, 0, 0.5, 0.9, 1.0]))
+    @patch('time.sleep', Mock())
+    def test_polling_loop_multiple_iterations(self):
+        self.assertEqual(list(polling_loop(1.0, interval=0.5)), [0, 1, 2])
+
```
CyberDem0n: I don't think we need to change it.

patroni/utils.py
```diff
@@ -713,7 +713,7 @@ def update_delay(self) -> None:
     @property
     def stoptime(self) -> float:
         """Get the current stop time."""
-        return self._cur_stoptime or 0
+        return self._cur_stoptime or float('-inf')
```
CyberDem0n:
```suggestion
        return float('-inf') if self._cur_stoptime is None else self._cur_stoptime
```

patroni/dcs/kubernetes.py
```diff
@@ -426,7 +427,7 @@ def request(
                     if TYPE_CHECKING:  # pragma: no cover
                         assert isinstance(retry, Retry)  # K8sConnectionFailed is raised only if retry is not None!
                     sleeptime = retry.sleeptime
-                    remaining_time = (retry.stoptime or time.time()) - sleeptime - time.time()
+                    remaining_time = (retry.stoptime or time.monotonic()) - sleeptime - time.monotonic()
```
CyberDem0n:
```suggestion
                    remaining_time = retry.stoptime - sleeptime - time.monotonic()
```

CyberDem0n (commented):
Submitting a few more findings.
We are getting really close to finishing it.

patroni/dcs/kubernetes.py
hanguannan: Modified

patroni/utils.py
hanguannan: Modified

tests/test_kubernetes.py
hanguannan: Modified

patroni/ha.py
hanguannan: Modified

patroni/ha.py
hanguannan: Modified

patroni/ha.py
```diff
@@ -2546,7 +2547,10 @@ def watch(self, timeout: float) -> bool:
         # watch on leader key changes if the postgres is running and leader is known and current node is not lock owner
         if not self._async_executor.busy and (not self.cluster or self.cluster.is_unlocked()):
             leader_version = None
-            time_left = self._primary_race_backoff_timestamp + global_config.primary_race_backoff - time.time()
+            if self._primary_race_backoff_timestamp > float('-inf'):
+                time_left = self._primary_race_backoff_timestamp + global_config.primary_race_backoff - time.monotonic()
+            else:
+                time_left = -1
```
CyberDem0n:
```suggestion
            time_left = self._primary_race_backoff_timestamp + global_config.primary_race_backoff - time.monotonic()
```

patroni/ha.py
```diff
@@ -229,15 +230,15 @@ def __init__(self, patroni: Patroni):
...
+        self._leader_expiry = float('-inf')
         self._leader_expiry_lock = RLock()
         self._failsafe = Failsafe(patroni.dcs)
         self._was_paused = False
         self._synchronous_strict_mode_activated = False
         self._leader_timeline = None
         self.recovering = False
         self._async_response = CriticalTask()
-        self._crash_recovery_started = 0
+        self._crash_recovery_started: float = float('-inf')
```
CyberDem0n:
```suggestion
        self._crash_recovery_started = float('-inf')
```

patroni/ha.py
```diff
@@ -259,7 +260,7 @@ def __init__(self, patroni: Patroni):
         # receive/flush/replay LSN from last cycle, is used to detect false positives of dead primary
         self._prev_wal_lsn: Optional[int] = None
         # timestamp when primary_race_backoff was triggered
-        self._primary_race_backoff_timestamp = 0
+        self._primary_race_backoff_timestamp: float = float('-inf')
```
CyberDem0n:
```suggestion
        self._primary_race_backoff_timestamp = float('-inf')
```

tests/test_ha.py
CyberDem0n:
```suggestion
```

tests/test_ha.py
```diff
@@ -24,6 +25,7 @@
 from patroni.postgresql.rewind import Rewind, REWIND_STATUS
 from patroni.postgresql.slots import SlotsHandler
 from patroni.postgresql.sync import _SyncState
+from patroni.quorum import Transition
```
CyberDem0n:
```suggestion
```

tests/test_ha.py
```diff
@@ -339,6 +341,12 @@ def test_crash_recovery(self):
             self.ha.cluster.config.data.update({'maximum_lag_on_failover': 10})
             self.assertEqual(self.ha.run_cycle(), 'terminated crash recovery because of startup timeout')
 
+            # Test handle_long_action_in_progress when _crash_recovery_started is float('-inf')
+            # (the else branch setting time_left = 0)
+            self.ha._crash_recovery_started = float('-inf')
+            self.ha.is_failover_possible = true
+            self.assertEqual(self.ha.run_cycle(), 'terminated crash recovery because of startup timeout')
+
```
CyberDem0n:
```suggestion
```

tests/test_ha.py
```diff
@@ -539,6 +547,8 @@ def test_no_dcs_connection_primary_demote(self):
         self.ha.load_cluster_from_dcs = Mock(side_effect=DCSError('Etcd is not responding properly'))
         self.assertEqual(self.ha.run_cycle(), 'demoted self because DCS is not accessible and I was a leader')
 
+    @patch('time.time', Mock(return_value=100))
+    @patch('patroni.ha.time.monotonic', Mock(return_value=100))
```
CyberDem0n:
```suggestion
```

tests/test_ha.py
```diff
@@ -559,6 +569,8 @@ def test_check_failsafe_topology(self):
         self.assertEqual(self.ha.run_cycle(),
                          'continue to run as a leader because failsafe mode is enabled and all members are accessible')
 
+    @patch('time.time', Mock(return_value=100))
+    @patch('patroni.ha.time.monotonic', Mock(return_value=100))
```
CyberDem0n:
```suggestion
```

tests/test_ha.py
```diff
@@ -571,6 +583,8 @@ def test_no_dcs_connection_primary_failsafe(self):
...
```
CyberDem0n:
```suggestion
```

tests/test_ha.py
```diff
@@ -579,6 +593,8 @@ def test_readonly_dcs_primary_failsafe(self):
...
```
CyberDem0n:
```suggestion
```

tests/test_ha.py
```diff
@@ -591,6 +607,8 @@ def test_no_dcs_connection_replica_failsafe(self):
             self.assertEqual(self.ha.run_cycle(), 'DCS is not accessible')
             self.assertEqual(mock_logger.call_args_list[0][0][0], 'Failed to fetch current wal lsn: %r')
 
+    @patch('time.time', Mock(return_value=100))
+    @patch('patroni.ha.time.monotonic', Mock(return_value=100))
```
CyberDem0n:
```suggestion
```

tests/test_ha.py
```diff
@@ -2069,6 +2087,7 @@ def test__process_quorum_replication(self):
                                                               _SyncState('quorum', 1, CaseInsensitiveSet(['foo']),
                                                                          CaseInsensitiveSet(['foo']),
                                                                          CaseInsensitiveSet(['foo']))])
+        self.ha._promote_timestamp = time.monotonic() - self.ha.dcs.loop_wait - 1
```
CyberDem0n:
```suggestion
```

tests/test_ha.py
```diff
@@ -2081,7 +2100,8 @@ def test__process_quorum_replication(self):
         self.assertEqual(mock_set_sync.call_count, 1)
         self.assertEqual(mock_set_sync.call_args_list[0][0], ('ANY 1 (other)',))
 
-        # Test that we stick with last known sync node when synchronous_mode_strict and no nodes available
+        # Test ANY 1 (*) when synchronous_mode_strict and no nodes available
+        self.ha._promote_timestamp = time.monotonic() - self.ha.dcs.loop_wait - 1
```
CyberDem0n:
```suggestion
```

tests/test_ha.py
```diff
@@ -2107,9 +2127,25 @@ def test__process_quorum_replication(self):
...
+        self.ha.dcs.write_sync_state = Mock(return_value=SyncState(1, 'leader', 'other', 0))
+        self.ha.cluster = get_cluster_initialized_with_leader(sync=('leader', 'other', 1))
+        # QuorumStateResolver produces two transitions: first 'quorum' (processed normally),
+        # second 'sync' (_check_timeout() returns True on this iteration).
+        mock_transitions = [Transition('quorum', 'leader', 0, CaseInsensitiveSet()),
+                            Transition('sync', 'leader', 1, CaseInsensitiveSet(['other']))]
+        with patch('patroni.ha.QuorumStateResolver', Mock(return_value=mock_transitions)), \
+                patch('time.monotonic', Mock(side_effect=[0, 0, self.ha.dcs.loop_wait + 1])):
+            self.ha._process_quorum_replication()
+
```
CyberDem0n:
```suggestion
```

tests/test_utils.py
```diff
@@ -10,9 +10,16 @@
 
 class TestUtils(unittest.TestCase):
 
+    @patch('time.monotonic', Mock(side_effect=[0, 0, 0.001]))
+    @patch('time.sleep', Mock())
```
CyberDem0n:
```suggestion
```

tests/test_utils.py
CyberDem0n:
```suggestion
```

tests/test_ha.py
hanguannan: deleted

patroni/ha.py
hanguannan: changed

patroni/ha.py
hanguannan: changed

tests/test_ha.py
hanguannan: changed

tests/test_ha.py
hanguannan: changed

patroni/ha.py
hanguannan: changed

tests/test_ha.py
hanguannan: changed

tests/test_ha.py
hanguannan: changed

tests/test_ha.py
hanguannan: changed

tests/test_ha.py
hanguannan: changed

tests/test_ha.py
hanguannan: changed

tests/test_ha.py
hanguannan: changed

tests/test_utils.py
hanguannan: changed

tests/test_ha.py
hanguannan: changed

tests/test_ha.py
hanguannan: changed

tests/test_ha.py
hanguannan: we should keep `@patch('patroni.ha.time.monotonic', Mock(return_value=100))` or will failed when test `TestHa.test_check_failsafe_topology`

tests/test_utils.py
hanguannan: we should keep `@patch('time.monotonic', Mock(side_effect=[0, 0, 0.001]))` or will failed when test `TestUtils.test_polling_loop`

tests/test_ha.py
```diff
@@ -2081,7 +2082,6 @@ def test__process_quorum_replication(self):
         self.assertEqual(mock_set_sync.call_count, 1)
         self.assertEqual(mock_set_sync.call_args_list[0][0], ('ANY 1 (other)',))
 
-        # Test that we stick with last known sync node when synchronous_mode_strict and no nodes available
         self.p.sync_handler.current_state = Mock(return_value=_SyncState('quorum', 1,
```
CyberDem0n, line 2186:
```suggestion
        # Test that we stick with last known sync node when synchronous_mode_strict and no nodes available
        self.p.sync_handler.current_state = Mock(return_value=_SyncState('quorum', 1,
```

patroni/ha.py
```diff
@@ -1226,7 +1227,7 @@ def enforce_primary_role(self, message: str, promote_message: str) -> str:
                 return 'Postponing promotion because synchronous replication state was updated by somebody else'
             if self.state_handler.role not in (PostgresqlRole.PRIMARY, PostgresqlRole.PROMOTED):
                 # reset failsafe state when promote
-                self._failsafe.set_is_active(0)
+                self._failsafe.set_is_active(float('-inf'))
```
hughcapet, line 1230: this is not the only place where it needs to be adjusted (see Ha._handle_dcs_error and Ha._run_cycle

hughcapet (changes requested):

patroni/ha.py
hanguannan: yes,I have checked and changed those places.

hughcapet: no, you haven't
https://github.com/hanguannan/patroni/blob/24803b01cd3aeb224b10186a0416550e393417a4/patroni/ha.py#L2443

https://github.com/hanguannan/patroni/blob/24803b01cd3aeb224b10186a0416550e393417a4/patroni/ha.py#L2455

hanguannan: Yes, I missed that. Thank you for the reminder. It has been revised.

patroni/dcs/kubernetes.py
```diff
@@ -917,8 +921,9 @@ def _cluster_from_nodes(self, group: str, nodes: Dict[str, K8sObject], pods: Col
             ttl = self._ttl
 
         # We want to check validity of the leader record only for our own cluster
+        # Using monotonic time for local TTL duration check (see above).
         if leader_path == self.leader_path and\
-                not (metadata and self._leader_observed_time and self._leader_observed_time + ttl >= time.time()):
+                not (metadata and self._leader_observed_time and self._leader_observed_time + ttl >= time.monotonic()):
```
CyberDem0n:
```suggestion
                not (metadata and self._leader_observed_time is not None
                     and self._leader_observed_time + ttl >= time.monotonic()):
```

CyberDem0n (approved):

hughcapet (approved):

PR #3614, opened 2026-05-12
Add patronictl.spec to mkbinary.sh and supplement the citus module.

feeling0713:
Add the generation of the ctl tool to the packaging script; supporting Citus is a core feature, I believe it should be included by default in the package.

### DIFF
--- a/mkbinary.sh
+++ b/mkbinary.sh
@@ -3,3 +3,4 @@ set -e
 
 pip install --ignore-installed pyinstaller
 pyinstaller --clean patroni.spec
+pyinstaller --clean patronictl.spec
--- a/patronictl.spec
+++ b/patronictl.spec
@@ -20,7 +20,7 @@ def resources():
     return [(res_dir + '*.' + e, res_dir) for e in exts if e.lower() in {'yml', 'yaml'}]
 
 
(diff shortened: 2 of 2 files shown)


patroni.spec
```diff
@@ -8,7 +8,7 @@ def hiddenimports():
     sys.path.insert(0, '.')
     try:
         import patroni.dcs
-        return patroni.dcs.dcs_modules() + ['http.server']
+        return patroni.dcs.dcs_modules() + ['http.server', 'patroni.postgresql.mpp.citus']
```
CyberDem0n, line 11: It is better to add iter_mpp_modules() function to `patroni/postgresql/mpp/__init__.py`.
Similar to how it is done here: https://github.com/patroni/patroni/blob/95e0af5c404326141d467659d52ffc5ee28768b1/patroni/dcs/__init__.py#L83-L90

CyberDem0n: I also have feeling that we don't need to explicitly list 'http.server' anymore.

CyberDem0n (commented):
Why does patronictl need a second spec file?

hughcapet:
@feeling0713 ping. Are you planning to follow up on this PR?

PR #3637, opened 2026-06-15
fix: retry statement_timeout cancellation instead of failing over

gclough:
A statement_timeout cancellation comes back from psycopg as QueryCanceled, which is a subclass of OperationalError. The exact-type check in NamedConnection.query (`type(exc) in (DatabaseError, OperationalError)`) did not match it, so it was re-raised raw rather than wrapped as PostgresConnectionException. Since Retry only retries PostgresConnectionException, a single slow monitoring query (e.g. when PostgreSQL is briefly overloaded) was never retried and could poison the heart-beat cycle, leading the leader to demote itself.

Catch QueryCanceled explicitly and wrap it as PostgresConnectionException so it is retried the same way as other transient failures. The connection is left open because a statement_timeout cancellation leaves the session healthy and reusable, and with autocommit there is no aborted transaction to clean up.

### DIFF
--- a/patroni/postgresql/connection.py
+++ b/patroni/postgresql/connection.py
@@ -73,6 +73,12 @@ class NamedConnection:
             with self.get().cursor() as cursor:
                 cursor.execute(sql.encode('utf-8'), params or None)
                 return cursor.fetchall() if cursor.rowcount and cursor.rowcount > 0 else []
+        except psycopg.QueryCanceled as exc:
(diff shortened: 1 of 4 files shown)


gclough:
Draft for staging, but let me validate before review.

CyberDem0n (commented):
@gclough the problem indeed exists, however the right way of solving it would be handling exception directly in Postgresql._cluster_info_state_get() method.

PR #3640, opened 2026-06-17
Include database name in connection key for CitusDatabaseHandler

XiuhuaRuan:
I used patroni with branch feature/multi-citus and deployed citus cluster with multi databases. But only the metadata of the first database could be updated. I noticed all CitusDatabaseHandler instances use 'citus' as the key, so only the first handler will create a new connection; all subsequent handlers will reuse the same NamedConnection object—which points to the first database. So I used the database name to distinguish connection keys and verified OK.

### DIFF
--- a/patroni/postgresql/mpp/citus.py
+++ b/patroni/postgresql/mpp/citus.py
@@ -399,8 +399,9 @@ class CitusDatabaseHandler(Citus, Thread):
         self.daemon = True
         if config:
             self._connection = postgresql.connection_pool.get(
-                'citus', {'dbname': config['database'],
-                          'options': '-c statement_timeout=0 -c idle_in_transaction_session_timeout=0'})
+                f'citus-{config["database"]}',
+                {'dbname': config['database'],
+                 'options': '-c statement_timeout=0 -c idle_in_transaction_session_timeout=0'})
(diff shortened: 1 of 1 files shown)


patroni/postgresql/mpp/citus.py
```diff
@@ -399,7 +399,7 @@ def __init__(self, postgresql: 'Postgresql', config: Dict[str, Any]) -> None:
         self.daemon = True
         if config:
             self._connection = postgresql.connection_pool.get(
-                'citus', {'dbname': config['database'],
+                 f'citus-{config["database"]}', {'dbname': config['database'],
                           'options': '-c statement_timeout=0 -c idle_in_transaction_session_timeout=0'})
```
CyberDem0n: Nice catch, @XiuhuaRuan!
Please format code, flake8 complains about it:
```
patroni/postgresql/mpp/citus.py:402:18: E126 continuation line over-indented for hanging indent
patroni/postgresql/mpp/citus.py:403:27: E128 continuation line under-indented for visual indent
```

XiuhuaRuan:
@CyberDem0n Thanks for your reminder.  Addressed flake8 complaints and committed again.

PR #3642, opened 2026-06-19
Fix Prometheus metric type for `patroni_postgres_timeline`

kylorend3r:
While I was reviewing the api.py to review prometheus metrics I realised `patroni_postgres_timeline` is declared as `counter` in the `/metrics` endpoint. Prometheus counters must only ever increase, but the metric's own HELP text documents that it drops to `0` when PostgreSQL is not running: "Postgres timeline of this node (if running), 0 otherwise."

Every PostgreSQL restart — via `patronictl restart`, a config change requiring restart, or crash recovery — produces the scrape sequence `1 → 0 → 1`. Prometheus interprets the drop as a counter reset and applies reset compensation:

```bash
  increase(patroni_postgres_timeline[5m]) = 2   (should be 0)
  rate(patroni_postgres_timeline[5m])     > 0   (should be 0)
```

**Any alert on timeline changes such as `increase(patroni_postgres_timeline[5m]) > 0` fires on every routine restart rather than only on actual failovers.**

The metric represents current state, not a cumulative count. Change its declared type from `counter` to `gauge`.

To prevent false positive timeline alerts on the prometheus side we can update the type of the metric.

### DIFF
--- a/tests/test_api.py
+++ b/tests/test_api.py
@@ -426,9 +426,6 @@ class TestRestApiHandler(unittest.TestCase):
 
     @patch.object(MockPatroni, 'dcs')
     def test_do_GET_metrics(self, mock_dcs):
-        with patch.object(RestApiHandler, 'write_response') as mock_write:
-            MockRestApiServer(RestApiHandler, 'GET /metrics')
-            self.assertIn('# TYPE patroni_postgres_timeline gauge', mock_write.call_args[0][1])
         self.assertIsNotNone(MockRestApiServer(RestApiHandler, 'GET /metrics'))
         # Test with failsafe_mode enabled
         with patch.object(MockHa, 'is_failsafe_mode', Mock(return_value=True)):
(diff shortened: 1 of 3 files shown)


CyberDem0n (commented):
Please also update docs: https://github.com/patroni/patroni/blob/5ae67a7b845b017631996b29bc7df09b9771374d/docs/rest_api.rst?plain=1#L287-L366

kylorend3r:
> Please also update docs:
> 
> https://github.com/patroni/patroni/blob/5ae67a7b845b017631996b29bc7df09b9771374d/docs/rest_api.rst?plain=1#L287-L366

Updated, thanks.

tests/test_api.py
```diff
@@ -426,6 +426,9 @@ def test_do_GET_config(self, mock_dcs):
 
     @patch.object(MockPatroni, 'dcs')
     def test_do_GET_metrics(self, mock_dcs):
+        with patch.object(RestApiHandler, 'write_response') as mock_write:
```
hughcapet: there is no need for this test, let's remove it

hughcapet (approved):

PR #3645, opened 2026-06-22
Extend `Assigning synchronous_standby_names` log with added/removed members

kylorend3r:
I was reading the patroni logs during an investigation and realised that emitted logs about synchronous_standby_names doesn't reveal the difference. Put another way: I couldn't see the which instance removed and added to synchronous_standby_names during the problem. 

This refactoring is about enhancing the log only. It does not change where the log is emitted, does not add a new log location, and does not alter any existing behavior or log message. **It extends the existing `"Assigning synchronous_standby_names to …"` line which has always been emitted from `set_synchronous_standby_names()`**

New logs will be emitted as follow

```
INFO: Assigning synchronous_standby_names to patroni1 (added=['patroni1'] removed=[])
INFO: Assigning synchronous_standby_names to 2 (patroni1,patroni2) (added=['patroni2'] removed=[])
INFO: Assigning synchronous_standby_names to patroni1 (added=[] removed=['patroni2'])
```

When the member set is unchanged (e.g. Patroni re-confirms the same replica) the original plain form is preserved:

```
INFO: Assigning synchronous_standby_names to patroni1
```

With the enhanced log we can immediately answer:

- *Which replica just became synchronous?* → look for `added=(...)` 
- *Which replica left the sync set?* → look for `removed=(...)`
- *Was patroni2 removed before or after the outage?* → timestamp on the line with `removed=('patroni2')` answers it directly

### DIFF
--- a/patroni/postgresql/sync.py
+++ b/patroni/postgresql/sync.py
@@ -311,18 +311,12 @@ class SyncHandler(object):
         if synchronous_standby_names == self._synchronous_standby_names:
             return
 
-        old_members = self._ssn_data.members
         self._synchronous_standby_names = synchronous_standby_names
         try:
             self._ssn_data = parse_sync_standby_names(synchronous_standby_names)
         except ValueError as e:
             logger.warning('%s', e)
             self._ssn_data = deepcopy(_EMPTY_SSN)
-
-        added = sorted(self._ssn_data.members - old_members)
-        removed = sorted(old_members - self._ssn_data.members)
-        if added or removed:
-            logger.info('`synchronous_standby_names` changed: added=%s removed=%s', added, removed)
 
         # Invalidate cache of "sync" connections
         for app_name in list(self._ready_replicas.keys()):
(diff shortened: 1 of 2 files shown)


hughcapet:
Does this solve the problem of not having an access to the previous logs and thus not being able to compare with the previous `Assigning synchronous_standby_names to …` record? Or am I missing something?

kylorend3r:
Ah I see, no it doesn't solve the problem you mentioned. It's an enhancement about current log emission.  

But I understood the issue you pointed out. I can also work on this. 

After a patroni/postgresql restart it can read the current ssn and compare the diff on conf change.

hughcapet:
i only mean that as long as you have access to the old logs, you can always compare with the previous log line about the value change and see what was added/removed

kylorend3r:
Yes but during the debugging this enhancement can help you to check the last log and see the diff. 

Instead of comparing the logs we can enhance the log line. Also, if we rotate the logs (for some reason) we may not be able to compare If I got your point correctly.

hughcapet:
imho, the goal is not worth adding this additional complexity to the code

Issue #2278, opened 2022-04-18
Patroni throws a traceback to a TCP RST segment from haproxy


victor-sudakov:
**Describe the bug**
Haproxy 2.5.1 performs a [httpchk](https://www.haproxy.com/documentation/hapee/latest/onepage/#4.2-option%20httpchk) to the Patroni REST API and finishes it with a TCP RST segment. This looks like normal behavior. Patroni however reacts with a Traceback in its log complaning about "ConnectionResetError: [Errno 104] Connection reset by peer"

**TCP Session**
![haproxy_rst](https://user-images.githubusercontent.com/12457866/163882682-52aaaf97-0c1d-4c5c-b271-1406cf46728e.png)

**Patroni log snippet**
```
2022-04-15 01:58:48,093 WARNING: Exception happened during processing of request from 172.19.69.216:52962 
2022-04-15 01:58:48,094 WARNING: Exception happened during processing of request from 172.19.69.216:40907 
2022-04-15 01:58:48,094 WARNING: Traceback (most recent call last):
  File "/usr/lib/python3.9/socketserver.py", line 650, in process_request_thread
    self.finish_request(request, client_address)
  File "/usr/lib/python3.9/socketserver.py", line 360, in finish_request
    self.RequestHandlerClass(request, client_address, self)
  File "/usr/lib/python3.9/socketserver.py", line 720, in __init__
    self.handle()
  File "/usr/lib/python3.9/http/server.py", line 427, in handle  
    self.handle_one_request()
  File "/usr/lib/python3/dist-packages/patroni/api.py", line 652, in handle_one_request
    BaseHTTPRequestHandler.handle_one_request(self)
  File "/usr/lib/python3.9/http/server.py", line 415, in handle_one_request
    method()
  File "/usr/lib/python3/dist-packages/patroni/api.py", line 178, in do_GET
    self._write_status_response(status_code, response)
  File "/usr/lib/python3/dist-packages/patroni/api.py", line 94, in _write_status_response
    self._write_json_response(status_code, response)
  File "/usr/lib/python3/dist-packages/patroni/api.py", line 53, in _write_json_response
    self._write_response(status_code, json.dumps(response, default=str), content_type='application/json')
  File "/usr/lib/python3/dist-packages/patroni/api.py", line 50, in _write_response
    self.wfile.write(body.encode('utf-8'))
  File "/usr/lib/python3.9/socketserver.py", line 799, in write  
    self._sock.sendall(b)
[...]

Issue #2229, opened 2022-02-25
Excessive tracebacks when HAProxy polls (TLS-) REST API


bluthg:
**Describe the bug**
Since setting up HAProxy (with check-ssl), we're being flooded with messages like
```
2022-02-25 11:54:58,601 WARNING: Exception happened during processing of request from <redacted>
2022-02-25 11:54:58,603 WARNING: Traceback (most recent call last):
  File "/usr/lib/python3.8/socketserver.py", line 683, in process_request_thread
    self.finish_request(request, client_address)
  File "/usr/lib/python3.8/socketserver.py", line 360, in finish_request
    self.RequestHandlerClass(request, client_address, self)
  File "/usr/lib/python3.8/socketserver.py", line 747, in __init__
    self.handle()
  File "/usr/lib/python3.8/http/server.py", line 427, in handle
    self.handle_one_request()
  File "/usr/lib/python3/dist-packages/patroni/api.py", line 642, in handle_one_request
    BaseHTTPRequestHandler.handle_one_request(self)
  File "/usr/lib/python3.8/http/server.py", line 415, in handle_one_request
    method()
  File "/usr/lib/python3/dist-packages/patroni/api.py", line 178, in do_GET
    self._write_status_response(status_code, response)
  File "/usr/lib/python3/dist-packages/patroni/api.py", line 94, in _write_status_response
    self._write_json_response(status_code, response)
  File "/usr/lib/python3/dist-packages/patroni/api.py", line 53, in _write_json_response
    self._write_response(status_code, json.dumps(response, default=str), content_type='application/json')
  File "/usr/lib/python3/dist-packages/patroni/api.py", line 50, in _write_response
    self.wfile.write(body.encode('utf-8'))
  File "/usr/lib/python3.8/socketserver.py", line 826, in write
    self._sock.sendall(b)
  File "/usr/lib/python3.8/ssl.py", line 1204, in sendall
    v = self.send(byte_view[count:])
  File "/usr/lib/python3.8/ssl.py", line 1173, in send
    return self._sslobj.write(data)
ConnectionResetError: [Errno 104] Connection reset by peer
```
Did not check if the same happens when the REST API is not TLS-enabled!

**To Reproduce**
* Configure a HAProxy with e.g.
`server xxxxx_5432 x.x.x.x:5432 maxconn 1000 check check-ssl ca-file XXXXXXXXX.pem port xxxx`
* `tail -f  var/log/{syslog,messages}`

**Expected behavior**
[...]

PR #3647, opened 2026-06-23
Log REST API connection resets at DEBUG instead of WARNING

kylemclaren:
## What

Patroni's REST API logs a `WARNING` with a full Python stack trace every time a client resets the TCP connection before the response has been fully written:

```
WARNING: Exception happened during processing of request from <ip>
...
  File ".../patroni/api.py", line 179, in write_response
    self.wfile.write(body.encode('utf-8'))
...
ConnectionResetError: [Errno 104] Connection reset by peer
```

This happens routinely whenever a load balancer / health-checker (HAProxy, a cloud synthetic probe, a Kubernetes readiness probe, ...) closes the connection right after reading the health-check response. The node and the cluster are perfectly healthy, the only impact is log noise, which buries genuine warnings and can trip log-based alerting.

## Why

`write_response()` already carried a standing `# TODO: try-catch ConnectionResetError: [Errno 104] Connection reset by peer and log it in DEBUG level`. This PR implements it.

The exception is raised by `self.wfile.write()` (and the `end_headers()` flush — both `sendall()` to the socket), propagates up through `do_GET` → `handle_one_request`, and is finally logged by `handle_error()` as a WARNING plus a traceback.

## What this changes

Wrap the response-writing in `RestApiHandler.write_response()` in `try/except ConnectionError` and log it at `DEBUG` instead. `ConnectionError` is the base class of `ConnectionResetError`, `BrokenPipeError` and `ConnectionAbortedError`, so the common "client went away mid-write" variants are all covered. Any genuine (non-connection) error still propagates exactly as before.

## Test

Added `tests/test_api.py::test_write_response_connection_reset`: it makes the mocked socket `sendall` raise `ConnectionResetError(104)` and asserts the handler swallows it and logs at DEBUG rather than propagating. Verified it **fails with the original `ConnectionResetError` without the fix and passes with it**. Full `tests/test_api.py` suite: 37 passed; `flake8` clean.

Refs: 
- patroni/patroni#3313
- patroni/patroni#2278
- patroni/patroni#2229

### DIFF
--- a/tests/test_api.py
+++ b/tests/test_api.py
@@ -796,6 +796,20 @@ class TestRestApiServer(unittest.TestCase):
         except Exception:
             self.assertIsNone(self.srv.handle_error(None, ('127.0.0.1', 55555)))
 
+    def test_finish_request_connection_reset(self):
+        import ssl
+        # A client (e.g. a load-balancer performing health-checks) may reset the connection at any
+        # point while the request is handled, not only in write_response(). finish_request() must
+        # swallow a ConnectionError (plain HTTP) or ssl.SSLError (TLS) raised anywhere during handling
+        # and log it at DEBUG, instead of letting it propagate to handle_error() as a WARNING.
+        for exc in (ConnectionResetError(104, 'Connection reset by peer'), ssl.SSLError('reset')):
+            self.srv.RequestHandlerClass = Mock(side_effect=exc)
+            with patch('patroni.api.logger.debug') as mock_debug:
+                self.assertIsNone(self.srv.finish_request(Mock(), ('127.0.0.1', 55555)))
+            self.srv.RequestHandlerClass.assert_called_once()
+            mock_debug.assert_called_once()
+            self.assertIn('was reset', mock_debug.call_args[0][0])
+
     @patch.object(HTTPServer, '__init__', Mock(side_effect=socket.error))
     def test_socket_error(self):
         self.assertRaises(socket.error, MockRestApiServer, Mock(), '', {'listen': '*:8008'})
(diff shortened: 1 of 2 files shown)


CyberDem0n (changes requested):
It is not the only place where exception should be handled.
To cover everything please override finish_request() method in RestApiServer class.
It should call method from the parent class and intercept ConnectionError

kylemclaren:
@CyberDem0n thanks for the review! I moved the handling out of `write_response()` and into a `finish_request()` override on `RestApiServer` that calls the parent implementation and intercepts `ConnectionError`, logging it at DEBUG. As you pointed out, this now covers a reset from any write site during request handling (`write_response`, `_write_status_code_only`, `send_error`, ...) rather than just `write_response`.

One follow-up I'd like your call on: on a **TLS-enabled** REST API a client reset surfaces as `ssl.SSLError` / `ssl.SSLEOFError`, which are `OSError` subclasses but **not** `ConnectionError`, so those would still reach `handle_error()` (relevant to #2229 / #2278). Would you prefer I widen the `except` to `(ConnectionError, ssl.SSLError)` here to cover the HTTPS case too, or keep this PR scoped to `ConnectionError` and address the SSL variant separately?

CyberDem0n:
@kylemclaren lets do it here.
Actually, process_request_thread()->request.do_handshake() may probably also raise something

kylemclaren:
@CyberDem0n moved both `finish_request()` and the `do_handshake()` to catch OSError, which should cover TLS. And yes, a failed handshake was skipping `shutdown_request()` and leaking the socket, so it now shuts down and logs at DEBUG there too.

tests/test_api.py
```diff
@@ -797,6 +797,20 @@ def test_handle_error(self):
         except Exception:
             self.assertIsNone(self.srv.handle_error(None, ('127.0.0.1', 55555)))
 
+    def test_finish_request_connection_reset(self):
+        import ssl
```
hughcapet, line 801:
```suggestion
        import ssl

```

tests/test_api.py
```diff
@@ -818,6 +832,25 @@ def test_process_request(self):
         with patch.object(self.srv._executor, 'submit', lambda f, r, c: f(r, c)):
             self.srv.process_request(self.__create_socket(), ('2', 54321))
 
+    def test_process_request_thread_ssl_handshake_reset(self):
+        import ssl
```
hughcapet, line 837:
```suggestion
        import ssl

```

hughcapet (approved):

CyberDem0n (approved):

PR #3655, opened 2026-06-28
Add `patroni_logical_slot_active` Prometheus metric

kylorend3r:
## The Idea Behind Adding the Metric

``patroni_logical_slot_active`` is a per-slot gauge that exposes the ``active`` field from ``pg_catalog.pg_replication_slots`` for every logical slot on the node. The value is ``1`` when a consumer is connected and ``0`` when the slot has no active consumer. The metric is emitted only when PostgreSQL is reachable and at least one logical slot exists; if the query fails, the metric is omitted for that scrape and all other metrics are unaffected.

Logical replication slots are a common source of unplanned disk-full incidents: an abandoned CDC connector leaves a slot with no consumer, and PostgreSQL retains WAL indefinitely until it reconnects.On Patroni clusters this is compounded by failover risk on PostgreSQL < 17, slots not tracked by Patroni are silently lost when the primary changes.

The postgres_exporter already exposes raw slot metrics, but it has no awareness of Patroni's DCS configuration and cannot distinguish slots that Patroni manages from slots created directly by applications.

Each series carries a `managedby` label derived from Patroni's own `permanent_slots` DCS config:

| `managedby` | Source | Failover behaviour |
|-------------|--------|--------------------|
| `patroni` | Listed in `permanent_slots` in DCS config | Patroni recreates the slot on the new primary automatically |
| `other` | Created directly by an application, not in Patroni config | **Lost after failover on PG < 17. Never cleaned up. WAL accumulates if consumer disconnects.** |

This makes it possible to write alerts that were previously impossible:
```
    # WAL accumulating — unmanaged slot has no consumer
    patroni_logical_slot_active{managedby="other"} == 0

    # Active unmanaged slots that will break on failover (PG < 17)
    patroni_logical_slot_active{managedby="other"} == 1
```

The metric is emitted only when PostgreSQL is reachable and at least one logical slot exists. If the slot query fails the metric is omitted for that scrape and all other metrics are unaffected.

## The Example Metric Output

```
# HELP patroni_logical_slot_active Value is 1 if the logical replication slot has an active consumer, 0 otherwise.
# TYPE patroni_logical_slot_active gauge
patroni_logical_slot_active{scope="demo",name="patroni2",slot="patroni_cdc_slot",managedby="patroni"} 1
patroni_logical_slot_active{scope="demo",name="patroni2",slot="unmanaged_slot",managedby="other"} 0
```

### DIFF
--- a/docs/rest_api.rst
+++ b/docs/rest_api.rst
@@ -465,20 +465,6 @@ The ``managedby`` label bridges this gap: it is derived by cross-referencing eac
 against ``permanent_slots`` — information already in memory inside the Patroni process,
 requiring no extra DCS call.
 
-This makes it possible to write alert rules that were previously not expressible:
-
-.. code-block:: yaml
-
-    # Unmanaged slot has lost its consumer — WAL will accumulate indefinitely
-    - alert: LogicalSlotOtherInactive
-      expr: patroni_logical_slot_active{managedby="other"} == 0
-      for: 5m
-
-    # Active unmanaged slots exist — these will be lost on failover (PG < 17)
-    - alert: LogicalSlotOtherActivePreFailover
-      expr: patroni_logical_slot_active{managedby="other"} == 1
-      for: 0m
-
 
 Cluster status endpoints
 ------------------------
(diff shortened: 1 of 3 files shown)


hughcapet:
we do not want to have this in Patroni. One should have a proper monitoring for logical replication slots, while Patroni is not a monitoring system

PR #3658, opened 2026-06-29
Honor primary_stop_timeout during primary restart

### LINKED ISSUE #3657: REST API restart does not apply primary_stop_timeout when stopping a primary
### What happened?

When Patroni restarts PostgreSQL through the REST API `/restart` path, `Postgresql.restart()` calls `Postgresql.stop()` without passing `stop_timeout`.

That differs from the demotion/shutdown paths, which apply `primary_stop_timeout()` when stopping a primary.

As a result, a primary restart requested through `/restart` can wait indefinitely for PostgreSQL shutdown, even when `primary_stop_timeout` is configured and the same stop during demotion/shutdown would be bounded.

### Why I think this is unexpected

The dynamic configuration docs describe `primary_stop_timeout` as:

> The number of seconds Patroni is allowed to wait when stopping Postgres and effective only when synchronous_mode is enabled. When set to > 0 and the synchronous_mode is enabled, Patroni sends SIGKILL to the postmaster if the stop operation is running for more than the value set by primary_stop_timeout.

https://patroni.readthedocs.io/en/latest/dynamic_configuration.html

The `/restart` endpoint is documented as the way to restart PostgreSQL on a specific node:

https://patroni.readthedocs.io/en/latest/rest_api.html#restart-endpoint

So I expected a primary restart to use the same stop timeout semantics as demotion/shutdown, or for the difference to be documented explicitly.

### How can we reproduce it?

I created a small Docker-based reproduction with more details:

https://gist.github.com/sebastianwebber/9ff908975ece526d801f03769f36f14d

It builds a local image with Patroni, instruments `Postgresql.stop()`, calls `Postgresql.restart(timeout=30, block_callbacks=True, role="primary")`, and prints the arguments passed to `stop()`.

Relevant output:

```text
Expected:
  Postgresql.restart() should call stop(..., stop_timeout=<primary_stop_timeout>)

Actual:
  Postgresql.restart() called stop(..., stop_timeout=None)
  Full stop() kwargs: {'block_callbacks': True, 'before_shutdown': None}

Result: BUG REPRODUCED
  restart() does not pass stop_timeout to stop().
  Therefore, a primary restart cannot apply primary_stop_timeout in this path.
```

I also confirmed this end-to-end with a disposable three-node Patroni cluster by setting:

```yaml
synchronous_mode: quorum
primary_stop_timeout: 31
```

Then I stopped a normal SQL backend on the leader with `SIGSTOP` and called:

```bash
curl -v --max-time 90 -X POST http://127.0.0.1:8008/restart \
  -H 'Content-Type: application/json' \
  -d '{"role":"primary"}'
```

The request timed out after 90 seconds with no response, even though `primary_stop_timeout` was set to 31 seconds. After resuming the stopped backend with `SIGCONT`, the node returned to `running`.

### Code references

`Postgresql.stop()` accepts `stop_timeout` and uses it to bound shutdown behavior:

https://github.com/patroni/patroni/blob/v4.1.0/patroni/postgresql/__init__.py#L844-L861

`Postgresql.restart()` calls `stop()` without `stop_timeout`:

https://github.com/patroni/patroni/blob/v4.1.0/patroni/postgresql/__init__.py#L1024-L1039

`Ha.primary_stop_timeout()` returns the configured timeout only when synchronous mode is enabled:

https://github.com/patroni/patroni/blob/v4.1.0/patroni/ha.py#L267-L270

The demotion path passes `stop_timeout=self.primary_stop_timeout()`:

https://github.com/patroni/patroni/blob/v4.1.0/patroni/ha.py#L1611-L1615

The restart path calls `self.state_handler.restart(...)` without threading a stop timeout:

https://github.com/patroni/patroni/blob/v4.1.0/patroni/ha.py#L1905-L1938

When restarting a primary, Patroni should apply `primary_stop_timeout()` to the stop phase, matching demotion/shutdown behavior.

### What happened instead?

The `/restart` path does not pass any stop timeout to `Postgresql.stop()`, so the configured `primary_stop_timeout` is not applied.

Confirmed against Patroni `4.1.0` and `4.1.2`.

This appears independent of PostgreSQL major version because the issue is in Patroni’s restart call path.

This seems related to, but distinct from, https://github.com/patroni/patroni/issues/3639.

That issue concerns watchdog behavior after `primary_stop_timeout` expires. This issue is about the `/restart` path not applying `primary_stop_timeout` at all.

Would you be open to a PR that threads an optional `stop_timeout` through `Postgresql.restart()` and passes `primary_stop_timeout()` from the HA restart path when restarting a primary?

sebawebber:
## Summary
- Apply `primary_stop_timeout` to primary restarts by passing it through `Ha.restart()` and `Postgresql.restart()` into `Postgresql.stop()`.
- Add unit coverage for the HA restart path and a Behave scenario that verifies `/restart` returns while a backend is stuck during shutdown.

Fixes #3657

## Test plan
- `python -m flake8 features/environment.py features/steps/patroni_api.py patroni/ha.py patroni/postgresql/__init__.py tests/test_ha.py`
- `python -m pytest tests/test_ha.py -q`
- `python -m pytest tests/test_postgresql.py -q`
- `python -m behave features/patroni_api.feature --name 'primary restart honors primary_stop_timeout'`

### DIFF
--- a/features/environment.py
+++ b/features/environment.py
@@ -345,8 +345,9 @@ class PatroniController(AbstractController):
         except Exception:
             return None
 
-    def patroni_hang(self, timeout):
-        hang = ProcessHang(self._handle.pid, timeout)
+    def patroni_hang(self, timeout, pid=None):
+        pid = pid or self._handle.pid
+        hang = ProcessHang(pid, timeout)
         self._closables.append(hang)
         hang.start()
 
@@ -386,14 +387,17 @@ class ProcessHang(object):
         self.timeout = timeout
 
     def start(self):
+        os.kill(self.pid, signal.SIGSTOP)
         self._thread.start()
 
     def run(self):
-        os.kill(self.pid, signal.SIGSTOP)
         try:
             self._cancelled.wait(self.timeout)
         finally:
-            os.kill(self.pid, signal.SIGCONT)
+            try:
+                os.kill(self.pid, signal.SIGCONT)
+            except ProcessLookupError:
+                pass
 
     def close(self):
         self._cancelled.set()
--- a/features/patroni_api.feature
+++ b/features/patroni_api.feature
@@ -60,6 +60,17 @@ Scenario: check the scheduled restart
 	And Response on GET http://127.0.0.1:8008/patroni does not contain pending_restart after 10 seconds
 	And postgres-0 role is the primary after 10 seconds
 
+Scenario: primary restart honors primary_stop_timeout
+	Given I start postgres-0
+	And postgres-0 is a leader after 10 seconds
+	When I issue a PATCH request to http://127.0.0.1:8008/config with {"synchronous_mode": true, "primary_stop_timeout": 3}
+	Then I receive a response code 200
+	And Response on GET http://127.0.0.1:8008/config contains primary_stop_timeout=3 after 10 seconds
+	When I suspend a backend on postgres-0 for 20 seconds
+	And I issue a POST request to http://127.0.0.1:8008/restart with {"role": "primary"} and client timeout 12 seconds
+	Then I receive a response code 200
+	And postgres-0 role is the primary after 10 seconds
+
 Scenario: check API requests for the primary-replica pair in the pause mode
 	Given I start postgres-1
 	Then replication works from postgres-0 to postgres-1 after 20 seconds
--- a/features/steps/patroni_api.py
+++ b/features/steps/patroni_api.py
@@ -70,6 +70,19 @@ def do_post_empty(context, url):
     do_request(context, 'POST', url, None)
 
 
+@step('I issue a {request_method:w} request to {url:url} with {data} and client timeout {timeout:d} seconds')
+def do_request_with_timeout(context, request_method, url, data, timeout):
+    if context.certfile:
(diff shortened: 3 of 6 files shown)


CyberDem0n (commented):
Sorry, but no.
I explained yesterday in the open [issue](https://github.com/patroni/patroni/issues/3657) and copying my reply here:

---
The thing is, that in case of restart our intention is to do a restart, not failover.
When stop is taking too long using `pg_ctl stop -m immediate` or just killing all postgres processes is not the best option.
After unclean shutdown recovery may take significant time (much longer than stop).

I know only two reasons whey shutdown may hang:
1. Postgres is flushing dirty buffers
2. Postgres is waiting for replication connections to confirm

Especially bad could be misbehaving logical replication clients (e.g. Debezium).

Maybe in case of restart we can wait for pg_controldata reporting `'Database cluster state: shut down'` and safely terminate remaining processes (replication connections).

hughcapet:
@sebastianwebber will you continue working on this pr, implementing 
 
 > Maybe in case of restart we can wait for pg_controldata reporting 'Database cluster state: shut down' and safely terminate remaining processes (replication connections).
 
 or should we close this?

PR #3660, opened 2026-06-30
Refactor connection options

CyberDem0n:
In many places we are explicitly using `connect_timeout=3` and `options='-c statement_timeout=2000'`, which are essentially a default.
Therefore, it is logical to use them as default when setting `ConnectionPool.conn_kwargs` or when calling `get_connection_cursor()` function.

We explicitly pass `options` and `connect_timeout` only in case we want to override defaults.

### DIFF
--- a/patroni/postgresql/__init__.py
+++ b/patroni/postgresql/__init__.py
@@ -1099,6 +1099,7 @@ class Postgresql(object):
     @contextmanager
     def get_replication_connection_cursor(self, host: Optional[str] = None, port: Union[int, str] = 5432,
                                           **kwargs: Any) -> Generator[Union['cursor', 'Cursor[Any]'], None, None]:
+        # We use RemoteMember here because it has the logic to map and sanitize connection parameters
         member = RemoteMember('', {'conn_kwargs': {'host': host, 'port': port}})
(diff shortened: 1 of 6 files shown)


hughcapet (dismissed):

patroni/postgresql/__init__.py
```diff
@@ -1103,9 +1099,9 @@ def waldump(self, timeline: Union[int, str], lsn: str, limit: int) -> Tuple[Opti
     @contextmanager
     def get_replication_connection_cursor(self, host: Optional[str] = None, port: Union[int, str] = 5432,
                                           **kwargs: Any) -> Generator[Union['cursor', 'Cursor[Any]'], None, None]:
-        conn_kwargs = self.config.replication.copy()
-        conn_kwargs.update(host=host, port=int(port) if port else None, user=conn_kwargs.pop('username'),
-                           connect_timeout=3, replication=1, options='-c statement_timeout=2000')
+        member = RemoteMember('', {'conn_kwargs': {'host': host, 'port': port}})
```
hughcapet, line 1103: can we have a comment here why RemoteMember is used

hughcapet (approved):

PR #3673, opened 2026-07-09
Stream pg_rewind and pg_basebackup output to logs while running

### LINKED ISSUE #3543: pg_rewind progress output
### What happened?

I saw in #3484 that support was added for custom pg_rewind options, including `--progress`, but it's not possible to see the output of pg_rewind until it exits.

I understand that patroni is capturing the output in order to parse it for some messages, but it would be helpful if the user could see the output in the logs as it runs as well. In particular, progress reporting would be nice to have for large rewinds.

For some context, in my use case we will be deliberately promoting and running a cluster member outside of patroni's control for some period of time, before handing back to patroni to rejoin the cluster. In this case the rewind takes a good amount of time, so having progress information would be helpful.

My current workaround is to write a wrapper script as the `bin_name.pg_rewind` config that adds the `-P` flag (because the latest release doesn't have the changes in #3484), and tees the stdout and stderr like this
```bash
#!/bin/bash
[ "$1" == '--help' ] && exec pg_rewind "$@"

exec pg_rewind "$@" -P \
  1> >(tee >(systemd-cat -t pg_rewind -p info)) \
  2> >(tee >(systemd-cat -t pg_rewind -p info) >&2)
```

Would a PR that adds logging for the pg_rewind stdout/stderr as it is received be welcome?

n/a

n/a

n/a

```yaml
n/a
```

### patronictl show-config

```yaml
n/a
```

### Patroni log files

```shell
n/a
```

### PostgreSQL log files

```shell
n/a
```

- [x] Yes

_No response_

Sharvash:
Closes https://github.com/patroni/patroni/issues/3543

Rebased on top of master now that #3684 (the `stream_cb` plumbing in `CancellableSubprocess`) is merged, so this PR is the consumer side plus one fix in the reader.

When the `progress` option is present in the custom `pg_rewind` options (supported since #3484) or in the `basebackup` options, Patroni streams the output of the command to its log as it arrives, one line at a time, prefixed with `pg_rewind:` / `pg_basebackup:`, instead of dumping it only after the process exits (pg_rewind) or writing it to the terminal Patroni inherited (pg_basebackup).

Implementation notes:

- `Rewind._log_output_line()` and `Bootstrap._log_output_line()` are passed as `stream_cb` to `CancellableSubprocess.call()` only when `--progress` is among the user options. Without it nothing changes: `stream_cb` is None, pg_basebackup stdio is not redirected at all and pg_rewind still logs the captured stdout/stderr after exit.
- Streamed lines are logged at INFO, the level already used for the post-exit dump. For pg_rewind that post-exit `stdout=`/`stderr=` dump is skipped while streaming, so the output is not logged twice.
- `communicate` still returns the byte-exact output of each pipe separately, so `_find_missing_wal()` and its pre-PG15 retry loop are unaffected.
- Progress reports of pg_basebackup (PostgreSQL 10 and older) and pg_rewind (PostgreSQL 11 and older) are terminated with a bare CR rather than LF, because those versions do not check whether the output is a terminal. The reader in `CancellableSubprocess` therefore splits records on LF, CRLF and CR; otherwise the whole progress of these versions would reach the log only at exit. The bytes handed back through `communicate` keep the original line endings.

Verified with unit tests, including one that proves a CR-terminated record is delivered while the process is still writing, with 100% coverage of `patroni/postgresql/cancellable.py`, and against a real subprocess emitting CR-terminated progress through a pipe.

### DIFF
--- a/patroni/postgresql/cancellable.py
+++ b/patroni/postgresql/cancellable.py
@@ -1,4 +1,5 @@
 import logging
+import re
 import subprocess
 
 from io import BufferedReader
@@ -12,6 +13,11 @@ from patroni.exceptions import PostgresException
 from patroni.utils import polling_loop
 
 logger = logging.getLogger(__name__)
+
+# Records are terminated with LF, CRLF (Windows), or a bare CR. The latter is used by progress reports of
+# pg_basebackup (PostgreSQL 10 and older) and pg_rewind (PostgreSQL 11 and older), which don't check whether
+# the output is a terminal and therefore keep writing CR even when it is a pipe.
+LINE_SEPARATOR_RE = re.compile(b'\r\n|\r|\n')
 
 
 class CancellableExecutor(object):
@@ -83,8 +89,9 @@ class CancellableSubprocess(CancellableExecutor):
 
         ``stderr`` is drained by a single helper thread while ``stdout`` is read inline, so both pipes are drained
         concurrently with only one extra thread. ``read1()`` returns as soon as any data is available, so slowly
-        produced lines are delivered promptly. Exceptions from ``stream_cb`` are swallowed so streaming continues;
-        only read/OS errors abort (after both streams are drained).
+        produced lines are delivered promptly. Lines are split on LF, CRLF, or a bare CR, so that progress reports
+        of old ``pg_basebackup``/``pg_rewind`` versions are streamed as well. Exceptions from ``stream_cb`` are
+        swallowed so streaming continues; only read/OS errors abort (after both streams are drained).
 
         :param stream_cb: called as ``stream_cb(stream_name, line)`` for every complete line of *stdout*/*stderr*.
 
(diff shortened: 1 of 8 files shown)


patroni/postgresql/cancellable.py
```diff
@@ -76,11 +78,98 @@ def __init__(self) -> None:
...
+        stream_output: Optional[Callable[[str], None]] = kwargs.pop('stream_output', None)
+        communicate: Optional[Dict[str, Any]] = kwargs.pop('communicate', None)
+        if stream_output is not None:
+            if not isinstance(communicate, dict):
+                raise ValueError('stream_output requires a communicate dictionary')
+            if communicate.get('input'):
+                raise ValueError('stream_output is incompatible with communicate input')
+            if kwargs.get('text') or kwargs.get('universal_newlines') \
+                    or kwargs.get('encoding') is not None or kwargs.get('errors') is not None:
+                raise ValueError('stream_output is incompatible with text mode')
```
CyberDem0n: I don't think we should be spending time on all these checks and writing tests for it.
It is enough to have a check on line https://github.com/patroni/patroni/pull/3673/changes#diff-55590afbb5df4f18d02162743bdac7927455ee6cada51e9276f3e486a4d892e3R195

patroni/postgresql/cancellable.py
```diff
@@ -102,12 +192,22 @@ def call(self, *args: Any, **kwargs: Any) -> Optional[int]:
...
+                process_finished = True
+                return return_code
         finally:
+            if not process_finished:
+                self._kill_process()
             with self._lock:
+                process = self._process
                 self._process = None
+            if process is not None and not process_finished:
+                process.wait()
```
CyberDem0n: Why was it added?

patroni/postgresql/cancellable.py
```diff
@@ -76,11 +78,98 @@ def __init__(self) -> None:
...
+                name, buf = pipes[fd]
+                if data:
+                    output[name].extend(data)
+                    self._stream_data(data, buf, stream_output)
+                else:  # EOF
+                    poller.unregister(fd)
+                    del pipes[fd]
+                    self._stream_lines([bytes(buf)], stream_output)
+
+        communicate['stdout'], communicate['stderr'] = bytes(output['stdout']), bytes(output['stderr'])
```
CyberDem0n: I don't particularly like the fact that it doesn't work on Windows.
There is a nice way of doing it without using non-blocking mode by starting one additional thread.
See: https://github.com/patroni/patroni/pull/3684
Once it is merged, - you can do the rest.
E.g., use it for pg_rewind and for pg_basebackup, which also supports `--progress` option.

hughcapet:
@Sharvash are you planning to keep working on this PR?

Sharvash:
> @Sharvash are you planning to keep working on this PR?

Yes - rebasing on #3684 and cutting this down to the two call sites

patroni/postgresql/cancellable.py
Sharvash: Gone. #3684 has just the one gate you pointed at

patroni/postgresql/cancellable.py
Sharvash: The finally block is back to what it was

patroni/postgresql/cancellable.py
Sharvash: Done: the thread-based reader from #3684, now used for both `pg_rewind` and `pg_basebackup`

Sharvash:
Rebased on current master, so the thread-based reader from #3684 is in place and now used for both pg_rewind and pg_basebackup, gated on --progress. One addition on top: the reader also splits on a bare CR now, since pg_basebackup up to PG 10 and pg_rewind up to PG 11 end progress reports with CR and nothing reached the log before exit. The earlier comments about the validation checks and the extra process.wait() are moot, that code is gone with the old implementation.

hughcapet (approved):

PR #3679, opened 2026-07-16
feat: Add topology-aware synchronous standby selection

### LINKED ISSUE #2157: Patroni with Two DataCenters and sync/async replication
Hi,

I am implementing a Patroni PostgreSQL cluster between two data centers.
I would like to have the Leader and one sync Standby on DC1 and remote async replica on DC2 with a sync replica for it in order to have HA for the second DC.

I am using the Patroni synchronous_mode: true, however it seems that there is no distinguish between the datacenters and mode of replication.

How can I achieve such a desired mode of replication between Data Center and inside each Data Center ?
Is it something that Patroni can managed with the synchronous_mode, or requires the PostgreSQL replication parameters ?

Thanks,

Ori

erayaraz:
Description
This PR introduces a new global configuration property synchronous_node_topology to enable topology-aware synchronous standby selection. This directly addresses the feature request described in #2157.

The Problem: Currently, Patroni selects the synchronous standby strictly based on the minimal replication lag. In multi-AZ or multi-DC deployments, the node with the lowest lag is almost always located in the same availability zone or rack as the primary due to network proximity. This defeats the primary purpose of multi-AZ disaster recovery: if an entire zone goes down, both the primary and its synchronous replica are lost simultaneously, potentially leading to data loss or a split-brain scenario.

The Solution: By utilizing the existing node tags feature, users can now enforce topological constraints when electing a synchronous standby. For instance, setting strategy: "different" and key: "dc" ensures that the synchronous standby MUST be located in a different datacenter than the primary, guaranteeing cross-AZ data redundancy before transactions are committed.

Implementation Details
Configuration: Added synchronous_node_topology under bootstrap.dcs and exposed it as a property in GlobalConfig.
Filtering Logic: Modified _apply_topology_filter inside sync.py to filter candidates based on the primary's tag value before evaluating their replication lag.
Graceful Fallback: Added a safety net. If the topology filter excludes all available candidates (e.g., all nodes in the other DC are temporarily down), the logic gracefully falls back to the unfiltered candidate list. This ensures the cluster doesn't get stuck waiting for a synchronous commit when cross-AZ nodes are unavailable.
Validation: Updated the schema validation in validator.py using EnumValidator to enforce strict values (different or same).
Tests: Added comprehensive unit tests in test_sync.py to cover all topology filtering branches, ensuring 100% test coverage for the new feature.

How to test
This feature can be verified by applying the following configuration:

bootstrap:
  dcs:
    synchronous_node_topology:
      key: "dc"
      strategy: "different"

If we have a cluster where patroni3 (in dc-b) is the Leader, Patroni will bypass patroni2 (in dc-a) if there is another node in dc-b, but if configured with strategy: "same", it will prefer the node in the same DC.

When strategy: "different" (Cross-DC Sync): Even if patroni2 has minimal lag, Patroni elects patroni1 as the sync standby because they are in different datacenters (dc-b vs dc-a).

+ Cluster: demo (7662465560473440284) +-----------+----+-------------+-----+------------+-----+----------+
| Member   | Host      | Role         | State     | TL | Receive LSN | Lag | Replay LSN | Lag | Tags     |
+----------+-----------+--------------+-----------+----+-------------+-----+------------+-----+----------+
| patroni1 | 10.89.0.6 | Sync Standby | streaming |  5 |   0/BB7D028 |   0 |  0/BB7D028 |   0 | dc: dc-a |
| patroni2 | 10.89.0.9 | Replica      | streaming |  5 |   0/BB7D028 |   0 |  0/BB7D028 |   0 | dc: dc-a |
| patroni3 | 10.89.0.8 | Leader       | running   |  5 |             |     |            |     | dc: dc-b |
+----------+-----------+--------------+-----------+----+-------------+-----+------------+-----+----------+

When strategy: "same" (Same-DC Sync): If patroni1 were the Leader in dc-a, Patroni would specifically elect patroni2 (also in dc-a) as the sync standby, bypassing patroni3 in dc-b.

+ Cluster: demo (7662465560473440284) +-----------+----+-------------+-----+------------+-----+----------+
| Member   | Host      | Role         | State     | TL | Receive LSN | Lag | Replay LSN | Lag | Tags     |
+----------+-----------+--------------+-----------+----+-------------+-----+------------+-----+----------+
| patroni1 | 10.89.0.6 | Leader       | running   |  5 |             |     |            |     | dc: dc-a |
| patroni2 | 10.89.0.9 | Sync Standby | streaming |  5 |   0/BB7D028 |   0 |  0/BB7D028 |   0 | dc: dc-a |
| patroni3 | 10.89.0.8 | Replica      | streaming |  5 |   0/BB7D028 |   0 |  0/BB7D028 |   0 | dc: dc-b |
+----------+-----------+--------------+-----------+----+-------------+-----+------------+-----+----------+

Related Issues
Fixes #2157

### DIFF
--- a/patroni/global_config.py
+++ b/patroni/global_config.py
@@ -163,6 +163,25 @@ class GlobalConfig(types.ModuleType):
         return max(self.get_int('synchronous_node_count', 1), self.min_synchronous_nodes)
 
     @property
+    def synchronous_node_topology(self) -> Optional[Dict[str, str]]:
+        """Currently configured value of ``synchronous_node_topology`` from the global configuration.
+
+        If configured, returns a dictionary with ``key`` and ``strategy`` fields.
+        The ``key`` specifies which member tag to use for topology comparison (e.g., ``dc``, ``zone``).
+        The ``strategy`` can be ``different`` (sync standby must be in a different topology group than the primary)
+        or ``same`` (sync standby must be in the same topology group as the primary).
+
+        When all candidates matching the topology filter are unavailable, the filter is bypassed
+        to prevent the cluster from becoming stuck.
+
+        :returns: topology configuration dictionary or ``None`` if not configured or invalid.
+        """
+        value = self.get('synchronous_node_topology')
+        if isinstance(value, dict) and value.get('key') and value.get('strategy') in ('different', 'same'):
+            return {'key': str(value['key']), 'strategy': str(value['strategy'])}
+        return None
+
+    @property
     def maximum_lag_on_failover(self) -> int:
         """Currently configured value of ``maximum_lag_on_failover`` from the global configuration.
 
--- a/patroni/postgresql/sync.py
+++ b/patroni/postgresql/sync.py
@@ -239,6 +239,7 @@ class _ReplicaList(List[_Replica]):
 
         members = CaseInsensitiveDict({m.name: m for m in cluster.members
                                        if m.is_running and not m.nosync and m.name.lower() != postgresql.name.lower()})
+        members = _ReplicaList._apply_topology_filter(members, cluster, postgresql)
         replication = CaseInsensitiveDict({row['application_name']: row for row in postgresql.pg_stat_replication()
                                            if row[sort_col] is not None and row['application_name'] in members})
         for row in replication.values():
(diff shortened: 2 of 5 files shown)


hughcapet:
Thank you for working on that.
But unfortunately it is more complex than this and we are already working on providing a solution for site-aware sync replication and failover

PR #3682, opened 2026-07-21
Prevent unclean shutdown when postmaster hangs during restart

### LINKED ISSUE #3657: REST API restart does not apply primary_stop_timeout when stopping a primary

erayaraz:
Previously, if stop_timeout was reached during a restart, Patroni would unconditionally send SIGKILL to the postmaster. This caused an unclean shutdown and triggered crash recovery if PostgreSQL was legitimately blocked (e.g. waiting for synchronous standby).

This commit introduces a safe_kill option. During a restart, Patroni now checks pg_controldata before sending SIGKILL. If the database is not in the "shut down" state, Patroni skips the kill to prevent crash recovery and fails the restart gracefully. If it is safely "shut down" but the postmaster is hung, Patroni proceeds with the kill.

Closes #3657

### DIFF
--- a/patroni/postgresql/__init__.py
+++ b/patroni/postgresql/__init__.py
@@ -854,7 +854,8 @@ class Postgresql(object):
 
     def stop(self, mode: str = 'fast', block_callbacks: bool = False, checkpoint: Optional[bool] = None,
              on_safepoint: Optional[Callable[..., Any]] = None, on_shutdown: Optional[Callable[[int, int], Any]] = None,
-             before_shutdown: Optional[Callable[..., Any]] = None, stop_timeout: Optional[int] = None) -> bool:
+             before_shutdown: Optional[Callable[..., Any]] = None, stop_timeout: Optional[int] = None,
+             safe_kill: bool = False) -> bool:
         """Stop PostgreSQL
 
         Supports a callback when a safepoint is reached. A safepoint is when no user backend can return a successful
@@ -864,12 +865,13 @@ class Postgresql(object):
         :param on_safepoint: This callback is called when no user backends are running.
         :param on_shutdown: is called when pg_controldata starts reporting `Database cluster state: shut down`
         :param before_shutdown: is called after running optional CHECKPOINT and before running pg_ctl stop
+        :param safe_kill: if True, only terminate postmaster after a clean shut down (used for restart)
         """
         if checkpoint is None:
             checkpoint = False if mode == 'immediate' else True
 
         success, pg_signaled = self._do_stop(mode, block_callbacks, checkpoint, on_safepoint,
-                                             on_shutdown, before_shutdown, stop_timeout)
+                                             on_shutdown, before_shutdown, stop_timeout, safe_kill)
         if success:
             # block_callbacks is used during restart to avoid
             # running start/stop callbacks in addition to restart ones
@@ -884,7 +886,8 @@ class Postgresql(object):
 
     def _do_stop(self, mode: str, block_callbacks: bool, checkpoint: bool,
                  on_safepoint: Optional[Callable[..., Any]], on_shutdown: Optional[Callable[[int, int], Any]],
-                 before_shutdown: Optional[Callable[..., Any]], stop_timeout: Optional[int]) -> Tuple[bool, bool]:
+                 before_shutdown: Optional[Callable[..., Any]], stop_timeout: Optional[int],
+                 safe_kill: bool = False) -> Tuple[bool, bool]:
         postmaster = self.is_running()
         if not postmaster:
             if on_safepoint:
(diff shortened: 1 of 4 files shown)


patroni/postgresql/__init__.py
```diff
@@ -941,8 +946,16 @@ def _do_stop(self, mode: str, block_callbacks: bool, checkpoint: bool,
...
-            postmaster.wait(timeout=stop_timeout)
+            # With safe_kill, once cleanly shut down there is nothing left to wait for,
+            # so terminate any lingering processes (e.g. replication connections) right away.
+            postmaster.wait(timeout=0 if safe_kill and clean_shutdown else stop_timeout)
         except TimeoutExpired:
+            # With safe_kill, never terminate before a clean shutdown: recovering from an
+            # unclean shutdown may take much longer than waiting for shutdown would have.
+            if safe_kill and not clean_shutdown:
+                logger.warning("Timeout during postmaster stop, but Postgres is not 'shut down'. Skipping kill")
+                return False, True
```
CyberDem0n, line 957: TBH, I am not sure if it is a good idea to kill replication connections even in synchronous_mode.
The problem is that after restart all transactions become visible even if they weren't replicated to enough standby nodes.
Waiting for walsender processes graceful shut down is the only way to make sure that transactions will not become visible until they are replicated to sufficient number of nodes.

PR #3683, opened 2026-07-21
Add site configuration and make cloning site-aware

hughcapet:
Part 1 of site-awareness work in Patroni

Now site can be configured via local configuration (incl.  environment variables) and make cloning site aware (prefer local reinit and replica bootstrap)
```
+ Cluster: demo (7666092842183090209) ---+-----------+----+-------------+-----+------------+-----+
| Site | Member   | Host       | Role    | State     | TL | Receive LSN | Lag | Replay LSN | Lag |
+------+----------+------------+---------+-----------+----+-------------+-----+------------+-----+
| dc1  | patroni1 | 172.19.0.3 | Leader  | running   |  1 |             |     |            |     |
| dc1  | patroni2 | 172.19.0.7 | Replica | streaming |  1 |   0/604FB68 |   0 |  0/604FB68 |   0 |
| dc2  | patroni3 | 172.19.0.2 | Replica | streaming |  1 |   0/604FB68 |   0 |  0/604FB68 |   0 |
| dc2  | patroni4 | 172.19.0.6 | Replica | streaming |  1 |   0/604FB68 |   0 |  0/604FB68 |   0 |
+------+----------+------------+---------+-----------+----+-------------+-----+------------+-----+
```

### DIFF
--- a/features/site_awareness.feature
+++ b/features/site_awareness.feature
@@ -1,0 +1,18 @@
+Feature: site awareness
+  Scenario: setup a mutisite cluster
+    When I start postgres-0 in site dc1
+    And postgres-0 is a leader after 10 seconds
+    And there is a non empty initialize key in DCS after 15 seconds
+    And "members/postgres-0" key in DCS has site=dc1 after 5 seconds
+    And I start postgres-1 in site dc1 with failover priority 2, sync priority 2
+    And I start postgres-2 in site dc2 with failover priority 3, sync priority 3
+    And I start postgres-3 in site dc2 with failover priority 4, sync priority 4
+    Then "members/postgres-1" key in DCS has site=dc1 after 5 seconds
+    And "members/postgres-2" key in DCS has site=dc2 after 5 seconds
+    And "members/postgres-3" key in DCS has site=dc2 after 5 seconds
+    And "status" key in DCS has dc1 in current_site
+
+  Scenario: test site-aware reinit
+    When I run patronictl.py reinit batman postgres-2 --force
+    Then I receive a response returncode 0
+    And there is one of ["bootstrapped from replica 'postgres-3'"] INFO in the postgres-2 patroni log after 25 seconds
--- a/features/site_awareness.feature
+++ b/features/site_awareness.feature
@@ -10,6 +10,9 @@ Feature: site awareness
     Then "members/postgres-1" key in DCS has site=dc1 after 5 seconds
     And "members/postgres-2" key in DCS has site=dc2 after 5 seconds
     And "members/postgres-3" key in DCS has site=dc2 after 5 seconds
+    And "members/postgres-1" key in DCS has state=running after 15 seconds
+    And "members/postgres-2" key in DCS has state=running after 15 seconds
+    And "members/postgres-3" key in DCS has state=running after 15 seconds
     And "status" key in DCS has dc1 in current_site
 
   Scenario: test site-aware reinit
--- a/features/site_awareness.feature
+++ b/features/site_awareness.feature
@@ -1,5 +1,5 @@
 Feature: site awareness
-  Scenario: setup a mutisite cluster
+  Scenario: setup a multisite cluster
     When I start postgres-0 in site dc1
     And postgres-0 is a leader after 10 seconds
     And there is a non empty initialize key in DCS after 15 seconds
--- a/patroni/config.py
+++ b/patroni/config.py
@@ -500,7 +500,7 @@ class Config(object):
             """
             return os.environ.pop(PATRONI_ENV_PREFIX + name.upper(), None)
 
-        for param in ('name', 'namespace', 'scope', 'thread_pool_size', 'thread_stack_size'):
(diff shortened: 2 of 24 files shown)


patroni/dcs/__init__.py
```diff
@@ -130,7 +130,7 @@ def get_dcs(config: Union['Config', Dict[str, Any]]) -> 'AbstractDCS':
     for name, dcs_class in iter_dcs_classes(config):
         # Propagate some parameters from top level of config if defined to the DCS specific config section.
         config[name].update({
-            p: config[p] for p in ('namespace', 'name', 'scope', 'loop_wait',
+            p: config[p] for p in ('namespace', 'name', 'scope', 'site', 'loop_wait',
```
CyberDem0n, line 134: I think we don't have to do it.
DCS doesn't really care about site.

patroni/dcs/__init__.py
```diff
@@ -1948,7 +1966,8 @@ def update_leader(self,
         ret = self._update_leader(cluster.leader)
         if ret and last_lsn:
             status: Dict[str, Any] = {self._OPTIME: last_lsn, 'slots': slots or None,
-                                      'retain_slots': self._build_retain_slots(cluster, slots)}
+                                      'retain_slots': self._build_retain_slots(cluster, slots),
+                                      'current_site': site}
```
CyberDem0n: If `site` is None, it will write {"current_site":null} to the /status key.
However, in case of K8s it will work differently, and remove "current_site" annotation, because we do PATCH.

patroni/dcs/__init__.py
CyberDem0n: From another side, if we propagate it, and store in AbstractDCS._site, it will allow to get rid of new argument in update_leader().

patroni/config.py
```diff
@@ -776,7 +776,8 @@ def _build_effective_configuration(self, dynamic_configuration: Dict[str, Any],
             'name',
             'scope',
             'retry_timeout',
-            'citus'
+            'citus',
+            'site'
```
CyberDem0n, line 780: It makes `site` available in `Postgresql` class.
If we combine this fact together with https://github.com/patroni/patroni/pull/3683/changes#diff-247edd60473c0933e528a1b4621cc612da54b6dc964720719d7623135d312474R133, it will make more and more sense to have a dedicated `PatroniSite`, class which will be used as a parent class for Patroni, AbstractDCS, Postgresql, and maybe Member.

It will also help to get rid of `SyncHandler._site` property, because it will be able to use self._postgresql.site.

patroni/validator.py
```diff
@@ -1066,6 +1066,7 @@ def __call__(self, value: Any) -> bool:
 schema = Schema({
     "name": validate_name,
     "scope": str,
+    Optional("site"): str,
```
CyberDem0n: IMO, we should require it to be non-empty.

patroni/ctl.py
```diff
@@ -1672,7 +1683,8 @@ def format_diff(param: str, values: Dict[str, str], hide_long: bool):
         title = 'Cluster'
         title_details = f' ({initialize})'
 
-    title = f' {title}: {name}{title_details} '
+    site = len(cluster_sites) == 1 and list(cluster_sites)[0] and ' Site: ' + list(cluster_sites)[0] or ''
+    title = f' {title}: {name}{title_details}{site} '
```
CyberDem0n: Lets display `Site: xyz` in front:
```suggestion
    site = next(iter(cluster_sites)) if len(cluster_sites) == 1 else ''
    site = f'Site: {site}, ' if site else ''
    title = f' {site}{title}: {name}{title_details} '
```

features/site_awareness.feature
```diff
@@ -0,0 +1,21 @@
+Feature: site awareness
+  Scenario: setup a mutisite cluster
```
CyberDem0n:
```suggestion
  Scenario: setup a multisite cluster
```

patroni/site.py
```diff
@@ -0,0 +1,32 @@
...
+        """
+        return self._site
+
+    def reload_site(self, site: Optional[str]) -> None:
+        """Load and set relevant site value from configuration.
+
+        :param site: string representing name of the site assigned to the current node.
+            loaded from the local configuration.
+        """
+        self._site = site
```
CyberDem0n: Hmm. I think `site` shouldn't be changed without restart.

patroni/dcs/__init__.py
```diff
@@ -1940,6 +1959,7 @@ def update_leader(self,
         :param last_lsn: absolute WAL LSN in bytes.
         :param slots: dictionary with permanent slots ``confirmed_flush_lsn``.
         :param failsafe: if defined dictionary passed to :meth:`~AbstractDCS.write_failsafe`.
+        :param site: the site to which the leader belongs.
```
CyberDem0n:
```suggestion
```

CyberDem0n (approved):

Issue #3462, opened 2025-10-07
patroni demoting master while having issues with k8s api (failsafe mode is on)


ovnozdrach:
### What happened?

Hi there!

Patroni demoted the master while there were issues with the Kubernetes API
failsafe mode was enabled in the Patroni configuration and it was activated when the primary node failed to contact the DCS. 
After about two minutes of operation in failsafe mode, Patroni demoted the master due to "Could not acquire TTL lock." 
Apparently all three etcd nodes were experiencing resource constraints at the time Patroni failed to update master lock. 
It seems the Kubernetes API responded intermittently, but for several minutes it was barely functional. Only one etcd node remained online (the other two crashed due to OOM), but it was throttled. 

```	
{"@timestamp": "2025-09-25 00:56:21,745", "level": "ERROR", "message": "Request to server https://[fd00::1]:443 failed: ReadTimeoutError(\"HTTPSConnectionPool(host='fd00::1', port=443): Read timed out. (read timeout=3.0106948520988226)\")", "module": "kubernetes", "app": "patroni"}
{"@timestamp": "2025-09-25 00:56:23,154", "level": "ERROR", "message": "Request to server https://[fd00::1]:443 failed: ReadTimeoutError(\"HTTPSConnectionPool(host='fd00::1', port=443): Read timed out. (read timeout=1.1928399801254272)\")", "module": "kubernetes", "app": "patroni"}
{"@timestamp": "2025-09-25 00:56:24,808", "level": "WARNING", "message": "Retrying (Retry(total=0, connect=None, read=None, redirect=0, status=None)) after connection broken by 'ConnectTimeoutError(<urllib3.connection.HTTPSConnection object at 0x7fa4e871b280>, 'Connection to fd00::1 timed out. (connect timeout=1.0)')': /api/v1/namespaces/stg-observability/endpoints/dragon-grafana-db", "module": "connectionpool", "app": "patroni"}
[...]

Issue #3344, opened 2025-04-28
Demoting self when failsafe mode is enabled and got ReadTimeoutError from k8s api.


shmelevall:
### What happened?

failsafe_mode don't work when got ReadTimeoutError from k8s api.

### How can we reproduce it (as minimally and precisely as possible)?

 maybe k8s is overload

### What did you expect to happen?

failsafe_mode could work 

### Patroni/PostgreSQL/DCS version

-   Patroni version: 4.0.4
-   PostgreSQL version: 
-   DCS (and its version): k8s

### Patroni configuration file

```yaml
bootstrap:
  dcs:
    failsafe_mode: true
    loop_wait: 10
    maximum_lag_on_failover: 33554432
    postgresql:
      parameters:
        archive_mode: 'on'
        archive_timeout: 1800s
        auto_explain.log_analyze: 'true'
        auto_explain.log_buffers: 'true'
        auto_explain.log_min_duration: 1000ms
        auto_explain.log_nested_statements: 'true'
        auto_explain.log_timing: 'true'
        auto_explain.log_triggers: 'true'
        auto_explain.log_wal: 'true'
        autovacuum_analyze_scale_factor: '0.01'
        autovacuum_max_workers: '8'
        autovacuum_multixact_freeze_max_age: '200000000'
        autovacuum_vacuum_cost_limit: '10000'
        autovacuum_vacuum_insert_scale_factor: '0.01'
        autovacuum_vacuum_scale_factor: '0.01'
        checkpoint_completion_target: '0.9'
        checkpoint_timeout: 20min
        effective_cache_size: 96GB
        effective_io_concurrency: '200'
        fsync: 'on'
        hot_standby: 'on'
        jit: 'on'
        log_autovacuum_min_duration: 0
        log_checkpoints: 'on'
        log_connections: 'on'
        log_disconnections: 'on'
        log_line_prefix: '%t [%p]: [%l-1] %c %x %d %u %a %h '
        log_lock_waits: 'on'
        log_min_duration_statement: 1000ms
        log_statement: ddl
        log_temp_files: 0
        maintenance_work_mem: 2GB
        max_connections: 1000
        max_parallel_maintenance_workers: '8'
        max_parallel_workers: '16'
        max_parallel_workers_per_gather: '8'
        max_prepared_transactions: '200'
        max_replication_slots: 10
        max_wal_senders: 10
        max_wal_size: 50GB
        max_worker_processes: '16'
        min_wal_size: 2GB
[...]

PR #3687, opened 2026-07-31
fix: check failsafe mode when update_lock() returns False (K8s DCS)

alexryndin:
## Problem

When using Kubernetes DCS, `failsafe_mode` is bypassed when `update_lock()` returns `False` instead of raising `DCSError`. This causes unnecessary leader demotion during K8s API timeouts (e.g., etcd defragmentation, API server overload).

### Root cause

`@catch_kubernetes_errors` on `_patch_or_create()` catches `KubernetesError` (which inherits from `DCSError`) and converts it to a `False` return:

```python
# kubernetes.py
@catch_kubernetes_errors                              # ← catches KubernetesError (= DCSError)
def _patch_or_create(self, name, annotations, ...):   # ← returns False instead of raising DCSError
    ...

def catch_kubernetes_errors(func):
    def wrapper(self, *args, **kwargs):
        try:
            return self._run_and_handle_exceptions(func, self, *args, **kwargs)
        except KubernetesError:   # ← KubernetesError inherits from DCSError (kubernetes.py:70)
            return False           # ← swallowed, update_lock() sees False, not DCSError
    return wrapper
```

As a result, `update_lock()` in `ha.py` returns `False` (not `DCSError`), and `process_healthy_cluster()` demotes without checking failsafe:

```python
# ha.py, process_healthy_cluster()
if self.update_lock(True):   # ← returns False (not DCSError) for K8s DCS
    ...                       # ← not reached
else:
    self.demote('immediate-nolock')   # ← demote WITHOUT failsafe check
```

The failsafe check in `_handle_dcs_error()` is only reachable via `except DCSError:` in `_run_cycle()` — which is never triggered because `@catch_kubernetes_errors` swallowed the exception.

### When does this happen?

Two conditions must be met simultaneously:
1. `load_cluster_from_dcs()` **succeeds** (ObjectCache has valid cached data → reads from cache)
2. `update_lock()` **fails** (K8s API PATCH times out → `@catch_kubernetes_errors` swallows error → returns `False`)

This is exactly what happens during etcd defragmentation: ObjectCache retains cached data (reads succeed), but etcd can't process writes (~5 seconds).

### When does failsafe work correctly?

When `load_cluster_from_dcs()` also fails (e.g., SSL errors, complete K8s API connection failure), `DCSError` IS raised (via `__load_cluster()` → `KubernetesError`), reaches `_run_cycle()`, and `_handle_dcs_error()` is called with failsafe check. This was confirmed in production logs.

### This is consistent with PR #2484

PR #2484 ("Raise DCSError when communication with DCS fails") explicitly stated the intent:

> *"Previously such an exception was raised only from the `get_cluster()` method, and now we will do the same from the `update_leader()` and `attempt_to_acquire_leader()` methods."*

However, for K8s DCS, `@catch_kubernetes_errors` on `_patch_or_create()` prevents `DCSError` from propagating out of `update_leader()`. The dead `except (RetryFailedError, K8sException)` block in `_update_leader_with_retry()` (added in the same PR) proves the intent was for exceptions to propagate.

### Related issues

- **#3344** — "Demoting self when failsafe mode is enabled and got ReadTimeoutError from k8s api" (Patroni 4.0.4)
- **#3462** — "patroni demoting master while having issues with k8s api (failsafe mode is on)" (Patroni 4.0.6)

In #3462, failsafe worked correctly for ~2 minutes (via read path), then the leader was demoted when `update_lock()` returned `False` (write path). The logs show `ReadTimeoutError` and `ConnectTimeoutError`, not HTTP 409.

- **#3404** — Similar fix for etcd DCS: *"We need to convert it to etcd.EtcdConnectionFailed for failsafe_mode to work."*
- **#3348** — Partial fix for `attempt_to_acquire_leader()` path (409 handling), but does not address `update_leader()` path.

## Fix

Add a failsafe check in `process_healthy_cluster()` before demoting, mirroring the logic in `_handle_dcs_error()`:

```python
else:
    logger.error('failed to update leader lock')
    if self.state_handler.is_primary():
        if self.is_paused():
            return 'continue to run as primary after failing to update leader lock in DCS'
        if self.is_failsafe_mode() and self.state_handler.is_running() \
                and self.check_failsafe_topology():
            self.set_is_leader(True)
            self._failsafe.set_is_active(time.time())
            self.watchdog.keepalive()
            self._sync_replication_slots(True)
            return 'continue to run as a leader because failsafe mode is enabled'\
                   ' and all members are accessible'
        self.demote('immediate-nolock')
        return 'demoted self because failed to update leader lock in DCS'
```

This is safe because `check_failsafe_topology()` contacts all replicas via REST. If another node has legitimately taken leadership, it responds with `Running as a leader` → `check_failsafe_topology()` returns `False` → demote proceeds correctly.

## Test

Added `test_update_lock_returns_false_failsafe` — verifies that when `update_leader` returns `False` (simulating K8s DCS behavior), failsafe is still checked and the leader continues running.

### DIFF
--- a/patroni/ha.py
+++ b/patroni/ha.py
@@ -1942,6 +1942,20 @@ class Ha(object):
                 if self.state_handler.is_primary():
                     if self.is_paused():
                         return 'continue to run as primary after failing to update leader lock in DCS'
+                    # When update_lock() returns False (rather than raising DCSError),
+                    # _handle_dcs_error() is never reached and failsafe_mode is bypassed.
+                    # This happens with Kubernetes DCS, where @catch_kubernetes_errors
+                    # on _patch_or_create() swallows KubernetesError (= DCSError) and
+                    # converts it to a False return before it can propagate to _run_cycle().
+                    # Check failsafe here to cover that gap.
+                    if self.is_failsafe_mode() and self.state_handler.is_running() \
+                            and self.check_failsafe_topology():
+                        self.set_is_leader(True)
+                        self._failsafe.set_is_active(time.time())
+                        self.watchdog.keepalive()
+                        self._sync_replication_slots(True)
+                        return 'continue to run as a leader because failsafe mode is enabled'\
+                               ' and all members are accessible'
                     self.demote('immediate-nolock')
                     return 'demoted self because failed to update leader lock in DCS'
                 else:
--- a/patroni/ha.py
+++ b/patroni/ha.py
@@ -1943,21 +1943,13 @@ class Ha(object):
                     if self.is_paused():
                         return 'continue to run as primary after failing to update leader lock in DCS'
                     # When update_lock() returns False (rather than raising DCSError),
-                    # _handle_dcs_error() is never reached and failsafe_mode is bypassed.
+                    # _handle_dcs_error() is never reached via the except DCSError path
+                    # in _run_cycle(), and failsafe_mode is bypassed.
                     # This happens with Kubernetes DCS, where @catch_kubernetes_errors
                     # on _patch_or_create() swallows KubernetesError (= DCSError) and
                     # converts it to a False return before it can propagate to _run_cycle().
-                    # Check failsafe here to cover that gap.
-                    if self.is_failsafe_mode() and self.state_handler.is_running() \
-                            and self.check_failsafe_topology():
-                        self.set_is_leader(True)
-                        self._failsafe.set_is_active(time.time())
-                        self.watchdog.keepalive()
-                        self._sync_replication_slots(True)
-                        return 'continue to run as a leader because failsafe mode is enabled'\
-                               ' and all members are accessible'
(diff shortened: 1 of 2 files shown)


CyberDem0n:
The failsafe_mode contract is that it is triggered when communication with DCS fails because DCS is not accessible.
#2484 made sure that all methods raise DCSError exception in this case, instead of returning something.

I don't see any prove (logs) that DCS wasn't accessible in your case.

> @catch_kubernetes_errors on _patch_or_create() catches KubernetesError (which inherits from DCSError) and converts it to a False return:

There is no decorator on _patch_or_create() method: https://github.com/patroni/patroni/blob/9b1533c0a60f9fcf0f0d1fc7b34ac2adec7fb8dc/patroni/dcs/kubernetes.py#L1083-L1087

PR #3688, opened 2026-08-04
Implement site-aware failover and manual switchover

hughcapet:
- site-aware automatic failover
  - prefer local site members
  - if no local members, perform cross-site failover
  - if local members are lagging (even if lag is under `maximum_lag_on_failover`), prefer remote replicas with lower lag
  - failover priority is validated within the local site first, remote member's priority is only considered when no local members left
- manual switchover to a site without specific candidate
  - in case of a switchover to a remote site, when there are no eligible local candidates in the specified site after primary is already demoted, prefer other remote sites over previous leader site's members to failover to
- manual failover still can only be performed to a specific candidate but there is an additional `--site` option for `patronictl` to filter candidates selection list

### DIFF
--- a/patroni/api.py
+++ b/patroni/api.py
@@ -1081,7 +1081,7 @@ class RestApiHandler(BaseHTTPRequestHandler):
         """
         failover = self.server.patroni.dcs.get_cluster().failover
         if failover and failover.scheduled_at:
-            if not self.server.patroni.dcs.manual_failover('', '', version=failover.version):
+            if not self.server.patroni.dcs.manual_failover('', '', '', version=failover.version):
                 return self.send_error(409)
             else:
                 data = "scheduled switchover deleted"
@@ -1157,12 +1157,13 @@ class RestApiHandler(BaseHTTPRequestHandler):
             action.title(), timeout * 2)
 
     def is_failover_possible(self, cluster: Cluster, leader: Optional[str], candidate: Optional[str],
-                             action: str) -> Optional[str]:
+                             site: Optional[str], action: str) -> Optional[str]:
         """Checks whether there are nodes that could take over after demoting the primary.
 
         :param cluster: the Patroni cluster.
         :param leader: name of the current Patroni leader.
         :param candidate: name of the Patroni node to be promoted.
+        :param site: name of the site to failover/switchover to.
         :param action: the action to be performed (``switchover`` or ``failover``).
 
         :returns: a string with the error message or ``None`` if good nodes are found.
@@ -1177,14 +1178,21 @@ class RestApiHandler(BaseHTTPRequestHandler):
             members = [m for m in cluster.members if m.name == candidate]
             if not members:
                 return 'candidate does not exists'
-        elif config.is_synchronous_mode and not config.is_quorum_commit_mode:
-            members = [m for m in cluster.members if cluster.sync.matches(m.name)]
-            if not members:
-                return action + ' is not possible: can not find sync_standby'
         else:
-            members = [m for m in cluster.members if not cluster.leader or m.name != cluster.leader.name and m.api_url]
-            if not members:
-                return action + ' is not possible: cluster does not have members except leader'
+            members = cluster.members
+            if site:
+                members = [m for m in cluster.members if m.site == site]
+                if not members:
+                    return action + ' is not possible: can not find members in site ' + site
+            if config.is_synchronous_mode and not config.is_quorum_commit_mode:
+                members = [m for m in members if cluster.sync.matches(m.name)]
+                if not members:
+                    return action + ' is not possible: can not find sync_standby'
+            else:
+                members = [m for m in members if not cluster.leader or m.name != cluster.leader.name and m.api_url]
+                if not members:
+                    return action + ' is not possible: cluster does not have members except leader'
+
         for st in self.server.patroni.ha.fetch_nodes_statuses(members):
             if st.failover_limitation() is None:
                 return None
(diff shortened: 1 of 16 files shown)


patroni/api.py
```diff
@@ -1251,16 +1261,16 @@ def do_POST_failover(self, action: str = 'failover') -> None:
             logger.warning('received failover request with leader specified - performing switchover instead')
             action = 'switchover'
 
-        if not data and leader == candidate:
+        if not data and leader and candidate and leader == candidate:
```
CyberDem0n: Why was it changed?

patroni/ha.py
```diff
@@ -1871,12 +1894,13 @@ def process_unhealthy_cluster(self) -> str:
             if self.acquire_lock():
                 failover = self.cluster.failover
                 if failover:
-                    if self.is_paused() and failover.leader and failover.candidate:
+                    if self.is_paused() and failover.leader and (failover.candidate or failover.site):
```
CyberDem0n: In pause we support only switchover to a specified node, not site.

patroni/api.py
hughcapet: leftover from the time when it was possible to perform failover to a site

patroni/ha.py
```diff
@@ -1437,40 +1439,63 @@ def _is_healthiest_node(self, members: Collection[Member],
...
 
         # When not in quorum commit we just want to return `True`.
         # In quorum commit the former leader is special and counted healthy even when there are no other nodes.
         # Otherwise check that the number of votes exceeds the quorum field from the /sync key.
         return not self.quorum_commit_mode_is_active() or quorum_votes >= quorum\
             or nodes_ahead == 0 and self.cluster.sync.leader == self.state_handler.name
 
-    def is_failover_possible(self, *, cluster_lsn: int = 0, exclude_failover_candidate: bool = False) -> bool:
+    def is_failover_possible(self, *, cluster_lsn: int = 0,
+                             exclude_failover_candidate: bool = False, local_only: bool = False) -> bool:
```
CyberDem0n: I am not sure is local_only is a good name for it.

patroni/ctl.py
```diff
@@ -1366,20 +1376,28 @@ def _do_failover_or_switchover(action: str, cluster_name: str, group: Optional[i
         failover_value['leader'] = switchover_leader
     if scheduled_at_str:
         failover_value['scheduled_at'] = scheduled_at_str
+    if site and action != 'failover':
+        failover_value['site'] = site
 
     logging.debug(failover_value)
 
-    # By now we have established that the leader exists and the candidate exists
```
CyberDem0n, line 1372: why was it removed?

patroni/ha.py
```diff
@@ -1437,40 +1439,63 @@ def _is_healthiest_node(self, members: Collection[Member],
...
+                    if my_wal_position == st.wal_position:
+                        eligible_members.append(st)
+
+        action = self._get_failover_action_name()
+        if self.cluster.failover and self.cluster.failover.site:
+            eligible_members = [st for st in eligible_members if st.data.get('site') == self.cluster.failover.site]
+            if eligible_members and self.patroni.site != self.cluster.failover.site:
+                logger.info('%s to the requested site %s is possible, while my site is %s',
+                            action.capitalize(), self.cluster.failover.site, self.patroni.site)
+                return False
```
CyberDem0n, line 1454: I think there is a problem.
You overwrite `eligible_members`, and the list might be empty after applying failover.site filter, while we have to do something similar to the next block: https://github.com/patroni/patroni/pull/3688/changes#diff-b6a3a0ef64783a6d32feb908d74a3a009e9f25008b8ca5313c9c9747e3b74a63R1456-R1464

patroni/ha.py
```diff
@@ -1846,14 +1877,14 @@ def process_manual_failover_from_leader(self) -> Optional[str]:
...
-        self.dcs.manual_failover('', '', version=failover.version)
+        self.dcs.manual_failover('', '', '', version=failover.version)
 
     def process_unhealthy_cluster(self) -> str:
         """Cluster has no leader key"""
         # First, we want to handle primary_race_backoff. Do it only for non-standby cluster,
         # not in maintenance mode and when there is no manual failover/switchover in progress.
         if not self.is_paused() and not self.is_standby_cluster() and \
-                not (self.cluster.failover and self.cluster.failover.candidate) and \
+                not (self.cluster.failover and (self.cluster.failover.candidate or self.cluster.failover.site)) and \
```
CyberDem0n: I am not sure we should add self.cluster.failover.site here.

patroni/ha.py
```diff
@@ -2611,4 +2645,6 @@ def is_eligible(node: Member) -> bool:
                 (exclude_failover_candidate or not failover
                  or not failover.candidate or node.name == failover.candidate)
 
-        return list(filter(is_eligible, self.cluster.members))
+        members = [m for m in self.cluster.members if not failover or not failover.site
```
CyberDem0n: IMO, this condition should be put into is_eligible() function.

patroni/ctl.py
hughcapet: with the site option added, it is not true anymore

patroni/dcs/__init__.py
```diff
@@ -2040,6 +2043,9 @@ def manual_failover(self, leader: Optional[str], candidate: Optional[str],
 
         if scheduled_at:
             failover_value['scheduled_at'] = scheduled_at.isoformat()
+
+        if not candidate and site:
+            failover_value['site'] = site
```
CyberDem0n: Lets move it up:
```python
        if candidate:
            failover_value['member'] = candidate
        elif site:
            failover_value['site'] = site
```

CyberDem0n (approved):

PR #3695, opened 2026-08-16
Reload REST API certificate when local config also changed

evkuzin:
```
if local or sighup and self.api.reload_local_certificate():
```

This parses as `local or (sighup and ...)`, so when the local config changed, reload_local_certificate() is never called. 

This is particularly painful when you have IaaC: you pushed changes to patroni config and at the same time cert expiration triggered a cert renewal, which will never be loaded.

### DIFF
--- a/patroni/__main__.py
+++ b/patroni/__main__.py
@@ -184,7 +184,8 @@ class Patroni(AbstractPatroniDaemon, ClusterSite, Tags):
             if local:
                 self._tags = self._get_tags()
                 self.request.reload_config(self.config)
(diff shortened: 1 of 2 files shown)


evkuzin:
Hm. I'm looking at the logs and it looks like a race to me, but I dont see obvious explanation how my change can lead to it. Is it possible its a flaky test?

evkuzin:
@CyberDem0n is it possible to rerun this failed one? I cant find any other explanation other than flaky test 🫠 Thx!

hughcapet (approved):

PR #3696, opened 2026-08-16
Expose cert expiry date for patroni api in metrics

evkuzin:
in addition to #3695 a proposal to have an expiry date for the patroni API cert exposed in metrics so it can be monitored.

### DIFF
--- a/patroni/api.py
+++ b/patroni/api.py
@@ -774,11 +774,11 @@ class RestApiHandler(BaseHTTPRequestHandler):
         metrics.append("# TYPE patroni_failover_priority gauge")
         metrics.append("patroni_failover_priority{0} {1}".format(labels, patroni.failover_priority))
 
-        if self.server._ssl_not_after is not None:
+        if self.server.ssl_not_after is not None:
             metrics.append("# HELP patroni_restapi_certificate_expiry Epoch timestamp when the REST API"
                            " certificate expires.")
             metrics.append("# TYPE patroni_restapi_certificate_expiry gauge")
-            metrics.append("patroni_restapi_certificate_expiry{0} {1}".format(labels, self.server._ssl_not_after))
+            metrics.append("patroni_restapi_certificate_expiry{0} {1}".format(labels, self.server.ssl_not_after))
 
         self.write_response(200, '\n'.join(metrics) + '\n', content_type='text/plain')
 
@@ -1540,10 +1540,19 @@ class RestApiServer(ThreadingMixIn, HTTPServer):
         self._executor = PatroniThreadPoolExecutor(max_workers=thread_pool_size + 1, thread_name_prefix='RestAPI')
         self.__ssl_options: Dict[str, Any] = {}
         self.__ssl_serial_number = None
-        self._ssl_not_after: Optional[float] = None
+        self.__ssl_not_after: Optional[int] = None
         self._received_new_cert = False
         self.reload_config(config)
         self.daemon = True
+
+    @property
+    def ssl_not_after(self) -> Optional[int]:
+        """Epoch timestamp when the certificate used by the REST API expires.
+
+        It is ``None`` when the REST API is not running in HTTPS mode, or when the expiry could not be read from the
+        certificate.
+        """
+        return self.__ssl_not_after
 
     def construct_server_tokens(self, token_config: str) -> str:
         """Construct the value for the ``Server`` HTTP header based on *server_tokens*.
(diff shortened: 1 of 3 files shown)


patroni/api.py
```diff
@@ -1901,19 +1913,26 @@ def shutdown_request(self, request: Union[socket.socket, Tuple[bytes, socket.soc
...
             except ssl.SSLError as e:
-                logger.error('Failed to get serial number from certificate %s: %r', self.__ssl_options['certfile'], e)
+                logger.error('Failed to decode certificate %s: %r', certfile, e)
+            else:
+                try:
+                    return crt.get('serialNumber'), float(ssl.cert_time_to_seconds(crt.get('notAfter')))
+                except (TypeError, ValueError) as e:
+                    logger.error('Failed to get expiry from certificate %s: %r', certfile, e)
+                    return crt.get('serialNumber'), None
+        return None, None
```
```suggestion
                return None, None
            sn = crt.get('serialNumber')
            try:
                exp = float(ssl.cert_time_to_seconds(crt.get('notAfter')))
            except (TypeError, ValueError) as e:
                logger.error('Failed to get expiry from certificate %s: %r', certfile, e)
                exp = None
            return sn, exp
```

Please also resolve pyright complain:
```python
Argument of type "Any | None" cannot be assigned to parameter "cert_time" of type "str" in function "cert_time_to_seconds"
  Type "Any | None" is not assignable to type "str"
    "None" is not assignable to "str" (reportArgumentType)
```

patroni/api.py
```diff
@@ -1901,19 +1913,26 @@ def shutdown_request(self, request: Union[socket.socket, Tuple[bytes, socket.soc
...
         super(RestApiServer, self).shutdown_request(request)
 
-    def get_certificate_serial_number(self) -> Optional[str]:
-        """Get serial number of the certificate used by the REST API.
+    def parse_certificate(self) -> Tuple[Optional[str], Optional[float]]:
+        """Parse the certificate used by the REST API.
 
-        :returns: serial number of the certificate configured through ``restapi.certfile`` setting.
+        :returns: serial number of the certificate configured through ``restapi.certfile`` setting and the
+            epoch timestamp when it expires.
```
CyberDem0n: It would be nice to explicitly mention that it returns a tuple and what are the values

patroni/api.py
```diff
@@ -1530,6 +1540,7 @@ def __init__(self, patroni: Patroni, config: Dict[str, Any]) -> None:
         self._executor = PatroniThreadPoolExecutor(max_workers=thread_pool_size + 1, thread_name_prefix='RestAPI')
         self.__ssl_options: Dict[str, Any] = {}
         self.__ssl_serial_number = None
+        self._ssl_not_after: Optional[float] = None
```
CyberDem0n: It deserves an explicit property method to resolve pyright complains when being accessed from RestApiHandler.

patroni/api.py
```diff
@@ -1922,7 +1941,7 @@ def reload_local_certificate(self) -> Optional[bool]:
             otherwise.
         """
         if self.__protocol == 'https':
-            on_disk_cert_serial_number = self.get_certificate_serial_number()
+            on_disk_cert_serial_number = self.parse_certificate()[0]
             if on_disk_cert_serial_number != self.__ssl_serial_number:
                 self._received_new_cert = True
                 self.__ssl_serial_number = on_disk_cert_serial_number
```
CyberDem0n: Don't we also need to reassign _ssl_not_after in this case?

patroni/api.py
```diff
@@ -1901,19 +1922,37 @@ def shutdown_request(self, request: Union[socket.socket, Tuple[bytes, socket.soc
...
+
+        import ssl
+        try:
+            crt = cast(Dict[str, Any], ssl._ssl._test_decode_cert(certfile))  # pyright: ignore
+        except ssl.SSLError as e:
+            logger.error('Failed to decode certificate %s: %r', certfile, e)
+            return None, None
+
+        serial_number: Optional[str] = crt.get('serialNumber')
+        not_after = crt.get('notAfter')
```
CyberDem0n, line 1946: it reports that this line isn't covered with tests.
something is fishy...

evkuzin: I think I fixed that. https://app.codecov.io/gh/patroni/patroni/pull/3696 is 🟢

evkuzin: yup. tests passed now. 🎉

evkuzin:
@CyberDem0n wdyt?

hughcapet (approved):

Issue #3603, opened 2026-05-06
Patroni is not checking or reporting on pg_is_in_recovery() for current leader.


guriandoro:
### What happened?

After a failover, the node that was flagged as leader by Patroni was in read-only mode. Patroni error log and patronictl list outputs showed as if the cluster was healthy, when in reality no writes were possible.

### How can we reproduce it (as minimally and precisely as possible)?

- Start with a 2 node patroni cluster and a pgbackrest to nfs
- Configure pgbackrest and set archive_command and restore_command as needed.
- using iptables on the replica, block returning traffic from the nfs
```
iptables -I INPUT 1 -p tcp -s "$SERVER" --sport 2049 -j DROP
``` 
- kill -9 pidof_patroni on the leader
- the replica will be correctly elected as leader from patroni side, but from postgres side the following will never complete:
```
archive-get command begin 2.58.0: [00000002.history, pg_wal/RECOVERYHISTORY] ...
```

Patroni shows a seemingly healthy node, wthen in reality the node is in read-only mode.

```
[root@node1 /]# patronictl list
+ Cluster: cluster1 (7636233032694706021) --------+----+-------------+-----+------------+-----+
|      Member     |    Host    |  Role  |  State  | TL | Receive LSN | Lag | Replay LSN | Lag |
+-----------------+------------+--------+---------+----+-------------+-----+------------+-----+
| cluster113949-1 | 172.16.2.3 | Leader | running |  1 |             |     |            |     |
+-----------------+------------+--------+---------+----+-------------+-----+------------+-----+

[root@node1 journalctl -u patroni | tail
May 06 02:51:53 node1 patroni[4000]: 2026-05-06 02:51:53,951 INFO: no action. I am (cluster113949-1), the leader with the lock
May 06 02:52:03 node1 patroni[4000]: 2026-05-06 02:52:03,951 INFO: no action. I am (cluster113949-1), the leader with the lock
May 06 02:52:13 node1 patroni[4000]: 2026-05-06 02:52:13,952 INFO: no action. I am (cluster113949-1), the leader with the lock
May 06 02:52:23 node1 patroni[4000]: 2026-05-06 02:52:23,951 INFO: no action. I am (cluster113949-1), the leader with the lock
May 06 02:52:33 node1 patroni[4000]: 2026-05-06 02:52:33,951 INFO: no action. I am (cluster113949-1), the leader with the lock
[...]

PR #3697, opened 2026-08-18
Report 'promoted' role in patronictl list and /cluster endpoint

guriandoro:
## What problem does this solve?

Fixes the visibility half of #3603 (the log warning is sent as a separate PR).

After a failover, a node can win the leader lock and receive the promote
request, yet Postgres never finishes the promotion because
`restore_command` hangs (in my case: a stalled NFS mount behind
pgBackRest's `archive-get`). Postgres keeps running as a standby,
`pg_is_in_recovery()` stays `true`, and the cluster has no writable node.

Meanwhile `patronictl list` and the `/cluster` REST endpoint show a
healthy `Leader` / `role: leader`, because `cluster_as_json()` hardcodes
the role for whichever member holds the lock. The DCS itself tells the
truth: the member key keeps `"role": "promoted"` until promotion
completes. That signal was being dropped at the presentation layer.

## What does this PR do?

In `cluster_as_json()` (`patroni/utils.py`), when the lock holder's
member key still reports `role: promoted`, pass it through instead of
hardcoding `leader`:

- `patronictl list` shows `Promoted` in the Role column while promotion
  has not completed, reverting to `Leader` once it does.
- `GET /cluster` returns `"role": "promoted"` for that member.

Side effect (arguably a second fix): since `replication_state` is only
suppressed for role `leader`, the State column for a stuck member now
shows the actual state (e.g. `in archive recovery`) instead of
`running` — which is exactly what the operator needs to see to find the
hanging `restore_command`.

During a normal failover the `Promoted` role is visible for at most a
cycle or two, same as the `promoted` role in the DCS member key today.

Standby clusters are unaffected (`standby_leader` takes precedence, as
before). All other surfaces (`/patroni`, `/primary`, `/metrics`, the DCS
member key) already report this state correctly and are untouched.

## Repro / verification

Original repro from the issue: 2-node cluster with pgBackRest over NFS,
block NFS return traffic on the replica with iptables, `kill -9` Patroni
on the leader. The replica takes the lock, promote is issued,
`archive-get 00000002.history` hangs forever, `pg_is_in_recovery()`
stays true.

Verified live against this branch (PG 18.4, etcd v3.5, 2-node local
cluster): the hanging `restore_command` was simulated with a script that
sleeps forever on timeline-history requests, then Patroni was `kill -9`'d
on the leader. While node2 was stuck (`pg_is_in_recovery()` = `t`):

```
$ patronictl list
+ Cluster: repro3603 (7675475385382874329) ----------------+----+-------------+-----+------------+-----+
| Member | Host           | Role     | State               | TL | Receive LSN | Lag | Replay LSN | Lag |
+--------+----------------+----------+---------------------+----+-------------+-----+------------+-----+
| node2  | 127.0.0.1:5434 | Promoted | in archive recovery |  1 |             |     |            |     |
+--------+----------------+----------+---------------------+----+-------------+-----+------------+-----+

$ curl -s http://127.0.0.1:8009/cluster
{"members": [{"name": "node2", "role": "promoted", "state": "in archive recovery", ...}], "scope": "repro3603"}
```

Once the restore_command hang was released, promotion completed and both
surfaces reverted on the next loop:

```
$ patronictl list
| node2  | 127.0.0.1:5434 | Leader | running |  2 | ...

$ curl -s http://127.0.0.1:8009/cluster
{"members": [{"name": "node2", "role": "leader", "state": "running", ...}], "scope": "repro3603"}
```

## Tests

Extended `test_output_members` in `tests/test_ctl.py`: a cluster whose
lock holder has `role: promoted` in its member data renders `Promoted`
in the Role column. Full suite passes (706 tests). Docs updated:
`docs/rest_api.rst` role enumeration, `cluster_as_json` and
`output_members` docstrings.

### DIFF
--- a/patroni/utils.py
+++ b/patroni/utils.py
@@ -922,7 +922,8 @@ def cluster_as_json(cluster: 'Cluster') -> Dict[str, Any]:
         * ``members``: list of members in the cluster. Each value is a :class:`dict` that may have the following keys:
 
             * ``name``: the name of the host (unique in the cluster). The ``members`` list is sorted by this key;
-            * ``role``: ``leader``, ``standby_leader``, ``sync_standby``, ``quorum_standby``, or ``replica``;
+            * ``role``: ``leader``, ``promoted`` (holds the leader lock but promotion has not completed yet),
+                ``standby_leader``, ``sync_standby``, ``quorum_standby``, or ``replica``;
             * ``state``: one of :class:`~patroni.postgresql.misc.PostgresqlState`;
             * ``api_url``: REST API URL based on ``restapi->connect_address`` configuration;
             * ``host``: PostgreSQL host based on ``postgresql->connect_address``;
@@ -948,7 +949,7 @@ def cluster_as_json(cluster: 'Cluster') -> Dict[str, Any]:
             * ``to``: name of the member to be promoted.
     """
     from . import global_config
-    from .postgresql.misc import format_lsn
+    from .postgresql.misc import format_lsn, PostgresqlRole
 
     config = global_config.from_cluster(cluster)
     leader_name = cluster.leader.name if cluster.leader else None
@@ -958,7 +959,13 @@ def cluster_as_json(cluster: 'Cluster') -> Dict[str, Any]:
     sync_role = 'quorum_standby' if config.is_quorum_commit_mode else 'sync_standby'
     for m in cluster.members:
         if m.name == leader_name:
-            role = 'standby_leader' if config.is_standby_cluster else 'leader'
+            if config.is_standby_cluster:
+                role = 'standby_leader'
+            elif m.data.get('role') == PostgresqlRole.PROMOTED:
+                # the member holds the leader lock, but Postgres promotion has not completed yet (still in recovery)
+                role = 'promoted'
+            else:
+                role = 'leader'
         elif config.is_synchronous_mode and cluster.sync.matches(m.name):
             role = sync_role
         else:
--- a/docs/rest_api.rst
+++ b/docs/rest_api.rst
@@ -479,6 +479,7 @@ Cluster status endpoints
       }
     }
 
+  The ``role`` of a member is one of ``leader``, ``promoted``, ``standby_leader``, ``sync_standby``, ``quorum_standby``, or ``replica``. The ``promoted`` role is reported for a member that already holds the leader lock while PostgreSQL promotion has not completed yet, i.e. PostgreSQL is still running in recovery and therefore is read-only.
 
 - The ``GET /history`` endpoint provides a view on the history of cluster switchovers/failovers. The format is very similar to the content of history files in the ``pg_wal`` directory. The only difference is the timestamp field showing when the new timeline was created.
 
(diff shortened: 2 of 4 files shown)


patroni/utils.py
```diff
@@ -958,7 +959,13 @@ def cluster_as_json(cluster: 'Cluster') -> Dict[str, Any]:
     sync_role = 'quorum_standby' if config.is_quorum_commit_mode else 'sync_standby'
     for m in cluster.members:
         if m.name == leader_name:
-            role = 'standby_leader' if config.is_standby_cluster else 'leader'
+            if config.is_standby_cluster:
+                role = 'standby_leader'
+            elif m.data.get('role') == PostgresqlRole.PROMOTED:
+                # the member holds the leader lock, but Postgres promotion has not completed yet (still in recovery)
+                role = 'promoted'
```
CyberDem0n, line 966: 1. Because role is used on line 974 now patronictl shows `Promoted | in archive recovery`. instead of `Promoted | running`
2. 'promoted` also breaks [logic](https://github.com/patroni/patroni/blob/5f2c94c82a9cbea388c40451bdc8444683bae367/patroni/ctl.py#L542-L549) in ctl.py:get_all_members()
3. it also affects [ctl.py:change_cluster_role()](https://github.com/patroni/patroni/blob/5f2c94c82a9cbea388c40451bdc8444683bae367/patroni/ctl.py#L2402-L2411)

guriandoro: >Because role is used on line 974 now patronictl shows Promoted | in archive recovery. instead of Promoted | running

That State change was intentional (mentioned in the PR description): with a hanging `restore_command` the member key keeps `state: running`, so `Promoted | running` would still hide where the problem is, while `in archive recovery` points the operator straight at it. But I see the flip side: every normal failover now transiently shows `Promoted | in archive recovery` for a cycle. If you prefer, I can suppress `replication_state` for `promoted` as well (i.e. `role not in ('leader', 'promoted')` on that line) so it shows `Promoted | running`. Which way do you want it?

>'promoted` also breaks [logic](https://github.com/patroni/patroni/blob/5f2c94c82a9cbea388c40451bdc8444683bae367/patroni/ctl.py#L542-L549) in ctl.py:get_all_members()

Agreed, fixed in a101cc76: a lock holder with role `promoted` no longer matches `--role primary` (it is still in recovery and read-only) nor `--role standby-leader`; it still matches `--role leader`, since it does hold the lock. Added a test case to `test_get_all_members`. Note this behavior predates this PR (`get_all_members()` reads the member key from the DCS directly, and the DCS already reports `promoted` there), but it belongs with this change.

>it also affects [ctl.py:change_cluster_role()](https://github.com/patroni/patroni/blob/5f2c94c82a9cbea388c40451bdc8444683bae367/patroni/ctl.py#L2402-L2411)

Same situation as 2: `change_cluster_role()`'s watch loop compares the DCS member key role, so it behaves identically with or without this PR. While the role is `promoted` it just keeps polling and prints `leader role: promoted` until promotion completes. Do you want it to handle `promoted` explicitly (e.g. report "promotion in progress" while waiting), and if so, in this PR or as a follow-up?

CyberDem0n: > That State change was intentional (mentioned in the PR description): with a hanging restore_command the member key keeps state: running, so Promoted | running would still hide where the problem.

`in archive recovery` just essentially duplicates `Promoted` and doesn't bring any additional value, the state should be `running`.

> Same situation as 2: change_cluster_role()'s watch loop compares the DCS member key role, so it behaves identically with or without this PR. While the role is promoted it just keeps polling and prints leader role: promoted until promotion completes.

That's exactly the problem. Old behavior - break the loop as soon as role changed to `primary` (when `pg_ctl promote` was called), and now it may wait much longer, impacting existing users.

PR #3699, opened 2026-08-19
Site-aware synchronous mode

hughcapet:
Make synchronous mode site-aware by introducing 6 strategies of synchronous replicas selection: any (old behavior, no site-awareness), balanced, prefer_local, local_only, prefer_remote, remote_only


patroni/dcs/__init__.py
```diff
@@ -639,7 +641,9 @@ def from_node(version: Optional[_Version], value: Union[str, Dict[str, Any], Non
                 raise PatroniAssertionError('not a dict')
             leader = value.get('leader')
             quorum = value.get('quorum')
-            return SyncState(version, leader, value.get('sync_standby'), int(quorum) if leader and quorum else 0)
+            cross_site_mode = SyncCrossSiteMode(value.get('cross_site_mode', 'any'))
+            return SyncState(version, leader, value.get('sync_standby'),
+                             int(quorum) if leader and quorum else 0, cross_site_mode)
         except (PatroniAssertionError, TypeError, ValueError):
             return SyncState.empty(version)
```
CyberDem0n: The `cross_site_mode` enum coercion runs *inside* the shared `try/except (PatroniAssertionError, TypeError, ValueError)` that also guards `leader`/`sync_standby`/`quorum` parsing. As a result, an **unrecognized** `cross_site_mode` value raises `ValueError`, is swallowed by the `except`, and returns `SyncState.empty(version)` — silently discarding an otherwise-valid `leader`,  sync_standby`, and `quorum`.

patroni/dcs/__init__.py
```diff
@@ -2139,19 +2144,21 @@ def delete_cluster(self) -> bool:
...
                        see at least :attr:`~SyncState.quorum` other nodes from the
                        :attr:`~SyncState.sync_standby` + :attr:`~SyncState.leader` list
+        :param cross_site_mode: synchronous_cross_site mode applied.
 
         :returns: dictionary that later could be serialized to JSON or saved directly to DCS.
         """
         return {'leader': leader, 'quorum': quorum,
-                'sync_standby': ','.join(sorted(sync_standby)) if sync_standby else None}
+                'sync_standby': ','.join(sorted(sync_standby)) if sync_standby else None,
+                'cross_site_mode': cross_site_mode.value if cross_site_mode else None}
```
CyberDem0n, line 2165: `write_sync_state` always serializes the *configured* mode (`global_config.sync_cross_site_mode`), but `current_state` **bypasses** site logic entirely when the primary has no site (`if not self._postgresql.site or cross_site_mode == SyncCrossSiteMode.ANY`). So a site-less primary configured with `local_only` selects candidates as if `any`, yet `/sync` records `cross_site_mode='local_only'`. The recorded value can misrepresent the strategy that actually took effect.

patroni/dcs/__init__.py
```diff
@@ -1961,6 +1965,7 @@ def update_leader(self,
         :param last_lsn: absolute WAL LSN in bytes.
         :param slots: dictionary with permanent slots ``confirmed_flush_lsn``.
         :param failsafe: if defined dictionary passed to :meth:`~AbstractDCS.write_failsafe`.
+        :param site: the site to which the leader belongs.
```
CyberDem0n: There is no such parameter

patroni/postgresql/sync.py
```diff
@@ -389,7 +409,28 @@ def current_state(self, cluster: Cluster) -> _SyncState:
         sync_node_maxlag = global_config.maximum_lag_on_syncnode
 
         # Prefer members without nofailover tag. We are relying on the fact that sorts are guaranteed to be stable.
-        for replica in sorted(replica_list, key=lambda x: x.nofailover):
+        sorted_replicas = sorted(replica_list, key=lambda x: x.nofailover)
+        current_site_replicas: List[_Replica] = [r for r in sorted_replicas if r.site == self._postgresql.site]
+        remote_replicas: List[_Replica] = [r for r in sorted_replicas if r.site != self._postgresql.site]
```
CyberDem0n: A replica with **no** `site` (`r.site is None`) is classified as *remote* by the `current_state` partition (`r.site != self._postgresql.site` is `True` for `None`), yet `pick_replicas_site_balanced` deliberately places site-less replicas **last**.

So the same "unknown site" member is a first-class *remote* candidate under `remote_only`/`prefer_remote` but a last-resort under `balanced`. Concretely, under `remote_only` a node that
is simply *missing* its `site:` config silently becomes an eligible synchronous standby.

The docs describe `remote_only` as "only replicas in a site *different from the primary*" - a member with *no* site arguably has no site relationship at all, so the current classification is easy to misread and inconsistent across modes.


**Suggestion:**
Pick one policy for site-less members and apply it uniformly - e.g. exclude them from the strict site modes:
```python
current_site_replicas = [r for r in sorted_replicas if r.site == self._postgresql.site]
remote_replicas = [r for r in sorted_replicas
                   if r.site is not None and r.site != self._postgresql.site]
```
or, if admitting them as last-resort (matching `balanced`) is intended, keep the current partition but state it explicitly in the per-mode list in docs/replication_modes.rst.

patroni/postgresql/sync.py
```diff
@@ -389,7 +409,28 @@ def current_state(self, cluster: Cluster) -> _SyncState:
...
+        elif cross_site_mode in (SyncCrossSiteMode.REMOTE_ONLY, SyncCrossSiteMode.PREFER_REMOTE):
+            selection_order = remote_replicas
+            if cross_site_mode == SyncCrossSiteMode.PREFER_REMOTE:
+                additional_replicas = current_site_replicas
+        else:  # SyncCrossSiteMode.LOCAL_ONLY, SyncCrossSiteMode.PREFER_LOCAL
+            selection_order = current_site_replicas
+            if cross_site_mode == SyncCrossSiteMode.PREFER_LOCAL:
+                additional_replicas = remote_replicas
+
+        additional_count = max(0, sync_node_count - len(selection_order))
```
CyberDem0n, line 435: `additional_count = max(0, sync_node_count - len(selection_order))` is computed from the *count* of primary-set candidates, before the loop applies the `maximum_lag_on_syncnode` filter.
In priority mode with `prefer_local`, if the local set has exactly `synchronous_node_count` members but one is filtered out by maxlag, `additional_count` is `0`, so no remote replica backfills the freed slot and the cluster ends up with fewer synchronous standbys than requested even though a qualifying remote existed.

**Suggestion:**
Compute the padding after the maxlag filter, or pad more generously and rely on the loop's existing `break` once the count is reached.

patroni/postgresql/sync.py
```diff
@@ -358,6 +362,22 @@ def _process_replica_readiness(self, cluster: Cluster, replica_list: _ReplicaLis
...
+    @staticmethod
+    def pick_replicas_site_balanced(current_site: str, replicas: List[_Replica]) -> List[_Replica]:
+        site_replicas: defaultdict[Optional[str], List[_Replica]] = defaultdict(list)
+        for replica in replicas:
+            site_replicas[replica.site or ''].append(replica)
+
+        # Create selection order: pick one from each remote site, then local, then repeat.
+        # Nodes with site not defined are added at the end.
+        # Ensure consistent sites order with sorting.
+        result: List[_Replica] = []
```
CyberDem0n:
```suggestion
```

patroni/validator.py
```diff
@@ -972,6 +973,10 @@ def validate_watchdog_mode(value: Any) -> None:
     assert_(value in (False, "off", "automatic", "required"))
 
 
+def validate_sync_cross_site(value: Any) -> None:
+    assert_(value in [a.value for a in SyncCrossSiteMode.__members__.values()])
```
CyberDem0n: Should we make it case insensitive?

docs/ha_multi_dc.rst
```diff
@@ -26,29 +26,189 @@ Regarding postgres, we must deploy at least 2 nodes, in different DC. Then you h
...
+
+- ``balanced``: replicas are picked round-robin across sites. Patroni builds a per-site list of eligible replicas (sorted deterministically for stability), then interleaves them: one replica from each remote site, then one from the local site, then back to the first remote site, and so on. Replicas without a configured site are added at the end of the list. ``balanced`` is intended for clusters that span three or more sites and want each site to contribute proportionally to the synchronous set.
+
+
+Interaction with ``synchronous_node_count`` and ``sync_priority``
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+The replica selection pipeline runs in five strict steps on every HA loop iteration:
+
+1. **Build the eligible set.** Patroni reads ``pg_stat_replication`` on the primary, joins each row with the corresponding member from DCS, and drops cascading replicas, ``nosync`` members, and the primary itself.
```
CyberDem0n, line 154: drops primary itself - sounds scary

docs/replication_modes.rst
```diff
@@ -220,3 +220,25 @@ If the primary (``node0``) failed, in the above example two of the ``node1``, ``
...
+- ``prefer_local``: local replicas are picked first; remote replicas fill any remaining slots. Suited to deployments that want a low-latency path under normal conditions but still meet ``synchronous_node_count`` after a local failure.
+- ``prefer_remote``: remote replicas are picked first; local replicas fill any remaining slots. The mirror of ``prefer_local`` — suited to deployments where durability is prioritised over latency but local replicas should still be used as fallback during a multi-site outage.
+- ``balanced``: replicas are interleaved round-robin across sites. Intended for clusters that span three or more sites and want each site to contribute proportionally to the synchronous set — one replica from each remote site first, then a local replica, repeating.
+
+If the *primary itself* has no ``site`` attribute configured, the cross-site logic is bypassed entirely and the selection falls back to the plain ``any`` ordering regardless of the configured mode.
```
CyberDem0n, line 242: Hmm. Isn't it an exact copy of what we have in ha_multi_dc.rst?

patroni/ha.py
```diff
@@ -891,12 +891,18 @@ def _handle_synchronous_strict_mode(self, dcs_state: SyncState, replication_stat
...
+            msg = site_msg = ''
             if voters != replication_state.sync or \
                     numsync != replication_state.numsync or \
                     sync_type != replication_state.sync_type:
                 self.state_handler.sync_handler.set_synchronous_standby_names(voters, numsync)
+            elif current_cross_site_mode != dcs_state.cross_site_mode and \
+                    current_cross_site_mode in (SyncCrossSiteMode.LOCAL_ONLY, SyncCrossSiteMode.REMOTE_ONLY):
+                self.dcs.write_sync_state(dcs_state.leader, CaseInsensitiveSet(), quorum, version=dcs_state.version)
+                self.state_handler.sync_handler.set_synchronous_standby_names(CaseInsensitiveSet(), numsync)
+                site_msg = 'that satisfy current synchronous_cross_site configuration '
```
CyberDem0n: I am not entirely sure about correctness of this branch.
Why do we set ssn and /sync key to empty sets here?

docs/ha_multi_dc.rst
```diff
@@ -26,29 +26,189 @@ Regarding postgres, we must deploy at least 2 nodes, in different DC. Then you h
...
+
+- ``prefer_local``: local replicas are preferred, but if there are not enough healthy local replicas to satisfy ``synchronous_node_count``, the remaining slots are filled with the highest-priority remote replicas. Useful when you want to minimise latency under normal conditions but still guarantee that ``synchronous_node_count`` synchronous standbys are available even after a local failure.
+
+- ``prefer_remote``: the mirror of ``prefer_local`` — remote replicas are preferred, but local replicas are used to fill any shortage. This is the right choice when durability is prioritised over latency but you still want to fall back to local replicas during a multi-site outage.
+
+- ``balanced``: replicas are picked round-robin across sites. Patroni builds a per-site list of eligible replicas (sorted deterministically for stability), then interleaves them: one replica from each remote site, then one from the local site, then back to the first remote site, and so on. Replicas without a configured site are added at the end of the list. ``balanced`` is intended for clusters that span three or more sites and want each site to contribute proportionally to the synchronous set.
```
CyberDem0n: I think it would be worth mentioning that balanced doesn't work well with synchronous_mode=quorum.

patroni/global_config.py
```diff
@@ -81,14 +81,24 @@ def from_cluster(self, cluster: Optional['Cluster']) -> 'GlobalConfig':
...
         :returns: configuration value or ``None`` if it is missing.
         """
-        return self.__config.get(name)
+        return self.__config.get(name, default)
+
+    @property
+    def sync_cross_site_mode(self) -> SyncCrossSiteMode:
+        """Currently configured value of ``synchronous_cross_site`` from the global configuration."""
+        try:
+            val = SyncCrossSiteMode((self.get('synchronous_cross_site', 'any')).lower())
```
CyberDem0n:
```suggestion
            val = SyncCrossSiteMode(str(self.get('synchronous_cross_site', 'any')).lower())
```

PR #3702, opened 2026-08-24
Add support for pg_hosts.conf

CyberDem0n:
PostgreSQL 19 introduces `pg_hosts.conf`. Manage it the same way as `pg_ident.conf` via the new `postgresql.pg_hosts` configuration, including role-specific overrides (pg_hosts_primary/pg_hosts_replica/standby_leader).

Patroni generates and reloads pg_hosts.conf only when running PostgreSQL 19+, so older versions are unaffected.


patroni/config_generator.py
```diff
@@ -462,6 +464,17 @@ def _set_conf_files(self) -> None:
...
+        if 'hosts_file' in self.config['postgresql']['parameters']:
+            default_hosts_path = os.path.join(self.config['postgresql']['data_dir'], 'pg_hosts.conf')
+            if self.config['postgresql']['parameters']['hosts_file'] == default_hosts_path:
+                try:
+                    self.config['postgresql']['pg_hosts'] = [i for i in read_stripped(default_hosts_path)
+                                                             if i and not i.startswith('#')]
+                except OSError as err:
+                    raise PatroniException(f'Failed to read pg_hosts.conf: {err}')
+                if not self.config['postgresql']['pg_hosts']:
+                    del self.config['postgresql']['pg_hosts']
```
hughcapet: maybe it is time to deduplicate this code for all auth files?

hughcapet (approved):

Issue #1636, opened 2020-08-03
basic auth in REST API does not work?


victor-sudakov:
I deliberately configured restapi.authentication.username and restapi.authentication.password to different random values on each node of the Patroni cluster, restarted Patroni, but the cluster for some reason still works. I expected some kind of authentication error and inability for the misconfigured nodes to form a cluster.

Do I need to enable something else to enable the basic auth (without certificates) between Patroni nodes? I don't expect it to be very secure but I expect it to protect the cluster from members joined in error.

CyberDem0n:
Basic auth is only checked for unsafe endpoints (which are serving non GET requests).
During normal operation the REST API is not even used across nodes, only during the leader race.
Unsafe endpoints are called only by patroninctl.

victor-sudakov:
So the best protection from a rogue node are "scope" and "namespace", not the username and password?

CyberDem0n:
What problem are you trying to solve?

victor-sudakov:
> What problem are you trying to solve?

I would like to provide some basic protection from untrusted users communicating to the Patroni cluster via its REST API (sending commands, queries, adding or deleting nodes etc).

CyberDem0n:
Perfect. Just configure basic auth credentials. As I told it will protect all non `GET` endpoints.

victor-sudakov:
> Perfect. Just configure basic auth credentials. As I told it will protect all non `GET` endpoints.
How can I verify that? For the present, I see that when I join a new Patroni  node with a deliberately incorrect 
restapi.authentication.username and restapi.authentication.password, it just joins successfully.

CyberDem0n:
> How can I verify that?

1. Read the [documentation](https://patroni.readthedocs.io/en/latest/rest_api.html).
2. Find any endpoint accepting non GET requests.
3. Try to execute this request without (or with the wrong) authorization header.

victor-sudakov:
> > How can I verify that?
> 
>     1. Read the [documentation](https://patroni.readthedocs.io/en/latest/rest_api.html).
> 
>     2. Find any endpoint accepting non GET requests.
> 
>     3. Try to execute this request without (or with the wrong) authorization header.

True, some non-GET commands (like `patronictl switchover` or `patronictl restart`) are prevented with the status code=401, (not authenticated).

But for example `patronictl edit-config` is not prevented, you can edit all you want including pg_hba, without knowing the right username/password. And "PATCH /config" is not a GET request.

[...]

PR #3704, opened 2026-08-26
Add restapi.authentication.mode for staged auth rollout

chaitanyabodlapati:
## What

Adds `restapi.authentication.mode` with three values — disabled,
permissive, strict — to allow enforcing REST API authentication on an
existing cluster without a flag-day cutover.

## Why

Today enabling REST API auth is effectively all-or-nothing across a
cluster: while some nodes require credentials and others don't, peer
calls during the leader race and switchovers can fail. There is no
supported migration window. (See #1636 for a case where per-node
credential differences did not behave as operators expected.)

permissive mode is that migration window: it accepts both
authenticated and unauthenticated requests (logging the latter) so
nodes can be rolled one at a time, then moved to strict once all are
done.

## Design

- disabled (default): current behavior, no enforcement. Existing
  configs are completely unaffected.
- permissive: accept both; warn (throttled) on unauthenticated access.
- strict: require valid credentials.

Auth state is an immutable snapshot swapped atomically on reload.
Outbound peer requests send credentials derived from
restapi.authentication so strict nodes authenticate to each other.
permissive/strict fail closed without complete credentials.

## Backwards compatibility

Default is disabled and the existing `auth` string keeps working, so
upgrading with no config change preserves current behavior.

## Testing

Unit tests cover all three modes, reload atomicity, the
allowlist/client-certificate non-bypass, outbound auth precedence,
and the throttled permissive warning. Validated on a real 3-node
cluster: rolling disabled -> permissive -> strict with no quorum loss,
switchover under strict, and failover recovery.

## Scope

This is the authentication mode only. A companion TLS mode
(permissive/strict for HTTPS migration) is planned as a separate PR.

### DIFF
--- a/docs/security.rst
+++ b/docs/security.rst
@@ -28,6 +28,8 @@ From the point of view of security, REST API contains safe (``GET`` requests, on
 
 The unsafe endpoints can be protected with HTTP basic-auth by setting the ``restapi.authentication.username`` and ``restapi.authentication.password`` parameters. There is no way to protect the safe endpoints without enabling TLS.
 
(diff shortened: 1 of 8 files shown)


CyberDem0n (commented):
It seems that you are trying to solve non-existing problem.
Nothing stops users from adding valid credentials to the `ctl` section (with Patroni reload) before doing so in `restapi` section.

PR #3706, opened 2026-08-28
Time out stalled REST API connections

### LINKED ISSUE #3705: Unresponsive HTTP API because of unstable clients like HAProxy
### What happened?

With TLS enabled, the REST API intermittently stops answering. TCP connects still succeed, so a load balancer (HAProxy in our case) sees the connection established, then times out waiting for a response and marks the backend down. Patroni itself stays healthy, keeps its leader lock, and logs nothing above WARNING.

A single client that is able to open a TCP connection can permanently consume one of the REST API worker slots. With the default `restapi.thread_pool_size` of 5, five such connections disable the REST API on that node until Patroni is restarted.

Since #3526 (`617a0330`) the REST API is served by a fixed thread pool instead of a thread per connection. `serve_forever()` permanently holds one worker, so the default `thread_pool_size` of 5 leaves exactly 5 request handlers, and no timeout is set: `RestApiHandler.timeout` and `BaseServer.timeout` are both `None`, and `socket.setdefaulttimeout()` is not called.

The listening socket is wrapped with `do_handshake_on_connect=False`, so the TLS handshake runs inside a pool worker on a blocking socket. A client that opens a TCP connection to the TLS port, holds it open, and never sends a ClientHello blocks that worker forever in `do_handshake()`. TCP keepalive reaps peers that are gone, not silent ones, so the worker is never released. A client that completes the handshake (or connects over plain HTTP) and then sends nothing does the same thing one step later, inside the request handler.

Once all 5 handler workers are held this way, further connections are accepted at the TCP layer but their work is queued and never picked up.

4.1.0 is unaffected because it used a thread per connection. Stalling one connection could not starve another. The stall is a property of the bounded pool introduced in #3526, not of the OS. Debian 12 alone is on OpenSSL 3.0, whose provider/property-fetch lock contention (fixed in 3.1/3.2; Debian 11 is on 1.1.1, Debian 13 on 3.5) plausibly amplifies per-handshake latency and makes saturation easier to hit under real load, but it is an amplifier, not the cause.

With `restapi.certfile` set and the default `thread_pool_size`, open five connections to the TLS port that never send a ClientHello, then try a normal request:

```bash
for i in $(seq 1 5); do (exec 3<>/dev/tcp/127.0.0.1/8443; sleep 3600) & done
curl -sk https://127.0.0.1:8443/liveness --max-time 5   # connects, then times out
```

`py-spy dump` shows the `RestAPI-*` threads parked in `do_handshake`. The same commands against 4.1.0 have no effect.

A client that opens a connection but never makes progress must not be able to hold the REST API worker indefinitely. The handshake and the request handling should each be bounded by a timeout so a stalled or hostile peer is dropped and the worker returned to the pool.

- Patroni: 4.1.1+
- PostgreSQL: any
- DCS: any
- TLS: `restapi.certfile` set
- Debian: any
- Baremetal and VM, with free memory, so not memory pressure

```yaml
scope: cluster_name
namespace: /patroni/
name: node-1

restapi:
  listen: 0.0.0.0:8443
  connect_address: xxx.xxx.xxx.xxx:8443
  certfile: /etc/patroni/cert.pem
  keyfile: /etc/patroni/key.pem
  cafile: /etc/patroni/ca.pem
  authentication:
    username: xxx
    password: xxx

ctl:
  insecure: true

log:
  level: INFO

consul:
  host: xxx.xxx.xxx.xxx:xxx
  port: xxx
  scheme: http
  service_check_interval: 5s
  consistency: consistent
  token: xxx


bootstrap:
  dcs:
    ttl: 30
    loop_wait: 10
    retry_timeout: 30
    master_start_timeout: 300
    synchronous_mode: true
    synchronous_node_count: 5
    failsafe_mode: true
    postgresql:
      use_pg_rewind: true
      use_slots: true
      parameters:
        wal_level: logical
        wal_keep_size: 512MB
        hot_standby: "on"
        max_connections: 2000
        max_wal_senders: 50
        max_replication_slots: 200
        max_locks_per_transaction: 64
        max_prepared_transactions: 1000
        max_worker_processes: 200
        track_commit_timestamp: "on"
        wal_log_hints: "on"
  method: pg_createcluster
  pg_createcluster:
    command: /path/to/pg_createcluster_script --encoding=UTF8 --version=14
  users:
    monitoring:
      password: xxx

postgresql:
  listen: 0.0.0.0:5432
  connect_address: xxx.xxx.xxx.xxx:5432
  data_dir: /data/postgresql/main
  bin_dir: /usr/lib/postgresql/14/bin
  config_dir: /etc/postgresql/14/main
  pgpass: /var/lib/postgresql/postgres.pgpass
  authentication:
    superuser:
      username: postgres
      password: xxx
    replication:
      username: repl
      password: xxx
  parameters:
    unix_socket_directories: '/var/run/postgresql'
    stats_temp_directory: pg_stat_tmp
    syslog_facility: LOCAL6
    syslog_ident: postgres
    logging_collector: "on"
    log_destination: syslog, csvlog
    log_directory: /var/log/postgresql
    log_filename: postgresql.log
    log_min_duration_statement: 250
    log_line_prefix: '%m postgres[%p]: [%c-%l] db=%d,user=%u,app=%a,client=%h '
    log_statement: ddl
    log_timezone: UTC
    log_checkpoints: "on"
    log_autovacuum_min_duration: 0
    log_lock_waits: "on"
    log_connections: "on"
    log_disconnections: "on"
    autovacuum_max_workers: 10
    autovacuum_work_mem: 51MB
    vacuum_cost_delay: 10
    DateStyle: "iso, mdy"
    TimeZone: UTC
    lc_messages: en_US.UTF-8
    lc_monetary: en_US.UTF-8
    lc_numeric: en_US.UTF-8
    lc_time: en_US.UTF-8
    default_text_search_config: pg_catalog.english
    shared_preload_libraries: "pg_stat_statements"
    pg_stat_statements.max: 10000
    pg_stat_statements.track: all
    random_page_cost: 4
    archive_mode: "on"
    archive_command: "xxx"
    archive_timeout: 1min
    min_wal_size: 512MB
    max_wal_size: 1GB
    wal_compression: on
    shared_buffers: 1505MB
    effective_cache_size: 4GB
    work_mem: 16MB
    maintenance_work_mem: 512MB
    dynamic_shared_memory_type: posix
    ssl: "on"
    ssl_cert_file: /etc/patroni/cert.pem
    ssl_key_file: /etc/patroni/key.pem
    ssl_ca_file: /etc/patroni/ca.pem
    enable_partitionwise_aggregate: "on"
    jit: "off"
    temp_file_limit: 10GB
    statement_timeout: 0
    max_parallel_workers: 4
    max_parallel_workers_per_gather: 2
    effective_io_concurrency: 256
    synchronous_commit: on
    restore_command: "xxx"
    recovery_target_timeline: latest
    password_encryption: scram-sha-256
    hot_standby_feedback: on
    max_logical_replication_workers: 30
    wal_sender_timeout: 0
    log_error_verbosity: VERBOSE
  pg_hba: [xxx]
  create_replica_method:
    - pg_clonecluster
  pg_clonecluster:
    command: /path/to/pg_clonecluster_script --version=14

tags:
  nofailover: false
  noloadbalance: false
  clonefrom: false
  nosync: false
```

### patronictl show-config

```yaml
failsafe_mode: true
loop_wait: 1
master_start_timeout: 300
postgresql:
  parameters:
    hot_standby: 'on'
    max_connections: 2000
    max_locks_per_transaction: 64
    max_prepared_transactions: 1000
    max_replication_slots: 50
    max_wal_senders: 50
    max_worker_processes: 200
    track_commit_timestamp: 'on'
    wal_keep_size: 512MB
    wal_level: logical
    wal_log_hints: 'on'
  use_pg_rewind: true
  use_slots: true
retry_timeout: 14
slots: {}
synchronous_mode: true
synchronous_node_count: 5
ttl: 30
```

### Patroni log files

```shell
N/A
```

### PostgreSQL log files

```shell
N/A
```

- [x] Yes

_No response_

jouir:
The REST API is served by a fixed thread pool, and no timeout is set anywhere. A client that connects to the TLS port and never sends a ClientHello blocks a pool worker forever in do_handshake(), and a client that connects and never sends a request does the same in the handler. With the default restapi.thread_pool_size of 5, five such connections make the REST API stop answering until Patroni is restarted, while Patroni itself stays healthy and logs nothing above WARNING.

Add two settings that bound how long a single connection may hold a worker without making progress. restapi.handshake_timeout (default 5) caps the TLS handshake, restapi.request_timeout (default 10) caps sending the request and reading the response, over HTTPS and plain HTTP alike. Both are applied on reload and never go below one second.

These are socket timeouts, so they bound individual reads and writes on the client connection, not the time spent in the handler. A slow query does not trigger them, and neither timeout logs above DEBUG.

Introduced in #3526.
Fixes #3705.

### DIFF
--- a/patroni/config.py
+++ b/patroni/config.py
@@ -549,7 +549,8 @@ class Config(object):
         _set_section_values('restapi', ['listen', 'connect_address', 'certfile', 'keyfile', 'keyfile_password',
                                         'cafile', 'ciphers', 'verify_client', 'http_extra_headers',
                                         'https_extra_headers', 'allowlist', 'allowlist_include_members',
-                                        'request_queue_size', 'server_tokens'])
+                                        'request_queue_size', 'handshake_timeout', 'request_timeout',
+                                        'server_tokens'])
         _set_section_values('ctl', ['insecure', 'cacert', 'certfile', 'keyfile', 'keyfile_password'])
         _set_section_values('postgresql', ['listen', 'connect_address', 'proxy_address',
                                            'config_dir', 'data_dir', 'pgpass', 'bin_dir'])
@@ -574,7 +575,8 @@ class Config(object):
                 if value is not None:
                     ret[first][second] = value
 
-        for first, params in (('restapi', ('request_queue_size', 'thread_pool_size')),
+        for first, params in (('restapi', ('request_queue_size', 'thread_pool_size',
+                                           'handshake_timeout', 'request_timeout')),
                               ('log', ('max_queue_size', 'file_size', 'file_num', 'mode'))):
             for second in params:
                 value = ret.get(first, {}).pop(second, None)
--- a/docs/ENVIRONMENT.rst
+++ b/docs/ENVIRONMENT.rst
@@ -222,6 +222,8 @@ REST API
 -  **PATRONI\_RESTAPI\_HTTP\_EXTRA\_HEADERS**: (optional) HTTP headers let the REST API server pass additional information with an HTTP response.
 -  **PATRONI\_RESTAPI\_HTTPS\_EXTRA\_HEADERS**: (optional) HTTPS headers let the REST API server pass additional information with an HTTP response when TLS is enabled. This will also pass additional information set in ``http_extra_headers``.
 -  **PATRONI\_RESTAPI\_REQUEST\_QUEUE\_SIZE**: (optional): Sets request queue size for TCP socket used by Patroni REST API.  Once the queue is full, further requests get a "Connection denied" error. The default value is 5.
+-  **PATRONI\_RESTAPI\_HANDSHAKE\_TIMEOUT**: (optional): Maximum time in seconds a single client is given to complete the TLS handshake. Connections that do not complete it in time are closed. It only applies when ``PATRONI_RESTAPI_CERTFILE`` is set. The default value is 2.
+-  **PATRONI\_RESTAPI\_REQUEST\_TIMEOUT**: (optional): Maximum time in seconds a single client is given to send its request, and to read the response. Connections that stay silent for longer are closed. The default value is 5.
 -  **PATRONI\_RESTAPI\_SERVER\_TOKENS**: (optional) Configures the value of the ``Server`` HTTP header.  ``Original`` (default) will expose the original behaviour and display the BaseHTTP and Python versions, e.g. ``BaseHTTP/0.6 Python/3.12.3``. ``Minimal``: The header will contain only the Patroni version, e.g. ``Patroni/4.0.0``. ``ProductOnly``: The header will contain only the product name, e.g. ``Patroni``.
 
 .. warning::
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -47,6 +47,8 @@ In addition, you may tune the following Patroni configuration parameters:
 - ``thread_stack_size`` - stack size used for threads started by Patroni. Lowering this value reduces memory usage of the Patroni process. The default value set by Patroni is ``512kB``. Increase ``thread_stack_size`` if Patroni experience stack-related crashes; otherwise the default value is sufficient.
 - ``thread_pool_size`` - size of the thread pool used by Patroni for asynchronous tasks and REST API communication with other members during leader race or failsafe checks. The default value is ``5``, which is sufficient for three-node clusters.
 - ``restapi.thread_pool_size`` - size of the thread pool used to process REST API requests. The default value is ``5``, allowing up to five parallel REST API requests. Note that requests involving SQL queries are effectively serialized because a single database connection is used, so increasing this value typically provides no benefit.
+- ``restapi.handshake_timeout`` - maximum time in seconds a single client is given to complete the TLS handshake. It bounds how long a connection may hold a thread of ``restapi.thread_pool_size`` before sending any data. The default value is ``2``.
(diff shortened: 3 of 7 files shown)


CyberDem0n:
Nice catch, @jouir.
However I think that default timeouts are too high.
I would suggest default handshake timeout 2 seconds, and default request timeout 5 second.

patroni/config.py
```diff
@@ -574,7 +574,8 @@ def _set_section_values(section: str, params: List[str]) -> None:
                 if value is not None:
                     ret[first][second] = value
 
-        for first, params in (('restapi', ('request_queue_size', 'thread_pool_size')),
+        for first, params in (('restapi', ('request_queue_size', 'thread_pool_size',
+                                           'handshake_timeout', 'request_timeout')),
```
CyberDem0n, line 579: You also need to add new parameters to:
```python
        _set_section_values('restapi', ['listen', 'connect_address', 'certfile', 'keyfile', 'keyfile_password',
                                        'cafile', 'ciphers', 'verify_client', 'http_extra_headers',
                                        'https_extra_headers', 'allowlist', 'allowlist_include_members',
                                        'request_queue_size', 'server_tokens'])
```

Line 549-552

jouir: Updated

jouir:
> Nice catch, @jouir. However I think that default timeouts are too high. I would suggest default handshake timeout 2 seconds, and default request timeout 5 second.

I have reduced the timeouts as suggested.

hughcapet (approved):

PR #3708, opened 2026-09-03
Fix premature 503 in poll_failover_result() during transient unlocked state

CyberDem0n:
The problem may happen with ZooKeeper, because it cleans up all ephemeral ZNodes when client disconnects from the server (even temporary).


hughcapet:
i bit more context on when it can happen would help:(

hughcapet (approved):

PR #3709, opened 2026-09-03
Fix flaky priority_failover behave test by waiting for equal WAL position

CyberDem0n:
and refactor site_awareness behave test


hughcapet:
can we maybe replace the one used in site_awareness with this one?

hughcapet (approved):

PR #3714, opened 2026-09-03
Add validator for standby cluster's port(s)

hughcapet:


patroni/validator.py
```diff
@@ -166,6 +166,30 @@ def validate_host_port(host_port: str, listen: bool = False, multiple_hosts: boo
...
+    :raises:
+        :class:`~patroni.exceptions.ConfigParseError`:
+            * If *port* is not an :class:`int`; or
+            * If *port* is not a :class:`str` with a comma-separated list of ints.
+    """
+    if isinstance(port, int):
+        return True
+    if isinstance(port, str):
+        try:
+            [int(p) for p in port.split(",")]
```
CyberDem0n: We have to check min and max

patroni/validator.py
```diff
@@ -166,6 +166,35 @@ def validate_host_port(host_port: str, listen: bool = False, multiple_hosts: boo
...
+        return True
+    if isinstance(port, str):
+        try:
+            ports = [int(p) for p in port.split(",")]
+        except ValueError:
+            raise ConfigParseError("contains a wrong value")
+        else:
+            if not all(1 <= p <= 65535 for p in ports):
+                raise ConfigParseError("contains a wrong value")
+            return True
```
CyberDem0n, line 184:
```suggestion
    if isinstance(port, int):
        if 1 <= port <= 65535:
            return True
    elif isinstance(port, str):
        try:
            if all(1 <= int(p) <= 65535 for p in port.split(",")):
                return True
        except Exception:
            pass
```

CyberDem0n (approved):

PR #3725, opened 2026-09-21
Handle overflowing values in parse_int

Cianidos:
## Summary

- return `None` when a floating-point value overflows during integer parsing
- cover overflow both before and during base-unit conversion

`strtod()` accepts values such as `1e999` and returns infinity. Calling `round()` on that value raises `OverflowError`, although `parse_int()` promises to return `None` when a value cannot be parsed as an integer.

## Tests

- `flake8 --config tox.ini patroni/utils.py tests/test_utils.py`


Cianidos:
Addressed CI isort failure and pushed corrected import ordering. Updated workflow now awaits maintainer approval because it runs from a fork.

PR #3726, opened 2026-09-21
Handle arithmetic errors in scheduled timestamps

Cianidos:
## Summary

- treat arithmetic failures from `dateutil.parser.parse()` as invalid scheduled timestamps
- cover both `OverflowError` and `decimal.InvalidOperation`

Malformed or extreme timestamp values can make `python-dateutil` raise arithmetic exceptions instead of `ValueError` or `TypeError`. Letting these exceptions escape aborts REST requests for scheduled restarts or switchovers rather than returning the documented `422` response.

## Tests


tests/test_api.py
```diff
@@ -544,6 +545,13 @@ def make_request(request=None, **kwargs):
         # incorrect schedule
         request = make_request(schedule='2016-08-42 12:45TZ+1', role=PostgresqlRole.PRIMARY)
         MockRestApiServer(RestApiHandler, request)
+        for error in (OverflowError, decimal.InvalidOperation):
```
hughcapet: please use the actual values that produce these errors

e.g., `request = make_request(schedule='100000000000000000000-01-01', role=PostgresqlRole.PRIMARY)`

Cianidos: Updated the regression test to exercise two actual values: one raises `OverflowError`, and the other raises `decimal.InvalidOperation`. Both now verify the documented 422 response without mocking `dateutil.parser.parse()`.

Cianidos:
Addressed review feedback with actual inputs for both arithmetic exceptions. Updated workflow now awaits maintainer approval because it runs from a fork.

CyberDem0n (approved):

PR #3728, opened 2026-09-21
Validate the restart PostgreSQL version type

Cianidos:
## Summary

- reject non-string `postgres_version` restart filters with HTTP 400
- add regression coverage for JSON object input

`postgres_version_to_int()` expects a string and calls `.split()`. A valid JSON request can provide another type, causing an uncaught `AttributeError` and aborting the REST request instead of reporting invalid input.

## Tests


Cianidos:
Reopening as a regular pull request; GitHub is refusing draft-state conversion for this cross-fork PR.

PR #3729, opened 2026-09-21
Validate the restart PostgreSQL version type

Cianidos:
## Summary

- reject non-string `postgres_version` restart filters with HTTP 400
- add regression coverage for JSON object input

`postgres_version_to_int()` expects a string and calls `.split()`. A valid JSON request can provide another type, causing an uncaught `AttributeError` and aborting the REST request instead of reporting invalid input.

## Tests


Cianidos:
GitHub currently prevents this cross-fork draft from being marked ready by its author. Maintainers: please mark it ready for review when convenient; implementation and focused regression tests are complete.

CyberDem0n (commented):
I don't like changing existing API (making it stricter) for no good reason.

Cianidos:
Thanks for the review. I agree compatibility is more important here: accepting non-string JSON values was not an intentional string-only contract, and this guard would make that API stricter. I will close this PR rather than change existing behavior.

